PR #12844, opened 2018-12-19
Add json to csv transform example

pamod:
## Purpose

Add sample to illustrate json to csv conversion through usage of CSV channel.


pamod:
This changes will be addressed with #12846

PR #12845, opened 2018-12-19
 Add socket timeout configuration in gRPC client and server

### LINKED ISSUE #11276: [gRPC] Enable support to configure socket timeout in service/client endpoint
**Description:**
Currently there is no way of defining timeout value for gRPC service and client when initializing. We need to give support to define timeout as endpoint configuration.

daneshk:
## Purpose
Resolves #11276


stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/ClientCall.java
```diff
@@ -185,7 +186,8 @@ public void sendMessage(Message message) {
         try {
             InputStream resp = method.streamRequest(message);
             outboundMessage.sendMessage(resp);
-
+        } catch (StatusRuntimeException ex) {
```
wggihan: Why we need to catch this if we don't do anything? Since this is a RuntimeException shall we leave without handling it?

daneshk: Earlier for any exception we throw StatusRuntimeException and we need to avoid reconstructing exception if it is StatusRuntimeException.

PR #12846, opened 2018-12-19
Add json to csv transform example

pamod:
## Purpose

Add sample to illustrate json to csv conversion through usage of CSV channel.

### DIFF
--- a/examples/json-csv/json_csv.bal
+++ b/examples/json-csv/json_csv.bal
@@ -1,0 +1,72 @@
+import ballerina/io;
+import ballerina/log;
+
+// Converts a json into an array of strings
+// Returns the result as a tuple which contains the headers and fields
+function getFields(json rec) returns (string[], string[]) {
+    int count =0;
+    string [] headers = [];
+    string [] fields = [];
+    headers = rec.getKeys();
+    foreach var field in headers {
+        fields[count] = rec[field].toString();
+        count = count + 1;
+    }
+    return (headers, fields);
+}
+
+// Example function which writes json content to csv
+function writeCsv(json content, string path) returns error? {
+    io:WritableCSVChannel csvch = io:openWritableCsvFile(path);
+    int recIndex =0;
+    int recLen = content.length();
+    while (recIndex < recLen) {
+        (string [], string []) result = getFields(content[recIndex]);
+        var (headers, fields) = result;
+        if (recIndex == 0) {
+             var wResult1 = check csvch.write(headers);
+        }
+        var wResult2 = check csvch.write(fields);
+        recIndex = recIndex + 1;
+    }
+    return;
+}
+
+public function main() {
+    // Sample json which will be written
+    json sample = {
+            "employees": {
+                "employee": [
+                    {
+                        "id": "1",
+                        "firstName": "Tom",
+                        "lastName": "Cruise",
+                        "photo": "https://ballerina-team/profile/3737.jpg"
+                    },
+                    {
+                        "id": "2",
+                        "firstName": "Maria",
+                        "lastName": "Sharapova",
+                        "photo": "https://ballerina-team/profile/5676.jpg"
+                    },
+                    {
+                        "id": "3",
+                        "firstName": "James",
+                        "lastName": "Bond",
+                        "photo": "https://ballerina-team/profile/6776.jpg"
+                    }
+                ]
+            }};
+    // Writes json into a csv
+    string path = "./files/sample.csv";
+    // Specify the json array which should be transformed into csv
+    // Also provide the location the csv should be written
+    var result = writeCsv(sample.employees.employee, path);
+    if (result is error) {
+        log:printError("Error occurred while writing csv record :",
+                        err = result);
+    } else {
+        io:println("json record successfully transformed to a csv, file could" +
+                    " be found in " + path);
+    }
+}
--- a/examples/json-csv/json_csv.bal
+++ b/examples/json-csv/json_csv.bal
@@ -25,9 +25,9 @@ function writeCsv(json content, string path) returns error? {
         var (headers, fields) = result;
         if (recIndex == 0) {
             //We ignore the result as this would mean a nill return
-             _  = check csvch.write(headers);
+            check csvch.write(headers);
         }
-        _  = check csvch.write(fields);
+        check csvch.write(fields);
         recIndex = recIndex + 1;
     }
     return;
--- a/examples/json-csv/json_csv.bal
+++ b/examples/json-csv/json_csv.bal
@@ -4,7 +4,7 @@ import ballerina/log;
 // Converts a json into an array of strings
 // Returns the result as a tuple which contains the headers and fields
 function getFields(json rec) returns (string[], string[]) {
-    int count =0;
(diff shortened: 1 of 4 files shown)


examples/json-csv/json_csv.bal
```diff
@@ -0,0 +1,72 @@
...
+        fields[count] = rec[field].toString();
+        count = count + 1;
+    }
+    return (headers, fields);
+}
+
+// Example function which writes json content to csv
+function writeCsv(json content, string path) returns error? {
+    io:WritableCSVChannel csvch = io:openWritableCsvFile(path);
+    int recIndex =0;
```
riyafa: formatting issue

examples/json-csv/json_csv.bal
```diff
@@ -0,0 +1,72 @@
...
+    int recLen = content.length();
+    while (recIndex < recLen) {
+        (string [], string []) result = getFields(content[recIndex]);
+        var (headers, fields) = result;
+        if (recIndex == 0) {
+             var wResult1 = check csvch.write(headers);
```
riyafa: Shall we change the variable names since it is an example. I don't see where these variables are used later.

examples/json-csv/json_csv.bal
```diff
@@ -0,0 +1,72 @@
+import ballerina/io;
+import ballerina/log;
+
+// Converts a json into an array of strings
+// Returns the result as a tuple which contains the headers and fields
+function getFields(json rec) returns (string[], string[]) {
+    int count =0;
```
riyafa: Formatting issue. Shall we fix in all places?

examples/json-csv/json_csv.bal
pamod: write contains error|() we need to assign the result to a variable ( which is enforced at compile time) if there's an error this will be propagated with the check

riyafa: If it is not used you can ignore using `_`

examples/json-csv/json_csv.bal
```diff
@@ -0,0 +1,73 @@
+import ballerina/io;
+import ballerina/log;
+
+// Converts a json into an array of strings
```
MaryamZi:
```suggestion
// Converts a `JSON` value to an array of strings.
```

examples/json-csv/json_csv.bal
```diff
@@ -0,0 +1,73 @@
...
+    string [] fields = [];
+    headers = rec.getKeys();
+    foreach var field in headers {
+        fields[count] = rec[field].toString();
+        count = count + 1;
+    }
+    return (headers, fields);
+}
+
+// Example function which writes json content to csv
```
MaryamZi:
```suggestion
// Writes `JSON` content to CSV.
```

examples/json-csv/json_csv.bal
```diff
@@ -0,0 +1,73 @@
...
+             _  = check csvch.write(headers);
+        }
+        _  = check csvch.write(fields);
+        recIndex = recIndex + 1;
+    }
+    return;
+}
+
+public function main() {
+    // Sample json which will be written
```
MaryamZi:
```suggestion
    // Sample `JSON` which will be written.
```

examples/json-csv/json_csv.bal
```diff
@@ -0,0 +1,73 @@
...
+                    },
+                    {
+                        "id": "3",
+                        "firstName": "James",
+                        "lastName": "Bond",
+                        "photo": "https://ballerina-team/profile/6776.jpg"
+                    }
+                ]
+            }};
+    // Writes json into a csv
```
MaryamZi: Should this comment be here? This will be highlighted only for the `string path = "./files/sample.csv";` line right?

examples/json-csv/json_csv.bal
```diff
@@ -0,0 +1,73 @@
...
+    string path = "./files/sample.csv";
+    // Specify the json array which should be transformed into csv
+    // Also provide the location the csv should be written
+    var result = writeCsv(sample.employees.employee, path);
+    if (result is error) {
+        log:printError("Error occurred while writing csv record :",
+                        err = result);
+    } else {
+        io:println("json record successfully transformed to a csv, file could" +
+                    " be found in " + path);
```
MaryamZi:
```suggestion
                    " be found at " + path);
```

examples/json-csv/json_csv.bal
```diff
@@ -0,0 +1,73 @@
...
+                        "lastName": "Bond",
+                        "photo": "https://ballerina-team/profile/6776.jpg"
+                    }
+                ]
+            }};
+    // Writes json into a csv
+    string path = "./files/sample.csv";
+    // Specify the json array which should be transformed into csv
+    // Also provide the location the csv should be written
+    var result = writeCsv(sample.employees.employee, path);
```
MaryamZi: Shall we rephrase the comments to say something along the lines of
```
// Invoke the function to write the `JSON` content to CSV, specifying the `JSON` array to write and the location to write to.
```

WDYT?

examples/json-csv/json_csv.description
```diff
@@ -0,0 +1 @@
+// Sample demonstrates how json array could be transformed into csv
```
MaryamZi: How about something like
```suggestion
// This sample demonstrates how a `JSON` array can be transformed to CSV.
```

examples/json-csv/json_csv.bal
pamod: Some of the suggestions were addressed in #12858

examples/json-csv/json_csv.bal
pamod: This comment would be applicable for LoC below it.

examples/json-csv/json_csv.bal
pamod: Since this a directory path in would be grammatically correct :) some of the example usages of a similar case 

1. https://www.gog.com/forum/general/the_file_could_not_be_located_in_the_installers_folder
2. https://www.mathworks.com/matlabcentral/answers/401697-file-not-found-in-current-directory

examples/json-csv/json_csv.bal
pamod: In this case wouldn't invocation of the function be obvious ? restating it might not make sense i presume, intention was to highlight the type of params which should be provided when calling the function

examples/json-csv/json_csv.bal
MaryamZi: Yeah but on the site it will highlight just the single `string path = ....` line right?

examples/json-csv/json_csv.bal
MaryamZi: Yeah, felt a bit disconnected though. Shall we add periods at the end of the comments if we are keeping as is?

examples/json-csv/json_csv.bal
MaryamZi: But here it's not the directory right?
```
json record successfully transformed to a csv, file could be found in ./files/sample.csv
```
Was wondering since we have the path to the file itself (`/files/sample.csv`)?

PR #12856, opened 2018-12-19
Trigger onClose for caller->close

wggihan:
## Purpose
$title.

### DIFF
--- a/stdlib/socket/src/main/java/org/ballerinalang/stdlib/socket/endpoint/tcp/client/Close.java
+++ b/stdlib/socket/src/main/java/org/ballerinalang/stdlib/socket/endpoint/tcp/client/Close.java
@@ -25,8 +25,9 @@ import org.ballerinalang.model.values.BMap;
 import org.ballerinalang.model.values.BValue;
 import org.ballerinalang.natives.annotations.BallerinaFunction;
 import org.ballerinalang.natives.annotations.Receiver;
-import org.ballerinalang.stdlib.socket.SocketConstants;
+import org.ballerinalang.stdlib.socket.tcp.SelectorDispatcher;
 import org.ballerinalang.stdlib.socket.tcp.SelectorManager;
+import org.ballerinalang.stdlib.socket.tcp.SocketService;
 import org.ballerinalang.stdlib.socket.tcp.SocketUtils;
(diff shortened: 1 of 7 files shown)


stdlib/socket/src/main/java/org/ballerinalang/stdlib/socket/endpoint/tcp/client/Close.java
```diff
@@ -68,6 +71,12 @@ public void execute(Context context) {
             if (client != null && Boolean.getBoolean(client.toString())) {
                 SelectorManager.getInstance().stop();
             }
+            SocketService socketService = (SocketService) clientEndpoint.getNativeData(SOCKET_SERVICE);
+            // Check whether client has the attached callback service.
+            // ResourceMap shouldn't be null if callback service attached.
+            if (socketService != null && socketService.getResources() != null) {
+                SelectorDispatcher.invokeOnClose(socketService);
```
pamod: If it's null should we handle the error i.e log ?

wggihan: No. That's not an error situation. Having a callback service for a client is an optional.

stdlib/socket/src/main/java/org/ballerinalang/stdlib/socket/tcp/SelectorDispatcher.java
```diff
@@ -141,28 +139,31 @@ static void invokeOnClose(SocketService socketService) {
...
-        BMap<String, BValue> caller = SocketUtils.createClient(programFile, client);
+    private static BValue[] getOnErrorResourceSignature(SocketService socketService, Resource errorResource,
+            String msg) {
+        ProgramFile programFile = errorResource.getResourceInfo().getPackageInfo().getProgramFile();
+        BMap<String, BValue> caller = SocketUtils.createClient(programFile, socketService);
         BError error = SocketUtils.createSocketError(programFile, msg);
         return new BValue[] { caller, error };
     }
+
+    private static BValue[] getReadyReadyResourceSignature(SocketService socketService, ByteBuffer buffer,
```
chamil321: A typo ? getReadReadyResourceSignature*

PR #12858, opened 2018-12-19
Fix doc comments

pamod:
## Purpose
 
Fix some of the document comments in json-csv example

### DIFF
--- a/examples/json-csv/json_csv.bal
+++ b/examples/json-csv/json_csv.bal
@@ -1,7 +1,7 @@
 import ballerina/io;
 import ballerina/log;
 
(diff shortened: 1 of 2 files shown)


examples/json-csv/json_csv.bal
```diff
@@ -1,7 +1,7 @@
 import ballerina/io;
 import ballerina/log;
 
-// Converts a json into an array of strings
+// Converts a `JSON` value to an array of strings.
```
Manuri: Should we keep json in simple? `json` is a keyword in Ballerina not JSON.

PR #12860, opened 2018-12-19
Fix comments in h2 and MySQL examples

Manuri:
## Purpose
$Subject

### DIFF
--- a/examples/h2-client/h2_client.bal
+++ b/examples/h2-client/h2_client.bal
@@ -1,7 +1,7 @@
 import ballerina/h2;
 import ballerina/io;
 
-// Create a client for h2 database. Change the path before running the sample.
+// Create a `client` for h2 database. Change the path before running the sample.
 // This will create a new database in the given path if not exists already.
 h2:Client testDB = new({
         path: "./h2-client",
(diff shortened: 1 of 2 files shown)


examples/h2-client/h2_client.bal
```diff
@@ -1,7 +1,7 @@
 import ballerina/h2;
 import ballerina/io;
 
-// Create a client for h2 database. Change the path before running the sample.
+// Create a `client` for h2 database. Change the path before running the sample.
```
anupama-pathirage: Shall we reword this to Create a client endpoint for h2 database. IMO highlighting is not required as "client" is not a keyword

Manuri: "client" is a keyword. h2:Client is a `client` type object. Should we still add "endpoint"?

anupama-pathirage: I was thinking since `client` object is not visible in the example as a keyword. (But the h2:Client) What we have is two types of endpoints - client endpoints and listener endpoints. So IMO it is better if we can have more descriptive word. But it is fine. :)

Manuri: Yeah, thats true. .Updated the examples

Issue #12842, opened 2018-12-19
`io:sprintf` fails for `nil` values 


MaryamZi:
**Description:**
In the [Sprintf implementation](https://github.com/ballerina-platform/ballerina-lang/blob/v0.990.0/stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/Sprintf.java), methods are called on `args.getRefValue(k)`. 

When the Ballerina value is `nil`, `args.getRefValue(k)` becomes `null` resulting in NPEs. We need to specifically handle `nil` scenarios.

**This issue also causes the following issue in testerina:**
Since testerina internally uses `io:sprintf` to format the error messages when panicking on failure for `test:assertEquals` and `test:assertNotEquals`, test failures currently have an incorrect reason.

e.g.,
```cmd
$ ballerina test values
Compiling tests
    ballerina-examples/values:0.0.1

Running tests
    ballerina-examples/values:0.0.1
        [fail] testFunc:
            error: {}
                at ballerina/io:sprintf(<native>)
                ballerina/test:assertEquals(assert.bal:74)
```

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main(string... args) {
    string s = io:sprintf("%s", ());
    io:println(s);
}
```
Fails with:
```cmd
$ ballerina run sprintf_bug.bal
error: {}        at ballerina/io:sprintf(<native>)
           main(sprintf_bug.bal:4)
```

**Affected Versions:**
0.990.0

aashikam:
Fixed [#12867](https://github.com/ballerina-platform/ballerina-lang/pull/12867)

PR #12867, opened 2018-12-19
Fix io:sprintf failing for nil values

aashikam:
Fix https://github.com/ballerina-platform/ballerina-lang/issues/12842

### DIFF
--- a/stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/Sprintf.java
+++ b/stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/Sprintf.java
@@ -95,19 +95,34 @@ public class Sprintf extends BlockingNativeCallableUnit {
                 }
                 try {
                     char formatSpecifier = format.charAt(j);
+                    BRefType ref = args.getRefValue(k);
                     switch (formatSpecifier) {
                         case 'b':
                         case 'B':
                         case 'd':
                         case 'f':
-                            result.append(String.format("%" + padding + formatSpecifier, args.getRefValue(k).value()));
+                            if (ref == null) {
+                                throw BLangExceptionHelper.getRuntimeException(
+                                        RuntimeErrors.ILLEGAL_FORMAT_CONVERSION,
+                                        format.charAt(j) + " != ()");
+                            } else {
+                                result.append(String.format("%" + padding + formatSpecifier, ref.value()));
+                            }
                             break;
                         case 'x':
                         case 'X':
-                            formatHexString(args, result, k, padding, formatSpecifier);
+                            if (ref == null) {
+                                throw BLangExceptionHelper.getRuntimeException(
+                                        RuntimeErrors.ILLEGAL_FORMAT_CONVERSION,
+                                        format.charAt(j) + " != ()");
+                            } else {
+                                formatHexString(args, result, k, padding, formatSpecifier);
+                            }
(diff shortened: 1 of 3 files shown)


stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/Sprintf.java
```diff
@@ -95,19 +95,34 @@ public final void execute(final Context context) {
...
                         case 'd':
                         case 'f':
-                            result.append(String.format("%" + padding + formatSpecifier, args.getRefValue(k).value()));
+                            if (ref != null) {
+                                result.append(String.format("%" + padding + formatSpecifier, ref.value()));
+                            } else {
+                                throw BLangExceptionHelper.getRuntimeException(
+                                        RuntimeErrors.ILLEGAL_FORMAT_CONVERSION,
+                                        format.charAt(j) + " != ()");
+                            }
```
anupama-pathirage: Shall we invert the if condition, and fail fast?

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/Sprintf.java
```diff
@@ -95,19 +95,34 @@ public final void execute(final Context context) {
...
                         case 'x':
                         case 'X':
-                            formatHexString(args, result, k, padding, formatSpecifier);
+                            if (ref != null) {
+                                formatHexString(args, result, k, padding, formatSpecifier);
+                            } else {
+                                throw BLangExceptionHelper.getRuntimeException(
+                                        RuntimeErrors.ILLEGAL_FORMAT_CONVERSION,
+                                        format.charAt(j) + " != ()");
+                            }
```
anupama-pathirage: Shall we invert the if condition, and fail fast?

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/Sprintf.java
Manuri: @anupama-pathirage Isn't it the most probable condition that should be put as the if condition?

anupama-pathirage: @Manuri The idea is to get invalid cases out of the way first, either simply exiting or raising exceptions as appropriate. Then add the real code which needs to get executed. With that we don't need else block at all.

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/Sprintf.java
```diff
@@ -95,19 +95,32 @@ public final void execute(final Context context) {
...
                     switch (formatSpecifier) {
                         case 'b':
                         case 'B':
                         case 'd':
                         case 'f':
-                            result.append(String.format("%" + padding + formatSpecifier, args.getRefValue(k).value()));
+                            if (ref == null) {
+                                throw BLangExceptionHelper.getRuntimeException(RuntimeErrors.ILLEGAL_FORMAT_CONVERSION,
+                                        format.charAt(j) + " != ()");
+                            } else {
```
wggihan: Let's remove the else part, since we are throwing the exception in the if condition.

Issue #12812, opened 2018-12-17
Bad. Sad. error for object's abstract init function


NipunaMarcus:
**Description:**
There is a `Bad. Sad.` error on build when an object has an abstract init function. Internal error log is as below.

```
java.lang.NullPointerException
	at org.wso2.ballerinalang.compiler.semantics.analyzer.DataflowAnalyzer.visit(DataflowAnalyzer.java:840)
	at org.wso2.ballerinalang.compiler.tree.types.BLangObjectTypeNode.accept(BLangObjectTypeNode.java:70)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.DataflowAnalyzer.analyzeNode(DataflowAnalyzer.java:1044)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.DataflowAnalyzer.visit(DataflowAnalyzer.java:286)
	at org.wso2.ballerinalang.compiler.tree.BLangTypeDefinition.accept(BLangTypeDefinition.java:122)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.DataflowAnalyzer.analyzeNode(DataflowAnalyzer.java:1044)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.DataflowAnalyzer.lambda$visit$1(DataflowAnalyzer.java:252)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.DataflowAnalyzer.visit(DataflowAnalyzer.java:252)
	at org.wso2.ballerinalang.compiler.tree.BLangPackage.accept(BLangPackage.java:159)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.DataflowAnalyzer.analyzeNode(DataflowAnalyzer.java:1044)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.DataflowAnalyzer.analyze(DataflowAnalyzer.java:235)
	at org.wso2.ballerinalang.compiler.CompilerDriver.dataflowAnalyze(CompilerDriver.java:207)
	at org.wso2.ballerinalang.compiler.CompilerDriver.compile(CompilerDriver.java:162)
	at org.wso2.ballerinalang.compiler.CompilerDriver.compilePackageSymbol(CompilerDriver.java:144)
	at org.wso2.ballerinalang.compiler.CompilerDriver.compilePackage(CompilerDriver.java:106)
[...]

PR #12868, opened 2018-12-19
Fix df analysis failure for object init method as outside method def and disallow extern object init methods

MaryamZi:
## Purpose
This PR
- fixes the failure at data flow analysis phase when an object init method is defined as an outside method
- disallows extern object init methods

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/12812


lankavitharana:
any reason for not allowing extern?

MaryamZi:
@lankavitharana 
> any reason for not allowing extern?

This is because dataflow analysis cannot be done with extern __init() functions.

lankavitharana:
it's kind of like a technicality right? my point is can we say we are not allowing it because we can't do data flow analysis? if we take data flow analysis, we need that to identify whether all the fields are initialized right? so there can be scenarios where we have default values for all the fields, in which case we don't need to analyze init function. Of course, if we don't have default values, then we have to have an analyzable init function.

MaryamZi:
Yeah agreed. It would be the same even on the off chance there is an object with all private/no fields. But do we need to allow this, given that we are not expecting many extern functions anyway? 

Also, IMO, whatever logic in the init function (including setting values for object fields) should ideally be written in Ballerina. If by any chance someone actually wants to perform some extern logic it could be a separate extern function invocation within the init function. WDYT?

lankavitharana:
I think that's a better answer than the previous one :) 
+1

PR #12885, opened 2018-12-20
Fix the array, tuple and map bbes

MaryamZi:
## Purpose
Fix the array, tuple and map bbes

### DIFF
--- a/examples/arrays/arrays.bal
+++ b/examples/arrays/arrays.bal
@@ -1,11 +1,11 @@
 import ballerina/io;
 
 public function main() {
-    // This creates an `int` array of length 0.
+    // Create an `int` array of length 0.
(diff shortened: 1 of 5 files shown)


examples/arrays/arrays.bal
```diff
@@ -16,25 +16,26 @@ public function main() {
     io:println(b[999]);
     io:println(b.length());
 
-    //This initializes an array of int arrays.
+    // Initialize an array of `int` arrays.
```
pamod:
```suggestion
    // Initialize two dimensional `int` array.
```

MaryamZi: Will change this to 
```suggestion
    // Initialize a two dimensional `int` array.
```
Thanks.

PR #12819, opened 2018-12-17
Fix issue where func call cp entry is wrongly overridden

manuranga:

PR #12889, opened 2018-12-20
Test fix #12819 - add test for fp cp entry

manuranga:


hasithaa:
@manuranga, Shall we rebase this to master. ?

tests/ballerina-unit-test/src/test/resources/test-src/workers/sync-and-async-fp-call.bal
```diff
@@ -0,0 +1,19 @@
+import ballerina/io;
+import ballerina/runtime;
+
+int counter = 0;
+
+function syncAsyncFpCall() returns boolean {
+    var incrementFp = increment;
+    any f;
+    f = incrementFp.call();
+    f = start incrementFp.call();
```
lankavitharana: shall we use a channel or something to ensure this async function invocation won't finish before hitting the next line? otherwise, there is a possibility that this test may fail intermittently right?

PR #12890, opened 2018-12-20
Track the identifier literals those contain '$' symbol when debugging 

### LINKED ISSUE #10336: [Debugger] Track the identifier literals those contain '$' symbol
**Description:**
Identifier literals those contain '$' symbol, are not tracked while debugging. This bug is introduced as a result of the effort carried out to exclude Ballerina meta-variables from debugger frames.    

**Affected Versions:**
0.981.2-SNAPSHOT

Component/Debugger, Type/Bug

Pranavan

pranavan15:
## Purpose
This PR fixes #10336.


lankavitharana:
Can we add a test case for this?

pranavan15:
> Can we add a test case for this?

Ack. added with commit 8b1fca8.

PR #12957, opened 2019-01-02
Remove additional error messages for incompatible conversions 

### LINKED ISSUE #11973: Additional error messages for incompatible conversions
Try to convert a value x to type T, where the conversions between two types are not supported. e.g.:
```ballerina
T.create(x)
```
This gives two error messages, and the second one shouldn't be there.
```
ERROR: .::constrained-map.bal:494:12:: incompatible types: 'json' cannot be converted to 'T'
ERROR: .::constrained-map.bal:494:12:: function invocation on type 'typedesc' is not supported
```

warunalakshitha:
Purpose

Resolves #11973


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1023,6 +1023,12 @@ public void visit(BLangInvocation iExpr) {
                         iExpr.expr.type);
                 resultType = symTable.semanticError;
                 break;
+            case TypeTags.TYPEDESC:
+                if (dlog.errorCount == 0) {
```
SupunS: This may not be a scalable solution. What if there were other errors logged previously that are not related to the type desc?

SupunS: This may not be a scalable solution. What if there were other errors logged previously that are not related to the type desc?

warunalakshitha: Yes. other option is to move this logic to default case.

SupunS: As discussed offline, lets create a dummy symbol for invalid-conversions and continue the flow.

PR #12983, opened 2019-01-04
Move crypto related unit tests

ldclakmal:
## Purpose
- Move crypto related unit tests into the crypto module itself

### DIFF
--- a/stdlib/crypto/src/test/resources/testng.xml
+++ b/stdlib/crypto/src/test/resources/testng.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+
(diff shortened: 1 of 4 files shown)


stdlib/crypto/src/test/resources/testng.xml
```diff
@@ -0,0 +1,34 @@
...
+specific language governing permissions and limitations
+under the License.
+-->
+
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
+
+<suite name="ballerina-test-suite">
+
+    <test name="ballerina-lang-test-suite" preserve-order="true" parallel="false">
+        <groups>
```
Bhashinee: Is this required as we don't have any tests marked as broken?

wggihan: This addressed through https://github.com/ballerina-platform/ballerina-lang/pull/13029

PR #12989, opened 2019-01-04
Make theme build isolated from the distribution build

jeradrutnam:


composer/packages/distribution/webpack.config.js
```diff
@@ -21,50 +14,31 @@ module.exports = {
...
+                        options: {
                             sourceMap: true
-                            }
-                        },
-                        {
-                            loader: 'less-loader', options: {
+                        }
+                    },
+                    {
+                        loader: 'less-loader',
```
kaviththiranga: do we still need less loader here?

jeradrutnam: Ermm.. not really. Kept it for safe side. But since we are enforcing the styling through theming, we can take out the entire style loader I guess.

kaviththiranga: +1 for removal as it will prevent less imports within diagram.

jeradrutnam: Removed.

composer/packages/theme/scripts/build-theme.js
```diff
@@ -0,0 +1,128 @@
+/**
+ * Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
pahans: copyright year 2019

composer/packages/theme/src/themes/playground/theme.config
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
pahans: copyright year 2019

composer/packages/theme/scripts/build-theme.js
jeradrutnam: Actually this was created last year (2018) and forgot to update. Yes need to change it since we are merging only now. I'll fix it. Thanks 👍

jeradrutnam: Done.

composer/packages/theme/src/themes/playground/theme.config
jeradrutnam: Done.

PR #12992, opened 2019-01-04
Move WebSub unit tests under respective modules

chamil321:


stdlib/websub/src/test/resources/testng.xml
```diff
@@ -0,0 +1,31 @@
...
+  ~ under the License.
+  -->
+
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
+
+<suite name="ballerina-test-suite">
+    <listeners>
+        <listener class-name="org.ballerinalang.net.websub.BLogInitializer"/>
+    </listeners>
+    <test name="ballerina-lang-test-suite" preserve-order="true" parallel="false">
```
Manuri: Should we exclude "broken" group?

PR #13001, opened 2019-01-06
Add no-entity body flag to outbound request

chamil321:
```ballerina
public type RequestMessage Request|string|xml|json|byte[]|io:ReadableByteChannel|mime:Entity[]|();

public type ResponseMessage Response|string|xml|json|byte[]|io:ReadableByteChannel|mime:Entity[]|();
 ```

### CODE REVIEW

File: stdlib/http/src/main/java/org/ballerinalang/net/http/actions/httpclient/Forward.java [hunk 1]
```diff
@@ -63,14 +64,20 @@ protected HttpCarbonMessage createOutboundRequestMsg(Context context) {
...
         } else {
-            prepareOutboundRequest(context, path, outboundRequestMsg);
+            prepareOutboundRequest(context, path, outboundRequestMsg, isNonEntityBodyReq(outboundRequestMsg));
             String httpVerb = (String) outboundRequestMsg.getProperty(HttpConstants.HTTP_METHOD);
             outboundRequestMsg.setProperty(HttpConstants.HTTP_METHOD, httpVerb.trim().toUpperCase(Locale.getDefault()));
         }
         return outboundRequestMsg;
     }
+
+    private Boolean isNonEntityBodyReq(HttpCarbonMessage outboundRequestMsg) {
```
arukshani: method name sounds a bit weird same as in NON_ENTITY_BODY_REQUEST constant

arukshani:
`OutboundRequestEntity` and `OutboundResponseEntity` gives out the meaning that we are talking about the entity of the said request and response. But the union includes the `Request` and the `Response` itself which doesn't fall under entity category. Isn't it better to use a term that gives out the meaning for both instead of using a misleading term?

stdlib/http/src/main/java/org/ballerinalang/net/http/HttpConstants.java
```diff
@@ -395,8 +397,8 @@
     public static final String NATIVE_DATA_WEBSOCKET_CONNECTION_MANAGER = "NATIVE_DATA_WEBSOCKET_CONNECTION_MANAGER";
 
     public static final int REQUEST_STRUCT_INDEX = 1;
-    public static final String REQUEST_REUSE_STATUS_INDEX = "dirtyRequest";
     public static final boolean DIRTY_REQUEST = true;
+    public static final String NON_ENTITY_BODY_REQUEST = "NON_ENTITY_BODY_REQUEST";
```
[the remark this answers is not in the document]
chamil321: What I meant by NON_ENTITY_BODY_REQUEST is a request that does not have an entity body. Basically an entity consists of entity-header fields and an entity-body. Here the main requirement is to flag the requests which does not contain an entity-body to handle content-length/transfer-encoding headers.
If it is confusing, do you have any better suggestion to replace with?

stdlib/http/src/main/java/org/ballerinalang/net/http/actions/httpclient/Forward.java
chamil321: The same elaboration applies here as well.

stdlib/http/src/main/java/org/ballerinalang/net/http/HttpConstants.java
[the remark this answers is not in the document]
arukshani: I would prefer a simpler term something like "REQUEST_WITHOUT_BODY" since the "Body" implicitly means the body of an entity, that is, the body of either a message or of a body part. 

To me, trouble with "NON_ENTITY_BODY_REQUEST" is when you read it aloud, it sounds weird. I guess it depends on how you read it, just like punctuations make all the difference to a meaning of a sentence. when I glanced it, I  automatically read it as "nonentity bodyrequest" which is totally weird but if you read it as "non entitybody request" it gives the proper meaning. 

Maybe I'm thinking too much :) , but I would personally prefer a simpler term that gives out the meaning at a glance.

chamil321:
@arukshani Yes I was thinking that too. At first I was thinking about having `Message` as the type name. But it was quite generic. 
How about `RequestMessage` and `ResponseMessage`? or do you have any better suggestions?

arukshani:
Ideally it should be `RequestOrBody` and `ResponseOrBody` but I guess that's ugly.

So +1 for `RequestMessage` and `ResponseMessage`

tests/ballerina-integration-test/src/test/resources/http/httpservices/13_http_header_test.bal
```diff
@@ -39,6 +40,62 @@ service headerService on new http:Listener(9106) {
...
+        var result = stockqEP -> execute("GET", "/sample/entitySizeChecker", "hello ballerina");
+        if (result is http:Response) {
+            _ = caller->respond(result);
+        } else if (result is error) {
+            _ = caller->respond(result.reason());
+        }
+    }
+
+    resource function noEntityExecute(http:Caller caller, http:Request req) {
+        var result = stockqEP -> execute("GET", "/sample/entitySizeChecker", ());
```
hemikak: Remove spaces around ->

tests/ballerina-integration-test/src/test/resources/http/httpservices/13_http_header_test.bal
```diff
@@ -39,6 +40,62 @@ service headerService on new http:Listener(9106) {
...
+        var result = stockqEP -> forward("/sample/entitySizeChecker", untaint req);
+        if (result is http:Response) {
+            _ = caller->respond(result);
+        } else if (result is error) {
+            _ = caller->respond(result.reason());
+        }
+    }
+
+    resource function entityExecute(http:Caller caller, http:Request req) {
+        var result = stockqEP -> execute("GET", "/sample/entitySizeChecker", "hello ballerina");
```
hemikak: Remove spaces around -> . Check other places.

Issue #12987, opened 2019-01-04
Data sent to workers should be cloned


NatashaWso2:
**Description:**
When sending data among workers the data sent should be cloned.

**Affected Versions:**
0.990.3-SNAPSHOT

PR #13027, opened 2019-01-07
Clone data before sending to a worker

vinok88:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/12987


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2465,7 +2466,8 @@ public void visit(BLangStringTemplateLiteral stringTemplateLiteral) {
 
     @Override
     public void visit(BLangWorkerSend workerSendNode) {
-        workerSendNode.expr = rewriteExpr(workerSendNode.expr);
+        List<BLangExpression> list = Arrays.asList(rewriteExpr(workerSendNode.expr));
```
warunalakshitha: Can we use Lists.of(rewriteExpr(workerSendNode.expr)) as we used in other places in Desugar class. So we can get rid of with import as well.

PR #13038, opened 2019-01-07
Migrate io unit tests

pamod:
## Purpose

Migrate io unit tests to its own module


stdlib/io/src/test/resources/test-src/io/csv_io.bal
```diff
@@ -0,0 +1,75 @@
+import ballerina/io;
```
daneshk: Shall we add License header?

PR #13042, opened 2019-01-07
Fix nested if statement flow problem in debugger

### LINKED ISSUE #13040: [Debugger] Incorrect line flow when debugging nested if statements
**Description:**
When debugging a code that contains nested if statements, an incorrect line flow occurs.

**Steps to reproduce:**
Check the below image:

debugger

After line number 5 when executing STEP_OVER command, it goes to line number 8, which is unexpected and incorrect. 

**Affected Versions:**
0.990.2

 Component/Debugger, Type/Bug, Severity/Major

@pranavan15

pranavan15:
## Purpose
This PR fixes #13040. The root cause for the aforementioned issue is `GOTO` instructions got assigned the line number of the last statement of the `if block` it belongs. Hence, the debugger shows the last line of the inner if block as a line hit. This issue is resolved by ignoring the `GOTO` instructions during the debugging process.

## Automation tests
 - Unit tests 
   > Added

### DIFF
--- a/bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
+++ b/bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
@@ -3090,8 +3090,7 @@ public class BVM {
             return false;
         }
 
-        int opcode = ctx.currentFrame.code[ctx.currentFrame.ip].getOpcode();
-        if (isIgnorableInstruction(opcode)) {
+        if (isIgnorableInstruction(ctx)) {
             return false;
         }
         
@@ -3137,7 +3136,8 @@ public class BVM {
         return false;
     }
 
-    private static boolean isIgnorableInstruction(int opcode) {
+    private static boolean isIgnorableInstruction(Strand ctx) {
+        int opcode = ctx.currentFrame.code[ctx.currentFrame.ip].getOpcode();
         return opcode == InstructionCodes.GOTO;
     }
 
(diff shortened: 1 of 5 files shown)


bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
```diff
@@ -3090,6 +3090,11 @@ private static boolean debug(Strand ctx) {
             return false;
         }
 
+        int opcode = ctx.currentFrame.code[ctx.currentFrame.ip].getOpcode();
```
lankavitharana: shall we move this to the method?

pranavan15: +1. Commit 726d308 addresses this.

tests/ballerina-unit-test/src/test/resources/test-src/debugger/test_nested_if.bal
```diff
@@ -0,0 +1,13 @@
+public function main(string... args) {
```
hemikak: Need copyright header

pranavan15: Added with commit 9ec303d.

lankavitharana:
I guess you have rebased this with the master(hence the force push message) I think it would be better to merge rather than rebasing in future.

PR #13057, opened 2019-01-08
Terminate record type rest fields by semicolon

vinok88:
**Old**
```ballerina
type UnionRec1 record {
    string var1;
    int...
};
```
**New**
```ballerina
type UnionRec1 record {
    string var1;
    int...;
};
```


compiler/ballerina-lang/src/main/resources/grammar/BallerinaParser.g4
```diff
@@ -98,12 +98,12 @@ fieldDefinition
...
 
 recordRestFieldDefinition
-    :   typeName restDescriptorPredicate ELLIPSIS
+    :   typeName restDescriptorPredicate ELLIPSIS SEMICOLON
     |   sealedLiteral
     ;
 
 sealedLiteral
-    :   NOT restDescriptorPredicate ELLIPSIS
+    :   NOT restDescriptorPredicate ELLIPSIS SEMICOLON
```
Anoukh: `sealedLiteral` has been used in places like `arrayTypeNameLabel` and `restBindingPattern`. Won't this change affect them?
Example of an array with fixed length: 
```ballerina
int[!...] x = [1, 2];
int[!...;] x = [1, 2];
```

vinok88: Yes, it is affecting. This PR is still WIP :) I did not send any PR review request yet :)

Anyway, thanks for pointing out.

Anoukh: Oh sorry didn't know it was WIP :D

PR #13088, opened 2019-01-08
Restrict private visibility in abstract objects

vinok88:
Private fields and methods are not allowed in an abstract object.

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -304,10 +304,28 @@ public class SemanticAnalyzer extends BLangNodeVisitor {
     @Override
     public void visit(BLangObjectTypeNode objectTypeNode) {
         SymbolEnv objectEnv = SymbolEnv.createTypeEnv(objectTypeNode, objectTypeNode.symbol.scope, env);
-        objectTypeNode.fields.forEach(field -> analyzeDef(field, objectEnv));
 
-        // Visit functions as they are not in the same scope/env as the object fields
-        objectTypeNode.functions.forEach(f -> analyzeDef(f, env));
+        if (!objectTypeNode.flagSet.contains(Flag.ABSTRACT)) {
+            objectTypeNode.fields.forEach(field -> analyzeDef(field, objectEnv));
+
+            // Visit functions as they are not in the same scope/env as the object fields
+            objectTypeNode.functions.forEach(f -> analyzeDef(f, env));
+        } else {
+            // Restrict private visibility in abstract objects
+            objectTypeNode.fields.forEach(field -> {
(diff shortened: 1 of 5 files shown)


tests/ballerina-unit-test/src/test/resources/test-src/object/abstract-object-negative.bal
```diff
@@ -67,3 +67,11 @@ type Person5 abstract object {
 function Person5.getName() returns string {
     return "my name";
 }
+
+// Abstract object with private field
+type Person6 abstract object {
```
SupunS: Shall we add a test case where, an abstract object contains a private method and that object is used as a `object type reference` in another object, and trying to implement/access that private method..

eg:
```ballerina
type Foo abstract object {
    private function getName() returns string;
}

type Bar abstract object {
    Foo*
}

function Bar.getName() returns string {
    returns "bar";
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -304,10 +304,28 @@ public void visit(BLangTypeDefinition typeDefinition) {
...
+        if (!objectTypeNode.flagSet.contains(Flag.ABSTRACT)) {
+            objectTypeNode.fields.forEach(field -> analyzeDef(field, objectEnv));
+
+            // Visit functions as they are not in the same scope/env as the object fields
+            objectTypeNode.functions.forEach(f -> analyzeDef(f, env));
+        } else {
+            // Restrict private visibility in abstract objects
+            objectTypeNode.fields.forEach(field -> {
+                analyzeDef(field, objectEnv);
+                if (field.flagSet.contains(Flag.PRIVATE)) {
```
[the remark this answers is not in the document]
vinok88: I did this to avoid additional overhead, let's merge since this is in compiler, do not have to worry much on performance.

tests/ballerina-unit-test/src/test/resources/test-src/object/abstract-object-negative.bal
vinok88: What do you look for in such a test? this will fail at Foo object since we do not allow it.

SupunS: Object type referencing will copy the members from the referenced types. Since getName is a private method and is not allowed, it should not get copied. 

Simply, 
```ballerina
function Bar.getName() returns string {
    returns "bar";
}
```
should give an error saying the getName() does not exists.

vinok88: Will write a test. it won't give the method does not exist because we do not allow private in abstract objects now.

vinok88: Test case added, stopped duplicating private methods while type referencing.

Issue #13089, opened 2019-01-09
The async test that waits for all cancelled futures to terminate is disabled since it fails intermittently


NatashaWso2:
**Description:**
$subject
Need to fix the test and enable it again

**Affected Versions:**
0.990.3-SNAPSHOT

PR #13090, opened 2019-01-09
Fix intermittently failing async test cases

NatashaWso2:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13089

### DIFF
--- a/tests/ballerina-unit-test/src/test/resources/test-src/expressions/async/basic-async-operations.bal
+++ b/tests/ballerina-unit-test/src/test/resources/test-src/expressions/async/basic-async-operations.bal
@@ -4,130 +4,130 @@ import ballerina/io;
 int globalResult = 0;
 
 function testAsyncNonNativeBasic1() returns int {
-  future<int> f1 = start add(5, 2);
-  int result = wait f1;
-  return result;
+    future<int> f1 = start add(5, 2);
+    int result = wait f1;
+    return result;
 }
 
 function testAsyncNonNativeBasic2() returns int {
-  future<int> f1 = start add(5, 2);
-  int result = wait f1;
-  future<int> f2 = start add(10, 2);
-  runtime:sleep(100);
-  result = result + wait f2;
-  return result;
+    future<int> f1 = start add(5, 2);
+    int result = wait f1;
+    future<int> f2 = start add(10, 2);
+    runtime:sleep(100);
+    result = result + wait f2;
+    return result;
 }
 
 function testAsyncNonNativeBasic3() returns int {
-  future<int> f1 = start add(5, 2);
-  int result = wait f1;
-  future<int> f2 = start add(10, 2);
-  result = result + wait f2;
-  result = result + wait f2;
-  return result;
+    future<int> f1 = start add(5, 2);
+    int result = wait f1;
+    future<int> f2 = start add(10, 2);
+    result = result + wait f2;
+    result = result + wait f2;
+    return result;
 }
 
 function testAsyncNonNativeBasic4() returns int {
-  future<any> f1 = start addGlobal(1, 2);
-  _ = wait f1;
-  return globalResult;
+    future<any> f1 = start addGlobal(1, 2);
+    _ = wait f1;
+    return globalResult;
 }
 
 function testAsyncNonNativeBasic5() returns float {
-  future<float> f1 = start addSlow(10.5, 20.5);
-  return wait f1;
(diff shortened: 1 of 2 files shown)


tests/ballerina-unit-test/src/test/resources/test-src/expressions/async/basic-async-operations.bal
```diff
@@ -79,8 +79,8 @@ function testAsyncNonNativeBasic9() returns int {
 
 function testAsyncNonNativeBasic10() returns any {
   future<int> f1 = start addNum(5, 2);
-  future<int> f2 = start addNum(5, 2);
-  boolean cancelled_f1 = f1.cancel();
+  future<int> f2 = start subtract(5, 2);
+  boolean cancelled_f2 = f2.cancel();
```
lankavitharana: do we have a test for wait for all with all futures canceled?

NatashaWso2: No we don't have I'll add one

NatashaWso2: Added a test with f5c0019. Thank you :)

tests/ballerina-unit-test/src/test/resources/test-src/expressions/async/basic-async-operations.bal
```diff
@@ -79,8 +79,17 @@ function testAsyncNonNativeBasic9() returns int {
...
   future<int> f1 = start addNum(5, 2);
-  future<int> f2 = start addNum(5, 2);
+  future<int> f2 = start subtract(5, 2);
+  boolean cancelled_f2 = f2.cancel();
+  any result = wait {f1,f2};
+  return result;
+}
+
+function testAsyncNonNativeBasic11() returns any {
+  future<int> f1 = start addNum(5, 2);
```
lankavitharana: I think you need to use subtract method here as well, (you cannot guarantee that addNum method will not be finished before calling cancel right?)

NatashaWso2: yeah +1 using the subtract method ensures that the method will never be finished. Added it with 738541d. Thank you :)

tests/ballerina-unit-test/src/test/resources/test-src/expressions/async/basic-async-operations.bal
```diff
@@ -61,8 +61,8 @@ function testAsyncNonNativeBasic7() returns int {
 }
 
 function testAsyncNonNativeBasic8() returns int {
-  future<int> f1 = start addNum(5, 2);
-  future<int> f2 = start addNum(5, 2);
+  future<int> f1 = start subtract(8, 3);
```
lankavitharana: do we have a test to see if the wait for any will return the successful future if only one future is canceled? I think there are tests for this kind of scenario when some panic happens in one future, I think we may need to add something like that for cancelation as well.

NatashaWso2: Yes we do have a test https://github.com/ballerina-platform/ballerina-lang/pull/13090/files#diff-c57bc8d6b3a06c2b98c705942dcafed3R72. This will return the future that was not cancelled

lankavitharana: in that test, we cannot guarantee the canceled future will be really canceled right? (it may successfully finish execution, therefore test may intermittently fail) I think you need to use substract method there as well.

tests/ballerina-unit-test/src/test/resources/test-src/expressions/async/basic-async-operations.bal
```diff
@@ -79,8 +79,17 @@ function testAsyncNonNativeBasic9() returns int {
...
-  future<int> f2 = start addNum(5, 2);
+  future<int> f2 = start subtract(5, 2);
+  boolean cancelled_f2 = f2.cancel();
+  any result = wait {f1,f2};
+  return result;
+}
+
+function testAsyncNonNativeBasic11() returns any {
+  future<int> f1 = start subtract(7, 2);
+  future<int> f2 = start subtract(5, 2);
```
hemikak: Something wrong with the formatting. Tab would have 4 spaces.

PR #13114, opened 2019-01-09
Validate for reserved org-names given when executing the 'init' and 'push' commands

NatashaWso2:
## Automation tests
 - Integration tests

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes
 - Ran FindSecurityBugs plugin and verified report? yes
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/util/RepoUtils.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/util/RepoUtils.java
@@ -48,6 +48,9 @@ public class RepoUtils {
     private static final boolean BALLERINA_DEV_STAGE_CENTRAL = Boolean.parseBoolean(
             System.getenv("BALLERINA_DEV_STAGE_CENTRAL"));
 
+    private static final String BALLERINA_ORG = "ballerina";
+    private static final String BALLERINAX_ORG = "ballerinax";
+
     /**
      * Create and get the home repository path.
      *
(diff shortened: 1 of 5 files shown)


MaryamZi:
IINM, even with this change, we still won't be validating for build, test, etc. right? IMO, we need to validate for all commands in a generic way if possible. Currently, saying `ballerina build` with "ballerina" as the org name fails silently.

For example:
```cmd
$ ballerina build 05-simple-values/
Compiling source
    ballerina/05-simple-values
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/util/RepoUtils.java
```diff
@@ -159,11 +159,21 @@ public static boolean validateOrg(String orgName) {
...
     }
+
+    /**
+     * Check if the org-name is a reserved org-name in ballerina.
+     *
+     * @param orgName The org-name
+     * @return True if the org-name is reserved, else false.
+     */
+    public static boolean isReservedOrgName(String orgName) {
+        return orgName.equals("ballerina") || orgName.equals("ballerinax");
```
MaryamZi: Shall we introduce two constants for "ballerina" and "ballerinax" ?

NatashaWso2: Fixed with ff52b4f

PR #13128, opened 2019-01-10
Improve the descriptions in the maps BBE #12984

### LINKED ISSUE #12984: Improve the descriptions in the maps BBE
Need to $subject. There are incorrect/improper terminology used there.

praneesha:
Closes #12984


examples/maps/maps.bal
```diff
@@ -1,48 +1,48 @@
...
     io:println(city);
 
-    // Add or update the value for a key.
+    // Add or update the value of a key.
     addrMap["postalCode"] = "00300";
     addrMap.postalCode = "00301";
     io:println(addrMap);
 
-    // Use the `keys()` function of the `map` to get the keys of the `map` as an array.
+    // Use the `keys()` method to print the keys of the `map` as an array.
```
SupunS: `keys()` method will **return** the keys. It will not **print** the keys.

examples/maps/maps.bal
```diff
@@ -1,48 +1,48 @@
...
     io:println(addrMap.keys());
 
-    // Print the number of keys in the `map`.
+    // Print the number of mappings in the `map`.
     io:println(addrMap.length());
 
-    // Remove a key using the `remove()` function.
+    // Keys can be removed using the `remove()` method.
```
SupunS: `remove()` method will remove both the key and value.

pubudu91:
```suggestion
    // Mappings can be removed using the `remove()` method.
```

Issue #13163, opened 2019-01-14
[Spec Deviation] Creating and array with a type with no implicit initial value should be restricted


Anoukh:
**Description:**
If a type T has no implicit initial value, an array of type T should not be possible. Instead, the array type should be nillable -> T, in which case the implicit initial value of the members of the array will be nil

```ballerina
public type FooObject object {
    public string fooFieldOne;

    public function __init(string fooFieldOne) { // Since the constructor requires an argument, there is no imlpicit initial value for FooObject
        self.fooFieldOne = fooFieldOne;
    }

    public function getFooFieldOne() returns string {
        return self.fooFieldOne;
    }
};

function testArrayMemberTypesBroken() {
    utils:FooObject[] objectArray = []; // This should fail at compile time
}
```

PR #13193, opened 2019-01-16
Mandate list element types to have implicit initial value 

rdhananjaya:
## Purpose
In Ballerina spec "Ballerina-Language-Specification-v0.990-2019-01-16", under list subsection it says:
```text
A type ​T[]​ contains a list shape if all members of the list shape are in ​T​. 
A type ​T[n] contains a list shape if in addition the length of the list shape is n.
The member type of an array type can be any type T (including arrays), 
provided only that T has an implicit initial value. 
(When T does not have an implicit initial value, an array of T? may be used instead; 
see Optional Types.)
```

Also form the spec under **5. Values, types and variables** there is a table summarizing implicit initial values for each type.

This PR add this constraint checking to Ballerina compiler.

```ballerina
type FT 1|2|3;
type FTN 1|2|3|();

FT[] ar_ft = [1,2]; // error: array element type 'FT' must have an implicit initial value, use 'FT?'
FT?[] ar_ft2 = [2,3]; // valid
FT[2] ar_ft2_sealed = [2,3]; // valid for sealed/closed arrays
FTN[] ar_ftn = []; // valid
```
 
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13163


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BFiniteType.java
```diff
@@ -65,6 +64,11 @@ public String toString() {
         return joiner.toString();
     }
 
+    @Override
+    public boolean hasImplicitInitialValue() {
+        return valueSpace.stream().anyMatch(value -> value.type.tag == TypeTags.NIL);
```
pubudu91: For finite types, checking for just nil is not enough. As per the spec, the following finite type's implicit initial value would be 0.
```ballerina
type STATE 0|1|2;
```
Here's what the spec says regarding implicit initial values of union types:
```
The rules for the implicit initial value of a union type are as follows:
● if () is in the value space of a union type, then the implicit initial value for the union
type is ();
● otherwise if the value space only contains only values of a single basic type, and one
of those values is the implicit initial value of that basic type, then the implicit initial
value for the union is the implicit initial value of the basic type of its values (for
example, the implicit initial value for 0|1 is 0)
● otherwise the union type does not have an implicit initial value.
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -84,4 +84,9 @@ public String getDesc() {
     public void setNullable(boolean nullable) {
         this.nullable = nullable;
     }
+
+    @Override
+    public boolean hasImplicitInitialValue() {
+        return memberTypes.stream().anyMatch(t -> t.tag == TypeTags.NIL);
```
pubudu91: Same comment I put for finite types applies here

pubudu91:
There's a drop in test coverage. Shall we check that out?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -341,8 +342,25 @@ private void visitFunction(BLangFunction funcNode) {
             return;
         }
         boolean isNilableReturn = funcNode.symbol.type.getReturnType().isNullable();
+
+        BLangType returnTypeNode = funcNode.returnTypeNode;
+        if (returnTypeNode.type.tag == TypeTags.ARRAY && returnTypeNode instanceof BLangArrayType) {
```
Shan1024: We can use node kind instead of the `instanceof` checks.

Eg - `funcNode.returnTypeNode.getKind()== NodeKind.ARRAY_TYPE`

Applicable to all other places.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -341,8 +342,25 @@ private void visitFunction(BLangFunction funcNode) {
...
+            analyzeArrayElemImplicitInitialValue(((BLangArrayType) returnTypeNode),
+                    returnTypeNode.pos);
```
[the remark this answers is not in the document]
rdhananjaya: You mean inline `returnTypeNode`?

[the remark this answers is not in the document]
Shan1024: Yes, AFAIS, it will not exceed 120 character limit.

Eg - `analyzeArrayElemImplicitInitialValue(((BLangArrayType) returnTypeNode), returnTypeNode.pos);`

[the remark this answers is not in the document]
rdhananjaya: Oh my bad, I didn't see that second argument is on a separate line (Y)

rdhananjaya:
Only array definitions were constrained initially and then constraints on returns, parameters were added as a tool to help me with the porting of stdlib code to comply with this constraint, and I didn't remove it later since it doesn't seems to violate the spec. 

I do agree that only unsealed arrays needs to be constrained like this, because sealed arrays cannot have empty slots.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -341,6 +342,7 @@ private void visitFunction(BLangFunction funcNode) {
             return;
         }
         boolean isNilableReturn = funcNode.symbol.type.getReturnType().isNullable();
+
```
NatashaWso2: Shall we remove this extra  line?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -999,6 +1003,45 @@ public void visit(BLangSimpleVariable varNode) {
         }
     }
 
+    private void analyzeArrayVariableImplicitInitialValue(BLangSimpleVariable varNode) {
+        if (varNode.typeNode == null || varNode.typeNode.type == null) {
```
NatashaWso2: Do we need to check for both the conditions? Can we only have the first condition `(varNode.typeNode == null`) since this would imply that the type is also null?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BArrayType.java
```diff
@@ -77,6 +77,15 @@ public TypeKind getKind() {
         return TypeKind.ARRAY;
     }
 
+    @Override
+    public boolean hasImplicitInitialValue() {
+        if (state == BArrayState.UNSEALED) {
```
NatashaWso2: We can remove this else block by modifying the code as below:
```java
if (state == BArrayState.UNSEALED) {
   return true; // Empty array.
}
return eType.hasImplicitInitialValue();
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BFiniteType.java
```diff
@@ -65,6 +65,41 @@ public String toString() {
...
     }
 
+    @Override
+    public boolean hasImplicitInitialValue() {
+        // Has NIL element as a member.
+        if (valueSpace.stream().anyMatch(value -> value.type.tag == TypeTags.NIL)) {
+            return true;
+        }
+        // All of the element are from same type. And elements contains implicit initial value.
+        if (!valueSpace.isEmpty()) {
```
NatashaWso2: Can we invert the condition and fail fast? 
```java
if (valueSpace.isEmpty()) { 
   return false;
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BFiniteType.java
```diff
@@ -65,6 +65,41 @@ public String toString() {
...
+            BLangExpression firstElement = valueSpace.iterator().next();
+            boolean sameType = valueSpace.stream().allMatch(value -> value.type.tag == firstElement.type.tag);
+            if (!sameType) {
+                return false;
+            }
+
+            switch (firstElement.type.tag) {
+                case TypeTags.STRING:
```
NatashaWso2: What will happen if the type is decimal?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -84,4 +85,39 @@ public String getDesc() {
...
+
+    @Override
+    public boolean hasImplicitInitialValue() {
+        // NIL is a member.
+        if (memberTypes.stream().anyMatch(t -> t.tag == TypeTags.NIL)) {
+            return true;
+        }
+
+        // All members are of same type and has the implicit initial value as a member.
+        if (!memberTypes.isEmpty()) {
```
NatashaWso2: Shall we invert the if condition and fail fast?
```java
 if (memberTypes.isEmpty()) { 
    return false;
 }
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -84,4 +85,39 @@ public String getDesc() {
...
+            Iterator<BType> iterator = memberTypes.iterator();
+            BType firstMember;
+            for (firstMember = iterator.next(); iterator.hasNext(); ) {
+                if (firstMember != iterator.next()) {
+                    return false;
+                }
+            }
+            // Control reaching here means there is only one type in the union.
+
```
NatashaWso2: Shall we remove this extra line?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -84,4 +85,39 @@ public String getDesc() {
...
+            return isBasicType(firstMember.tag) && firstMember.hasImplicitInitialValue();
+        }
+        return false;
+    }
+
+    private boolean isBasicType(int tag) {
+        switch (tag) {
+            case TypeTags.BOOLEAN:
```
NatashaWso2: Isn't decimal a simple type?

stdlib/http/src/main/ballerina/http/resiliency/failover_client_endpoint.bal
```diff
@@ -270,9 +270,17 @@ function performExecuteAction (string path, Request request, string httpVerb,
 // Handles all the actions exposed through the Failover connector.
 function performFailoverAction (string path, Request request, HttpOperation requestAction,
                                                 FailoverClient failoverClient) returns Response|error {
-    Client foClient = failoverClient.failoverInferredConfig.failoverClientsArray[failoverClient.succeededEndpointIndex];
+
```
NatashaWso2: Shall we remove this extra line?

stdlib/http/src/main/ballerina/http/resiliency/failover_client_endpoint.bal
```diff
@@ -383,13 +391,20 @@ function performFailoverAction (string path, Request request, HttpOperation requ
         }
         failoverRequest = check createFailoverRequest(failoverRequest, requestEntity);
         runtime:sleep(failoverInterval);
-        foClient = failoverClients[currentIndex];
+
```
NatashaWso2: Shall we remove this extra line?

stdlib/http/src/main/ballerina/http/resiliency/load_balance_client_endpoint.bal
```diff
@@ -375,13 +383,15 @@ function createClientEPConfigFromLoalBalanceEPConfig(LoadBalanceClientEndpointCo
 }
 
 function createLoadBalanceHttpClientArray(LoadBalanceClientEndpointConfiguration loadBalanceClientConfig)
-                                                                                    returns Client[]|error {
-    Client[] httpClients = [];
+                                                                                    returns Client?[]|error {
```
NatashaWso2: Shall we remove this extra line?

stdlib/transactions/src/main/ballerina/transactions/2pc_transaction.bal
```diff
@@ -158,13 +158,21 @@ type TwoPhaseCommitTransaction object {
...
-        foreach var f in results {
+        foreach var res in results {
+            future<((PrepareResult|error)?, Participant)> f;
+            if (res is future<((PrepareResult|error)?, Participant)>) {
+                f = res;
+            } else {
+                error err = error("Unexpected nil found");
+                panic err;
+            }
+
```
NatashaWso2: Shall we remove this extra line?

stdlib/transactions/src/main/ballerina/transactions/2pc_transaction.bal
```diff
@@ -207,13 +215,21 @@ type TwoPhaseCommitTransaction object {
...
-        foreach var f in results {
+        foreach var r in results {
+            future<(NotifyResult|error)?> f;
+            if (r is future<(NotifyResult|error)?>) {
+                f = r;
+            } else {
+                error err = error("Unexpected nil found");
+                panic err;
+            }
+
```
NatashaWso2: Shall we remove this extra line?

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/binaryoperations/EqualAndNotEqualOperationsTest.java
```diff
@@ -823,51 +823,49 @@ public void selfAndCyclicReferencingNegativeFunctions(String testFunctionName) {
 
     @Test(description = "Test equal and not equal with errors")
     public void testEqualAndNotEqualNegativeCases() {
-        Assert.assertEquals(resultNegative.getErrorCount(), 34);
+        //Assert.assertEquals(resultNegative.getErrorCount(), 34);
```
NatashaWso2: Why is this line commented?

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArrayTest.java
```diff
@@ -132,4 +135,47 @@ public void testConnectorNegativeCases() {
...
+
+    @Test(description = "Test arrays of types without implicit initial values")
+    public void testArrayImplicitInitialValuesOfFiniteType() {
+        CompileResult negResult = BCompileUtil.compile(
+                "test-src/statements/arrays/array-implicit-initial-value-finite-type-negative.bal");
+        Assert.assertEquals(negResult.getErrorCount(), 4);
+        BAssertUtil.validateError(negResult, 0,
+                "array element type 'allInitNonZero' does not have an implicit initial value, use 'allInitNonZero?'",
+                22, 1);
+
```
NatashaWso2: Shall we remove this extra line? Shall we do the same for all the extra new lines in this file?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
rdhananjaya: Does `varNode.typeNode != null` guarantee that `varNode.typeNode.type != null`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BFiniteType.java
rdhananjaya: I don't think we can define finite type for decimals as we don't have a decimal literal.

stdlib/transactions/src/main/ballerina/transactions/2pc_transaction.bal
rdhananjaya: That line separate 2 logical code units

stdlib/http/src/main/ballerina/http/resiliency/failover_client_endpoint.bal
rdhananjaya: This separate 2 logical blocks of code

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -999,6 +1002,45 @@ public void visit(BLangSimpleVariable varNode) {
...
+        // Variable is a array def, elements must have implicit initial value.
+        if (varNode.typeNode.type.tag == TypeTags.ARRAY) {
+            analyzeArrayElemImplicitInitialValue(varNode.typeNode, varNode.pos);
+        } else if (varNode.typeNode.type.tag == TypeTags.UNION
+                && varNode.typeNode.getKind() == NodeKind.UNION_TYPE_NODE) {
+            // Check each member of the union.
+            for (BLangType memberTypeNode : ((BLangUnionTypeNode) varNode.typeNode).memberTypeNodes) {
+                analyzeArrayElemImplicitInitialValue(memberTypeNode, memberTypeNode.pos);
+            }
+        } else if (varNode.typeNode.type.tag == TypeTags.UNION && varNode.typeNode.getKind() == NodeKind.ARRAY_TYPE) {
```
SupunS: If we check `varNode.typeNode.getKind()` only for all three cases, then i think we can merge this one with L#1010.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNilType.java
```diff
@@ -48,4 +48,9 @@ public String getDesc() {
     public String toString() {
         return Names.NIL_VALUE.value;
     }
+
+    @Override
+    public boolean hasImplicitInitialValue() {
```
SupunS: can we move this to the BType class as the default implementation?
Then anyone can override when needed..

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BFiniteType.java
```diff
@@ -65,6 +65,41 @@ public String toString() {
...
+            return false;
+        }
+        switch (firstElement.type.tag) {
+            case TypeTags.STRING:
+                return containsElement(valueSpace, "\"\"");
+            case TypeTags.INT:
+                return containsElement(valueSpace, "0");
+            case TypeTags.FLOAT:
+            case TypeTags.DECIMAL:
+                return containsElement(valueSpace, "0.0");
```
hasithaa: Please check, is this working in a scenario like this, 
```
 type Foo <decimal> 0 | 10;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNilType.java
hasithaa: +1.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BFiniteType.java
rdhananjaya: At the moment we don't support this syntax, will add check for decimal "0" for future proofing.

hasithaa:
There are some commit messages which are not starting with a verb. i.e "Array element type IIV refactor - *". Can we reword those ?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -999,6 +1002,49 @@ public void visit(BLangSimpleVariable varNode) {
         }
     }
 
+    private void analyzeArrayVariableImplicitInitialValue(BLangSimpleVariable varNode) {
+        BLangType varTypeNode = varNode.typeNode;
+        if (varTypeNode == null || varTypeNode.type == null) {
```
rdhananjaya: @SupunS @NatashaWso2 It turned out that for ignore variable (var _) varNode.typeNode is not null yet varNode.typeNode.type is null, hence re-adding the null check.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BObjectType.java
```diff
@@ -56,4 +58,20 @@ public String toString() {
         return this.tsymbol.toString();
     }
 
+    @Override
+    public boolean hasImplicitInitialValue() {
+        if (this.tsymbol instanceof BObjectTypeSymbol) { // To please the FindBugs warning, BC_UNCONFIRMED_CAST
+            BAttachedFunction initializerFunc = ((BObjectTypeSymbol) this.tsymbol).initializerFunc;
+            if (initializerFunc == null) {
+                // No __init function found.
```
hasithaa: An Object type can't have an implicit initial value if there any field with a type that doesn't have an implicit initial value. @SupunS pls confirm.

rdhananjaya: As per spec, it's not allowed for a field which doesn't have a IIV to be uninitialized when __init exit.

SupunS: Yes, currently for public fields, it is mandatory to be initialized in the object __init(). However i'm not sure whether we do the same check for the private fields (we don't, IIRC).

If it is the case, i guess we need to update that as well.

rdhananjaya:
Temporary closing to refactor code

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -84,4 +85,36 @@ public String getDesc() {
...
+        // NIL is a member.
+        if (memberTypes.stream().anyMatch(t -> t.tag == TypeTags.NIL)) {
+            return true;
+        }
+
+        // All members are of same type and has the implicit initial value as a member.
+        Iterator<BType> iterator = memberTypes.iterator();
+        BType firstMember;
+        for (firstMember = iterator.next(); iterator.hasNext(); ) {
+            if (firstMember != iterator.next()) {
```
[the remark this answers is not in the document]
rdhananjaya: +1

pom.xml
```diff
@@ -1836,7 +1836,7 @@
                 <module>tests/ballerina-compiler-plugin-test</module>
                 <module>cli/ballerina-cli-utils</module>
 
-                <module>language-server</module>
+                <!--<module>language-server</module>-->
```
hasithaa: @rdhananjaya Hope this was added mistakenly.

rdhananjaya: Yes yes, never intended to commit a pom here. Will remove

tests/ballerina-spec-conformance-tests/S05_type_descriptors/tests/anydata.bal
```diff
@@ -16,7 +16,7 @@
 
 import ballerina/test;
 
-(anydata|error)[] anydataOrErrorArray = [a, b, errValOne, c];
+(anydata|error)[4] anydataOrErrorArray = [a, b, errValOne, c];
```
hasithaa: @rdhananjaya anydata|error should have an implicit initial value, which is (). Please check.

tests/ballerina-spec-conformance-tests/S05_values_types_variables/tests/deviations/values_types_variables.bal
```diff
@@ -114,7 +114,7 @@ function testImplicitInitialValuesBroken() {
     //typedescArray[1] = int;
     //test:assertTrue(typedescArray[0], (), msg = "expected implicit initial value of typedesc to be ()");
 
-    One[] singletonArray = [];
+    One?[] singletonArray = [];
```
hasithaa: @MaryamZi, @Anoukh Pls check this logic.

tests/ballerina-spec-conformance-tests/S05_values_types_variables/tests/implicit_initial_value.bal
```diff
@@ -118,3 +118,27 @@ function testImplicitInitialValueOfJSONType() {
     jsonArray[1] = 3;
     test:assertEquals(jsonArray[0], (), msg = "expected implicit initial value of json type to be ()");
 }
+
+@test:Config {}
+function testArrayMemberTypesBroken() {
```
hasithaa: The function name should be renamed. It is not broken anymore :)

tests/ballerina-spec-conformance-tests/S05_values_types_variables/tests/implicit_initial_value.bal
```diff
@@ -118,3 +118,27 @@ function testImplicitInitialValueOfJSONType() {
...
+        self.fooFieldOne = fooFieldOne;
+    }
+
+    public function getFooFieldOne() returns string {
+        return self.fooFieldOne;
+    }
+};
+
+@test:Config {}
+function testErrorImplicitInitialValueBroken() {
```
hasithaa: The function name should be renamed. It is not broken anymore :)

tests/ballerina-spec-conformance-tests/S05_values_types_variables/tests/implicit_initial_value.bal
```diff
@@ -118,3 +118,27 @@ function testImplicitInitialValueOfJSONType() {
...
+    error?[] errorArray = [];
+    error er = error("Error One");
+    errorArray[1] = er;
+}
```
hasithaa: Please add a new line, As a practice, we end a file with a new line.

tests/ballerina-spec-conformance-tests/S05_values_types_variables/tests/implicit_initial_value.bal
rdhananjaya: I think we can remove this and the below test altogether right?
1. They are tested in array test tests anyway.
2. This code only confirm that `QuxObject?` have an implicit initial value, not that QuxObject doesn't have an IIV.

What do you think?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1934,9 +1936,146 @@ public BType getSafeType(BType type, boolean liftError) {
...
+            case TypeTags.NIL:
+            case TypeTags.TYPEDESC:
+            case TypeTags.JSON:
+            case TypeTags.MAP:
+            case TypeTags.NONE:
+            case TypeTags.STREAM:
+            case TypeTags.TABLE:
+            case TypeTags.XML:
+            case TypeTags.ANYDATA:
+            case TypeTags.ANY:
```
SupunS: can't we use `type.tag < TypeTags.INVOKABLE`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1934,9 +1936,146 @@ public BType getSafeType(BType type, boolean liftError) {
...
+                return true;
+            case TypeTags.ANNOTATION:
+            case TypeTags.CHANNEL:
+            case TypeTags.ERROR:
+            case TypeTags.FUTURE:
+            case TypeTags.INVOKABLE:
+            case TypeTags.PACKAGE:
+            case TypeTags.SEMANTIC_ERROR:
+            case TypeTags.XML_ATTRIBUTES:
```
SupunS: don't need this cases right? can replace these from the `analyzeImplicitInitialValue` method's body

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1934,9 +1936,146 @@ public BType getSafeType(BType type, boolean liftError) {
...
+        }
+        // Control reaching this point means there is only one type in the union.
+        return isValueType(firstMember) && hasImplicitInitialValue(firstMember);
+    }
+
+    private boolean analyzeObjectType(BObjectType type) {
+        if (type.getKind() == TypeKind.SERVICE) {
+            return false;
+        }
+        if (type.tsymbol instanceof BObjectTypeSymbol) { // To please the FindBugs warning, BC_UNCONFIRMED_CAST
```
SupunS: lets use the symbol kind for the comparison. Thats what we have done elsewhere.

rdhananjaya: IIRC before moving into Types class, travis failed with commented error message. Let me try again with symbol kind.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
rdhananjaya: I think having these tags are important for documenting that these types doesn't have IIV, what do you think? Because without performing any analysis we know that these types doesn't have IIVs.

SupunS: I don't think we need to add unnecessary code just for the purpose of documenting :)
'default' case means 'everything else'.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
rdhananjaya: We can go til type.tag < TypeTags.RECORD, as for record we need to look at members.
Still listing these constants feel more readable to me.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
rdhananjaya: For me this this reads like, 1st block does have IIVs for sure, 2nd block doesn't have for sure, for rest of it lets analyze the type to find out.

SupunS: in the type tags, if you re-order and move recod to just after the map, Your code will looks like below:
```ballerina
private boolean hasImplicitInitialValue(BType type) {
        if (type.tag < TypeTags.RECORD) {
            return true;
        }
        switch (type.tag) {
            case TypeTags.ARRAY:
                BArrayType arrayType = (BArrayType) type;
                if (arrayType.state == BArrayState.UNSEALED) {
                    // Empty array is the implicit initial value for arrays.
                    return true;
                }
                return hasImplicitInitialValue(arrayType.eType);
            case TypeTags.FINITE:
                return analyzeFiniteType((BFiniteType) type);
            case TypeTags.OBJECT:
                return analyzeObjectType((BObjectType) type);
            case TypeTags.RECORD:
                BRecordType recordType = (BRecordType) type;
                return recordType.fields.stream().allMatch(f -> hasImplicitInitialValue(f.type));
            case TypeTags.TUPLE:
                BTupleType tupleType = (BTupleType) type;
                return tupleType.tupleTypes.stream().allMatch(this::hasImplicitInitialValue);
            case TypeTags.UNION:
                return analyzeUnionType((BUnionType) type);
            default:
                return false;
        }
    }
``` 
Thats 50% reduction of LOC.
It still has the same three blocks you mentioned, but with positions swapped.

rdhananjaya: Re-ordering RECORD breaks the compiler, it looks like there are some other place that depends on TAG ordering.
We can go with:
```java
        if (type.tag < TypeTags.RECORD) {
            return true;
        }
        switch (type.tag) {
            case TypeTags.TYPEDESC:
            case TypeTags.STREAM:
            case TypeTags.MAP:
                return true;
            case TypeTags.ARRAY:
                BArrayType arrayType = (BArrayType) type;
                if (arrayType.state == BArrayState.UNSEALED) {
                    // Empty array is the implicit initial value for arrays.
                    return true;
                }
                return hasImplicitInitialValue(arrayType.eType);
            case TypeTags.FINITE:
                return analyzeFiniteType((BFiniteType) type);
            case TypeTags.OBJECT:
                return analyzeObjectType((BObjectType) type);
            case TypeTags.RECORD:
                BRecordType recordType = (BRecordType) type;
                return recordType.fields.stream().allMatch(f -> hasImplicitInitialValue(f.type));
            case TypeTags.TUPLE:
                BTupleType tupleType = (BTupleType) type;
                return tupleType.tupleTypes.stream().allMatch(this::hasImplicitInitialValue);
            case TypeTags.UNION:
                return analyzeUnionType((BUnionType) type);
            default:
                return false;
        }
```

tests/ballerina-spec-conformance-tests/S05_values_types_variables/tests/value_belongs_to_type.bal
```diff
@@ -23,9 +23,9 @@ const EXPECTED_VALUE_NOT_TO_BELONG_TO_TYPE_FAILURE_MESSAGE = "expected value to
 // look like the type no matter how the value is mutated
 @test:Config {}
 function testBelongsToWithArrays() {
-    (int|boolean)[] a1 = [true, false, false];
+    (int|boolean)[3] a1 = [true, false, false];
     any a2 = a1;
-    if !(a2 is anydata[]) || !(a2 is (int|boolean)[]) {
+    if !(a2 is anydata[3]) || !(a2 is (int|boolean)[3]) {
```
MaryamZi: `a2 is anydata[]` should also work (without the length). Seems like an issue with the `is` check.

Shall we create an issue if there isn't one already and add a comment here?

```suggestion
    // <GITHUB_ISSUE> 
    // if !(a2 is anydata[]) || !(a2 is (int|boolean)[3]) {
```

tests/ballerina-spec-conformance-tests/S05_values_types_variables/tests/implicit_initial_value.bal
```diff
@@ -118,3 +118,15 @@ function testImplicitInitialValueOfJSONType() {
     jsonArray[1] = 3;
     test:assertEquals(jsonArray[0], (), msg = "expected implicit initial value of json type to be ()");
 }
+
+public type QuxObject object {
+    public string fooFieldOne;
+
+    public function __init(string fooFieldOne) {
```
Anoukh: Shall we refactor this object so that it has an IIV and test it with a new function `testImplicitInitialValueOfObjectType()`?

rdhananjaya: I don't think we can do that at the moment, we need to wait for @pubudu91 s PR to be merged in order to get a IIV of an object. Lets add a deviation test, WDYT?

hasithaa:
@SupunS Can you please check whether requested changes are added?

tests/ballerina-spec-conformance-tests/S05_values_types_variables/tests/implicit_initial_value.bal
Anoukh: Ah yes we'll add a deviation test in that case (y)

PR #13194, opened 2019-01-16
Reuse http endpoint initialization in gRPC

daneshk:
## Purpose
$subject. This fix is to remove duplicates when initializing http endpoint. Previously same endpoint config population code is duplicated in both gRPC and http code.


stdlib/grpc/src/main/ballerina/grpc/grpc_commons.bal
```diff
@@ -67,3 +67,49 @@ public type ServiceOcspStapling record {
...
+# Defines the possible values for the chunking configuration in HTTP services and clients.
+#
+# `AUTO`: If the payload is less than 8KB, content-length header is set in the outbound request/response,
+#         otherwise chunking header is set in the outbound request/response
+# `ALWAYS`: Always set chunking header in the response
+# `NEVER`: Never set the chunking header even if the payload is larger than 8KB in the outbound request/response
+public type Chunking CHUNKING_AUTO|CHUNKING_ALWAYS|CHUNKING_NEVER;
+
+# If the payload is less than 8KB, content-length header is set in the outbound request/response,
+# otherwise chunking header is set in the outbound request/response.}
```
pamod:
```suggestion
# otherwise chunking header is set in the outbound request/response.
```

stdlib/grpc/src/main/ballerina/grpc/client_endpoint.bal
```diff
@@ -69,26 +69,61 @@ public type Client client object {
 # + secureSocket - The SSL configurations for the client endpoint.
 # + timeoutMillis - The maximum time to wait (in milliseconds) for a response before closing the connection.
```
wggihan: We need to add documentation for these new properties.

PR #13235, opened 2019-01-20
Remove clone instruction code

warunalakshitha:
## Purpose
This will be replaced by builtin private method. Anyway user syntax will be same.

## Approach
Built in methods will be desugar to private functions in builtin package. ex _clone


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/programfile/Mnemonics.java
```diff
@@ -270,8 +270,7 @@
         mnemonics[InstructionCodes.SNEWARRAY] = "snewarray";
         mnemonics[InstructionCodes.BNEWARRAY] = "bnewarray";
         mnemonics[InstructionCodes.RNEWARRAY] = "rnewarray";
-
-        mnemonics[InstructionCodes.CLONE] = "clone";
+
```
vinok88: Extra line. Shall we remove?

stdlib/builtin/src/main/java/org/ballerinalang/nativeimpl/builtin/operationslib/Clone.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Performs a deep copy, recursively copying all structural values and their members.
+ *
+ * @since 0.991.0
+ */
+@BallerinaFunction(
+        orgName = "ballerina",
+        packageName = "builtin",
+        functionName = "_clone",
+        returnType = { @ReturnType(type = TypeKind.ANYDATA) }
```
vinok88: Don't we need to define the argument 'anydata value' as well here?

stdlib/builtin/src/main/java/org/ballerinalang/nativeimpl/builtin/operationslib/Clone.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.HashMap;
+
+/**
+ * Performs a deep copy, recursively copying all structural values and their members.
+ *
+ * @since 0.991.0
+ */
+@BallerinaFunction(
+        orgName = "ballerina",
```
vinok88: Shall we use constants for these string literals?

warunalakshitha: I think we do not need to add constants here since these are annotations and do not used in code. I checked other built in methods as well where we did the same. WDYT?

warunalakshitha:
Closing this as I will send  new PR to refactor_native_functions branch with latest changes with new standard package.

PR #13238, opened 2019-01-21
Resolve spec violations in WebSocket

riyafa:


Manuri:
There is a compilation error. Shall we fix that?

PR #13242, opened 2019-01-21
Improve JMS code

riyafa:
>Resolve https://github.com/ballerina-platform/ballerina-lang/issues/13252

### DIFF
--- a/tests/ballerina-integration-test/src/test/java/org/ballerinalang/test/jms/util/JmsClientHandler.java
+++ b/tests/ballerina-integration-test/src/test/java/org/ballerinalang/test/jms/util/JmsClientHandler.java
@@ -28,10 +28,10 @@ import java.io.File;
 /**
  * Client handler class to handle JMS client related operations.
  */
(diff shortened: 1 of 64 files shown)


tests/ballerina-integration-test/src/test/java/org/ballerinalang/test/jms/util/JmsClientHandler.java
```diff
@@ -28,10 +28,10 @@
 /**
  * Client handler class to handle JMS client related operations.
  */
-public class JMSClientHandler {
+public class JmsClientHandler {
```
Manuri: Just wondering, why are we replacing JMS with Jms? This stands for, Java Message Service Client Handler right?

riyafa: It is to have the same convention in all places. We have classes named as Http* and not HTTP*. One other concern is having capital letters in classes make them harder to find in IDEs.

PR #13244, opened 2019-01-21
Add a record to retrieve mutual ssl status

Bhashinee:
## Purpose
Added a record to hold mutual SSL handshake status in the request object.


stdlib/http/src/main/ballerina/http/http_request.bal
```diff
@@ -416,3 +417,10 @@ public function Request.setPayload(string|xml|json|byte[]|io:ReadableByteChannel
     }
 }
 
+# A record for providing mutual ssl handshake results.
+#
+# + status - Status of the handshake.
+public type MutualSslHandshake record {
+    string status = "";
```
daneshk: Shall we define enum for the status.

PR #13250, opened 2019-01-21
Update ballerina script to take in an external environment variable for classpath extension

lafernando:
This is required in order to pass in an environment variable from outside the ballerina script to be added to the runtime Java classpath. For example, this is required for AWS Lambda integration.


lafernando:
> @lafernando We have to update installers as well. @keizer619 Can you point out correct location for that.

Hi, yes, Tharik already mentioned to me this, and this is already covered in the PR: https://github.com/ballerina-platform/ballerina-distribution/pull/125

PR #13272, opened 2019-01-22
Remove typeTag field of BLangLiteral and introduce BLangNumericLiteral

### LINKED ISSUE #13256: Finite types defined using decimal constants are represented using BFloat instead of BDecimal
```ballerina
const decimal ZERO = 0.0;
const decimal ONE_TWO = 1.2;
const decimal TWO_THREE = 2.3;

type DEC ZERO|ONE_TWO|TWO_THREE;
```
If we debug and check the `DEC` type, it is made up of 3 BFiniteType members. And in the value spaces of these BFiniteType types, there's a single value each. However, this value is a BFloat, when it should actually be a BDecimal, considering the constants' type is explicitly set as decimal.

pranavan15:
## Purpose
This PR removes the `literal.typeTag` as we already have `literal.type.tag` to hold the type tag information and having both is unnecessary and perplexing. This PR also introduces `BLangNumericLiteral`, which extends the `BLangLiteral` to separate the kinds of literals based on whether they are numeric literals or non-numeric literals. This refactor effort is carried out to fix a couple of issues related to the finite types those made up of decimal constants. 
Alongside, this PR also fixes #13256.


pubudu91:
There were a couple of issues with this. Discussed them offline with Pranavan. Those changes need to be incorporated before merging this.

pranavan15:
> There were a couple of issues with this. Discussed them offline with Pranavan. Those changes need to be incorporated before merging this.

@pubudu91: Identified issues have been fixed by the above commits. Could you please review again?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1555,6 +1555,40 @@ public boolean isAssignableToFiniteType(BType type,
...
+     * rather than checking the literal's type's tag. This allows us to assign a float literal directly to a finite
+     * type that consists of a corresponding decimal value. Use this method when we assign a BLangLiteral directly to
+     * a finite type.
+     *
+     * @param type        expected BType
+     * @param literalExpr literal expression
+     * @return returns true if assignable, false otherwise
+     */
+    boolean isLiteralAssignableToFiniteType(BType type,
+                                            BLangLiteral literalExpr) {
```
Shan1024: Let's merge this line with the above line.

tests/ballerina-unit-test/src/test/resources/test-src/types/constant/constant-negative.bal
```diff
@@ -246,3 +246,15 @@ function testStringTypeWithoutType() returns StringTypeWithoutType {
...
+
+const decimal decimalConst1 = 23.2;
+const decimal decimalConst2 = 12.1;
+const float floatConst = 23.2;
+
+type decimalLiteralType decimalConst1 | decimalConst2;
+
+function testAssigningFloatConstToDecimalFiniteType() {
+    decimalLiteralType d = floatConst; // Invalid value assingment.
+}
```
Shan1024: Shall we add a newline here?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1555,6 +1555,40 @@ public boolean isAssignableToFiniteType(BType type,
...
+        if (type.tag == TypeTags.FINITE) {
+            BFiniteType expType = (BFiniteType) type;
+            boolean foundMember = expType.valueSpace
+                    .stream()
```
Shan1024: Shall we move `.stream()` to the line above since it is the convention we use when formatting Java streams.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompiledPackageSymbolEnter.java
```diff
@@ -571,7 +571,8 @@ private void defineValueSpace(DataInputStream dataInStream, BFiniteType finiteTy
                 valueCPIndex = dataInStream.readInt();
                 UTF8CPEntry decimalEntry = (UTF8CPEntry) this.env.constantPool[valueCPIndex];
                 litExpr.value = decimalEntry.getValue();
-                litExpr.typeTag = TypeTags.DECIMAL;
+                // Type tag of the literal is float
+                litExpr.typeTag = TypeTags.FLOAT;
```
hasithaa: Can you explain the logic scenario/behind here?

tests/ballerina-unit-test/src/test/resources/test-src/types/constant/constant-negative.bal
hasithaa: Add new line.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
pranavan15: Sure; Commit ef08dee addresses all the review suggestions.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompiledPackageSymbolEnter.java
pranavan15: We have two different kinds of tags. One is the literal's tag, which we refer by `literal.tag` and the other one is literal's type's tag referred by `literal.type.tag`. The former basically gives the tag of the actual literal, which is `float` in case of both decimal and float literals and the latter denotes the tag of the type of the literal, which should be `decimal` in case of a decimal literal. The same argument holds when a decimal value comes directly from the same module rather coming from a `balo`. 

However, earlier for the decimal constants those come from a `balo` the `literal.tag` was wrongly set. Hence, it is corrected with this PR.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1555,6 +1555,38 @@ public boolean isAssignableToFiniteType(BType type,
...
+     * rather than checking the literal's type's tag. This allows us to assign a float literal directly to a finite
+     * type that consists of a corresponding decimal value. Use this method when we assign a BLangLiteral directly to
+     * a finite type.
+     *
+     * @param type        expected BType
+     * @param literalExpr literal expression
+     * @return returns true if assignable, false otherwise
+     */
+    boolean isLiteralAssignableToFiniteType(BType type, BLangLiteral literalExpr) {
+        if (type.tag == TypeTags.FINITE) {
```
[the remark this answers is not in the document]
pranavan15: The reason why we need two methods is, one method checks the `literal.type.tag` when checking the assignability and the latter uses `literal.typeTag`. Consider the following scenario, where we have a finite type made up of decimal constants and we try to assign a simple floating point literal to a variable of user-defined finite type. 

```ballerina
const decimal CONST1 = 9.23;
const decimal CONST2 = 21.4;

type DEC CONST1|CONST2;

public function main() {
    DEC d1 = 21.4; // Case 1: Expected to be successful

    float f1 = 21.4;
    DEC d2 = f1; // Case 2: Expected to fail at compile time
}
```
As shown above, we need to support assigning a simple floating point literal to the finite type made up of decimal constants. However, simple float literal 21.4 would have its `literal.type.tag`
as float's tag and the finite type `valuespace` values would have `literal.type.tag` of decimal's tag as it is made up of decimal constants. By considering the `literal.typeTag` when checking the assignability we can solve this problem as both decimal and float literals expected to have the same `literal.typeTag`. 

At the same time, we need to fail the case 2 at compile time as it tries to assign a float variable to the finite type made up of decimal types. Hence, in this case, we need to consider the `literal.type.tag` when checking the assignability as it would be different values for decimal and float types.

This is the reason why we need two different methods to handle both the above cases.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompiledPackageSymbolEnter.java
Shan1024: @pranavan15 - Shall we add this as a comment as well?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1555,6 +1555,38 @@ public boolean isAssignableToFiniteType(BType type,
...
+            BFiniteType expType = (BFiniteType) type;
+            boolean foundMember = expType.valueSpace.stream()
+                    .anyMatch(memberLiteral -> {
+                        if (((BLangLiteral) memberLiteral).value == null) {
+                            return literalExpr.value == null;
+                        } else {
```
NatashaWso2: Do we need the else block here since we are returning from if the condition? We can modify the code as below
```java
  if (((BLangLiteral) memberLiteral).value == null) {
     return literalExpr.value == null;
  }
  // Check the literal value and the literal typeTag (For both decimal and float,literal typeTag would be that of the float type)
  return (((BLangLiteral) memberLiteral).value.equals(literalExpr.value) &&
                                    ((BLangLiteral) memberLiteral).typeTag == literalExpr.typeTag);
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompiledPackageSymbolEnter.java
hasithaa: @pranavan15 ,  can we refactor usage of `litExpr.typeTag` as we discussed offline?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/expressions/BLangNumericLiteral.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.wso2.ballerinalang.compiler.tree.expressions;
+
+import org.ballerinalang.model.tree.NodeKind;
+
+/**
+ * Represents numeric literals in ballerina (int, byte, float, decimal).
+ * This class extends {@link BLangLiteral}.
+ *
+ * @since 0.991
```
NatashaWso2:
```suggestion
 * @since 0.990.4
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/codegen/CodeGenerator.java
```diff
@@ -2258,8 +2257,9 @@ private void createObjectTypeTypeDef(BLangTypeDefinition typeDefinition,
             objFieldInfo.fieldType = objField.type;
 
             // Populate default values
-            if (objField.expr != null && (objField.expr.getKind() == NodeKind.LITERAL &&
-                    objField.expr.type.getKind() != TypeKind.ARRAY)) {
+            if (objField.expr != null && (objField.expr.getKind() == NodeKind.LITERAL ||
+                    objField.expr.getKind() == NodeKind.NUMERIC_LITERAL) &&
+                    objField.expr.type.getKind() != TypeKind.ARRAY) {
```
SupunS: shall we move this (condition) to a util method? can re-use it at L#2346 as well.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangPackageBuilder.java
```diff
@@ -1193,10 +1196,16 @@ void addLiteralValue(DiagnosticPos pos, Set<Whitespace> ws, int typeTag, Object
     }
 
     void addLiteralValue(DiagnosticPos pos, Set<Whitespace> ws, int typeTag, Object value, String originalValue) {
-        BLangLiteral litExpr = (BLangLiteral) TreeBuilder.createLiteralExpression();
+        BLangLiteral litExpr;
+        // If numeric literal create a numeric literal expression; otherwise create a literal expression
+        if (typeTag < TypeTags.DECIMAL) {
```
SupunS: shouldn't this be `typeTag <= TypeTags.DECIMAL`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangPackageBuilder.java
```diff
@@ -1193,10 +1196,16 @@ void addLiteralValue(DiagnosticPos pos, Set<Whitespace> ws, int typeTag, Object
...
+            litExpr = (BLangNumericLiteral) TreeBuilder.createNumericLiteralExpression();
+        } else {
+            litExpr = (BLangLiteral) TreeBuilder.createLiteralExpression();
+        }
         litExpr.addWS(ws);
         litExpr.pos = pos;
-        litExpr.typeTag = typeTag;
+        litExpr.type = new BType(typeTag, null);
```
SupunS: type should be taken from the symbolTable.

Applicable for all places

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1403,8 +1403,9 @@ private BType checkStaticMatchPatternLiteralType(BLangExpression expression) {
                 BLangRecordLiteral recordLiteral = (BLangRecordLiteral) expression;
                 recordLiteral.type = new BMapType(TypeTags.MAP, symTable.anydataType, null);
                 for (BLangRecordLiteral.BLangRecordKeyValue recLiteralKeyValue : recordLiteral.keyValuePairs) {
-                    if (recLiteralKeyValue.key.expr.getKind() == NodeKind.SIMPLE_VARIABLE_REF || (
-                            recLiteralKeyValue.key.expr.getKind() == NodeKind.LITERAL
+                    NodeKind kind = recLiteralKeyValue.key.expr.getKind();
+                    if (kind == NodeKind.SIMPLE_VARIABLE_REF || (
+                            (kind == NodeKind.LITERAL || kind == NodeKind.NUMERIC_LITERAL)
                                     && typeChecker.checkExpr(recLiteralKeyValue.key.expr, this.env).tag
                                     == TypeTags.STRING)) {
```
SupunS: better to move the condition to a separate method

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1545,23 +1545,31 @@ private boolean isAssignableToUnionType(BType source, BType target, List<TypePai
...
+            }
+            // Check whether the member literal and the literal that needs to be tested are of same kind and check
+            // the value equality between them.
+            return isSameLiteralKind((BLangLiteral) memberLiteral, literalExpr) &&
+                    ((BLangLiteral) memberLiteral).value.equals(literalExpr.value);
+        }).count();
+
+        // If more than one match means the value space contains ambiguous values.
+        if (matchCount > 1) {
+            dlog.error(literalExpr.pos, DiagnosticCode.AMBIGUOUS_TYPES, type);
```
SupunS: Shall we add some test cases?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1545,23 +1545,31 @@ private boolean isAssignableToUnionType(BType source, BType target, List<TypePai
...
+        }
+        return matchCount == 1;
+    }
+
+    private boolean isSameLiteralKind(BLangLiteral literal1, BLangLiteral literal2) {
```
SupunS: do we need this method? can't we inline it?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangPackageBuilder.java
pranavan15: Since we are at the PackageBuilder level we can't get the type from the SymbolTable for this case. As we are removing the `typeTag` field from the `BLangLiteral` to hold the `type/type tag` information, we have to use the `type` field. However, at this level the type is null. 
To tackle this we require to create a `new BType()` here, which will anyway be modified always at the `TypeChecker` phase.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
pranavan15: This needs to be revisited as per the comments of `James` in the mail thread "Clarification regarding union types" in Ballerina-dev group. Basically, we should not treat this scenario as ambiguous. There will be some other logic changes as well, which I am planning to handle in a different PR. I will add test cases when I revisit this logic for the fulfillness.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangPackageBuilder.java
SupunS: > Since we are at the PackageBuilder level we can't get the type from the SymbolTable for this case.

Why not? :) 
The catch is, value types are singletons, but here we are creating new instances of same type.

It seems the cleanest approach is to have separate kinds of nodes for different literal types. At the moment we are losing this information at packageBuilder, and trying to fix it by the 'type' has caused these complications.. 
But lets do that as an improvement separately.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
SupunS: lets add a test(s) for the current behavior. We can alter it later once we do the improvements

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangPackageBuilder.java
pranavan15: +1; As you suggested the best way to handle this to have separate node kinds for different literal types. Type's tag and Node kind need to be separated out and type related checks/changes should happen only after type checker phase.

PR #13281, opened 2019-01-23
Add subscriber availability check before publishUpdate

chamil321:

### DIFF
--- a/tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
+++ b/tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
@@ -17,8 +17,8 @@
 import ballerina/http;
 import ballerina/io;
 import ballerina/log;
-import ballerina/websub;
 import ballerina/runtime;
+import ballerina/websub;
 
 const string WEBSUB_TOPIC_ONE = "http://one.websub.topic.com";
 const string WEBSUB_TOPIC_TWO = "http://two.websub.topic.com";
@@ -232,23 +232,22 @@ function getPayloadContent(string contentType, string mode) returns string|xml|j
 }
 
 function checkSubscriberAvailability(string topic, string callback) {
+    int count = 0;
     boolean subscriberAvailable = false;
-    while(!subscriberAvailable) {
+    while (!subscriberAvailable && count < 60) {
         websub:SubscriberDetails[] topicDetails = webSubHub.getSubscribers(topic);
         if (isSubscriberAvailable(topicDetails, callback)) {
-            subscriberAvailable = true;
-        } else {
-            runtime:sleep(1000);
+            return;
         }
+        runtime:sleep(1000);
+        count += 1;
     }
 }
 
 function isSubscriberAvailable(websub:SubscriberDetails[] topicDetails, string callback) returns boolean {
     foreach var detail in topicDetails {
-        if(detail.callback == callback) {
+        if (detail.callback == callback) {
             return true;
-        } else {
-            continue;
         }
     }
     return false;
--- a/tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
+++ b/tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
@@ -18,6 +18,7 @@ import ballerina/http;
 import ballerina/io;
 import ballerina/log;
 import ballerina/websub;
+import ballerina/runtime;
 
 const string WEBSUB_TOPIC_ONE = "http://one.websub.topic.com";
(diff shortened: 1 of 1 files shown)


tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
```diff
@@ -18,6 +18,7 @@ import ballerina/http;
 import ballerina/io;
 import ballerina/log;
 import ballerina/websub;
+import ballerina/runtime;
```
MaryamZi: shall we alphabetically order the imports?

tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
```diff
@@ -227,3 +230,26 @@ function getPayloadContent(string contentType, string mode) returns string|xml|j
     error e = error(websub:WEBSUB_ERROR_CODE, { message : errorMessage });
     panic e;
 }
+
+function checkSubscriberAvailability(string topic, string callback) {
+    boolean subscriberAvailable = false;
+    while(!subscriberAvailable) {
+        websub:SubscriberDetails[] topicDetails = webSubHub.getSubscribers(topic);
+        if (isSubscriberAvailable(topicDetails, callback)) {
+            subscriberAvailable = true;
```
MaryamZi: can we return here and remove the else block to have just `runtime:sleep(1000);`?

tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
```diff
@@ -227,3 +230,26 @@ function getPayloadContent(string contentType, string mode) returns string|xml|j
     error e = error(websub:WEBSUB_ERROR_CODE, { message : errorMessage });
     panic e;
 }
+
+function checkSubscriberAvailability(string topic, string callback) {
+    boolean subscriberAvailable = false;
+    while(!subscriberAvailable) {
```
MaryamZi:
```suggestion
    while (!subscriberAvailable) {
```

tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
```diff
@@ -227,3 +230,26 @@ function getPayloadContent(string contentType, string mode) returns string|xml|j
...
+        websub:SubscriberDetails[] topicDetails = webSubHub.getSubscribers(topic);
+        if (isSubscriberAvailable(topicDetails, callback)) {
+            subscriberAvailable = true;
+        } else {
+            runtime:sleep(1000);
+        }
+    }
```
daneshk: Shall we add maximum loop count to avoid endless loop if subscriber not available. if max loop count reached, break the loop and fail the test case.

tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
```diff
@@ -227,3 +230,26 @@ function getPayloadContent(string contentType, string mode) returns string|xml|j
...
+}
+
+function isSubscriberAvailable(websub:SubscriberDetails[] topicDetails, string callback) returns boolean {
+    foreach var detail in topicDetails {
+        if(detail.callback == callback) {
+            return true;
+        } else {
+            continue;
+        }
+    }
```
MaryamZi: Can't we do something like
```ballerina
    foreach var detail in topicDetails {
        if (detail.callback == callback) {
             return true;
         }
    }
    return false;
```

tests/ballerina-integration-test/src/test/resources/websub/services/01_websub_publisher.bal
chamil321: The test fails when the LogLeecher timeouts and task clears during the server cleanup. So should we add maxLoopCount too?

MaryamZi: The LogLeecher time out would be after we respond to the request at https://github.com/ballerina-platform/ballerina-lang/pull/13281/files#diff-d319b3fb1a8c4ff418213f19622aaadaR155 right? So the test will fail when HttpClientRequest.doPost times out?

+1 to add a maximum count here instead of relying on timing out.

chamil321: HttpClientRequest timeout is 30s. But the Logleecher timeout is 45s. So the test depends on  Logleecher timeout basically.

However added the MaxCount for safety purposes ;)

PR #13286, opened 2019-01-23
[WIP] Align the implicit initial values of types with the spec

### LINKED ISSUE #13166: [Spec Deviation] Implicit initial value of tuple, map, record, table, typedesc, singleton, error, object and union not working as expected
**Description:**
```ballerina
type One 1;
type Union 0|1|2;

@test:Config {}
function testImplicitInitialValuesBroken() {
    // Implicit initial value of a tuple should be a tuple value with its member values taking their respective implicit initial values
    (int, boolean, string)[] tupleArray = [];
    tupleArray[1] = (200, true, "test string");
    test:assertEquals(tupleArray[0], (),
        msg = "expected implicit initial value of (int, boolean, string) to be (0, false, \"\")");

    // Implicit initial value of a map should be an empty map
    map<any>[] mapArray = [];
    mapArray[1] = { fieldOne: "valueOne" };
    map<any> expectedMap = {};
    test:assertEquals(mapArray[0], (), msg = "expected implicit initial value of map to be {}");

    // Implicit initial value of a record should be a record where each field has the implicit initial value for its type
    utils:FooRecord[] fooRecordArray = [];
    fooRecordArray[1] = { fooFieldOne: "valueOne" };
    test:assertEquals(fooRecordArray[0], (),
        msg = "expected implicit initial value of FooRecord to { fooFieldOne: \"\" }");

    // Implicit initial value of a table should be a table with no rows
    table<utils:BarRecord>[] tableArray = [];
    tableArray[1] = table{};
    test:assertEquals(tableArray[0], (), msg = "expected implicit initial value of string to be an empty string");

    // Implicit initial value of typedesc should be ()
    //typedesc[] typedescArray = [];
    //typedescArray[1] = int;
   
    // Implicit initial value of a singleton type should be the single value
    One[] singletonArray = [];
    singletonArray[1] = 1;
    test:assertEquals(singletonArray[0], (),
        msg = "expected implicit initial value of a singleton to be the singleton value");

    error[] errorArray = []; // should fail at compile time

    // The rules for the implicit initial value of a union type are as follows:
    // 1. if () is in the value space of a union type, then the implicit initial value for the union type is ();
    // 2. otherwise if the value space only contains only values of a single basic type, and one of those values is the implicit initial value of that basic type, then the implicit initial value for the union is the implicit initial value of the basic type of its values (for example, the implicit initial value for 0|1 is 0)
    // 3. otherwise the union type does not have an implicit initial value.
    Union[] unionArray2 = [];
    unionArray2[1] = 2;
    test:assertEquals(unionArray2[0], (), msg = "expected implicit initial value of this union should be 0");
}
```
**Affected Versions:**
0.990.2

pubudu91:

### DIFF
--- a/bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BArrayType.java
+++ b/bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BArrayType.java
@@ -98,7 +98,7 @@ public class BArrayType extends BType implements BIndexedType {
             case TypeTags.DECIMAL_TAG:
                 return (V) new BDecimalArray();
             default:
-                return (V) new BValueArray();
+                return (V) new BValueArray(this);
         }
     }
 
--- a/bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BFiniteType.java
+++ b/bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BFiniteType.java
@@ -80,7 +80,7 @@ public class BFiniteType extends BType {
 
         Iterator<BValue> valueIterator = valueSpace.iterator();
         BValue firstVal = valueIterator.next();
-        BValue implicitInitValOfType = firstVal.getType().getZeroValue();
+        BValue implicitInitValOfType = firstVal.getType().getEmptyValue();
 
         if (implicitInitValOfType.equals(firstVal)) {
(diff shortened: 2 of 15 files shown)


pubudu91:
Closing this for now. Will update and reopen once PR #13193 gets merged.

PR #13294, opened 2019-01-24
Emit a () when generating array's string representation

rdhananjaya:
## Purpose
An array that contains a nil value has an invalid string representation.
This PR fixes that.

```ballerina
    string?[] ar = ["abc", "d", (), "s"];
    io:println(ar);
```
Previous output: 
["abc", "d", "s"]

With this PR, new output:
["abc", "d", (), "s"]

This PR depends on https://github.com/ballerina-platform/ballerina-lang/pull/13193 and https://github.com/ballerina-platform/ballerina-lang/pull/13286


SupunS:
Is this issue only specific to arrays? Have we checked the behavior for other scenarios as well?
eg: printing just nil, nil values in other structural types, etc..

MaryamZi:
Should we be printing nothing instead of `()` (`["abc", "d", , "s"]`)? 

For example with this approach, how do we differentiate between
```ballerina
io:println("()");
```

and 
```ballerina
io:println(());
```

rdhananjaya:
> Should we be printing nothing instead of `()` (`["abc", "d", , "s"]`)?
> 
> For example with this approach, how do we differentiate between
> 
> ```ballerina
> io:println("()");
> ```
> 
> and
> 
> ```ballerina
> io:println(());
> ```

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArrayTest.java
```diff
@@ -102,6 +102,13 @@ public void testJSONArrayLength() {
         Assert.assertEquals(((BInteger) returnVals[0]).intValue(), 2, "Length didn't match");
         Assert.assertEquals(((BInteger) returnVals[1]).intValue(), 3, "Length didn't match");
     }
+
+    @Test()
```
NatashaWso2: Shall we add a description for the test?

hasithaa:
Some improvements points were discussed here. https://groups.google.com/d/msg/ballerina-dev/4QEGRga-nT4/7L2V1JfnAgAJ Shall we create an issue to capture this?

PR #13296, opened 2019-01-24
Update values types and variables module in spec tests

Anoukh:


spec-conformance-tests/S05_values_types_variables/tests/initial_implicit_value.bal
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) 2019 WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
MaryamZi: should this file be `implicit_initial_values.bal` instead?

PR #13300, opened 2019-01-24
API Designer Bug Fixes

DimalChandrasiri:


composer/packages/api-designer/src/api-designer/components/context/open-api-context.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
nadeeshaan: Change year

composer/packages/api-designer/src/api-designer/components/header/header.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from "react";
```
NatashaWso2: Shall we add license headers for all the newly added files?

composer/packages/api-designer/src/api-designer/components/context/open-api-context.ts
DimalChandrasiri: Fixed with commit 4d0cc36

composer/packages/api-designer/src/api-designer/components/header/header.tsx
DimalChandrasiri: Fixed with commit 4d0cc36

PR #13303, opened 2019-01-25
[IntelliJ Plugin]

NipunaRanasinghe:
## Purpose
This PR updates IDEA plugin version and change notes.

### DIFF
--- a/tool-plugins/intellij/gradle.properties
+++ b/tool-plugins/intellij/gradle.properties
@@ -15,6 +15,6 @@
 #
 
 ballerinaPluginName = ballerina-intellij-idea-plugin
-ballerinaPluginVersion = 0.990.1
+ballerinaPluginVersion = 0.991.0
 
 ideaVersion = 2017.3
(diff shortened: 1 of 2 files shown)


tool-plugins/intellij/gradle.properties
```diff
@@ -15,6 +15,6 @@
 #
 
 ballerinaPluginName = ballerina-intellij-idea-plugin
-ballerinaPluginVersion = 0.990.1
+ballerinaPluginVersion = 0.991.0
```
hasithaa: this has to be 0.990-r3

NipunaRanasinghe: As per the discussion `Ballerina versioning strategy for future releases`,  there is an issue with vscode plugin where as npm package.json supports semver. It does not allow versioning like 0.990-r3 as it has two digits. In that case for ballerina ide plugins (vscode and idea) we'll be using semver and represent r1 as part of patch number, as follows.

|Ballerina version |	VSCode version|
|--                     |--                     |
| 0.990-r3	| 0.990.3 |
| 0.995-r1 | 0.990.1 |

Therefore updated the version accordingly.

PR #13308, opened 2019-01-25
Add map support in constants

Shan1024:
## Purpose
This PR adds map support in constants.

Eg - 
```ballerina
const map<string> m1 = { "k": "v" };

const value = "val";
const map<string> m2 = { "k": value };

const map<map<string>> m3 = { "k": m2 };
```


Shan1024:
Closing since some design level changes have been discussed.

PR #13339, opened 2019-01-28
Improve arrays BBE description

rdhananjaya:
## Purpose
Improve description about unbound arrays

### DIFF
--- a/examples/arrays/arrays.bal
+++ b/examples/arrays/arrays.bal
@@ -10,7 +10,7 @@ public function main() {
     io:println(b[0]);
     io:println(b.length());
 
-    // Unless length is explicitly specified or is expected to be inferred, arrays are unbounded in length.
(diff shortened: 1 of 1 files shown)


examples/arrays/arrays.bal
```diff
@@ -10,8 +10,9 @@ public function main() {
     io:println(b[0]);
     io:println(b.length());
 
-    // Arrays are unbounded in length. They can grow up to any length based on 
-    // the given index. In this example, the length of the array is 1000.
+    // Unless length is specified, arrays are unbounded in length.
```
MaryamZi: Should we say something like
```suggestion
    // Unless length is explicitly specified or is expected to be inferred, arrays are unbounded in length.
```

praneesha:
```suggestion
    // Unless the length is specified, arrays are unbounded in length.
```

praneesha: +1 for this:

PR #13369, opened 2019-01-28
Add BBE tests for Type Check Expression

warunalakshitha:
## Purpose
$Subject

### DIFF
--- a/examples/type-check-expression/tests/type_check_expression_test.bal
+++ b/examples/type-check-expression/tests/type_check_expression_test.bal
@@ -1,0 +1,31 @@
+import ballerina/test;
+import ballerina/io;
+
+(any|error)[] outputs = [];
+int counter = 0;
+
+// This is the mock function that replaces the real function.
+@test:Mock {
+    moduleName: "ballerina/io",
+    functionName: "println"
+}
+public function mockPrint(any|error... s) {
+    foreach var elem in s {
+        outputs[counter] = elem;
+        counter += 1;
+    }
+}
+
+@test:Config
+function testFunc() {
+    // Invoking the main function
+    main();
+    io:println(outputs[1]);
+    test:assertEquals(outputs[0], "Is 'a' a string? ");
+    test:assertEquals(outputs[1], true);
+    test:assertEquals(outputs[2], "'a' is a string with value: ");
+    test:assertEquals(outputs[3], "Hello, world!");
+    test:assertEquals(outputs[4], "Alex is a student");
+    test:assertEquals(outputs[5], "Alex is a person");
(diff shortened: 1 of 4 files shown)


MaryamZi:
Is this BBE listed on bio? 

We have https://ballerina.io/learn/by-example/type-guard.html for the `is` check.

SupunS:
we can do 'type test' for any expression. eg: `foo() is T`. 'implicit type narrowing' will not always applies to type test expressions.
Type guard is a special case of type test, where the expr is a variable-ref.

May be we should update the example stating that?

examples/type-check-expression/type_check_expression.description
```diff
@@ -1,2 +1,2 @@
-// Type check expression can be used to assert the runtime type of a value in Ballerina.
+// Type check expression can be used to assert the runtime type of a value or any expression in Ballerina.
```
SupunS: The previous statement sound right to me. It always checks the type of a value at runtime.

Being able to give any expression during compile time is different to that IMO.

examples/type-check-expression/tests/type_check_expression_test.bal
```diff
@@ -0,0 +1,35 @@
...
+public function mockPrint(any|error... s) {
+    foreach var elem in s {
+        outputs[counter] = elem;
+        counter += 1;
+    }
+}
+
+@test:Config
+function testFunc() {
+    // Invoking the main function
```
Shan1024:
```suggestion
    // Invoking the main function.
```

warunalakshitha: Fixed

examples/type-check-expression/type_check_expression.description
warunalakshitha: Revert to previous statement.

examples/type-check-expression/type_check_expression.bal
```diff
@@ -24,7 +24,7 @@ public function main() {
     if (x is Student) {
         io:println("Alex is a student");
     } else {
-        io:println("Alex is a student");
+        io:println("Alex is not a student");
     }
 
     // Type of `x` is `Student`. But it is structurally equivalent to `Person`.
```
praneesha: Type of `x` is `Student`. However, it is structurally equivalent to `Person`.

examples/type-check-expression/type_check_expression.bal
```diff
@@ -41,6 +41,12 @@ public function main() {
     } else {
         io:println("Alex is not a vehicle");
     }
+
+    // Type check expression can be used for any expression.
```
praneesha: Type `check` can be used for any expression.

PR #13389, opened 2019-01-29
[IntelliJ Plugin] Add minor fixes

NipunaRanasinghe:
## Purpose
This PR adds minor fixes improvements related to deprecated usages, LS based signature help, hover support and new ballerina versioning strategy.

### DIFF
--- a/tool-plugins/intellij/src/main/java/io/ballerina/plugins/idea/inspections/VersionMismatchNotificationProvider.java
+++ b/tool-plugins/intellij/src/main/java/io/ballerina/plugins/idea/inspections/VersionMismatchNotificationProvider.java
@@ -84,7 +84,7 @@ public class VersionMismatchNotificationProvider extends EditorNotifications.Pro
 
     @NotNull
     private String getMinorVersion(@NotNull String version) {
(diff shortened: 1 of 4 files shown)


tool-plugins/intellij/src/main/java/io/ballerina/plugins/idea/inspections/VersionMismatchNotificationProvider.java
```diff
@@ -84,7 +84,7 @@ private static String getBallerinaPluginVersion() {
 
     @NotNull
     private String getMinorVersion(@NotNull String version) {
-        return version.contains(".") ? version.split("[.]")[1] : "";
+        return version.contains(".") ? version.replace('-','.').split("[.]")[1] : "";
```
Shan1024:
```suggestion
        return version.contains(".") ? version.replace('-', '.').split("[.]")[1] : "";
```

Don't we need a space? Is the gradle checkstyle plugin is running?

NipunaRanasinghe: It was disabled locally due a config change. Fixed the checkstyle issue

PR #13398, opened 2019-01-29
Remove freeze and isFrozen instructions

vinok88:

### DIFF
--- a/bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
+++ b/bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
@@ -513,10 +513,6 @@ public class BVM {
                 case InstructionCodes.DETAIL:
                     handleErrorBuiltinMethods(opcode, operands, sf);
                     break;
-                case InstructionCodes.IS_FROZEN:
-                case InstructionCodes.FREEZE:
-                    handleFreezeBuiltinMethods(strand, opcode, operands, sf);
-                    break;
                 case InstructionCodes.STAMP:
                     handleStampBuildInMethod(strand, operands, sf);
                     break;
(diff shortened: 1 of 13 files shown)


bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
```diff
@@ -5399,11 +5356,11 @@ public FreezeStatus(State state) {
             this.currentState = state;
         }
 
-        private void setFrozen() {
+        public void setFrozen() {
```
MaryamZi: Is there a possibility of moving `FreezeStatus` somewhere else and leaving these methods private, or will there be a circular dependency? 

Since this is now public, the freeze status can be modified via the BValue right? This was left private to only allow modification via the `.freeze()` call itself.

stdlib/utils/src/main/ballerina/utils/built-in-methods.bal
```diff
@@ -19,3 +19,15 @@
...
+#
+# + value - Value to be frozen
+# + return - Frozen value
+extern function freeze(anydata value) returns anydata;
+
+# Check freeze status of given value
+#
+# + value - Value to check freeze status
+# + return - True for a frozen value
+extern function isFrozen(anydata value) returns boolean;
```
MaryamZi: needs a new line?

stdlib/utils/src/main/ballerina/utils/built-in-methods.bal
```diff
@@ -19,3 +19,15 @@
...
 # + return - Clone of the given value
 extern function clone(anydata value) returns anydata;
+
+# Freeze a given value.
+#
+# + value - Value to be frozen
+# + return - Frozen value
+extern function freeze(anydata value) returns anydata;
+
+# Check freeze status of given value
```
MaryamZi: Shall we say 
```
# Check the freeze status of a given value.
```

stdlib/utils/src/main/java/org/ballerinalang/utils/IsFrozen.java
```diff
@@ -0,0 +1,48 @@
...
+        packageName = "utils",
+        functionName = "isFrozen",
+        args = {@Argument(name = "value", type = TypeKind.ANYDATA)},
+        returnType = {@ReturnType(type = TypeKind.BOOLEAN)})
+public class IsFrozen extends BlockingNativeCallableUnit {
+
+    @Override
+    public void execute(Context ctx) {
+        BValue value = ctx.getNullableRefArgument(0);
+        boolean frozen = (value == null || value.isFrozen()) ? true : false;
```
MaryamZi: Shall we change this to
```java
boolean frozen = value == null || value.isFrozen();
```

stdlib/utils/src/main/ballerina/utils/built-in-methods.bal
```diff
@@ -19,3 +19,15 @@
 # + value - Value to be cloned
 # + return - Clone of the given value
 extern function clone(anydata value) returns anydata;
+
+# Freeze a given value.
+#
+# + value - Value to be frozen
+# + return - Frozen value
+extern function freeze(anydata value) returns anydata;
```
MaryamZi: We also allow `.freeze()` and `.isFrozen()` on non-anydata typed variables which can hold `anydata` values.
For example, a `map<any>` that only holds `anydata` values.

Isn't saying `anydata` here an issue?

bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
vinok88: In order to keep it private, we need to move FreezeStatus to utils stdlib. Then we need to add a dependency from core to stdlib, which will introduce a circular dependency. Hence, it's not possible.

stdlib/utils/src/main/ballerina/utils/built-in-methods.bal
vinok88: Have we covered them in tests? All the tests got passed. What is the datatype we will use here then?

MaryamZi: Yeah these are covered, for example:
- https://github.com/ballerina-platform/ballerina-lang/blob/refactor_builtin_functions/tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/builtinfunctions/FreezeAndIsFrozenTest.java#L367 
- https://github.com/ballerina-platform/ballerina-lang/blob/refactor_builtin_functions/tests/ballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/freeze-and-isfrozen.bal#L399

In such scenarios, the return type is `T|error`, not just `T` too. So the return type could be `any|error` too, not just `any`.

Wonder how this worked in the first place?

vinok88: There is an issue with the test then :) Will fix the method anyway.

MaryamZi: Ideally compilation should have failed right? According to the log, `freeze-and-isfrozen.bal` seems to have compiled without issues. Maybe the types specified here are not validated?

PR #13401, opened 2019-01-30
Remove Length instruction

warunalakshitha:
## Purpose
This will remove LENGTH instruction specifically used for v.length() built in method. Method will be desugar into extern private function. For string we will reuse built in string.length method. 

This also refactor length() and size() methods in BValues and rename them to size() to have generic length method.


stdlib/utils/src/main/java/org/ballerinalang/utils/Length.java
```diff
@@ -0,0 +1,56 @@
...
+import org.ballerinalang.natives.annotations.ReturnType;
+import org.ballerinalang.util.exceptions.BLangRuntimeException;
+
+import static org.ballerinalang.bre.bvm.BLangVMErrors.NULL_REF_EXCEPTION;
+
+/**
+ * Returns an integer representing the number of items that a ref value contains, where the meaning of item depends on
+ * the basic type of value.
+ *
+ * @since 0.991.0
```
NatashaWso2: Shall we update the version to the latest?

warunalakshitha: Fixed with moving version 0.990.4 since this will be added to 0.990.4. Thanks

PR #13405, opened 2019-01-30
Remove REASON,DETAIL instructioin codes

vinok88:
Introduce native functions instead.

### DIFF
--- a/stdlib/utils/src/main/java/org/ballerinalang/utils/Detail.java
+++ b/stdlib/utils/src/main/java/org/ballerinalang/utils/Detail.java
@@ -1,0 +1,49 @@
+/**
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * <p>
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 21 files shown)


stdlib/utils/src/main/java/org/ballerinalang/utils/Detail.java
```diff
@@ -0,0 +1,49 @@
...
+        functionName = "detail",
+        args = {@Argument(name = "value", type = TypeKind.ERROR)},
+        returnType = {@ReturnType(type = TypeKind.ANYDATA)})
+public class Detail extends BlockingNativeCallableUnit {
+
+    @Override
+    public void execute(Context context) {
+        BValue refRegVal = context.getNullableRefArgument(0);
+        if (refRegVal == null) {
+            return;
```
warunalakshitha: Don't we need to  throw  BLangRuntimeException with BLangVMErrors.NULL_REF_EXCEPTION here?

warunalakshitha:
Reason.java and Detail.java code coverage is less than 80%. shall we check that. We might need to write some tests for those( ex. null values)

stdlib/utils/src/main/java/org/ballerinalang/utils/Detail.java
vinok88: Removed null checks since the existing BVM impl do not have any null chacks.

PR #13408, opened 2019-01-30
Update missing spec tests and remove assertErrorReason util method

Anoukh:
## Purpose
Update missing spec tests and remove assertErrorReason util method


spec-conformance-tests/S05_behavioral_values/tests/services.bal
```diff
@@ -0,0 +1,23 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+import ballerina/test;
+
+// service-type-descriptor := service
+service sampleService = service {
+    function serviceFunction() returns int {
```
MaryamZi: shall we add a `resource function` also? We can just add one similar to a normal function right?

spec-conformance-tests/S05_behavioral_values/tests/functions/function_parameters_and_arguments.bal
```diff
@@ -100,7 +102,7 @@ function testFunctionInvocationWithAllArgs() {
         msg = EXPECTED_RETURN_VALUE_TO_BE_EQUAL_TO_SUM_FAILURE_MESSAGE);
 }
 
-function funcWithNamedAndRestParams(int i, int j = 0, int k = 0, int l, int... m) returns int {
+function funcWithNamedAndRestParams(int i, int j = 0, int k = 10, int l, int... m) returns int {
```
MaryamZi: Shall we have `10` as a constant, and use it here and in the tests? O:)

PR #13418, opened 2019-01-30
Fix SO error when calling same function inside mocking function

rasika:

### DIFF
--- a/misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestAnnotationProcessor.java
+++ b/misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestAnnotationProcessor.java
@@ -271,9 +271,9 @@ public class TestAnnotationProcessor extends AbstractCompilerPlugin {
             }
 
             for (PackageInfo packageInfo : programFile.getPackageInfoEntries()) {
-                int instrEndLimit = getInstructionsEndLimit(packageInfo);
+                int limit = getTestInstructionsPosition(packageInfo);
                 Instruction[] instructions = packageInfo.getInstructions();
-                for (int i = 0; i < instrEndLimit; i++) {
+                for (int i = 0; i < limit; i++) {
                     Instruction ins = instructions[i];
                     if (ins instanceof Instruction.InstructionCALL) {
                         // replace the function pointer of the instruction with the mock function pointer
@@ -544,7 +544,7 @@ public class TestAnnotationProcessor extends AbstractCompilerPlugin {
         return bLangPackage.packageID.toString();
     }
 
-    private static int getInstructionsEndLimit(PackageInfo packageInfo) {
+    private static int getTestInstructionsPosition(PackageInfo packageInfo) {
         FunctionInfo testInitFunctionInfo = packageInfo.getTestInitFunctionInfo();
         if (testInitFunctionInfo != null) {
             return testInitFunctionInfo.getDefaultWorkerInfo().getCodeAttributeInfo().getCodeAddrs();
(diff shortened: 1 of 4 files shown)


misc/testerina/modules/testerina-core/src/test/resources/functionmocktest4.pkg/tests/test-sub-program.bal
```diff
@@ -0,0 +1,20 @@
+
```
Shan1024: Shall we remove this newline?

misc/testerina/modules/testerina-core/src/test/resources/functionmocktest4.pkg/sub-program.bal
```diff
@@ -0,0 +1,6 @@
+import ballerina/io;
+
+// this function should not get called as mocked
```
Shan1024:
```suggestion
// This function should not get called as mocked.
```

misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestAnnotationProcessor.java
```diff
@@ -540,4 +543,12 @@ private String getPackageName(PackageNode packageNode) {
         BLangPackage bLangPackage = ((BLangPackage) packageNode);
         return bLangPackage.packageID.toString();
     }
+
+    private static int getInstructionsEndLimit(PackageInfo packageInfo) {
```
Shan1024: Better to add some explanation.

misc/testerina/modules/testerina-core/src/test/resources/functionmocktest4.pkg/tests/test-sub-program.bal
```diff
@@ -0,0 +1,20 @@
...
+}
+public function mockPrint(any... s) {
+    io:println("io:println: Mocked");
+    success = true;
+}
+
+@test:Config{}
+function testAssertIntEquals () {
+    printSomething();
+    test:assertTrue(success, msg = "using same function inside mocking function failed");
```
NatashaWso2:
```suggestion
    test:assertTrue(success, msg = "using the same function inside the mocking function failed");
```

misc/testerina/modules/testerina-core/src/test/resources/functionmocktest4.pkg/tests/test-sub-program.bal
```diff
@@ -0,0 +1,20 @@
...
+import ballerina/io;
+
+boolean success = false;
+
+@test:Mock {
+    moduleName: "ballerina/io",
+    functionName: "println"
+}
+public function mockPrint(any... s) {
+    io:println("io:println: Mocked");
```
NatashaWso2:
```suggestion
    io:println("The function 'io:println' was mocked");
```

misc/testerina/modules/testerina-core/src/test/resources/functionmocktest4.pkg/sub-program.bal
NatashaWso2:
```suggestion
// This function should not get called as it is being mocked.
```

misc/testerina/modules/testerina-core/src/test/resources/functionmocktest4.pkg/tests/test-sub-program.bal
rasika: Removed in https://github.com/ballerina-platform/ballerina-lang/pull/13418/commits/8f2497b1b2034b18e20f61a0e73bbc140bafb680

misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestAnnotationProcessor.java
rasika: Since this is a private function I am bit reluctant to add explanations. As per my belief most of the time just the name of the function should be expressible enough. WDYT? Renamed the function name  in https://github.com/ballerina-platform/ballerina-lang/pull/13418/commits/8901a7c96c5aaba4d0958db616d4f4333c19be3e.

hasithaa:
@rasika This is based on an Assumption, where we mock only function calls in the module, except in test sources. is that correct?

rasika:
@hasithaa Correct. We don't need to mock the **target function** inside the **mocking function**. This also reduces the work-load. For the same example I used in PR; I don't need to implement a new `io:println` inside the mocking function. Rather I can tweak the parameters for the target function.

PR #13161, opened 2019-01-13
[Do Not Merge] Replace type assertion/explicitly typed expressions with the type cast expression

MaryamZi:
## Purpose
This PR replaces type assertion and explicitly typed expressions with the type cast expression as specified in the v0.990:2018-12-DD spec.

To Do:
- [ ] **update balo version**
- [x] add float to byte, byte to float, decimal to byte and byte to decimal conversions
- [ ] update type assertion BBE as the type cast BBE
- [x] rename op code
- [ ] rename method names - assertion to cast
- [ ] rename conversion nodes as cast nodes
- [ ] **NaN/Infinite decimal to int/byte**

Issue #12929, opened 2018-12-24
Validate assertion as union types, and fail disallowed assertions at compile time


MaryamZi:
**Description:**
Currently, type assertion looks for an exact match between asserted type and storage type.

Thus asserting to a union type will always panic:

For example:
```ballerina
public function main() {
    any a = "hi";
    string? s = <string?> a; // string vs string?
}
```

However currently this does not fail at compile time. We need to finalize obvious panic scenarios with assertion to union types, and fail compilation.

**Affected Versions:**
0.990.1

MaryamZi:
This is no longer valid with the changes to the type cast expression that will be introduced with #13479. 

The said cast is now valid.

PR #13479, opened 2019-02-06
Replace type assertion/explicitly typed expressions with the type cast expression

MaryamZi:
Opening a new PR in place of #13161.

Invalidates https://github.com/ballerina-platform/ballerina-lang/issues/12929

## Purpose
This PR replaces type assertion and explicitly typed expressions with the type cast expression as specified in the v0.990.0 spec.

## Examples
```ballerina
import ballerina/io;

public function main() {
    // 1) OK
    any a = "this is a string value";
    string? s = <string?> a;  // The value (string value) belongs to the type string?. 
    io:println("1: ", s); 

    // 2) OK
    a = (); // nil value
    s = <string?> a; // The value (nil value) belongs to the type string?. 
    io:println("2: ", s); 

    // 3) OK, numeric conversion from int to float
    int i = 10;
    float f = <float> i; 
    io:println("3: ", f); 

    // 4) OK, numeric conversion from int to float
    i = 10;
    string|float sf = <string|float> i; 
    io:println("4: ", sf); 

    // // 5) Compilation failure, because 'float|decimal' contains more than one numeric type.
    // i = 10;
    // float|decimal d = <float|decimal> i;

    // // 6) Compilation failure, int does not belong to string, nor can int hold a string.
    // i = 10;
    // string s2 = <string> i;

    // // 7) Panic since int does not belong to string.
    // a = 10;
    // string s3 = <string> a;
    // io:println("7: ", s3); 

    // 8) Unions with no direct match
    string|boolean sb = "string as string|boolean";
    string|map<string|int> sm = <string|map<string|int>> sb;
    io:println("8: ", sm);

    // 9) Unions with no direct match - 2
    BarRecord b1 = { name: "em", age:20 };
    FooRecord|string v1 = b1;
    BarRecord|int v2 = <BarRecord|int> v1; 
    io:println("9: ", v2);

    // // 10) Write failure scenario
    // map<string> m = { one: "test string" };
    // map<string|int> m2 = <map<string|int>> m;
    // io:println("10: ", m2.one);
    // m2.two = 1;

    // 11) Target type with more than one numeric type allowed
    a = 100;
    int|float u = <int|float> a;
    io:println("11: ", u);
}

type FooRecord record {
    string name;
};

type BarRecord record {
    string name;
    int age;
};
```

## To Do
- [x] **update balo version**
- [x] update type assertion BBE as the type cast BBE - changes on https://github.com/MaryamZi/ballerina/tree/fix-type-cast-BBE, will send a separate PR.
- [ ] rename method names - assertion to cast
- [ ] rename conversion nodes as cast nodes


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -539,34 +540,41 @@ BOperatorSymbol createBuiltinMethodSymbol(BLangBuiltInMethod method, BType type,
...
-        int sourceTypeTag = sourceType.tag;
-        if (types.isValueType(sourceType)) {
-            if (sourceType == targetType) {
-                return Symbols.createCastOperatorSymbol(sourceType, targetType, symTable.errorType, false, true, 
-                                                        InstructionCodes.NOP, null, null);
-            }
+    BSymbol getNumericConversionOrCastSymbol(BLangTypeConversionExpr conversionExpr, BType sourceType,
+                                             BType targetType) {
+        if (targetType.tag == TypeTags.UNION &&
+                ((BUnionType) targetType).memberTypes.stream()
```
Shan1024: Better to refactor the second condition to a separate variable.

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/typecast/TypeCastExpressionsTest.java
```diff
@@ -0,0 +1,313 @@
+/*
+ * Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
Shan1024:
```suggestion
 * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/typecast/TypeCastExpressionsTest.java
```diff
@@ -0,0 +1,313 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.ballerinalang.launcher.util.BAssertUtil.validateError;
+
+/**
+ * Class to test type cast expressions.
+ *
+ * @since 0.985.0
```
warunalakshitha: Version seems to be incorrect.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
MaryamZi: Do you mean moving
```java
((BUnionType) targetType).memberTypes.stream()
                        .filter(memType -> types.isBasicNumericType(memType)).count() > 1
```

This depends on the first condition (`targetType.tag == TypeTags.UNION`) being true - since we are casting to union type.

Shan1024: Yes, I meant the stream. Maybe we can split the 2nd conditions and move it inside the `if` condition and use a nested `if` block? I don't think we've used an stream result as an condition anywhere else (not 100% sure).

@SupunS - WDYT?

PR #13481, opened 2019-02-06
Code refactor and rename constants and bal file

Kishanthan:
Code refactoring and proper naming of constants and bal files


compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_constants.bal
```diff
@@ -0,0 +1,195 @@
+// jvm bytecode instructions
```
manuranga: would it be a good idea to prefix to indicate type? since there seems to be multiple kinds of consts.

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_constants.bal
```diff
@@ -0,0 +1,195 @@
...
+const int ACC_SYNTHETIC = 4096;
+const int ACC_ANNOTATION = 8192;
+const int ACC_ENUM = 16384;
+const int ACC_MANDATED = 32768;
+const int ACC_MODULE = 32768;
+const int ACC_DEPRECATED = 131072;
+
+
+// jump instruction related constants
+const int JUMP = 0;
```
manuranga: How about using the java byte code spec value `167` here, then we don't need a switch case in java side.

Issue #13493, opened 2019-02-06
LocksInMainTest.testThrowErrorInsideLock test fails intermittently


vinok88:
**Description:**
ERROR] testThrowErrorInsideLock(org.ballerinalang.test.lock.LocksInMainTest)  Time elapsed: 0.545 s  <<< FAILURE!
java.lang.AssertionError: expected [51] but found [50]
        at org.ballerinalang.test.lock.LocksInMainTest.testThrowErrorInsideLock(LocksInMainTest.java:95)

PR #13502, opened 2019-02-07
Fix intermittent failure in LocksInMainTest.testThrowErrorInsideLock

vinok88:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13493


tests/ballerina-unit-test/src/test/resources/test-src/lock/locks-in-functions.bal
```diff
@@ -106,6 +106,7 @@ function throwErrorInsideLock() returns (int, string) {
         lockWithinLockString1 = "second worker string";
         lockWithinLockInt1 = lockWithinLockInt1 + 50;
     }
+    error? waitResult = trap wait w1;
```
Shan1024: Don't we need to check whether any error has occurred?

vinok88: We have written this to intentionally get an error. But this is a lock test. No point of checking it. We only need to check the locks are getting released properly after a panic within a lock.

Issue #13501, opened 2019-02-06
WaitForAnyActionsTest test case fails intermittently 


anupama-pathirage:
**Description:**
```
[ERROR] Tests run: 4112, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 259.821 s <<< FAILURE! - in TestSuite
[ERROR] waitTest21(org.ballerinalang.test.worker.WaitForAnyActionsTest)  Time elapsed: 0.017 s  <<< FAILURE!
org.testng.TestException: 

Method WaitForAnyActionsTest.waitTest21()[pri:0, instance:org.ballerinalang.test.worker.WaitForAnyActionsTest@4f0ee1eb] should have thrown an exception of type class org.ballerinalang.util.exceptions.BLangRuntimeException

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   WaitForAnyActionsTest.waitTest21 Â» Test 
Method WaitForAnyActionsTest.waitTest...
[INFO] 
[ERROR] Tests run: 4112, Failures: 1, Errors: 0, Skipped: 0
```

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Issue #13490, opened 2019-02-06
WaitForAllWorkersTest.sendToFork test fails intermittently


vinok88:
**Description:**
Error:
"[ERROR] sendToFork(org.ballerinalang.test.worker.WaitForAllWorkersTest)  Time elapsed: 0.002 s  <<< FAILURE!
java.lang.AssertionError: 
expected [Finishing Default Worker
Finishing Worker w2
] but found [Finishing Default Worker
]
	at org.ballerinalang.test.worker.WaitForAllWorkersTest.sendToFork(WaitForAllWorkersTest.java:63)"

Issue #13463, opened 2019-02-05
ballerina-examples/async failing intermittently


Shan1024:
**Description:**
```
Generating executable
    ./target/fork-variable-access.balx
Compiling source
    ballerina-examples/async:0.0.1

Running tests
    ballerina-examples/async:0.0.1
	[fail] testFunc:
	    error: not enough format arguments {}
	    	at ballerina/io:sprintf(<native>)
	    	ballerina/test:assertEquals(assert.bal:80)
	    	ballerina-examples/async:0.0.1:testFunc(tests/async_test.bal:33)

	0 passing
	1 failing
	0 skipped

[ERROR] Command execution failed.
org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1)
    at org.apache.commons.exec.DefaultExecutor.executeInternal (DefaultExecutor.java:404)
    at org.apache.commons.exec.DefaultExecutor.execute (DefaultExecutor.java:166)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:804)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:751)
    at org.codehaus.mojo.exec.ExecMojo.execute (ExecMojo.java:313)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:134)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:208)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:51)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:309)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:194)
[...]

PR #13513, opened 2019-02-07
Fix intermittent test failures in workers

NatashaWso2:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13463
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13501
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13490

### DIFF
--- a/examples/async/async.bal
+++ b/examples/async/async.bal
@@ -19,8 +19,6 @@ public function main() {
     // Call the `countInfinity()` function, that runs forever in asynchronous
     // mode.
     future<()> f2 = start countInfinity();
-    // A `runtime:sleep` is added to delay the execution.
-    runtime:sleep(2000);
     // Check if the function call is done.
     io:println(f2.isDone());
     // Check if someone cancelled the asynchronous execution.
--- a/examples/async/async.bal
+++ b/examples/async/async.bal
@@ -12,12 +12,15 @@ public function main() {
     // You can pass the value of the `future` variable around
     // and call its results later.
     int result = squarePlusCube(f1);
+    // Wait for future `f1` to finish.
+    _ = wait f1;
     io:println("SQ + CB = ", result);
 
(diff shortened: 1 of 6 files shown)


examples/async/async.bal
```diff
@@ -122,4 +122,4 @@ function countInfinity() {
     while (true) {
         count += 1;
     }
-}
+}
```
vinok88: Add a new line.

examples/async/async.bal
```diff
@@ -12,12 +12,15 @@ public function main() {
...
+    _ = wait f1;
     io:println("SQ + CB = ", result);
 
-    // Call the `countInfinity()` function, that runs forever in asynchronous 
+    // Call the `countInfinity()` function, that runs forever in asynchronous
     // mode.
     future<()> f2 = start countInfinity();
-    runtime:sleep(1000);
+    // A `runtime:sleep` is added to delay the execution.
+    runtime:sleep(2000);
```
vinok88: Shall we avid using sleeps for passing tests? Using a sleep is not a permanent solution. It should we used only if there is no other solution. Since this is a example test, what you can do is add all the possible results due to the parallel behavior, to the test.

vinok88: Or, you can join all the returned outputs and check each line exists.

examples/async/async.bal
NatashaWso2: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/13513/commits/dd5ce2b5a90f3a614a31d74ffe9360b8a90b7986. Thanks :)

examples/async/async.bal
NatashaWso2: Fixed with [dd5ce2b](https://github.com/ballerina-platform/ballerina-lang/commit/dd5ce2b5a90f3a614a31d74ffe9360b8a90b7986). Thanks :)

PR #13516, opened 2019-02-07
Add decimal support for cursor tables

anupama-pathirage:
## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no


stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDataIterator.java
```diff
@@ -603,6 +612,24 @@ private void handleDoubleValue(double fValue, BMap<String, BValue> bStruct, Stri
...
+    private void handleDecimalValue(BigDecimal fValue, BMap<String, BValue> bStruct, String fieldName, BType fieldType)
+            throws SQLException {
+        boolean isOriginalValueNull = rs.wasNull();
+        int fieldTypeTag = fieldType.getTag();
+        if (fieldTypeTag == TypeTags.UNION_TAG) {
+            BRefType refValue = isOriginalValueNull ? null : new BDecimal(fValue);
+            validateAndSetRefRecordField(bStruct, fieldName, TypeTags.DECIMAL_TAG,
+                    retrieveNonNilTypeTag(fieldType), refValue, UNASSIGNABLE_UNIONTYPE_EXCEPTION);
+        } else {
+            if (isOriginalValueNull) {
```
NatashaWso2: Shall we can join this if condition to the previous if block?

```java
else if (isOriginalValueNull) {
     handleNilToNonNillableFieldAssignment();
} else {
     validateAndSetRefRecordField(bStruct, fieldName, TypeTags.DECIMAL_TAG, fieldTypeTag,
                        new BDecimal(fValue), MISMATCHING_FIELD_ASSIGNMENT);
}
```

anupama-pathirage: IMO it will reduce the clarity of the code as the else block contains the case where the field type is not a union type. If we move to a else if block that idea is not clear as we are checking different unrelated conditions within the if and else if

Manuri: There are several similar methods for other data types and this is how we've done there as well, due to the same reason. So +1 to keep this logic.

PR #13542, opened 2019-02-11
Refactor CONVERT and STAMP instruction codes

warunalakshitha:
## Purpose
This will refactor  CONVERT and STAMP instruction codes to util built in functions.

## Goal
According to current impl we use BVM specific instruction codes to execute both built in methods which is tightly couple to bvm. We need to remove that coupling.

## Approach
Desugar to util built in package functions.

## Automation tests
 - Unit tests 
   > Yes


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -472,7 +472,7 @@ private BSymbol generateStampSymbol(Name name, BType variableSourceType, BType t
         if (types.isAssignable(variableSourceType, targetType)) {
             List<BType> paramTypes = new ArrayList<>();
             paramTypes.add(variableSourceType);
-            return symTable.createOperator(name, paramTypes, targetType, InstructionCodes.STAMP);
+            return symTable.createOperator(name, paramTypes, targetType, -1);
```
vinok88: I think the convention is to use the `InstructionCodes.NOOP` instead of `-1`

warunalakshitha: Fixed

PR #13602, opened 2019-02-13
Support index based access via expressions for tuples

### LINKED ISSUE #13164: [Spec Deviation] Index based access in Tuples should support expressions
**Description:**
```ballerina
function testTupleMemberReferenceByInvalidIntegerIndex() {
    (float, float, float) tuple = (1.1, 0.0, 2.20);
    int index = 1;
    float value = tuple[index]; // This should be supported
}
```
Results in
```
invalid index expression: expected integer literal
```

**Affected Versions:**
0.990.2

Anoukh:
## Purpose
Fixes #13164
Supports the following:
```ballerina
function tupleAccessTest() {
    (int, string, boolean, ()) tuple = (100, "string_value", true, ());
    int index = 2;
    int|string|boolean|() x = tuple[index]; // If value of index is 4, the tuple access will result in a panic with reason IndexOutOfRange
}

```


bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
```diff
@@ -1840,9 +1840,16 @@ private static void execStoreOpcodes(Strand ctx, StackFrame sf, int opcode, int[
...
                 long index = sf.longRegs[j];
                 BRefType refReg = sf.refRegs[k];
-                BType elementType = (list.getType().getTag() == TypeTags.ARRAY_TAG)
-                        ? ((BArrayType) list.getType()).getElementType()
-                        : ((BTupleType) list.getType()).getTupleTypes().get((int) index);
+                BType elementType;
+                if (list.getType().getTag() == TypeTags.ARRAY_TAG) {
+                    elementType = ((BArrayType) list.getType()).getElementType();
+                } else {
+                    BTupleType tuple = getTupleFieldType(ctx, (BTupleType) list.getType(), index);
```
MaryamZi: Isn't the method name `getTupleFieldType` a bit misleading? 

Also isn't the method more like a range check? How about updating it to return a `boolean` instead? 

Should we also consider refactoring `execListAddOperation` for tuples? Or we will be repeating the range check right?

bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
```diff
@@ -1943,6 +1950,25 @@ private static void execStoreOpcodes(Strand ctx, StackFrame sf, int opcode, int[
         }
     }
 
+    /**
+     * Method used to get the type of a specific member in a tuple specified by a given index.
+     *
+     * @param ctx   strand
+     * @param tuple the tuple
+     * @param index the index of the member
+     * @return the type of the member
```
MaryamZi: Currently returns the tuple itself right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -1981,7 +1981,7 @@ public void visit(BLangIndexBasedAccess indexAccessExpr) {
         BType varRefType = indexAccessExpr.expr.type;
         if (varRefType.tag == TypeTags.OBJECT || varRefType.tag == TypeTags.RECORD) {
             targetVarRef = new BLangStructFieldAccessExpr(indexAccessExpr.pos,
```
Shan1024: Looks like that the RHS of this statement can be refactored into two lines.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2789,6 +2777,10 @@ private BType checkIndexAccessExpr(BLangIndexBasedAccess indexBasedAccessExpr, B
                     actualType = ((BArrayType) varRefType).getElementType();
                 }
                 break;
+            case TypeTags.TUPLE:
+                indexExprType = checkExpr(indexExpr, this.env, symTable.noType);
```
Shan1024: This will set the type to the `indexExpr`. Then we don't have to pass the `indexExprType` to the `checkTupleIndexBasedAccess()` below since we can access the type. WDYT?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2818,6 +2803,56 @@ private BType checkIndexAccessExpr(BLangIndexBasedAccess indexBasedAccessExpr, B
         return actualType;
     }
 
+    private BType checkTupleIndexBasedAccess(BLangIndexBasedAccess accessExpr, BType indexExprType, BTupleType tuple) {
+        BType actualType = symTable.semanticError;
+        BLangExpression indexExpr = accessExpr.indexExpr;
+        switch (indexExprType.tag) {
```
Shan1024: If we remove the `indexExprType` parameter(), we can use `indexExpr` to get the type tag here. WDYT?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2818,6 +2803,56 @@ private BType checkIndexAccessExpr(BLangIndexBasedAccess indexBasedAccessExpr, B
...
+                    actualType = tupleTypes.size() == 1 ? tupleTypes.iterator().next() :
+                            new BUnionType(null, tupleTypes, false);
+                }
+                break;
+            case TypeTags.FINITE:
+                BFiniteType finiteIndexExpr = (BFiniteType) indexExprType;
+                LinkedHashSet<BType> possibleTypes = new LinkedHashSet<>();
+                for (BLangExpression finiteMember : finiteIndexExpr.valueSpace) {
+                    if (finiteMember.type.tag != TypeTags.INT) {
+                        dlog.error(indexExpr.pos, DiagnosticCode.INVALID_FINITE_TYPE_INDEX_EXPR, indexExprType);
```
Shan1024: Something like this will fail, right - 

```ballerina
type V 0|"x";
V v = 0;

(int, string) t = (1, "one");
int i = t[v];
```

Is that the correct behavior according to the spec? This looks like a valid scenario.

Anoukh: Index expressions are not completed yet in the Spec. Went with this restrictive approach so as to avoid adding more checks at the runtime. Added a test case for it as well. Yes, it is valid but personally didn't see a critical use case for such a scenario.

Shan1024: Yup, it is not critical. Maybe later we can allow it.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -359,6 +359,9 @@ error.invalid.index.expr.struct.field.access=\
 error.invalid.index.expr.tuple.field.access=\
   invalid index expression: expected integer literal
 
+error.invalid.finite.type.index.expr=\
+  invalid index expression: invalid finite type value space ''{0}''
```
hasithaa: In the spec, we don't use the word "finite type", This should give an error message, which indicates, given value is doesn't belong to tuple size.

Issue #13579, opened 2019-02-12
ballerina-examples/hello-world-parallel:0.0.1:testFunc fails intermittently


MaryamZi:
**Description:**
```Ballerina project initialized
Compiling source
    ballerina-examples/hello-world-service:0.0.1

Running tests
    ballerina-examples/hello-world-service:0.0.1
[ballerina/http] started HTTP/WS endpoint 0.0.0.0:9090
        [pass] testFunc

        1 passing
        0 failing
        0 skipped

Generating executable
    ./target/hello-world-service.balx
Compiling source
    ballerina-examples/hello-world:0.0.1

Running tests
    ballerina-examples/hello-world:0.0.1
        [pass] testFunc

        1 passing
        0 failing
        0 skipped

Generating executable
    ./target/hello-world.balx
Compiling source
    ballerina-examples/hello-world-parallel:0.0.1

Running tests
    ballerina-examples/hello-world-parallel:0.0.1
        [fail] testFunc:
            error: NullReferenceException {}
                    at equalsIgnoreCase(<native>)
                    ballerina-examples/hello-world-parallel:0.0.1:testFunc(tests/hello_world_parallel_test.bal:24)

        0 passing
        1 failing
        0 skipped
```

Failed on Jenkins.

PR #13609, opened 2019-02-14
 Fix intermittent test failure in hello-world-parallel and fork variable access BBE

NatashaWso2:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13579
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13898


examples/hello-world-parallel/hello_world_parallel.bal
```diff
@@ -14,4 +14,7 @@ public function main() {
     worker w3 {
         io:println("Hello, World! #k");
     }
+
+    // Wait for all workers to finish.
+    _ = wait {w1, w2, w3};
```
lankavitharana: why do we need this? AFAIR we anyway wait for all the workers to finish

vinok88: It does not wait, within the test invocations. It does wait if we run it through a ballerina run only.

NatashaWso2: The main function is invoked from tests then we can't guarantee that all workers in the main function have finished. To ensure that they are finished added a wait so that it would wait for all the workers to complete

lankavitharana: So what I'm wondering is whether the fix would be to simply wait for whatever workers started to be finished in test runs? isn't that the case?

hasithaa: Shall we fix this in test case itself, without introducing wait functionality? 

A better solution is, add multiple attempts with a fixed time interval to validate the output.

NatashaWso2: @hasithaa as discussed added the proper fix with 88c1c7c

PR #13618, opened 2019-02-15
Align the implicit initial values of types with the spec

### LINKED ISSUE #13166: [Spec Deviation] Implicit initial value of tuple, map, record, table, typedesc, singleton, error, object and union not working as expected

pubudu91:

### DIFF
--- a/bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BFiniteType.java
+++ b/bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BFiniteType.java
@@ -45,12 +45,12 @@ public class BFiniteType extends BType {
             return false;
         }
         BFiniteType that = (BFiniteType) o;
-        return valueSpace.containsAll(that.valueSpace) && that.valueSpace.containsAll(valueSpace);
+        return this.valueSpace.size() == that.valueSpace.size() && this.valueSpace.containsAll(that.valueSpace);
     }
 
     @Override
     public <V extends BValue> V getZeroValue() {
-        if (valueSpace.stream().anyMatch(val -> val == null || BTypes.isImplicitInitValueNil(val.getType()))) {
+        if (valueSpace.stream().anyMatch(val -> val == null || val.getType().isNilable())) {
             return null;
         }
 
@@ -78,7 +78,7 @@ public class BFiniteType extends BType {
 
     @Override
     public <V extends BValue> V getEmptyValue() {
-        if (valueSpace.stream().anyMatch(val -> val == null || BTypes.isImplicitInitValueNil(val.getType()))) {
+        if (valueSpace.stream().anyMatch(val -> val == null || val.getType().isNilable())) {
             return null;
         }
 
--- a/bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BFiniteType.java
+++ b/bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BFiniteType.java
@@ -56,8 +56,12 @@ public class BFiniteType extends BType {
 
         Iterator<BValue> valueIterator = valueSpace.iterator();
         BValue firstVal = valueIterator.next();
-        BValue implicitInitValOfType = firstVal.getType().getZeroValue();
 
+        if (isSingletonType()) {
+            return (V) firstVal;
+        }
+
+        BValue implicitInitValOfType = firstVal.getType().getZeroValue();
         if (implicitInitValOfType.equals(firstVal)) {
             return (V) implicitInitValOfType;
         }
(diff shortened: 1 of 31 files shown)


bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BFiniteType.java
```diff
@@ -35,18 +36,80 @@ public BFiniteType(String typeName, String pkgPath) {
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof BFiniteType)) {
+            return false;
+        }
+        BFiniteType that = (BFiniteType) o;
+        return valueSpace.containsAll(that.valueSpace) && that.valueSpace.containsAll(valueSpace);
```
SupunS: checking the length, and containsAll() for one way would be faster?

bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BRecordType.java
```diff
@@ -53,12 +54,28 @@ public TypeInfo getTypeInfo() {
...
+                    implicitInitValue.put(entry.getKey(), value);
+                });
+
+        return (V) implicitInitValue;
     }
 
     @Override
     public <V extends BValue> V getEmptyValue() {
-        return (V) new BMap<>(this);
+        BMap<String, BValue> implicitInitValue = new BMap<>(this);
```
SupunS: can replace with `return getZeroValue()`

bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BTypes.java
```diff
@@ -111,4 +111,14 @@ public static BType fromString(String typeName) {
         }
         return getTypeFromName(typeName);
     }
+
+    public static boolean isImplicitInitValueNil(BType type) {
```
SupunS: this is equivalent to checking nillable right? so can we call it `isNillable()` and make it generic?

bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BUnionType.java
```diff
@@ -62,12 +62,20 @@ public boolean isNullable() {
 
     @Override
     public <V extends BValue> V getZeroValue() {
-        return null;
+        if (nullable || memberTypes.stream().anyMatch(BTypes::isImplicitInitValueNil)) {
```
SupunS: if we do as my above comment, this will be simply `if(BTypes.isNillable(this))`

bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BRecordType.java
pubudu91: Can't do it for record type since we are calling `getEmptyValue()` recursively on the fields. But can do this for some of the others like `BTableType`

bvm/ballerina-core/src/main/java/org/ballerinalang/model/types/BTypes.java
pubudu91: This is in the runtime. The `isNullable()` method is there only in the compile time `BType`. Will have to add it to the runtime `BType` as well

PR #13661, opened 2019-02-16
Bump balo version in refactor builtin functions branch

warunalakshitha:
## Purpose
Bump class version and balo version.


warunalakshitha:
There are some spec testes added by new master changes and after merging master those got failed. We need to fix them first. Will close this and reopen once fixed those.

PR #13662, opened 2019-02-16
Refactor Builtin Functions

warunalakshitha:
## Purpose
Some of ballerina operation use bvm specific instruction codes, because of that run time implementation was tightly coupled with BVM logic. This PR will refactor builtin functions by rewriting in ballerina it self or desugar to util methods .

##Goals
Covers following instructions.

REASON
DETAIL         
LENGTH           
FREEZE           
IS_FROZEN           
CLONE           
STAMP      
CONVERT    
ITERATE            

## Approch
Desugar to util package built in function.

## Automation tests
 - Unit tests 
   > Yes


warunalakshitha:
Need to address some review suggestions hence closed this. will add pr directly with my branch.

PR #13679, opened 2019-02-18
Improve object type infrence

rdhananjaya:
## Purpose
This PR improves type inference of `new` when LHS is an union type.

```ballerina
type Obj0 object {
    int val;
    function __init(int i, int j = 0) {
        self.val = 0;
    }
};

type Obj2 object {
    int val;
    function __init() {
        self.val = 2;
    }
};

// Will infer 'obj0Instance' to be of type 'Obj0'
Obj0|Obj2 obj0Instance = new(5, j=0);

// Will infer 'obj2Instance' to be of type 'Obj2'
Obj0|Obj2 obj2Instance = new();

// Will emit error: 'cannot infer type of the object from 'Obj|Obj2'
Obj0|Obj2 objxInstance = new(100, j="zero");
```


tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/object/ObjectTest.java
```diff
@@ -539,17 +537,7 @@ public void testObjectInitFunctionNegative() {
...
-        BAssertUtil.validateError(result, 1, "cannot infer type of the object from 'Person?'", 2, 14);
-        BAssertUtil.validateError(result, 2, "cannot infer type of the object from 'Person?'", 5, 18);
-        BAssertUtil.validateError(result, 3, "cannot infer type of the object from 'Person?'", 6, 18);
-        BAssertUtil.validateError(result, 4, "cannot infer type of the object from 'Person?'", 8, 10);
-        BAssertUtil.validateError(result, 5, "cannot infer type of the object from 'Person?'", 10, 10);
-        BAssertUtil.validateError(result, 6, "cannot infer type of the object from 'Person?'", 19, 25);
-        BAssertUtil.validateError(result, 7, "cannot infer type of the object from 'Person?'", 20, 25);
-        BAssertUtil.validateError(result, 8, "cannot infer type of the object from 'Person?'", 25, 19);
-        BAssertUtil.validateError(result, 9, "cannot infer type of the object from 'Person?'", 26, 19);
+        Assert.assertEquals(result.getErrorCount(), 0);
```
pubudu91: Instead of simply checking the compile error count, shall we move this to the positive test cases and get rid of this now-invalid negative test?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1112,12 +1113,27 @@ public void visit(BLangTypeInit cIExpr) {
...
             case TypeTags.CHANNEL:
-                if (cIExpr.initInvocation.argExprs.size() > 0) {
+                if (!cIExpr.initInvocation.argExprs.isEmpty()) {
                     dlog.error(cIExpr.pos, DiagnosticCode.TOO_MANY_ARGS_FUNC_CALL, cIExpr.initInvocation.name);
                     resultType = symTable.semanticError;
                     return;
                 }
                 break;
+            case TypeTags.UNION:
+                List<BType> matchingMembers = findMembersWithMatchingInitFunc(cIExpr, (BUnionType) actualType);
```
warunalakshitha: What if I have union type with objecttype and non object type like map ?

rdhananjaya: If we have `Object1|Object2|map<string> variable = new(args..);`
And if arguments can match one of the `__init` functions of either Object1 or Object2 relevant __init will be called. Otherwise error similar to `cannot infer type of the object from 'Obj|Obj2|Obj3|Obj4'` will be emitted.

We already had type inference support for record/map literals. `Obj|map<string> objAr = { item: "item line" };` this code works without this PR.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1128,6 +1144,115 @@ public void visit(BLangTypeInit cIExpr) {
         resultType = types.checkType(cIExpr, actualType, expType);
     }
 
+    private List<BType> findMembersWithMatchingInitFunc(BLangTypeInit cIExpr, BUnionType lhsUnionType) {
+        List<BType> matchingLhsMemberTypes = new ArrayList<>();
+        for (BType memberType : lhsUnionType.memberTypes) {
+            if (memberType.tsymbol == null || memberType.tsymbol.kind != SymbolKind.OBJECT) {
```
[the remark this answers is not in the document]
rdhananjaya: I guess we can, will do the change.
BTW what's the difference between the `tag` and the `kind`, I mean in when to use which kind of a way?

pubudu91:
The code coverage report says there's a 3% decrease in coverage. Shall we look into that as well?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1113,12 +1114,27 @@ public void visit(BLangTypeInit cIExpr) {
...
                     resultType = symTable.semanticError;
                     return;
                 }
                 break;
+            case TypeTags.UNION:
+                List<BType> matchingMembers = findMembersWithMatchingInitFunc(cIExpr, (BUnionType) actualType);
+                BType matchedType = getMatchedTypeOrEmitError(matchingMembers, cIExpr, actualType);
+
+                expType = matchedType;
+                if (matchedType.tsymbol.type.tag == TypeTags.OBJECT
```
pubudu91: This can simply be `matchedType.tag`. Getting the type from the type symbol will again return the `matchedType`.

rdhananjaya:
> The code coverage report says there's a 3% decrease in coverage. Shall we look into that as well?

Maybe I'm mis-reading test coverage report, I cannot figure out how LoggingExceptionListener's coverage drop due to this PR.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1129,6 +1145,115 @@ public void visit(BLangTypeInit cIExpr) {
...
+        // More named args given than function can accept.
+        if (function.symbol.defaultableParams.size() < namedArgs.size()) {
+            return false;
+        }
+
+        int matchedParamterCount = 0;
+        for (BVarSymbol defaultableParam : function.symbol.defaultableParams) {
+            for (BLangExpression namedArg : namedArgs) {
+                BLangNamedArgsExpression namedArgExpr = (BLangNamedArgsExpression) namedArg;
+                if (namedArgExpr.name.value.equals(defaultableParam.name.value)) {
```
SupunS: fail fast

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1129,6 +1145,115 @@ public void visit(BLangTypeInit cIExpr) {
...
+
+            BAttachedFunction initializerFunc = ((BObjectTypeSymbol) memberType.tsymbol).initializerFunc;
+            if (isArgsMatchesFunction(cIExpr.argsExpr, initializerFunc)) {
+                matchingLhsMemberTypes.add(memberType);
+            }
+        }
+        return matchingLhsMemberTypes;
+    }
+
+    private BType getMatchedTypeOrEmitError(List<BType> matchingLhsMembers, BLangTypeInit cIExpr, BType lhsUnion) {
```
SupunS: `getMatchingType()` would be better?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1129,6 +1145,115 @@ public void visit(BLangTypeInit cIExpr) {
...
+        if (requiredParamCount > unnamedArgs.size()) {
+            return false;
+        }
+
+        // No rest params, all (unnamed) args must match params.
+        if (function.symbol.restParam == null && requiredParamCount != unnamedArgs.size()) {
+            return false;
+        }
+
+        // Each arguments must be assignable to required parameter at respective position.
```
SupunS: order doesn't matter for named-args. if the user pass the named-args in a different order than the function definition, will it get handled here?

rdhananjaya: It's handled in `matchDefaultableParameters`.
We can't pass arguments to ordinary parameters by name right?

SupunS: Ack. 
btw, where do we check the restParam Type?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
rdhananjaya: Failing fast (`continue`) at outer nested `if` didn't make the code any better IMO.
I also find this too nested, I tried a hash-join kind of thing and found it's way more complicated for this.

Early exit:
```java
        int matchedParamterCount = 0;
        for (BVarSymbol defaultableParam : function.symbol.defaultableParams) {
            for (BLangNamedArgsExpression namedArg : namedArgs) {
                if (!namedArg.name.value.equals(defaultableParam.name.value)) {
                    continue;
                }
                BType namedArgExprType = checkExpr(namedArg.expr, env);
                if (types.isAssignable(defaultableParam.type, namedArgExprType)) {
                    matchedParamterCount++;
                } else {
                    // Name matched, type mismatched.
                    return false;
                }
            }
        }
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
rdhananjaya: Yes after choosing a member type from union we run `types.checkType(cIExpr, matchedType, actualType);`

SupunS: shouldn't we check the rest param when choosing a member? 
Lets say `Foo` has a constructor `__init(string... x)` and `Bar` has a constructor `__init(int... x)`
then `Foo|Bar y = new(5)`. What would be the output?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
SupunS: This is better right? one less level.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
rdhananjaya: Yes we need to consider that, I completely missed that scenario.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1129,6 +1145,136 @@ public void visit(BLangTypeInit cIExpr) {
         resultType = types.checkType(cIExpr, actualType, expType);
     }
 
+    private List<BType> findMembersWithMatchingInitFunc(BLangTypeInit cIExpr, BUnionType lhsUnionType) {
+        List<BType> matchingLhsMemberTypes = new ArrayList<>();
+        for (BType memberType : lhsUnionType.memberTypes) {
+            if (memberType.tsymbol == null || memberType.tag != TypeTags.OBJECT) {
```
hasithaa: why can't simply check memberType.tag != TypeTags.OBJECT only.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1113,12 +1114,27 @@ public void visit(BLangTypeInit cIExpr) {
...
                     resultType = symTable.semanticError;
                     return;
                 }
                 break;
+            case TypeTags.UNION:
+                List<BType> matchingMembers = findMembersWithMatchingInitFunc(cIExpr, (BUnionType) actualType);
+                BType matchedType = getMatchingType(matchingMembers, cIExpr, actualType);
+
+                expType = matchedType;
+                if (matchedType.tag == TypeTags.OBJECT
```
hasithaa: Checking TypeTag.Object is enough, because, we know that, the matched type should have an initializer function from previous validations. isn't it ?

hasithaa: Ahh, there can be a scenario, the user hasn't mentioned the init function. In that case we need to check.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1129,6 +1145,136 @@ public void visit(BLangTypeInit cIExpr) {
...
+            BLangExpression argument = unnamedArgs.get(i);
+            if (!types.isAssignable(argument.type, param.type)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean restArgTypesMatch(List<BLangExpression> unnamedArgs, int requiredParamCount, BType restParamType) {
+        if ((unnamedArgs.size() - requiredParamCount) == 0) {
```
hasithaa: why can't we just check unnamedArgs.size() == requiredParamCount :) Above looks weired.

rdhananjaya: It does look weird, will update.

PR #13693, opened 2019-02-18
Fix vscode dependency issues

pahans:
- Moved vscode dependency to devDependencies
- cross-env and glob are both dev and production dependencies so moved them to dependencies


pahans:
closing the PR because build fails locally. will reopen after fixing

PR #13703, opened 2019-02-18
Refactor to use same symbol for type casting and conversion for unsafe numeric types

warunalakshitha:
## Purpose

According to spec, Convert method will return an error for failed conversion and type casting will be panic. When we use same symbol for both cases, type casting compilation will fail since we check for error as well.

This fix will provide to use same conversion symbol for type casting and conversion.
We need to remove error type from conversion symbol for type casting since type casting will be panic in unsafe conversion. 

This pr will also make float to int and decimal to int conversions unsafe as per spec.

This will fix the tests accordingly.

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/StreamingCodeDesugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/StreamingCodeDesugar.java
@@ -93,6 +93,7 @@ import org.wso2.ballerinalang.compiler.util.CompilerContext;
 import org.wso2.ballerinalang.compiler.util.FieldKind;
 import org.wso2.ballerinalang.compiler.util.Name;
 import org.wso2.ballerinalang.compiler.util.Names;
+import org.wso2.ballerinalang.compiler.util.TypeTags;
 import org.wso2.ballerinalang.compiler.util.diagnotic.BLangDiagnosticLog;
 import org.wso2.ballerinalang.compiler.util.diagnotic.DiagnosticPos;
 import org.wso2.ballerinalang.util.Lists;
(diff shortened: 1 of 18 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/StreamingCodeDesugar.java
```diff
@@ -93,6 +93,7 @@
 import org.wso2.ballerinalang.compiler.util.FieldKind;
 import org.wso2.ballerinalang.compiler.util.Name;
 import org.wso2.ballerinalang.compiler.util.Names;
+import org.wso2.ballerinalang.compiler.util.TypeTags;
```
vinok88: Do we need this import? there no other changes in this class.

stdlib/utils/pom.xml
```diff
@@ -22,7 +22,7 @@
     <parent>
         <artifactId>ballerina-parent</artifactId>
         <groupId>org.ballerinalang</groupId>
-        <version>0.990.3-SNAPSHOT</version>
+        <version>0.990.4-SNAPSHOT</version>
```
vinok88: shouldn't this be `paren.version`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/StreamingCodeDesugar.java
warunalakshitha: I merged the latest master changes to branch and some how this import is missing. Need to add this to since some earlier changes in refactor branch need this, else build will be failed.

stdlib/utils/pom.xml
warunalakshitha: We use same way for other modules as well. I think this version automatically change with builder plugin hence should be ok.

PR #13711, opened 2019-02-19
Allow mapping string to json in table constraint

### LINKED ISSUE #13710: Allow having JSON field in the table constraint to map string values from db
**Description:**
$Subject since json is a union containing string.
i.e. Following shouldn't throw mismatching field assignment runtime error

```ballerina
type ResultJson record {
   json STRING_TYPE;
};

var result = testDB->select("SELECT string_type from DataTable WHERE row_id = 1", ResultJson);
    json val = "";
    if (result is table<ResultJson>) {
        foreach var rs in result {
            val = rs.STRING_TYPE;
        }
    }
```

Manuri:
## Purpose
Fix #13710

### DIFF
--- a/stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDataIterator.java
+++ b/stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDataIterator.java
@@ -53,6 +53,7 @@ import java.sql.Struct;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.sql.Types;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.List;
 import javax.sql.rowset.CachedRowSet;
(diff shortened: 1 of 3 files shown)


anupama-pathirage:
@Manuri IMO the proper fix should contain all the inclusive types of json such as number, boolean etc

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDataIterator.java
```diff
@@ -526,11 +538,13 @@ private void handleStringValue(String stringValue, String fieldName,
         int fieldTypeTag = fieldType.getTag();
         if (fieldTypeTag == TypeTags.UNION_TAG) {
             BRefType refValue = stringValue == null ? null : new BString(stringValue);
-            validateAndSetRefRecordField(bStruct, fieldName, TypeTags.STRING_TAG,
+            int[] expectedTypeTags = { TypeTags.STRING_TAG, TypeTags.JSON_TAG };
```
anupama-pathirage: Any and Anydata should be also part of this?

Manuri: As discussed offline we'll do the complete fix (which requires some refactoring) in a separate PR and merge this considering the urgency of the fix.

PR #13712, opened 2019-02-20
Introduce global, shared db connection pool

Manuri:
## Purpose
**Previous Design**
Each database endpoint will have a separate database connection pool with properties which can be defined by the user. 
Eg: 
```ballerina
mysql:Client testDB = new({
       host: "localhost",
       port: 3306,
       name: "testdb",
       username: "test",
       password: "test",
       dbOptions: { useSSL: false }
   });
```

For this endpoint, a new connection pool is created with the properties provided in the poolOptions record. The pool creation happens when a client is created.

**Proposed Design**
For an application with several database endpoints with the same database backend, creating separate database connection pools may not be required. It might be a waste of resources. Also, some users may not want to mess with the pool options at all.

We can instead have a default global connection pool per database with a set of default configuration values. A given connection pool maps with a database endpoint identified by host, port, database name and database/driver properties(eg: IFEXISTS property in H2, maxReconnects property in MySQL). These additional properties have to be taken into account since setting these properties is coupled with internal pool creation. 

A developer is free to change the configuration values of the default global pool. However, this is possible only at the program startup through config API.

There might be a real requirement to increase/decrease the pool size for some db endpoint without changing the global setting. User can simply pass his custom pool config to the client constructor, which would create a separate pool for him. 

```ballerina
// Since pool options are not provided, this will use the default global pool if exists, or will result in the 
// creation of the default global pool. This pool will be mapped against a JDBC URL.
mysql:Client testDB = new({
       host: "localhost",
       port: 3306,
       name: "testdb",
       username: "test",
       password: "test",
       dbOptions: { useSSL: false }
   });

// Following will result in the creation of a pool that can be used only by this client
mysql:Client testDB = new({
       host: "localhost",
       port: 3306,
       name: "testdb",
       username: "test",
       password: "test",
       dbOptions: { useSSL: false }
   });

// Following pool configuration can be shared among multiple clients.
PoolOptions p = { maximumPoolSize: 5 };

// During the initialization of this client, a pool will be created 
mysql:Client testDB1 = new({
       host: "localhost",
       port: 3306,
       name: "testdb",
       username: "test",
       password: "test",
       poolOptions: p,
       dbOptions: { useSSL: false }
   });

// Since this client also connects to the same destination db and has same database options,
// this will reuse the pool created during the initialization of previous client.
mysql:Client testDB2 = new({
       host: "localhost",
       port: 3306,
       name: "testdb",
       username: "test",
       password: "test",
       poolOptions: p,
       dbOptions: { useSSL: false }
   });

// Since this client is pointing to a different destination db, even though the pool config is the same, a 
// separate pool will be created for this client.
mysql:Client testDB3 = new({
       host: "someotherhost",
       port: 3306,
       name: "testdb",
       username: "test",
       password: "test",
       poolOptions: p,
       dbOptions: { useSSL: false }
   });
```
Resolves https://github.com/ballerina-platform/ballerina-lang/issues/13817


examples/h2-client/h2_client.bal
```diff
@@ -9,8 +9,7 @@ h2:Client testDB = new({
         path: "./h2-client",
         name: "testdb",
         username: "SA",
-        password: "",
-        poolOptions: { maximumPoolSize: 5 }
+        password: ""
```
anupama-pathirage: Shall we add testDB.stop() for all the samples?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/actions/Close.java
```diff
@@ -42,6 +44,17 @@
...
     @Override
     public void execute(Context context) {
-        closeConnections(retrieveDatasource(context));
+        SQLDatasource datasource = retrieveDatasource(context);
+        // When an exception is thrown during database endpoint init (eg: driver not present) stop operation
+        // of the endpoint is automatically called. But at this point, datasource is null therefore to handle that
+        // situation following null check is needed.
+        if (datasource != null && !datasource.isGlobalDatasource()) {
+            try {
+                datasource.decrementClientCounter();
```
anupama-pathirage: Shall we rename the method `decrementClientCounter()` to something like `releaseConnectionPool()`. Decrementing is only a part of that operation.

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/actions/Close.java
```diff
@@ -42,6 +44,17 @@
...
+            } catch (InterruptedException e) {
+                context.setReturnValues(
+                        SQLDatasourceUtils.getSQLConnectorError(context, "Error while stopping the database client"));
```
anupama-pathirage: Shall we add more information to the error message based on the message in InterruptedException?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/actions/Close.java
Manuri: When a client decides to stop, datasource decrement the client counter and eventually when it reaches zero pool is shut down.
So isn't it more like, pool shutdown happens as a result of client counter reaching zero?

stdlib/database/h2/src/test/java/org/ballerinalang/stdlib/database/sql/SQLConnectionPoolTest.java
```diff
@@ -0,0 +1,212 @@
...
+/**
+ * Tests scenarios of global pool, local shared pool, local unshared pool usage.
+ */
+public class SQLConnectionPoolTest {
+    private CompileResult result;
+    private static final String DB1_NAME = "TEST_SQL_CONNECTION_POOL_GLOBAL_1";
+    private static final String DB2_NAME = "TEST_SQL_CONNECTION_POOL_GLOBAL_2";
+    private static final String DB3_NAME = "TEST_SQL_CONNECTION_POOL_LOCAL_SHARED_1";
+    private static final String DB4_NAME = "TEST_SQL_CONNECTION_POOL_LOCAL_SHARED_2";
+    private Path ballerinaConfPath;
```
anupama-pathirage: Can this be a local variable?

stdlib/database/sql/src/main/ballerina/sql/types.bal
```diff
@@ -14,35 +14,71 @@
...
+    string connectionInitSql = config:getAsString("connection.init.sql", default = "");
+    string dataSourceClassName = config:getAsString("datasource.class.name", default = "");
+    boolean autoCommit = config:getAsBoolean("connection.auto.commit", default = true);
+    boolean isXA = config:getAsBoolean("xa.enabled", default = false);
+    int maximumPoolSize = config:getAsInt("max.pool.size", default = 10);
+    int connectionTimeout = config:getAsInt("connection.time.out", default = 30000);
+    int idleTimeout =  config:getAsInt("connection.idle.time.out", default = 600000);
+    int minimumIdle = config:getAsInt("connection.min.idle.count", default = 10);
+    int maxLifetime = config:getAsInt("connection.max.life.time", default = 1800000);
+    int validationTimeout = config:getAsInt("validation.time.out", default = 5000);
```
anupama-pathirage: IIRC we had a convention to start all conf with "b7a."

Manuri: +1. May be we should add "b7a.database" as the prefix to make it easy to understand where ther property belongs.

stdlib/database/h2/src/test/resources/ballerina.conf
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+connection.time.out=1000
```
anupama-pathirage: What is the purpose of this file? Are we using them for a test?

Manuri: Yes, this is used in SQLConnectionPoolTest class.

stdlib/database/sql/src/main/ballerina/sql/types.bal
anupama-pathirage: How about "b7a.sql" prefix as it is short and those are actually belonging to sql package?

stdlib/database/h2/src/test/resources/ballerina.conf
anupama-pathirage: IMO we need to think about a way to redesign the test to validate whether the parameters have taken from the file or not

Manuri: Actually, this is used not just to test whether the value is taken from the conf file but also to set the connection timeout to a low value to avoid the tests taking a lot of time. 

This value gets automatically tested in many tests when asserting the timeout log.

stdlib/database/sql/src/main/ballerina/sql/types.bal
```diff
@@ -14,35 +14,71 @@
...
+#                       Default value of this field can be set through the configuration API with the key "connection.time.out"
+# + idleTimeout - Maximum amount of time that a connection is allowed to sit idle in the pool. Default is 10 minutes.
+#                 Default value of this field can be set through the configuration API with the key "connection.idle.time.out"
+# + minimumIdle - Minimum number of idle connections that pool tries to maintain in the pool. Default is same as maximumPoolSize.
+#                 Default value of this field can be set through the configuration API with the key "connection.min.idle.count"
 # + maxLifetime - Maximum lifetime of a connection in the pool. Default is 30 minutes
+#                 Default value of this field can be set through the configuration API with the key "connection.max.life.time"
 # + validationTimeout - Maximum amount of time that a connection will be tested for aliveness. Default 5 seconds
+#                       Default value of this field can be set through the configuration API with the key "validation.time.out"
```
anupama-pathirage: Check the formatting for above doc comments (max 120 char)

stdlib/database/sql/src/main/ballerina/sql/types.bal
```diff
@@ -14,35 +14,71 @@
...
+# + maximumPoolSize - Maximum size that the pool is allowed to reach, including both idle and in-use connections.
+#                     Default value of this field can be set through the configuration API with the key "max.pool.size"
+# + connectionTimeout - Maximum number of milliseconds that a client will wait for a connection from the pool. Default is 30 seconds.
+#                       Default value of this field can be set through the configuration API with the key "connection.time.out"
+# + idleTimeout - Maximum amount of time that a connection is allowed to sit idle in the pool. Default is 10 minutes.
+#                 Default value of this field can be set through the configuration API with the key "connection.idle.time.out"
+# + minimumIdle - Minimum number of idle connections that pool tries to maintain in the pool. Default is same as maximumPoolSize.
+#                 Default value of this field can be set through the configuration API with the key "connection.min.idle.count"
 # + maxLifetime - Maximum lifetime of a connection in the pool. Default is 30 minutes
```
anupama-pathirage:
```suggestion
# + maxLifetime - Maximum lifetime of a connection in the pool. Default is 30 minutes.
```

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/actions/Close.java
Manuri: As discussed offline, will rename the method to imply we are decrementing the client count and attempting to shut the pool down.

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/PoolOptionsWrapper.java
```diff
@@ -0,0 +1,112 @@
...
+    private SQLDatasource createAndInitDatasource(SQLDatasource.SQLDatasourceParams sqlDatasourceParams) {
+        SQLDatasource newSqlDatasource = new SQLDatasource().init(sqlDatasourceParams);
+        newSqlDatasource.incrementClientCounter();
+        return newSqlDatasource;
+    }
+
+    /**
+     * Retrieve the value of the field corresponding to the provided field.
+     *
+     * @param fieldName The field name
```
anupama-pathirage: Can we have a better description here?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasource.java
```diff
@@ -136,71 +162,83 @@ private void buildDataSource(SQLDatasourceParams sqlDatasourceParams) {
...
-                long connectionTimeout = sqlDatasourceParams.options.getIntField(Constants.Options.CONNECTION_TIMEOUT);
+                long connectionTimeout = ((BInteger) sqlDatasourceParams.poolOptionsWrapper
+                        .get(Constants.Options.CONNECTION_TIMEOUT)).intValue();
                 if (connectionTimeout != -1) {
                     config.setConnectionTimeout(connectionTimeout);
                 }
-                long idleTimeout = sqlDatasourceParams.options.getIntField(Constants.Options.IDLE_TIMEOUT);
+                long idleTimeout = ((BInteger) sqlDatasourceParams.poolOptionsWrapper
+                        .get(Constants.Options.IDLE_TIMEOUT))
+                        .intValue();
```
anupama-pathirage: Reformat to two lines?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasource.java
```diff
@@ -136,71 +162,83 @@ private void buildDataSource(SQLDatasourceParams sqlDatasourceParams) {
...
+                    config.addDataSourceProperty(Constants.USER, sqlDatasourceParams.username);
+                    config.addDataSourceProperty(Constants.PASSWORD, sqlDatasourceParams.password);
                 } else {
                     config.setJdbcUrl(sqlDatasourceParams.jdbcUrl);
                 }
-                String connectionInitSQL = sqlDatasourceParams.options
-                        .getStringField(Constants.Options.CONNECTION_INIT_SQL);
+                String connectionInitSQL = sqlDatasourceParams.poolOptionsWrapper
+                        .get(Constants.Options.CONNECTION_INIT_SQL)
+                        .stringValue();
```
anupama-pathirage: Reformat to two lines?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -94,6 +97,7 @@
     public static final String POSTGRES_OID_COLUMN_TYPE_NAME = "oid";
     private static final int ORACLE_CURSOR_TYPE = -10;
     private static final String TIME_FIELD = "time";
+    private static final String POOL_MAP_KEY = UUID.randomUUID().toString();
```
anupama-pathirage: Shall we move this to Constants.java class. (Similar to https://github.com/Manuri/ballerina/blob/connection-pool-change/stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/Constants.java#L199)

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -1153,67 +1161,176 @@ private static void notifyTxMarkForAbort(Context context, TransactionLocalContex
...
+            BMap<String, BValue> clientEndpointConfig, String urlOptions,
+            BMap<String, BRefType> globalPoolOptions) {
+        String host = clientEndpointConfig.get(Constants.EndpointConfig.HOST).stringValue();
+        int port = (int) ((BInteger) clientEndpointConfig.get(Constants.EndpointConfig.PORT)).intValue();
+        String name = clientEndpointConfig.get(Constants.EndpointConfig.NAME).stringValue();
+        String username = clientEndpointConfig.get(Constants.EndpointConfig.USERNAME).stringValue();
+        String password = clientEndpointConfig.get(Constants.EndpointConfig.PASSWORD).stringValue();
+        BMap<String, BRefType> poolOptions = (BMap<String, BRefType>) clientEndpointConfig
+                .get(Constants.EndpointConfig.POOL_OPTIONS);
+        boolean userProvidedPoolOptionsNotPresent;
```
anupama-pathirage:
```suggestion
        boolean userProvidedPoolOptionsNotPresent = poolOptions == null;
```

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -1153,67 +1161,176 @@ private static void notifyTxMarkForAbort(Context context, TransactionLocalContex
...
+        if (userProvidedPoolOptionsNotPresent = (poolOptions == null)) {
```
anupama-pathirage:
```suggestion
        if (userProvidedPoolOptionsNotPresent) {
```

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
Manuri: This is not used anywhere else and we don't need to make this key public so that anyone who is doing a pool related code change will always have to go through the two methods provided in this class to add/retrieve pool.

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -1153,67 +1161,176 @@ private static void notifyTxMarkForAbort(Context context, TransactionLocalContex
...
-        Map<String, Value> dbOptions = clientEndpointConfig.getMapField(Constants.EndpointConfig.DB_OPTIONS);
-        org.ballerinalang.connector.api.Struct options = clientEndpointConfig
-                .getStructField(Constants.EndpointConfig.POOL_OPTIONS);
+        SQLDatasource.SQLDatasourceParams sqlDatasourceParams = builder.withJdbcUrl(jdbcUrl)
+                .withPoolOptions(poolOptionsWrapper).withUsername(username).withPassword(password).withDbName(name)
+                .withIsGlobalDatasource(userProvidedPoolOptionsNotPresent).build();
+        return createSQLClient(context, sqlDatasourceParams);
+    }
+
+    public static BMap<String, BValue> createSQLDBClient(Context context, BMap<String, BValue> clientEndpointConfig,
```
anupama-pathirage: Unused method?

Manuri: Used in JDBC module

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -1153,67 +1161,176 @@ private static void notifyTxMarkForAbort(Context context, TransactionLocalContex
...
-        String password = clientEndpointConfig.getStringField(Constants.EndpointConfig.PASSWORD);
-        org.ballerinalang.connector.api.Struct options = clientEndpointConfig
-                .getStructField(Constants.EndpointConfig.POOL_OPTIONS);
-
+        String name = clientEndpointConfig.get(Constants.EndpointConfig.NAME).stringValue();
+        String username = clientEndpointConfig.get(Constants.EndpointConfig.USERNAME).stringValue();
+        String password = clientEndpointConfig.get(Constants.EndpointConfig.PASSWORD).stringValue();
+        BMap<String, BRefType> poolOptions = (BMap<String, BRefType>) clientEndpointConfig
+                .get(Constants.EndpointConfig.POOL_OPTIONS);
+        boolean userProvidedPoolOptionsNotPresent;
```
anupama-pathirage: Shall we simplified the if condition block as suggested earlier?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -1627,28 +1744,32 @@ public static Connection getDatabaseConnection(Context context, SQLDatasource da
...
-                case BOOLEAN:
-                    dataValue = Boolean.toString(value.getBooleanValue());
-                    break;
-                default:
-                    dataValue = value.getStringValue();
-                }
+        dbOptions.getMap().forEach((key, value) -> {
+            if (isSupportedDbOptionType(value)) {
+                dbOptionsStringJoiner
+                        .add(key + Constants.JDBCUrlSeparators.EQUAL_SYMBOL + String.valueOf(value.value()));
```
anupama-pathirage:
```suggestion
                        .add(key + Constants.JDBCUrlSeparators.EQUAL_SYMBOL + value.value());
```

tests/ballerina-unit-test/src/test/resources/test-src/types/table/table_type_negative.bal
```diff
@@ -193,6 +193,6 @@ function testWrongOrder(string queryStr) returns error? {
         }
         selectRet.close();
     }
-    testDB.stop();
+    _ = testDB.stop();
```
anupama-pathirage: Shall we add one test case to check type return value of stop method as well?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/actions/Close.java
Manuri: If that happens, I doubt how much a ballerina developer can do about a java thread interruption.
Also, in this case, the InterruptedException object doesn't have a message so all we know is thread was interrupted while waiting for the mutex, hence the exception.

PR #13839, opened 2019-02-24
Refactor Builtin Functions

warunalakshitha:
## Purpose
Some of ballerina operation use bvm specific instruction codes, because of that run time implementation was tightly coupled with BVM logic. This PR will refactor builtin functions by rewriting in ballerina it self or desugar to util methods .

##Goals
Covers following instructions.

REASON
DETAIL         
LENGTH           
FREEZE           
IS_FROZEN           
CLONE           
STAMP      
CONVERT    
ITERATE            

## Approch
Desugar to util package built in function.

## Automation tests
 - Unit tests 
   > Yes


bvm/ballerina-core/src/main/resources/MessagesBundle.properties
```diff
@@ -151,6 +151,9 @@ task.not.running = The task in not running
 illegal.format.conversion = illegal format conversion ''{0}''
 incompatible.stamp.operation = incompatible stamp operation: ''{0}'' value cannot be stamped as ''{1}''
 cannot.stamp.null = cannot stamp ''null'' value to type ''{0}''
+cannot.convert.null = cannot convert ''null'' value to type ''{0}''
```
hasithaa: nil instead of null ?

warunalakshitha: Added same error message similar to existing error message in stamp. I think we need to change both. So I will create issue and will send a pr to fix that separately.

warunalakshitha: https://github.com/ballerina-platform/ballerina-lang/issues/14076

PR #13907, opened 2019-02-26
Flatten the union type

### LINKED ISSUE #13902: Flatten the union type
**Description:**
Currently, the union types which are created internally aren't flattened upon its creation. This leaves that task to the users of union types. There are many places in the compiler where the union type is special cased because of that. Flattening it will solve this issue.

pubudu91:

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
@@ -3435,7 +3435,7 @@ public class Desugar extends BLangNodeVisitor {
         // Owner of the variable symbol must be an invokable symbol
         BType enclosingFuncReturnType = ((BInvokableType) invokableSymbol.type).retType;
         Set<BType> returnTypeSet = enclosingFuncReturnType.tag == TypeTags.UNION ?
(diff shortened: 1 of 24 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -3716,7 +3716,7 @@ private BLangExpression createPatternIfCondition(BLangMatchBindingPatternClause
         BType[] memberTypes;
         if (patternType.tag == TypeTags.UNION) {
             BUnionType unionType = (BUnionType) patternType;
-            memberTypes = unionType.memberTypes.toArray(new BType[0]);
+            memberTypes = unionType.getMemberTypes().toArray(new BType[0]);
```
hasithaa: @pubudu91 Shall we return a Collections.unmodifiableList(list), to prevent adding into list my mistake?

pubudu91: +1

pubudu91: Addressed with 46f4db7

Issue #13607, opened 2019-02-14
ballerina-examples/transactions-distributed - transactions_distributed_test is failing


SupunS:
```
Running tests
    ballerina-examples/xa-transactions:0.0.1
[ballerina/http] started HTTP/WS endpoint 172.17.0.1:35593
2019-02-13 13:22:42,869 INFO  [ballerina/transactions] - Created transaction: 05c97bf4-9aa8-4e42-a424-6aefe8f917cb 
2019-02-13 13:22:42,903 INFO  [ballerina/transactions] - Running 2-phase commit for transaction: 05c97bf4-9aa8-4e42-a424-6aefe8f917cb:ballerina-examples$xa-transactions$1 
	[pass] testFunc

	1 passing
	0 failing
	0 skipped

Generating executable
    ./target/xa-transactions.balx
Compiling source
    ballerina-examples/transactions-distributed:0.0.1

Running tests
    ballerina-examples/transactions-distributed:0.0.1
[ballerina/http] started HTTP/WS endpoint 172.17.0.1:38842
[ballerina/http] started HTTP/WS endpoint 0.0.0.0:8889
[ballerina/http] started HTTP/WS endpoint 0.0.0.0:8080
	[fail] testFunc:
	    error: Assertion Failed! {}
	    	at ballerina/test:createBallerinaError(assert.bal:42)
	    	ballerina/test:assertTrue(assert.bal:52)
	    	ballerina-examples/transactions-distributed:0.0.1:testFunc(tests/transactions_distributed_test.bal:44)

	0 passing
	1 failing
	0 skipped

[ERROR] Command execution failed.
```

rdhananjaya:
This is fixed in https://github.com/ballerina-platform/ballerina-lang/pull/13193 [fix](https://github.com/ballerina-platform/ballerina-lang/pull/13193/files#diff-ff92e09a624e5a2806a9fc082ea15f20).

Shan1024:
@rdhananjaya - Seems like this is not yet fixed. This occurred in https://api.travis-ci.org/v3/job/493644201/log.txt.

MaryamZi:
Reopening since this failed again (on Travis):
```
Compiling source
    ballerina-examples/transactions-distributed:0.0.1

Running tests
    ballerina-examples/transactions-distributed:0.0.1
[ballerina/http] started HTTP/WS endpoint 172.17.0.1:34464
[...]

PR #13919, opened 2019-02-26
Fix intermitent failure in transactions BBE test

rdhananjaya:
## Purpose
$subject.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13607

### DIFF
--- a/examples/transactions-distributed/tests/transactions_distributed_test.bal
+++ b/examples/transactions-distributed/tests/transactions_distributed_test.bal
@@ -15,7 +15,6 @@ public function mockLogInfo(string | (function() returns (string)) msg) {
             outputs[count] = msg;
             count += 1;
         }
-        log:printWarn(msg);
     }
 }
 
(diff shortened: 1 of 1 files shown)


examples/transactions-distributed/tests/transactions_distributed_test.bal
```diff
@@ -46,7 +45,6 @@ function testFunc() {
     test:assertTrue(foundMatch(untaint outputs, "Got response from bizservice"));
 
     test:assertTrue(foundMatch(untaint outputs,  "Initiated transaction committed"));
-    test:assertTrue(foundMatch(untaint outputs, "Sent response back to client"));
```
rdhananjaya: Since we can't guarantee the happens before relationship between initiator (initiator logging  "Sent response back to client" message) and test case executing this assertion. To avoid intermittent test failure due to "happens before" violation, only option I see is to remove the check on this line.

I think this is fine because this logged message "Sent response back to client" has nothing to do with "distributed-transactions" logic.

Issue #13918, opened 2019-02-26
Test that searches for a module in Ballerina Central fails intermittently


NatashaWso2:
**Description:**
$subject

![screenshot from 2019-02-26 12-19-53](https://user-images.githubusercontent.com/13446645/53393015-2d1f2d00-39c1-11e9-95ad-f430be76c31d.png)

**Affected Versions:**
0.990.4 SNAPSHOT

Issue #13915, opened 2019-02-26
 Tests associated with building and running a module which is not inside a project fails in Mac OS  


NatashaWso2:
**Description:**
The test for building a module which is not inside a project in "ModuleBuildTestCase and the test for running a module which is not inside a project in "ModuleRunTestCase" fails in Mac OS. 

**Affected Versions:**
0.990.4 SNAPSHOT

**OS, DB, other environment details and versions:**
Mac OS

PR #13921, opened 2019-02-26
Fix intermittent test failures in ballerina build, run and search commands

NatashaWso2:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13915
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13918

### DIFF
--- a/tests/ballerina-tools-integration-test/src/test/java/org/ballerinalang/test/packaging/ListDependencyTestCase.java
+++ b/tests/ballerina-tools-integration-test/src/test/java/org/ballerinalang/test/packaging/ListDependencyTestCase.java
@@ -60,7 +60,7 @@ public class ListDependencyTestCase extends BaseTest {
                 "├── ballerina/io\n" +
                 "└── ballerina/math\n";
 
-        String actualOut = balClient.runMainAndReadStdOut("list", clientArgs, tempProjectDirectory.toString());
+        String actualOut = balClient.runMainAndReadStdOut("list", clientArgs, null, tempProjectDirectory.toString());
         Assert.assertEquals(actualOut, expectedOut);
     }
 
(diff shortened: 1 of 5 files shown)


tests/ballerina-tools-integration-test/src/test/java/org/ballerinalang/test/packaging/ListDependencyTestCase.java
```diff
@@ -60,7 +60,7 @@ public void testListDependenciesOfModule() throws Exception {
                 "├── ballerina/io\n" +
                 "└── ballerina/math\n";
 
-        String actualOut = balClient.runMainAndReadStdOut("list", clientArgs, tempProjectDirectory.toString());
+        String actualOut = balClient.runMainAndReadStdOut("list", clientArgs, null, tempProjectDirectory.toString());
```
Shan1024: I think it would be better to pass an empty map here and then we won't need the null check in the `runMainAndReadStdOut` method. Or we can create an overloaded method which will pass an empty map to the other overloaded method. In that case, we won't have to create an empty map as well.

Eg -
```java
public String runMainAndReadStdOut(String command, String[] args, Map<String, String> envProperties,
                                       String commandDir) throws BallerinaTestException {
    ...
}

public String runMainAndReadStdOut(String command, String[] args, String commandDir) throws BallerinaTestException {
    runMainAndReadStdOut(command, args, new HashMap<>(), commandDir);
}
```
WDYT?

hemikak: +1 to add another overloaded method.

NatashaWso2: Added with 5ed76dd

NatashaWso2:
Closing this PR in favor of https://github.com/ballerina-platform/ballerina-lang/pull/13995

Issue #13820, opened 2019-02-22
Inter-assignability between float, decimal constants and byte, int constants in a union type should fail


pranavan15:
**Description:**
Consider the below code,

```ballerina
const byte b = 45;
const float f = 34.3;
const int i = 45;
const decimal d = 34.3;

type T b|f;

public function main() {
    T t1 = i; // Case1
    T t2 = d; // Case2
}
```
The above two cases are invalid and should fail at compile time. However, that is not the case now as these cases compile and run successfully.

**Related Issues (optional):**
#13317 

**Suggested Labels (optional):**
<Type/Bug><Component/Compiler>

**Suggested Assignees (optional):**
@pranavan15

Issue #13587, opened 2019-02-12
Revisit isAssignableToFiniteType method logic in Types


pranavan15:
**Description:**
The logic of the method `isAssignableToFiniteType` of `Types.java` file needs to be revisited to ensure it allows  scenarios like below,

#### Scenario 1
```ballerina 
    const int ICONST = 5; 
    const float FCONST = 5;
    const decimal DCONST = 5;

    type NUM DCONST | FCONST | ICONST;  

    NUM n = 5;
```

#### Scenario 2
```ballerina 
    const float FCON = 5.0;
    const string SCON = "STR";

    type FOO SCON | FCON; 
    
    FOO f = 5.0000;
```
Alongside, new test cases need to be added for these scenarios.

**Suggested Labels (optional):**
<Type/Task>, <Component/Compiler>

**Suggested Assignees (optional):**
@pranavan15

Issue #13317, opened 2019-01-25
Finite type made up of decimal constants allows assigning a float constant with the same value


pranavan15:
**Description:**
Check the below code:

```ballerina
const decimal CONST1 = 23.2;
const decimal CONST2 = 2.12;
const float FCONST = 23.2;

type DEC CONST1|CONST2;

public function main() {
    DEC d1 = FCONST;
}
```
The above code is expected to fail at compile time with the error message `incompatible types: expected '23.2|2.12', found 'float'`. However, currently, it compiles and runs successfully.

**Related Issues (optional):**
https://github.com/ballerina-platform/ballerina-lang/issues/13256

**Suggested Labels (optional):**
<Type/Bug>, <Component/Compiler>, <Component/BVM>

**Suggested Assignees (optional):**
@pranavan15

Issue #10923, opened 2018-10-20
Cannot assign a byte value to a union typed variable


MaryamZi:
**Description:**
```ballerina
byte|float x = 4;
```
Compilation fails for the above with the following error:
```cmd
incompatible types: expected 'byte|float', found 'int'
```

**Affected Version**
0.982.0

pranavan15:
Reopening the issue as the above-referenced PR would only fix the finite type assignability. Also, it only checks for the assignability without changing the actual type of the literal.

MaryamZi:
Closing this since it will be fixed once we fix https://github.com/ballerina-platform/ballerina-lang/issues/14376.

PR #13967, opened 2019-02-27
Modify the finite type assignability logic

pranavan15:
## Purpose
This PR modifies the finite type assignability logic to check the literal assignability based on types.

* Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13317
* Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13820
* Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13587
* Fixes https://github.com/ballerina-platform/ballerina-lang/issues/10923

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -68,6 +68,8 @@ import org.wso2.ballerinalang.programfile.InstructionCodes;
 import org.wso2.ballerinalang.util.Flags;
 import org.wso2.ballerinalang.util.Lists;
 
+import java.math.BigDecimal;
+import java.math.MathContext;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -80,6 +82,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.Stack;
 import java.util.stream.Collectors;
+
+import static org.wso2.ballerinalang.compiler.semantics.model.SymbolTable.BBYTE_MAX_VALUE;
+import static org.wso2.ballerinalang.compiler.semantics.model.SymbolTable.BBYTE_MIN_VALUE;
 
 /**
  * This class consists of utility methods which operate on types.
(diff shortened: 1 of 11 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1618,21 +1623,88 @@ boolean isAssignableToFiniteType(BType type, BLangLiteral literalExpr) {
...
+                    return (byte) baseValue == ((Long) candidateValue).byteValue();
+                }
+                break;
+            case TypeTags.INT:
+                if (candidateTypeTag == TypeTags.INT) {
+                    return (long) baseValue == (long) candidateValue;
+                }
+                break;
+            case TypeTags.FLOAT:
+                double baseDoubleVal = Double.parseDouble(String.valueOf(baseValue));
```
hasithaa: Can this fail with number format exception? If yes, we need to handle this case.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1618,21 +1623,88 @@ boolean isAssignableToFiniteType(BType type, BLangLiteral literalExpr) {
...
+        int candidateTypeTag = candidateLiteral.type.tag;
+
+        // Numeric literal assignability is based on assignable type and numeric equivalency of values.
+        // If the base numeric literal is,
+        // (1) byte: we can assign byte or a int simple literal (Not an int constant) with the same value.
+        // (2) int: we can assign int literal or int constants with the same value.
+        // (3) float: we can assign int simple literal(Not an int constant) or a float literal/constant with same value.
+        // (4) decimal: we can assign int simple literal or float simple literal (Not int/float constants) or decimal
+        // with the same value.
+        switch (baseLiteral.type.tag) {
```
hasithaa: Here is an improvement. Check whether we can simplify this logic by taking out its core validation. IIRC, we have written the same validation for type checker as well. We might able to reuse the same core validation in both places.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
pranavan15: This parsing happens after checking the type tag of the literal. If we have `float` type tag for a literal then the literal value is guaranteed to be a numeric and a double parsable value.  For very large or small numbers those are out of the double range, the `parseDouble` method returns `Double.POSITIVE_INFINITY` or `Double.NEGATIVE_INFINITY` respectively. Hence, this **cannot** fail with a number format exception.

hasithaa: Ack.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
pranavan15: Ack. I will check on the possibilities of uniforming this logic to decouple value/type assignment and type compatibility checking logic currently we have in the type checker when visiting literals.

PR #13989, opened 2019-02-28
Fix dependency issues in gradle

pahans:
## Purpose
This PR fixes all existing dependency issues in gradle build

### DIFF
--- a/stdlib/database/h2/build.gradle
+++ b/stdlib/database/h2/build.gradle
@@ -4,11 +4,16 @@ dependencies {
     baloImplementation project(path: ':ballerina-builtin', configuration: 'baloImplementation')
     baloImplementation project(path: ':ballerina-sql', configuration: 'baloImplementation')
 
+    // transitive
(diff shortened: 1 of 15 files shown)


stdlib/database/h2/build.gradle
```diff
@@ -4,11 +4,16 @@ dependencies {
     baloImplementation project(path: ':ballerina-builtin', configuration: 'baloImplementation')
     baloImplementation project(path: ':ballerina-sql', configuration: 'baloImplementation')
 
+    // transitive
+    baloImplementation project(path: ':ballerina-sql', configuration: 'baloImplementation')
+    baloImplementation project(path: ':ballerina-config-api', configuration: 'baloImplementation')
+    baloImplementation project(path: ':ballerina-system', configuration: 'baloImplementation')
```
hasithaa: ballerina-system is not an implementation, is that correct. ?

pahans: should be a testCompile dependency. Ill revisit this and reopen PR

PR #14023, opened 2019-03-01
Remove 'maven-scm-publish-plugin' dependency

Shan1024:
## Purpose
This PR removes `maven-scm-publish-plugin` dependency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2134,12 +2134,6 @@
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-scm-publish-plugin</artifactId>
-                    <version>${maven.scm.publish.plugin.version}</version>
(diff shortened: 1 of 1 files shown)


hasithaa:
@keizer619 @madhuramendis Don't we use this for update site? If not +1 to remove this.

Manuri:
@hasithaa I don't think so.. AFAIR this was added to ballerina-parent along with the generic plugin set we had in wso2-parent when we were creating the separate ballerina-parent. 

Also, I did not specifically use this for any site update work when I was implementing the release pipeline back then. Even if it is used, AFAIK the site generation is not coupled to building this repo, so having maven plugin in the pom of ballerina repo is not a requirement.  
Better verify with @keizer619 @madhuramendis

keizer619:
@Manuri Yes this is not used for website generation and we can remove that

Issue #13619, opened 2019-02-15
[Spec Deviation] Module level variables cannot be public


MaryamZi:
**Description:**
According to the "Module variable declaration" section of the spec, module level variables cannot be public.

```
module-var-decl :=
   metadata
   [final]
   typed-binding-pattern = expression ;
```

"The scope of variables declared in a module-var-decl is the entire module. **Note that module variables are not allowed to be public.**"

But `public` module variables are currently allowed in the implementation.

**Affected Versions:**
0.990.3

MaryamZi:
The expression is also currently optional. But according to the spec, it is not.

PR #14032, opened 2019-03-01
Update module level variable/listener grammar

MaryamZi:
## Purpose
According to the "Module variable declaration" section of the spec, module level variables cannot be public.

```
module-var-decl :=
   metadata
   [final]
   typed-binding-pattern = expression ;
```

"The scope of variables declared in a module-var-decl is the entire module. **Note that module variables are not allowed to be public.**"

This PR disallows marking module level variables as public. Getters/setters are introduced where applicable in tests.

Also, the expression is now mandatory. This change was also introduced for listeners according to the spec:
```
listener-decl :=
   metadata
   public? listener [type-descriptor] identifier = expression ;

```

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13619


compiler/ballerina-lang/src/main/resources/grammar/BallerinaParser.g4
```diff
@@ -121,8 +121,9 @@ constantDefinition
     ;
 
 globalVariableDefinition
-    :   PUBLIC? LISTENER? typeName Identifier (ASSIGN expression)? SEMICOLON
-    |   PUBLIC? FINAL? (typeName | VAR) Identifier ASSIGN expression SEMICOLON
+    :   PUBLIC? LISTENER typeName Identifier (ASSIGN expression)? SEMICOLON
+    |   FINAL (typeName | VAR) Identifier ASSIGN expression SEMICOLON
+    |   (typeName | VAR) Identifier (ASSIGN expression)? SEMICOLON
```
Shan1024: If I'm not mistaken. in this rule (3rd) if `var` is used, the expression can be optional. But it should be mandatory.

MaryamZi: Thanks, missed this. Fixed with updates to the grammar and added tests.

Issue #13605, opened 2019-02-14
Intermittent test failure - WebSubSecureHubTestCase.testContentReceipt


Shan1024:
**Description:**
```
[ERROR] Tests run: 419, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 990.337 s <<< FAILURE! - in TestSuite
[ERROR] testContentReceipt(org.ballerinalang.test.service.websub.advanced.WebSubSecureHubTestCase)  Time elapsed: 45.019 s  <<< FAILURE!
org.ballerinalang.test.context.BallerinaTestException: Timeout expired waiting for matching log: WebSub Notification Received by One: {"mode":"internal", "content_type":"json"}
	at org.ballerinalang.test.service.websub.advanced.WebSubSecureHubTestCase.testContentReceipt(WebSubSecureHubTestCase.java:106)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   WebSubSecureHubTestCase.testContentReceipt:106 Â» BallerinaTest Timeout expired...
[INFO] 
[ERROR] Tests run: 419, Failures: 1, Errors: 0, Skipped: 0
[INFO] 
```

Complete log - [log-WebSubSecureHubTestCase.txt](https://github.com/ballerina-platform/ballerina-lang/files/2863066/log-WebSubSecureHubTestCase.txt)

chamil321:
This was not observed after that and could not reproduce. We've introduced logs to identify the possible cause by #14033, hence closes the issue.

PR #14033, opened 2019-03-01
[WebSub] Add log to identify error reported in #13605

chamil321:


stdlib/websub/src/main/java/org/ballerinalang/net/websub/nativeimpl/GetSubscribers.java
```diff
@@ -57,24 +59,32 @@
...
+                if (topic.equals(subscriber.getTopic())) {
+                    BMap<String, BValue> subscriberDetail = BLangVMStructs.createBStruct(
+                            structInfo, subscriber.getCallback(),
+                            subscriber.getSubscriptionDetails().get(SUBSCRIPTION_DETAILS_LEASE_SECONDS),
+                            subscriber.getSubscriptionDetails().get(SUBSCRIPTION_DETAILS_CREATED_AT));
+                    subscriberDetailArray.append(subscriberDetail);
+                }
             }
+        } catch (Exception ex) {
+            log.error("Error occurred while getting available subscribers: " + ex.getMessage(), ex);
```
riyafa: Printing the message is not needed if we are passing the exception to the logger. If we do this there will be duplicates.

PR #14039, opened 2019-03-01
Add array new and array store support

vinok88:
```
public function main(int i) returns string[] {
        int[5] c = [3,2,1,1,4];
        string[2] d = ["abc", "a"];
        return d;
}
```
This only supports fixed length arrays.


compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_instruction_gen.bal
```diff
@@ -335,6 +339,59 @@ type InstructionGenerator object {
...
+        // self.mv.visitVarInsn(LLOAD, valueIndex); //need to fix, this is only for int
+        int valueIndex = self.getJVMIndexOfVarRef(inst.rhsOp.variableDcl);
+        bir:BType valueType = inst.rhsOp.variableDcl.typeValue;
+        self.loadFromLocalVar(valueType, valueIndex);
+
+        string valueDesc = getMethodArgDesc(valueType); //pass the value type
+        self.mv.visitMethodInsn(INVOKEVIRTUAL, ARRAY_VALUE, "add", io:sprintf("(J%s)V", valueDesc), false);
+    }
+
+    function generateArrayValueLoad() {
```
warunalakshitha: Are we using this function in any place?

Kishanthan: i think it will be used when we have array load instruction

vinok88: Yes, not used ATM. Need to revisit once the array load available from BIR.

PR #14082, opened 2019-03-04
Fix gRPC module build failure due to missing dependent modules

daneshk:
## Purpose
Fix gRPC module build failure due to missing dependent modules
Sub task : #14081

PR #14087, opened 2019-03-04
Fix gradle build issues in runtime module

VijithaEkanayake:
## Purpose
$Subject.


stdlib/runtime/build.gradle
```diff
@@ -1,22 +1,27 @@
...
-    // TODO: Fix tests and remove test skip block
-    exclude '**/*'
-    
+    dependsOn copyGeneratedBalo
     useTestNG() {
         suites 'src/test/resources/testng.xml'
     }
-}
\ No newline at end of file
+    systemProperty "ballerina.home", "$buildDir"
```
pahans: @daneshk Have moved this line to [balNativeLibProject plugin.](https://github.com/ballerina-platform/ballerina-lang/pull/14082/files) so we don't need this line anymore

PR #14120, opened 2019-03-05
Add object value class generation

SupunS:


compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_constants.bal
```diff
@@ -206,4 +207,4 @@ const string MAP_TYPE = "org/ballerinalang/jvm/types/BMapType";
 const string UNION_TYPE = "org/ballerinalang/jvm/types/BUnionType";
 const string RECORD_TYPE = "org/ballerinalang/jvm/types/BRecordType";
 const string OBJECT_TYPE = "org/ballerinalang/jvm/types/BObjectType";
-const string BFIELD = "org/ballerinalang/jvm/types/BField";
+const string BFIELD = "org/ballerinalang/jvm/types/BField";
```
vinok88: Add newline

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_constants.bal
```diff
@@ -191,9 +191,10 @@ final int V1_8 = 0 << 16 | 52;
 // jvm runtime values related classes
 const string MAP_VALUE = "org/ballerinalang/jvm/values/MapValue";
 const string ARRAY_VALUE = "org/ballerinalang/jvm/values/ArrayValue";
+const string OBJECT_VALUE = "org/ballerinalang/jvm/values/ObjectValue";
```
Kishanthan: Shall we use some consistence naming convention for the constants here? May be we an use one pattern for jvm classes and one or ballerina types, values ?

SupunS: The current convention is as follows:

- Values: \<typename\>_TYPE 
- Type names: \<typename\>_VALUE 
- Other JVM classes: same name as their java class name

There are some cases that belongs to two of these categories. eg `java.lang.Object`. Those were named according to the 3rd category.

Kishanthan: ack, if we have a identifiable pattern

Issue #14079, opened 2019-03-04
Ballerina init command is not working properly on Desktop


rnavagamuwa:
**Description:**
`ballerina init` should create a sample project which includes a file called `hello_service.bal`. But this `hello_service.bal` file is not created when the `ballerina init` command is executd on desktop.

**Steps to reproduce:**
Execute `ballerina init` command on desktop.

**Affected Versions:**
- Ballerina 0.990.3

**OS, DB, other environment details and versions:**
- OS
    - ProductName:    Mac OS X
    - ProductVersion: 10.14.3
    - BuildVersion:   18D109

NatashaWso2:
@rnavagamuwa 'ballerina init' is designed to be like 'git init' where we initialize the current directory as a ballerina project. We do not explicitly create a project directory and initialize it. We just initialize the directory where 'ballerina init'  was executed as a ballerina project. So if you want to initialize a ballerina project using the 'ballerina init' command, first create a project directory and then run 'ballerina init' inside it.

rnavagamuwa:
@NatashaWso2 thanks for the detailed explanation. Since desktop is also a directory my personal opinion is `ballerina init` should work properly on the desktop too. As per my understanding `hello_service.bal` will be created only if the directory is empty but in Mac OS there is a file called `.DS_Store` which stores custom attributes of its containing folder. Please correct me if I'm wrong here but due to this `.DS_Store` file, `hello_service.bal` will not be created.

NatashaWso2:
@rnavagamuwa apologies for the late reply. Yes you are correct we need to ignore these hidden files. We will provide a proper fix for this since we will have to consider hidden files in other platforms as well.

pubudu91:
Closing since this was fixed with #14139

PR #14139, opened 2019-03-06
[Fix] Ballerina init command is not working properly on Mac OS

rnavagamuwa:
## Purpose
This PR is related to the issue [14079](https://github.com/ballerina-platform/ballerina-lang/issues/14079).


cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,26 @@
     private boolean helpFlag;
 
     private static boolean isDirEmpty(final Path directory) throws IOException {
+
         try (DirectoryStream<Path> dirStream = Files.newDirectoryStream(directory)) {
-            return !dirStream.iterator().hasNext();
+            //Check whether the OS is Mac OS and the folder contains .DS_Store file
```
Shan1024:
```suggestion
            // Check whether the OS is MacOS and the folder contains `.DS_Store` file.
```

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,26 @@
...
+            if (pathIterator.hasNext()) {
+
+                Path path = (Path) pathIterator.next();
+                Path filePath = path.getFileName();
+                if (filePath != null && filePath.toString().equals(ProjectDirConstants.MAC_OS_DS_STORE_FILE)) {
+
+                    return System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("mac") &&
+                            !pathIterator.hasNext();
+                } else {
+
```
Shan1024: I think it is better to remove this newline.

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,26 @@
...
+                    return false;
+                }
+
```
Shan1024: I think it is better to remove this newline.

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -161,7 +180,8 @@ public void execute() {
                         do {
                             out.print("Module for the service: (no module) ");
                             packageName = scanner.nextLine().trim();
-                        } while (!validatePkgName(projectPath, packageName));
+                        }
+                        while (!validatePkgName(projectPath, packageName));
```
Shan1024: Better to revert this change. Can be applied to few `while` loops which has been modified below as well.

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,26 @@
     private boolean helpFlag;
 
     private static boolean isDirEmpty(final Path directory) throws IOException {
+
```
Shan1024: Shall we remove this empty newline which is added to all of the methods?

hasithaa:
@NatashaWso2 @hemikak Can you check this PR?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,25 @@
...
         try (DirectoryStream<Path> dirStream = Files.newDirectoryStream(directory)) {
-            return !dirStream.iterator().hasNext();
+            //Check whether the OS is MacOS and the folder contains .DS_Store file
+            Iterator pathIterator = dirStream.iterator();
+            if (pathIterator.hasNext()) {
+
+                Path path = (Path) pathIterator.next();
+                Path filePath = path.getFileName();
+                if (filePath != null && filePath.toString().equals(ProjectDirConstants.MAC_OS_DS_STORE_FILE)) {
```
NatashaWso2: We can simplify this if condition as below:
```java
if (filePath != null && filePath.toString().equals(ProjectDirConstants.MAC_OS_DS_STORE_FILE)) {
   return true;
}
return false;
```

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,25 @@
...
+
+                    return System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("mac") &&
+                            !pathIterator.hasNext();
```
NatashaWso2: Why do we need to check this condition again? 
```java
 !pathIterator.hasNext()
```

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,25 @@
     private boolean helpFlag;
 
     private static boolean isDirEmpty(final Path directory) throws IOException {
+
         try (DirectoryStream<Path> dirStream = Files.newDirectoryStream(directory)) {
-            return !dirStream.iterator().hasNext();
+            //Check whether the OS is MacOS and the folder contains .DS_Store file
+            Iterator pathIterator = dirStream.iterator();
+            if (pathIterator.hasNext()) {
+
```
NatashaWso2: Shall we remove this extra line?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/util/ProjectDirConstants.java
```diff
@@ -56,4 +56,7 @@ private ProjectDirConstants() {
     public static final String BALLERINA_SOURCE_ROOT = "ballerina.source.root";
 
     public static final String NIGHTLY_BUILD = "nightly.build";
+
+    public static final String MAC_OS_DS_STORE_FILE = ".DS_Store";
```
NatashaWso2: Shall we rename the variable as 'DS_STORE_FILE' without specifying the os name?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -256,6 +277,7 @@ private Manifest createManifest(Scanner scanner, String createToml) {
      */
     @Override
     public String getName() {
+
```
NatashaWso2: Shall we remove this empty newline?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -264,6 +286,7 @@ public String getName() {
      */
     @Override
     public void printLongDesc(StringBuilder out) {
+
```
NatashaWso2: Shall we remove this empty newline?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -274,6 +297,7 @@ public void printLongDesc(StringBuilder out) {
      */
     @Override
     public void printUsage(StringBuilder out) {
+
```
NatashaWso2: Shall we remove this empty newline?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,25 @@
     private boolean helpFlag;
 
     private static boolean isDirEmpty(final Path directory) throws IOException {
+
         try (DirectoryStream<Path> dirStream = Files.newDirectoryStream(directory)) {
-            return !dirStream.iterator().hasNext();
+            //Check whether the OS is MacOS and the folder contains .DS_Store file
+            Iterator pathIterator = dirStream.iterator();
+            if (pathIterator.hasNext()) {
```
NatashaWso2: We can fail fast here by returning 'true' if there are no files inside the directory?

```java
if (!pathIterator.hasNext()) { 
 return true;
}
```

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
rnavagamuwa: Otherwise, we won't be able to detect if there are more files other than the `.DS_Store`. Assume there are two files and one is the `.DS_Store`. Iterator picks the `.DS_Store` as its first item. Then there should be a way to check the additional files. In this case `!pathIterator.hasNext()` will return false, but if this is not here then the whole function will return `true` which is not the case.

NatashaWso2: Okay got it :) So we can remove the check for the os and have only the check for the file name

rnavagamuwa: Don't we have to check the OS as well? What happens if there is a manually created `.DS_Store` file in a Linux system?

NatashaWso2: There is a rare chance of that happening right :) Even if such a file was created we would want to generate the ballerina source file. So I think it would be better if we remove the check added for the os name.

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,21 @@
...
-            return !dirStream.iterator().hasNext();
+            //Check whether the OS is MacOS and the folder contains .DS_Store file
+            Iterator pathIterator = dirStream.iterator();
+            if (!pathIterator.hasNext()) {
+                return true;
+            }
+            Path path = (Path) pathIterator.next();
+            Path filePath = path.getFileName();
+            if (filePath != null && filePath.toString().equals(ProjectDirConstants.DS_STORE_FILE)) {
+
```
NatashaWso2: Shall we remove all extra new lines added?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -236,15 +251,17 @@ private Manifest createManifest(Scanner scanner, String createToml) {
             do {
                 out.print("Organization name: (" + defaultOrg + ") ");
                 orgName = scanner.nextLine().trim();
-            } while (!validateOrgName(orgName));
+            }
```
NatashaWso2: Why have we moved the while to a new line?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -236,15 +251,17 @@ private Manifest createManifest(Scanner scanner, String createToml) {
...
             // Set org-name
             manifest.setName(orgName.isEmpty() ? defaultOrg : orgName);
             String version;
             do {
                 out.print("Version: (" + DEFAULT_VERSION + ") ");
                 version = scanner.nextLine().trim();
                 version = version.isEmpty() ? DEFAULT_VERSION : version;
-            } while (!validateVersion(out, version));
+            }
+            while (!validateVersion(out, version));
```
NatashaWso2: Why have we moved the while to a new line?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
```diff
@@ -67,8 +68,21 @@
...
+
         try (DirectoryStream<Path> dirStream = Files.newDirectoryStream(directory)) {
-            return !dirStream.iterator().hasNext();
+            //Check whether the OS is MacOS and the folder contains .DS_Store file
+            Iterator pathIterator = dirStream.iterator();
+            if (!pathIterator.hasNext()) {
+                return true;
+            }
+            Path path = (Path) pathIterator.next();
+            Path filePath = path.getFileName();
```
NatashaWso2: Shall we change this variable name to 'fileName' since we are referring to the file name?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/InitCommand.java
rnavagamuwa: If that's the case will remove the OS check. Then the logic can be simplified to 

`return fileName != null && fileName.toString().equals(ProjectDirConstants.DS_STORE_FILE) && !pathIterator.hasNext();`

NatashaWso2:
@rnavagamuwa can you resolve the conflicts?

rnavagamuwa:
@NatashaWso2 All the conflicts have been resolved right?

PR #14143, opened 2019-03-06
Support error constructor in BIR and JVM target

warunalakshitha:
## Purpose
This will add following.

1. Support error literal in BIR
2. JVM Codegen

ex. 

```
public function main() returns error {
    map<string> m = {"message":"cause"};
    error e = error("Reason", m);
    error x = foo(e);
    return x;
}

function foo(error e) returns error{
    return e;
}
```
will print Reason {"message":"cause"}


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BTypes.java
```diff
@@ -41,7 +41,7 @@
     public static BType typeXMLAttributes = new BXMLAttributesType(TypeConstants.XML_ATTRIBUTES_TNAME, null);
     // public static BType typeIterator = new BIteratorType(TypeConstants.ITERATOR_TNAME, null);
     // public static BType typeChannel = new BChannelType(TypeConstants.CHANNEL, null);
-    // public static BErrorType typeError = new BErrorType(TypeConstants.ERROR, typeString, typeMap, null);
+    public static BErrorType typeError = new BErrorType(TypeConstants.ERROR, null, typeString, typeMap);
```
vinok88: If we do not need the package path, shall we remove it from the constructor itself.

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_terminator_gen.bal
```diff
@@ -140,11 +144,14 @@ type TerminatorGenerator object {
...
             } else if (bType is bir:BArrayType ||
                         bType is bir:BMapType ||
+                        bType is bir:BErrorType ||
                         bType is bir:BTypeAny ||
                         bType is bir:BTypeNil ||
                         bType is bir:BObjectType ||
                         bType is bir:BUnionType) {
                 self.mv.visitVarInsn(ASTORE, lhsLndex);
+            } else if (bType is bir:BErrorType) {
+                self.mv.visitVarInsn(ASTORE, lhsLndex);
```
SupunS: we don't need this line, do we?

btw, this should give a compile error, if not its a bug in type narrowing i believe.

warunalakshitha: Yes. We do not need this line since line 147 covers. This seems to be added again when fixing some conflicts. Any we need line 147 to return error type if not you will get an exception when returning the error.

 JVM generation is not supported for type {reasonType:"string", detailType:{constraint:"any"}} {}
	at compiler_backend_jvm:0.0.0:genCallTerm(jvm_terminator_gen.bal:154)

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BTypes.java
warunalakshitha: Fixed

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_terminator_gen.bal
warunalakshitha: Fixed with last commit

PR #14155, opened 2019-03-06
Fix gradle build test failures in HTTP module

chamil321:

### DIFF
--- a/stdlib/http/build.gradle
+++ b/stdlib/http/build.gradle
@@ -1,9 +1,6 @@
 apply from: "$rootDir/gradle/balNativeLibProject.gradle"
 
 dependencies {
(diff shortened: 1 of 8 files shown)


stdlib/http/build.gradle
```diff
@@ -55,22 +58,21 @@ dependencies {
...
 
-task copyGeneratedBalo(type: org.gradle.api.tasks.Copy) {
-    from "$buildDir/generated-balo/repo/ballerina"
-    into "$buildDir/lib/repo/ballerina"
-}
-
 test {
-    // TODO: Fix tests and remove test skip block
-    exclude '**/*'
+    minHeapSize = "256m"
```
pahans: shall we move this to `balNativeLibProject.gradle`?

PR #14156, opened 2019-03-06
Set file encoding to utf-8 in gradle.properties file

Anoukh:
## Purpose
$title

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,3 @@
 org.gradle.caching=true
 org.gradle.parallel=true
+org.gradle.jvmargs='-Dfile.encoding=UTF-8'


gradle.properties
```diff
@@ -1,2 +1,3 @@
 org.gradle.caching=true
 org.gradle.parallel=true
+org.gradle.jvmargs='-Dfile.encoding=UTF-8'
```
pahans: any specific reason for this?

Anoukh: Unit tests are failing in windows OS due to an encoding issue. @hasithaa can verify if this solves the issue?

hasithaa: Yes, this solves test failures in Windows. Still, we need https://github.com/wso2/code-quality-tools/pull/9 to build on windows.

PR #14178, opened 2019-03-07
Fix testerina and spec conformance tests for gradle build

Anoukh:
## Purpose
Fix testerina and spec test gradle build


hasithaa:
@pahans Can you review this PR ?

gradle/balNativeLibProject.gradle
```diff
@@ -1,36 +1,4 @@
...
-    }
-}
-
-
-task createBalHome(type: org.gradle.api.tasks.Copy) {
-    from configurations.baloImplementation
-    into "$buildDir/lib"
-
-}
+apply from: "$rootDir/gradle/testableBaloProject.gradle"
```
pahans: shall we rename this to 'baloProject'?

Anoukh: Shall I rename to `javaProjectWithExtBalo.gradle` since this will not be used for balo projects, but only for java projects that need to include balos from other modules?

PR #14186, opened 2019-03-07
Update checkstyle rule version

hasithaa:


pom.xml
```diff
@@ -2003,7 +2003,7 @@
                             <phase>validate</phase>
                             <configuration>
                                 <configLocation>
-                                    https://raw.githubusercontent.com/wso2/code-quality-tools/v1.1/checkstyle/checkstyle.xml
+                                    https://raw.githubusercontent.com/wso2/code-quality-tools/v1.2/checkstyle/checkstyle.xml
```
pahans: with this we can remove encoding="UTF-8" also. right?

hasithaa: yes, +1.

PR #14188, opened 2019-03-07
Add fields to configure ssl session timeout and handshake timeout

Bhashinee:
## Purpose
This PR is to introduce SSL session and handshake timeout configurations.

Two new fields were introduced in secure socket record. Those are,
1. handshakeTimeout
2. sessionTimeout


bvm/ballerina-core/src/main/java/org/ballerinalang/connector/impl/StructImpl.java
```diff
@@ -79,6 +79,13 @@ public long getIntField(String fieldName) {
         return ((BInteger) value.get(fieldName)).intValue();
     }
 
+    public long getDefaultableIntField(String fieldName) {
```
VijithaEkanayake: Shall we use ternary operator here? WDYT?
return (value.get(fieldName) != null) ? getIntField(fieldName) : 0;

stdlib/http/src/main/java/org/ballerinalang/net/http/HttpUtil.java
```diff
@@ -1408,6 +1412,14 @@ public static void populateSSLConfiguration(SslConfiguration sslConfiguration, S
         sslConfiguration.setOcspStaplingEnabled(ocspStaplingEnabled);
         sslConfiguration.setHostNameVerificationEnabled(hostNameVerificationEnabled);
 
+        if (sessionTimeout > 0) {
```
VijithaEkanayake: Could you check whether this (sessionTimeout > 0) check is required?

tests/ballerina-integration-test/src/test/resources/mutualSSL/mutual_ssl_client.bal
```diff
@@ -35,7 +35,9 @@ http:ClientEndpointConfig mutualSslClientConf = {
         certValidation: {
             enable: false
         },
-        ocspStapling: false
+        ocspStapling: false,
+        handshakeTimeout: 20,
+        sessionTimeout:30
```
VijithaEkanayake:
```suggestion
        sessionTimeout: 30
```

Issue #13814, opened 2019-02-22
[Messaging] Implement Artemis Connector 


pamod:
**Description:**
Implement Artemis connector for Ballerina.

PR #14193, opened 2019-03-07
Implement ActiveMQ Artemis Connector

riyafa:
## Purpose
This has the basic implementation for the ActiveMQ Artemis connector based on [this](https://docs.google.com/document/d/1mDFfm4CAJDYP3ZQoe-zCCFLw1hix4TtcSi7h9zOs6RI/edit?usp=sharing) design.

Related issue https://github.com/ballerina-platform/ballerina-lang/issues/13814

**Examples**
- [point_to_point_consumer.bal](https://gist.github.com/riyafa/00a0673b12d18f59a283b1eace441f56) 
- [point_to_point_producer.bal](https://gist.github.com/riyafa/701f23b50f3a2e899f2771e44ef556b2)
- [pub_sub_producer.bal](https://gist.github.com/riyafa/b647186dfc256778f2c61c080877021a)
- [pub_sub_consumer.bal](https://gist.github.com/riyafa/52f2aa7d432ff7f2b462d03afcf481a0)


stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/artemis_commons.bal
```diff
@@ -0,0 +1,50 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/io;
+
+# Constant for the artemis error code.
+public const string ARTEMIS_ERROR_CODE = "{ballerina/artemis}ArtemisError";
```
MaryamZi: We can just say

```suggestion
public const ARTEMIS_ERROR_CODE = "{ballerina/artemis}ArtemisError";
```

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/artemis_commons.bal
```diff
@@ -0,0 +1,50 @@
...
+
+# The Artemis error record.
+#
+# + message - the error message.
+public type ArtemisError record {
+    string message = "";
```
MaryamZi: Shall we make this an optional field if the message is not mandatory instead of setting a default value?

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/artemis_commons.bal
```diff
@@ -0,0 +1,50 @@
...
+#
+# + host - The host
+# + port - The port
+# + username - The username
+# + password - The password
+public type URLConfiguration record {
+    string host;
+    int port;
+    string? username = ();
+    string? password = ();
```
MaryamZi: Again, should `username` and `password` be optional fields? Applicable elsewhere too.

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/message.bal
```diff
@@ -0,0 +1,130 @@
...
+            self.createMessage(session, data, self.configuration);
+        }
+    }
+
+    extern function createMessage(Session session, string | byte[] | map<string | int | float | byte | boolean | byte[]> | io:ReadableByteChannel data, MessageConfiguration config);
+    # Acknowledges reception of this message.
+    #
+    # + return - If an error occurred while acknowledging the message
+
+    public remote extern function acknowledge() returns error?;
```
aashikam: Small formatting issue

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/connection.bal
```diff
@@ -0,0 +1,68 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+# Represents ActiveMQ Artemis Connection
+public type Connection client object {
+
+    public function __init(string url, ConnectionConfiguration? config = ()) {
```
MaryamZi: Shall we add documentation?

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/message.bal
```diff
@@ -0,0 +1,130 @@
...
+
+    # The message payload
+    #
+    # + return - The message payload or error on failure to retrieve payload or if the type is unsupported.
+    #  A map payload can contain an error if the type is unsupported.
+    public extern function getPayload() returns string | byte[] | map<string | int | float | byte | boolean | byte[]> | error;
+
+    # Call this function to save to a WritableByteChannel if the message is `STREAM` type
+    #
+    # + ch - The byte channel to save to
```
aashikam: Shall we make this variable channel?

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/message.bal
```diff
@@ -0,0 +1,130 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+# Represents ActiveMQ Artemis Message
+public type Message client object {
+
+    private MessageType messageType = TEXT;
+    private MessageConfiguration configuration;
+
+    public function __init(Session session, io:ReadableByteChannel | int | float | byte | boolean | string |
```
MaryamZi: I believe the format we've followed is to not have spaces around `|` with union types.

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/message.bal
```diff
@@ -0,0 +1,130 @@
...
+            self.messageType = BYTES;
+            self.createMessage(session, data, self.configuration);
+        }
+    }
+
+    extern function createMessage(Session session, string | byte[] | map<string | int | float | byte | boolean | byte[]> | io:ReadableByteChannel data, MessageConfiguration config);
+    # Acknowledges reception of this message.
+    #
+    # + return - If an error occurred while acknowledging the message
+
```
MaryamZi: extra new line? A new line is needed at L54.

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/message.bal
riyafa: `channel` is a keyword in ballerina

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/artemis_commons.bal
```diff
@@ -0,0 +1,50 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/io;
+
```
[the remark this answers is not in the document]
riyafa: It is used in the package so importing it in one place here.

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/producer.bal
```diff
@@ -0,0 +1,78 @@
...
+
+    extern function createProducer(string addressName, AddressConfiguration addressConfig, int rate);
+
+    # Sends a message to the producer's address
+    #
+    # + data - the `Message` or data to send 
+    # + return - `error` on failure
+    public remote function send(int | float | string | json | xml | byte | byte[] | map<string | int | float | byte
+| boolean | byte[]> | io:ReadableByteChannel | Message data) returns error? {
+        Message msg;
```
chamil321:
```suggestion
        Message msg = data is Message ? data : new(self.session, data);
```

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/producer.bal
```diff
@@ -0,0 +1,78 @@
...
+        if (data is Message) {
+            msg = data;
+        } else {
+            msg = new(self.session, data);
+        }
```
chamil321: Shall we achieved above using ternary operator?

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/ArtemisUtils.java
```diff
@@ -0,0 +1,159 @@
...
+        } else if (obj instanceof Double) {
+            return new BFloat((double) obj);
+        } else if (obj instanceof Boolean) {
+            return new BBoolean((boolean) obj);
+        } else if (obj instanceof Byte) {
+            return new BByte((byte) obj);
+        } else if (obj instanceof byte[]) {
+            return new BValueArray((byte[]) obj);
+        } else {
+            return ArtemisUtils.getError(context, new BallerinaException("Unsupported type"));
```
chamil321:
```suggestion
            return ArtemisUtils.getError(context, "Unsupported type");
```
Better if the error message is sent directly as you have defined two methods to create error

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/ArtemisUtils.java
```diff
@@ -0,0 +1,159 @@
...
+        @SuppressWarnings(ArtemisConstants.UNCHECKED)
+        BMap<String, BValue> sessionObj = (BMap<String, BValue>) obj.get("session");
+        return (ClientSession) sessionObj.getNativeData(ArtemisConstants.ARTEMIS_SESSION);
+    }
+
+    public static void closeIfAnonymousSession(BMap<String, BValue> obj) throws ActiveMQException {
+        boolean anonymousSession = ((BBoolean) obj.get("anonymousSession")).booleanValue();
+        if (anonymousSession) {
+            ClientSession session = ArtemisUtils.getClientSessionFromBMap(obj);
+            if (session.isClosed()) {
```
chamil321: Why do you close again if the session is already closed? Shouldn't the condition be negation?

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/connection/Close.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+@BallerinaFunction(
+        orgName = ArtemisConstants.BALLERINA, packageName = ArtemisConstants.ARTEMIS,
+        functionName = "close",
+        receiver = @Receiver(type = TypeKind.OBJECT, structType = ArtemisConstants.CONNECTION_OBJ,
+                             structPackage = ArtemisConstants.PROTOCOL_PACKAGE_ARTEMIS),
+        isPublic = true
+)
+public class Close implements NativeCallableUnit {
```
chamil321: Any particular reason to implement this blocking function using NativeCallableUnit? Why not extend BlockingNativeCallableUnit?

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/connection/CreateConnection.java
```diff
@@ -0,0 +1,114 @@
...
+            connectionPool.setInitialConnectAttempts(initialConnectAttempts);
+            connectionPool.setConfirmationWindowSize(1024);
+
+            ClientSessionFactory factory = connectionPool.createSessionFactory();
+
+            connection.addNativeData(ArtemisConstants.ARTEMIS_CONNECTION_POOL, connectionPool);
+            connection.addNativeData(ArtemisConstants.ARTEMIS_SESSION_FACTORY, factory);
+
+        } catch (Exception e) {
+            ArtemisUtils.throwBallerinaException("Error occurred while starting connection.", context, e);
```
chamil321: Shouldn't it better if we log error from the class which the error is occurred rather than calling util class to log?

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/CreateMessage.java
```diff
@@ -0,0 +1,159 @@
...
+                @Argument(name = "session", type = TypeKind.OBJECT, structType = ArtemisConstants.SESSION_OBJ),
+                @Argument(name = "data", type = TypeKind.UNION),
+                @Argument(name = "config", type = TypeKind.RECORD, structType = "ConnectionConfiguration")
+        }
+)
+public class CreateMessage implements NativeCallableUnit {
+
+    @Override
+    public void execute(Context context, CallableUnitCallback callableUnitCallback) {
+        @SuppressWarnings(ArtemisConstants.UNCHECKED)
```
chamil321: Shall we exclude the warning having single SuppressWarning for this method?

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/CreateMessage.java
```diff
@@ -0,0 +1,159 @@
...
+        args = {
+                @Argument(name = "session", type = TypeKind.OBJECT, structType = ArtemisConstants.SESSION_OBJ),
+                @Argument(name = "data", type = TypeKind.UNION),
+                @Argument(name = "config", type = TypeKind.RECORD, structType = "ConnectionConfiguration")
+        }
+)
+public class CreateMessage implements NativeCallableUnit {
+
+    @Override
+    public void execute(Context context, CallableUnitCallback callableUnitCallback) {
```
chamil321:
```suggestion
    @SuppressWarnings(ArtemisConstants.UNCHECKED)
    public void execute(Context context, CallableUnitCallback callableUnitCallback) {
```

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/message.bal
riyafa: It seems to be the formatting style in VSCode. Refer mail: https://groups.google.com/d/msg/ballerina-dev/Ayv1J9QdI0s/Bxm1PyB7GQAJ
I am using VSCode for formatting

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/connection.bal
```diff
@@ -0,0 +1,68 @@
...
+};
+
+# Configurations related to a Artemis `Connection`
+#
+# + connectionTTL - Connection's time-to-live. negative to disable or greater or equals to 0
+# + callTimeout - The blocking calls timeout in milliseconds
+# + consumerWindowSize - Window size in bytes for flow control of the consumers created through this `Connection`
+# + consumerMaxRate - Maximum rate of message consumption for consumers created through this `Connection`
+# + producerWindowSize - Window size for flow control of the producers created through this `Connection`
+# + producerMaxRate - The maximum rate of message production for producers created through this `Connection`
```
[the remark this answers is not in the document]
riyafa: I think `producerMaxRate` makes more sense because we are not doing **production** here :)

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/connection.bal
```diff
@@ -0,0 +1,68 @@
...
+    # + return - `error` if an error occurs closing the connection or nil
+    public remote extern function close() returns error?;
+};
+
+# Configurations related to a Artemis `Connection`
+#
+# + connectionTTL - Connection's time-to-live. negative to disable or greater or equals to 0
+# + callTimeout - The blocking calls timeout in milliseconds
+# + consumerWindowSize - Window size in bytes for flow control of the consumers created through this `Connection`
+# + consumerMaxRate - Maximum rate of message consumption for consumers created through this `Connection`
```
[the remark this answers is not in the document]
riyafa: On similar ground as for producer I think it is better to leave this as `consumerMaxRate`

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/listener.bal
```diff
@@ -0,0 +1,93 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+# Represents ActiveMQ Artemis Listener. This is an abstraction for that includes the connection and session. 
+# Consumers are represented by the service attaching to this listener.
+public type Listener object {
+    *AbstractListener;
+    private Session session;
+    private boolean anonymousSession;
```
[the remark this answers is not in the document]
riyafa: It means that it has not been created explicitly by the user but is created implicitly and is anonymous to the user. Do you have a better suggestion for a name?

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/message.bal
```diff
@@ -0,0 +1,130 @@
...
+    # + key - The name of the property
+    # + value - The value of the property 
+    # + return - If an error occures while setting the property
+    public extern function putProperty(string key, string | int | float | boolean | byte | byte[] value) returns error?;
+
+    # Get a message property
+    #
+    # + key - The name of the property
+    # + return - The value of the property or nil if not found
+    public extern function getProperty(string key) returns string | int | float | boolean | byte | byte[] | () | error;
```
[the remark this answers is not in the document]
riyafa: It has been concluded to have spaces around `|` in Ballrina formatting. Refer mail: https://groups.google.com/d/msg/ballerina-dev/Ayv1J9QdI0s/Bxm1PyB7GQAJ

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/CreateMessage.java
riyafa: If we have it for the method it will tend to suppress all warnings. I would like to suppress only the relevant warnings.

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/CreateMessage.java
riyafa: explained in previous comment

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/producer.bal
riyafa: This conversion is not possible because `data` is not recognized as type Message inside the ternary statement

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/producer.bal
riyafa: explained in above comment

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/artemis_commons.bal
riyafa: If we use the optional fields we need to do a lot of boiler plate error handling because of a key missing instead of being nil.  In this case it is more suitable to have a nil value instead of an optional field.

MaryamZi: If we use index based access, I don't see any difference. 

Say we have a record value `urlConfig` (of type `URLConfiguration`).

Currently when you say `urlConfig.username` the resultant value is of type `string?`. It will be the same if `username` is an optional field in the record and you say `urlConfig["username"]`. 

Could you explain what you meant by "boiler plate error handling" ?

With this approach we allow the user to set the `username` as `()` which is wrong IMO. I believe we should be minimizing room for user error even if we have to write a bit of extra code.

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/producer.bal
MaryamZi: Can you share the error please, this should work IINM? Tried with the following simple example:
```ballerina
import ballerina/io;

    string|int|Message m = "hello";
    testTernary(m);
}

function testTernary(string|int|Message m) {
    Message m2 = m is Message ? m : new Message(m);
    m2->printText();    
}

public type Message client object {
    string|int text;

    public function __init(string|int s) {
        self.text = s;
    }

    remote function printText() {
        io:println(self.text);
    }
};
```

riyafa: That seems to work. I have been mistaken. Thanks..

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/artemis_commons.bal
riyafa: I have fixed this in this place. But it happens to be that username and password could be null in the Artemis broker protocol which they handle. So didn't change it in the session.

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/connection.bal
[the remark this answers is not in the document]
aashikam: "The maximum rate of message production"

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/consumer/CreateConsumer.java
```diff
@@ -0,0 +1,194 @@
...
+            @SuppressWarnings(ArtemisConstants.UNCHECKED)
+            BMap<String, BValue> consumerObj = (BMap<String, BValue>) context.getRefArgument(0);
+
+            ClientSession session = ArtemisUtils.getClientSessionFromBMap(consumerObj);
+
+            Service service = BLangConnectorSPIUtil.getServiceRegistered(context);
+
+            Annotation serviceAnnotation = getServiceConfigAnnotation(service);
+            Struct annotationValue = serviceAnnotation.getValue();
+            boolean autoAck = annotationValue.getBooleanField("autoAck");
```
wggihan: Shall we create constants for these fields as well? Check the other places.

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/consumer/Start.java
```diff
@@ -0,0 +1,71 @@
...
+            @SuppressWarnings(ArtemisConstants.UNCHECKED)
+            BMap<String, BValue> listenerObj = (BMap<String, BValue>) context.getRefArgument(0);
+            listenerObj.addNativeData(ArtemisConstants.COUNTDOWN_LATCH, countDownLatch);
+            @SuppressWarnings(ArtemisConstants.UNCHECKED)
+            BMap<String, BValue> sessionObj = (BMap<String, BValue>) listenerObj.get("session");
+            ClientSession session = (ClientSession) sessionObj.getNativeData(ArtemisConstants.ARTEMIS_SESSION);
+            session.start();
+            new Thread(() -> {
+                try {
+                    countDownLatch.await();
```
wggihan: Why we can't use NativeCallableUnit and implement this in nonblocking manner?

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/GetBodySize.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public void execute(Context context) {
+        try {
+            @SuppressWarnings(ArtemisConstants.UNCHECKED)
+            BMap<String, BValue> messageObj = (BMap<String, BValue>) context.getRefArgument(0);
+            ClientMessage message = (ClientMessage) messageObj.getNativeData(ArtemisConstants.ARTEMIS_MESSAGE);
+            context.setReturnValues(new BInteger(message.getBodySize()));
+        } catch (Exception e) {
+            context.setReturnValues(ArtemisUtils.getError(context, e));
```
wggihan: Why we can't handle all the specific exceptions here instead of this? Check the other places as well.

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/GetProperty.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public void execute(Context context) {
+        @SuppressWarnings(ArtemisConstants.UNCHECKED)
+        BMap<String, BValue> messageObj = (BMap<String, BValue>) context.getRefArgument(0);
+        ClientMessage message = (ClientMessage) messageObj.getNativeData(ArtemisConstants.ARTEMIS_MESSAGE);
+
+        String key = context.getStringArgument(0);
+        Object property = message.getObjectProperty(key);
+        if (property != null) {
```
wggihan: Don't we need to handle null part and send an error?

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/session/CreateSession.java
```diff
@@ -0,0 +1,98 @@
...
+        } catch (Exception e) {
+            ArtemisUtils.throwBallerinaException("Error occurred while starting session", context, e, logger);
+        }
+    }
+
+    @Override
+    public boolean isBlocking() {
+        return true;
+    }
+
```
wggihan: Extra lines. Please check other places as well.

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/GetProperty.java
riyafa: If it is null it returns nil.

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/GetBodySize.java
riyafa: Since this method returns an error we do not want this method to throw an exception and fail. Instead we make sure it returns all errors. This is the way most of the Ballerina functions are written including http ones. 
Eg:
https://github.com/ballerina-platform/ballerina-lang/blob/019bb4a98312a3a8eb860b9374f672dd6fca2023/stdlib/mime/src/main/java/org/ballerinalang/mime/nativeimpl/GetBodyAsString.java#L68-L71
https://github.com/ballerina-platform/ballerina-lang/blob/019bb4a98312a3a8eb860b9374f672dd6fca2023/stdlib/mime/src/main/java/org/ballerinalang/mime/nativeimpl/GetByteArray.java#L92-L95

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/GetProperty.java
wggihan: I think it will be much more readable if we handle the else part as well. WDYT?

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/GetBodySize.java
wggihan: We need to come to a final decision about this and apply to all the places.

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/consumer/Start.java
riyafa: If we do that the start function never returns. In integration tests service deployment is checked using returning of this function. If it does not return the integration test suite hangs. Hence this is the only viable solution we could think of. @lankavitharana might be able to explain.

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/consumer/CreateConsumer.java
riyafa: I am using these values in only one place. Hence it is redundant to define constants for each of these values. For values that I am using more than once I have created constants as it is the best practice.

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/message/GetProperty.java
riyafa: Fixed

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/session/CreateSession.java
```diff
@@ -0,0 +1,96 @@
...
+        BMap<String, BValue> config = (BMap<String, BValue>) context.getRefArgument(2);
+
+        ServerLocator serverLocator = (ServerLocator) connection.getNativeData(
+                ArtemisConstants.ARTEMIS_CONNECTION_POOL);
+        ClientSessionFactory sessionFactory =
+                (ClientSessionFactory) connection.getNativeData(ArtemisConstants.ARTEMIS_SESSION_FACTORY);
+        try {
+            String username = null;
+            String password = null;
+            BValue userValue = config.get("username");
```
pamod: Shall we define constants ?

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/consumer/Start.java
wggihan: Can you explain the functionality of this start method? Shall we schedule a code review for this? I think that will be much easier to discuss this. Main implementation shouldn't depend on outside factors like test suites.

stdlib/messaging/activemq-artemis/src/main/ballerina/artemis/listener.bal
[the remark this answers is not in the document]
daneshk: If this is a property of the session, shall we move it to Session object and call it `anonymous`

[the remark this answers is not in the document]
riyafa: It is a property of the Listener not the session

stdlib/messaging/activemq-artemis/src/main/java/org/ballerinalang/messaging/artemis/externimpl/consumer/Start.java
riyafa: As discussed offline I have added a developer comment

PR #14205, opened 2019-03-08
Add attached functions to object type in bir 

manuranga:


stdlib/bir/src/main/ballerina/bir/type_parser.bal
```diff
@@ -120,7 +120,28 @@ public type TypeParser object {
...
+    function parseObjectAttachedFunctions() returns BAttachedFunction[] {
+        int size = self.reader.readInt32();
+        int c = 0;
+        BAttachedFunction[] attachedFunctions = [];
+        while c < size {
+            var funcName = self.reader.readStringCpRef();
+            var visibility = parseVisibility(self.reader);
+
+            var typeTag = self.reader.readInt8();
+            if(typeTag != 16 ){
```
SupunS: shall we use constant `TYPE_TAG_INVOKABLE`?

PR #14220, opened 2019-03-08
Allow returning errors from object __init() 

### LINKED ISSUE #13342: [Spec Deviation] Object __init() methods return should be a subtype of error|()
**Description:**
As per the spec (v0.990:2019-01-16), The return type of the __init method must be a subtype of the union of error and nil.
```ballerina
type InitMethodInObject object {
    public function __init() returns error? {

    }
};
```

Tasks
- [ ] validate for Returning from __init() body with a ‘nil’ while there are potentially uninitialized fields should result in a compile error. This prevents using “check” expression inside the object initialization function."

**Affected Versions:**
0.990.2

pubudu91:
```ballerina
Person|error p = new Person();
----------------------------------------------
// new Person() is desugared to
{
    Person|error $result$; // The expr of the expression statement
    Person $obj$ = new Person();
    error? $temp$ = $obj$.__init();

    if ($temp$ is error) {
        $result$ = $temp$;
    } else {
        $result$ = $obj$;
    }
}
```
> With this change, only the `NEWSTRUCT` instruction is emitted for the `BLangTypeInit`. The invocation of the `__init()` is handled in the normal flow.
> For `__init()` functions which do not return errors, the type test mentioned above is not done. It's desugared to the following:
```ballerina
{
    Person $obj$ = new Person(); // $obj$ is the expr of the expression stmt in this case
    $obj$.__init();
}
```

## Samples
```ballerina
import ballerina/io;

public function main() {
    Person|error p = new Person();

    if (p is Person) {
        io:println("Person created");
    } else {
        io:println("Failed to create a Person: ", p);
    }
}

type Person object {
    string name = "";

    function __init() returns error? {
        self.name = check test();
    }

    function foo() returns string {
        return "foo";
    }
};

function test() returns string|error {
    error e = error("test");
    return e;
}
```

### DIFF
--- a/bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
+++ b/bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
@@ -4892,7 +4892,6 @@ public class BVM {
             case TypeTags.BYTE_TAG:
             case TypeTags.NULL_TAG:
             case TypeTags.XML_TAG:
-            case TypeTags.SERVICE_TAG:
                 if (sourceType.getTag() == TypeTags.FINITE_TYPE_TAG) {
                     return ((BFiniteType) sourceType).valueSpace.stream()
                             .allMatch(bValue -> checkIsType(bValue, targetType));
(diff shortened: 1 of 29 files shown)


hasithaa:
@pubudu91 There is a conflict, can you fix it.

PR #14243, opened 2019-03-09
Replace type assertion BBE with type cast BBE

MaryamZi:
## Purpose
Replace type assertion BBE with type cast BBE.

Related PR: https://github.com/ballerina-platform/ballerina-lang/pull/13479


hasithaa:
The Sample looks ok, @praneesha could you please check for the content?

examples/type-cast/type_cast.bal
```diff
@@ -0,0 +1,56 @@
...
+    io:println("Integer Value: ", i);
+
+    // Use the type cast expression with `value` to cast it to `float`.
+    // A numeric conversion would happen from `int` to `float`.
+    float f = <float> value;
+    io:println("Converted Float Value: ", f);
+
+    // Casting to a union type would also work similarly.
+    // If `value` belongs to the union type, the resultant value would
+    // be `value` itself. Else, if applicable a numeric conversion will
```
praneesha:
```suggestion
    // be `value` itself. Else, if applicable, a numeric conversion will
```

examples/type-cast/type_cast.bal
```diff
@@ -0,0 +1,56 @@
...
+    // be attempted.
+    // A numeric conversion would happen from `int` to `float` here.
+    float|boolean u = <float|boolean> value;
+    io:println("Converted Float Value: ", u);
+
+    // Assign `employee` to `value`.
+    value = employee;
+    // Casting a value to an incorrect type (the value does not belong
+    // to the type and numeric conversion is not possible) would result
+    // in abrupt completion with a panic.
```
praneesha:
```suggestion
    // in an abrupt completion with a panic.
```

examples/type-cast/type_cast.description
```diff
@@ -0,0 +1,6 @@
+// Type cast attempts casting a value to a target type, performing a numeric conversion if required.
```
praneesha:
```suggestion
// Type cast casts a value to a target type performing a numeric conversion if required.
```

examples/type-cast/type_cast.description
```diff
@@ -0,0 +1,6 @@
...
+// If the value belongs to the target type, the result of the type cast expression is the value itself.
+// Else, if the target type contains exactly one basic numeric type, and the value belongs to a numeric type,
+// the resultant value would be the numeric conversion of the value to the target numeric type.
+// If neither is satisfied or if numeric conversion returns an error, the type cast expression completes abruptly with
```
praneesha:
```suggestion
// If neither is satisfied or if the numeric conversion returns an error, the type cast expression completes abruptly with
```

examples/type-cast/type_cast.out
```diff
@@ -0,0 +1,12 @@
+# To run this sample, navigate to the directory that contains the
+# `.bal` file and issue the `ballerina run` command.
```
praneesha:
```suggestion
# `.bal` file and execute the `ballerina run` command.
```

examples/type-cast/type_cast.bal
```diff
@@ -0,0 +1,56 @@
...
+    int age;
+    int empNo;
+};
+
+type Department record {
+    string code;
+};
+
+public function main() {
+    // Define an `Employee` and cast it to a `Person` typed variable.
```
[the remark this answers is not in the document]
MaryamZi: Will change to 
```suggestion
    // Define an `Employee` value and cast it to a `Person`.
```

examples/type-cast/type_cast.bal
```diff
@@ -0,0 +1,56 @@
...
+    io:println("Person Name: ", person.name);
+    io:println("Person Age: ", person.age);
+
+    // Cast the value held by the `person` variable to an `Employee`.
+    // This should succeed since `person` holds an `Employee` value.
+    Employee employeeTwo = <Employee> person;
+    io:println("Employee Name: ", employeeTwo.name);
+    io:println("Employee Age: ", employeeTwo.age);
+
+    // Cast an `int` value held by an `anydata` typed variable as an
```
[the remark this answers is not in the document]
MaryamZi: Not sure the change will reflect what is done here. 

The `anydata` typed variable holds an `int` value. We are casting that to an `int`.

examples/type-cast/type_cast.description
MaryamZi: Don't we need "attempts" since the cast could fail as explained later on.

examples/type-cast/type_cast.out
MaryamZi: "issue" is what we've used in all the BBEs. If "execute" is more accurate shall we fix all the BBEs separately?

examples/type-cast/type_cast.description
hasithaa: This is the exact text from the spec.

"A type-cast-expr casts the value resulting from evaluating expression to the type described by the type-descriptor, performing a numeric conversion if required."  

Then in the next paragraph, spec explains the failure behavior. So having just cast is enough in this case.  WDYT?

examples/type-cast/type_cast.out
hasithaa: It seems like people are using both "issue" and "execute" words for the same purpose. It seems like, The word "execute" is used more frequently on the web in the context of computing.

examples/type-cast/type_cast.description
MaryamZi: Got it, will change to say "casts".

examples/type-cast/type_cast.out
MaryamZi: Yeah, "execute" or even "use" seems natural with the usage of a command compared to "issue". So I'll update this here, and send a separate PR to update the rest of the BBEs.

examples/type-cast/tests/type_cast_test.bal
```diff
@@ -0,0 +1,44 @@
+import ballerina/test;
+import ballerina/io;
```
Anoukh: Would it be better to reorder the imports alphabetically?

MaryamZi: Fixed!

PR #14260, opened 2019-03-11
Support panic statement in BIR and JVM codegen

warunalakshitha:
## Purpose
Support panic statement in bir and initial implementation of jvm codegen.

ex. 
```
type AccountNotFoundError error<string, string>;
function getAccountBalance(int accountID) returns int {
      
        string errorReason = "Account Not Found";
        AccountNotFoundError accountNotFoundError =
                                            error(errorReason,"3");
        if(accountID == 0){
            panic accountNotFoundError;
        }
        return accountID;
    
}

    var result = getAccountBalance(0);
	return result;
}

```

this will result in 

Exception in thread "main" Account Not Found 3
	at hello.getAccountBalance(Unknown Source)
	at hello.main(Unknown Source)
	at hello.main(Unknown Source)

## Todos
We use java stacktrace directly when printing. This required to be formatted to comply with bal stacktrace.


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ErrorValue.java
```diff
@@ -28,13 +28,15 @@
...
-public class ErrorValue implements RefValue {
+public class ErrorValue extends RuntimeException implements RefValue {
 
-    private BType type;
-    private String reason;
-    private RefValue details;
+    private static final long serialVersionUID = 1L;
+    private final BType type;
+    private final String reason;
+    private final Object details;
```
vinok88: Details are modeled as bal records, right? Then shouldn't this be of MapValur type?

warunalakshitha: Detail is actually ref value. So it can be any type.

warunalakshitha:
Will send new PR with trap changes

PR #14265, opened 2019-03-11
Tuple type support in JVM target

vinok88:


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLAttributes.java
```diff
@@ -60,6 +60,9 @@ public boolean equals(Object obj) {
         if (obj == null || !(obj instanceof XMLAttributes)) {
             return false;
         }
+        ArrayValue ara = new ArrayValue(BTypes.typeMap, 5);
+        ara.add(0,new MapValue<>());
+        MapValue val = (MapValue) ara.getRefValue(0);
```
SupunS: Do we need this code snippet?

vinok88: Oops, yes that's by mistake. Removed now.

PR #14277, opened 2019-03-11
[WIP] Add is like and is checks

lankavitharana:


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/model/BIRNonTerminator.java
```diff
@@ -262,7 +262,7 @@ public void accept(BIRVisitor visitor) {
     }
 
     /**
-     * A type cast expression.
+     * A type cast expression
```
Kishanthan: Add .dot

PR #14278, opened 2019-03-11
 Fix unit test cache updating issue

Anoukh:
## Purpose
Unit tests will always rerun due to balo tests updating .zip files in the test resource src which also gets reflected in the build resource directory. This fix will delete the respective newly created files after each balo test.


hasithaa:
@pahans Could you have a look into this? I verified the fix.

Issue #13946, opened 2019-02-27
XML Expression Improvements for 0.990.0 Spec


hasithaa:
* [ ] #13718 Update XML interpolation syntax. use `${expr}` instead of `{{expr}}`
* [ ]  Interpolated expressions occurring at ‘content’ can be xml or any simple type. Currently, it only supports xml and string.
* [ ]  Interpolated expressions should not be allowed in the value of a namespace attribute.

Issue #10935, opened 2018-10-22
XML attributes are considered namespace declarations, with default namespaces without prefixes


MaryamZi:
**Description:**
When a default namespace is declared (without a prefix), un-prefixed XML attributes of subsequent elements are considered namespace declarations.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
  xmlns "http://ballerina.com/";
 
  xml x1 = xml `<book year="1940">The Lost World</book>`;
  io:println(x1);
}
```

produces output (note **xmlns:year="1940"**):
```cmd
<book xmlns="http://ballerina.com/" xmlns:year="1940">The Lost World</book>
```

**Affected Versions:**
0.982.0

MaryamZi:
Happens at https://github.com/ballerina-platform/ballerina-lang/blob/v0.990.0/bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BXMLItem.java#L269. 

Need to revisit and validate expected behaviour.

Issue #13718, opened 2019-02-20
Update XML expression grammar


pranavan15:
**Description:**
XML expression needs to be checked and updated to comply with the Ballerina specification.

**Suggested Labels (optional):**
Type/Task, Component/Compiler

pranavan15:
This issue is fixed by the above-referenced PR. Hence, closing the issue.

PR #14281, opened 2019-03-12
XML interpolation expression improvement

rdhananjaya:
## Purpose
Update XML interpolation syntax from {{expr}} to ${expr}
```ballerina
//xml x = xml `<Person name={{name}}>{{content}}</Person>`;
xml x = xml `<Person name=${name}>${content}</Person>`;
```

Remove the ability to create XML elements types and attributes by interpolating tag name
```ballerina
\\ We can no longer interpolate XML tags
xml x = xml `<${elemName}>content</${elemName}>`;
xml y = xml `<elem ${attName}="value" />`;
```

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13946 and https://github.com/ballerina-platform/ballerina-lang/issues/13718
Also fixes https://github.com/ballerina-platform/ballerina-lang/issues/10935


bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BXMLItem.java
```diff
@@ -263,15 +263,6 @@ public void setAttribute(String localName, String namespaceUri, String prefix, S
             return;
         }
 
-        // Attributes cannot cannot be belong to default namespace. Hence, if the current namespace is the default one,
```
rdhananjaya: We no longer allow default namespace URI of the element to be used as 'xmlns'. 
When user actually use 'xmlns' URI it's handled in prefix handling code at the end of this method.

tests/ballerina-unit-test/src/test/resources/test-src/types/xml/xml-literals-negative.bal
```diff
@@ -22,15 +22,15 @@ function defineEmptyNamespaceInline() {
 
 function testExpressionAsElementName() {
     int v1 = 11;
-    // xml x1 = xml `<ns0:{{v1}}>hello</ns0:{{v1}}>`;
+    // xml x1 = xml `<ns0:${v1}>hello</ns0:${v1}>`;
```
MaryamZi: Isn't this also invalid now? Shall we remove this commented line and also the TODO in the `XMLLiteralTest` if it is no longer valid?

PR #14324, opened 2019-03-13
Upgrade netty version in Ballerina

Bhashinee:
## Purpose
$subject.


tests/ballerina-integration-test/src/test/resources/grpc/clients/grpc_secured_unary_client.bal
```diff
@@ -31,7 +31,7 @@ function testUnarySecuredBlocking() returns (string) {
                 name: "TLSv1.2",
                 versions: ["TLSv1.2","TLSv1.1"]
             },
-            ciphers:["TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA"],
+            ciphers:["TLS_RSA_WITH_AES_128_CBC_SHA"],
```
Manuri: Just for my understanding, why we had to switch from Diffie Hellman to RSA?

Bhashinee: Upgraded openssl version does not support that cipher.

riyafa:
I think we need to update the netty version in gradle also

PR #14330, opened 2019-03-13
Add support for package level variables

Kishanthan:
With this PR, now we can define package level variables and use them with jvm target as below.

```ballerina
int a = 33;
  
        a = 44;
        int b = a;
        return b;
}
```


compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_instruction_gen.bal
```diff
@@ -420,83 +326,119 @@ type InstructionGenerator object {
...
+
+        if (varDcl.kind == "GLOBAL") {
+            string varName = varDcl.name.value;
+            string className = lookupFullQualifiedClassName(self.currentPackageName + varName);
+
+            if (bType is bir:BTypeInt) {
+                self.mv.visitFieldInsn(PUTSTATIC, className, varName, "J");
+            } else if (bType is bir:BMapType) {
+                self.mv.visitFieldInsn(PUTSTATIC, className, varName, io:sprintf("L%s;", MAP_VALUE));
+            } else {
```
vinok88: Need to add global var support for other types as well.

Issue #13947, opened 2019-02-27
 x@ should return a map. Currently, it has to be converted to a map.


hasithaa:

rdhananjaya:
From spec:
```
xml-attributes-expr := expression @

Returns the attributes map of a singleton xml value, or nil if the operand is not a singleton xml value. 
The result type is map<string>?
```

PR #14350, opened 2019-03-13
XML attribute access expression xmlVal@ returns a map<string>?

rdhananjaya:
## Purpose
This PR fixes https://github.com/ballerina-platform/ballerina-lang/issues/13947

```ballerina
var x1 = xml `<child xmlns:p1="http://wso2.com/" xmlns:p2="http://sample.com/wso2/a1/" p1:foo="bar"/>`;
map<string>? s = x1@;
```
`s` is a map containing namespace p1, p2 and attribute `p1:foo`.

```ballerina
function nonSingletonXmlAttributeAccess() returns boolean {
    xml x = xml `<someEle>cont</someEle>`;
    xml y = xml `<elem>More-Stuff</elem>`;
    xml nonSingleton = x + y;
    if (nonSingleton@ is ()) {
        return true;
    }
    return false;
}
```
`nonSingletonXmlAttributeAccess` returns true as nonSingleton is a xml sequence.


bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BXMLSequence.java
```diff
@@ -167,7 +166,7 @@ public void setAttribute(String localName, String namespace, String prefix, Stri
             return ((BXMLItem) sequence.getRefValue(0)).getAttributesMap();
         }
 
-        return new BMap<>(new BMapType(BTypes.typeString));
+        return null;
```
rdhananjaya: Only XML items are allowed to return a map when ask for attribute map, XML sequence should return a nil value.

rdhananjaya:
> If an update is done on top of the resulting map, should that reflect in the original xml as well?

Updated the PR to propagate attribute map updates back to BXMLItem.

bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BXMLItem.java
```diff
@@ -890,4 +891,42 @@ public synchronized void attemptFreeze(BVM.FreezeStatus freezeStatus) {
...
+
+        void finishConstruction() {
+            constructed = true;
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public void put(Object key, BValue value) {
+            super.put(key, value);
+            if (constructed) {
```
pubudu91: What's the purpose of this check?

rdhananjaya: This maps put method work under 2 states, 
1. While being populated with attribute values, using this put method.
2. Assigning new attributes to the XMLItem via this attribute map instance. (constructed is true)

bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BXMLItem.java
```diff
@@ -890,4 +891,42 @@ public synchronized void attemptFreeze(BVM.FreezeStatus freezeStatus) {
...
+                setAttribute((String) key, value.stringValue());
+            }
+        }
+
+        private void setAttribute(String key, String value) {
+            String url = null;
```
pubudu91: Isn't it incorrect to assume namespaces are URLs all the time?

bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BXMLItem.java
```diff
@@ -890,4 +891,42 @@ public synchronized void attemptFreeze(BVM.FreezeStatus freezeStatus) {
...
+            String localName = key;
+
+            int endOfUrl = key.indexOf('}');
```
pubudu91: What happens if the namespace contains braces? For example,
```ballerina
xmlns "foo}}bar" as foobar;
```
Then the `key` will be `{foo}}bar}age` and `url` will be `foo`.

bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BXMLItem.java
rdhananjaya: [https://www.w3.org/TR/xml-names/](https://www.w3.org/TR/xml-names/)
Namespace is a URI reference, it doesn't have to be of http://... format

PR #14359, opened 2019-03-13
Allow defining array vars of types without implicit init vals

### LINKED ISSUE #14296: Record fields with non-defaultable type array are not allowed
```ballerina
type Foo record {
    Bar[] bar;
}
```
above is not allowed by the compiler, if `Bar` does not have an implicit initial value.

pubudu91:


SupunS:
Shall we add some positive test case? e.g: T[] is used as varDef, parameters, record/object fields, where T doesn't have an implicit initial value

Issue #14304, opened 2019-03-12
Update Time module APIs with new language changes.


anupama-pathirage:
**Description:**
Review and update the current time module API
- It still throws errors instead of returning
- It using string types as input parameters, Try to convert them to const types where applicable

PR #14362, opened 2019-03-13
Refactor the Time API

anupama-pathirage:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/14304

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no


stdlib/time/src/main/java/org/ballerinalang/stdlib/time/nativeimpl/Parse.java
```diff
@@ -46,7 +48,10 @@
         functionName = "parse",
         args = {@Argument(name = "timestamp", type = TypeKind.STRING),
                 @Argument(name = "format", type = TypeKind.UNION)},
-        returnType = {@ReturnType(type = TypeKind.RECORD, structType = "Time", structPackage = "ballerina/time")},
+        returnType = {
+                @ReturnType(type = TypeKind.RECORD, structType = "Time", structPackage = "ballerina/time"),
+                @ReturnType(type = TypeKind.RECORD, structType = "error", structPackage = BALLERINA_BUILTIN_PKG)
```
warunalakshitha: Is Type Kind should be  TypeKind.ERROR? Same with other native functions as well.

stdlib/http/src/main/ballerina/http/caching/http_caching_client.bal
```diff
@@ -556,11 +556,14 @@ function getFreshnessLifetime(Response cachedResponse, boolean isSharedCache) re
             if (cachedResponse.hasHeader(DATE)) {
                 string[] dateHeader = cachedResponse.getHeaders(DATE);
 
+                var tExpiresHeader = time:parse(expiresHeader[0], time:TIME_FORMAT_RFC_1123);
+                var tDateHeader = time:parse(dateHeader[0], time:TIME_FORMAT_RFC_1123);
```
pubudu91: Shall we bring these two inside the if condition below?

stdlib/http/src/main/ballerina/http/caching/http_caching_client.bal
```diff
@@ -556,11 +556,14 @@ function getFreshnessLifetime(Response cachedResponse, boolean isSharedCache) re
...
+
                 if (dateHeader.length() == 1) {
-                    // TODO: See if time parsing errors need to be handled
-                    int freshnessLifetime = (time:parse(expiresHeader[0], time:TIME_FORMAT_RFC_1123).time
-                            - time:parse(dateHeader[0], time:TIME_FORMAT_RFC_1123).time) / 1000;
-                    return freshnessLifetime;
+                    if (tExpiresHeader is time:Time && tDateHeader is time:Time) {
+                        int freshnessLifetime = (tExpiresHeader.time - tDateHeader.time) /1000;
+                        return freshnessLifetime;
+                    }
```
pubudu91: Shall we add an else clause here as well? The else clause should simply be `return STALE;`

stdlib/http/src/main/ballerina/http/caching/http_caching_client.bal
```diff
@@ -750,14 +753,21 @@ function getDateValue(Response inboundResponse) returns int {
...
 
     string dateHeader = inboundResponse.getHeader(DATE);
     // TODO: May need to handle invalid date headers
-    time:Time dateHeaderTime = time:parse(dateHeader, time:TIME_FORMAT_RFC_1123);
-    return dateHeaderTime.time;
+    var dateHeaderTime = time:parse(dateHeader, time:TIME_FORMAT_RFC_1123);
+    if (dateHeaderTime is time:Time) {
+        return dateHeaderTime.time;
+    } else {
+        return 0;
```
Manuri: What is the intended meaning of returning 0 here?

stdlib/http/src/main/ballerina/http/http_response.bal
```diff
@@ -316,8 +316,10 @@ public function Response.setETag(json|xml|string|byte[] payload) {
 
 public function Response.setLastModified() {
     time:Time currentT = time:currentTime();
-    string lastModified = time:format(currentT, time:TIME_FORMAT_RFC_1123);
-    self.setHeader(LAST_MODIFIED, lastModified);
+    var lastModified = time:format(currentT, time:TIME_FORMAT_RFC_1123);
+    if (lastModified is string) {
+        self.setHeader(LAST_MODIFIED, lastModified);
+    }
```
Manuri: How the error scenario should be handled in this case?

examples/index.json
```diff
@@ -447,7 +447,7 @@
         "category": "Language concepts",
         "samples": [{
                 "name": "Date Time",
```
MaryamZi: Should we also update the name?

examples/time/time.bal
```diff
@@ -2,28 +2,36 @@ import ballerina/io;
...
     io:println("Current system time in milliseconds: " + currentTimeMills);
     // Specify a time with the required year, month, date,
     // time, and timezone information.
-    time:Time timeCreated = time:createTime(2017, 3, 28, 23, 42, 45, 554,
-                                            "America/Panama");
-    io:println("Created Time: " + time:toString(timeCreated));
+    time:Time|error timeCreated = time:createTime(2017, 3, 28, 23, 42, 45,
+        554, "America/Panama");
+    if (timeCreated is time:Time) {
+        io:println("Created Time: " + time:toString(timeCreated));
```
MaryamZi:
```suggestion
        io:println("Created Time: ", time:toString(timeCreated));
```
Applicable elsewhere too.

stdlib/http/src/main/ballerina/http/http_response.bal
anupama-pathirage: Practically it is unlikely an error occurred here since the time is a current time and format is a constant.  Theoretically if some error occurred last modified time header won't get set in the response. But since this error cannot be happened no need of handling it explicitly

stdlib/http/src/main/ballerina/http/caching/http_caching_client.bal
anupama-pathirage: IMO we don't need to have else clause here as all these else clauses will be handled by the return statement at the end of the function. 
https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/http/src/main/ballerina/http/caching/http_caching_client.bal#L571

pubudu91: Ah yes. Sorry my bad! Totally missed that final return.

stdlib/http/src/main/ballerina/http/caching/http_caching_client.bal
pubudu91: Here we want to parse the date header and get the time in milliseconds. If the date header is invalid, we simply return 0.

stdlib/http/src/main/ballerina/http/http_response.bal
pubudu91: Yes, it's unlikely that this would result in an error. IIRC, the spec doesn't say anything about handling a case where the current time cannot be parsed.

Manuri: I wasn't talking about the exact context here.
I'd say, if we an error scenario is possible, but we most certainly know it doesn't happen depending on the context, it would be best if we have an assertion for the error case due to the following reasons.
1. Someone may later modify return values of the relevant function and won't handle all cases. 
2. If the implementation changes someday in a way that an error probability is higher for the same scenario, that will not go unnoticed easily.

I'm suggesting this considering how we treat default case in java switch in a similar scenario. 
I know the above can be avoided if whoever changes it does that the proper way.

anupama-pathirage: @Manuri Agreed to your point here. Added a panic for error case (even  though it is unlikely to happen)

stdlib/time/src/main/ballerina/time/timelib.bal
```diff
@@ -48,7 +56,7 @@ public extern function toString(Time time) returns string;
 # + time - The Time record to be formatted
 # + timeFormat - The format which is used to format the time represented by this object
 # + return - The formatted string of the given time
-public extern function format(Time time, string|TimeFormat timeFormat) returns string;
+public extern function format(Time time, string|TimeFormat timeFormat) returns string|error;
```
wggihan: Don't we need to specify in the return statement that this will return an error?

PR #14364, opened 2019-03-13
Add UDP client support

wggihan:
## Samples
```
import ballerina/io;
import ballerina/socket;

public function main() {
    // Create a new socket client.
    // Optionally, you can provide port that this socket need to bind or
    // both interface and port as follows.
    // socket:UdpClient client = new({ localAddress: { port: 48828 } });
    // socket:UdpClient client = new({ localAddress: { host: "localhost", port: 48828 } });
    socket:UdpClient socketClient = new({});
    string msg = "Hello from UDP client";
    byte[] c1 = msg.toByteArray("utf-8");
    // Send data to remote host.
    // Second parameter is the address of the remote host.
    var sendResult =
        socketClient->sendTo(c1, { host: "localhost", port: 48826 });
    if (sendResult is int) {
        io:println("Number of bytes written: ", sendResult);
    } else {
        panic sendResult;
    }
    // Wait until data receive from remote host.
    // This will block until receive at least a single byte.
    // Optionally, you can specify the length as below.
    // socketClient->receiveFrom(length = 30)
    // This will block until specified length of bytes receive from host.
    var result = socketClient->receiveFrom();
    if (result is (byte[], int, socket:Address)) {
        var (content, length, address) = result;
        io:ReadableByteChannel byteChannel =
            io:createReadableChannel(content);
        io:ReadableCharacterChannel characterChannel =
            new io:ReadableCharacterChannel(byteChannel, "UTF-8");
        var str = characterChannel.read(60);
        if (str is string) {
            io:println("Received: ", untaint str);
        } else {
            io:println(str.detail().message);
        }
    } else {
        io:println("An error occured while receiving the data ",
            result);
    }
    // Close the client and release the bound port.
    var closeResult = socketClient->close();
    if (closeResult is error) {
        io:println("An error occured while closing the connection ",
            closeResult);
    }
}
```

### DIFF
--- a/examples/udp-socket-client/udp_socket_client.bal
+++ b/examples/udp-socket-client/udp_socket_client.bal
@@ -6,9 +6,9 @@ public function main() {
     // Create a new socket client.
     // Optionally, you can provide port that this socket need to bind or
     // both interface and port as follows.
-    // socket:UdpClient client = new({ localAddress: { port: 48828 } });
-    // socket:UdpClient client = new({ localAddress: { host: "localhost", port: 48828 } });
-    socket:UdpClient socketClient = new({});
+    // socket:UdpClient client = new(localAddress = { port: 48828 });
+    // socket:UdpClient client = new(localAddress = { host: "localhost", port: 48828 });
+    socket:UdpClient socketClient = new;
     string msg = "Hello from UDP client";
     byte[] c1 = msg.toByteArray("utf-8");
     // Send data to remote host.
@@ -36,7 +36,7 @@ public function main() {
         if (str is string) {
             io:println("Received: ", untaint str);
         } else {
-            io:println(str.detail().message);
+            io:println("Error: ", str.detail().message);
         }
     } else {
         io:println("An error occured while receiving the data ",
--- a/examples/udp-socket-client/udp_socket_client.bal
+++ b/examples/udp-socket-client/udp_socket_client.bal
@@ -1,0 +1,51 @@
+// This is the client implementation for the UDP socket.
+import ballerina/io;
+import ballerina/socket;
+
+public function main() {
+    // Create a new socket client.
+    // Optionally, you can provide port that this socket need to bind or
+    // both interface and port as follows.
+    // socket:UdpClient client = new({ localAddress: { port: 48828 } });
+    // socket:UdpClient client = new({ localAddress: { host: "localhost", port: 48828 } });
+    socket:UdpClient socketClient = new({});
+    string msg = "Hello from UDP client";
+    byte[] c1 = msg.toByteArray("utf-8");
+    // Send data to remote host.
+    // Second parameter is the address of the remote host.
+    var sendResult =
+        socketClient->sendTo(c1, { host: "localhost", port: 48826 });
+    if (sendResult is int) {
+        io:println("Number of bytes written: ", sendResult);
+    } else {
+        panic sendResult;
+    }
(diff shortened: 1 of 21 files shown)


stdlib/socket/src/main/ballerina/socket/udp_client_socket.bal
```diff
@@ -0,0 +1,80 @@
...
+# + localAddress - the local IP address string in textual presentation to which the socket is bound
+# + id - a unique identifier to identify each client
+public type UdpClient client object {
+
+    private UdpClientConfig config;
+    public int localPort = 0;
+    public string? localAddress = ();
+    public int id = 0;
+
+    public function __init(UdpClientConfig? clientConfig) {
```
chamil321: Shall we document this method?

examples/udp-socket-client/udp_socket_client.bal
```diff
@@ -0,0 +1,51 @@
...
+        var (content, length, address) = result;
+        io:ReadableByteChannel byteChannel =
+            io:createReadableChannel(content);
+        io:ReadableCharacterChannel characterChannel =
+            new io:ReadableCharacterChannel(byteChannel, "UTF-8");
+        var str = characterChannel.read(60);
+        if (str is string) {
+            io:println("Received: ", untaint str);
+        } else {
+            io:println(str.detail().message);
```
MaryamZi: Shall we say something to indicate that an error occurred, in addition to printing something from the detail?
```suggestion
            io:println("Error: ", str.detail().message);
```

stdlib/socket/src/main/ballerina/socket/udp_client_socket.bal
MaryamZi: Should this be a deaultable param? 
```suggestion
    public function __init(UdpClientConfig? clientConfig = ()) {
```

That way the user would not have to give an empty record literal like in https://github.com/ballerina-platform/ballerina-lang/pull/14364/files#diff-925cdfc5930b3e9edcc7a9e998fda05bR11 when initiating the client, if he doesn't want to specify anything additional 
```ballerina
    socket:UdpClient socketClient = new();
```

stdlib/socket/src/main/ballerina/socket/udp_client_socket.bal
```diff
@@ -0,0 +1,80 @@
...
+        if (clientConfig is UdpClientConfig) {
+            self.config = clientConfig;
+            var initResult = self.initEndpoint(clientConfig);
+            if (initResult is error) {
+                panic initResult;
+            }
+        }
+        return ();
```
MaryamZi: This `return ()` is not needed right? It should work without.

stdlib/socket/src/main/ballerina/socket/udp_client_socket.bal
```diff
@@ -0,0 +1,80 @@
...
+    string host?;
+    int port;
+    !...;
+};
+
+# Configuration for UDP socket client endpoint.
+#
+# + localAddress - The address information this UDP socket bind locally
+public type UdpClientConfig record {
+    Address? localAddress = ();
```
MaryamZi: Given that `localAddress`is the only field in the record and since we only use this record in `__init()` IINM, do we need this record at all?

Should we maybe update `__init()` to accept `Address?` and remove `UdpClientConfig`? 

```ballerina
     public function __init(Address? localAddress = ()) {
```

PR #14388, opened 2019-03-14
Add test listener for HTTP module tests

pubudu91:


stdlib/http/src/test/resources/testng.xml
```diff
@@ -21,6 +21,10 @@ under the License.
 
 <suite name="ballerina-test-suite">
 
+    <!--<listeners>-->
```
chamil321: Shall we add a comment explaining the process when debugging?

pubudu91: Added with 6f65404

PR #14407, opened 2019-03-14
Fix language server gradle build issues

nadeeshaan:


language-server/modules/langserver-core/build.gradle
```diff
@@ -56,6 +84,11 @@ CopySpec copyResourceSpec = copySpec {
     into("resources")
 }
 
+task generateIndex(type: JavaExec) {
+    classpath = sourceSets.main.runtimeClasspath
```
pahans: define inputs and outputs of this task and enable caching for this task.

PR #14410, opened 2019-03-15
[WIP] Updates to anydata, open records and error detail

### LINKED ISSUE #13181: [Spec Deviation] Default record rest field type should be anydata|error (pure)
**Description:**
As per the spec, "if the record-rest-descriptor is empty, then the value may contain extra fields
belonging to any pure type".

Currently this is limited to `anydata` only, and errors cannot be added.

i.e., the current default rest field type is `anydata`, and this needs to be changed to `pure` types.

### LINKED ISSUE #13231: [Spec Deviation] (anydata|error)[] and map<anydata|error> need to be considered as anydata
**Description:**
As per the spec (v0.990:2019-01-16 - pages 30, 31), 
"anydata-type-descriptor := anydata

The type descriptor anydata describes the type of all pure values other than errors. The type anydata contains a shape if and only if the shape is pure and is not the shape of an error value. 

Thus the type anydata|error is the supertype of all pure types. The type anydata is equivalent to the union

() | boolean | int | float | decimal | string | **(anydata|error)[]** | **map<anydata|error>** | xml | table"

But, currently `(anydata|error)[]` and `map<anydata|error>` are not considered `anydata`.

**Affected Versions:**
0.990.2

MaryamZi:
## Purpose
This PR
- Allows structures with pure types (anydata + error) as anydata 
Fixes #13231 - (anydata|error)[] and map<anydata|error> need to be considered as anydata

- Updates the default record rest field to be anydata|error (pure type)
Fixes #13181 - Default record rest field type should be anydata|error (pure)

- Validates that the detail expression specified in the error constructor is a subtype of map<anydata|error>


MaryamZi:
Closing this PR since changes need to be introduced to the error type, type signature and info, to support detail types containing the same errors, Will open a new PR once the changes are done.

PR #14451, opened 2019-03-17
Improve scripts

Manuri:
## Purpose
$Subject


Manuri:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/14535

Issue #13495, opened 2019-02-06
AESCipherToolNegativeTest.testEncryptionAndDecryptionWithTwoCipherTools fails intermittently


vinok88:
**Description:**
Error:
"[ERROR] testEncryptionAndDecryptionWithTwoCipherTools(org.ballerinalang.stdlib.config.AESCipherToolNegativeTest)  Time elapsed: 0.503 s  <<< FAILURE!
java.lang.AssertionError: Should throw an error expected [true] but found [false]
	at org.ballerinalang.stdlib.config.AESCipherToolNegativeTest.testEncryptionAndDecryptionWithTwoCipherTools(AESCipherToolNegativeTest.java:43)"

vinok88:
There seems to be no issue with this test case. I could reproduce the intermittent behavior by running the test within a loop (100). Look like there is a bug in the cipher tool.

ldclakmal:
The reason for the issue is that we expect the test case should throw an exception since the decryption key is invalid. But in some cases, the given decryption key is able to decrypt the given encrypted string with the use of padding. But the output (decrypted text) does not match with the plain text. Therefore in the test case, no exception is thrown and assertion gets failed.

vinok88:
@ldclakmal, you mean decryption key matches intermittently? Then that's a huge security issue. Correct me if I'm wrong.

ldclakmal:
@vinok88 Decryption key does not match with encryption key intermittently. What happens is that the decryption key may end up in a key that the encoded byte array can be decrypted properly. But the result never matches with the original plain text.
In the test cases, it expected an exception each and every time. But in some cases, the exception does not occur. That is the reason why the test was failed intermittently. So, if the exception does not occur, the decrypted plain text should not be matched with the original plain text.

PR #14492, opened 2019-03-18
Fix the test case for decryption bug

ldclakmal:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13495 and moved the test case into `AESCipherToolTest.java` file as a negative test case.


stdlib/config-api/src/test/java/org/ballerinalang/stdlib/config/AESCipherToolTest.java
```diff
@@ -60,4 +60,23 @@ public void testEncryptionAndDecryptionWithLongSecret() throws AESCipherToolExce
...
+        try {
+            AESCipherTool encryptionCipherTool = new AESCipherTool(encryptionKey);
+            String encryptedStr = encryptionCipherTool.encrypt(plainText);
+            AESCipherTool decryptionCipherTool = new AESCipherTool(decryptionKey);
+            String decryptedText = decryptionCipherTool.decrypt(encryptedStr);
+            // We expect the above line should throw an exception since the decryption key is invalid. But in some cases
+            // the given decryption key is able to decrypt the given encrypted string with the use of padding. Then, the
+            // output (decrypted text) does not match with the plain text.
+            Assert.assertNotEquals(decryptedText, plainText);
+        } catch (AESCipherToolException e) {
```
Manuri: Can't we make this an expected exception?

ldclakmal: Since it does not throw an exception always, we can't do that. The reason for the intermittent failure was also not throwing exception always.

PR #14500, opened 2019-03-18
[stdlib] [cache] Panic Cache initilization when the Scheduler causes an error

ThisaruGuruge:

### DIFF
--- a/stdlib/cache/src/main/ballerina/cache/cache.bal
+++ b/stdlib/cache/src/main/ballerina/cache/cache.bal
@@ -80,7 +80,7 @@ public type Cache object {
 
         var attachCacheCleanerResult = cacheCleanupTimer.attach(cacheCleanupService);
         if (attachCacheCleanerResult is error) {
-            error e = error("Failed to create cache cleanup scheduler.", { message : attachCacheCleanerResult.detail().message });
+            error e = error("Failed to create the cache cleanup task.", { message : attachCacheCleanerResult.detail().message });
             panic e;
         }
         var timerStartResult = cacheCleanupTimer.start();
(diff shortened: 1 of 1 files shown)


stdlib/cache/src/main/ballerina/cache/cache.bal
```diff
@@ -78,9 +78,16 @@ public type Cache object {
         };
         task:Scheduler cacheCleanupTimer = new(cacheCleanupTimerConfiguration);
 
-        _ = cacheCleanupTimer.attach(cacheCleanupService);
-        _ = cacheCleanupTimer.start();
-
+        var attachCacheCleanerResult = cacheCleanupTimer.attach(cacheCleanupService);
+        if (attachCacheCleanerResult is error) {
+            error e = error("Failed to create cache cleanup scheduler.", { message : attachCacheCleanerResult.detail().message });
```
Shan1024: This is actually not a `scheduler`, it is a `task`, right? (we just try to schedule the task here)

```suggestion
            error e = error("Failed to create the cache cleanup task.", { message : attachCacheCleanerResult.detail().message });
```

stdlib/cache/src/main/ballerina/cache/cache.bal
```diff
@@ -78,9 +78,16 @@ public type Cache object {
...
+            panic e;
+        }
+        var timerStartResult = cacheCleanupTimer.start();
+        if (timerStartResult is error) {
+            error e = error("Failed to start cache cleanup scheduler.", { message : timerStartResult.detail().message });
```
Shan1024:
```suggestion
            error e = error("Failed to start the cache cleanup task.", { message : timerStartResult.detail().message });
```

Issue #13441, opened 2019-02-01
[Spec Deviation] Abstract Listener Object's mismatching method signature


Anoukh:
**Description:**
As per the spec (v0.990:2019-01-16), AbstractListener objects __attach() signature is as follows
```ballerina
function __attach (service s, string? name = ()) returns error?;
```
Implemented function signature is as follows
```ballerina
public function __attach(service s, map<any> annotationData) returns error?
```

**Affected Versions:**
0.990.2

PR #14532, opened 2019-03-19
Change abstract listener's __attach signature

rdhananjaya:
## Purpose
Current signature of __attach function of abstract listener is as follows:
```ballerina
public function __attach(service s, map<any> annotationData) returns error?
```

In Ballerina spec it should be as follows:
```ballerina
function __attach (service s, string? name = ()) returns error?;
```

This PR fixes https://github.com/ballerina-platform/ballerina-lang/issues/13441


tests/ballerina-spec-conformance-tests/S05_builtin_abstract_types/tests/deviations/listener.bal
```diff
@@ -1,66 +0,0 @@
-// Copyright (c) 2019 WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
MaryamZi: Instead of removing this, shall we move it as a conforming test?

rdhananjaya: Will move to tests/ballerina-spec-conformance-tests/S05_builtin_abstract_types/tests/ with expected  __attach signature.

stdlib/http/src/main/ballerina/http/non_listening_service_endpoint.bal
```diff
@@ -43,7 +43,7 @@ public type MockListener object {
 
     public function init (ServiceEndpointConfiguration c);
     public extern function initEndpoint () returns (error?);
-    public extern function register (service s, map<any> annotationData) returns error?;
+    public extern function register (service s, string? name) returns error?;
```
hasithaa: @wggihan why is this a public function? Shall we make this module private.

stdlib/http/src/main/ballerina/http/service_endpoint.bal
```diff
@@ -61,8 +61,9 @@ public type Listener object {
     # Gets invoked when attaching a service to the endpoint.
     #
     # + s - The service that needs to be attached
+    # + name - Name of the service
     # + return - An `error` if there is any error occured during the service attachment process or else nil
-    extern function register(service s, map<any> annotationData) returns error?;
+    extern function register(service s, string? name = ()) returns error?;
```
hasithaa: In some places, we have used `string?` name, shall we unify it.

stdlib/task/src/main/ballerina/task/listener.bal
```diff
@@ -33,8 +33,9 @@ public type Listener object {
         }
     }
 
-    public function __attach(service s, map<any> annotationData) returns error? {
-        return self.register(s, annotationData);
+    public function __attach(service s, string? name = ()) returns error? {
+        // ignore param 'name'
+        return self.register(s, {});
```
hasithaa: Is using `{}` this intentional? I think parsing service name to register function is useful. @ThisaruGuruge @rdhananjaya wdyt ?

ThisaruGuruge: We need an `attachment` parameter to pass inside the service. We used this to pass that `attachment` parameter, if the developer provide it. We use it in [Task Scheduler](https://github.com/ballerina-platform/ballerina-lang/blob/746d2424ecd17f83b6d9b324c6203016a3d13967/stdlib/task/src/main/ballerina/task/scheduler.bal#L33)

hasithaa: Shall we create an issue to track this, Ideally, with the new "Annotation and reflection API" design we should be able to handle this inside the method, without user providing this.

ThisaruGuruge: Sure. I'll create one.

stdlib/http/src/main/ballerina/http/non_listening_service_endpoint.bal
wggihan: @chamil321 Can you please check the feasibility of making this private?

chamil321: @wggihan Sure I'll check

hasithaa:
@rdhananjaya. Example nats-basic-client is added recently and it is failing. Can you fix that. ?

PR #14544, opened 2019-03-19
[Documentation] [stdlib] [task] - Review Module.md file after redesign

ThisaruGuruge:


stdlib/task/src/main/ballerina/task/Module.md
```diff
@@ -1,18 +1,18 @@
 ## Module overview
 
-This module includes functions to manage Task Listeners and Task Schedulers.
+This module provides functionality to configure and manage Task Listeners and Task Schedulers, which will be executed periodically.
```
pubudu91: "which will be executed periodically." <- What is meant by this? What does it mean to execute a task listener or a task scheduler periodically?

stdlib/task/src/main/ballerina/task/Module.md
```diff
@@ -1,18 +1,18 @@
...
+There are two types of configurations for a Listener.
 - `TimerConfiguration`
 - `AppointmentConfiguration`
 
-If a task is needed to be run periodically, `TimerConfiguration` can be used. `TimerConfiguration` consists of three fields.
-- `interval` - Timer interval by which the listener should trigger. This should be in Milliseconds.
-- `initialDelay` - This is an optional field to state an initial delay, before the task should trigger. If this is set to `0`, task will run immediately. If the field is not set, interval will be taken as the initial delay as well. This should be given in Milliseconds.
-- `noOfRecurrences` - If there's a requirement to run a particular task only for a number of times, this field can be used. This should be given as an int.
+If a task is needed to be run periodically, `TimerConfiguration` can be used. `TimerConfiguration` consists of three fields, from which two are optional.
+- `interval` - Timer interval by which the listener should trigger. This should be given in Milliseconds.
```
pubudu91:
```suggestion
- `interval` - Timer interval by which the listener should trigger. This should be given in milliseconds.
```

stdlib/task/src/main/ballerina/task/Module.md
```diff
@@ -1,18 +1,18 @@
...
+- `initialDelay` - [Optional] Initial delay before the task should trigger. If this is set to `0`, task will run immediately. If the field is not set, interval will be taken as the initial delay. This should be given in Milliseconds.
```
pubudu91:
```suggestion
- `initialDelay` - [Optional] Initial delay before the task should trigger. If this is set to `0`, task will run immediately. If the field is not set, interval will be taken as the initial delay. This should be given in milliseconds.
```

stdlib/task/src/main/ballerina/task/Module.md
```diff
@@ -1,18 +1,18 @@
...
+- `noOfRecurrences` - [Optional] If there's a requirement to run a particular task only for a number of times, this field can be used. This should be given as an int.
```
pubudu91:
```suggestion
- `noOfRecurrences` - [Optional] If there is a requirement to run a particular task only for a number of times, this field can be used. This should be given as an `int`.
```
Avoid using shortened forms of phrases such as "there's"

stdlib/task/src/main/ballerina/task/Module.md
praneesha:
```suggestion
This module provides the functionality to configure and manage Task Listeners and Task Schedulers, which will be executed periodically.
```

stdlib/task/src/main/ballerina/task/Module.md
```diff
@@ -1,109 +1,25 @@
...
-    }
-```
-
-### Task Scheduler
-
-A task `Scheduler` can be used to create timers / appointments dynamically. Then a service can be attached to the `Scheduler`. Scheduler can also be created as a timer or an appointment. The configurations are as same as the `Listener`.
+A Task `Listener` can be used to create a service listener, which will be triggered at specified times.
 
-Code snippet for an example timer `task:Scheduler`. 
+Below are the two types of configurations that an be used to configure a Task Listener, either as a timer or as an appointment.
```
Bhashinee:
```suggestion
Below are the two types of configurations that can be used to configure a Task Listener, either as a timer or as an appointment.
```

PR #14549, opened 2019-03-19
Enable standard libraries tests in gradle

keizer619:


stdlib/io/build.gradle
```diff
@@ -1,15 +1,36 @@
...
+    doFirst {
+        copy {
+            from "$buildDir/generated-balo/repo/ballerina"
+            into "$buildDir/lib/repo/ballerina"
+        }
+    }
+    useTestNG() {
+        suites 'src/test/resources/testng.xml'
+    }
+}
```
riyafa: Shall we add new line in all places?

PR #14580, opened 2019-03-20
[jBallerina] Add initial classes required to load Java packages imported in Ballerina code

sameerajayasoma:
## Purpose
The purpose is to load Java packages imported in Ballerina source files.

## Approach
This design piggybacks onto the current Ballerina package loader
design. We can improve this further with the compiler redesign.

In the current design, PackageLoader loads Ballerina PackageSymbols
which can be categorized into two kinds: Ballerina source
package(from source code in the module directory), Ballerina binary
package(from Balos). I modeled PackageSymbols loaded from Java packages
as the third kind. Therefore I've introduced a class called
JPackageSymbolEnter to load/define PackageSymbols for Java packages.

Each PackageSymbol has a Scope which contains all the child symbols
defined in the package. The Scope interface has two methods.
The first one 'define()' allows you to add child symbols to the scope.
The second one 'lookup()' allows you to lookup for symbols defined
in the scope. Now in the case of Java packages, I introduced a new class
called JScope which extends from the current Scope class. Whenever
the JScope.lookup() method receives a request to lookup a Java class,
this method check whether it is already loaded, if not use class loaders
to load the requested class.

For each loaded/request Java class we need to create a corresponding
type and a symbol in the Ballerina compiler. I will include such details
in the next PR.

##Sample 
```ballerina
import ballerina/io;
import jballerina/java.util;

// The `main` function, which acts as the entry point.
public function main() {
    io:println("Hello, World!");
}
```


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/javainterop/JPackageClassLoader.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.wso2.ballerinalang.compiler.util.CompilerContext;
+
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+/**
+ * A class loader which loads java packages and classes used in Ballerina code.
+ *
+ * @since 0.991.0
```
Kishanthan: 0.995.0 ?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/javainterop/JPackageSymbol.java
```diff
@@ -0,0 +1,36 @@
...
+import org.ballerinalang.model.elements.PackageID;
+import org.wso2.ballerinalang.compiler.semantics.model.symbols.BPackageSymbol;
+import org.wso2.ballerinalang.compiler.semantics.model.symbols.BSymbol;
+
+// TODO We need to introduce a common package symbol interface/class
+// TODO Then define BallerinaPackageSymbol and JavaPackageSymbol from the base interface/class
+/**
+ * This class represents a package symbol of a Java package.
+ *
+ * @since 0.991.0
```
Kishanthan: 0.995.0 ?

Kishanthan:
@sameerajayasoma , we may have to rebase with master.

Kishanthan:
Closing this and the PR needs to be rebased with master.

PR #14586, opened 2019-03-21
Enable gradle test build in config-api

daneshk:
## Purpose
Enable gradle test build in config-api

### DIFF
--- a/stdlib/config-api/src/test/java/org/ballerinalang/stdlib/config/ComplexConfigKeyTest.java
+++ b/stdlib/config-api/src/test/java/org/ballerinalang/stdlib/config/ComplexConfigKeyTest.java
@@ -27,7 +27,6 @@ import org.testng.Assert;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
-import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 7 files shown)


stdlib/config-api/src/test/java/org/ballerinalang/stdlib/config/ComplexConfigKeyTest.java
```diff
@@ -42,8 +41,7 @@
 
     @BeforeClass
     public void setup() throws IOException {
-        resourceRoot = new File(getClass().getProtectionDomain().getCodeSource().getLocation().getPath())
-                .getAbsolutePath();
+        resourceRoot = Paths.get("src", "test", "resources").toAbsolutePath().toString();
```
pubudu91: Why  was this changed? The intention of the previous one was to take the sources and the config files from the `/target` directory. This will take it straight from the `src/test/resources` directory. This may be problematic in cases where the files get deleted.

daneshk: How built artifacts structured in output directory(/target, /build) is different in maven and gradle. We can't use the same way to get test resources in gradle. So we need to get it from src/test/resources directory. 
Do we have any use case to delete test resources?

pubudu91: Yes, there's a test for reading the secret for decrypting through a txt file. This secret file is deleted after reading once. If we point it to the resources dir, the resource will be deleted.

Issue #14596, opened 2019-03-21
Extern functions need to be defined as `= external` instead of `extern function`


MaryamZi:
**Description:**

The "Function definition" section of the specification specifies the following
```
function-defn := 
   metadata
   [public]
   function identifier function-signature function-body
function-body := function-body-block | external-function-body
external-function-body := = annots external ;
```

According to the above, functions with external implementations need to be specified as follows:
```ballerina
function externFunction() = external;
```

With the current implementation extern functions are defined as follows:
```ballerina
extern function externFunction();
```

This needs to be updated to be spec compliant.

PR #14608, opened 2019-03-21
Update external function syntax

MaryamZi:
## Purpose
$title.

Previous syntax:
```ballerina
extern function externFunction();
```

New syntax:
```ballerina
function externFunction() = external;
```

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/14596


compiler/ballerina-lang/src/main/resources/grammar/BallerinaParser.g4
```diff
@@ -109,7 +110,7 @@ sealedLiteral
 restDescriptorPredicate : {_input.get(_input.index() -1).getType() != WS}? ;
 
 objectFunctionDefinition
-    :   documentationString? annotationAttachment* deprecatedAttachment? (PUBLIC | PRIVATE)? (REMOTE | RESOURCE)? EXTERN? FUNCTION callableUnitSignature (callableUnitBody | SEMICOLON)
+    :   documentationString? annotationAttachment* deprecatedAttachment? (PUBLIC | PRIVATE)? (REMOTE | RESOURCE)? FUNCTION callableUnitSignature (callableUnitBody | (ASSIGN EXTERNAL)? SEMICOLON)
```
hasithaa: This allows having an external resource method, which is not supported by the runtime. I have created issue  https://github.com/ballerina-platform/ballerina-spec/issues/32 to capture this in the spec. Shall we disable in the compiler, saying this implementation doesn't support this functionality?

MaryamZi: Disallowed external resource functions with e8b0cbb, 371e4b0.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -628,8 +628,11 @@ error.cannot.attach.functions.to.abstract.object=\
 error.abstract.object.function.cannot.have.body=\
   function ''{0}'' in abstract object ''{1}'' cannot have a body
 
+error.resource.function.cannot.be.extern=\
+  resource function ''{0}'' cannot be an ''external'' function
```
hasithaa: Shall we say this is not supported in this implementation? We may have to allow this in the future. Also create an issue to track this, since this is allowed in the spec.

MaryamZi: Updated with 8cadac7.

Issue #13948, opened 2019-02-27
The =+ operator is not defined to concatenate a string and an xml.


hasithaa:

rdhananjaya:
How do we handle '<', '>' etc.,

xml e = xml `<elem>abc</elem>`;
e += "<";

hasithaa:
Fixed with #14627.

Issue #14612, opened 2019-03-21
Improve XML literal to allow '&lt;' '&gt;' and '&amp;'


rdhananjaya:
**Description:**
```ballerina
xml a = xml `&lt;&gt;&amp`;
```

- a is a xml sequence containing 3 code points (<,>,&)
- length of a is 3
- `a.elements()` should return 3 xml elements (<,>,&)
- string representation of a is "&amp;lt;&amp;gt;&amp;amp;"

Suggested assignee: @rdhananjaya

rdhananjaya:
Fixed in https://github.com/ballerina-platform/ballerina-lang/pull/14627

rdhananjaya:
Fixed in https://github.com/ballerina-platform/ballerina-lang/pull/14627

PR #14627, opened 2019-03-22
Allow addition and compound assigment of string and xml

rdhananjaya:
## Purpose
This PR fixes https://github.com/ballerina-platform/ballerina-lang/issues/13948 and https://github.com/ballerina-platform/ballerina-lang/issues/14612

```ballerina
xml a = xml `<a>hello</a>`;
string b = "stringLit";
xml c = a + b; // c = <a>hello</a>stringLit
c += "-str-2" // c = <a>hello</a>stringLit-str-2

```

This also fixes xml.length() function returning incorrect number of codePoints when applied to char sequence.

```ballerina
xml a = xml `abc`;
xml b = xml `def`;
xml c = a + b;

int nA = a.length(); // prev: 1 | after this pr: 3
int nB = b.length(); // prev: 1 | after this pr: 3
int nC = c.length(); // prev: 2 | after this pr: 6
```


bvm/ballerina-core/src/main/java/org/ballerinalang/model/util/XMLUtils.java
```diff
@@ -335,7 +335,11 @@ public static BXML tableToXML(BTable table) {
      */
     public static BXML<?> createXMLText(String content) {
         // Remove carriage return on windows environments to eliminate additional &#xd; being added
-        content = content.replace("\r\n", "\n");
+        content = content
+                .replace("\r\n", "\n")
+                .replace("&gt;", ">")
```
rdhananjaya: <, >, & symbols even though represented as escaped characters in xml (as &lt; &gt; &amp), to xml infoset they are just the plain character (<, >, &). And they are counted  as one character each.

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/statements/compoundassignment/CompoundAssignmentTest.java
```diff
@@ -321,6 +322,16 @@ public void testCompoundAssignmentAdditionStructElements() {
         Assert.assertEquals(((BInteger) returns[0]).intValue(), 500);
     }
 
+    @Test(description = "Test compound assignment with addition.")
+    public void testCompoundAssignmentOfXmlAndString() {
+        BValue[] returns = BRunUtil.invoke(result, "xmlCompoundAssignment");
+        Assert.assertTrue(returns[0] instanceof BXMLSequence);
+        BXMLSequence sequence = (BXMLSequence) returns[0];
+        Assert.assertEquals(sequence.size(), 4);
```
MaryamZi: Just to clarify based on this example, according to the spec, "if one operand is a string and one is xml, the string is treated as if it were an xml sequence with one character item for each code point in the string".

So when we are adding "hah" and "blah", is it not like adding two sequences of lengths 3 and 4 respectively?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2502,6 +2512,16 @@ public void visit(BLangBinaryExpr binaryExpr) {
         }
     }
 
+    private BLangXMLTextLiteral createXMLTextLiteralNode(
```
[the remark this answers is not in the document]
rdhananjaya: I think we could leave it here as there is not much of a re-use for this method. I'm OK going with any.

tests/ballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/add-operation.bal
```diff
@@ -20,4 +20,20 @@ function intFloatAdd(int a, float b) returns (float) {
 
 function floatIntAdd(float a, int b) returns (float) {
     return a + b;
-}
\ No newline at end of file
+}
+
+function xmlXmlAdd(xml a, xml b) returns (xml) {
```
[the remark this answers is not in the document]
rdhananjaya: I was just following the files convention here.

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/statements/compoundassignment/CompoundAssignmentTest.java
rdhananjaya: Yes you are correct, there is a bug in BXMLSequence.size() method. Will fix that.

hasithaa:
@rdhananjaya I have merged, Maryams's external change PR, which caused this to conflict. Can you update you PR?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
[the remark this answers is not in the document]
hasithaa: Seems like a static method, Let's move this to ASTBuilderUtils.

composer/package-lock.json
```diff
@@ -955,8 +955,8 @@
       "integrity": "sha512-r5160ogAvGyHsal38Kux7YYtodEKOj89RGb28ht1jh3SJb08VwRwAKKJL0bGb04Zd/3r9FL3BFIc3bBidYffCA==",
       "dev": true,
       "requires": {
-        "exec-sh": "^0.3.2",
-        "minimist": "^1.2.0"
+        "exec-sh": "0.3.2",
```
hasithaa: Seems like this is added by mistake.  Let's revert this change.

tests/ballerina-integration-test/src/test/java/org/ballerinalang/test/socket/SocketBaseTest.java
```diff
@@ -55,6 +54,7 @@ public void start() throws BallerinaTestException {
         String[] args = new String[] { "-e", "certificate.key=" + privateKey, "-e", "public.cert=" + publicCert };
         serverInstance = new BServerInstance(balServer);
         serverInstance.startServer(balFile, "services", args, requiredPorts);
+        executor.execute(mockSocketServer);
```
hasithaa: Any reason to do this change? seems like this is not related to XML, string concatenation.

tool-plugins/vscode/package-lock.json
```diff
@@ -257,7 +257,7 @@
     },
     "async": {
       "version": "1.2.1",
-      "resolved": "https://registry.npmjs.org/async/-/async-1.2.1.tgz",
+      "resolved": "http://registry.npmjs.org/async/-/async-1.2.1.tgz",
```
hasithaa: Mistakenly committed.

rdhananjaya: I think these files are committed with merge. Will re-merge and push

PR #14631, opened 2019-03-22
Add BIR generation for future type and start construct

vinok88:
```
future<string> p = start foo();
```


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/model/BIRTerminator.java
```diff
@@ -99,6 +99,44 @@ public void accept(BIRVisitor visitor) {
...
     }
 
+    /**
+     * A async function call instruction.
+     * <p>
+     * e.g., _4 = call doSomethingAsynchronously _1 _2 _3
+     *
+     * @since 0.995.0
+     */
+    public static class AsyncCall extends BIRTerminator implements BIRAssignInstruction {
```
lankavitharana: shall we extend Call class?

vinok88: Fixed.

PR #14632, opened 2019-03-22
Fix issue with implicit init val check for cyclic dependencies

### LINKED ISSUE #14620: Creating an array when two records depend on each other crashes the compiler
This was originally raised in Slack: https://ballerina-platform.slack.com/archives/C47EAELR1/p1553195283093800

```ballerina
type A record {
    B b?;
};

type B record {
    A a?;
};

public function main() {
    A[] aArray = [];
}
```

pubudu91:


hasithaa:
@pubudu91 shall we check whether we can prevent this at compile time for the objects

pubudu91:
@hasithaa Created #14633 to track this

Issue #13810, opened 2019-02-22
[Messaging] Implement NATS Connector 


pamod:
**Description:**
Introduce NATS connector for ballerina. NATS connector would support streaming API.

Issue #14668, opened 2019-03-25
Include test cases pertaining to NATS module in testgrid


pamod:
**Description:**

NATS does not contain an in-memory version of server. Hence tests should be written via testgrid by spawning NATS streaming sever docker image. 

Following scenarios should be addressed,

- [x] Basic producer/consumer
- [ ] Durable topic consumer 
- [x] Request/Response pattern
- [ ] Manual and Auto Ack modes

PR #14669, opened 2019-03-25
Add NATS module

pamod:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/13810

## Approach
NATS producer and consumer was implemented to demonstrate the following messaging domains. 

- Point to point communication (Queues)
- Pub/Sub (Topics)
- Request/Reply

## Samples

**Producer**

```
import ballerina/nats;
import ballerina/io;

public function main() {
        nats:Producer producer = new({ host: "localhost", port: 4222, clientId: "p0" });
        var result = producer->send("subject_foo", "Hello Ballerina !");
        if (result is error) {
                io:println("Error occurred while producing the message");
        } else {
                io:println("Message produced successfully");
        }
}
```

**Consumer** 

```
import ballerina/nats;
import ballerina/io;

listener nats:Consumer subscription = new({ host: "localhost", port: 4222, clientId: "s0" });

@nats:ServiceConfig { subject: "demo" }
service demo on subscription {
    resource function onMessage(nats:Message msg) {
        io:println("Recived message : " + msg.getData());
    }
}
```

## Remarks

Test cases will be addressed through https://github.com/ballerina-platform/ballerina-lang/issues/14668 as writing test cases for NATS would require spawning NATS streaming server docker images. This will be done as a part of testgrid effort.


examples/nats-basic-client/publisher.bal
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2019 WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
riyafa: We do not include license headers for examples since they will be displayed on the website.

riyafa:
We need to include the examples in the index.json file

stdlib/nats/pom.xml
```diff
@@ -0,0 +1,333 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>${maven.compiler.plugin.version}</version>
+                <configuration>
+                    <compilerArgument>-proc:none</compilerArgument>
+                </configuration>
+            </plugin>
+            <!--plugin>
```
riyafa: Do we have a reason for having this commented code?

stdlib/nats/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+    implementation 'io.nats:jnats:2.1.0'
+
+    baloImplementation project(path: ':ballerina-builtin', configuration: 'baloImplementation')
+    baloImplementation project(path: ':ballerina-io', configuration: 'baloImplementation')
+    baloImplementation project(path: ':ballerina-utils', configuration: 'baloImplementation')
+}
+
+description = 'Ballerina - NATS'
+
+test {
```
riyafa: Shall we remove test related dependencies if we are not including tests in this PR?

stdlib/nats/pom.xml
```diff
@@ -0,0 +1,333 @@
...
+        <dependency>
+            <groupId>org.ballerinalang</groupId>
+            <artifactId>ballerina-io</artifactId>
+            <type>zip</type>
+            <classifier>ballerina-binary-repo</classifier>
+        </dependency>
+        <dependency>
+            <groupId>org.ballerinalang</groupId>
+            <artifactId>ballerina-launcher</artifactId>
+            <scope>test</scope>
```
riyafa: Shall we remove test related dependencies if we are not including tests in this PR?

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Constants.java
```diff
@@ -0,0 +1,69 @@
...
+     */
+    static final String IO_ERROR = "IOError";
+
+    /**
+     * Error code for i/o.
+     */
+    public static final String NATS_ERROR_CODE = "{ballerina/nats}NATSError";
+
+    /**
+     * Represents the struct which holds the connection.
```
riyafa: Shall we not use the term struct because we no longer have structs in Ballerina?

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
```diff
@@ -0,0 +1,102 @@
...
+    private static final String MESSAGE = "message";
+
+    /**
+     * Constructs a BallerinaException and throw it.
+     *
+     * @param message error message which should be included.
+     * @param context context of the ballerina call.
+     * @param throwable the exception stacktrace.
+     */
+    public static void throwBallerinaException(String message, Context context, Throwable throwable) {
```
riyafa: Shall we use exception instead of throwable?

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
```diff
@@ -0,0 +1,102 @@
...
+import org.ballerinalang.model.values.BString;
+import org.ballerinalang.model.values.BValue;
+import org.ballerinalang.util.exceptions.BallerinaException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Utilities for producing and consuming via NATS sever.
+ */
+public class Utils {
```
riyafa: Shall we rename to NatsUtils as generic class names like Utils are hard to search for in a larger project?

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Constructs a BallerinaException and throw it.
+     *
+     * @param message error message which should be included.
+     * @param context context of the ballerina call.
+     * @param throwable the exception stacktrace.
+     */
+    public static void throwBallerinaException(String message, Context context, Throwable throwable) {
+        LOGGER.error(message, throwable);
```
riyafa: Shall we pass in the logger from the respective class in the method? Otherwise all logs would be from the Utils class?

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Retrieve the receiver object.
+     *
+     * @param context Ballerina context which should be received.
+     * @return the receiver object struct.
+     */
+    public static Struct getReceiverObject(Context context) {
+        return BLangConnectorSPIUtil.getConnectorEndpointStruct(context);
```
riyafa: Currently it is recommended to use the BMap rather than the struct you can use the following to get the BMap:
```suggestion
        return (BMap<String, BValue>) context.getRefArgument(0);
```

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/connection/Init.java
```diff
@@ -0,0 +1,118 @@
...
+            BMap<String, BValue> serverConfig = (BMap<String, BValue>) context.getRefArgument(1);
+            Struct connectorEndpointStruct = BLangConnectorSPIUtil.getConnectorEndpointStruct(context);
+            StreamingConnection connection = connect(serverConfig);
+            connectorEndpointStruct.addNativeData(Constants.NATS_CONNECTION, connection);
+        } catch (IOException e) {
+            Utils.throwBallerinaException("Error occurred while establishing connection", context, e);
+        } catch (InterruptedException ignore) {
+            // ignore
+            Thread.currentThread().interrupt();
+        } catch (Throwable e) {
```
riyafa: Shall we catch Exception instead of Throwable?

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/consumer/Acknowledgment.java
```diff
@@ -0,0 +1,73 @@
...
+import org.ballerinalang.model.values.BValue;
+import org.ballerinalang.natives.annotations.BallerinaFunction;
+import org.ballerinalang.natives.annotations.Receiver;
+import org.ballerinalang.nats.nativeimpl.Constants;
+import org.ballerinalang.nats.nativeimpl.Utils;
+
+import java.io.IOException;
+
+/**
+ * Acknowledges to a given message.
```
riyafa:
```suggestion
 * Acknowledge a given message.
```

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/consumer/Create.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final String DURABLE_NAME = "durableName";
+    public static final int ZERO = 0;
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void execute(Context context, CallableUnitCallback callableUnitCallback) {
+        try {
+            Struct consumer = Utils.getReceiverObject(context);
```
riyafa: Shall we use BMap instead of Struct?

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/consumer/Listener.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+
+/**
+ * Handles incoming message for a given subscription.
+ */
+public class Listener implements MessageHandler {
+    /**
+     * Resource which the message should be dispatched.
```
riyafa:
```suggestion
     * Resource to which the message should be dispatched.
```

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/producer/RequestReply.java
```diff
@@ -0,0 +1,93 @@
...
+            callback.notifySuccess();
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isBlocking() {
+        return false;
```
riyafa: I guess this function can be blocking if there are no particular reason for making it non-blocking.

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/producer/Send.java
```diff
@@ -0,0 +1,78 @@
...
+import org.ballerinalang.nats.nativeimpl.Constants;
+import org.ballerinalang.nats.nativeimpl.Utils;
+
+import java.io.IOException;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Sends message to a given subject.
+ *
+ * @since 0.995
```
riyafa: Shall we add `@since` to other classes as well to be consistent?

examples/nats-basic-client/publisher.bal
```diff
@@ -0,0 +1,37 @@
...
+import ballerina/nats;
+import ballerina/io;
+
+// Produce a message to a subject in NATS sever.
+public function main() {
+    string message = "";
+    string subject = io:readln("Subject : ");
+    // Initializes a producer
+    nats:Producer producer = new({ host: "localhost", port: 4222, clientId: "p0" });
+    while (message != "!q") {
```
Manuri: Can we use a constant for "!q" with a suitable name implying what it means?

stdlib/nats/src/main/ballerina/nats/producer/producer.bal
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    # Produces a message to a NATS streaming server for the given subject.
+    #
+    # + subject - Could also be referred as the 'topic/queue' name.
+    # + message - Message could be either a string or json representation.
+    # + charset - Encoding of the message (by defaults it would be UTF-8).
+    # + return -  GUID of acknowledgment or the specific error.
+    public remote function send(string subject, string|json message, string charset = "UTF-8") returns string|error {
```
Manuri: Shouldn't charset be a finite type? I'm not sure whether finite types are allowed to be defaultable though..

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/consumer/Create.java
```diff
@@ -0,0 +1,115 @@
...
+        isPublic = true
+)
+public class Create implements NativeCallableUnit {
+
+    private static final String START_SEQ = "startSeq";
+    private static final String SUBJECT_FIELD = "subject";
+    private static final String MANUAL_ACK = "manualAck";
+    private static final String ACK_WAIT = "ackWait";
+    private static final String DURABLE_NAME = "durableName";
+    public static final int ZERO = 0;
```
Manuri: Can't we make this a private constant?

stdlib/nats/src/main/ballerina/nats/Module.md
```diff
@@ -0,0 +1,45 @@
...
+import ballerina/nats;
+import ballerina/io;
+
+public function main() {
+    nats:Producer producer = new({ host: "localhost", port: 4222, clientId: "p0" });
+    var result = producer->send("demo", "Hello Ballerina !!");
+    if (result is error) {
+       io:println("Error occurred while producing the message.");
+    } else {
+       io:println("GUID "+result+" received for the produced message.");
```
MaryamZi:
```suggestion
       io:println("GUID " + result + " received for the produced message.");
```

stdlib/nats/src/main/ballerina/nats/connection/connection.bal
```diff
@@ -0,0 +1,28 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+# Represents a NATS connection.
+public type Connection object {
+
+    # Initializes a connection from the given server config.
+    #
+    # + config - Information neccasary for NATS client to establish a connection with the server.
+    public function __init(ConnectionConfig config) {
```
MaryamZi: Since `host` and `port` seem to be required fields in the `ConnectionConfig` record, should we maybe remove them from the record and make them positional params for `__init()`.

stdlib/nats/src/main/ballerina/nats/consumer/listener.bal
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    extern function create(service serviceType, map<any> data) returns error?;
+
+    # Starts the listener in the lifecyle.
+    #
+    # + return - Error or ().
+    public function __start() returns error? {
+        //ignore : since connection can be re-used between multiple listeners
+        return ();
```
MaryamZi: We don't need these `return ()`s right? L58 too.

stdlib/nats/src/main/ballerina/nats/message/message.bal
```diff
@@ -0,0 +1,32 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+# Represents a message which will be pushed from the NATS server to the consumer.
+public type Message object {
+    private string content = "";
```
MaryamZi: Shouldn't we be introducing an `__init()` method with a string parameter to create the message instead of setting an empty string as content?

stdlib/nats/src/main/ballerina/nats/connection/connection.bal
pamod: There will be two main paradigms to use connection, 

a. We basically create a connection and reuse them between 1.* producers, 1.* consumers and combination of both of them (advance use case)
b. We introduce a simplified API to create a connection per producer or a consumer 

In order for us to fulfill both the aforementioned use cases, creation of the producer or the consumer should either accept an already established connection or accept the list of fields necessary to create a connection which is represented through ConnectionConfig and internally create a new connection  

If we extract the mandatory params included in the ConnectionConfig as positional parameters the downside would be the same list of parameters should be reflected in both producer and the consumer creation APIs to make it consistent and to ensure that API is more readable instead of having it flooded with too many parameters.

This would however be a subjective choice :)

stdlib/nats/src/main/ballerina/nats/message/message.bal
pamod: Message here represents the content received via the NATS server to the consumer. Content here would be initialized internally. There will be more operations which will be added in future to process the content i.e serialize content into JSON etc. Hence did not include the content as native data

stdlib/nats/src/main/ballerina/nats/producer/producer.bal
pamod: There're tons of charset encodings supported, hence did not attempt to restrict the encodings into a finite region, kept it as a string. 

However, your point is valid we should carefully study the list of charsets and include them into a set of finite values. This will also be applicable for the current socket, io and byte APIs which we should do in future.

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/producer/Send.java
pamod: This is added to all the classes which represents ballerina native function implementations.

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/producer/RequestReply.java
pamod: We use the callback to return the values here, we would gradually introduce the separate executor to handle this asynchronously which would be addressed in future.

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
pamod: As discussed offline, currently with reference to the ballerina native implantation it would be important to handle the error conditions which is captured in throwable.

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
pamod: Given Utils falls under nats package. It might not make sense to repeat the package + class in the naming. 

Also we would be re writing all of this in Ballerina in the next iteration. Current java based implementation would only be introduce to validated the high-level ballerina APIs

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
pamod: Same as described in https://github.com/ballerina-platform/ballerina-lang/pull/14669#discussion_r268972377

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/connection/Init.java
pamod: Same as described in https://github.com/ballerina-platform/ballerina-lang/pull/14669#discussion_r268928258

stdlib/nats/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+    implementation project(':ballerina-builtin')
+    implementation project(':ballerina-utils')
+    implementation 'io.nats:java-nats-streaming:2.1.0'
+    implementation 'io.nats:jnats:2.1.0'
+
+    baloImplementation project(path: ':ballerina-builtin', configuration: 'baloImplementation')
+    baloImplementation project(path: ':ballerina-io', configuration: 'baloImplementation')
+    baloImplementation project(path: ':ballerina-utils', configuration: 'baloImplementation')
+}
+
```
chamil321: Shouldn't we add test task and copy generated balo to ballerina repo(lib/repo/ballerina) here?

pamod: All tests related to NATS would go under testgrid. Explained that in the PR. Hence the test phase was removed as its not necessary

stdlib/nats/src/main/ballerina/nats/connection/config.bal
```diff
@@ -0,0 +1,34 @@
...
+# + port - Remote server port.
+# + clusterId - Name of the NATS server cluster.
+# + clientId - Unique identifier for the client.
+# + connectionTimeout - Number of seconds to hold an idle connection.
+# + maxPubAcksInFlight - Number of messages which could be dispatched without receiving acks.
+# + ackTimeout - Time (in seconds) to wait for an acknowledment before retrying.
+public type ConnectionConfig record {|
+    string host;
+    int port;
+    string clusterId = "test-cluster";
```
pubudu91: Should this be a field with a default value or an optional field?

pamod: Default. You need cluster id, typically the default NATS server starts with "test-cluster"

stdlib/nats/src/main/ballerina/nats/error.bal
```diff
@@ -0,0 +1,24 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+# Represents an I/O error.
+#
+# + message - error message.
+# + id - error id.
+public type IOError record {|
```
pubudu91: A bit misleading to call a record an error isn't it?

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
pubudu91: Why log this at all? This is getting logged to the `ballerina.internal` log file. That's only meant for logging unexpected exceptions/abnormal conditions in the runtime which caused it to crash and there's nothing the user can do about it (i.e., a bug on our part). Throwing a BallerinaException means it's intended for the user. It will end up as a panic. These should not be logged in the `ballerina.internal` log file

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Creates an error message.
+     *
+     * @param context context which is invoked.
+     * @param errCode the error code.
+     * @param errMsg  the cause for the error.
+     * @return an error which will be propagated to ballerina user.
+     */
+    public static BError createError(Context context, String errCode, String errMsg) {
```
pubudu91: There's an API for creating errors in `BLangVMErrors`. Why not use that?

stdlib/nats/src/main/ballerina/nats/error.bal
pamod: When we return errors we include error details in a record. Typically the convention we follow to include the detail error messages was in the following format which is prefixed by *Error

https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/crypto/src/main/ballerina/crypto/crypto.bal#L114

https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/http/src/main/ballerina/http/http_commons.bal#L178

The record explains the IOError thrown by NATS.

stdlib/nats/src/main/java/org/ballerinalang/nats/nativeimpl/Utils.java
pamod: This method uses BLangVMErrors :) 

What it does is to include the error message to the IOError record and create an error through BLangVMErrors. We call this in several places hence introduced this as a util method

PR #14746, opened 2019-03-28
Include ballerina examples to the gradle build

keizer619:


tests/ballerina-examples-test/build.gradle
```diff
@@ -0,0 +1,102 @@
...
+
+
+configurations {
+    ballerinaTools {
+        transitive false
+    }
+}
+
+dependencies {
+    ballerinaTools project(path: ':ballerina-tools', configuration: 'zip')
```
Anoukh: I think we also need to include implementation dependency on :ballerina-tools to make sure that the zip get's created before running the examples tests

PR #14759, opened 2019-03-28
Support panic and trap for jvm target 

warunalakshitha:
## Purpose
Support panic and trap statement in bir and jvm codegen.

## Approach
This section covers the error type and value, panic and trap constructs. Here error handling will be described in two sub sections with BIR and JVM target.

### BIR Design
**Error type and literal constructor**

```
%8 = new_error %9 %10;
```
**Panic**
Panic will be a terminator in BIR.

```
%4 = new_error %5 %6;
panic %4;
```

**Trap**
We will use error table to mark the trap expression in BIR. New basic block will be created for each trap expression. 
```
	Error Table 
		BB	|errorOp
		bb1	|%6
		bb3	|%11
		bb4	|%10
```

### JVM Code Gen
**Ballerina construct** | **JVM Representation**
-- | --
error | Runtime Exception
panic | Throw
trap | Try catch

## Samples

**Panic example** 
ex. 
```
type AccountNotFoundError error<string, string>;
function getAccountBalance(int accountID) returns int {
      
        string errorReason = "Account Not Found";
        AccountNotFoundError accountNotFoundError =  error(errorReason,"3");
        if (accountID == 0){
            panic accountNotFoundError;
        }
        return accountID;
    
}

    var result = getAccountBalance(0);
	return result;
}

```
this will result in 

Exception in thread "main" Account Not Found 3
	at hello.getAccountBalance(Unknown Source)
	at hello.main(Unknown Source)
	at hello.main(Unknown Source)

**Trap sample** 
```
   var x =   trap foo(0);
   if ( x is error) {
       return x;
   }
}
public function foo(int x) returns error|int  {
     if(x == 0){
        panic error("reason", {"message":"detail"});
    } else {
        return x;
    }
}
```

this will result in

reason {"message":"detail"}

## Remarks
We use java stacktrace directly when printing. This required to be formatted to comply with bal stacktrace.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


Kishanthan:
@warunalakshitha , conflicts, lets fix them

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/BuilderUtils.java
```diff
@@ -146,13 +146,17 @@ public static void compileAndWriteJar(Path sourceRootPath,
                                           boolean offline,
                                           boolean lockEnabled,
                                           boolean skipTests,
-                                          boolean enableExperimentalFeatures) {
+                                          boolean enableExperimentalFeatures, 
+                                          boolean dumpBIR) {
```
vinok88: Did we decide to add this?

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_method_gen.bal
```diff
@@ -26,6 +26,7 @@ function generateMethod(bir:Function func, jvm:ClassWriter cw, bir:Package modul
     // generate method desc
     string desc = getMethodDesc(func);
     jvm:MethodVisitor mv = cw.visitMethod(ACC_PUBLIC + ACC_STATIC, funcName, desc, (), ());
+    InstructionGenerator instGen = new(mv, indexMap, currentPackageName);
```
vinok88: Why did we need this change?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/BuilderUtils.java
warunalakshitha: Yes

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_method_gen.bal
warunalakshitha: Previously InstructionGenerator was created for each instruction (inside the loop) . Hence I moved out of the loop so we can reuse same object.

PR #14768, opened 2019-03-28
Add object attached function calls

manuranga:
Following will generate bir now (but it will still fail at jvm gen level)
```
type MyObj object {
   function a (int i){}
};
function f(){
    MyObj myObj = new;
    myObj.a(8);
}
```


Kishanthan:
@manuranga , BIR reading part for attached functions is already added?

Kishanthan:
@manuranga , btw, gradle has failed

manuranga:
Passed locally, might be an intermittent issue in `':ballerina-h2:test'`.

Issue #14779, opened 2019-03-29
ClosureDesugar throw NullPointer exception 


rdhananjaya:
**Description:**
```ballerina
public function foo(boolean test) {
    string a;
    if test {
        a = "a";
    } else {
        a = "b";
    }

    var bar = function () returns string {
        a = a + "a";
        return a;
    };
}
```

NullPointerException thrown at ```string a;``` line, when ClosureDesugar try to rewrite ```string a;``` with a map access.

PR #14781, opened 2019-03-29
Closure desugar: Rewrite only var decls with RHS expression

rdhananjaya:
## Purpose
```ballerina
public function laterInitCapture() returns string {
    string a; // This should not be re-written to map access.
    boolean test = true;
    if test {
        a = "a";
    } else {
        a = "b";
    }

    var bar = function () returns string {
        a = a + "a";
        return a;
    };

    return bar.call();
}
```

Declaration of ```string s;``` should not re-written to map access, as we do for other closure variable references.

This PR fixes https://github.com/ballerina-platform/ballerina-lang/issues/14779


hasithaa:
@rdhananjaya @NatashaWso2  Did we test for an uninitialized variable inside a closure? It seems like it is not working. Shall we fix that in a separate PR?

NatashaWso2:
@hasithaa Sure I'll fix it and send another PR.

PR #14788, opened 2019-03-29
[NATS] Add close connection

pamod:
## Purpose
Add capability to close a given NATS connection


examples/nats-basic-client/publisher.bal
```diff
@@ -21,4 +22,6 @@ public function main() {
             }
         }
     }
+    // Close the publisher connection.
+    var result = producer.close();
```
Manuri: If an error occurs shouldn't we log it?

Issue #14180, opened 2019-03-07
`any|error` is considered to be `any`


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    error e1 = error("error reason");
    any|error x = e1;
    io:println(x is any);
}
```

This fails with
```cmd
$ ballerina run test.bal
error: .::test.bal:6:16: unnecessary condition: expression will always evaluate to 'true'
```

But this should be allowed since error does not belong to `any`.

**Affected Versions:**
0.990.3

MaryamZi:
This is an issue with the assignability check.

Even the following works at the moment!

```ballerina
import ballerina/io;

public function main() {
    error e1 = error("error reason");
    any|error e2 = e1;
    any e3 = e2; // should fail!
    io:println(e3);
}
```

```cmd
$ ballerina run test.bal
error reason {}
```

MaryamZi:
The runtime check also fails with errors inserted to maps:
```ballerina
import ballerina/io;

public function main(string... args) {
    map<any> m1 = { one: "a", two: "b" };
    error errValOne = error("error reason one");

    insertMemberToMap(m1, "three", errValOne); // expected to panic
    io:println(m1);
}

public function insertMemberToMap(map<any|error> mapVal, string index, any|error member) {
    mapVal[index] = member;
}
```

Moved from https://github.com/ballerina-platform/ballerina-lang/issues/13230

PR #14844, opened 2019-04-01
 Fix assigning of error values to any constrained maps at runtime

Anoukh:
## Purpose
Fix the runtime check for inserting error values to an any constrained map, and nil values to non-nil constrained maps

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/14180

## Approach
Runtime always returns true if the map constraint is type any. This check was removed.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
```diff
@@ -4408,7 +4408,7 @@ private static boolean isValidMapInsertion(BType mapType, BValue value) {
         }
```
MaryamZi: Not related to changes introduced in this PR, but we should check the constraint when the value is `null` right, whether it is a nil-able type?

The following doesn't fail atm, and allows an inherent type violating insertion!

```ballerina
import ballerina/io;

    map<string> m = { one: "one" };
    insertNilToMap(m);
    io:println(m);
}

function insertNilToMap(map<any> m) {
    m.two = ();
}
```

```cmd
$ ballerina run test.bal
{"one":"one", "two":()}
```

Anoukh: Addressed with 2627ff2

bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVM.java
```diff
@@ -4403,12 +4403,8 @@ private static void insertToMap(Strand ctx, BMap bMap, String fieldName, BValue
     }
 
     private static boolean isValidMapInsertion(BType mapType, BValue value) {
-        if (value == null) {
-            return true;
-        }
-
         BType constraintType = ((BMapType) mapType).getConstrainedType();
-        if (constraintType == BTypes.typeAny || constraintType.equals(value.getType())) {
+        if (value != null && constraintType.equals(value.getType())) {
```
MaryamZi: The first check in `checkIsType` (after checking whether the value is mutable) is `sourceType == targetType || sourceType.equals(targetType)`. So I guess we can remove this check and just call `checkIsType`, WDYT? We can do it separately maybe.

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/object/ObjectEquivalencyTest.java
```diff
@@ -146,4 +146,13 @@ public void testObjectMemberOrder() {
         Assert.assertEquals(returns[0].stringValue(), "{age:45, name:\"Doe\", address:\"\"}");
         Assert.assertEquals(returns[1].stringValue(), "{age:35, name:\"John\", address:\"\"}");
     }
+
+
```
Anoukh:
```suggestion
```

PR #14845, opened 2019-04-01
Add recursive union and object types for BIR

manuranga:
see `types.bal` for examples.


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/writer/BIRTypeWriter.java
```diff
@@ -52,18 +53,22 @@
...
  * A ConstPool is used to store string typed information.
  * 
  * @since 0.995.0
  */
 public class BIRTypeWriter implements TypeVisitor {
+    public static final int TYPE_TAG_SELF = 50;
     private final ByteBuf buff;
 
     private final ConstantPool cp;
     private final BMapType anydataMapType =
```
MaryamZi: We can remove this too right, since we do not have to overwrite the error detail now?

manuranga: Thanks, fixed in #14885

PR #14861, opened 2019-04-02
Switch thread model based on runtime parameter

chamil321:
```
-e b7a.http.worker.nonblocking=true
```
> The default behavior is false which represents the current thread model with the ballerina worker pool. When user specifies the parameter as true, worker pool will not be applied but the single thread will perform operations in a non-blocking manner.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVMExecutor.java
+++ b/bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVMExecutor.java
@@ -18,6 +18,7 @@
 package org.ballerinalang.bre.bvm;
 
 import org.ballerinalang.bre.bvm.Strand.State;
+import org.ballerinalang.config.ConfigRegistry;
 import org.ballerinalang.model.types.BType;
 import org.ballerinalang.model.types.TypeTags;
 import org.ballerinalang.model.values.BBoolean;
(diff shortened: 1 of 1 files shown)


bvm/ballerina-core/src/main/java/org/ballerinalang/bre/bvm/BVMExecutor.java
```diff
@@ -148,7 +149,11 @@ public static void executeResource(ProgramFile programFile, FunctionInfo resourc
         ObserveUtils.startResourceObservation(strand, observerContext);
 
         BVMScheduler.stateChange(strand, State.NEW, State.RUNNABLE);
-        BVMScheduler.schedule(strand);
+        if (ConfigRegistry.getInstance().getAsBoolean("b7a.http.worker.nonblocking")) {
```
lankavitharana: This is wrong IMO, you should do this in http level, in ballerina core level there should be two API methods exposed to run in same thread or to submit to a pool. Can we please correct this?

Issue #14860, opened 2019-04-02
Allow users to specify a configuration file to the build command


NatashaWso2:
**Description:**
As of now there's no way to specify a configuration file for the build command. But we should allow the user's to specify a config file if needed using the '--config' flag.

**Affected Versions:**
0.990.4-SNAPSHOT

PR #14862, opened 2019-04-02
Add '--config' flag to specify a configuration file to the build command

NatashaWso2:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/14860

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


pubudu91:
We didn't agree upon anything for this right? Like would this be misleading (considering the config file here is a runtime thing required for the tests)?

NatashaWso2:
@pubudu91 added the config flag here since there is an inconsistency where we allow the '--config' flag in the test command but not in the build command.

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/BuildCommand.java
```diff
@@ -182,6 +185,10 @@ public void execute() {
                 BuilderUtils.compileAndWriteJar(sourceRootPath, pkgName, targetFileName, buildCompiledPkg,
                         offline, lockEnabled, skiptests, experimentalFlag, dumpBIR);
             } else {
+                // Load the configuration file. If no config file is given then the default config file i.e.
+                // "ballerina.conf" in the source root path is taken.
+                LauncherUtils.loadConfigurations(sourceRootPath, configFilePath);
```
pubudu91: Doesn't this mean that we won't be able to give a config file to the build command if the target is the JVM?

misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/util/TesterinaUtils.java
```diff
@@ -198,10 +196,7 @@ public static void setManifestConfigs(Path sourceRoot) {
      */
     public static void executeTests(Path sourceRootPath, Map<BLangPackage, CompiledBinaryFile.ProgramFile>
             programFileMap) {
-        // Load configuration file. The default config file is taken "ballerina.conf" in the source root path
-        LauncherUtils.loadConfigurations(sourceRootPath, new HashMap<>(), null, false);
```
pubudu91: Why was this removed? Wouldn't this cause problems if the user runs `ballerina test` with a config file provided?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/BuildCommand.java
NatashaWso2: We should allow it. Fixed with 0071468. Thanks :)

PR #14863, opened 2019-04-02
Update CHANGELOG.md

rasika:


pahans:
please use [skip ci] for commits for like this. (which bypasses unnecessary CI/CD )

PR #14888, opened 2019-04-03
Fix xml attribute xml test

rdhananjaya:
## Purpose
Fix nil print mismatch in ballerina examples test

### DIFF
--- a/examples/xml-attributes/tests/xml_attributes_test.bal
+++ b/examples/xml-attributes/tests/xml_attributes_test.bal
@@ -11,6 +11,7 @@ int counter = 0;
 }
 public function mockPrint(any... s) {
     if (s[0] is ()) {
+        // Cannot convert () to string.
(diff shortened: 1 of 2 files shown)


examples/xml-attributes/tests/xml_attributes_test.bal
```diff
@@ -11,6 +11,7 @@ int counter = 0;
 }
 public function mockPrint(any... s) {
     if (s[0] is ()) {
+        // Cannot convert () to string.
```
MaryamZi: Maybe we can eventually improve this test to assert value equality without converting to strings.

rdhananjaya: +1

Manuri:
Merging without waiting for travis since this doesn't have an effect on any tests.

PR #14976, opened 2019-04-08
Downgrade mimepull version to 1.9.7

Manuri:
## Purpose
$Subject. This is due to the incompatible license change in 1.9.11


Manuri:
Closing since the EDL license has been identified as compatible with Apache 2.0, hence there is no need to downgrade the version.

PR #15006, opened 2019-04-11
Add Path api resolve() and match() support

### LINKED ISSUE #13811: Introduce PATH APIs for Ballerina
**Description:**
Introduces the PATH APIs for Ballerina. Currently they're offered as a part of the ballerina/internal module 

#### Design
- [x] Finalize API V1 Design document

#### Implementation
- [x] absolute() function
- [x] isAbsolute() function
- [x] filename() function
- [x] parent() function
- [x] normalize() function
- [x] resolve() function
- [x] split() function
- [x] build() function
- [x] extension() function
- [x] relative() function
- [x] match() function

daneshk:
## Purpose
This PR is to add resolve() and match() function support for Path API.

Resolves #13811 

## Approach
Implemented native functions using java path api.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [x] Ballerina By Example Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/filepath/src/main/java/org/ballerinalang/stdlib/filepath/Utils.java
```diff
@@ -45,18 +45,38 @@
      * @return      Ballerina error object.
      */
     public static BError getPathError(String reason, Throwable error) {
+        BError bError;
```
chamil321: Shall we simplify this method as follows?
```ballerina
string errorMsg = error != null && error.getMessage() != null ? error.getMessage() : UNKNOWN_MESSAGE;
return getPathError(reason, errorMsg);
```

daneshk: Fixed in bd349a5dc3dedcb64a08b3c49422502898666a7a

stdlib/filepath/src/main/java/org/ballerinalang/stdlib/filepath/Utils.java
```diff
@@ -45,18 +45,29 @@
      * @return      Ballerina error object.
      */
     public static BError getPathError(String reason, Throwable error) {
+        String errorMsg = error != null && error.getMessage() != null ? error.getMessage() : UNKNOWN_MESSAGE;
+        return getPathError(reason, errorMsg);
+    }
+
+    /**
+     * Returns error struct of input type.
```
riyafa: Shall we not use the term struct as it is no more in Ballerina?

stdlib/filepath/src/main/java/org/ballerinalang/stdlib/filepath/nativeimpl/Matches.java
```diff
@@ -0,0 +1,75 @@
...
+public class Matches extends BlockingNativeCallableUnit {
+
+    private static final String GLOB_SYNTAX_FLAVOR = "glob:";
+
+    @Override
+    public void execute(Context context) {
+        String inputPath = context.getStringArgument(0);
+        String pattern = context.getStringArgument(1);
+        FileSystem fs = FileSystems.getDefault();
+        if (pattern != null) {
```
ThisaruGuruge: Can pattern be null here ?

stdlib/filepath/src/main/java/org/ballerinalang/stdlib/filepath/Utils.java
```diff
@@ -45,18 +45,29 @@
...
+     * Error type is generic ballerina error type. This utility to construct error struct from message.
+     *
+     * @param reason    Reason for creating the error object. If the reason is null, "UNKNOWN" sets by
```
riyafa:
```suggestion
     * @param reason    Reason for creating the error object. If the reason is null, value "UNKNOWN" is set by
```

stdlib/filepath/src/main/java/org/ballerinalang/stdlib/filepath/Utils.java
```diff
@@ -45,18 +45,29 @@
...
+     *                  default.
+     * @param details     Java throwable object to capture description of error struct. If throwable object is null,
+     *                  "Unknown Error" sets to message by default.
```
riyafa:
```suggestion
     *                  "Unknown Error" is set to message by default.
```

stdlib/filepath/src/main/ballerina/filepath/file_path.bal
```diff
@@ -347,6 +347,22 @@ public function relative(string base, string target) returns string|error {
     return cleanTarget.substring(t0, tl);
 }
 
+# Returns the filepath after the evaluation of any symbolic links.
+# If path is relative the result will be relative to the current directory, unless one of the components is an absolute symbolic link.
```
riyafa: Line is longer than 120 characters?

stdlib/filepath/src/main/java/org/ballerinalang/stdlib/filepath/nativeimpl/Matches.java
daneshk: Yes, we don't need null check here. This is always true. I will fix it

PR #15026, opened 2019-04-12
Add APIs for runtime value creation at JVM side

Kishanthan:
This PR adds APIs for runtime value creation, which are needed at stdlibs when calling functions of an object (eg: service dispatching).


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ValueCreator.java
```diff
@@ -0,0 +1,37 @@
+package org.ballerinalang.jvm.values;
```
warunalakshitha: Shall we add Licence header

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_method_gen.bal
```diff
@@ -49,6 +49,16 @@ function generateMethod(bir:Function func, jvm:ClassWriter cw, bir:Package modul
...
         generateInitFunctionInvocation(module, mv);
         generateUserDefinedTypes(mv, module.typeDefs);
+
+        mv.visitTypeInsn(NEW, typeOwnerClass);
+        mv.visitInsn(DUP);
+        mv.visitMethodInsn(INVOKESPECIAL, typeOwnerClass, "<init>", "()V", false);
+        mv.visitVarInsn(ASTORE, 1);
+        mv.visitLdcInsn(currentPackageName);
+        mv.visitVarInsn(ALOAD, 1);
+        mv.visitMethodInsn(INVOKESTATIC, "org/ballerinalang/jvm/values/ValueCreator", "addValueCreator",
```
warunalakshitha: We can use VALUE_CREATOR constant here as well.

PR #15069, opened 2019-04-17
Update observe/config Module.mds

Manuri:
## Purpose
$Subject


Manuri:
Merging without wating for travis since this is a doc change

PR #15085, opened 2019-04-18
Remove forkEvery to reduce build time

Anoukh:
## Purpose
$subject


pahans:
forkEvery was added due to intermittent failures in tests. +1 If this passes travis build.

PR #15123, opened 2019-04-20
Support positions info and source file name to code generation

warunalakshitha:
## Purpose
This pr will support positions in bir and code gen line numbers and source files in JVM target. In addition to that, position info will be reflected in stack-trace. This also support generating printable stack-trace using Java stack trace which will be printed based on current ballerina stack-trace format.
Implementation will also support generation of classes based on source file name.

## Approach

Module classes will be generated per source file. Line numbers will be added with labels.

## Samples

**Module restucture**
```
_module_

 myorg/mymodule

foo.bal
bar.bal
```

Generated classes

```
myorg.mymodule.___init.class
myorg.mymodule.foo.class
myorg.mymodule.bar.class
```
Additionally frame classes  will be generated.

**Ballerina stacktrace**
```
error: reason bar 1 {"message":"bar"}
	at bar(hello.bal:46)
	   foo(hello.bal:36)
	   bas(hello.bal:40)
	   main(hello.bal:4)

```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/services/ErrorHandlerUtils.java
```diff
@@ -0,0 +1,52 @@
...
+package org.ballerinalang.jvm.services;
+
+import java.io.PrintStream;
+
+/**
+ * Class contains utility methods for ballerina server error handling.
+ *
+ * @since 0.995.0
+ */
+public class ErrorHandlerUtils {
```
Kishanthan: Shall we merge this with the BLangVMErrors class it self? Looks like both having some util methods for error handling in JVM?

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_method_gen.bal
```diff
@@ -1069,6 +1088,14 @@ function generateDefaultConstructor(jvm:ClassWriter cw, string ownerClass) {
     mv.visitEnd();
 }
 
+function generateDiagnosticPos(bir:DiagnosticPos pos, jvm:MethodVisitor mv) {
+    if (pos.sLine != -1) {
```
Kishanthan: Why this check is needed? When can it be -1?

tests/ballerina-test-utils/src/main/java/org/ballerinalang/test/util/BCompileUtil.java
```diff
@@ -570,30 +570,20 @@ private static CompileResult compileOnJBallerina(CompilerContext context, String
...
         try {
-            Method method = clazz.getDeclaredMethod(funcName, Strand.class);
+            Method method = initClazz.getDeclaredMethod(funcName, Strand.class);
             method.invoke(null, new Strand(new Scheduler()));
         } catch (Exception e) {
             throw new RuntimeException("Error while invoking function '" + funcName + "'", e);
         }
-
-        compileResult.setEntryClass(clazz);
+        compileResult.setClassLoader(classLoader);
```
warunalakshitha: Previously we had only one class file. Hence string entry class will be ok. But now when we have multiple class files and during test we need to load the classes on demand ( multiple entry points) . Hence I am storing class loader . If not static values created during init method will be lost.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/services/ErrorHandlerUtils.java
warunalakshitha: In BVM impl also we had two classes BLangVMErrors and ErrorHandlerUtils. ErrorHandlerUtils class was used in other stdlib modules like http, grpc, websub etc. +1 to move. But I think better to keep  same where migration will be easy. We can move once migration is done. WDYT?

Kishanthan: ack

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_method_gen.bal
warunalakshitha: We have Goto  terminators without position hence this will have default values -1. These terminators will be skipped during code gen.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIREmitter.java
```diff
@@ -283,10 +303,25 @@ public void visit(BIROperand birOp) {
...
+            sb.append(":");
+            appendPos(sb, pos.sCol);
+            sb.append("-");
+            appendPos(sb, pos.eCol);
+        }
+    }
+
+    private void appendPos(StringBuilder sb, int line) {
+        if (line < 10) {
+            sb.append("0");
```
vinok88: What is the logic behind this?

warunalakshitha: This is to print the position data with two digits with bir dump option mainly for readability purpose. 

ex. 
04-04:21-22	%2 = CONST_LOAD <int> 2;
04-04:24-25	%3 = CONST_LOAD <int> 3;

PR #15182, opened 2019-04-30
Enable tests related to Add operation

sameerajayasoma:
## Purpose
This PR enables unit tests written to test the Add operation in jBallerina.

## Remarks
In this process, I found an issue in the XMLFactory.concatenate method and fixed it. 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/XMLFactory.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/XMLFactory.java
@@ -84,23 +84,15 @@ public class XMLFactory {
     static {
         Canonicalizer.init();
         try {
-            if (Canonicalizer.getInstance(CANONICALIZER_OMIT_COMMENTS) == null) {
-                Canonicalizer.register(CANONICALIZER_OMIT_COMMENTS,
-                        "org.apache.axiom.c14n.impl.Canonicalizer20010315OmitComments");
-            }
-            if (Canonicalizer.getInstance(CANONICALIZER_WITH_COMMENTS) == null) {
-                Canonicalizer.register(CANONICALIZER_WITH_COMMENTS,
-                        "org.apache.axiom.c14n.impl.Canonicalizer20010315WithComments");
-            }
-            if (Canonicalizer.getInstance(CANONICALIZER_EXCL_OMIT_COMMENTS) == null) {
-                Canonicalizer.register(CANONICALIZER_EXCL_OMIT_COMMENTS,
-                        "org.apache.axiom.c14n.impl.Canonicalizer20010315ExclOmitComments");
-            }
-            if (Canonicalizer.getInstance(CANONICALIZER_EXCL_WITH_COMMENTS) == null) {
-                Canonicalizer.register(CANONICALIZER_EXCL_WITH_COMMENTS,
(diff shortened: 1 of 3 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/XMLFactory.java
```diff
@@ -84,23 +84,15 @@
...
-            if (Canonicalizer.getInstance(CANONICALIZER_EXCL_OMIT_COMMENTS) == null) {
-                Canonicalizer.register(CANONICALIZER_EXCL_OMIT_COMMENTS,
-                        "org.apache.axiom.c14n.impl.Canonicalizer20010315ExclOmitComments");
-            }
-            if (Canonicalizer.getInstance(CANONICALIZER_EXCL_WITH_COMMENTS) == null) {
-                Canonicalizer.register(CANONICALIZER_EXCL_WITH_COMMENTS,
-                        "org.apache.axiom.c14n.impl.Canonicalizer20010315ExclWithComments");
-            }
-            canonicalizer = Canonicalizer.getInstance(CANONICALIZER_WITH_COMMENTS);
+            Canonicalizer.register("http://www.w3.org/TR/2001/REC-xml-c14n-20010315",
```
sameerajayasoma: I had to do this change to make it work. The original version was giving NPE. Please review this. @SupunS

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/XMLFactory.java
```diff
@@ -239,7 +231,7 @@
      * @return Concatenated XML sequence
      */
     public static XMLValue<?> concatenate(XMLValue<?> firstSeq, XMLValue<?> secondSeq) {
-        ArrayValue concatSeq = new ArrayValue();
+        ArrayValue concatSeq = new ArrayValue(new BArrayType(firstSeq.getType()), -1);
```
sameerajayasoma: In the earlier version, ArrayType was missing from the ArrayValue. I got an NPE down the line. Is this the best way to solve this? @SupunS

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/XMLFactory.java
SupunS: I had to do that change because with the new jvm runtime, now we have two XMLFactory classes, and both register the same set of canonicalizers twice. That was giving errors saying the same canonicalizer is already present.
Should we silently fail on `AlgorithmAlreadyRegisteredException`?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/XMLFactory.java
SupunS: +1. can use the constructor that takes only the array-type.

I think we need to remove these additional constructors. Will create an issue for that.

Kishanthan:
Closing as this is added.

Issue #15150, opened 2019-04-25
Implement typeof unary operator


rdhananjaya:
**Description:**
`typeof` operator is not implemented.

```
typeof-expr := typeof expression
```
The result of a typeof-expr is a typedesc value for the runtime type of v, where v is the result of evaluating expression.

The runtime type of v is the narrowest type to which v belongs. 
```ballerina
typedesc td0 = typeof i;
```

See section "Typeof expression" in Ballerina spec for more details.

Issue #15193, opened 2019-05-02
Add tooling support for typeof unary operator


rdhananjaya:
**Description:**
$subject.

```ballerina
type RecType0 record {
    string name;
};
typedesc t1 = typeof 22;

RecType0 i = { name: "theName"};
typedesc t2 = typeof i;
```

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Issue #15160, opened 2019-04-26
Complete implementation of BIR UnaryOp


rdhananjaya:
**Description:**
Complete BIR instruction UnaryOp implementation.

Blocker for https://github.com/ballerina-platform/ballerina-lang/issues/15150

PR #15194, opened 2019-05-02
Introduce typeof operator

rdhananjaya:
## Purpose
This PR introduce `typeof` operator as shown by following code fragment.

```ballerina
type RecType0 record {
    string name;
};
typedesc t1 = typeof 22;

RecType0 i = { name: "theName"};
typedesc t2 = typeof i;
```

Fixes: https://github.com/ballerina-platform/ballerina-lang/issues/15150 and https://github.com/ballerina-platform/ballerina-lang/issues/15160

## Check List 
- [ ] Updated Change Log
- [x] Checked Tooling Support (https://github.com/ballerina-platform/ballerina-lang/issues/15193)
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypedescProvider.java
```diff
@@ -0,0 +1,50 @@
...
+import org.ballerinalang.jvm.values.TypedescValue;
+
+/**
+ * Runtime impl to create typedesc values out of Ballerina values.
+ *
+ * @since 0.995.0
+ */
+public class TypedescProvider {
+    public static TypedescValue typeof(Object value) {
+        if (value instanceof RefValue) {
```
SupunS: Can use the `TypeChecker.getType(...)` method to get the type.

SupunS: We can also then move this method to the TypeChecker class it self, since it will be a one-liner :)

rdhananjaya: With that we can codegen to invoke TypeChecker.getTypedesc(), skipping this class altogether right?

SupunS: yes

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/TypedescValue.java
```diff
@@ -0,0 +1,72 @@
...
+        return this;
+    }
+
+    @Override
+    public boolean isFrozen() {
+        return true;
+    }
+
+    @Override
+    public Object freeze() {
```
MaryamZi: Since typedesc is not anydata, freeze should not be allowed on typedesc right? Any reason to allow freeze on typedesc, including returning `true` on `isFrozen()` and overriding `freeze()` instead of `attemptFreeze()`?

rdhananjaya: This is a mistake, I think I miss interpret the abstract freeze's immutable clause, with the builtin method's definition.

PR #15202, opened 2019-05-03
Fix sample issues

anupama-pathirage:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/examples/index.json
+++ b/examples/index.json
@@ -1014,7 +1014,7 @@
         "column": 3,
         "category": "Working over the network",
         "samples": [{
-                "name": "TCP Socket Basic Example",
+                "name": "Basic TCP Socket",
                 "url": "tcp-socket-listener-client"
             },
             {
--- a/examples/index.json
+++ b/examples/index.json
@@ -1018,7 +1018,7 @@
                 "url": "tcp-socket-listener-client"
             },
             {
-                "name": "UDP Client Socket Basic Example",
+                "name": "Basic UDP Client Socket",
                 "url": "udp-socket-client"
             }
         ]
--- a/examples/tcp-socket-listener-client/tcp_socket_client.out
+++ b/examples/tcp-socket-listener-client/tcp_socket_client.out
@@ -2,8 +2,8 @@
 # `.bal` file and use the `ballerina run` command.
 $ ballerina run tcp_socket_client.bal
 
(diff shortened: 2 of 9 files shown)


examples/tcp-socket-listener-client/tcp_socket_client.out
```diff
@@ -2,8 +2,8 @@
 # `.bal` file and use the `ballerina run` command.
 $ ballerina run tcp_socket_client.bal
 
-# `onConnect` invokes and prints the remote server port once the connection to the server is successful.
+// `onConnect` invokes and prints the remote server port once the connection to the server is successful.
```
MaryamZi: Will `//` render the comments separately with the .out files? We seem to have used `#` in other .out files too, should we update all if `//` works?

examples/index.json
```diff
@@ -1014,7 +1014,7 @@
         "column": 3,
         "category": "Working over the network",
         "samples": [{
-                "name": "TCP Socket Basic Example",
+                "name": "Basic TCP Socket",
```
MaryamZi: We seem to have several other "Xxx Basic Example" titled BBEs. Shall we rename them too?

examples/tcp-socket-listener-client/tcp_socket_client.out
Shan1024: I think `#` has been used to indicate that a particular line is not a command, but a comment. So I think it is better to use the `#` in all the places.

anupama-pathirage: If we use // instead of # in when rendering in site, they will appear as comment boxes towards the right hand side of the page. (Similar to the code comments) But yes, we need to change all the places if we agreed to do that change. Hence will revert this change and we can fix all later if needed

examples/index.json
anupama-pathirage: Fixed the remaining socket example as well. Stdlib team will revisit the remaining samples and will provide meaningful names. @wggihan Please note.

Issue #15084, opened 2019-04-18
jBallerina sprint - Week 17 - 2019/04/22 to 2019/04/26


Kishanthan:
- [ ] BIR Symbol Enter - @lankavitharana 
- [ ] Locks BIR - @manuranga 
- [ ] Locks Codegen - @manuranga
- [ ] Http stdlib migration - @chamil321 
- [x] Function Pointers Codegen - @vinok88
- [ ] jBallerina dist and commands naming - @warunalakshitha 
- [ ] Remaining builtin methods - @warunalakshitha 
- [ ] I/O module compilation - @SupunS 
- [ ] Records test cases enable ? - @SupunS
- [ ] Table type BIR model - @gimantha 
- [ ] Generic native function mapping at module level - @Kishanthan

PR #15238, opened 2019-05-07
Add compiled package symbol enter initial phase

lankavitharana:
## Purpose
This is to add new Symbol enter phase for bir files.

Fixes #<Issue Number>
Part of https://github.com/ballerina-platform/ballerina-lang/issues/15084

## Approach
Save the BIR in the balo zip file along with the current balo(this is for the time being, should remove the balo later), then use that bir for the symbol enter when some other package depends on this package. For the time being, we use --jvmTarget flag to differentiate these two flows (like when u build some package with --jvmTarget flag, then it will look bir files for dependent packages as well.

## Remarks
This PR also modifies the build process as well, change is as follows,
if you build a project with --jvmTarget flag, then it will generate bir (and include that for every balo zip libraries) and generate jar files where applicable (for packages with entry points)

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/model/BIRNode.java
```diff
@@ -125,6 +125,25 @@ public void accept(BIRVisitor visitor) {
...
+        public Name name;
+
+        public BIRParameter(DiagnosticPos pos, Name name) {
+            super(pos);
+            this.name = name;
+        }
+
+        @Override
+        public void accept(BIRVisitor visitor) {
+//            visitor.visit(this);
```
Kishanthan: This visit is commented out?

lankavitharana: yeah right now it's serialized without the help of the visitor, it only writes an integer, hence thought of writing that then and there, can do with the visitor as well. If the visitor is the better way, will add that  in the next PR

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -0,0 +1,1394 @@
...
+        dataInStream.readInt();
+    }
+
+    private int visibilityAsMask(int flags, byte visibility) {
+        switch (visibility) {
+            case 0:
+                return flags;
+            case 1:
+                return flags | Flags.PRIVATE;
+            case 2:
```
SupunS: shall we use constants?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -0,0 +1,1394 @@
...
+//        defineSymbols(dataInStream, rethrow(this::defineService));
+
+        // Resolve unresolved types.
+//        resolveTypes();
+
+        // Read resource info entries.
+//        defineSymbols(dataInStream, rethrow(this::defineResource));
+
+        // Define constants.
+//        defineSymbols(dataInStream, rethrow(this::defineConstants));
```
SupunS: How do we know the number of bytes to be ignored in these commented out cases?

lankavitharana: those are not serialized yet, only what is serialized is getting deserialized now, I just kept these so that I know these are needed to be symbol entered, anyway there are some code cleanups to be done, will do them in a separate PR

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
lankavitharana: +1 shall we do that in a separate PR? basically if no conceptual issue, shall we merge this? because I'm anyway working on code cleaning and improving the symbol enter, will improve these with that

PR #15251, opened 2019-05-08
Update records BBE to comply with the latest spec changes

### LINKED ISSUE #15162: Update the record BBEs
With the new changes introduced by PR #14872, the record related BBEs need to be updated as well.

pubudu91:
Fix #15162

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [x] Ballerina By Example Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/examples/records/records.bal
+++ b/examples/records/records.bal
@@ -10,7 +10,7 @@ type Student record {
 };
 
 // This defines a closed record type named `Address`. The `{|` and `|}` delimiters indicate that this record type
-// exclusively allows mapping values which contain the described fields.
+// allows mapping values which contain only the described fields.
 type Address record {|
     string city;
     string country;
@@ -22,7 +22,7 @@ type Grades record {|
     int maths;
     int physics;
     int chemistry;
-    // This is an `int` typed rest field. All additional fields should be of the rest field's type or a subtype of it.
+    // This is a rest field of the type`int`. All additional fields should be of the rest field's type or a subtype of it.
     int...;
 |};
 
@@ -59,6 +59,6 @@ public function main() {
     // This adds an additional field not defined in the record type descriptor above.
     // Note that an attempt to add additional fields to a closed record results in compile errors.
     // e.g., `peter.address.street = "Palm Grove";` will result in a compile error.
-    peter.address = <Address>{city: "Colombo", country: "Sri Lanka"};
+    peter.address = <Address>{ city: "Colombo", country: "Sri Lanka" };
     io:println(peter);
 }
--- a/examples/records/records.bal
+++ b/examples/records/records.bal
@@ -1,22 +1,22 @@
 import ballerina/io;
 
-// This defines an open record type named `Student`. The `{` and `}` delimiters indicate that this record type
-// allows mapping values with additional pure-typed (i.e., `anydata|error`) fields in addition to the fields described
-// in the descriptor. The descriptor `record { }` is equivalent to `record {| (anydata|error)...; |}`.
+// Define an open record type named `Student`. The `{` and `}` delimiters indicate that in addition to the fields
+// defined here, this record type allows additional fields with pure-typed (i.e., `anydata|error`) values.
+// The descriptor `record { }` is equivalent to `record {| (anydata|error)...; |}`.
 type Student record {
     string name;
     int age;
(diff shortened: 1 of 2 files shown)


examples/records/records.bal
```diff
@@ -1,23 +1,32 @@
...
     string name;
     int age;
     Grades grades;
-    // This is a `string` typed rest field. All additional fields should be of the rest field's type or a subtype of it.
-    // In the absence of an explicit rest field, an implicit rest field of type `anydata|error` is assumed.
-    string...;
+};
+
+// This defines a closed record type named `Address`. The `{|` and `|}` delimiters indicate that this record type
+// exclusively allows mapping values which contain the described fields.
```
praneesha:
```suggestion
// exclusively allows mapping values, which contain the described fields.
```

examples/records/records.bal
```diff
@@ -1,23 +1,32 @@
...
 |};
 
-// This defines a closed record type named `Grades`. Closed records are defined using the `{|` and `|}` delimiters.
+// This defines an open record type named `Grades`. Although it is defined using the `{|` and `|}` delimiters, it has
+// an `int` rest field as well. Therefore, this is an open record type.
 type Grades record {|
     int maths;
     int physics;
     int chemistry;
+    // This is an `int` typed rest field. All additional fields should be of the rest field's type or a subtype of it.
```
praneesha:
```suggestion
    // This is a rest field of the type`int`. All additional fields should be of the rest field's type or a subtype of it.
```

examples/records/records.bal
```diff
@@ -1,23 +1,32 @@
 import ballerina/io;
 
-// This defines an open record type named `Student` which in addition to the fields defined here, allows additional
-// `string` fields.
-type Student record {|
+// This defines an open record type named `Student`. The `{` and `}` delimiters indicate that this record type
```
MaryamZi: Should we change "This defines an " to "Define an " ?

Given that this is also the first records example, should we start with what a record is? By saying if can hold values for the particular the fields, and then talk about additional fields?

examples/records/records.bal
MaryamZi: By exclusively are we trying to say we only allow the fields specified? How about something like the following instead?
```suggestion
// allows mapping values which contain only the described fields.
```

IMO, "exclusively allows mapping values" sounds like only this record type allows mapping values.

examples/records/records.bal
```diff
@@ -49,7 +58,7 @@ public function main() {
 
     // This adds an additional field not defined in the record type descriptor above.
     // Note that an attempt to add additional fields to a closed record results in compile errors.
-    // e.g., `peter.grades.ict = 77;` will result in a compile error.
-    peter.department = "Computer Science";
+    // e.g., `peter.address.street = "Palm Grove";` will result in a compile error.
+    peter.address = <Address>{city: "Colombo", country: "Sri Lanka"};
```
MaryamZi:
```suggestion
    peter.address = <Address>{ city: "Colombo", country: "Sri Lanka" };
```

examples/records/records.bal
pubudu91: "This defines an" is fine right? Given that we are referring to a record type descriptor by "this".

A brief explanation about what a record is given in the records.description file

examples/records/records.bal
pubudu91: Yeah, meant that only mapping values with those fixed set of fields are allowed

examples/records/records.bal
MaryamZi: IINM, we've used "Define a ...." in general instead of saying "This ...". We can decide on a convention maybe? I personally prefer the former.

IMO it'll still be useful to start with the fields here too. How about swapping to have something like:
```
// Define an open record type named `Student`. The `{` and `}` delimiters indicate that in addition to the fields defined here
// this record type allows values with additional pure-typed (i.e., `anydata|error`) fields.
```

pubudu91: +1 will change

PR #15316, opened 2019-05-15
Add azure pipelines configs

pahans:


Shan1024:
@pahans - Was there any specific reason to remove MacOS build? If not, shall we add MacOS build as well?

pahans:
> @pahans - Was there any specific reason to remove MacOS build? If not, shall we add MacOS build as well?

The plan is to fix windows build first then add OSX build.

PR #15338, opened 2019-05-18
Migrate Config API module into JVM types/values

chamil321:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/config-api/src/main/java/org/ballerinalang/stdlib/config/Contains.java
+++ b/stdlib/config-api/src/main/java/org/ballerinalang/stdlib/config/Contains.java
@@ -21,6 +21,7 @@ package org.ballerinalang.stdlib.config;
 import org.ballerinalang.bre.Context;
 import org.ballerinalang.bre.bvm.BlockingNativeCallableUnit;
 import org.ballerinalang.config.ConfigRegistry;
+import org.ballerinalang.jvm.Strand;
 import org.ballerinalang.model.types.TypeKind;
 import org.ballerinalang.model.values.BBoolean;
 import org.ballerinalang.natives.annotations.Argument;
@@ -46,5 +47,9 @@ public class Contains extends BlockingNativeCallableUnit {
         String configKey = context.getStringArgument(0);
         boolean containsKey = ConfigRegistry.getInstance().contains(configKey);
         context.setReturnValues(new BBoolean(containsKey));
+    }
+
+    public static boolean contains(Strand strand, String configKey) {
+        return ConfigRegistry.getInstance().contains(configKey);
     }
 }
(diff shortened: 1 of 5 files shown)


stdlib/config-api/src/main/java/org/ballerinalang/stdlib/config/GetConfig.java
```diff
@@ -92,4 +95,35 @@ private BMap buildBMap(Map<String, Object> section) {
...
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    private static MapValue buildMapValue(Map<String, Object> section) {
+        MapValue map = new MapValueImpl<String, Object>();
+
+        section.forEach((key, val) -> {
+            if (val instanceof String || val instanceof Long || val instanceof Double || val instanceof Boolean) {
+                map.put(key, val);
```
Kishanthan: What about other ref types?

stdlib/config-api/src/main/java/org/ballerinalang/stdlib/config/SetConfig.java
```diff
@@ -65,4 +66,8 @@ public void execute(Context context) {
 
         context.setReturnValues();
     }
+
+    public static void setConfig(Strand strand, String configKey, Object configValue) {
```
pubudu91: Again, what's the purpose of this method? As with the contains implementation, a `Strand` is passed as a param but not made use of.

stdlib/config-api/src/main/java/org/ballerinalang/stdlib/config/GetConfig.java
pubudu91: For configs, we only consider value types + maps of value types.

stdlib/config-api/src/main/java/org/ballerinalang/stdlib/config/SetConfig.java
Kishanthan: Strand is a first argument to any methods in ballerina. The impl may or may not use the strand argument based on the requirement. For example, if you want to perform a non-blocking invocation within the impl, then you may use the strand ref.

stdlib/config-api/src/main/java/org/ballerinalang/stdlib/config/Contains.java
```diff
@@ -47,4 +48,8 @@ public void execute(Context context) {
         boolean containsKey = ConfigRegistry.getInstance().contains(configKey);
         context.setReturnValues(new BBoolean(containsKey));
     }
+
+    public static boolean contains(Strand strand, String configKey) {
```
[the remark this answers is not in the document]
vinok88: That's how we model functions in ballerina. Strand should be the first param, regardless of its usage. Users may use strand to yield the execution etc..

PR #15408, opened 2019-05-24
Support untaint and add unary ops with tests enabled

Kishanthan:
$subject.


Kishanthan:
Resolves #15408 15382

PR #15410, opened 2019-05-24
Add support to insert byte[] as a direct parameter in update operation

### LINKED ISSUE #14710: Incorrect handling of BIT data in SQL Data Connectors
**Description:**
$subject in data connectors.

In SQL Server bit contains a boolean or null [1]. But in MySQL it is different.

In mysql a type of BIT(M) enables storage of M-bit values. M can range from 1 to 64. [2]
So if BIT(1) it should get mapped to boolean, if BIT(> 1) it should get mapped to byte[] [3]
But currently both cases match into boolean which is incorrect. We need to use GetColumnClassName to get the exact type.

[1] http://www.sqlservertutorial.net/sql-server-basics/sql-server-bit/
[2] https://dev.mysql.com/doc/refman/8.0/en/bit-type.html
[3] https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-reference-type-conversions.html

**Steps to reproduce:**

```ballerina
import ballerina/io;
import ballerina/mysql;
import ballerina/sql;
mysql:Client testDB = new({
    host: "localhost",
    port: 3306,
    name: "",
    username: "test",
    password: "test",
    dbOptions: { useSSL: false }
});

type BitTypes record {
    int id;
    boolean bitVal1;
    boolean bitVal2; //This should be a byte[]
};

    createDatabase();
    var ret = testDB->update("CREATE TABLE testdb.bittypes(
        id INT,
        bitVal1 BIT(1),
        bitVal2 BIT(2),
        PRIMARY KEY (id)
    )");
    handleUpdate(ret, "Create bittypes table");

    ret = testDB->update("INSERT INTO testdb.bittypes(
                            id, bitVal1, bitVal2
                        ) VALUES (
                            1, b'0', b'10' )");

    var selectRet = testDB->select("Select  id, bitVal1, bitVal2 from testdb.bittypes", BitTypes);
    if (selectRet is table<BitTypes>) {
        int index = 0;
        foreach var row in selectRet {
            io:println(row.id);
            io:println(row.bitVal1);
            io:println(row.bitVal2);
        }
    } else {
        io:print("ERROR OCCURED!!!:");
        io:println(selectRet);
    }                        
    dropDatabase();
}

function createDatabase() {
    var ret = testDB->update("CREATE DATABASE testdb");
    handleUpdate(ret, "Create database testdb");
}

function dropDatabase() {
    var ret = testDB->update("DROP DATABASE testdb");
    handleUpdate(ret, "Drop Database testdb");
}

function handleUpdate(sql:UpdateResult|error returned, string message) {
    if (returned is sql:UpdateResult) {
        io:println(message + " status: " + returned.updatedRowCount);
    } else {
        io:println(message + " failed: " + <string>returned.detail().message);
    }
}

```
**OUTPUT**
```
Create database testdb status: 1
Create bittypes table status: 0
1
false
true
Drop Database testdb status: 1
```

Similarly we should be able to insert byte[] as a direct parameter in the form of a byte[] or as a sql:Parameter. Need to check on the other usage of this column (ex : in / out parameters of call method) 

```ballerina
    int id = 2;
    boolean bitVal1 = true;
    byte[] bitVal2 = [1, 2];
    ret = testDB->update("INSERT INTO testdb.bittypes(
                            id,
                            bitVal1,
                            bitVal2
                        ) VALUES (
                            ?,
                            ?,
                            ?
                        )", id, bitVal1, bitVal2);
```
**Affected Versions:**
0.990.3

### LINKED ISSUE #15400: Problem on having fixed array size as 100 for byteValues in BValueArray
**Description:**

When I try to insert the byte array([1, 2, 3]) into mysql db, it returns the error "Data truncation: Data too long for column 'bitVal2' at row 1". This is because of the byte array specified in the mysql query contains bit values larger than the maximum bit values allowed. Here, A type of BIT(64) enables storage of 64-bit values.

When the provided byte array gets passed to the getByteArray method [1], that array contains byteValues with the fixed array size (100). At this line [2], the return byte array will be [1, 2, 3, 0, 0,.....0] with the size 100. Since 100 > 64, I'm getting that error output.

[1] https://github.com/ballerina-platform/ballerina-lang/blob/v0.991.0/stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java#L599
[2] https://github.com/ballerina-platform/ballerina-lang/blob/v0.991.0/stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java#L602

**Steps to reproduce:**
```
import ballerina/io;
import ballerina/mysql;
import ballerina/sql;
mysql:Client testDB = new({
        host: "localhost",
        port: 3306,
        name: "",
        username: "root",
        password: "root",
        dbOptions: { useSSL: false }
    });


    createDatabase();
    var ret = testDB->update("CREATE TABLE testdb.bittypes(
        id INT,
        bitVal1 BIT(1),
        bitVal2 BIT(64),
        PRIMARY KEY (id)
    )");

    handleUpdate(ret, "Create bittypes table");

    int id = 1;
    boolean bitVal1 = true;
    byte[] bitVal2 = [1, 2, 3];
    sql:Parameter para1 = { sqlType: sql:TYPE_INTEGER, value: id };
    sql:Parameter para2 = { sqlType: sql:TYPE_BOOLEAN, value: bitVal1 };
    sql:Parameter para3 = { sqlType: sql:TYPE_BINARY, value: bitVal2};

    ret = testDB->update("INSERT INTO testdb.bittypes(
                            id,
                            bitVal1,
                            bitVal2
                        ) VALUES (
                            ?,
                            ?,
                            ?
                        )", para1, para2, para3);
    handleUpdate(ret, "Update bittypes table");

    dropDatabase();
}

function createDatabase() {
    var ret = testDB->update("CREATE DATABASE testdb");
    handleUpdate(ret, "Create database testdb");
}

function dropDatabase() {
    var ret = testDB->update("DROP DATABASE testdb");
    handleUpdate(ret, "Drop Database testdb");
}

function handleUpdate(sql:UpdateResult|error returned, string message) {
    if (returned is sql:UpdateResult) {
        io:println(message + " status: " + returned.updatedRowCount);
    } else {
        io:println(message + " failed: " + <string>returned.detail().message);
    }
}
```
**OUTPUT**

Create database testdb status: 1
Create bittypes table status: 0
Update bittypes table failed: execute update failed: **Data truncation: Data too long for column 'bitVal2' at row 1**
Drop Database testdb status: 1

**Affected Versions:**

Biruntha:
## Purpose
Add support to insert byte[] as a direct parameter in update operation

Resolves #14710 
Resolves #15400

## Samples
```
import ballerina/io;
import ballerina/sql;
import ballerinax/jdbc;

jdbc:Client testDB = new({
    url: "jdbc:mysql://localhost:3306/testdb",
    username: "root",
    password: "root",
    dbOptions: { useSSL: false }
});

public function main() {
    // Create a table using the `update` remote function. If the DDL
    // statement execution is successful, the `update` remote function returns 0.
    io:println("Creating table and procedures:");
    var ret = testDB->update("CREATE TABLE testdb.bittypes(
        row_id INT,
        bit_val_1 BIT(1),
        bit_val_2 BIT(64),
        PRIMARY KEY (row_id)
    )");
    handleUpdate(ret, "Create bittypes table");

    ret = testDB->update("INSERT INTO testdb.bittypes(
                                row_id, bit_val_1, bit_val_2
                            ) VALUES (
                                1, b'1', b'01110' )");
    handleUpdate(ret, "Update bittypes table");

    // Insert byte[] as a sql:Parameter.
    int rowId = 2;
    boolean bitValue1 = true;
    byte[] bitValue2 = [1, 2];
    sql:Parameter para1 = { sqlType: sql:TYPE_INTEGER, value: rowId };
    sql:Parameter para2 = { sqlType: sql:TYPE_BOOLEAN, value: bitValue1};
    sql:Parameter para3 = { sqlType: sql:TYPE_BINARY, value: bitValue2};

    ret = testDB->update("INSERT INTO testdb.bittypes(row_id, bit_val_1, bit_val_2
                        ) VALUES (?, ?, ?)", para1, para2, para3);
    handleUpdate(ret, "Update bittypes table using sql:Parameter.");

    // Insert byte[] as a direct parameter
    rowId = 3;
    bitValue2 = [1, 2, 4];
    ret = testDB->update("INSERT INTO testdb.bittypes(row_id, bit_val_1, bit_val_2
                        ) VALUES (?, ?, ?)", rowId, bitValue1, bitValue2);
    handleUpdate(ret, "Update bittypes table by passing as direct parameter.");

    //Create the stored procedure with IN/OUT parameters.
    ret = testDB->update("CREATE PROCEDURE TestBitOutParam (IN id INT, OUT bitVal1 BOOLEAN, OUT bitVal2 BIT(64))
                  READS SQL DATA
                  BEGIN
                  SELECT bit_val_1 INTO bitVal1 from bittypes WHERE row_id = id;
                  SELECT bit_val_2 INTO bitVal2 from bittypes WHERE row_id = id;
                  END");

    handleUpdate(ret, "Stored procedure with IN/OUT param creation");

    sql:Parameter id = { sqlType: sql:TYPE_INTEGER, direction: sql:DIRECTION_IN, value: 2 };
    sql:Parameter bitVal1 = { sqlType: sql:TYPE_BOOLEAN, direction: sql:DIRECTION_OUT };
    sql:Parameter bitVal2 = { sqlType: sql:TYPE_BINARY, direction: sql:DIRECTION_OUT };

    var retCall = testDB->call("{call TestBitOutParam (?,?,?)}",(), id, bitVal1, bitVal2);

    if (retCall is ()|table<record {}>[]) {
        io:println("Call operation with OUT params successful");
        io:println("=============bitVal1============= ", bitVal1.value);
        io:println("=============bitVal2============= ", bitVal2.value);
    } else {
        io:println("Stored procedure call failed: " + <string>retCall.detail().message);
    }

    ret = testDB->update("DROP PROCEDURE TestBitOutParam");
    handleUpdate(ret, "Drop stored procedure TestBitOutParam");
}

// Function to handle the return value of the `update` remote function.
function handleUpdate(sql:UpdateResult|error returned, string message) {
    if (returned is sql:UpdateResult) {
        io:println(message + " status: " + returned.updatedRowCount);
    } else {
        io:println(message + " failed: " + <string>returned.detail().message);
    }
}
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -995,6 +997,14 @@ public static String getSQLType(BType value) {
             return Constants.SQLDataTypes.BOOLEAN;
         case TypeTags.DECIMAL_TAG:
             return Constants.SQLDataTypes.DECIMAL;
+        case TypeTags.ARRAY_TAG:
+            if (value instanceof BArrayType) {
+                if (((BArrayType) value).getElementType().getTag() == TypeTags.BYTE_TAG) {
```
Manuri: Can't we merge the two if conditions?

Also I think the exception should be thrown in the else block of that if condition. This error should be more specific such that it indicates that the problem is passing an array of a type other than byte.

Also, can we add a negative test to verify this exception is thrown appropriately?

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/types/table/TableTest.java
```diff
@@ -1427,6 +1427,18 @@ public void testRemoveOp() {
...
+        BValue[] returns = BRunUtil.invoke(result, "testByteArrayInsertAsDirectParameter");
+        Assert.assertEquals(((BInteger) returns[0]).intValue(), 1);
+    }
+
+    @Test(groups = TABLE_TEST, description = "Check inserting byte array data as SqlParameter.")
+    public void testByteArrayInsertAsSqlParameter() {
+        BValue[] returns = BRunUtil.invoke(result, "testByteArrayInsertAsSqlParameter");
+        Assert.assertEquals(((BInteger) returns[0]).intValue(), 1);
+    }
+
```
Manuri: Is there a possibility to make these two tests a part of an existing test? Do we have a specific reason to make these separate tests?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
Biruntha: I have tried to add negative test to check the exception is thrown by adding int[] into database as direct parameter.

but it gives compilation error as,
```
incompatible types: expected 'string|int|boolean|float|decimal|byte[]|ballerina/sql:Parameter', found 'int[]'
```

tests/ballerina-unit-test/src/test/java/org/ballerinalang/test/types/table/TableTest.java
Biruntha: Moved testByteArrayInsertAsSqlParameter test case to an existing test case. 
Changed other test to do the sql operation in the existing db but keeping it as separate test, since there are no test case written by passing params as direct params .

Biruntha: Moved to part of an existing test.
https://github.com/ballerina-platform/ballerina-lang/pull/15410/commits/4c5751d7c86e5153c50efc46c775120e159d4101

Biruntha:
Resolves #15400

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -995,6 +997,13 @@ public static String getSQLType(BType value) {
             return Constants.SQLDataTypes.BOOLEAN;
         case TypeTags.DECIMAL_TAG:
             return Constants.SQLDataTypes.DECIMAL;
+        case TypeTags.ARRAY_TAG:
+            if ((value instanceof BArrayType) && ((BArrayType) value).getElementType().getTag() == TypeTags.BYTE_TAG) {
+                return Constants.SQLDataTypes.BINARY;
+            } else {
+                throw new BallerinaException("Found unsupported array data type as direct value and supported array " +
```
Manuri:
```suggestion
                throw new BallerinaException("Array data type as direct value is supported only with byte type elements, use sql:Parameter " +
```

Can we have something like this?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -995,6 +997,13 @@ public static String getSQLType(BType value) {
             return Constants.SQLDataTypes.BOOLEAN;
         case TypeTags.DECIMAL_TAG:
             return Constants.SQLDataTypes.DECIMAL;
+        case TypeTags.ARRAY_TAG:
+            if ((value instanceof BArrayType) && ((BArrayType) value).getElementType().getTag() == TypeTags.BYTE_TAG) {
```
Manuri: I think we don't need to run instanceof here. If the tag is ARRAY_TAG then it should be an instance of BArrayType.

Biruntha: Without this instanceof check i got the following error.
```
Unchecked/unconfirmed cast from org.ballerinalang.model.types.BType to org.ballerinalang.model.types.BArrayType in org.ballerinalang.database.sql.SQLDatasourceUtils.getSQLType(BType) [org.ballerinalang.database.sql.SQLDatasourceUtils] At SQLDatasourceUtils.java:[line 1001] BC_UNCONFIRMED_CAST
```

So, now updated the spotbug rule of sql module with this https://github.com/ballerina-platform/ballerina-lang/pull/15410/commits/cee73a092e226aeb37d10021adfb885100a9e44a
and removed the  instanceof check

anupama-pathirage:
It seems this PR hasn't addressed the data retrieval part. When selecting data, it should be possible to assign bit(N) type data into a ballerina byte[] instead of boolean. So this PR is not resolving https://github.com/ballerina-platform/ballerina-lang/issues/14710 fully. @Manuri ? WDYT?

Manuri:
According to @Biruntha 's findings, in case of MySQL this is due to the behavior in JDBC driver which outputs the column type and the class as Boolean even when the actual type is byte array. However, she observed that, this occurs in MySQL 8 whereas in MySQL 5 mapping to byte[] worked as expected.

@Biruntha As discussed offline could you put the observations relevant to H2? If H2 also displays the same behavior as MySQL 8, we need to debug and see if this is due to the same reason. 

Also, if that is the case, for both MySQL 8 and H2 we need to check whether there are reported issues for this, since this looks like a bug. And check a bit whether there is any workaround we can apply by obtaining any other available meta data.

Biruntha:
This is the observations in H2 db. Tested with h2-1.4.197.jar version.

Here in [1], I have tried to insert byte array into bittypes table and the result outcome of ColumnType, ClassName, ColumnTypeName values from  ResultSetMetaData as in [2]. As per this output, byte array is going as hex X'01020528' to the backend. Here, I have validated that X'01020528' is equalant to the inserted [1, 2, 5, 40] byte array by converting hex to byte array. 

Here also when try to get the classname of the BIT(> 1) cloumn (bitVal2), it return as Boolean. 

[1]
```
import java.io.ByteArrayInputStream;
import java.sql.Connection;
import java.sql.DriverManager;
import java.sql.PreparedStatement;
import java.sql.ResultSet;
import java.sql.ResultSetMetaData;
import java.sql.Statement;
import java.util.Arrays;

public class H2jdbcCreateDemo {
    // JDBC driver name and database URL
    static final String JDBC_DRIVER = "org.h2.Driver";
    static final String DB_URL = "jdbc:h2:~/test";

    //  Database credentials
    static final String USER = "sa";
    static final String PASS = "";

    public static void main(String[] args) {
        Connection con;
        Statement stmt;
        ResultSet rs;
        try {
            Class.forName(JDBC_DRIVER);
            con = DriverManager.getConnection(DB_URL, USER, PASS);

            String sql = "INSERT INTO bittypes(id, bitVal1, bitVal2) VALUES (5, true, ?)";
            PreparedStatement pstmt = con.prepareStatement(sql);

            byte[] buffer = new byte[] {1, 2, 5, 40};
            pstmt.setBinaryStream(1, new ByteArrayInputStream(buffer), buffer.length);
            System.out.println("====PreparedStatement=====");
            System.out.println(pstmt);
            pstmt.executeUpdate();
            pstmt.close();


            stmt=con.createStatement();
            rs = stmt.executeQuery("SELECT * FROM bittypes");
            ResultSetMetaData md = rs.getMetaData();
            int counter = md.getColumnCount();
            String colName[] = new String[counter];
            System.out.println("\n The column names are as follows:");
            for (int loop = 1; loop <= counter; loop++) {
                colName[loop-1] = md.getColumnLabel(loop);
                System.out.println("############ColumnName######### " + colName[loop-1]);
                System.out.println("ColumnType " + md.getColumnType(loop));
                System.out.println("ClassName " + md.getColumnClassName(loop));
                System.out.println("ColumnTypeName " + md.getColumnTypeName(loop));
            }
        } catch(Exception e) {
            e.printStackTrace();
        }
    }
```
[2]
```
====PreparedStatement=====
prep0: INSERT INTO bittypes(id, bitVal1, bitVal2) VALUES (5, true, ?) {1: X'01020528'}

 The column names are as follows:
############ColumnName######### ID
ColumnType 4
ClassName java.lang.Integer
ColumnTypeName INTEGER
############ColumnName######### BITVAL1
ColumnType 16
ClassName java.lang.Boolean
ColumnTypeName BOOLEAN
############ColumnName######### BITVAL2
ColumnType 16
ClassName java.lang.Boolean
ColumnTypeName BOOLEAN
```

Biruntha:
This is the observations for mysql server version 8.0.16 with mysql-connector-java-8.0.16.jar. For the following sample[1], I'm getting output as [2] and it is returning the value for bitVal2 column as boolean. So inorder to validate this further, again I have try out this java sample mentioned in https://github.com/ballerina-platform/ballerina-lang/issues/14710. It's again returning the class name as Boolean [3].

[1]
```
import ballerina/io;
import ballerina/mysql;
import ballerina/sql;
import ballerinax/jdbc;

jdbc:Client testDB = new({
	url: "jdbc:mysql://127.0.0.1:6603/testdb",
	username: "root",
	password: "876",
	dbOptions: { useSSL: false }
});

    int id = 1;
    boolean bitVal1 = true;
    byte[] bitVal2 = [1, 2];
    var ret = testDB->update("INSERT INTO testdb.bittypes(id, bitVal1, bitVal2
                        ) VALUES (?, ?, ?)", id, bitVal1, bitVal2);
    handleUpdate(ret, "Update bittypes table");

    var selectRet = testDB->select("Select id, bitVal1, bitVal2 from testdb.bittypes", ());
    json j = getJsonConversionResult(selectRet);
    string jsonStr = io:sprintf("%s", j);
    io:println(jsonStr);
}

function handleUpdate(sql:UpdateResult|error returned, string message) {
    if (returned is sql:UpdateResult) {
        io:println(message + " status: " + returned.updatedRowCount);
    } else {
        io:println(message + " failed: " + <string>returned.detail().message);
    }
}

function getJsonConversionResult(table<record {}>|error tableOrError) returns json {
    json retVal = {};
    if (tableOrError is table<record {}>) {
        var jsonConversionResult = json.convert(tableOrError);
        if (jsonConversionResult is json) {
            // Converting to string to make sure the json is built before returning.
            _ = jsonConversionResult.toString();
            retVal = jsonConversionResult;
        } else {
            retVal = {"Error" : <string>jsonConversionResult.detail().message};
        }
    } else {
        retVal = {"Error" : <string>tableOrError.detail().message};
    }
    return retVal;
}
```
[2]
```
Update bittypes table status: 1
[{"id":1, "bitVal1":true, "bitVal2":true}]
```
[3]
```
The column names are as follows:
############ColumnName######### id
ColumnType 4
ClassName java.lang.Integer
ColumnTypeName INT
############ColumnName######### bitVal1
GetBoolean-------true
GetBytes-------[1]
ColumnType -7
ClassName java.lang.Boolean
ColumnTypeName BIT
############ColumnName######### bitVal2
ColumnType -7
ClassName java.lang.Boolean
ColumnTypeName BIT
```

Manuri:
@Biruntha So it seems, this is due to the behavior in respective drivers. 
As mentioned offline as well, we can do is,

1. Check whether there are reported bugs in relevant database/driver bug trackers/forums, if there aren't any, we should report bugs if possible
2. Ask regarding this behavior in their forums/git repositories
3. Regardless of this being an issue in database/driver/ballerina code, ballerina developers may face this issue so we need to create a separate git issue explaining the problem and the reason we found with workarounds if any.
4. Write a separate test case for this and keep that disabled with a link to the aforementioned git issue.

PR #15427, opened 2019-05-28
Validate in-object definitions of referenced functions

### LINKED ISSUE #13179: `is` produces incorrect results for functions when the param/return type is a union
Edited 23/06/2019

**Description:**
Case I:
```ballerina
import ballerina/io;

function funcReturningString(string|int s) returns string  {
    return "string";
}

    any a = funcReturningString;
    io:println(a is function(string s) returns string);
}
```

Case II:
```ballerina
import ballerina/io;

function funcReturningString() returns string  {
    return "string";
}

    any a = funcReturningString;
    io:println(a is function() returns string|int);
}
```

Both the above print `false`.

**Affected Versions:**
0.990.2

### LINKED ISSUE #13930: Abstract object implementation chekcs only for function name
**Description:**
The concrete implementation of a abstract clients only checks for function names when referring an abstract object . They are not validated against the input parameters and the return types of the functions. Same issue exists with the abstract objects as well.

The below code doesn't give an compiler error, but it should as the function contains different input parameters and return types.
```ballerina
public type AbstractSQLClient abstract client object {
    public remote function update(string queryStr, int count) returns int|error;
};

public type Client client object {
    *AbstractSQLClient;

    public remote function update(string sqlQuery) returns string|error {
        return "test";
    }
};

```
Same issue is there with abstrct objects as well.
```ballerina
public type AbstractSQLObj abstract object {
    public function update(string queryStr, int count) returns int|error;
};

public type Obj object {
    *AbstractSQLObj;

    public function update(string sqlQuery) returns string|error {
        return "test";
    }
};
```

**Steps to reproduce:**

**Affected Versions:**

pubudu91:
Fixes #13930
Fixes #13179 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


pubudu91:
Marked this as WIP because the plan is to include the removal of support for outside method definition with this PR as well

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -691,12 +691,12 @@ private static boolean checkIsFunctionType(BType sourceType, BFunctionType targe
...
 
         for (int i = 0; i < source.paramTypes.length; i++) {
-            if (!isSameType(source.paramTypes[i], targetType.paramTypes[i])) {
+            if (!checkIsType(targetType.paramTypes[i], source.paramTypes[i], new ArrayList<>())) {
                 return false;
             }
         }
 
-        return isSameType(source.retType, targetType.retType);
+        return checkIsType(source.retType, targetType.retType, new ArrayList<>());
```
MaryamZi: These changes would fix https://github.com/ballerina-platform/ballerina-lang/issues/13179 too right? Shall we verify and add tests for the cases mentioned in the issue too?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -676,6 +676,9 @@ error.redeclared.type.reference=\
 error.redeclared.function.from.type.reference=\
   redeclared symbol ''{0}'': trying to copy a duplicate function through referenced type ''{1}''
 
+error.referred.function.signature.mismatch=\
+  function signatures mismatched: expected ''{0}'', found ''{1}''
```
MaryamZi: Shouldn't this be 
```suggestion
  mismatched function signatures: expected ''{0}'', found ''{1}''
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
pubudu91: Added with 4a207ce

pubudu91:
Closing this PR in favour of PR #16002

PR #15507, opened 2019-06-04
Enable Streams Values and Type unit tests

### LINKED ISSUE #15473: Cannot run not-Nil returning function if that function body contains a lambda function reference
Please consider the following code

```ballerina
import ballerina/io;

public function main() {
    io:println("Test");
}

function testStreamPublishingAndSubscriptionForAssignableTupleTypeStream(string s1) returns int {
    test("lambda", zzz);
    return 0;
}

function test(string val, function (string) x) {
    x.call(val);
}

function zzz((string) val) {
    io:println(val);
}
```

Above code will be compiled. But cannot be run. Following will be thrown

java.lang.VerifyError: Inconsistent stackmap frames at branch target 223
    Exception Details:
      Location:
 ```
test.testStreamPublishingAndSubscriptionForAssignableTupleTypeStream(Lorg/ballerinalang/jvm/Strand;Ljava/lang/String;Z)J @223: aload_0
      Reason:
        Type null (current frame, locals[5]) is not assignable to integer (stack map, locals[5])
      Current Frame:
        bci: @28
        flags: { }
        locals: { 'org/ballerinalang/jvm/Strand', 'java/lang/String', long, long_2nd, null, null, integer }
        stack: { integer }
      Stackmap Frame:
        bci: @223
        flags: { }
        locals: { 'org/ballerinalang/jvm/Strand', 'java/lang/String', long, long_2nd, null, integer }
        stack: { }
```

### PR DESCRIPTION
**Pull request by gimantha**
## Purpose
This PR enables Streams Values and Types related unit tests in Jballerina module. Also fixes #15473 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples

### CODE REVIEW

File: bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/StreamValue.java [hunk 1]
```diff
@@ -121,14 +125,22 @@ public void publish(RefValue data) {
      * @param functionPointer represents the function pointer reference for the function to be invoked on receiving
      *                        messages
      */
-    public void subscribe(Consumer functionPointer) {
+    public void subscribe(FPValue<Object[], Object> functionPointer) {
+        BType[] parameters = ((BFunctionType) functionPointer.type).paramTypes;
+        int lastArrayIndex = parameters.length - 1;
+        if (!TypeChecker.isAssignable(constraintType, ((BFunctionType) functionPointer.getType())
```
warunalakshitha: IMO we can replace this by checkIsLikeType method. So we do not need isAssignable .

warunalakshitha: We also need to update those methods with stream type.

gimantha: @warunalakshitha CheckIsType is used to check if the given value can be assigned to a given type.. In this method we check two types.

gimantha: Removed redundant code

PR #15523, opened 2019-06-05
Add future.cancel() and fix worker send issue

vinok88:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/Strand.java
```diff
@@ -52,6 +52,7 @@
     public boolean blockedOnExtern;
     public ChannelDetails[] channelDetails;
     private Map<String, Object> globalProps;
+    public boolean doCancel;
```
Kishanthan: just "cancel" ?

vinok88: Done. Thanks.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/FutureValue.java
```diff
@@ -72,4 +75,18 @@ public void stamp(BType type, List<TypeValuePair> unresolvedValues) {
...
+     public boolean isDone() {
+         return isDone;
+     }
+
+     public boolean isCancelled() {
+         return isCancelled;
+     }
+
+     public boolean cancel() {
+         this.strand.cancel = true;
```
Kishanthan: why we need the strand also to have the cancel flag aprt from FutureValue having it?

vinok88: FutureValue is the reference to a strand that others can use it to cancel the strand. Within functions, strand is the one that carry this meta info. We need this info within functions. A strand cannot access it's own future. Hence this cancel flag is used to track that there is a cancel requested.

PR #15594, opened 2019-06-09
[IntelliJ Plugin] Fix plugin gradle build for offline mode

NipunaRanasinghe:
## Purpose
This PR adds a minor change which fixes the offline mode gradle build of ballerina intellij plugin.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tool-plugins/intellij/build.gradle
+++ b/tool-plugins/intellij/build.gradle
@@ -59,30 +59,6 @@ intellij {
     // alternativeIdePath localIdePath
 }
 
-apply plugin: 'de.undercouch.download'
-task downloadPsiViewerPlugin() {
-    download {
-        src 'https://plugins.jetbrains.com/plugin/download?updateId=31087'
-        dest new File("${buildDir}/tmp/plugins/", 'PsiViewer.jar')
-        onlyIfNewer true
-    }
-}
-
-apply plugin: "jacoco"
-jacoco {
-    toolVersion = "0.8.1"
-    reportsDir = file("$buildDir/customJacocoReportDir")
-}
-
-task copyPsiViewerPluginToSandBox(type: Copy) {
-    from "${buildDir}/tmp/plugins/PsiViewer.jar"
-    into "${buildDir}/idea-sandbox/plugins/"
-}
-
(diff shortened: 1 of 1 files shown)


Shan1024:
@NipunaMarcus - Why are we removing these tasks? `runIde` task should not have any impact on building the plugin IIRC. Unless we specifically invoke that task (to test), it will not have any impact on building the plugin offline.

tool-plugins/intellij/build.gradle
```diff
@@ -59,30 +59,6 @@ intellij {
     // alternativeIdePath localIdePath
 }
 
-apply plugin: 'de.undercouch.download'
-task downloadPsiViewerPlugin() {
```
Shan1024: Why are we removing these tasks? `runIde` task should not have any impact on building the plugin IIRC. Unless we specifically invoke that task (to test), it will not have any impact on building the plugin offline.

manuranga: This downloads beginning at every build even if we compile a single jar. This is because it's written as a generic task not a download task. So you should be able to fix it by 
```groove
task downloadPsiViewerPlugin(type: Download) {
    src 'https://plugins.jetbrains.com/plugin/download?updateId=31087'
    dest new File("${buildDir}/tmp/plugins/", 'PsiViewer.jar')
}

```

NipunaRanasinghe: When trying to build in offline mode, build fails with the following exception.

```gradle
FAILURE: Build failed with an exception.

* Where:
Build file '/home/nino/Desktop/ballerina/tool-plugins/intellij/build.gradle' line: 100

* What went wrong:
A problem occurred evaluating project ':ballerina-intellij-idea-plugin'.
> Unable to download file 'https://plugins.jetbrains.com/plugin/download?updateId=31087' in offline mode.
```

NipunaRanasinghe: @manuranga Your fix works. :) Thanks for the suggestion!

PR #15689, opened 2019-06-12
Introduce non-blocking db interactions from Ballerina worker threads' POV

### LINKED ISSUE #8593: Implement and do a performance test for having intermediate thread pool between ballerina workers and db connection pool
**Description:**
At the moment almost all connectors are non-blocking, but DB connectors aren't, because there are only synchronous libraries available.

This means that we need enough threads so that a worker can be blocked while waiting on a database connection.

What if we allocated separate threads for DB pools? 

If I define a MySQL connection with a poolsize of 50, there will be 50 threads dedicated to the MySQL connection that are separate from the normal worker threads. When I call the DB, the worker thread will pass the request over to one of these threads. If they are all busy then it is the same as if the pool is busy (since each thread can have its own connection to the database as NumThreads==NumConnections in the pool). 

This way we can have a much more limited number of worker threads all asynchronous and only use larger pools of threads specifically to handle the blocking case of a DB.

- Quoted from "Making DB connections non-blocking" @ballerina-dev by Paul Fremantle

Manuri:
## Purpose
Fixes #8593

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


Manuri:
Closing until design changes are further discussed.

PR #15698, opened 2019-06-12
Enable some IO tests

wggihan:
## Purpose
$title.


stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/HasNextTextRecord.java
```diff
@@ -110,7 +76,7 @@ public static Object hasNext(Strand strand, ObjectValue channel) {
             Register register = EventRegister.getFactory().register(hasNextEvent, HasNextTextRecord::getResponse);
             eventContext.setRegister(register);
             register.submit();
-            return null;
+            return false;
```
chamil321: We can remove this return as function return from #81

Issue #15429, opened 2019-05-29
Use diagram libs from distribution


kaviththiranga:
Currently VSCode extension packs diagram library inside vsix itself. But with https://github.com/ballerina-platform/ballerina-lang/pull/15412, diagram library is now getting shipped in the distribution. We need to change VSCode extension to pick them from distribution and avoid bundling it in the vsix.

kaviththiranga:
fixed with https://github.com/ballerina-platform/ballerina-lang/pull/15727

PR #15412, opened 2019-05-24
Add composer library resources to the ballerina distribution

NipunaRanasinghe:
## Purpose
This PR updates gradle and maven build configurations to add composer library resources to the ballerina distribution.

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Required Balo version update
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #15727, opened 2019-06-13
Use composer libs from Ballerina home

kaviththiranga:
Previously composer libs were packed in vscode extension.
With https://github.com/ballerina-platform/ballerina-lang/pull/15412,
they have been added to Ballerina distribution. This is to change VScode
extension along with that. Includes some refactoring to webview utils
as well.

Fixes: https://github.com/ballerina-platform/ballerina-lang/issues/15429

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tool-plugins/vscode/.vscodeignore
+++ b/tool-plugins/vscode/.vscodeignore
@@ -6,6 +6,7 @@ src/
 target/
 docs/
 extractedDistribution/
+resources/composer
 .gitignore
 tsconfig.json
 tslint.json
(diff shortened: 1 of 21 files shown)


tool-plugins/vscode/.vscodeignore
```diff
@@ -6,6 +6,7 @@ src/
 target/
 docs/
 extractedDistribution/
+resources/composer
```
roonyh: Do we need this? If the assets are coming from the distribution will we need to create a resources/compoer directory?

kaviththiranga: @roonyh good catch, yeah we don't need it. I guess I needed it because composer files were already there from existing builds. But if we do a clean build this shouldn't be needed.

roonyh: Got it 👍

PR #15768, opened 2019-06-14
Use BServiceType for services and fix using services as values in annotations

MaryamZi:
## Purpose
$title. 

Note: Changes to AnnotationDesugar.java, are temporary and will be changed with the annotation change.


compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_type_gen.bal
```diff
@@ -372,6 +372,39 @@ function createObjectType(jvm:MethodVisitor mv, bir:BObjectType objectType, bir:
     return;
 }
 
+# Create a runtime type instance for the service.
+#
+# + mv - method visitor
+# + objectType - object type
+# + typeDef - type definition of the service
+# + pkgName - name of the module the service belongs to
+function createServiceType(jvm:MethodVisitor mv, bir:BObjectType objectType, bir:TypeDef typeDef, string pkgName) {
```
lankavitharana: we can unify this with object type right? only change is the new instruction and the init, (better to do that with a separate PR)

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_type_gen.bal
```diff
@@ -590,6 +623,10 @@ function loadType(jvm:MethodVisitor mv, bir:BType? bType) {
     } else if (bType is bir:BTypeDesc) {
         typeFieldName = "typeTypedesc";
     }  else if (bType is bir:BServiceType) {
+        if (getTypeFieldName(bType.oType.name.value) != "$type$service") {
```
lankavitharana: even here, i think it's better to unify(like write default service type as an object type, so you don't need to differentiate here, anyway do that refactoring later with a different PR)

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
```diff
@@ -107,6 +102,14 @@ protected void rewritePackageAnnotations(BLangPackage pkgNode) {
             initFunction.body.stmts.add(index++, stmt);
         }
 
+        // Handle service annotations
```
lankavitharana: here the order matters right?, better to put a comment specifying that, and the order

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
```diff
@@ -223,4 +260,28 @@ private void initAnnotation(BLangAnnotationAttachment attachment, BLangSimpleVar
         indexAccessNode.type = annotationMapEntryVar.symbol.type;
         assignmentStmt.varRef = indexAccessNode;
     }
+
+    private int calculateIndex(List<BLangStatement> statements, BLangService service) {
+        for (int i = 0; i < statements.size(); i++) {
```
lankavitharana: i think u can simplify these as well

Issue #15780, opened 2019-06-14
Record type referencing does not work in jBallerina


Kishanthan:
```ballerina
type Person record {
    string name;
    int age;
};

type Student record {
    *Person;
    float gpa;
};
```

Above doesn't work and jvm compiler backend fails with the an error.

```
error: cannot find function definition for : <init> {}
	at compiler_backend_jvm:0.0.0:isBIRFunctionExtern(jvm_package_gen.bal:78)
	   compiler_backend_jvm:0.0.0:genStaticCall(jvm_terminator_gen.bal:330)
	   compiler_backend_jvm:0.0.0:genBuiltinTypeAttachedFuncCall(jvm_terminator_gen.bal:322)
	   compiler_backend_jvm:0.0.0:genCall(jvm_terminator_gen.bal:308)
	   compiler_backend_jvm:0.0.0:genCallTerm(jvm_terminator_gen.bal:244)
	   compiler_backend_jvm:0.0.0:genTerminator(jvm_terminator_gen.bal:44)
	   compiler_backend_jvm:0.0.0:generateBasicBlocks(jvm_method_gen.bal:578)
	   compiler_backend_jvm:0.0.0:generateMethod(jvm_method_gen.bal:192)
	   compiler_backend_jvm:0.0.0:createRecordMethods(jvm_value_gen.bal:364)
	   compiler_backend_jvm:0.0.0:createRecordValueClass(jvm_value_gen.bal:353)
	   compiler_backend_jvm:0.0.0:generateValueClasses(jvm_value_gen.bal:44)
	   compiler_backend_jvm:0.0.0:generatePackage(jvm_package_gen.bal:128)
	   compiler_backend_jvm:0.0.0:generateJarBinary(main.bal:50)
```

warunalakshitha:
Once fix the issue lets revert the change in  #15782 in tests/jballerina-unit-test/src/test/resources/test-src/statements/ifelse/type-guard.bal file.

SupunS:
Object Referencing currently works, but does not copy the default-values of the expressions.

Need to figure out a way to get the default values..

PR #15782, opened 2019-06-14
Enable comment, ifelse and package import statement tests

Kishanthan:
$subject.

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/statements/ifelse/type-guard.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/statements/ifelse/type-guard.bal
@@ -367,7 +367,8 @@ type Person record {
 };
 
 type Student record {
-    *Person;
+    string name;
+    int age;
     float gpa;
 };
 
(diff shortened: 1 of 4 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/statements/ifelse/type-guard.bal
```diff
@@ -367,7 +367,8 @@ type Person record {
 };
 
 type Student record {
-    *Person;
+    string name;
+    int age;
```
Kishanthan: There is a L1 identified with this - https://github.com/ballerina-platform/ballerina-lang/issues/15780

tests/jballerina-unit-test/src/test/resources/test-src/statements/ifelse/type-guard.bal
```diff
@@ -367,7 +367,8 @@ type Person record {
 };
 
 type Student record {
-    *Person;
```
warunalakshitha: Do we need this change to test file?

Kishanthan: There is a L1 identified with this - #15780

PR #15841, opened 2019-06-19
Improve Ballerina SQL error

Biruntha:
## Purpose
Enhancing Ballerina SQL error

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/BatchUpdateStatement.java
```diff
@@ -89,6 +89,10 @@ public void execute() {
             // ignore a few failed commands in the batch and let the rest of the commands run if driver allows it.
             updatedCount = e.getUpdateCounts();
             processAndSetBatchUpdateResult(updatedCount, paramArrayCount);
+        } catch (SQLException e) {
+            context.setReturnValues(
+                    SQLDatasourceUtils.getSQLDatabaseError(context, e, "execute batch update failed: "));
+            checkAndObserveSQLError(context, e.getMessage());
```
anupama-pathirage: Why are we omitting `handleErrorOnTransaction` from this catch clause?

anupama-pathirage:
@Biruntha Shall we add test cases to this change?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
```diff
@@ -280,18 +280,35 @@ public static String createJDBCDbOptions(String propertiesBeginSymbol, String se
...
+                throwable.getMessage() != null ? throwable.getMessage() : Constants.APPLICATION_ERROR_MESSAGE;
         return getSQLConnectorError(context, messagePrefix + detailedErrorMessage);
     }
 
-    public static BError getSQLConnectorError(Context context, String detailedErrorMessage) {
-        BMap<String, BValue> sqlClientErrorDetailRecord = BLangConnectorSPIUtil
-                .createBStruct(context, Constants.SQL_PACKAGE_PATH, Constants.DATABASE_ERROR_DATA_RECORD_NAME,
-                        detailedErrorMessage);
-        return BLangVMErrors.createError(context, true, BTypes.typeError, Constants.DATABASE_ERROR_CODE,
+    public static BError getSQLConnectorError(Context context, String detailedErrorMessage, Object... values) {
```
anupama-pathirage: Remove this common method and move the respective logic to the `getSQLDatabaseError` and `getSQLApplicationError` methods directly. Having Object... and if checks within the method adds no value

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/CallStatement.java
```diff
@@ -120,10 +120,15 @@ public void execute() {
                 cleanupResources(resultSets, stmt, conn, !isInTransaction);
                 context.setReturnValues();
             }
+        } catch (SQLException e) {
```
anupama-pathirage: We cannot simply stop by checking SQLException at the outer layer. There are multiple places in [1] where SQLExceptions are re-thrown as BallerinaException. So we need to change them as well.

[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/AbstractSQLStatement.java

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDatasourceUtils.java
Biruntha: This getSQLConnectorError method was used in this line [1]. I haven't remove this method in this https://github.com/ballerina-platform/ballerina-lang/pull/15841/commits/b3309499bc1a5fd8905e6b505b094ad552e25f1d . But i have moved database error logic to getSQLDatabaseError method

[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/actions/Close.java#L56

Biruntha:
> @Biruntha Shall we add test cases to this change?

There are some issues while running the test cases. After Migration to jballerina is done, I'll add the test cases for this.

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/AbstractSQLStatement.java
```diff
@@ -427,28 +430,34 @@ protected void cleanupResources(Statement stmt, Connection conn, boolean connect
                 conn.close();
             }
         } catch (SQLException e) {
-            throw new BallerinaException("error in cleaning sql resources: " + e.getMessage(), e);
+            String errorMessage = "error in cleaning sql resources: ";
+            context.setReturnValues(SQLDatasourceUtils.getSQLDatabaseError(context, e, errorMessage));
+            checkAndObserveSQLError(context, errorMessage + e.getMessage());
```
Manuri: I don't think we should be doing this at this level. IMO we should rather propagate this to upper layer and handle at the concrete Statement (eg: SelectStatement, UpdateStatement) level. Otherwise it will be hard to manage. This applies to other relevant places as well.

May be for database modules, we should define two java exception types for the two Ballerina error types and throw them appropriately, then catch at Statement level and create Ballerina errors. WDYT?

PR #15843, opened 2019-06-19
Fix log-api module 

Bhashinee:
## Purpose
$subject.


pubudu91:
Should this be sent to master?

Issue #14352, opened 2019-03-13
Implement new Annotation syntax


hasithaa:
Coppied from the spec 2019r1

annots := annotation*
annotation := @ annot-tag-reference annot-value

Annotations provide structured metadata about a particular construct. Multiple annotations can be applied to a single construct. An annotation consists of a tag and a value.

annotation-decl :=
   metadata
   [public] [const] annotation [type-descriptor] annot-tag 
   [on annot-attach-points]
annot-tag := identifier
 
An annotation-decl declares an annotation tag. The type-descriptor specifies the type of the annotation tag. The type must be a subtype of one of the following three types: true, map<anydata>, map<anydata>[]. If the type-descriptor is omitted, then the type is true.

annot-tag-reference := qualified-identifier
annot-value := [mapping-constructor-expr]

An annot-tag-reference in an annotation must refer to an annot-tag declared in an annotation declaration. If the annotation includes a mapping-constructor-expr, then the value of the annotation is the mapping value resulting from evaluating the mapping-constructor-expr; otherwise the value is the boolean value true.

For every construct that has an annotation with a particular tag, there is an effective value for that annotation tag, which is constructed from all the annotations with that tag that were attached to that construct. The effective value belongs to the type of the annotation tag.

The type of the annotation tag constrains both the annotation value and the occurrence of multiple annotations with the same tag on a single construct as follows.
If the type of the annotation tag is T[] for some T, then the mapping-constructor-expr must be specified and its static type must be a subtype of T; multiple annotations are allowed; the effective value is a list of all annotations with the tag in the order specified.
Otherwise, multiple annotations with this tag are not allowed and the effective value for the tag is the value of the single annotation with the tag.
If the type of the annotation tag is true, then the mapping-constructor-expr must not be specified.
Otherwise, the mapping-constructor-expr must be specified and its static type must be a subtype of the type of the annotation tag.

If the annotation-decl for a tag specifies const, then a mapping-constructor-expr in annotations with that tag must be a const-expr and is evaluated at compile-time with the semantics of a const-expr. Otherwise, the mapping-constructor-expr is evaluated when the annotation is evaluated and the resulting value is automatically frozen.

An annotations applied to a module-level declaration is evaluated when the module is initialized. An annotation applied to a service constructor is evaluated when the service constructor is evaluated. An annotation occurring within a type descriptor in a block is evaluated when the type descriptor is resolved.

annot-attach-points := annot-attach-point (, annot-attach-point)*
[...]

PR #15884, opened 2019-06-20
Update annotation declaration syntax and introduce the annotation access expression

MaryamZi:
## Purpose
$title, according to the spec.

- Annotations - ["Annotations" section under metadata](https://htmlpreview.github.io/?https://raw.githubusercontent.com/ballerina-platform/ballerina-spec/v2019R1/lang/spec.html#metadata)
- Annotation Access Expression - ["Annotation access expression" section under Expressions](https://htmlpreview.github.io/?https://raw.githubusercontent.com/ballerina-platform/ballerina-spec/v2019R1/lang/spec.html#expressions) 

This PR also:
- introduces `invoke()` methods in BRunUtil to accept JVM values to run functions
- migrates mime, http utils to use JVM values instead of BValues

Note: .balx generating tasks/modules, where annotations are used, have been disabled since CodeGen (phase) does not work with these changes.

Initial implementation addressing https://github.com/ballerina-platform/ballerina-lang/issues/14352

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


MaryamZi:
@arukshani, @chamil321 - Can y'all please have a look at the changes to HTTP and MIME tests?

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_instruction_gen.bal
```diff
@@ -357,6 +359,18 @@ type InstructionGenerator object {
         self.storeToVar(binaryIns.lhsOp.variableDcl);
     }
 
+    function generateAnnotAccessIns(bir:BinaryOp binaryIns) {
+        self.loadVar(binaryIns.rhsOp1.variableDcl);
+        addBoxInsn(self.mv, binaryIns.rhsOp1.variableDcl.typeValue);
```
SupunS: may be.. might not need addBoxInsn(). just double check

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_type_gen.bal
```diff
@@ -583,7 +618,7 @@ function addObjectAtatchedFunctions(jvm:MethodVisitor mv, bir:BAttachedFunction?
 # + mv - method visitor
 # + attachedFunc - object attached function
 function createObjectAttachedFunction(jvm:MethodVisitor mv, bir:BAttachedFunction attachedFunc,
-                                        bir:BObjectType objType) {
+                                      bir:BObjectType objType, string pkgName) {
```
SupunS: pkgName is not used?

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_type_gen.bal
```diff
@@ -468,10 +470,43 @@ function createServiceType(jvm:MethodVisitor mv, bir:BObjectType objectType, bir
     mv.visitLdcInsn(typeDef.flags);
     mv.visitInsn(L2I);
 
-    // Initialize the object
-    mv.visitMethodInsn(INVOKESPECIAL, SERVICE_TYPE, "<init>",
-        io:sprintf("(L%s;L%s;I)V", STRING_VALUE, PACKAGE_TYPE),
-        false);
+    // initialize the object
+    if (strandIndex is int) {
```
SupunS: Shall we do this in two steps? - with second step only when required.
As in, extract this logic to a separate method. When you want set the annotation values, simply call the old method just like before, and then call the extracted method. So the initialization and setting annotation values happens in two steps.

Suggesting this to avoid having large methods that does multiple things..

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_instruction_gen.bal
MaryamZi: Yeah, this shouldn't be needed. Removed with cb6db58.

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_type_gen.bal
MaryamZi: Removed this and few other unnecessary pkgName params in type creation functions with cb6db58.

compiler/ballerina-backend-jvm/src/main/ballerina/compiler_backend_jvm/jvm_type_gen.bal
MaryamZi: Kept the existing function as is and introduced another function which calls the existing function and then does annotation related stuff - https://github.com/ballerina-platform/ballerina-lang/commit/cb6db589f44e95f1f94e298fca2f0d0e80800e19#diff-eaee420729d9b10c600ac8582e8a3a99R470

PR #15949, opened 2019-06-24
Add error reason const literal match support

rdhananjaya:
## Purpose

Allow error to be matched based on constant literal.

```ballerina
error <string, map<string>> err1 = error("Error Code", message = "Msg");
match err1 {
    var error(reason) => return reason;
    error("Error Code", message = m, ... var rest) => return "reason matched";
}
return "Default";
```

In this case `err1` will match the second error match pattern, binding to m = "Msg", and rest = {}


hasithaa:
package-lock.json is committed.

PR #15981, opened 2019-06-24
Refactor HttpClient usage

chamil321:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/http/src/main/ballerina/http/http_client.bal
```diff
@@ -20,17 +20,16 @@
 # generic HTTP actions implementation.
 #
 # + url - The URL of the remote HTTP endpoint
-# + config - The configurations associated with the HttpCaller
-public type HttpCaller client object {
```
pubudu91: Wasn't `HttpCaller` what we used for the service to interact with the client?

chamil321: That is the Caller object which is used in the services and it has no relation with this HttpClient. In order to avoid that confusion, we thought of replacing the HttpCaller with HttpClient. HttpClient is basically the simple client which wraps the java client connector. And it is the base object for other clients(Caching, Retry,...etc)

PR #15996, opened 2019-06-25
[WIP] Move implicit initial value check to runtime

### LINKED ISSUE #15885: Relax var length array rules
https://github.com/ballerina-platform/ballerina-spec/issues/216

grainier:
## Purpose
Purpose of this PR is to relax var length array rules, by move implicit initial value check to runtime. This allows initializing variable length arrays without filler values. However, it will panic if we try to create a sparse array without filler values.

Fixes #15885 

## Remarks
- https://github.com/ballerina-platform/ballerina-spec/issues/216
- https://github.com/ballerina-platform/ballerina-lang/issues/15885


bvm/ballerina-core/src/main/resources/MessagesBundle.properties
```diff
@@ -161,3 +161,4 @@ invalid.object.field.addition = invalid value for object field ''{0}'': expected
 dynamically.nested.transactions.are.not.allowed = dynamically nested transactions are not allowed
 cyclic.value.reference = ''{0}'' value has cyclic reference
 tuple.index.out.of.range = tuple index out of range: index: {0}, size: {1}
+illegal.array.insertion= cannot insert an element into index ''{0}'' of an array without filler values
```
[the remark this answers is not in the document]
MaryamZi: Not sure if this message conveys what we want to. The index alone doesn't matter here right? Since the issue arises based on a combination of the current size and the index?

Should we include something that indicates this is related to expanding the array?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValue.java
```diff
@@ -662,6 +662,13 @@ private void rangeCheck(long index, int size) {
         }
     }
 
+    private void fillerValueCheck(int index, int size) {
+        if (!TypeChecker.hasImplicitInitialValue(elementType) && (index > size)) {
```
rdhananjaya: `(index > size)` this is the consecutive insertion check right?
Shall we extract a variable or add a comment explaining this non consecutive add rule.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1295,4 +1298,127 @@ public boolean equals(Object otherPair) {
             return true;
         }
     }
+
+    /**
+     * Checks whether a given {@link Byte} has an implicit initial value or not.
```
MaryamZi: BType? Needs to be fixed in the param description too.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1295,4 +1298,127 @@ public boolean equals(Object otherPair) {
...
+     * @param type {@link Byte} to be analyzed.
+     * @return whether there's an implicit initial value or not.
+     */
+    public static boolean hasImplicitInitialValue(BType type) {
+        return hasImplicitInitialValue(type, new ArrayList<>());
+    }
+
+    private static boolean hasImplicitInitialValue(BType type, List<BType> unAnalyzedTypes) {
+        if (type == null) {
```
MaryamZi: Ideally, type shouldn't be null right? Do we know where they become null? Is it due to array types being set incorrectly?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1295,4 +1298,127 @@ public boolean equals(Object otherPair) {
...
+            return true;
+        }
+        if (type.getTag() < TypeTags.RECORD_TYPE_TAG) {
+            return true;
+        }
+        switch (type.getTag()) {
+            case TypeTags.TYPEDESC_TAG:
+            case TypeTags.STREAM_TAG:
```
MaryamZi: We had an issue with new-ing up streams, since the name was required. Shall we validate if an array of streams works as expected?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1295,4 +1298,127 @@ public boolean equals(Object otherPair) {
...
+                BTupleType tupleType = (BTupleType) type;
+                return tupleType.getTupleTypes().stream().allMatch(TypeChecker::hasImplicitInitialValue);
+            case TypeTags.UNION_TAG:
+                return analyzeUnionType((BUnionType) type);
+            default:
+                return false;
+        }
+    }
+
+    private static boolean analyzeUnionType(BUnionType type) {
```
MaryamZi: Shall we rename this to indicate we are checking for the existence of a filler value?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1295,4 +1298,127 @@ public boolean equals(Object otherPair) {
...
+        // Control reaching this point means there is only one type in the union.
+        return BTypes.isValueType(firstMember) && hasImplicitInitialValue(firstMember);
+    }
+
+    private static boolean analyzeRecordType(BStructureType type, List<BType> unanalyzedTypes) {
+        if (unanalyzedTypes.contains(type)) {
+            return true;
+        }
+        unanalyzedTypes.add(type);
+        return type.getFields().values().stream().allMatch(f -> hasImplicitInitialValue(f.type, unanalyzedTypes));
```
MaryamZi: According to the spec, for mapping types the filler value is `{ }` if that is a valid constructor for the type.

Does that mean we need to consider records with fields with default values, even though they have no filler values, as records with a filler value?

```ballerina
type Bar object {
    function __init(string s) {
        // ....
    }
};

type Bar record {
    Foo f = new("test");
};
```
Since `{}` is a valid constructor for `Bar`, shouldn't we allow `Bar[]`

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1295,4 +1298,127 @@ public boolean equals(Object otherPair) {
...
+    }
+
+    private static boolean analyzeObjectType(BObjectType type) {
+        if (type.getTag() == TypeTags.SERVICE_TAG) {
+            return false;
+        }
+        if (type.getTag() == TypeTags.OBJECT_TYPE_TAG) {
```
MaryamZi: Do we need this if check? We can assume `BObjectType` has the tag `OBJECT_TYPE_TAG`, if it is not `SERVICE_TAG` right?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1295,4 +1298,127 @@ public boolean equals(Object otherPair) {
...
+        return false;
+    }
+
+    private static boolean analyzeFiniteType(BFiniteType type) {
+        // Has NIL element as a member.
+        if (type.valueSpace.stream().anyMatch(Objects::isNull)) {
+            return true;
+        }
+        // All of the element are from same type. And elements contains implicit initial value.
+        if (type.valueSpace.isEmpty()) {
```
MaryamZi: Is this ever satisfied? Also not sure if the comment is valid?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValue.java
MaryamZi: Shall we rename `hasImplicitInitialValue()` also to `hasFillerValue()`

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array_fill_test.bal
```diff
@@ -269,3 +269,28 @@ function testSingletonTypeArrayFill(int index) returns One[] {
     ar[index] = 1;
     return ar;
 }
+
+function testIllegalArrayInsertion(int index) returns (int|string|boolean)[] {
```
MaryamZi: Shall we update these tests to define the index within the function instead of passing it as an argument?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
pubudu91: `Bar[]` is ok in my view. The following was allowed before as well.
```ballerina
type Foo object {
    string f;

    function __init() {
        self.f = "foo";
    }
};
```
`Foo[]` was allowed since `Foo` has a no-arg `__init()` method.

MaryamZi: Yeah, for objects the same rule regarding the no-arg init still applies. My point is that for records we may also need to check for default values?

rdhananjaya: +1 for allowing Bar[]
When this check was first implemented, rec field default values were not fully supported (I think we only had default literals or something) , that's why it's not checked here.
Is record field default value support completed now?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
pubudu91: At runtime, no.. There wouldn't be empty value spaces. But even in compile time, this should've thrown an exception, instead of simply returning false IMO.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1295,4 +1298,127 @@ public boolean equals(Object otherPair) {
...
+        Object firstElement = type.valueSpace.iterator().next();
+        boolean sameType = type.valueSpace.stream().allMatch(value -> value.getClass() == firstElement.getClass());
+        if (!sameType) {
+            return false;
+        }
+        if (firstElement instanceof String) {
+            // check empty string for strings, and 0.0 for decimals
+            return containsElement(type.valueSpace, "\"\"");
+        } else if (firstElement instanceof Byte
+                || firstElement instanceof Integer
```
pubudu91: We don't need to check for `Byte` right? IIRC, the Ballerina `byte` is represented by `int` in Java.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1295,4 +1298,127 @@ public boolean equals(Object otherPair) {
...
+                || firstElement instanceof Long) {
+            return containsElement(type.valueSpace, "0");
+        } else if (firstElement instanceof Float
```
pubudu91: Again, no need to check for `Float` I think

grainier:
@hasithaa, @rdhananjaya, @MaryamZi I've fixed the PR according to the above suggestions and sent a  new PR to ballerina-platform:jballerina [#16118]. Hence, closing this PR.

Issue #16030, opened 2019-06-26
Function Variable Scoping


Anoukh:
**Description:**
Given a set of function parameters, all previous parameter symbols should be visible to the expression that is being evaluated for a default value of a defaultable parameter.

```ballerina
function test(int a, int b = 10, int c = a*2+b, int d = 10) { // valid

}

function test(int a, int b = 10, int c = a*2+b+d, int d = 10) { // invalid, undefined symbol d for default expression used for parameter c

}
```

Shan1024:
Closing the issue since it is already fixed by #16030.

PR #16031, opened 2019-06-26
 Initial implementation of function arguments refactor

Anoukh:
## Purpose
This PR will support providing positional arguments for both required and defaultable parameters as well as providing named arguments for both required and defaultable parameters.

## Checklist
- [x] Provide required parameters using named arguments.
- [x] Provide defaultable parameters using positional arguments.
- [ ] The rest-arg is not restricted to supplying the part of the argument list that will be bound to a rest-param and its static type is not restricted to being an array type. If there is rest-arg, then no parameter defaults are added.
- [ ] Function parameter scoping need to be revisited. Related issue #16030. With this PR, forward referencing also works for a given defaultable parameter. It should be restricted.
- [ ] `org.ballerinalang.test.main.function.ArgumentParserPositiveTest` and `org.ballerinalang.test.main.function.MainFunctionsTest` need to be migrated to run in JVM.
- [x] `public` keyword support for individual parameters.


hasithaa:
This is already merged to https://github.com/ballerina-platform/ballerina-lang/tree/15374_lang_module_refactor branch

hasithaa:
This is fixed in master.

PR #16067, opened 2019-06-27
[Intellij Plugin] Detach plugin build from the gradle build

NipunaRanasinghe:
## Purpose
This PR proposes to detach the ballerina intelliJ plugin build from the ballerina gradle build due to the following reasons.

- With the diagram support, intellij plugin is using javafx libraries which are not available in the openJDK and so the plugin is now compilable only with the oracle JDK (Please note that this won't affect the plugin runtime since the IDEA runtime environment comes with the required javafx libraries)
- Eventually ballerina tooling components will have separate repositories in the ballerina platform organization, hence this decoupling will be required eventually.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -99,7 +99,7 @@ include(':jballerina-integration-test')
 // and 'mustRunAfter' of vs-code-pluggin
 
 include(':plugin-vscode')
-include(':ballerina-intellij-idea-plugin')
+// include(':ballerina-intellij-idea-plugin')
 include(':benchmarks')
 include(':ballerina-activemq-artemis')
 include(':ballerina-rabbitmq')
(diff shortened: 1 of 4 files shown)


tool-plugins/intellij/src/main/resources/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,407 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    # Copyright 2015 WSO2 Inc. (http://wso2.org)
```
Shan1024: Shall we change the year?

settings.gradle
```diff
@@ -98,7 +98,7 @@ include(':jballerina-integration-test')
 // and 'mustRunAfter' of vs-code-pluggin
 
 include(':plugin-vscode')
-include(':ballerina-intellij-idea-plugin')
+// include(':ballerina-intellij-idea-plugin')
```
Shan1024: If this detachment is permanent, better to remove these commented lines as well.

tool-plugins/intellij/src/main/resources/checkstyle/checkstyle.xml
NipunaRanasinghe: Sorry I've missed it. Changed :)

settings.gradle
NipunaRanasinghe: I left it intentionally in case we need to enable the plugin build. But now I also think its better to be removed. Done

PR #16093, opened 2019-06-28
Improve Ballerina SQL error

Biruntha:
## Purpose
Improve Ballerina SQL error.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/exceptions/DatabaseException.java
```diff
@@ -0,0 +1,76 @@
...
+ * under the License.
+ */
+package org.ballerinalang.database.sql.exceptions;
+
+import java.sql.SQLException;
+
+/**
+ * The exception to throw Ballerina SQL database related exceptions.
+ *
+ * @since 0.8.0
```
Manuri: This should be 1.0.0.

stdlib/database/sql/src/main/ballerina/sql/types.bal
```diff
@@ -206,6 +206,20 @@ public type UpdateResult record {|
 # The parameter passed into the operations.
 type Param string|int|boolean|float|decimal|byte[]|Parameter;
 
+public const DATABASE_ERROR_REASON = "{ballerina/sql}DatabaseError";
```
anupama-pathirage: I think reason should be `"{ballerina/jdbc}DatabaseError"`

stdlib/database/sql/src/main/ballerina/sql/types.bal
```diff
@@ -206,6 +206,20 @@ public type UpdateResult record {|
...
+public type DatabaseError error<DATABASE_ERROR_REASON, DatabaseErrorData>;
+
+public const APPLICATION_ERROR_REASON = "{ballerina/sql}ApplicationError";
```
anupama-pathirage: I think reason should be "{ballerina/jdbc}ApplicationError"

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/Constants.java
```diff
@@ -213,8 +213,11 @@
     public static final String SQL_FILE_DB_POSTFIX = "_FILE";
     public static final String SQL_MEMORY_DB_POSTFIX = "_MEMORY";
 
+    public static final String APPLICATION_ERROR_DATA_RECORD_NAME = "ApplicationErrorData";
     public static final String DATABASE_ERROR_DATA_RECORD_NAME = "DatabaseErrorData";
-    public static final String DATABASE_ERROR_MESSAGE = "SQL Error Occurred";
+    public static final String APPLICATION_ERROR_MESSAGE = "Application Error Occurred";
+    public static final String DATABASE_ERROR_MESSAGE = "Database Error Occurred";
+    public static final String APPLICATION_ERROR_CODE = "{ballerina/sql}ApplicationError";
```
anupama-pathirage: We need to change these constants too to `ballerina/jdbc`

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/BatchUpdateStatement.java
```diff
@@ -105,7 +107,31 @@ public Object execute() {
...
+            return SQLDatasourceUtils.getSQLDatabaseError(e, errorMessage + ": ");
+        } catch (DatabaseException e) {
+            String errorMessage = "execute batch update failed";
+            if (!isInTransaction) {
+                try {
+                    conn.rollback();
+                } catch (SQLException ex) {
+                    errorMessage += ", failed to rollback any changes happened in-between";
+                }
+            }
```
Manuri: aren't this catch block and the previous catch block the same?

stdlib/database/h2/src/test/resources/test-src/sql/sql_actions_negative_test.bal
```diff
@@ -65,7 +65,7 @@ function testGeneratedKeyOnInsert() returns int|string {
     if (x is sql:UpdateResult) {
         ret = x.generatedKeys.length();
     } else {
-        ret = <string> x.detail().message;
+        ret = io:sprintf("%s", x);
```
anupama-pathirage: Instead of `io:sprintf` shall we try to get the message and return it?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/CallStatement.java
```diff
@@ -116,11 +118,21 @@ public Object execute() {
...
+        } catch (SQLException e) {
+            cleanupResources(resultSets, stmt, conn, !isInTransaction);
+            // handleErrorOnTransaction(context);
+            // checkAndObserveSQLError(context, "execute stored procedure failed: " + e.getMessage());
+            return SQLDatasourceUtils.getSQLDatabaseError(e, "execute stored procedure failed: ");
+        } catch (DatabaseException e) {
+            cleanupResources(resultSets, stmt, conn, !isInTransaction);
+            // handleErrorOnTransaction(context);
+            // checkAndObserveSQLError(context, "execute stored procedure failed: " + e.getMessage());
+            return SQLDatasourceUtils.getSQLDatabaseError(e, "execute stored procedure failed: ");
```
Manuri: Isn't the logic in the two catch blocks duplicated?

stdlib/database/h2/src/test/java/org/ballerinalang/stdlib/database/sql/SQLActionsTest.java
```diff
@@ -359,18 +359,27 @@ public void testComplexTypeRetrieval() {
     public void testFailedSelect() {
         BValue[] returns = BRunUtil.invoke(resultNegative, "testSelectData");
         Assert.assertTrue(returns[0].stringValue().contains("execute query failed:"));
+        Assert.assertTrue(returns[0].stringValue().contains("{ballerina\\/sql}DatabaseError"));
+        Assert.assertTrue(returns[0].stringValue().contains("sqlErrorCode:"));
+        Assert.assertTrue(returns[0].stringValue().contains("sqlState:"));
```
Manuri: Shall we keep these tests as they are, and add a few new tests (IMO it is OK to duplicate these tests if needed) and verify that Database/Application errors are thrown whenever applicable?

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/exceptions/ApplicationException.java
```diff
@@ -0,0 +1,56 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.ballerinalang.database.sql.exceptions;
+
+/**
+ * The exception to throw Ballerina SQL application related exceptions.
+ *
+ * @since 0.8.0
```
Manuri: This should be 1.0.0

anupama-pathirage: Invalid version

Manuri:
We need to update the ballerina database client API to return JDBCClientError instead of generic error where applicable.

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/UpdateStatement.java
```diff
@@ -97,8 +99,16 @@ public Object execute() {
...
-        } catch (Throwable e) {
-            return SQLDatasourceUtils.getSQLConnectorError(e, "execute update failed: ");
+        } catch (SQLException e) {
+            return SQLDatasourceUtils.getSQLDatabaseError(e, "execute update failed: ");
+            //handleErrorOnTransaction(context);
+           // checkAndObserveSQLError(context, "execute update failed: " + e.getMessage());
+        }  catch (DatabaseException e) {
+            return SQLDatasourceUtils.getSQLDatabaseError(e, "execute update failed: ");
+            //handleErrorOnTransaction(context);
+            // checkAndObserveSQLError(context, "execute update failed: " + e.getMessage());
```
anupama-pathirage: Above two catch blocks has the same logic

stdlib/database/sql/src/main/ballerina/sql/types.bal
```diff
@@ -206,6 +206,20 @@ public type UpdateResult record {|
...
+public type ApplicationError error<APPLICATION_ERROR_REASON, ApplicationErrorData>;
+
+public type JDBCClientError DatabaseError|ApplicationError;
```
Manuri: For these types, I suppose we should add ballerina doc descriptions. These will appear in the API docs so we should have clear explanations on when these error types are returned.

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/SQLDataIterator.java
```diff
@@ -482,7 +483,7 @@ private void handleBooleanValue(MapValue<String, Object> bStruct, String fieldNa
     }
 
     private void handleDateValue(MapValue<String, Object> record, String fieldName, java.util.Date date,
-                                 BType fieldType) {
+                                 BType fieldType) throws ApplicationException {
```
anupama-pathirage: If we throw ApplicationException for the methods which should panic it will be a issue, as trap scenarios won't work properly. (This method is used by both remote methods and normal methods)

Biruntha: Since I'm throwing from [1], I have changed.
[1] https://github.com/ballerina-platform/ballerina-lang/pull/16093/files#diff-19329368634f636768c0b86cba12a8f5R459

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/BatchUpdateStatement.java
Biruntha: In both catch I'm calling getSQLDatabaseError method with different exception. So it is calling the overloaded getSQLDatabaseError method.

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/CallStatement.java
Biruntha: In both catch I'm calling getSQLDatabaseError method with different exception. So it is calling the overloaded getSQLDatabaseError method

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/UpdateStatement.java
Biruntha: But In both catch I'm calling getSQLDatabaseError method with different exception. So it is calling the overloaded getSQLDatabaseError method

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/AbstractSQLStatement.java
```diff
@@ -404,28 +407,30 @@ protected void cleanupResources(Statement stmt, Connection conn, boolean connect
                 conn.close();
             }
         } catch (SQLException e) {
-            throw new BallerinaException("error in cleaning sql resources: " + e.getMessage(), e);
+            throw new Error(errorMessagePrefix + "error in cleaning sql resources: " + e.getMessage(), e);
```
Manuri: How about we let the SQL exception to be thrown from this method and handle it where this is called? 

We can't throw java Error instances.
Quoting java documentation.

> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch. Most such errors are abnormal conditions. The ThreadDeath error, though a "normal" condition, is also a subclass of Error because most applications should not try to catch it.

> A method is not required to declare in its throws clause any subclasses of Error that might be thrown during the execution of the method but not caught, since these errors are abnormal conditions that should never occur.

stdlib/database/h2/src/test/resources/test-src/sql/sql_actions_negative_test.bal
```diff
@@ -47,6 +47,34 @@ function testSelectData() returns string {
...
+    if (x is table<record {}>) {
+        var jsonConversionResult = json.convert(x);
+        if (jsonConversionResult is json) {
+            retVal = jsonConversionResult;
+        } else {
+            retVal = { "Error": <string> jsonConversionResult.detail().message };
+        }
+    } else {
+        error e = x;
+        retVal = { "Error": io:sprintf("%s", e) };
```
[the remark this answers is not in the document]
Biruntha: If we  return the error message only , we can't check whether it is DatabaseError or ApplicationError

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/AbstractSQLStatement.java
```diff
@@ -404,28 +407,30 @@ protected void cleanupResources(Statement stmt, Connection conn, boolean connect
...
+                                    boolean connectionClosable) {
         try {
             if (rs != null && !rs.isClosed()) {
                 rs.close();
             }
-            cleanupResources(stmt, conn, connectionClosable);
+            cleanupResources(errorMessagePrefix, stmt, conn, connectionClosable);
         } catch (SQLException e) {
-            throw new BallerinaException("error in cleaning sql resources: " + e.getMessage(), e);
+            throw new Error(errorMessagePrefix + "error in cleaning sql resources: " + e.getMessage(), e);
```
anupama-pathirage: Why we need to throw an `Error` here? Since this `cleanupResources` method is getting called within catch blocks and finally blocks we need to throw `BallerinaException` here to make it a panic scenario.

stdlib/database/sql/src/main/java/org/ballerinalang/database/sql/statement/AbstractSQLStatement.java
anupama-pathirage: IMO, Since this cleanupResources method is getting called within catch blocks and finally blocks we need to throw BallerinaException here to make it a panic scenario.

Manuri: I was having doubts whether we should let it panic or return it. May be thats the way to go. Since there is nothing user can do about that scenario.

stdlib/database/h2/src/test/java/org/ballerinalang/stdlib/database/sql/SQLConnectionPoolTest.java
```diff
@@ -188,6 +189,6 @@ public void testLocalSharedConnectionPoolStopInitInterleave() {
     public void testLocalConnectionPoolShutDown() {
         BValue[] returns = BRunUtil.invokeFunction(result, "testLocalConnectionPoolShutDown");
         Assert.assertTrue(returns[0] instanceof BValueArray);
-        Assert.assertEquals(returns[0].stringValue(), "(1, 1)");
+        Assert.assertEquals(returns[0].stringValue(), "(6, 4)");
```
Manuri: What is the reason for changing the expected number of open connections? 
I think this is supposed to be 1, 1 because this test test is supposed to run at the end and verify that all connections are closed except for the 2 connections we are using to obtain the number of connections.

Biruntha: Some of the connection created will not be closed due the Runtime exception thrown for some test cases. that's why I change this count.

Biruntha: Fixed the above issue with https://github.com/ballerina-platform/ballerina-lang/pull/16093/commits/67411a901dc1a65133156820d659251b3683516e

Manuri:
@Biruntha As further improvements, we need to,

1. Check for a way to avoid duplicating logic in catch statements
2. Improve error messages

Let's create issues for these to track.

anupama-pathirage:
There are not tests for checking the returned errors. Shall we add a test/s for checking whether the returned error is a `JdbcClientError` , `error` or the specific errors

stdlib/database/h2/src/test/resources/test-src/sql/sql_actions_negative_test.bal
```diff
@@ -158,15 +257,102 @@ function testInvalidArrayofQueryParameters() returns string {
...
+        });
+    json retVal;
+    var x = testDB->select("SELECT Name from Customers where registrationID = 1", ());
+
+    boolean isJdbcClientError = false;
+    boolean isDatabaseError = false;
+
+    if (x is sql:JdbcClientError) {
+        isJdbcClientError = true;
+        if (x is sql:DatabaseError) {
```
anupama-pathirage: Instead of having this within the `if (x is sql:JdbcClientError)` block, shall we have this check outside?

stdlib/database/h2/src/test/resources/test-src/sql/sql_actions_negative_test.bal
```diff
@@ -158,15 +257,102 @@ function testInvalidArrayofQueryParameters() returns string {
...
+    xml[] xmlDataArray = [x1, x2];
+    sql:Parameter para0 = { sqlType: sql:TYPE_INTEGER, value: xmlDataArray };
+    var x = testDB->select("SELECT FirstName from Customers where registrationID in (?)", (), para0);
+
+    boolean isJdbcClientError = false;
+    boolean isApplicationError = false;
+
+    if (x is sql:JdbcClientError) {
+        isJdbcClientError = true;
+        if (x is sql:ApplicationError) {
```
anupama-pathirage: Instead of having this within the `if (x is sql:JdbcClientError)` block, shall we have this check outside? 
Also lets check this for `error` as well.

PR #16118, opened 2019-06-29
Move implicit initial value check to runtime

### LINKED ISSUE #15885: Relax var length array rules

grainier:
## Purpose
Purpose of this PR is to relax var length array rules, by move implicit initial value check to runtime. This allows initializing variable length arrays without filler values. However, it will panic if we try to create a sparse array without filler values.

Fixes #15885 

## Remarks
- https://github.com/ballerina-platform/ballerina-spec/issues/216
- https://github.com/ballerina-platform/ballerina-lang/issues/15885

PS: This is the updated PR of #15996 for JBallerina.


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1361,4 +1357,125 @@ public boolean equals(Object otherPair) {
...
+
+    private static boolean hasFillerValue(BType type, List<BType> unanalyzedTypes) {
+        if (type == null) {
+            return true;
+        }
+        if (type.getTag() < TypeTags.RECORD_TYPE_TAG) {
+            return true;
+        }
+        switch (type.getTag()) {
+            case TypeTags.TYPEDESC_TAG:
```
hasithaa: typedesc doesn't have a filler value.

grainier: fixed in 1b1d65b

PR #15881, opened 2019-06-20
Fix tuple syntax changes

grainier:
## Purpose
Purpose of this PR is to

1. Migrate tuple syntax changes
2. Fix issue with tuples in unions
3. Remove support for tuple destructuring when defined with `var` 

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Required Balo version update
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #16123, opened 2019-06-30
Fix tuple syntax changes

grainier:
## Purpose
Purpose of this PR is to

1. Migrate tuple syntax changes
2. Fix issue with tuples in unions
3. Remove support for tuple destructuring when defined with `var` 

Note: This is the updated PR of #15881 for JBallerina.


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -589,6 +586,18 @@ public void visit(BLangTupleVariable varNode) {
         typeChecker.checkExpr(varNode.expr, env, varNode.type);
     }
 
+    private BType resolveTupleType(BLangTupleVariable varNode) {
+        List<BType> memberTypes = new ArrayList<>(varNode.memberVariables.size());
+        for (BLangVariable memberVariable : varNode.memberVariables) {
+            if (memberVariable.getKind() == NodeKind.TUPLE_VARIABLE) {
```
hasithaa: Need to support record and error binding pattern as well. Please create an issue.

PR #16182, opened 2019-07-02
include kid in JWT Header

JensWalter:
## Purpose
In the current implementation of issuing a JWT, the key id (`kid`) is supported through the JwtHeader type, but it is not rendered into the resulting JWT.

## Approach
Just adding the `kid` to the header, if a value is present.

## Samples
modified from the "issue JWT test case"
```
jwt:JwtHeader header = {};
header.alg = jwt:RS256;
header.typ = "JWT";
header.kid = "custom-key-1";

```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/jwt/src/main/ballerina/jwt/jwt_issuer.bal
+++ b/stdlib/jwt/src/main/ballerina/jwt/jwt_issuer.bal
@@ -95,6 +95,9 @@ function buildHeaderString(JwtHeader header) returns (string|error) {
         return prepareError("Unsupported JWS algorithm.");
     }
     headerJson[TYP] = "JWT";
+    if (header.kid != ()) {
+        headerJson[kid] = header.kid;
+    }
     string headerValInString = headerJson.toString();
     string encodedPayload = encoding:encodeBase64Url(headerValInString.toByteArray("UTF-8"));
     return encodedPayload;
--- a/stdlib/jwt/src/main/ballerina/jwt/jwt_issuer.bal
+++ b/stdlib/jwt/src/main/ballerina/jwt/jwt_issuer.bal
@@ -95,7 +95,7 @@ function buildHeaderString(JwtHeader header) returns (string|error) {
         return prepareError("Unsupported JWS algorithm.");
     }
     headerJson[TYP] = "JWT";
-    if (header.kid is string) {
+    if (header["kid"] is string) {
         headerJson[KID] = header.kid;
     }
     string headerValInString = headerJson.toString();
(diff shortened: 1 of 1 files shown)


JensWalter:
AppVeyor reports an issue with the optional element check:
```
  ERROR: ballerina/jwt::jwt_issuer.bal:98:9:: unnecessary condition: expression will always evaluate to 'true'
	at org.ballerinalang.stdlib.utils.GenerateBalo.genBalo(GenerateBalo.java:130)
	at org.ballerinalang.stdlib.utils.GenerateBalo.main(GenerateBalo.java:81)
:ballerina-jwt:createBalo (Thread[Daemon worker,5,main]) completed. Took 16.001 secs.
```
code section:
```
public type JwtHeader record {|
    JwtSigningAlgorithm alg?;
    string typ?;
    string cty?;
    string kid?;
|};
...
function buildHeaderString(JwtHeader header) returns (string|error) {

    if (header.kid is string) {
        headerJson[KID] = header.kid;
    }
...
```

So the check works as expected (executes if the value is non-empty, skipped if empty), but how must the check look like to conform with the AppVeyor check?

ldclakmal:
@JensWalter Thanks for fixing this bug.

In the `JwtHeader` code section, the `kid` filed is marked as **optional**, but not **nilable**.
```ballerina
string kid?;
```
**Nilable** (union of `string` and `nil`) field looks as follows:
```ballerina
string? kid;
```

Therefore, the `kid` filed is always `string`, but it may or may not exists. That is why the _AppVeyor_ and _Travis_ build fails.

You can fix that with index based access as follows:
```ballerina
    if (header["kid"] is string) {
        headerJson[KID] = header.kid;
    }
```

stdlib/jwt/src/main/ballerina/jwt/jwt_issuer.bal
```diff
@@ -95,6 +95,9 @@ function buildHeaderString(JwtHeader header) returns (string|error) {
         return prepareError("Unsupported JWS algorithm.");
     }
     headerJson[TYP] = "JWT";
+    if (header.kid is string) {
```
ldclakmal:
```suggestion
    if (header["kid"] is string) {
```

JensWalter:
Thanks @ldclakmal
I applied the change.

Issue #16175, opened 2019-07-02
Iterator functions mistakenly mark closure variables tainted


rdhananjaya:
**Description:**
Consider following:
```ballerina
type ClosedPerson record {|
    string name = "";
    int age = 0;
|};

function testForeachOpWithClosedRecords() returns map<any> {
    ClosedPerson p = { name: "John Doe", age: 25};
    map<any> rec = {};

    p.foreach(function ((string, any) entry) {
            var (field, value) = entry;
            rec[field] = value;
        });

    sensitiveF(rec);
    return rec;
}

function sensitiveF(@sensitive map<any> p0) {

}
```
Compiler complain `tainted value passed to sensitive parameter 'p0'`.

It seems when taint analyzing the lambda function argument to `foreach`, symbol of `rec` (from outer function) is marked tainted, and when the lambda analyzis is complete taint status of `rec` doesn't seems to get restored.

**Steps to reproduce:**

**Affected Versions:**
0.991.0

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Issue #15242, opened 2019-05-07
Remove untaint expression and Implement taint analysis using annotations.


hasithaa:
New Design is captured here.

https://docs.google.com/document/d/1ZVSOpsiv2vewk8HL9jcbNZIVqI06OZJzxG4CeE9WVwE/edit#

PR #16200, opened 2019-07-03
Taint analyzer change

rdhananjaya:
## Purpose
Resolve https://github.com/ballerina-platform/ballerina-lang/issues/15242
Also fixes https://github.com/ballerina-platform/ballerina-lang/issues/16175

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -554,12 +554,21 @@ error.invalid.action.invocation=\
...
+  tainted value passed to untainted parameter ''{0}''
+
+error.tainted.value.passed.to.untainted.param.in.obj.method=\
+  tainted value ''{0}'' passed to untainted parameter originating from object method invocation
 
 error.tainted.value.passed.to.global.variable=\
   tainted value passed to global variable ''{0}''
 
+error.tainted.value.passed.to.global.object=\
+  tainted value passed to global object ''{0}''
```
hasithaa: Do we need says which tainted value? 

How about rewording, this to `tainted value ''{0}'' passed to module object variable ''{0}''`

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -554,12 +554,21 @@ error.invalid.action.invocation=\
 error.undefined.action=\
   undefined remote function ''{0}'' in client object {1}
 
-error.tainted.value.passed.to.sensitive.parameter=\
-  tainted value passed to sensitive parameter ''{0}''
+error.tainted.value.passed.to.untainted.parameter=\
+  tainted value passed to untainted parameter ''{0}''
+
+error.tainted.value.passed.to.untainted.param.in.obj.method=\
+  tainted value ''{0}'' passed to untainted parameter originating from object method invocation
```
hasithaa: Can we say which parameter?

Can we say, `tainted value ''{0}'' passed to untainted method parameter ''{1}''`?

compiler/ballerina-lang/src/main/resources/compiler.properties
rdhananjaya: As tainted value could be coming from any expression, we can't say which tainted value. I'll change the other part.

compiler/ballerina-lang/src/main/resources/compiler.properties
rdhananjaya: Here again we can't name the value as it can be result of any expression.
Message changed to `"tainted value passed to untainted parameter ''{0}'' originating from object method ''{1}'' invocation`, which identify name of the untainted parameter being tainted and which method does that.

PR #16295, opened 2019-07-06
Add test cases for int lib functions

pubudu91:


hasithaa:
verified the build locally.

PR #16319, opened 2019-07-08
Remove extra groupings in function return syntax

rdhananjaya:
$subject.


stdlib/io/src/main/ballerina/io/strings.bal
```diff
@@ -13,6 +13,7 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
+import ballerina/time;
```
MaryamZi: Do we need this import?

rdhananjaya: That was a mistake, removed

PR #16349, opened 2019-07-09
[WIP]Introduce Error Types for HTTP Module

### LINKED ISSUE #16145: Add Error types to HTTP module
**Description:**
Introduce Error types to Http module, instead of using built-in `error` type. The module then have two main types of errors:
1. ClientError
2. ListenerError

These errors will be unions of the other errors. Users can use type-guard to differentiate these errors when handling the errors.

@ThisaruGuruge

ThisaruGuruge:
These errors have the following structure:
```ballerina
error<Reason, Detail>
```

The `Reason` is a constant string, which defines the error type. The Detail record is defined as follows:
```ballerina
public type Detail record {
    string message;
    error cause?;
}
```
The `message` field holds the specific error message. 
If this particular error is occurred due to another error (probably from another module) will be held in `cause` field.
This is an open record, hence there can be other fields as well. 

Fixes #16145 

## Samples
```ballerina
var result = someFunction();
if (result is http:ClientError) {
    // User can handle ClientError or go into sub errors
    if (result is http:GenericClientError) {
            // User can handle generic http client error or see the cause
            var cause = result.detail().cause;
            if (cause is io:IoError) {
                // Handle IoError or go into subsets
            }
        } else if (result is http:SecurityError) {
            io:println(result);
        }
    } else {
        io:println("ConnectionTimedOutError Error received.");
    }
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


ThisaruGuruge:
Closed as this is tracked by #16672

PR #16408, opened 2019-07-10
Langlib decimal test

rdhananjaya:
## Purpose
Add few test cases for langlib.decimal functions


langlib/langlib-test/src/test/resources/test-src/decimallib_test.bal
```diff
@@ -0,0 +1,58 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/'lang\.decimal as decimals;
+
+
+function testSum(decimal p1, decimal p2) returns decimal {
+    return decimals:sum(p1, p2);
```
hasithaa: Shall we add test like, `p1.sum(p2)`;

langlib/langlib-test/src/test/resources/test-src/decimallib_test.bal
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function testOneArgMax(decimal arg) returns decimal {
+    return decimals:max(arg);
+}
+
+function testMultiArgMax(decimal arg, decimal[] otherArgs) returns decimal {
+    return decimals:max(arg, ...otherArgs);
```
hasithaa: Add test cases like `arg.max(...otherArgs)`

PR #16442, opened 2019-07-10
Introduce new semantic changes from the spec R2 to foreach loop

gimantha:
## Purpose
This PR introduces the new semantic changes from the R2 spec.

## Approach
Lets say we have an iterable collection itr and we write the foreach as below

```ballerina
foreach var v in itr {
        //do something
}
```

The above foreach block will be desugared to below code. Note that all iterable types have iterator() extern function

```ballerina
$data$ = itr;
abstract object {public function next() returns record {|int value;|}? $iterator$ = $data$.iterator();
record {|int value;|}? $result$ = $iterator$.next();
while $result$ is record {|int value;|} {
     int v = $result$.value;
     $result$ = $iterator$.next();
     //do something
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Next.java
```diff
@@ -0,0 +1,70 @@
...
+ * Native implementation of lang.array.ArrayIterator:next().
+ *
+ * @since 1.0
+ */
+@BallerinaFunction(
+        orgName = "ballerina", packageName = "lang.array", functionName = "next",
+        receiver = @Receiver(type = TypeKind.OBJECT, structType = "ArrayIterator",
+                structPackage = "ballerina/lang.array"),
+        args = {@Argument(name = "arr", type = TypeKind.OBJECT)},
+        returnType = {@ReturnType(type = TypeKind.INT)},
```
pubudu91: Wrong args and returnType attributes?

gimantha: fixed

PR #16470, opened 2019-07-11
Add test cases for map lib functions

pubudu91:


pubudu91:
Set to WIP because there an issue with error that needs to be fixed before merging this PR

PR #16499, opened 2019-07-12
Remove loadToMemory option in jdbc select remote function

anupama-pathirage:
## Purpose
Existing 'loadToMemory' option in select remote function only provides the capability of re-iterating the same table. 

There are some issues with that option.
- That will load all the data to the memory and there is a risk that application goes OOM, if the database has large set of data matched to the given query.
- With the 2019 r2 lang spec, we can't have both optional parameters and rest parameters together in the same function signature. So to support this feature, we need to make this a mandatory parameter. But this is not needed for 99% use cases of the select function. So having it as a mandatory parameter is a bad design.

After removing this parameter, the alternative to re-iterate the same data set twice is:
- invoke the select function twice (There is a chance that data-set has changed in the DB when the second invocation happens)
- Copy the data of the table to record array. It can be iterated over and over

Fixes #<Issue Number>

## Check List 
- [x] Required Balo version update
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/SelectStatement.java
```diff
@@ -74,19 +70,13 @@ public Object execute() {
...
-            if (loadSQLTableToMemory) {
-                CachedRowSet cachedRowSet = RowSetProvider.newFactory().createCachedRowSet();
-                cachedRowSet.populate(rs);
-                rs = cachedRowSet;
-                rm.gracefullyReleaseResources();
-            } else {
-                rm.addResultSet(rs);
-            }
+            rm.addResultSet(rs);
+            rm.addResultSet(rs);
```
Manuri: I guess we don't have to add twice..?

Manuri:
We have a check for CachedRowSet in SQLDataItarator[1][2]. We can remove that as well since this is not required anymore.

[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/jdbc/src/main/java/org/ballerinax/jdbc/SQLDataIterator.java#L93
[2] https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/jdbc/src/main/java/org/ballerinax/jdbc/SQLDataIterator.java#L81

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/SelectStatement.java
anupama-pathirage: I see. Yes. Seems a mistake

PR #16558, opened 2019-07-13
Introduce http error types

### LINKED ISSUE #16145: Add Error types to HTTP module

ThisaruGuruge:
> These errors have the following structure:
> ```ballerina
> error<Reason, Detail>
> ```

> The `Reason` is a constant string, which defines the error type. The Detail record is defined as follows:
> ```ballerina
> public type Detail record {
>     string message;
>     error cause?;
> }
> ```

> The `message` field holds the specific error message. 
> If this particular error is occurred due to another error (probably from another module) will be held in `cause` field.
> This is an open record, hence there can be other fields as well. 

Fixes #16145 

## Samples

> ```ballerina
> var result = someFunction();
> if (result is http:ClientError) {
>     // User can handle ClientError or go into sub errors
>     if (result is http:GenericClientError) {
>             // User can handle generic http client error or see the cause
>             var cause = result.detail().cause;
>             if (cause is io:Error) {
>                 // Handle io:Error or go into subsets
>             }
>         } else if (result is http:SecurityError) {
>             io:println(result);
>         }
>     } else {
>         io:println("ConnectionTimedOutError Error received.");
>     }
> ```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/http/src/main/ballerina/http/http_client.bal
```diff
@@ -38,7 +38,7 @@ public type HttpClient client object {
     # + message - An HTTP outbound request message or any payload of type `string`, `xml`, `json`, `byte[]`,
     #             `io:ReadableByteChannel` or `mime:Entity[]`
     # + return - The response for the request or an `error` if failed to establish communication with the upstream server
```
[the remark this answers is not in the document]
chamil321: Shall we change it in other places too

stdlib/http/src/main/ballerina/http/http_errors.bal
```diff
@@ -0,0 +1,151 @@
+// Copyright (c) 2018 WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
wggihan:
```suggestion
// Copyright (c) 2019 WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

stdlib/http/src/main/ballerina/http/http_secure_client.bal
```diff
@@ -277,23 +278,23 @@ public function createHttpSecureClient(string url, ClientEndpointConfig config)
 #
 # + req - An HTTP outbound request message
 # + config - Client endpoint configurations
-# + return - Prepared HTTP request or `error` if an error occurred at auth handler invocation
+# + return - Prepared HTTP request or `http:ClientError` if an error occurred at auth handler invocation
 function prepareSecureRequest(Request req, ClientEndpointConfig config) returns Request|error {
```
chamil321: Any reason why have we returned `error` instead of return type of prepareAuthenticationError method?

stdlib/http/src/main/ballerina/http/resiliency/failover_client_endpoint.bal
```diff
@@ -240,7 +240,7 @@ public type FailoverClient client object {
     #
     # + httpFuture - The `HttpFuture` related to a previous asynchronous invocation
     # + return - An HTTP response message, or an `error` if the invocation fails
```
chamil321:
```suggestion
    # + return - An HTTP response message, or an `http:ClientError` if the invocation fails
```

chamil321: Fix other places too

tests/jballerina-unit-test/src/test/resources/test-src/taintchecking/connectors/httpclient.bal
```diff
@@ -15,10 +15,12 @@ public function main (string... args) {
...
             normalFunction(msg);
         } else {
-            panic msg;
+            error err = msg;
+            panic err;
         }
     } else {
-        panic response;
+        error err = response;
+        panic err;
```
chamil321:
```suggestion
        panic <error> response;
```

stdlib/http/src/main/ballerina/http/resiliency/load_balance_client_endpoint.bal
```diff
@@ -217,35 +221,36 @@ public type LoadBalanceActionError error<string, LoadBalanceActionErrorData>;
...
 function performLoadBalanceExecuteAction(LoadBalanceClient lb, string path, Request request,
-                                         string httpVerb) returns @tainted Response|error {
+                                         string httpVerb) returns Response|ClientError {
     HttpOperation connectorAction = extractHttpOperation(httpVerb);
     if (connectorAction != HTTP_NONE) {
         return performLoadBalanceAction(lb, path, request, connectorAction);
     } else {
-        error httpActionErr = error("Unsupported connector action received.");
-        return httpActionErr;
+        string message = "Load balancer client not supported for connector action: " + httpVerb;
```
wggihan: What's meant by the connector action? It is possible to rename that with more user friendly manner wording?

tests/jballerina-unit-test/src/test/resources/test-src/taintchecking/connectors/httpclient.bal
chamil321: We can get rid of additional variable creation, if we do the cast and panic. WDYT?

stdlib/http/src/main/java/org/ballerinalang/net/http/HttpUtil.java
```diff
@@ -495,29 +518,124 @@ private static void setHttpStatusCodes(int statusCode, HttpCarbonMessage respons
...
+            return createHttpError(throwable.getMessage());
+        }
+    }
+
+    public static ErrorValue createHttpError(String errorMessage) {
+        HttpErrorType errorType = getErrorType(errorMessage);
+        return createHttpError(errorMessage, errorType);
+    }
+
+    public static ErrorValue createHttpError(Throwable throwable) {
```
Manuri: Why not write a set of overloaded methods for each exception type instead of checking instanceOf inside one method?

stdlib/http/src/main/ballerina/http/http_secure_client.bal
ThisaruGuruge: It's a mistake. Will update.

stdlib/http/src/main/ballerina/http/resiliency/load_balance_client_endpoint.bal
ThisaruGuruge: Will update

stdlib/http/src/main/java/org/ballerinalang/net/http/HttpUtil.java
ThisaruGuruge: Done

stdlib/http/src/main/ballerina/http/http_errors.bal
```diff
@@ -0,0 +1,149 @@
...
+public type UpstreamServiceUnavailableError error<UPSTREAM_SERVICE_UNAVAILABLE, Detail>;
+
+public const ALL_LOAD_BALANCE_ENDPOINTS_FAILED = "{ballerina/http}AllLoadBalanceEndpointsFailed";
+public type AllLoadBalanceEndpointsFailedError error<ALL_LOAD_BALANCE_ENDPOINTS_FAILED, Detail>;
+
+public const ALL_RETRY_ATTEMPTS_FAILED = "{ballerina/http}AllRetryAttemptsFailed";
+public type AllRetryAttemptsFailed error<ALL_RETRY_ATTEMPTS_FAILED, Detail>;
+
+public const IDLE_TIMEOUT_TRIGGERED = "{ballerina/http}IdleTimeoutTriggeredError";
+public type IdleTimeoutTriggeredError error<IDLE_TIMEOUT_TRIGGERED, Detail>;
```
shafreenAnfar: Should we re-name this to IdleTimeoutError ?

ThisaruGuruge: Done

tests/jballerina-unit-test/src/test/resources/test-src/taintchecking/connectors/httpclient.bal
ThisaruGuruge: Agreed, done

stdlib/http/src/main/ballerina/http/resiliency/failover_client_endpoint.bal
ThisaruGuruge: Done

ThisaruGuruge:
This is merged with #16672

PR #16564, opened 2019-07-13
Constructfrom native impl

rdhananjaya:
## Purpose
$subject.

```ballerina
type Person record {
    string name;
    int age;
};

function testRecToJson() returns json|error {
    Person  p = {name: "N", age: 3};
    json|error j = typedesc<json>.constructFrom(p);
    return j;
}
```


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -555,7 +556,14 @@ private BType getMatchingErrorBoundType(BErrorType expType, SymbolEnv env, HashS
         }
         BType reasonType = getMatchingBoundType(expType.reasonType, env, resolvedTypes);
         BType detailType = getMatchingBoundType(expType.detailType, env, resolvedTypes);
-        return new BErrorType(null, reasonType, detailType);
+        BErrorTypeSymbol typeSymbol = new BErrorTypeSymbol(SymTag.ERROR,
```
hasithaa: copying symboltable.error symbol's seems not right, it may print the wrong error type. But printing its structure has its own problems. So let's keep this. Ideally, we should fix the structure problem. 

Other option is check 
```
if( reasonType == symTable.string && types.isSame(detailType, symTable.detailType) {
  return symTable.errorType;
}
```

langlib/langlib-test/src/test/resources/test-src/typedesclib_test.bal
```diff
@@ -0,0 +1,26 @@
...
+// under the License.
+
+type Person record {
+    string name;
+    int age;
+};
+
+function testRecToJson() returns json|error {
+    Person  p = {name: "N", age: 3};
+    json|error j = typedesc<json>.constructFrom(p);
```
hasithaa: Can you write another test case using json.constructFrom(p);

PR #16570, opened 2019-07-13
Refactor Integer range expression according to the R2 spec

gimantha:
## Purpose
With this PR a range expression cannot be assigned to an integer array as we used to do before.  The result type of an integer range expression is an Iterable object.

**Old code**

`int[] intArray = 20...30;`

**New code**

```ballerina
abstract object {public function next () returns (record {| int value; |}?);} rangeAsArray = 20 ... 30;
```

## Approach
A object called "IntRange" which is an iterable object is added to the internal module. In the desugar phase the integer range is replaced with this object which is initialized with the start and end value. By calling the next() function, the next value of the integer range is returned. With this approach an integer range can be iterated using foreach loop in the same way maps, arrays ...etc are iterated.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -140,6 +140,7 @@
 import org.wso2.ballerinalang.compiler.util.TypeTags;
 import org.wso2.ballerinalang.compiler.util.diagnotic.BLangDiagnosticLog;
 import org.wso2.ballerinalang.compiler.util.diagnotic.DiagnosticPos;
+import org.wso2.ballerinalang.programfile.InstructionCodes;
```
hasithaa: Unused import. ?

PR #16616, opened 2019-07-14
Enable jBallerina array stmt tests

grainier:
## Purpose
Purpose of this PR is to $subject. This will also temporarily disable some negative test cases as well.

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
@@ -1508,8 +1508,8 @@ public class TypeChecker {
         } else {
             AttachedFunction initializerFunc = type.initializer;
             if (initializerFunc == null) {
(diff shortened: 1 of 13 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1509,7 +1509,7 @@ private static boolean checkFillerValue(BObjectType type) {
             AttachedFunction initializerFunc = type.initializer;
             if (initializerFunc == null) {
                 // No __init function found.
-                return true;
+                return !Flags.isFlagOn(Flags.ABSTRACT, type.flags);
```
pubudu91: Doesn't the initializerFunc become null for abstract object types only?

grainier: Yep, removed redundant check.

PR #16671, opened 2019-07-14
Fix compilation issues in socket integration test resources

Bhashinee:
## Purpose
$subject.


tests/jballerina-integration-test/src/test/resources/socket/services/01_client_callback.bal
```diff
@@ -17,6 +17,7 @@
 import ballerina/http;
 import ballerina/io;
 import ballerina/socket;
+import ballerina/internal;
```
MaryamZi: Shall we move this to L18?

PR #16682, opened 2019-07-15
Fix object test cases

rdhananjaya:
## Purpose
Fix object test cases


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/object/AnonymousObjectTest.java
```diff
@@ -148,12 +148,13 @@ public void testObjectWithSelfReference() {
         Assert.assertEquals(returns[1].stringValue(), "Tyler Jewell");
     }
 
-    @Test(description = "Negative test to test un-defaultable anon object")
+    @Test(description = "Negative test to test un-defaultable anon object", enabled = false)
```
SupunS: better to add to the  "brokenOnLangLibChange", rather than disabling

rdhananjaya: Done

PR #16757, opened 2019-07-16
Remove unused method in sym resolver

pubudu91:


pubudu91:
Closing this PR since the change done in this is incorrect

PR #16795, opened 2019-07-16
Enable error var ref test cases

rdhananjaya:
$subject.

Also fixes error rest param ref deconstruction.


tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/error_variable_reference.bal
```diff
@@ -176,51 +180,8 @@ function testErrorInRecordWithDestructure2() returns [int, string, anydata|error
     return [x, reason, message, extra];
 }
 
-function testBasicErrorVariableWithFieldBasedRef() returns map<anydata|error> {
```
rdhananjaya: Error deconstruction assignment into index access ref are not supported.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/error_variable_reference.bal
```diff
@@ -242,15 +203,3 @@ function testErrorWithUnderscore() returns [string, map<string>] {
 
     return [reason, detail];
 }
-
```
rdhananjaya: Error detail type does not support map anymore

PR #16879, opened 2019-07-17
Cleanup modules

wggihan:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/crypto/src/main/java/org/ballerinalang/stdlib/crypto/nativeimpl/Crc32b.java
+++ b/stdlib/crypto/src/main/java/org/ballerinalang/stdlib/crypto/nativeimpl/Crc32b.java
@@ -18,8 +18,6 @@
 
 package org.ballerinalang.stdlib.crypto.nativeimpl;
 
-import org.ballerinalang.bre.Context;
-import org.ballerinalang.bre.bvm.BlockingNativeCallableUnit;
 import org.ballerinalang.jvm.Strand;
 import org.ballerinalang.jvm.TypeChecker;
 import org.ballerinalang.jvm.types.BArrayType;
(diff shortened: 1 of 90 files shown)


stdlib/crypto/src/main/java/org/ballerinalang/stdlib/crypto/nativeimpl/Crc32b.java
```diff
@@ -42,11 +40,7 @@
 @BallerinaFunction(
         orgName = "ballerina", packageName = "crypto",
         functionName = "crc32b", isPublic = true)
-public class Crc32b extends BlockingNativeCallableUnit {
-
-    @Override
-    public void execute(Context context) {
-    }
+public class Crc32b {
```
riyafa: Since the classes only have static functions we can add a private constructor to make sure these classes cannot be initialized.

wggihan: What is the use case that involves initializing these classes?

riyafa: That is a violation of a sonar rule. If a class has only static functions we need to prevent it from being intialized: https://rules.sonarsource.com/java/RSPEC-1118

wggihan: This is not utility class. This is a part of the framework implementation.

Sonar rule says "Utility classes should not have public constructors"

riyafa: It is to prevent a class from being instantiated. I think these classes should not be instantiated.

PR #16884, opened 2019-07-17
Fix and enable disabled database tests/ Cleanup

Manuri:
## Purpose
$Subject


stdlib/jdbc/src/main/java/org/ballerinax/jdbc/datasource/SQLDatasource.java
```diff
@@ -214,6 +214,32 @@ private void buildDataSource(SQLDatasourceParams sqlDatasourceParams) {
...
+            // Clarification on behavior with parameters.
+            // 1. Adding an invalid param in the JDBC URL. This will result in an error getting returned
+            //    eg: jdbc:Client testDB = new({
+            //        url: "jdbc:h2:file:./target/tempdb/TEST_SQL_CONNECTOR_INIT;INVALID_PARAM=-1",
+            //        username: "SA",
+            //        password: ""
+            //    });
+            // 2. Providing an invalid param with dataSourceClassName provided. This will result in an error
+            // returned because when hikaricp tries to call setINVALID_PARAM method on the given datasource
+            // class name, it will fail since there is no such method
```
anupama-pathirage:
```suggestion
            // class name, it will fail since there is no such method.
```

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/datasource/SQLDatasource.java
```diff
@@ -214,6 +214,32 @@ private void buildDataSource(SQLDatasourceParams sqlDatasourceParams) {
                     }
                 });
             }
+            // Clarification on behavior with parameters.
+            // 1. Adding an invalid param in the JDBC URL. This will result in an error getting returned
```
anupama-pathirage:
```suggestion
            // 1. Adding an invalid param in the JDBC URL. This will result in an error getting returned.
```

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/datasource/SQLDatasource.java
```diff
@@ -214,6 +214,32 @@ private void buildDataSource(SQLDatasourceParams sqlDatasourceParams) {
...
+            // class name, it will fail since there is no such method
+            // eg: jdbc:Client testDB = new({
+            //        url: "jdbc:h2:file:./target/tempdb/TEST_SQL_CONNECTOR_INIT",
+            //        username: "SA",
+            //        password: "",
+            //        poolOptions: { dataSourceClassName: "org.h2.jdbcx.JdbcDataSource" },
+            //        dbOptions: { "INVALID_PARAM": -1 }
+            //    });
+            // 3. Providing an invalid param WITHOUT dataSourceClassName provided. This may not return any error.
+            // Because this will result in the INVALID_PARAM being passed to Driver.Connect which may not recognize
```
anupama-pathirage:
```suggestion
            // Because this will result in the INVALID_PARAM being passed to Driver.Connect() which may not recognize
```

PR #16901, opened 2019-07-17
Fix spotbug issue in formatter tool

NipunaMarcus:
## Purpose
This will fix the spotbug issue in formatter tool.

### DIFF
--- a/misc/ballerina-formatter/src/main/java/org/ballerinalang/format/FormatUtil.java
+++ b/misc/ballerina-formatter/src/main/java/org/ballerinalang/format/FormatUtil.java
@@ -403,7 +403,7 @@ public class FormatUtil {
     }
 
     private static String getModuleName(String moduleName) {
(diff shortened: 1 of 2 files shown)


misc/ballerina-formatter/src/main/java/org/ballerinalang/format/FormatUtil.java
```diff
@@ -403,7 +403,7 @@ private static void generateChangeReport(List<String> formattedFiles, boolean dr
     }
 
     private static String getModuleName(String moduleName) {
-        String[] splitedTokens = moduleName.split(File.separator);
+        String[] splitedTokens = moduleName.split("" + File.separator);
```
SupunS: Is it safe to use `File.separator` here?
What about in windows env?

NipunaMarcus: This will work in both Unix and windows environment...

PR #16919, opened 2019-07-18
Enable local transactions in RabbitmQ connector

aashikam:
## Purpose
Enables local transaction support in RabbitMQ connector

## Tasks
- [x] Enable local transactions in RabbitMQ module
- [ ] Test functionality 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/messaging/rabbitmq/src/main/java/org/ballerinalang/messaging/rabbitmq/nativeimpl/channel/ExchangeDeclare.java
```diff
@@ -55,7 +55,7 @@ public static Object exchangeDeclare(Strand strand, ObjectValue channelObjectVal
         try {
             ChannelUtils.exchangeDeclare(channel, exchangeConfig);
             if (transactionContext != null) {
-                transactionContext.handleTransactionBlock();
+                transactionContext.handleTransactionBlock(strand);
```
Manuri: Why are we always checking whether a transaction context is present before checking whether we are inside a transaction?
I think it would be better if we first, check whether we are inside a transaction block and then check whether there is a transaction context.

stdlib/messaging/rabbitmq/src/main/java/org/ballerinalang/messaging/rabbitmq/RabbitMQTransactionContext.java
```diff
@@ -78,8 +84,26 @@ public XAResource getXAResource() {
 
     /**
      * Handles the transaction block if the context is in transaction.
+     *
+     * @param strand Strand.
      */
-    public void handleTransactionBlock() {
-     // TODO: Transaction handling logic
+    public void handleTransactionBlock(Strand strand) {
+        boolean isInTransaction = strand.isInTransaction();
```
wggihan: We are already checking this in outer method before calling this method. So why need this check again?

stdlib/messaging/rabbitmq/src/main/java/org/ballerinalang/messaging/rabbitmq/nativeimpl/channel/BasicPublish.java
```diff
@@ -60,8 +61,8 @@ public static Object basicPublish(Strand strand, ObjectValue channelObjectValue,
         try {
             ChannelUtils.basicPublish(channel, routingKey, messageContent.toString().getBytes(StandardCharsets.UTF_8),
                     defaultExchangeName, properties);
-            if (transactionContext != null) {
-                transactionContext.handleTransactionBlock();
+            if (isInTransaction && transactionContext != null) {
+                transactionContext.handleTransactionBlock(strand);
             }
```
daneshk: `transactionContext` cannot be null, because we are creating at channel creation. Shall we stop publishing the message and give an error if any case it is null and `isInTransaction`.

PR #16950, opened 2019-07-20
Remove old ballerina distributions and tests

keizer619:


warunalakshitha:
IMO  we haven't migrate ballerina-tools-integration-test yet. this needs to be recheck.

PR #16977, opened 2019-07-22
Implement lang.xml functions

rdhananjaya:
Add rest of lang.xml functions and test cases for them


rdhananjaya:
@rasika few function completion tests had to be commented out, probably because this PR add bunch of lang.xml functions.

rasika:
@rdhananjaya will look into it

langlib/lang.xml/src/main/java/org/ballerinalang/langlib/xml/GetName.java
```diff
@@ -0,0 +1,57 @@
...
+        if (!IsElement.isElement(strand, xmlVal)) {
+            throw BLangExceptionHelper.getRuntimeException(RuntimeErrors.XML_FUNC_TYPE_ERROR, "getName", "element");
+        }
+        try {
+            return xmlVal.getElementName();
+        } catch (Throwable e) {
+            BLangExceptionHelper.handleXMLException(OPERATION, e);
+        }
+
+        return null;
```
pubudu91: Shouldn't be returning `null` right? What's the appropriate value here? Empty string?

rdhananjaya: `BLangExceptionHelper.handleXMLException(OPERATION, e)` always throws so control never reach this point.
I would like to put a `assert false` here but since it's not being used anywhere else just went with a null.

PR #17008, opened 2019-07-23
Add rest descriptor support for tuple types

grainier:
## Purpose
Purpose of this pr is to $subject. With this pr, it allows the following;

## Samples
```ballerina
[int, string, boolean...] t = [1, "s", true, true];
[int, string, boolean...] t1 = [2, "s", true];
[int, string, boolean...] t2 = [3, "s"];
[int...] t3 = [1, 2];
[string...] t4 = ["s", "s"];
[string, Employee...] t5 = ["s", employee1, employee1, employee1];
[string, Person...] t6 = t5;

function foo([string, float, boolean...] t) {
    // foo body
}

function bar() returns [string, float, boolean...] {
    [string, float, boolean...] t = ["a", 5.0, true, false];
    return t;
}
```

## Tasks
- [x] Add rest descriptor support for tuple types
- [x] Tuple variable definition with rest descriptors
- [x] Tuple variable reference with rest descriptors
- [x] Unions of tuple with rest descriptors
- [x] Iteration (foreach, index-based) of tuple with rest descriptors
- [x] Tuple filler values
- [x] Tuple binding patterns (`list-binding-pattern`) with `match`.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BTupleType.java
```diff
@@ -30,21 +30,28 @@
...
+    private BType restType;
 
     /**
      * Create a {@code BTupleType} which represents the tuple type.
      *
      * @param typeList of the tuple type
+     * @param restType of the tuple type
      */
-    public BTupleType(List<BType> typeList) {
+    public BTupleType(List<BType> typeList, BType restType) {
```
MaryamZi: Wouldn't it be better to have two constructors instead? We can avoid all the calls passing null when there is not rest descriptor.

stdlib/bir/src/main/ballerina/src/bir/bir_model.bal
```diff
@@ -411,6 +411,7 @@ public type BUnionType record {|
 
 public type BTupleType record {|
    BType?[]  tupleTypes;
+   BType?  restType;
```
MaryamZi:
```suggestion
   BType?  restType = ();
```

We can set `()` as the initial value, keep https://github.com/ballerina-platform/ballerina-lang/pull/17008/files#diff-c55ac83ca455bbaf05c3333c10c991bdL210 as is and set https://github.com/ballerina-platform/ballerina-lang/pull/17008/files#diff-c55ac83ca455bbaf05c3333c10c991bdR214 only if `restPresent` is true.

MaryamZi:
`FillMember` now applies for tuples too right?

```ballerina
public function main() {
    [string, int...] t = ["hello", 1];
    t[3] = 3; // `t[2]` should be filled as `0` ?
}
```

If so, shall we validate and add tests?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BTupleType.java
grainier: Fixed in 507ce84

stdlib/bir/src/main/ballerina/src/bir/bir_model.bal
grainier: Fixed in 22233bb

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -935,27 +938,35 @@ private boolean checkTypeAndVarCountConsistency(BLangTupleVariable varNode, BTup
             }
         }
 
-        if (tupleTypeNode.tupleTypes.size() != varNode.memberVariables.size()) {
+        if (tupleTypeNode.tupleTypes.size() != varNode.memberVariables.size()
+                || (tupleTypeNode.restType == null && varNode.restVariable != null)
+                ||  (tupleTypeNode.restType != null && varNode.restVariable == null)) {
```
Anoukh: I think we should support the scenario where `tupleTypeNode.restType != null && varNode.restVariable == null`, which would allow the following:
```ballerina
    [int, string, string...] x = getTuple();
}

function getTuple() returns [int, string, string, string, string] {
    return [1, "A", "B", "C", "D"];
}

```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -817,17 +830,28 @@ private boolean checkTupleType(BLangExpression expression, BType type) {
...
+            }
+
+            for (int i = 0; i < tupleExpr.exprs.size(); i++) {
+                BLangExpression expr = tupleExpr.exprs.get(i);
+                if (i < tupleType.tupleTypes.size()) {
                     if (!checkTupleType(expr, tupleType.tupleTypes.get(i))) {
                         return false;
                     }
+                } else {
+                    if (tupleType.restType != null) {
```
Anoukh: I think we can simplify as follows
```ballerina
if (tupleType.restType == null || !checkTupleType(expr, tupleType.restType) return false;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1365,7 +1389,17 @@ public void visit(BLangTupleVarRef varRefExpr) {
...
-        BType actualType = new BTupleType(results);
+        BTupleType actualType = new BTupleType(results);
+        if (varRefExpr.restParam != null) {
+            BLangExpression restExpr = (BLangExpression) varRefExpr.restParam;
+            ((BLangVariableReference) restExpr).lhsVar = true;
+            BType checkedType = checkExpr(restExpr, env, symTable.noType);
+            if (checkedType.tag == TypeTags.ARRAY) {
+                actualType.restType = ((BArrayType) checkedType).eType;
+            } else {
+                actualType.restType = checkedType;
```
Anoukh: I think we can dlog here since rest param can only be of array type?

grainier: Fixed in 9423d9f

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
grainier: Fixed in 9423d9f

PR #17058, opened 2019-07-25
Fix few BBEs

irshadnilam:
## Purpose
$title

### DIFF
--- a/examples/strings/strings.bal
+++ b/examples/strings/strings.bal
@@ -5,17 +5,11 @@ public function main() {
 
     string statement = "Lion in Town. Catch the Lion";
 
-    string s1 = statement.toUpper();
-    io:println("ToUpper: " + s1);
+    string s1 = statement.toUpperAscii();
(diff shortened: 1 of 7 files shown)


examples/strings/strings.bal
```diff
@@ -5,84 +5,53 @@ public function main() {
 
     string statement = "Lion in Town. Catch the Lion";
 
-    string s1 = statement.toUpper();
-    io:println("ToUpper: " + s1);
+    string s1 = statement.toUpperAscii();
```
MaryamZi: Shall we also add examples for anything else that is supported in the string lang lib?

https://github.com/ballerina-platform/ballerina-lang/blob/master/langlib/lang.string/src/main/ballerina/src/lang.string/string.bal

MaryamZi: Let's add them in a separate PR.

irshadnilam: https://github.com/ballerina-platform/ballerina-lang/pull/17071

PR #17071, opened 2019-07-26
Add few more example to string BBE

irshadnilam:
## Purpose
$title

### DIFF
--- a/examples/strings/strings.bal
+++ b/examples/strings/strings.bal
@@ -1,5 +1,5 @@
-import ballerina/encoding;
 import ballerina/io;
+import ballerina/lang.'string as strings;
 
 public function main() {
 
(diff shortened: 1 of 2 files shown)


examples/strings/strings.bal
```diff
@@ -22,14 +22,34 @@ public function main() {
...
-    
+
     // Retrieves the length of the `string`.
     int length = statement.length();
     io:println("Length: ", length);
 
+    string hello = "Hello";
+    string ballerina = "Ballerina!";
+
+    // Concats multiple strings.
```
MaryamZi:
```suggestion
    // Concat multiple strings.
```

MaryamZi: Please change the comments in the rest of the BBE similarly.

examples/strings/strings.bal
```diff
@@ -22,14 +22,34 @@ public function main() {
...
+    string s4 = strings:concat(hello, " ", ballerina);
```
MaryamZi: Shall we change this as follows as discussed offline?

```suggestion
    string s4 = hello.concat(" ", ballerina);
```

Let's import langlibs only if it is not possible to do - as above - without the import. Please check other changes too.

examples/strings/strings.bal
```diff
@@ -22,14 +22,34 @@ public function main() {
...
+    io:println(s4);
```
MaryamZi: Let's prefix what we print with the new examples with 
```
Xxx: 
```
similar to the rest of the examples.

The output would be uniform then.

examples/strings/strings.bal
```diff
@@ -22,14 +22,34 @@ public function main() {
...
+    string|error s6 = strings:fromBytes(bArray);
+    if (s6 is string) {
+        io:println(s6);
+    }
+
     // Removes any leading and trailing white spaces.
-    string s8 = statement.trim();
-    io:println("Trim: " + s8);
+    string s7 = statement.trim();
+    io:println("Trim: " + s7);
```
MaryamZi:
```suggestion
    io:println("Trim: ", s7);
```

examples/strings/strings.bal
```diff
@@ -11,47 +11,60 @@ public function main() {
...
+    // Format a `string` according to the given format arguments.
     string name = "Sam";
     int marks = 90;
     string[] subjects = ["English", "Science"];
     float average = 71.5;
-    string s11 = io:sprintf("%s scored %d for %s and has an average of %.2f.",
+    string s8 = io:sprintf("%s scored %d for %s and has an average of %.2f.",
      name, marks, subjects[0], average);
-    io:println("Sprintf: " + s11);
+    io:println("Sprintf: " + s8);
```
MaryamZi:
```suggestion
    io:println("Sprintf: ", s8);
```

examples/strings/strings.bal
```diff
@@ -11,47 +11,60 @@ public function main() {
...
+    string hello = "Hello";
+    string ballerina = "Ballerina!";
+
+    // Concat multiple strings.
+    string s4 = hello.concat(" ", ballerina);
+    io:println("Concat: ", s4);
+
+    // Join strings with a separator.
+    string s5 = strings:'join(",", hello, ballerina);
+    io:println("Join:", s5);
```
MaryamZi:
```suggestion
    io:println("Join: ", s5);
```

Please fix the .out file too.

examples/strings/strings.bal
```diff
@@ -11,47 +11,60 @@ public function main() {
...
-    io:println("Trim: " + s8);
+    string hello = "Hello";
+    string ballerina = "Ballerina!";
+
+    // Concat multiple strings.
+    string s4 = hello.concat(" ", ballerina);
+    io:println("Concat: ", s4);
+
+    // Join strings with a separator.
+    string s5 = strings:'join(",", hello, ballerina);
```
MaryamZi: This can also be 
```suggestion
    string s5 = ",".'join(hello, ballerina);
```
right?

examples/strings/strings.bal
```diff
@@ -11,47 +11,60 @@ public function main() {
...
+    // Concat multiple strings.
+    string s4 = hello.concat(" ", ballerina);
+    io:println("Concat: ", s4);
+
+    // Join strings with a separator.
+    string s5 = ",".'join(hello, ballerina);
+    io:println("Join: ", s5);
 
-    // Checks whether the given `string` ends with the suffix "Lion".
+    // Convert the `string` to a `byte` array.
```
MaryamZi:
```suggestion
    // Convert `hello` to a `byte` array.
```

examples/strings/strings.bal
```diff
@@ -11,47 +11,60 @@ public function main() {
...
+    byte[] bArray = hello.toBytes();
+
+    // Convert a `byte` array to a `string`.
+    string|error s6 = strings:fromBytes(bArray);
+    if (s6 is string) {
+        io:println("From bytes: ", s6);
+    }
+
+    // Remove any leading and trailing white spaces.
```
MaryamZi:
```suggestion
    // Remove leading and trailing white spaces.
```

MaryamZi: Also `statement` does not have any leading and trailing white spaces right? Shall we declare another variable maybe with leading white spaces and use that instead?

PR #17081, opened 2019-07-26
Add error.stackTrace langlib function

rdhananjaya:
$subject


hasithaa:
There is a check style issue.

hasithaa:
Need to remove the old function from runtime.

PR #17102, opened 2019-07-28
Add the json-access BBE

MaryamZi:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/examples/json-access/json_access.bal
+++ b/examples/json-access/json_access.bal
@@ -1,0 +1,52 @@
+import ballerina/io;
+
+public function main() {
+    // Define a variable of type `json` that holds a mapping value.
+    json j1 = {
+        fname: "Mary",
+        lname: "Jane",
+        address: {
+            line: "20 Palm Grove",
+            city: "Colombo 03",
+            country: "Sri Lanka"
+        }
+    };
+
+    // Field access is allowed on the `json` typed variable, but the return
+    // type would be a union of `json` and `error`. An error will be returned
+    // either if the value on which the access is done is not a mapping or if
+    // it is a mapping, but it does not contain the particular key.
+    json|error r1 = j1.fname;
+    // `r1` is of type `json` since `j1` is a mapping and contains the field `fname`.
+    io:println(r1);
+
+    // Similarly, chained field access is also allowed for `json`. If an `error`
+    // is encountered at any point during the evaluation, evaluation stops at
+    // that point and the result would be an error.
+    json|error r2 = j1.address.city;
+    // `r2` is of type `json` since `j1` is a mapping with the field `address`, which in
+    // turn is a mapping with the field `city` which will be the resultant value.
+    io:println(r2);
+
+    // Attempting to access a non-existent field would result in an error.
+    json|error r3 = j1.age.year;
+    // `r3` is of type `error` since `j1` does not have a field with the key `age`.
+    io:println(r3);
+
+    // Similarly, optional field access is also allowed on lax types.
+    // The main difference between field and optional field access for lax types
+    // is that when a key is not found in a mapping, while the former returns
+    // an `error`, the latter returns `null` (`()`) instead.
+    // Similar to how errors are lifted with field access, `()` is lifted
+    // for optional field access, allowing chained access.
+    json|error r4 = j1?.age?.year;
+    // `r4` is `()` since `j1` does not have a field with the key `age`.
+    io:println(r4);
+
+    // Moreover, when optional field access is done on `null` (`()`), the result would
+    // also be `()`, whereas for field access it would have been an `error`.
+    json j2 = null;
+    json|error r5 = j2?.name;
+    // `r5` is `()` since `j2` is `()`.
+    io:println(r5);
+}
--- a/examples/json-access/json_access.description
+++ b/examples/json-access/json_access.description
@@ -1,0 +1,5 @@
+// Ballerina defines certain types as lax types, for which static typing rules are less strict.
+// For example field access (`.`) and optional field access (`?.`), which are generally allowed on
+// records and objects for fields that are defined in the type descriptors, are also additionally
+// allowed on lax types, moving some of the type checking to runtime from compile time.
+// `json` is defined to be a lax type along with any `map<T>` where `T` is a lax type.
--- a/examples/index.json
+++ b/examples/index.json
@@ -296,6 +296,10 @@
                 "url": "json-arrays"
             },
             {
+                "name": "JSON Access",
+                "url": "json-access"
+            },
+            {
                 "name": "JSON/Record/Map Conversion",
                 "url": "json-record-map-conversion"
             },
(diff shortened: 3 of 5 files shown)


examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+        fname: "Mary",
+        lname: "Jane",
+        address: {
+            line: "20 Palm Grove",
+            city: "Colombo 03",
+            country: "Sri Lanka"
+        }
+    };
+
+    // Field access is allowed on the `json` typed variable, but the return
```
praneesha: Field access is allowed on the `json`-typed variable. However, the return

examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+    // type would be a union of `json` and `error`. An error will be returned
+    // either if the value on which the access is done is not a mapping or if
+    // it is a mapping, but it does not contain the particular key.
```
praneesha: it is mapping. However, it does not contain the particular key.

**Not clear why mapping is used as a noun here**

examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+    json|error r1 = j1.fname;
+    // `r1` is of type `json` since `j1` is a mapping and contains the field `fname`.
```
praneesha: `r1` is of the type `json` since `j1` is mapping and contains the field `fname`.

examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+    io:println(r1);
+
+    // Similarly, chained field access is also allowed for `json`. If an `error`
+    // is encountered at any point during the evaluation, evaluation stops at
+    // that point and the result would be an error.
+    json|error r2 = j1.address.city;
+    // `r2` is of type `json` since `j1` is a mapping with the field `address`, which in
```
praneesha: `r2` is of the type `json` since `j1` maps with the field `address`, which in

examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+        address: {
+            line: "20 Palm Grove",
+            city: "Colombo 03",
+            country: "Sri Lanka"
+        }
+    };
+
+    // Field access is allowed on the `json` typed variable, but the return
+    // type would be a union of `json` and `error`. An error will be returned
+    // either if the value on which the access is done is not a mapping or if
```
praneesha: either if the value on which the access is done is not mapping or if

examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+    json|error r1 = j1.fname;
+    // `r1` is of type `json` since `j1` is a mapping and contains the field `fname`.
+    io:println(r1);
+
+    // Similarly, chained field access is also allowed for `json`. If an `error`
+    // is encountered at any point during the evaluation, evaluation stops at
+    // that point and the result would be an error.
+    json|error r2 = j1.address.city;
+    // `r2` is of type `json` since `j1` is a mapping with the field `address`, which in
+    // turn is a mapping with the field `city` which will be the resultant value.
```
praneesha: turn maps with the field `city`, which will be the resultant value.

examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+    io:println(r2);
+
+    // Attempting to access a non-existent field would result in an error.
+    json|error r3 = j1.age.year;
+    // `r3` is of type `error` since `j1` does not have a field with the key `age`.
```
praneesha: `r3` is of the type `error` since `j1` does not have a field with the key `age`.

examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+    io:println(r3);
+
+    // Similarly, optional field access is also allowed on lax types.
+    // The main difference between field and optional field access for lax types
+    // is that when a key is not found in a mapping, while the former returns
```
praneesha: is that when a key is not found when mapping, while the former returns

examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+    // an `error`, the latter returns `null` (`()`) instead.
+    // Similar to how errors are lifted with field access, `()` is lifted
+    // for optional field access, allowing chained access.
```
praneesha: for optional field access allowing chained access.

examples/json-access/json_access.bal
```diff
@@ -0,0 +1,52 @@
...
+    json|error r4 = j1?.age?.year;
+    // `r4` is `()` since `j1` does not have a field with the key `age`.
+    io:println(r4);
+
+    // Moreover, when optional field access is done on `null` (`()`), the result would
```
praneesha: Moreover, when optional field access occurs on `null` (`()`), the result would

examples/json-access/json_access.description
```diff
@@ -0,0 +1,5 @@
+// Ballerina defines certain types as lax types, for which static typing rules are less strict.
```
praneesha: Ballerina defines certain types as lax types for which static typing rules are not being applied strictly.

examples/json-access/json_access.description
```diff
@@ -0,0 +1,5 @@
...
+// For example field access (`.`) and optional field access (`?.`), which are generally allowed on
```
praneesha: For example, field access (`.`) and optional field access (`?.`), which are generally allowed on

examples/json-access/json_access.description
```diff
@@ -0,0 +1,5 @@
...
+// records and objects for fields that are defined in the type descriptors, are also additionally
+// allowed on lax types, moving some of the type checking to runtime from compile time.
```
praneesha: allowed on lax types. This moves some of the type checking to the runtime from compile time.

(Chunked it because the sentence is too long)

examples/json-access/json_access.bal
MaryamZi: This is pretty much from the spec. https://github.com/ballerina-platform/ballerina-spec/blob/master/lang/lib/value.bal#L85

We've used mappings to represent structural values that map a key to a value - maps, records.

examples/json-access/json_access.bal
MaryamZi: Again, this change isn't correct with our usage of mapping.

examples/json-access/json_access.description
MaryamZi: Not sure why we need this change? Isn't it the same?

praneesha: Chunked it because the sentence is too long

Issue #17100, opened 2019-07-28
Error using .copy() function with an XML comment


irshadnilam:
**Description:**
$title

**Steps to reproduce:**
```
import ballerina/io;

public function main() {
    xml bookComment = xml `<!--some comment-->`;

    // Makes a copy of an XML element.
    xml x = bookComment.copy();
    io:println(x);
}
```
Getting the following error
```
error: {ballerina}XMLOperationError message=Failed to get children from xml: org.apache.axiom.om.impl.dom.CommentImpl: method <init>()V not found
       at main(main.bal:7)
```
[ballerina-internal.log](https://github.com/ballerina-platform/ballerina-lang/files/3439015/ballerina-internal.log)

**Affected Versions:**
Ballerina 0.992.0-m2-SNAPSHOT
Language specification 2019R2
Ballerina tool 1.0.0

**OS, DB, other environment details and versions:**
MacOs 10.14.5
java version "1.8.0_212"
**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #17110, opened 2019-07-29
Fix axiom missing constructor error in distro

rdhananjaya:
Fix https://github.com/ballerina-platform/ballerina-lang/issues/17100


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLItem.java
```diff
@@ -639,20 +636,15 @@ public Object copy(Map<Object, Object> refs) {
                 clonedNode = ((OMElement) omNode).cloneOMElement();
                 break;
             case TEXT:
-                TextImpl text = new TextImpl();
-                text.setTextContent(((OMText) omNode).getText());
-                clonedNode = text;
+                clonedNode = (OMNode) XMLFactory.createXMLText(((OMText) omNode).getText()).value();
```
SupunS: `XMLFactory.createXMLText()` will return the `XMLItem` instance. We can directly return it.
Getting the value of it and creating a another `XMLItem` from it seems redundant and expensive.

Same for below two cases as well.

rdhananjaya: There is this wrapping in OMDocument not before returning thing, is it ok skipping it?
```java 
        // adding the document element as parent, to get xpPaths work
        OMDocument doc = new OMDocumentImpl();
        doc.addChild(clonedNode);
        return new XMLItem(clonedNode);
```

Issue #17089, opened 2019-07-27
Getting java.lang.VerifyError: Bad type on operand stack


irshadnilam:
**Description:**

$title

**Steps to reproduce:**
```
import ballerina/io;

xmlns "http://ballerina.com/aa" as ns0;

public function main() {
    // Creates an XML element, which has attributes that are bound to a namespace as well as ones that are not.
    xml x1 = xml `<ns0:book ns0:status="available" count="5"/>`;
    
    // An attribute can also be accessed using the string representation of the qualified name. 
    string? s = x1@["{http://ballerina.com/aa}status"];
    if (s is string) {
        io:println(s);
    }
}
```

Getting the following error printed on console.

```
ballerina run examples/main.bal                                                                                     1 ↵  1214  09:50:05
java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    main.main(Lorg/ballerinalang/jvm/Strand;)V @1065: putfield
  Reason:
    Type 'java/lang/Object' (current frame, stack[2]) is not assignable to 'java/lang/String'
  Current Frame:
    bci: @1065
    flags: { }
    locals: { 'org/ballerinalang/jvm/Strand', 'org/ballerinalang/jvm/values/XMLValue', 'java/lang/Object', 'java/lang/String', 'java/lang/String', 'java/lang/String', 'java/lang/Object', 'java/lang/Object', 'java/lang/Object', 'java/lang/String', 'java/lang/String', 'java/lang/String', 'java/lang/Object', 'java/lang/String', 'java/lang/String', 'java/lang/String', 'java/lang/String', 'java/lang/Object', 'java/lang/Object', 'java/lang/String', 'java/lang/String', 'java/lang/String', 'java/lang/Object', 'org/ballerinalang/jvm/values/XMLValue', 'java/lang/String', 'java/lang/Object', integer, 'java/lang/Object', 'org/ballerinalang/jvm/values/ArrayValue', long, long_2nd, 'java/lang/Object', 'java/lang/String', 'java/lang/Object', long, long_2nd, 'java/lang/Object', integer, null }
    stack: { 'mainFrame', 'mainFrame', 'java/lang/Object' }
  Bytecode:
    0x0000000: 2a10 00bd 000c b600 122a b400 1699 0007
    0x0000010: b800 1cbf 014c 014d 014e 013a 0401 3a05
    0x0000020: 013a 0601 3a07 013a 0801 3a09 013a 0a01
[...]

PR #17119, opened 2019-07-29
Fix class verify error in xml attrib access

rdhananjaya:
## Purpose
Fix https://github.com/ballerina-platform/ballerina-lang/issues/17089

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
@@ -2317,7 +2317,7 @@ public class BIRGen extends BLangNodeVisitor {
             return keyRegIndex;
         }
 
-        BIRVariableDcl tempQNameVarDcl = new BIRVariableDcl(qnameExpr.type,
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -2317,7 +2317,7 @@ private BIROperand getQNameOP(BLangExpression qnameExpr, BIROperand keyRegIndex)
             return keyRegIndex;
         }
 
-        BIRVariableDcl tempQNameVarDcl = new BIRVariableDcl(qnameExpr.type,
+        BIRVariableDcl tempQNameVarDcl = new BIRVariableDcl(symTable.anyType,
```
hasithaa: why this is anyType ?

rdhananjaya: Couldn't find a corresponding type for `XMLQName`. It's implemented as a RefValue, so used anyType to mimic that.

hasithaa: Ack. An interesting problem, the spec doesn't mention a type for this. I think it should be an XML attribute type. Need to verify with James.

PR #17136, opened 2019-07-30
Fix while loop issue when loop has few instructions

lankavitharana:
## Purpose
Below code was failing even though it generates correct java code. The issue is when the while body has a small number of instructions, then changing the while condition through another worker wouldn't cause the while to break. This happens when the condition is simple boolean variable and we try to change it's value from another worker.

```ballerina
import ballerina/runtime;

boolean a = true;

public function main() {
   worker w1 {
      runtime:sleep(2000);
      a = false;      
   }
   while (a) {
      int b  = 9999;
   } 
}
```
Anyway, this seems to be an issue in java, even the below simple java code is also failing
```java
public class Main {
    public static boolean classLevelVar = true;

    public static int breakWhileUsingClassVar() {
        (new Thread(() -> {
            try {
                Thread.sleep(2000);
            } catch (InterruptedException e) {
            }
            classLevelVar = false;
        })).start();
        while (classLevelVar) {
            int a = 9;
        }
        // Below statement doesn't get hit, making the class level variable volatile solves the issue.
        return 8;
    }
}
```

Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


anupama-pathirage:
Closing this until we get the fix fully done after checking with @lankavitharana .

Issue #15147, opened 2019-04-24
Need an option for HTTP client endpoint to disable the SSL Certificate Validation.


hejunchen:
**Description:**
I am coding in Ballerina to make HTTP request against an internal web services. The internal web service has self-signed SSL certificate.

I can add that self-signed certificate into my truststore.p12 file, but I want to see if there is an option to allow me disable the SSL validation?

**Steps to reproduce:**
![image](https://user-images.githubusercontent.com/6847102/56674390-cfc8f480-666e-11e9-886a-d173707b7456.png)

I don't want to attach the p12 file for each application written in Ballerina, neither I want to manually add each Self-Signed SSL Cert to Ballerina's Home directory's p12 truststore. There shall be an easy way to disable the SSL Cert. If we are writing Ballerina code for integration parts, there are a lot of chances that the code is only wiring up with the internal service. Self-Signed SSL Cert is widely used if the application is only implemented inside company's internal network. 

other languages such as Python allows to disable the SSL validation via an optional parameter

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**
SSL, Self-Signed Certificate

Bhashinee:
@hejunchen Thanks for reporting this issue. 
We are discussing [this](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!searchin/ballerina-dev/Disable$20SSL$20validation$20check%7Csort:date/ballerina-dev/lMG516IV6lM/YNU1mPGbAgAJ) with the team. Hopefully this feature will be added in a future release.

PR #17138, opened 2019-07-30
Add a configuration to disable ssl

Bhashinee:
## Purpose
$subject. This will fix https://github.com/ballerina-platform/ballerina-lang/issues/15147

### DIFF
--- a/tests/jballerina-integration-test/src/test/resources/http/src/httpservices/33_disable_https_service.bal
+++ b/tests/jballerina-integration-test/src/test/resources/http/src/httpservices/33_disable_https_service.bal
@@ -1,0 +1,45 @@
+// Copyright (c) 2019 WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+//
+// WSO2 Inc. licenses this file to you under the Apache License,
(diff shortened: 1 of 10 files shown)


tests/jballerina-integration-test/src/test/resources/http/src/httpservices/33_disable_https_service.bal
```diff
@@ -0,0 +1,45 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/http;
+import ballerina/io;
+
+http:ServiceEndpointConfiguration serviceConf = {
+    secureSocket: {
```
wggihan: Don't we need disable:true here?

Bhashinee: Disabling is only applicable to client side. So we don't need it here.

hejunchen:
thanks, this is a handy configuration that I have been waiting for.

PR #17156, opened 2019-07-30
 Improve type inferring with type param. 

### LINKED ISSUE #16401: Type params not supported for arrow functions
Consider the following program:
```ballerina
function testInvalidIterableOpChain() {
    int[] arr = [10, 20, 30, 40];
    string s = arr.'map(v => v/10).reduce(function (string accum, string val) returns string {
        return accum + val;
    }, "");
}
```
In the above program, the type of `v` is `any|error`, instead of `int`

### LINKED ISSUE #16540: Arrow function doesn't play well with @typeparams
**Description:**
```ballerina
function testFilterIterable() returns float {
    int[6] numberArray = [5, 6, 7, 8, 9, 10];

    index = 0;

    [int, int][] arr = numberArray.'map(function (int num) returns [int, int] {
        [int, int] value = [index, num];
        index += 1;
        return value;
    });
    var arr2 = arr.filter(num => num[0] >= 3)
    .'map(num => num[1]);

    float avg = floats:fromBitsInt(ints:sum(...arr))/floats:fromBitsInt(arr.length());
    return avg;
}
````

In `var arr2 = arr.filter(num => num[0] >= 3)` `num` is inferred to be a any|error, where as it should be [int, int]

### LINKED ISSUE #17070: Can not use array constructor in reduce function
**Description:**
```ballerina
import ballerina/io;
    int[] arr = [1, 2, 3, 4, 5];
    var res = arr.reduce(function (int[] a, int i) returns int[] {
        a.push(i);
        return a;
    }, []);
    io:println(res);
}
```

this produce error:
error: .::arrayReduce.bal:8:8: incompatible types: expected '(any|error)', found 'other'

hasithaa:
Fixes #16540
Fixes #17070
Fixes #16401


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1078,11 +1078,14 @@ warning.usage.of.deprecated.function=\
 error.object.outside.methods.not.allowed=\
   outside object method definitions are not allowed
 
+error.cannot.infer.type=\
+  cannot infer type here
```
pubudu91: Can't we be more specific in this error message? Like give the invokable type signature for which the compiler failed to infer the type? Also shall we rephrase this to something like `failed to infer type` or `failed to infer type of arrow expression`?

hasithaa: This can fail for record literals and array literals as well. Let's say we have a function like this in `lang.string` module

```
@TypeParam type T1 string;
@TypeParam type T2 anydata[];

function foo(T1 srt, T2 v) returns [T1,T2];

// other package
string ss = "abc";
var x = ss.foo([]); // Here we can't infer the type.
```

Ref to the error message: I used the same pattern where we start with "cannot" word for error messages. Yes, I think we can improve messages better.

PR #17197, opened 2019-07-31
Fix gen client generation in openApi

DimalChandrasiri:


misc/openapi-ballerina/modules/openapi-to-ballerina-generator/src/main/java/org/ballerinalang/openapi/CodeGenerator.java
```diff
@@ -122,7 +138,7 @@ public void generate(GenType type, String definitionPath, String serviceName, St
      */
     public void generate(GenType gt, String definitionPath, String outPath)
             throws IOException, BallerinaOpenApiException {
-        generate(gt, definitionPath, outPath, null);
+        generate(gt, definitionPath, null , outPath);
```
NipunaMarcus: @DimalChandrasiri any idea why we put service name as null?

DimalChandrasiri: This is a common method and the service name is required in the service generation path and is handled to use the service name when executed on the service generation path.

PR #17200, opened 2019-07-31
Update README with new messaging

Samallama:
Edited the beginning of the README file with the new messaging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,16 +7,19 @@
     [![codecov](https://codecov.io/gh/ballerina-platform/ballerina-lang/branch/master/graph/badge.svg)](https://codecov.io/gh/ballerina-platform/ballerina-lang)
   [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
 
-Ballerina is a compiled, type safe, concurrent programming language designed to make it simple to write microservices that integrate APIs.
+Ballerina is a general-purpose programming language with features optimized for network distributed applications.
 
-#### Integration Syntax
-A compiled, transactional, statically and strongly typed programming language with textual and graphical syntaxes. Ballerina incorporates fundamental concepts of distributed system integration and offers a type safe, concurrent environment to implement microservices.
+#### Services
+It has first-class language constructs for providing and consuming services,
 
-#### Networked Type System
-A type system that embraces network payload variability with primitive, object, union, and tuple types.
+#### Sequence diagrams
+The structure of functions and programs allows them to be viewed as sequence diagrams showing the interactions between concurrently executing parts.
 
-#### Concurrency
-An execution model composed of lightweight parallel worker units that are non-blocking where no function can lock an executing thread manifesting sequence concurrency. 
+#### Structural typing
+It allows for looser coupling between distributed components and eliminates the friction of data binding.
+
+#### Metadata
+Extensible metadata enables easy integration of Ballerina programs with cloud platforms. 
  
 ## Table of contents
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,16 +7,19 @@
     [![codecov](https://codecov.io/gh/ballerina-platform/ballerina-lang/branch/master/graph/badge.svg)](https://codecov.io/gh/ballerina-platform/ballerina-lang)
   [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
 
-Ballerina is a compiled, type safe, concurrent programming language designed to make it simple to write microservices that integrate APIs.
+In the cloud, applications interact with their environment over the network. Ballerina is a general-purpose programming language with features optimized for these network distributed applications. It is easy to write and modify and is suitable for application programmers.
```
praneesha:
```suggestion
In the cloud, applications interact with their environment over the network. Ballerina is a general-purpose programming language with features optimized for these network-distributed applications. It is easy to write and modify and is suitable for application programmers.
```

README.md
```diff
@@ -7,16 +7,19 @@
...
 
-#### Integration Syntax
-A compiled, transactional, statically and strongly typed programming language with textual and graphical syntaxes. Ballerina incorporates fundamental concepts of distributed system integration and offers a type safe, concurrent environment to implement microservices.
+#### Services
+It has first-class language constructs for providing and consuming services,
```
praneesha:
```suggestion
It has first-class language constructs for providing and consuming services.
```

README.md
```diff
@@ -7,16 +7,19 @@
...
 
-#### Networked Type System
-A type system that embraces network payload variability with primitive, object, union, and tuple types.
+#### Sequence diagrams
+The structure of functions and programs allows them to be viewed as sequence diagrams showing the interactions between concurrently executing parts.
```
praneesha:
```suggestion
The structure of the functions and programs allows them to be viewed as sequence diagrams showing the interactions between the components that execute concurrently.
```

README.md
```diff
@@ -7,16 +7,19 @@
...
 
-#### Concurrency
-An execution model composed of lightweight parallel worker units that are non-blocking where no function can lock an executing thread manifesting sequence concurrency. 
+#### Structural typing
+It allows for looser coupling between distributed components and eliminates the friction of data binding.
```
praneesha:
```suggestion
It allows looser coupling between distributed components and eliminates the friction of data binding.
```

README.md
Samallama: I consulted Rebecca and she said it can be written both ways. Since our tagline and most materials say network distributed (without the hyphen), shall we keep as is?

README.md
Samallama: - First edit: This isn't referring to any functions or programs, it's referring to all. Suggest not to add 'the'.
- Second edit: Suggest to keep as is, because this is the messaging reviewed by the Ballerina team.

README.md
Samallama: "allows" and "allows for" hold two different meanings.  Since this messaging has been approved already shall we keep it as it is?

README.md
praneesha: You can write it in both ways when the term "network-distributed" is used on its own. Here, since it is used as an adjective to describe "applications", we need to say "network-distributed applications."

Anyway, I am -1 to use it as our tagline as it linguistically doesn't sound right to say "network distributed". We do get such a word even when we do a Google search. The correct term is "distributed networks".  I doubt whether it should be "distributed-network applications". To be on the safe side, we can say "applications distributed over a network" to avoid distorting the actual meaning.

README.md
praneesha: - First edit: You need to use "the" before a noun.
- It shouldn't be reviewed linguistically by the Ballerina team as if done it should be done by myself. Saying "concurrently-executing parts" is just shorthand (we need the hyphen there). It is better to say it in the proper way as "the components that execute concurrently."

README.md
praneesha: I am not sure who approved it. For me, it is right to say, "It allows looser coupling " as the meaning becomes straightforward that Ballerina allows it.

PR #17238, opened 2019-07-31
Adapt table query bbe to new ballerina syntax

rdhananjaya:
## Purpose
$subject


hasithaa:
This is fixed in https://github.com/ballerina-platform/ballerina-lang/pull/17239/files

PR #17322, opened 2019-08-02
Add test cases for http2

Bhashinee:
## Purpose
Add test cases for client actions and http2 over ALPN.


arukshani:
Shall we just change the purpose statement to "Add test cases for client actions and http2 over ALPN". Content of the PR is fine. Let's remove the prior knowledge part in the purpose statement since it is only applicable for cleartext

PR #17401, opened 2019-08-05
Fix BBE issues and address suggestions

MaryamZi:
## Purpose
$title.

Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


examples/json-access/json_access.bal
```diff
@@ -12,38 +12,38 @@ public function main() {
...
     };
 
-    // Field access is allowed on the `json` typed variable, but the return
+    // Field access is allowed on the `json`-typed variable. However, the return
     // type would be a union of `json` and `error`. An error will be returned
     // either if the value on which the access is done is not a mapping or if
     // it is a mapping, but it does not contain the particular key.
     json|error r1 = j1.fname;
-    // `r1` is of type `json` since `j1` is a mapping and contains the field `fname`.
+    // `r1` is of the type `json` since `j1` is a mapping and contains the field `fname`.
```
pubudu91: Doesn't "the" sound odd here?

examples/json-access/json_access.bal
```diff
@@ -12,38 +12,38 @@ public function main() {
...
     io:println(r1);
 
     // Similarly, chained field access is also allowed for `json`. If an `error`
     // is encountered at any point during the evaluation, evaluation stops at
     // that point and the result would be an error.
     json|error r2 = j1.address.city;
-    // `r2` is of type `json` since `j1` is a mapping with the field `address`, which in
-    // turn is a mapping with the field `city` which will be the resultant value.
+    // `r2` is of the type `json` since `j1` is a mapping with the field `address`, which in
```
pubudu91: Same comment as above here

examples/json-access/json_access.bal
MaryamZi: These changes are as suggested by @praneesha in https://github.com/ballerina-platform/ballerina-lang/pull/17102/files#r309103462

examples/optional-type/optional_type.bal
```diff
@@ -16,12 +16,12 @@ function getValue(string key) returns string? {
 }
 
 public function main() {
-    // It is optional for `getValue()` to return a value of type `string`. Thus, the value could be either
+    // It is optional for `getValue()` to return a value of the type `string`. Thus, the value could be either
```
pubudu91:
```suggestion
    // It is optional for `getValue()` to return a value of type `string`. Thus, the value could be either
```

examples/range-expressions/range_expressions.bal
```diff
@@ -27,7 +27,7 @@ public function main() {
         if (r is record {| int value; |}) {
             io:println(r.value);
         } else {
-            // `r` is `()`, implying the end of the iteration.
+            // `r` is `()` implying the end of the iteration.
```
pubudu91:
```suggestion
            // `r` being `()` implies the end of the iteration.
```

examples/json-access/json_access.bal
praneesha: We need to have "the" before a noun.

examples/json-access/json_access.bal
praneesha: We need to have "the" before a noun.

examples/json-access/json_access.bal
pubudu91: It just sounds wrong and weird here though. The idea is to say that `r1`'s type is `json`

praneesha: The way it is worded here needs "the".

examples/range-expressions/range_expressions.bal
MaryamZi: I prefer the former. Specifically because this is a side by side explanation of the code.

examples/json-access/json_access.bal
pubudu91: Again, the context matters. In this context, putting "the" here sounds weird and wrong. See https://english.stackexchange.com/questions/57335/a-variable-of-type-int-versus-a-variable-of-an-int-type for example.

examples/range-expressions/range_expressions.bal
praneesha: +1

examples/json-access/json_access.bal
gimantha: https://english.stackexchange.com/questions/358223/use-of-the-after-the-prepostion-of

I think we only have to add "the" if there many "json" types and if we have specifically mentioned one before. but we have only one json type. so we don't need "the" IMO

praneesha: It is exactly why we need to have "the" (because we have only one type of JSON).

sameerajayasoma:
Looks like there are comments that need to be addressed. Please don't merge this PR.

examples/json-access/json_access.bal
MaryamZi: Shall we come to a conclusion on this please? :) I'm OK with either version, but a quick search of something like "of type int" seems to have more results than "of **the** type int".

IMO, the existing form reads OK -  
```
`r1` is of type `json`
```

However, IMO again, if it was phrased as follows, we would need the "the"
```
`r1` is of the `json` type
```

Thoughts please?

pubudu91: IMO, we should be going with what's normally used in programming. The first form is what I have seen in computing related literature/articles. Plus reads easy too.

praneesha: We need to have "the" for grammatical correctness.

anupama-pathirage: Here "`r1` is of type `json`" looks more natural and correct to me. This is what used in any other technical documentation as well. I think the reason is these are not Standard English, but rather technical terms used commonly. So we shouldn't make it odd by deviating common technical documentations. 

Refer below examples (Search for "of type JSON"). All are official technical documentations.
1. https://dev.mysql.com/doc/refman/5.7/en/json.html
2. https://v1.realworldocaml.org/v1/en/html/handling-json-data.html
3. https://www.postgresql.org/docs/9.4/datatype-json.html
4. https://books.google.lk/books?id=9dh6DAAAQBAJ&pg=PA153&lpg=PA153&dq=%22of+type+json%22&source=bl&ots=6Nyed9xHht&sig=ACfU3U0bpP-JCoPSnSMhgfKChtVNuWAnWg&hl=en&sa=X&ved=2ahUKEwiowPCD1JDkAhXFqI8KHWVjCQEQ6AEwCnoECAkQAQ#v=onepage&q=%22of%20type%20json%22&f=false

MaryamZi: Thanks all, going ahead without the "the" since that seems to be how it is commonly used.

PR #17415, opened 2019-08-05
Process Execute Support

lafernando:
This is to support executing commands as sub-processes in Ballerina.


stdlib/system/src/main/ballerina/src/system/Module.md
```diff
@@ -60,5 +60,8 @@ public function main() {
     
     // Copy file ot directory to new path.
     error? results = system:copy(“/A/B/C”, “/A/B/D”, replaceExisting = true);
+
+    // Execute an OS command as a subprocess
+    system:Process|error proc = check system:exec("ls", {}, "/", "-la");
```
wggihan:
```suggestion
    system:Process|system:Error proc = check system:exec("ls", {}, "/", "-la");
```

Issue #11172, opened 2018-11-02
Private types defined in a package can be used outside of the package


pubudu91:
Consider the following type def in a package `foo`.
```ballerina
type Bar string|xml|json;
```
The above type can be used outside of the `foo` package without a problem. For example:
```ballerina
import ballerina/io;
import pubudu/foo;

public function main() {
	json j = {"a":25};
	foo:Bar f = j;
	io:println(f);
}
```
Running the above program prints `{"a":25}`

Related issue: #11170

rdhananjaya:
Reopening as previous PR was reverted in https://github.com/ballerina-platform/ballerina-lang/pull/17603

PR #17575, opened 2019-08-11
Enforce pkg only access for user defined types

rdhananjaya:
## Purpose
Fix https://github.com/ballerina-platform/ballerina-lang/issues/11172


langlib/lang.array/src/main/ballerina/src/lang.array/array.bal
```diff
@@ -18,19 +18,19 @@
 # Has the special semantic that when used in a declaration
 # all uses in the declaration must refer to same type.
 @typeParam
-type Type any|error;
+public type Type any|error;
```
sameerajayasoma: We shouldn't make these types `public` IMO. They are not part of the `lang.array` API. 

Here is an attempt to clarify the scenario. 
```ballerina
// likewise, we can consider records, objects, etc,.
public function foo(T1 t1,.... Tn tn) returns Tr ;
```
`T` belongs to `{T1, ...., Tn, Tr}`

- If `T` has public visibility, then T can be part of a public construct.
- If `T` has module-level visibility, then T can be part of a public construct except for the following cases:
  - `T` is an object with members that have private visibility.
  - `T` is an object with members that have module-level visibility.

PR #17615, opened 2019-08-12
Enforce pkg only access for user defined types

rdhananjaya:
## Purpose
Fix https://github.com/ballerina-platform/ballerina-lang/issues/11172


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -647,6 +651,16 @@ public void visit(BLangTypeDefinition typeDefinition) {
         }
     }
 
+    private int getAccessModifierMask(Set<Flag> flagSet, BLangType typeNode) {
```
sameerajayasoma: Would be able to explain the logic behind this method?

rdhananjaya: If this type is defined to a public type or this is a anonymous type, return int with all bits set to 1, so that we can `bitwise and` it with any flag and the original flag will not change.
If the type is not a public type then return a mask where public flag is set to zero and all others are set to 1 so that we can perform bitwise and operation to remove the public flag from given flag.

Added the same comment to the method. 
When we create the type/tsymbol most of the time we don't have information to decide the public-ness of the type. We have defaulted to use `public` flag there. Then we can use this mask to filter the public-ness when we know for certain that it's a public or private/pkg type.

PR #17801, opened 2019-08-15
 Fix invalid record break issue

wggihan:
## Purpose
$title.


stdlib/io/src/test/resources/test-src/io/csv_io.bal
```diff
@@ -139,3 +161,21 @@ function getTableWithNill(string filePath) returns @tainted [string, string]|err
...
+    string[] keys = [];
+    if (rCsvChannel is io:ReadableCSVChannel) {
+        var tblResult = rCsvChannel.getTable(CommonApp);
+        if (tblResult is table<CommonApp>) {
+            foreach var rec in tblResult {
+                keys.push(rec.appId);
+            }
+            error? closeResult = rCsvChannel.close();
+        }
+        else {
```
chamil321: Shall we move `else` to the above line?

PR #17883, opened 2019-08-16
Enable few disabled jballerina unit tests

irshadnilam:
## Purpose
$title


tests/jballerina-unit-test/src/test/resources/test-src/expressions/lambda/iterable/basic-iterable-with-variable-mutability.bal
```diff
@@ -64,81 +64,75 @@ function testBasicArray2(string[] values) returns string {
...
-//}
+function testBasicMap1() returns [int, string[]] {
+    map<string> m = {a:"A", b:"B", c:"C", d:"D", e:"E"};
+    int count = 0;
+    string[] values = m.keys().'map(function (string key) returns string {
+                                count = count + 10;
+                                var k = key;
+                                return k.toLowerAscii();
+                            })
+                      .filter(function (string v) returns boolean {
```
Shan1024: It is better if this line is aligned with the `.keys`.

@NipunaRanasinghe Please check whether this formatting change can be supported via the IDEA(and LS) in the future.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/lambda/iterable/basic-iterable-with-variable-mutability.bal
```diff
@@ -64,81 +64,75 @@ function testBasicArray2(string[] values) returns string {
...
-//                            return ["", xml ` `];
-//                      });
-//    str = str + "-end";
-//    return [str, m];
-//}
+function testBasicMap1() returns [int, string[]] {
+    map<string> m = {a:"A", b:"B", c:"C", d:"D", e:"E"};
+    int count = 0;
+    string[] values = m.keys().'map(function (string key) returns string {
+                                count = count + 10;
```
Shan1024: Indentation seems to be incorrect here. The indentation should start from where the `function` starts.
Eg - 
```
string[] values = m.keys().'map(function (string key) returns string {
                                    count = count + 10;
```

@NipunaRanasinghe WDYT? Please check whether this formatting change can be supported via the IDEA(and LS) in the future if seems good.

NipunaRanasinghe: @Shan1024 Yeah +1. It would be better if the function body can be indented w.r.t. where the function starts, as Shan has suggested. :) 

@NipunaMarcus WDYT? Should we define these sub-rules in [Ballerina Style Guide](https://ballerina.io/learn/style-guide/definitions/#function-definition)?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/lambda/iterable/basic-iterable-with-variable-mutability.bal
NipunaRanasinghe: @Shan1024 +1. To be consistent, its better if we can try to indent w.r.t. the beginning of the RHS of the assignment statement, as below.
```
function foo() {
    var x = variableX.foo().bar()
                .match()
                .filter();
}
```

@NipunaMarcus  WDYT? Do we have already have predefined rules when formatting the assignment statements, in our formatting guide?,

NipunaMarcus: @NipunaRanasinghe yes we already have style guide rules on how to format assignment and how to format line breaks

tests/jballerina-unit-test/src/test/resources/test-src/expressions/lambda/iterable/basic-iterable-with-variable-mutability.bal
NipunaMarcus: @NipunaRanasinghe we don't have this specific rule ... we have a general rule how we should handle line breaks ... when a line break happens we indent the broken line with four spaces related to the parent of it.

hasithaa:
@irshadnilam Seems like there is a check style issue.

irshadnilam:
> @irshadnilam Seems like there is a check style issue.

Fixed.

PR #17973, opened 2019-08-19
Return anonymous record from table#getNext instead of any type

### LINKED ISSUE #17526: Table getNext() does not return the parameterized type
The following example:

```ballerina
    table<Product> dataTable = check clientDB->select("SELECT * FROM PRODUCT 
                                                                                        WHERE id = ?", Product, param);
    if (dataTable.hasNext()) { 
        return Product.constructFrom(dataTable.getNext());
```
The above code results in a compile error, since "dataTable.getNext()" returns "any" rather than returning a "Product" value. This works for doing a foreach iteration of the table, but not here. Specially in the time of checking for a single value, we have to use the above approach. This is not consistent behavior in the language. For the above to get to work, we have to do an explicit cast there.

Manuri:
## Purpose
Return anonymous record from table#getNext instead of any type

Fixes #17526

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/jdbc/src/test/resources/test-src/sql/table/table_type.bal
+++ b/stdlib/jdbc/src/test/resources/test-src/sql/table/table_type.bal
@@ -1423,6 +1423,40 @@ function testMultipleRowsWithForEach() returns @tainted [int, int] {
     return [rs1.INT_TYPE, rs2.INT_TYPE];
 }
 
+function testGetPrimitiveTypesWithWhileLoopAndConstructFrom() returns @tainted [int, int, float, float, boolean, string, decimal] {
+    jdbc:Client testDB = new({
+        url: "jdbc:h2:file:./target/tempdb/TEST_DATA_TABLE_H2",
+        username: "SA",
+        password: "",
+        poolOptions: { maximumPoolSize: 1 }
+    });
+
+    var selectRet = testDB->select("SELECT int_type, long_type, float_type, double_type,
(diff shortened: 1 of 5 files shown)


stdlib/jdbc/src/test/resources/test-src/sql/table/table_type.bal
```diff
@@ -1423,6 +1423,40 @@ function testMultipleRowsWithForEach() returns @tainted [int, int] {
     return [rs1.INT_TYPE, rs2.INT_TYPE];
 }
 
+function testGetPrimitiveTypesWithWhileLoopAndConstructFrom() returns @tainted [int, int, float, float, boolean, string, decimal] {
+    jdbc:Client testDB = new({
+        url: "jdbc:h2:file:./target/tempdb/TEST_DATA_TABLE_H2",
```
wggihan: This type of hardcoded jdbc urls will fail this test on windows. Do you have a plan to fix this it?

anupama-pathirage: I fixed it with https://github.com/ballerina-platform/ballerina-lang/pull/18020

Manuri: This is fixed in Anupama's PR which refactors tests.

PR #17994, opened 2019-08-19
Fix faulty unique symbol check

### LINKED ISSUE #17992: Incorrect redeclared symbol check logic for lambdas
The below check is incorrect. The arguments should be swapped and the flags of the owner of the symbol should be checked.
https://github.com/ballerina-platform/ballerina-lang/blob/df71d42b7ccf294e8804f053ba8f8d024ab3f8ca/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java#L262

pubudu91:
Fixes #17992

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
@@ -218,12 +218,18 @@ public class SymbolResolver extends BLangNodeVisitor {
             return true;
         }
 
+        // Type names should be unique and cannot be shadowed
+        if ((foundSym.tag & SymTag.TYPE) == SymTag.TYPE) {
+            dlog.error(pos, DiagnosticCode.REDECLARED_SYMBOL, symbol.name);
+            return false;
+        }
+
         if (isSymbolDefinedInRootPkgLvl(foundSym)) {
             dlog.error(pos, DiagnosticCode.REDECLARED_BUILTIN_SYMBOL, symbol.name);
             return false;
         }
 
-        if (isSymbolOwnersSame(symbol, foundSym)) {
+        if (hasSameOwner(symbol, foundSym)) {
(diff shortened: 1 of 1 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -168,15 +168,15 @@ public boolean checkForUniqueSymbol(DiagnosticPos pos, SymbolEnv env, BSymbol sy
         }
 
         // if a symbol is found, then check whether it is unique
-        return isSameSymbol(pos, symbol, foundSym);
+        return !isDuplicateSymbol(pos, symbol, foundSym);
```
hasithaa: All the usage of `isDuplicateSymbol` starts with a `!` hence, Consider renaming this.

hasithaa:
Shall we add test cases as well?

pubudu91:
The test cases are already there in https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/jballerina-unit-test/src/test/resources/test-src/variable/shadowing/variable-shadowing-negative.bal

It's just that even with the previous incorrect check, one couldn't write code in a way which went in the wrong branch of the check. It would have had to be something like below:
```ballerina
    string '\$lambda\$0 = "foo";

    var fn = function () returns string {
        return "bar";
    };
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
pubudu91: renamed and inverted to isDistinctSymbol()

PR #18015, opened 2019-08-20
Panic when attempt to change size of a tuple

rdhananjaya:
## Purpose
$subject.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValue.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValue.java
@@ -1072,8 +1072,8 @@ public class ArrayValue implements RefValue, CollectionValue {
     public void setLength(long length) {
         handleFrozenArrayValue();
         int newLength = (int) length;
-        rangeCheck(length, size);
         checkFixedLength(length);
+        rangeCheck(length, size);
         fillerValueCheck(newLength, size);
         resizeInternalArray(newLength);
         fillValues(newLength);
@@ -1081,6 +1081,10 @@ public class ArrayValue implements RefValue, CollectionValue {
     }
 
     private void checkFixedLength(long length) {
+        if (arrayType != null && arrayType.getTag() == TypeTags.TUPLE_TAG) {
+            throw BLangExceptionHelper.getRuntimeException(BallerinaErrorReasons.INHERENT_TYPE_VIOLATION_ERROR,
+                    RuntimeErrors.CANNOT_CHANGE_TUPLE_SIZE);
+        }
         if (((BArrayType) this.arrayType).getState() == ArrayState.CLOSED_SEALED) {
             throw BLangExceptionHelper.getRuntimeException(BallerinaErrorReasons.INHERENT_TYPE_VIOLATION_ERROR,
                     RuntimeErrors.ILLEGAL_ARRAY_SIZE, length);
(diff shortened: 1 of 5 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValue.java
```diff
@@ -1072,15 +1072,19 @@ public IteratorValue getIterator() {
...
         checkFixedLength(length);
+        rangeCheck(length, size);
         fillerValueCheck(newLength, size);
         resizeInternalArray(newLength);
         fillValues(newLength);
         size = newLength;
     }
 
     private void checkFixedLength(long length) {
+        if (arrayType != null && arrayType.getTag() == TypeTags.TUPLE_TAG) {
```
MaryamZi: Not all tuples should fail right? Shouldn't we be able to remove all from a tuple with only a rest field?

```
[int...] i = [2, 3];
i.removeAll(); // this is valid?
```

langlib/langlib-test/src/test/resources/test-src/arraylib_test.bal
```diff
@@ -249,3 +249,21 @@ function testRemoveAllFixedLengthArray() returns int[] {
     ar.removeAll();
     return ar;
 }
+
+function testTuppleResize() returns [int, string] {
```
MaryamZi:
```suggestion
function testTupleResize() returns [int, string] {
```

Several other occurrences too.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValue.java
rdhananjaya: My understanding is that in above example [2,3] is a tuple value of type [int, int] and removing/setlength changes the type/shape of the value, IMO is illegal.

MaryamZi: The type is still `[int...]` right? The shape is `[int, int]`. Updates that change the shape can be done as long as they don't violate the inherent type (given the value is not immutable).

IMO, `[int...]` is the same as `int[]`.

rdhananjaya: Create a spec issue to clarify https://github.com/ballerina-platform/ballerina-spec/issues/309

bvm/ballerina-runtime/src/main/resources/MessagesBundle.properties
```diff
@@ -166,3 +166,4 @@ illegal.tuple.insertion = tuple of length {0} cannot be expanded into tuple of l
 unexpected.xml.type = lang.xml function ''{0}'' only accept xml type ''{1}''
 java.null.reference = handle refers to Java null
 illegal.array.size = cannot change length of fixed length array to ''{0}''
+cannot.change.tuple.size = cannot change tuple size
```
MaryamZi: Can we make this consistent with the array error message?

Issue #12618, opened 2018-12-07
Write an integration test for HTTP caching


pubudu91:
Need to write test scenarios for following cache controls.
- [ ] max-age - Wait until the cache expires and see whether client gets the response after calling the server without getting it from cache.
- [ ] no-cache - Before delivering the cached response, always need to check with the back-end server.
- [ ] no-store - Should not deliver the cached response. Always call the back-end server and retrieve the response (Need to fully download the response).
- [ ] Mark the response as 'private'. Response can be only cached by a single user.  
- [ ] must-revalidate - If the response is stale, re-validate with the back-end. Otherwise send the cached response.
- [ ] proxy-revalidate 
- [ ] privateFields 
- [ ] noCacheFields 
- [ ] no-transform

pubudu91:
Added with #17687

pubudu91:
Closing since more tests were added with PR #18114

PR #18114, opened 2019-08-21
Add test cases for http-cache

Bhashinee:
## Purpose
$subject.
Related issue: https://github.com/ballerina-platform/ballerina-lang/issues/12618

Include test cases for following cache controls.
- max-age
- must-revalidate
- no-cache


tests/jballerina-integration-test/src/test/java/org/ballerinalang/test/service/http/sample/HTTPCachingTestCase.java
```diff
@@ -61,4 +61,45 @@ public void testPassthroughServiceByBasePath() throws IOException, InterruptedEx
...
+
+    @Test(description = "Test max-age cache control")
+    public void testMaxAgeCacheControl() throws IOException, InterruptedException {
+        HttpResponse response = HttpClientRequest.doGet(serverInstance.getServiceURLHttp(9245, "maxAge"));
+        assertEquals(response.getData(), "public,max-age=5");
+
+        response = HttpClientRequest.doGet(serverInstance.getServiceURLHttp(9245, "maxAge"));
+        assertEquals(response.getData(), "public,max-age=5");
+
+        Thread.sleep(5000);
```
shafreenAnfar: Guys, for tests with thread sleep, shall we follow the convention mentioned bellow,
https://martinfowler.com/articles/nonDeterminism.html

tests/jballerina-integration-test/src/test/java/org/ballerinalang/test/service/http/sample/HTTPCachingTestCase.java
```diff
@@ -61,4 +61,45 @@ public void testPassthroughServiceByBasePath() throws IOException, InterruptedEx
...
+    @Test(description = "Test must-revalidate cache control")
+    public void testMMustRevalidateCacheControl()
+            throws IOException {
+        HttpResponse response = HttpClientRequest.doGet(serverInstance.getServiceURLHttp(9247, "mustRevalidate"));
+        assertEquals(response.getData(), payload);
+        assertEquals(response.getHeaders().get(serviceHitCount), "1");
+
+        response = HttpClientRequest.doGet(serverInstance.getServiceURLHttp(9247, "mustRevalidate"));
+        assertEquals(response.getData(), payload);
+        assertEquals(response.getHeaders().get(serviceHitCount), "2");
```
pubudu91: This doesn't really capture what `must-revalidate` directive does. There should be 3 requests minimum: the initial request, another request sent before the cache entry expires, a 3rd request sent after the cache entry expires. So at the end of 3 requests, the service hit count should be 2. And the proxy hit count should be 3

tests/jballerina-integration-test/src/test/java/org/ballerinalang/test/service/http/sample/HTTPCachingTestCase.java
Bhashinee: In here, Thread.sleep is not added to wait for a response. We are sending the request after 5 seconds. So that the max-age would expire, and we need to test the response after the cache expiration.

PR #18117, opened 2019-08-21
Remove IO specific core based thread pool

### LINKED ISSUE #15814: Missing info for io:ReadableCharacterChannel object methods
The documentation of the methods in io:ReadableCharacterChannel object are incomplete. For example, the doc comments for the `read()` method just simply says that it reads the specified no. of characters. It doesn't specify what the behaviour in following cases:
- When the chars to read is greater than actual available chars to read.
- What's returned when EOF is reached.

### LINKED ISSUE #17276: Deadlock/Starvation in Ballerina I/O Blocking Calls
If we do a blocking I/O call in Ballerina, such as doing a blocking read on a io:ReadableCharacterChannel, this will use a full thread, that is defined in the thread pool at [1]. For example, in a 4 core CPU, this thread pool size would be 1. So if one blocking call happens, all other I/O calls in the Ballerina process will also block, and not continue. So if by any chance, the first I/O operation is blocked due to another operation whose I/O cannot proceed due to a lack of I/O thread, this situation is effectively deadlocked. 

This needs to be fixed in a way, the same thread pool at [1] should not be used, but rather, probably an unbounded thread pool should be used for those scenarios. 

Also, the general Ballerina I/O channel read/write operations should be checked to do direct Java's non-blocking channel operations in the same thread, without doing additional thread context changes. 

[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/io/src/main/java/org/ballerinalang/stdlib/io/events/EventManager.java

### LINKED ISSUE #17846: IO error for file not found is not user friendly
When you try to open a non-existent file for reading, it returns the following error:
```
{ballerina/io}GenericError message=Failed to open file: file not found: ./nofile cause=null
```
With this error, it is not easy to do anything based on the type of the error. For example, in my case, I'm getting a file name from the user and try to open that file for reading. And if the file was not found, I want to convey the same message to the user, saying the this particular file you requested could not be found. But with the above structure of the error, there's no easy way to do it. What I suggest is a structure like as follows:
```
{ballerina/io}FileNotFoundError message=Failed to open file filename=nofile path=path/to/file cause=null
```
With a structure like above, I can easily filter out file not found errors by checking the error reason and then get the `filename` from the error details and construct my own message.

wggihan:
Fixes #17276
Fixes #15814
Fixes #17846

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/test/StartServiceSkeleton.java
```diff
@@ -80,7 +79,7 @@ public void execute(Context ctx) {
             // TODO: find how to give the service name in to service generation function.
             generator.generateService(userDir, openApiFilePath, "", rootDir.toString());
         } catch (IOException | BallerinaOpenApiException e) {
-            throw new BallerinaIOException(String.format("Service skeleton creation failed. Failed to generate the "
+            throw new BallerinaException(String.format("Service skeleton creation failed. Failed to generate the "
```
riyafa: We can throw ErrorValue? It was recommended to be used if I am not mistaken.

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/channels/base/Buffer.java
```diff
@@ -151,12 +152,11 @@ public void reverse(int count) throws BallerinaIOException {
             if (reversedByteBufferPosition >= minimumBytePosition) {
                 byteBuffer.position(reversedByteBufferPosition);
             } else {
-                String message = "The specified byte count " + count + " has not being read,hence cannot reverse";
+                String message = "The specified byte count " + count + " has not being read, hence cannot reverse";
```
riyafa:
```suggestion
                String message = "The specified byte count " + count + " is not being read, hence cannot be reversed";
```

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/channels/base/CharacterChannel.java
```diff
@@ -286,38 +291,34 @@ private void processChars(int numberOfCharsRequired, ByteBuffer buffer, int numb
...
-            //We need to ensure that the required amount of characters are available in the buffer
-            if (charBuffer.limit() < charsRequiredToBeReadFromChannel) {
-                //This means the amount of chars required are not available
-                charsRequiredToBeReadFromChannel = charBuffer.limit();
-            }
-            appendCharsToString(content, charsRequiredToBeReadFromChannel);
-        } catch (IOException e) {
-            throw new IOException("Error occurred while reading characters from buffer", e);
+    public String read(int numberOfCharacters) throws BallerinaIOException {
+        //Will identify the number of characters required
```
riyafa:
```suggestion
        // Identify the number of characters required
```

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/channels/base/CharacterChannel.java
```diff
@@ -286,38 +291,34 @@ private void processChars(int numberOfCharsRequired, ByteBuffer buffer, int numb
...
+        int charsRequiredToBeReadFromChannel;
+        StringBuilder content = new StringBuilder(numberOfCharacters);
+        int numberOfBytesRequired = numberOfCharacters * MAX_BYTES_PER_CHAR;
+        //First the remaining buffer would be get and the characters remaining in the buffer will be written
```
riyafa:
```suggestion
        // Initially the remaining buffer would be obtained and the characters remaining in the buffer will be written
```

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/channels/base/CharacterChannel.java
```diff
@@ -286,38 +291,34 @@ private void processChars(int numberOfCharsRequired, ByteBuffer buffer, int numb
...
+        if (charsRequiredToBeReadFromChannel == 0) {
+            //This means there's no requirement to get the characters from channel
+            return content.toString();
+        }
+        if (log.isDebugEnabled()) {
+            log.debug(String.format("number of chars required to be get from the channel %d",
+                    charsRequiredToBeReadFromChannel));
         }
+        asyncReadBytesFromChannel(numberOfBytesRequired, numberOfCharacters);
+        //We need to ensure that the required amount of characters are available in the buffer
```
riyafa:
```suggestion
        // Ensure that the required amount of characters are available in the buffer
```

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/CloseWritableCharacterChannel.java
```diff
@@ -52,25 +49,14 @@
...
-        Register register = EventRegister.getFactory().register(closeEvent,
-                                                                CloseWritableCharacterChannel::closeChannel);
-        eventContext.setRegister(register);
-        register.submit();
+        try {
+            charChannel.close();
+        } catch (ClosedChannelException e) {
+            return IOUtils.createError("channel already closed.");
+        } catch (IOException e) {
+            return IOUtils.createError(e.getMessage());
```
riyafa: Since in several places we are using `IOUtils.createError(e.getMessage());` Wouldn't it make sense to have a util function that takes a throwable?

misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/test/StartServiceSkeleton.java
```diff
@@ -111,7 +110,7 @@ private void initTempDir(String sourceRoot) {
             try {
                 Files.createDirectories(projectRoot);
             } catch (IOException e) {
-                throw new BallerinaIOException(String.format("Service skeleton creation failed. Failed to create " +
+                throw new BallerinaException(String.format("Service skeleton creation failed. Failed to create " +
```
Manuri: We shouldn't be throwing `BallerinaException`. If this logic is somewhere deep down in the module I'd suggest introducing a separate exception type and then catch it at an outer level and throw `ErrorValue`

stdlib/io/src/main/ballerina/src/io/io_error.bal
```diff
@@ -25,4 +25,13 @@ public type ConnectionTimedOutError error<CONNECTION_TIMED_OUT, Detail>;
...
 
-public type Error GenericError|ConnectionTimedOutError;
+public const ACCESS_DENIED_ERROR = "{ballerina/io}AccessDeniedError";
+public type AccessDeniedError error<ACCESS_DENIED_ERROR, Detail>;
+
+public const FILE_NOT_FOUND_ERROR = "{ballerina/io}FileNotFoundError";
+public type FileNotFoundError error<FILE_NOT_FOUND_ERROR, Detail>;
+
+public const END_OF_FILE_ERROR = "{ballerina/io}EoF";
+public type EoFError error<END_OF_FILE_ERROR, Detail>;
```
Manuri: According to the convention shouldn't this be `EofError`?

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/channels/TempFileIOChannel.java
```diff
@@ -46,8 +45,8 @@ public void close() throws IOException {
             File tempFile = new File(tempFilePath);
             boolean isFileDeleted = tempFile.delete();
             if (!isFileDeleted) {
-                throw new BallerinaException("Temporary file deletion failure occurred while closing " +
-                        "TempFileIOChannel! ");
+                throw new BallerinaException(
```
Manuri: We should be throwing `ErrorValue`.

stdlib/io/src/main/ballerina/src/io/io_error.bal
Manuri: When I did a quick search what I saw was, in python, there is `EOFError`[1],
in Java there is `EOFException`[2] and in Rust `UnexpectedEof`.

If the error they intend is the same as our `EoFError`, AFAIU we are good to use our convention.

[1] https://docs.python.org/3/library/exceptions.html
[2] https://docs.oracle.com/javase/7/docs/api/java/io/EOFException.html
[3] https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.UnexpectedEof

Issue #18074, opened 2019-08-21
Absence of rest field marked with null in BTupleType


pubudu91:
For tuples, `null` is used to identify the absence of a tuple-rest-descriptor. Ideally, NoType should've been used to indicate this, as is done with records.

hasithaa:
NoType is used for inferring, Ideally, we should use NIL type here.

pubudu91:
But that’s a type itself. Adding setting the rest descriptor to nil would mean that the tuple has a nil rest descriptor

MaryamZi:
We can now use `never`.

MaryamZi:
> We can now use `never`.

But if we do this we wouldn't be able to differentiate between the absence of a rest descriptor and explicitly using `never` as the rest descriptor. IMO, using `null` here may be OK.

Issue #17192, opened 2019-07-31
Types.isJSONContext fails with a Bad Sad Error


hasithaa:
Try following program

```
function get() returns [string, int, boolean]|string {
    return ["",5,true, false];
}
```

```
[2019-07-31 12:13:22,656] ERROR {org.ballerinalang.tool.Main} - null
java.lang.NullPointerException
        at org.wso2.ballerinalang.compiler.semantics.analyzer.Types.isJSONContext(Types.java:179)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.setLiteralValueAndGetType(TypeChecker.java:275)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.visit(TypeChecker.java:264)
        at org.wso2.ballerinalang.compiler.tree.expressions.BLangLiteral.accept(BLangLiteral.java:51)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.checkExpr(TypeChecker.java:250)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.checkExpr(TypeChecker.java:217)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.visit(TypeChecker.java:642)
        at org.wso2.ballerinalang.compiler.tree.expressions.BLangListConstructorExpr.accept(BLangListConstructorExpr.java:51)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.checkExpr(TypeChecker.java:250)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.checkExpr(TypeChecker.java:217)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.visit(SemanticAnalyzer.java:2177)
        at org.wso2.ballerinalang.compiler.tree.statements.BLangReturn.accept(BLangReturn.java:44)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.analyzeNode(SemanticAnalyzer.java:2225)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.analyzeNode(SemanticAnalyzer.java:2190)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.analyzeStmt(SemanticAnalyzer.java:2186)
[...]

PR #18153, opened 2019-08-23
Improve type checking in tuple types

### LINKED ISSUE #18076: When LHS is a union with a tuple type and RHS is a tuple, the sizes don't get checked
The following program works
```ballerina
import ballerina/io;

public function main() {
    var x = get();
    io:println(x);

    [string, int, boolean]|string tup = ["foo", 5, true, false]; 
    io:println(tup);
}

function get() returns [string, int, boolean]|string {
    return ["foo",5,true, false];
}
```
And prints
```
foo 5 true false
foo 5 true false
```

grainier:
## Purpose
$subject, to check tuple size.

Fixes #18076 

## Samples
```ballerina
function testTupleInUnionReturn() returns [string, int, boolean]|string {
    return ["", 5, true, false]; // this should fail
}
```

## Remarks
#18076 
#17192
#18074

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -643,9 +643,6 @@ public class TypeChecker extends BLangNodeVisitor {
                             actType = checkExpr(listConstructor.exprs.get(i), env, expType);
                             results.add(expType.tag != TypeTags.NONE ? expType : actType);
                         } else {
-                            // TODO: 8/21/19 The fix below is a temporary one. The rest type should be NoType if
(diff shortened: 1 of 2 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -646,8 +646,15 @@ public void visit(BLangListConstructorExpr listConstructor) {
                             // TODO: 8/21/19 The fix below is a temporary one. The rest type should be NoType if
```
hasithaa: This comments no longer apply, isn't it ?

grainier: Removed in 000740f

Issue #17001, opened 2019-07-23
`.toString()` returns null for `()` values in structures


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
Code
```ballerina
import ballerina/io;

type Foo record {
    string? s = ();
};

public function main() {
    Foo f = {};

    io:println(f);
    io:println(f.toString());
}
```

Output
```cmd
$ ballerina run err_test.bal
s=null
s=null
```

**Affected Versions:**
0.992.0-m8

PR #18154, opened 2019-08-23
Add toString to object and Fix tests

### LINKED ISSUE #17003: Can't call `.toString()` on an object
**Description:**
$title.

`io:println(objectValue)` works, but prints an invalid representation.

According to the spec, `.toString()` for an object is as follows:
```
# - if `v` is an object, then
#     - if `v` provides a toString method with a string return type and no required methods,
#       then the result of calling that method on `v`
#     - otherwise, `object` followed by some implementation-dependent string
```

**Steps to reproduce:**
```ballerina
import ballerina/io;

public type Foo object {
    public string s = "xyz";
};

public function main() {
    Foo f = new Foo();

    io:println(f);
    io:println(f.toString());
}
```

**Affected Versions:**
0.992.0-m8

gimantha:
## Purpose
This PR adds toString() method to ballerina objects. For more details about the behavior of toString on object, refer to https://raw.githubusercontent.com/ballerina-platform/ballerina-spec/v2019R2/lang/lib/value.bal
Fixes #17003, #17001

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ErrorValue.java
```diff
@@ -67,7 +69,16 @@ public ErrorValue(BType type, String reason, Object details) {
 
     @Override
     public String stringValue() {
-        return "error " + reason + Optional.ofNullable(details).map(details -> " " + details).orElse("");
+        BType type = TypeChecker.getType(details);
+        return "error " + reason + Optional.ofNullable(details).map(details -> " " + StringUtils.getStringValue(null,
```
hasithaa: This can be replaced with return stringValue(null);

gimantha: fixed

PR #18180, opened 2019-08-26
Clone stream data before sending to each subscriber

### LINKED ISSUE #17064: Use a clone of the published data for each stream subscription
**Description:**
According to the spec, for each subscription to a stream, "Each subscriber receives a separate clone of the data."

Currently data is not cloned before sending. 

**Affected Versions:**
0.992.0-m8

gimantha:
Fixes #17064

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/CloneUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/CloneUtils.java
@@ -18,13 +18,6 @@
 
 package org.ballerinalang.jvm.values;
 
-import org.ballerinalang.jvm.TypeChecker;
-import org.ballerinalang.jvm.types.BTypes;
-import org.ballerinalang.jvm.types.TypeTags;
-import org.ballerinalang.jvm.util.exceptions.BLangExceptionHelper;
-import org.ballerinalang.jvm.util.exceptions.BallerinaErrorReasons;
-import org.ballerinalang.jvm.util.exceptions.RuntimeErrors;
-
 import java.util.HashMap;
 
 /**
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/CloneUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/CloneUtils.java
@@ -27,6 +27,11 @@ import org.ballerinalang.jvm.util.exceptions.RuntimeErrors;
 
 import java.util.HashMap;
 
+/**
+ * This class contains the functions related to cloning Ballerina values.
+ *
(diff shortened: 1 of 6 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/CloneUtils.java
```diff
@@ -0,0 +1,63 @@
...
+            return null;
+        }
+
+        if (!(value instanceof RefValue)) {
+            return value;
+        }
+
+        RefValue refValue = (RefValue) value;
+
+        if (refValue.getType().getTag() == TypeTags.ERROR_TAG ||
```
MaryamZi: Not related to this PR. But do we need this check now, since we only allow `.clone()` on `anydata` now?

Seems like an unnecessary and expensive check now. 

Even if we desugar to this method, I believe we should only do so on `anydata` values.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/stream/BStreamValueTest.java
```diff
@@ -240,21 +243,30 @@ public void testStreamPublishingAndSubscriptionForAssignableUnionTypeStream() {
         Assert.assertEquals(receivedEvents.size(), publishedEvents.size(), "Number of Events received does not "
                 + "match the number published");
         for (int i = 0; i < publishedEvents.size() - 1; i++) {
-            Assert.assertEquals(receivedEvents.getRefValue(i), publishedEvents.getRefValue(i),
-                                "Received event does not match the published event");
+            Assert.assertTrue(BVM.isEqual(receivedEvents.getRefValue(i), publishedEvents.getRefValue(i),
```
MaryamZi: Since the BVM will be deprecated, and we can't use BValues with the TypeChecker, shall we update the test to check equality on the Ballerina side and return a `boolean` instead?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/CloneUtils.java
gimantha: fixed

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/stream/BStreamValueTest.java
gimantha: fixed

PR #18204, opened 2019-08-26
Execute select on same connection as other operations in a transaction

### LINKED ISSUE #18183: Use same database connection for select statements as other statements inside a transaction
**Description:**
Need to remove isSelectQuery condition at [1], because, [2] is a behavior enforced from the driver itself.

Therefore, if the user needs to perform another operation he must fully iterate the result set, or he can set `clobberStreamingResults` to `true` in `dbOptions` in the JDBC client configuration which would cause a 'streaming' ResultSet to be automatically closed, and any outstanding data stillstreaming from the server to be discarded if another query is executed before all the data has been readfrom the server.

Also, with the change done in https://github.com/ballerina-platform/ballerina-lang/pull/16702 there is a requirement to be able to execute a select operation inside a transaction and the changes done inside the same transaction to be visible to that operation. For that also this would be required

[1] https://github.com/ballerina-platform/ballerina-lang/blob/v1.0.0-alpha3/stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/AbstractSQLStatement.java#L371

[2] https://github.com/ballerina-platform/ballerina-lang/blob/v1.0.0-alpha3/stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/AbstractSQLStatement.java#L361

**Steps to reproduce:**

**Affected Versions:**

**Related Issues/PRs (optional):**
https://github.com/ballerina-platform/ballerina-lang/pull/10888
https://github.com/ballerina-platform/ballerina-lang/issues/10611

Manuri:
Purpose
Fixes #18183

When executing operations following a select inside a transaction, on a MySQL database, set `clobberStreamingResults` property to true so that, 'streaming' ResultSet to be automatically closed, and any outstanding data still streaming from the server to be discarded if another query is executed before all the data has been read from the server.

```ballerina
jdbc:Client testDB = new({
    url: "jdbc:mysql://localhost:3306/testdb",
    username: "root",
    password: "123",
    dbOptions: { useSSL: false, clobberStreamingResults: true }
});
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TableResourceManager.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TableResourceManager.java
@@ -36,12 +36,11 @@ public class TableResourceManager {
     private Connection connection;
     private Statement statement;
     private Set<ResultSet> resultSets;
-    // Connection is closable if the table associated with the connection was retrieved through a select operation,
-    // because for select operations a separate db connection is obtained which independent of the connection used for
-    // rest of the operations in case of a transaction.
-    // If the table was retrieved through a call operation the connection will not be closable as the connection is
-    // shared with all the operations inside the transaction except for the select operations. In this case, connection
-    // closure happens through the {@code TransactionResourceManager} at the end of the transaction.
+
+    // Connection is closable only if the operation is not inside of a transaction because the connection is shared
+    // with all the operations inside the transaction.
+    // In the case where connection is not closable, connection closure happens through the
+    // {@code TransactionResourceManager} at the end of the transaction.
     private boolean connectionClosable;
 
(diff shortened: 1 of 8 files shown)


anupama-pathirage:
@Manuri : Do we have transactions tests  to test different scenarios such as

-  both select (iterate fully and not iterated) and update in transaction
-  select, and close within transaction etc

Also make sure these tests have running in mysql as this issue occurred in mysql

Manuri:
I've manually tested those cases on MySQL. 

I think we can add some more to the unit tests we have. I'll look into that.

Issue #18168, opened 2019-08-25
[Spec Deviation] wildcard-match-pattern matches errors


MaryamZi:
**Description:**
The spec says "Matching a wildcard-match-pattern against a value succeeds if the value belongs to type any, in other words if the basic type of the value is not error.", but $title.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main(string... args) {
    string|error s = error("err");

    match s {
        var x if x is string => {
            io:println("string");
        }
        _ => {
            io:println("wildcard");
        }
    }
}
```

```cmd
$ ballerina run test.bal 
wildcard
```

**Affected Versions:**
```cmd
$ ballerina -v
Ballerina 1.0.0-alpha3
Language specification 2019R2
Ballerina tool 1.0.0
```

PR #18214, opened 2019-08-27
Wildcard match pattern only match 'any' type

rdhananjaya:
## Purpose
Constraint the range of values matched by wildcard match pattern to type of `any`.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/18168

```ballerina
    any|error v = error("{TheError}Error");
    match v {
        5 => { return 0; }
        2 => { return 1; }
        3 => { return 2; }
        _ => { return -1; } // This would not match
    }
```

And same is done for 'var _' binding pattern as well.

This PR also adds a missing parser listener error check. Hence fixes https://github.com/ballerina-platform/ballerina-lang/issues/18225

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/structured_error_match_patterns.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/structured_error_match_patterns.bal
@@ -203,6 +203,7 @@ function foo5(any|error e) returns string {
             }
         }
     }
+    return "Not-matched";
 }
 
 function testBasicErrorMatch7() returns string[] {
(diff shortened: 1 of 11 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/structured_error_match_patterns.bal
```diff
@@ -203,6 +203,7 @@ function foo5(any|error e) returns string {
             }
         }
     }
+    return "Not-matched";
```
hasithaa: Is this correct? var x will match the error value as well. So every brach is returning. Hence this is unreachable code.

PR #18218, opened 2019-08-27
Limit constants from using '_' as identifier

### LINKED ISSUE #18157: _ is allowed on const
**Description:**
Following is doesn't give a compiler error. 

**Steps to reproduce:**
```
const _ = 10;

}
```

Tested on ballerina-1.0.0-Alpha3

irshadnilam:
## Purpose
$title

Fixes #18157

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -2280,6 +2280,9 @@ public class SemanticAnalyzer extends BLangNodeVisitor {
 
     @Override
     public void visit(BLangConstant constant) {
+        if (names.fromIdNode(constant.name) == Names.IGNORE) {
+            dlog.error(constant.name.pos, DiagnosticCode.UNDERSCORE_NOT_ALLOWED);
+        }
         if (constant.typeNode != null && !types.isAllowedConstantType(constant.typeNode.type)) {
             dlog.error(constant.typeNode.pos, DiagnosticCode.CANNOT_DEFINE_CONSTANT_WITH_TYPE, constant.typeNode);
         }
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
@@ -1055,6 +1055,10 @@ public class SymbolEnter extends BLangNodeVisitor {
             return;
         }
 
+        if (constant.symbol.name == Names.IGNORE) {
+            // Avoid symbol definition for constants with name '_'
+            return;
+        }
         // Add the symbol to the enclosing scope.
         env.scope.define(constantSymbol.name, constantSymbol);
     }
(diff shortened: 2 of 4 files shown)


compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
```diff
@@ -420,6 +420,7 @@
     CANNOT_UPDATE_CONSTANT_VALUE("cannot.update.constant.value"),
     CANNOT_ASSIGN_VALUE_TO_CONSTANT("cannot.assign.value.to.constant"),
     CANNOT_DEFINE_CONSTANT_WITH_TYPE("cannot.define.constant.with.type"),
+    CANNOT_DEFINE_CONSTANT_WITH_IDENTIFIER("cannot.define.constant.with.identifier"),
```
MaryamZi: We already have a `UNDERSCORE_NOT_ALLOWED` error, how about re-using that instead? The error would be "underscore is not allowed here". @hasithaa 

If we need a separate error shall we change this to
```suggestion
    CANNOT_DEFINE_CONSTANT_WITH_UNDERSCORE("cannot.define.constant.with.underscore"),
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2280,6 +2280,11 @@ public void visit(BLangForever foreverStatement) {
 
     @Override
     public void visit(BLangConstant constant) {
+        Name constantName = names.fromIdNode(constant.name);
```
MaryamZi: We use this variable only in L2284 right? 

How about we define a variable for `constant.name` instead, since that seems to be used several times.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2280,6 +2280,11 @@ public void visit(BLangForever foreverStatement) {
...
+        if (constantName == Names.IGNORE) {
+            dlog.error(constant.name.pos, DiagnosticCode.CANNOT_DEFINE_CONSTANT_WITH_IDENTIFIER, constant.name);
+            return;
```
MaryamZi: Do we need the return here? We can allow the next check also right?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1138,6 +1138,9 @@ error.cannot.assign.value.to.constant=\
 error.cannot.define.constant.with.type=\
   constant cannot be defined with type ''{0}'', expected a simple basic types or a map of a simple basic type
 
+error.cannot.define.constant.with.identifier=\
+  constant cannot be defined with identifier ''{0}''
```
MaryamZi: Again, if we are going with a new error I think we can change this to something like
```suggestion
  constant cannot be defined with ''_''
```

since ''{0}'' will always be ''_''

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
hasithaa: +1, better to use "underscore is not allowed here"

irshadnilam: Fixed. Used "Underscore not allowed" message

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
irshadnilam: Fixed.

PR #18231, opened 2019-08-27
Fix permission issue and improve cli messages

keizer619:


cli/ballerina-tool/src/main/java/org/ballerinalang/tool/util/ToolUtil.java
```diff
@@ -118,7 +123,7 @@ private static void setCurrentBallerinaVersion(String version) throws IOExceptio
      */
     private static String getCurrentToolsVersion() {
         //TODO: Need to read folder
-        return "1.0.0-beta";
+        return "1.0.0-beta-SNAPSHOT";
```
hemikak: Can we get this from a system property or env ?

keizer619: We have to do something like that once we separate tool jar from ballerina lang, until that we might need at least read a file and get the tool version.

PR #18277, opened 2019-08-28
Remove the logic of setting default truststore

Bhashinee:
## Purpose
$subject.


stdlib/http/src/main/java/org/ballerinalang/net/http/HttpUtil.java
```diff
@@ -1162,15 +1162,15 @@ public static ObjectValue createResponseStruct(HttpCarbonMessage httpCarbonMessa
...
+            MapValue<String, Object> clientEndpointConfig, String scheme) {
         ProxyServerConfiguration proxyServerConfiguration;
         MapValue secureSocket = clientEndpointConfig.getMapValue(HttpConstants.ENDPOINT_CONFIG_SECURE_SOCKET);
 
         if (secureSocket != null) {
             HttpUtil.populateSSLConfiguration(senderConfiguration, secureSocket);
-        } else {
-            HttpUtil.setDefaultTrustStore(senderConfiguration);
+        } else if (scheme.equals(PROTOCOL_HTTPS)) {
+            throw new ErrorValue("To enable https you need to configure secureSocket record", null);
```
ThisaruGuruge: Instead of creating a plain `ErrorValue`, shall we create an _http-module-specific_ `ErrorValue` here ?
We can use [`createHttpError()`](https://github.com/ballerina-platform/ballerina-lang/blob/73babe4728890d9fbc85b04c7b234c03853b40d8/stdlib/http/src/main/java/org/ballerinalang/net/http/HttpUtil.java#L487) method to create this.

PR #18317, opened 2019-08-29
Limit object __init function from invoking outside of self

### LINKED ISSUE #17918: [Spec Deviation] Object __init call should only be allowed on self
**Description:**
The 2019r3 spec says
```
 The __init method can be called in a method-call-expr only when the expression 
preceding the . is self.
```

But the implementation currently supports calling `__init()` on an object value.

**Steps to reproduce:**
```ballerina
import ballerina/io;

type Foo object {

    public function __init() {
        io:println("__init");    
    }

    function name() {
        io:println("name");    
        self.__init(); // valid
    }
};

public function main() {
    Foo f = new;
    f.__init(); // should fail at compile time
}
```

```cmd
$ ballerina run test.bal 
__init
__init
```

**Affected Versions:**
1.0.0-beta-SNAPSHOT

irshadnilam:
## Purpose
$title

Fixes #17918 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/workers/worker-cancelled.bal
```diff
@@ -1,10 +1,13 @@
+import ballerina/runtime;
+
 // Test if worker actions are panicked if the worker is cancelled before sending
 function workerCancelledBeforeSend() {
         worker wy {
             string aa = <- default;
             "message" -> default;
         }
         wy.cancel();
+        runtime:sleep(5);
```
irshadnilam: This test fails intermittently. 
I added this sleep following @vinok88 pointers.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -3077,6 +3077,15 @@ private void checkObjectFunctionInvocationExpr(BLangInvocation iExpr, BObjectTyp
                 return;
             }
         }
+
+        // __init method can be called in a method-call-expr only when the expression
+        // preceding the . is self.
```
irshadnilam: Is this comment okay?
I will remove the ending **.** with other suggestions.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -572,6 +572,9 @@ error.invalid.listener.attachment=\
 error.invalid.action.invocation.syntax=\
   invalid remote function invocation syntax, use ''->'' operator
 
+error.invalid.init.invocation=\
+  call to function ''__init'' is not allowed here
```
irshadnilam: Should we add a more descriptive error message?

MaryamZi:
```suggestion
  object ''__init()'' cannot be called outside the object
```

tests/jballerina-unit-test/src/test/resources/test-src/object/object_initializer_negative.bal
```diff
@@ -92,3 +94,17 @@ type Person4 object {
         return e;
     }
 };
+
+type Too object {
+    public function __init() {
+        io:println("__init");
```
MaryamZi: We don't need this for the test right? We usually try to minimize imports in language unit tests. IMO, we can have an empty method here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -3077,6 +3077,15 @@ private void checkObjectFunctionInvocationExpr(BLangInvocation iExpr, BObjectTyp
...
+        if (iExpr.name.value.equals(Names.USER_DEFINED_INIT_SUFFIX.value) &&
+                !(iExpr.expr.getKind() == NodeKind.SIMPLE_VARIABLE_REF &&
+                ((BLangSimpleVarRef) iExpr.expr).getVariableName().value.equals(Names.SELF.value))) {
```
pubudu91: We usually use the symbol for these checks. e.g.,
```java
Names.SELF.equals(((BLangSimpleVarRef) iExpr.expr).symbol.name)
```

tests/jballerina-unit-test/src/test/resources/test-src/object/object_initializer_negative.bal
```diff
@@ -92,3 +94,17 @@ type Person4 object {
...
+    }
+    function name() {
+        self.__init(); // valid
```
MaryamZi: Shall we add a positive test for this? To check if the `__init()` is actually invoked.

compiler/ballerina-lang/src/main/resources/compiler.properties
MaryamZi: @pubudu91 thoughts?

pubudu91: +1 for a more descriptive error message. Maybe something like below:
```
call to function ''__init'' is only allowed within its object type descriptor
```

tests/jballerina-unit-test/src/test/resources/test-src/workers/worker-cancelled.bal
pubudu91: 5 milliseconds is enough?

irshadnilam: Yes, from what @vinok88 mentioned, when there is **wait** like event, the worker gets the handle to cancel the task. 

@vinok88 please correct if i am wrong

compiler/ballerina-lang/src/main/resources/compiler.properties
irshadnilam: Changed the message to `
''__init'' function invocation is only allowed within its object type descriptor`

tests/jballerina-unit-test/src/test/resources/test-src/object/object_initializer_negative.bal
irshadnilam: Removed the print statement and import.

tests/jballerina-unit-test/src/test/resources/test-src/object/object_initializer_negative.bal
irshadnilam: Positive test added.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
irshadnilam: Changed.

compiler/ballerina-lang/src/main/resources/compiler.properties
MaryamZi: With other error messages related to init we seem to have used ''__init()''. Shall we keep this uniform?

Also saying "its" doesn't feel right here, unless we start with "object ''__init()''..." This should also be "method" and not "function".

How about
```suggestion
  object ''__init()'' method invocation is only allowed within the type descriptor
```

@irshadnilam, let's finalize the error message before doing any more changes :)

MaryamZi: As discussed offline let's change this to 
```suggestion
  object ''__init'' method invocation is only allowed within the type descriptor
```

Shall we also please change
```
error.object.init.function.cannot.be.extern=\
  object ''__init()'' function cannot be an ''external'' function
```

to 

```
error.object.init.function.cannot.be.extern=\
  object ''__init'' method cannot be an ''external'' function
```

irshadnilam: Changed.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -704,7 +707,7 @@ error.resource.function.cannot.be.extern=\
   external resource functions are not supported by the implementation
 
 error.object.init.function.cannot.be.extern=\
-  object ''__init()'' function cannot be an ''external'' function
+  object ''__init'' function cannot be an ''external'' function
```
```suggestion
  object ''__init'' method cannot have an ''external'' implementation
```

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -566,6 +566,9 @@ error.invalid.listener.attachment=\
 error.invalid.action.invocation.syntax=\
   invalid remote function invocation syntax, use ''->'' operator
 
+error.invalid.init.invocation=\
+  object ''__init'' method invocation is only allowed within the type descriptor
```
```suggestion
  object ''__init'' method call is only allowed within the type descriptor
```

MaryamZi:
I guess this should be specifically restricted in tooling now. Can you create a tooling issue specifying the valid and invalid calls please?

irshadnilam:
> I guess this should be specifically restricted in tooling now. Can you create a tooling issue specifying the valid and invalid calls please?

Created
https://github.com/ballerina-platform/ballerina-lang/issues/18369

PR #18333, opened 2019-08-29
Improve tuple & array inter-assignability

### LINKED ISSUE #17891: Incompatible open array to tuple (without rest field) assignment should be disallowed
**Description:**
$title. Currently allowed as a side-effect of allowing langlib invocations on lists.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    int[] i = [1];
    [int, int] j = i;
    io:println(j[0]);
    io:println(j[1]); // panics
}
```

**Affected Versions:**
Ballerina 1.0.0-beta-SNAPSHOT - 15th August build

grainier:
## Purpose
$subject

Fixes #17891

## Remarks
#17891


tests/jballerina-unit-test/src/test/resources/test-src/types/tuples/tuple_negative_test.bal
```diff
@@ -50,12 +50,6 @@ function testTupleToArrayAssignmentNegative1() returns string[] {
     return y;
 }
 
-function testTupleToArrayAssignmentNegative2() returns string[] {
-    [string, string] x = ["a", "b"];
-    string[3] y = x;
```
hasithaa: This is a valid negative case, why it is removed?

grainier: I was under the impression that since `string[] y = x;`, this should also be allowed (with fillers). fixed in cacae9a

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -753,8 +753,11 @@ private boolean isTupleTypeAssignableToArrayType(BTupleType source, BArrayType t
 
     private boolean isArrayTypeAssignableToTupleType(BArrayType source, BTupleType target,
                                                      List<TypePair> unresolvedTypes) {
-        if (source.state != BArrayState.UNSEALED
-                && (target.restType != null || target.tupleTypes.size() != source.size)) {
+        if (source.state == BArrayState.UNSEALED) {
+            if (target.restType == null) {
+                return false;
+            }
+        } else if (target.restType != null || target.tupleTypes.size() != source.size) {
```
MaryamZi: Closed array to a tuple with a rest type can pass in a scenario like the following right?

```ballerina
int[4] i = [1, 2, 3, 4];
[int, int, int...] j = i; // this is valid
```

I don't think we can simply return `false` if `target.restType != null`.

grainier: Yes m'am. 🙏

grainier: Fixed in cf480c6

tests/jballerina-unit-test/src/test/resources/test-src/types/tuples/tuple_negative_test.bal
```diff
@@ -86,6 +86,23 @@ function testArrayToTupleAssignmentNegative5() returns [int...] {
...
+
+    int[1] d = [1];
+    [int, int, int...] e = d;
+
+    [int, int] f = d;
+
+    int[3] g = [1, 2, 3];
+    [int, int] h = g;
+
+    [int...] i = d;
```
MaryamZi: Shall we add a positive test for this?

PR #18529, opened 2019-09-04
Fix client endpoint BBE

Bhashinee:
## Purpose
$subject.


examples/http-client-endpoint/http_client_endpoint.bal
```diff
@@ -1,8 +1,19 @@
...
+        trustStore:{
+            path: "${ballerina.home}/bre/security/ballerinaTruststore.p12",
+            password: "ballerina"
+        }
+    }
+};
+
 // Creates a new client with the backend URL.
-http:Client clientEndpoint = new("https://postman-echo.com");
+http:Client clientEndpoint = new("https://postman-echo.com", clientConf);
```
shafreenAnfar: Guys, can we use some http endpoint. Like below,

http://api.sunrise-sunset.org

PR #18797, opened 2019-09-07
Fix observability enable flag

### LINKED ISSUE #18798: Observability enable flag not working
**Description:**
Observability enable flag `--b7a.observability.enabled=true` not working in RC1.

**Steps to reproduce:**
`ballerina run myService --b7a.observability.enabled=true` would not enable observability

grainier:
## Purpose
Change of observability flag from `--observe` to `--b7a.observability.enabled=true`, was broken with #18633. This will fix it.

Fix #18798


warunalakshitha:
Shall we add a test for this. I thought we already have tests. :)

PR #18820, opened 2019-09-07
Fix add command help text

anupama-pathirage:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
+++ b/cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
@@ -11,7 +11,7 @@ DESCRIPTION
        Add a new Ballerina module to the project. The command should
        be run inside a Ballerina project.
 
-       Also different templates can be selected for the new module.
+       Also different templates can be selected to create the new module.
 
 
 OPTIONS
@@ -29,4 +29,3 @@ EXAMPLES
 
        Add a module with a service
        $ ballerina add myservice --template service
-       
--- a/cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
+++ b/cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
@@ -3,7 +3,7 @@ NAME
 
 
 SYNOPSIS
-       ballerina add <module-name> [-t | --template <template-name>]
+       ballerina add <module-name> [-t <template-name> | --template <template-name>]
                      [-l | --list]
 
 
--- a/cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
+++ b/cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
@@ -29,3 +29,4 @@ EXAMPLES
 
        Add a module with a service
        $ ballerina add myservice --template service
+       
--- a/cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
+++ b/cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
@@ -4,7 +4,7 @@ NAME
 
 SYNOPSIS
        ballerina add <module-name> [-t | --template <template-name>]
-                     [-l | --list ]
+                     [-l | --list]
 
 
 DESCRIPTION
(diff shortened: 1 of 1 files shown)


cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
```diff
@@ -1,16 +1,17 @@
 NAME
-       Ballerina add - Add a Ballerina module in a project
+       ballerina add - Add a new Ballerina module to a project
```
MaryamZi:
```suggestion
       ballerina add - add a new module to a Ballerina project
```

cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
```diff
@@ -1,16 +1,17 @@
...
+
 
 SYNOPSIS
-       ballerina add <module-name> [-t|--template <template-name>]
+       ballerina add <module-name> [-t | --template <template-name>]
+                     [-l | --list ]
```
MaryamZi:
```suggestion
                     [-l | --list]
```

cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
```diff
@@ -1,16 +1,17 @@
 NAME
-       Ballerina add - Add a Ballerina module in a project
+       ballerina add - Add a new Ballerina module to a project
+
 
 SYNOPSIS
-       ballerina add <module-name> [-t|--template <template-name>]
+       ballerina add <module-name> [-t | --template <template-name>]
```
MaryamZi: Is it just me or does this seem like

`-t` 
OR
`--template <template-name>`

Should this be 
`[-t <template-name> | --template <template-name>]` instead?

cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
```diff
@@ -1,16 +1,17 @@
...
+                     [-l | --list ]
 
 
 DESCRIPTION
-       Ballerina add command will initialize a new Ballerina module
-       inside the Project. The command should be run inside a Ballerina
-       project.
+       Add a new Ballerina module to the project. The command should
```
MaryamZi:
```suggestion
       Add a new module to a Ballerina project. The command should
```

cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
```diff
@@ -1,16 +1,17 @@
...
+       be run inside a Ballerina project.
 
-       You can also select different templates for the module.
+       Also different templates can be selected to create the new module.
```
MaryamZi:
```suggestion
       A template can also be specified when creating the new module.
```

cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
anupama-pathirage: Shouldn't it be a valid complete sentence?

cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
pubudu91: It seems like we are not following the convention used for these options. Have a look at https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

If an option starts with `--`, it's a long option. And for long options, you give values as `--opt=<value>`. Not `--opt <value>`. Checked in man pages of git, grep etc. and they all follow the same convention

pubudu91: Ah nvm. The key=value convention seem to be used when the value is optional.

cli/ballerina-tool/src/main/resources/cli-help/ballerina-add.help
MaryamZi: `man cat` produces something like

```
$ man cat
NAME
       cat - concatenate files and print on the standard output
```

If it starts with an upper case letter we have to finish with a full-stop too?

Issue #18989, opened 2019-09-10
Error match works without var


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    string|error se = error("test reason");

    match se {
        error(reason) => {
            io:println("this works without var!");
        }
        var value => {
            io:println("value");
        }
    }
}
```

**Affected Versions:**
```cmd
$ ballerina -v
Ballerina 1.0.0
Language specification 2019R3
```

**Related Issue(s):**
https://github.com/ballerina-platform/ballerina-lang/issues/15962

MaryamZi:
Fixed with #18996.

PR #18996, opened 2019-09-11
Add error match pattern 'var reason' validation

rdhananjaya:
## Purpose
$subject.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/18989 and https://github.com/ballerina-platform/ballerina-lang/issues/18997

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -581,7 +581,7 @@ public class SemanticAnalyzer extends BLangNodeVisitor {
 
         validateWorkerAnnAttachments(varNode.expr);
 
-        if (varNode.name.value.equals(Names.IGNORE.value) || varNode.name.value.equals("")) {
+        if (isIgnoredOrEmpty(varNode)) {
             // Fake symbol to prevent runtime failures down the line.
             varNode.symbol = new BVarSymbol(0, Names.IGNORE, env.enclPkg.packageID, symTable.anyType, env.scope.owner);
         }
@@ -1354,7 +1354,7 @@ public class SemanticAnalyzer extends BLangNodeVisitor {
                 errorVariable.restDetail.accept(this);
             }
 
-            if (isNotIgnoredOrNotSpecified(errorVariable)
+            if (isReasonSpecified(errorVariable)
                     && !errorVariable.reasonVarPrefixAvailable
                     && errorVariable.reasonMatchConst == null
                     && errorVariable.isInMatchStmt) {
@@ -1387,9 +1387,12 @@ public class SemanticAnalyzer extends BLangNodeVisitor {
         return true;
     }
 
-    private boolean isNotIgnoredOrNotSpecified(BLangErrorVariable errorVariable) {
-        String value = errorVariable.reason.name.value;
-        return !(value.equals(Names.IGNORE.value) || value.equals(""));
+    private boolean isReasonSpecified(BLangErrorVariable errorVariable) {
+        return !isIgnoredOrEmpty(errorVariable.reason);
+    }
+
+    private boolean isIgnoredOrEmpty(BLangSimpleVariable varNode) {
+        return varNode.name.value.equals(Names.IGNORE.value) || varNode.name.value.equals("");
     }
 
     private boolean isRestDetailBindingAvailable(BLangErrorVariable errorVariable) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -1353,6 +1353,22 @@ public class SemanticAnalyzer extends BLangNodeVisitor {
                 errorVariable.restDetail.type = restType;
                 errorVariable.restDetail.accept(this);
             }
+
+            if (!errorVariable.reason.name.value.equals(Names.IGNORE.value)
+                    && !errorVariable.reasonVarPrefixAvailable
+                    && errorVariable.reasonMatchConst == null
+                    && errorVariable.isInMatchStmt) {
+
+                BSymbol reasonConst = symResolver.lookupSymbol(
+                        this.env.enclEnv, names.fromString(errorVariable.reason.name.value), SymTag.CONSTANT);
+                if (reasonConst == symTable.notFoundSymbol) {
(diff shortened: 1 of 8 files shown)


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -634,6 +634,9 @@ error.undefined.parameter=\
 error.invalid.error.reason.type=\
   invalid error reason type ''{0}'', expected a subtype of ''string''
 
+error.error.match.over.const.reason.ref.not.supported=\
+  error match pattern with reason as a ''constant'' reference is currently not supported
```
```suggestion
  error match pattern with a constant reference as the reason is not yet supported
```

pubudu91: How about something like 
```
matching error reason to a ''const'' pattern is currently not supported.
```
?

tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/structured_error_match_patterns.bal
```diff
@@ -329,16 +329,28 @@ function testErrorConstReasonMatchPattern() returns string {
...
             return reason;
         }
     }
     return "Default";
 }
+
+public function testErrorMatchWihtoutReason() returns string {
+    string|error se = error("test reason", message="error detail message");
+
+    match se {
```
pubudu91: Does the wild card match pattern work for the error reason? If so, shall we add a test case for that as well?

compiler/ballerina-lang/src/main/resources/compiler.properties
rdhananjaya: Matching to a const literal is supported

tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/structured_error_match_patterns.bal
rdhananjaya: I think it's already there https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/structured_error_match_patterns.bal#L151

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/matchstmt/MatchStructuredErrorPatternsTest.java
```diff
@@ -199,4 +199,22 @@ public void testUnreachablePatterns() {
...
         BAssertUtil.validateError(resultNegative, ++i, unreachablePattern, 85, 13);
     }
+
+    @Test(description = "Test unsupported error match pattern")
+    public void testErrorMatchPatternNotSupportedErrors() {
+        CompileResult result = BCompileUtil.compile(
+                "test-src/statements/matchstmt/structured_error_match_patterns_negative2.bal");
+        int i = 0;
+        BAssertUtil.validateError(result, i++,
+                "error match pattern with reason as a 'constant' reference is currently not supported", 33, 15);
```
MaryamZi: tests needs to be updated for the error message change?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1353,6 +1353,22 @@ private boolean validateErrorVariable(BLangErrorVariable errorVariable) {
                 errorVariable.restDetail.type = restType;
                 errorVariable.restDetail.accept(this);
             }
+
+            if (isReasonSpecified(errorVariable)
+                    && !errorVariable.reasonVarPrefixAvailable
+                    && errorVariable.reasonMatchConst == null
+                    && errorVariable.isInMatchStmt) {
```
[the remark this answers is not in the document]
MaryamZi: Do we need a new field btw? Haven't we come across the requirement to identify if we are in a match statement before?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/matchstmt/MatchStructuredErrorPatternsTest.java
rdhananjaya: Yes, I forgot that

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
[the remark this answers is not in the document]
rdhananjaya: > Do we need a new field btw? Haven't we come across the requirement to identify if we are in a match statement before?

AFAIK we don't support const matching in other match patterns and treat all names as a variable to be bound to members of matched value. I think that's why didn't wanted something like this before.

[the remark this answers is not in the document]
rdhananjaya: > Shouldn't this be the first check?

Will move to top as it makes more sense to check reason, detail, rest in that order.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1318,6 +1318,22 @@ private boolean validateErrorVariable(BLangErrorVariable errorVariable) {
             Map<String, BField> fieldMap = recordType.fields.stream()
                     .collect(Collectors.toMap(f -> f.name.value, f -> f));
 
+            if (isReasonSpecified(errorVariable)
```
MaryamZi: We can move this to L1317 right? The variables defined are not used in this block?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1318,6 +1318,22 @@ private boolean validateErrorVariable(BLangErrorVariable errorVariable) {
...
+                    && !errorVariable.reasonVarPrefixAvailable
+                    && errorVariable.reasonMatchConst == null
+                    && errorVariable.isInMatchStmt) {
```
MaryamZi: What I originally meant is whether `errorVariable.isInMatchStmt` should be the first check in the if block?

rdhananjaya: No preferences from me, will move up

PR #19085, opened 2019-09-17
Migrate RabbitMQ module to use java interoperability

aashikam:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/messaging/rabbitmq/src/main/ballerina/src/rabbitmq/channel.bal
```diff
@@ -37,24 +39,31 @@ public type Channel client object {
     # + queueConfig - Holds the parameters required to declare a queue.
     # + return - The name of the queue if autogenerated or nil if the queue was successfully
     #               generated with the given parameters. An error is returned if an I/O error is encountered.
-    public remote function queueDeclare(QueueConfiguration? queueConfig = ())
-                                        returns string | Error? = external;
+    public remote function queueDeclare(QueueConfiguration? queueConfig = ()) returns string | Error? =
```
SupunS: return type should be `handle | Error?`.
java type `String` maps to ballerina type `handle`. 

This would work for now due to a bug in the compiler, which is fixed in the 1.0.1 patch version. So this will fail when we merge the two branches. So better to fix it properly here.

Applicable to all places...

PR #19098, opened 2019-09-18
Fix a compound assignment issue in function calling

### LINKED ISSUE #15459: A bug in compound operator implementation

```ballerina
a[f(i)] += y;
```
Here `f(i)` is not a side-effect free function. When I run this code, `f(i)` is executed twice. That is a bug.

KavinduZoysa:
## Purpose
Fixes #15459

## Approach
If my compound assignment is ```a[f(1, foo)][3][2] += y```, index expressions should be extracted and evaluated them separately. Finally, create a block statement as follows.
```
var $temp3$ = a[f(1, foo)];
var $temp2$ = 3;
var $temp1$ = 2;
a[$temp3$][$temp2$][$temp1$] = a[$temp3$][$temp2$][$temp1$] + y;
```
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
@@ -286,6 +286,7 @@ public class Desugar extends BLangNodeVisitor {
     private int errorCount = 0;
     private int annonVarCount = 0;
     private int initFuncIndex = 0;
+    private int indexExprNumber = 0;
 
     // Safe navigation related variables
     private Stack<BLangMatch> matchStmtStack = new Stack<>();
@@ -2349,7 +2350,6 @@ public class Desugar extends BLangNodeVisitor {
         // var $temp2$ = 3;
         // var $temp1$ = 2;
         // a[$temp3$][$temp2$][$temp1$] = a[$temp3$][$temp2$][$temp1$] + y;
-        int indexExprNumber = 0;
         List<BLangStatement> statements = new ArrayList<>();
         List<BLangSimpleVarRef> varRefs = new ArrayList<>();
         List<BType> types = new ArrayList<>();
@@ -2369,15 +2369,11 @@ public class Desugar extends BLangNodeVisitor {
             varRefs.add(0, tempVarRef);
             types.add(0, varRef.type);
 
-            if (((BLangIndexBasedAccess) varRef).expr.getKind() == NodeKind.INDEX_BASED_ACCESS_EXPR) {
-                varRef = (BLangVariableReference) ((BLangIndexBasedAccess) varRef).expr;
-                continue;
-            }
-            break;
-        } while (true);
+            varRef = (BLangVariableReference) ((BLangIndexBasedAccess) varRef).expr;
+        } while (varRef.getKind() == NodeKind.INDEX_BASED_ACCESS_EXPR);
 
         // Create the index access expression. ex: c[$temp3$][$temp2$][$temp1$]
-        BLangVariableReference var = (BLangVariableReference) ((BLangIndexBasedAccess) varRef).expr;
+        BLangVariableReference var = varRef;
         for (int ref = 0; ref < varRefs.size(); ref++) {
             var = ASTBuilderUtil.createIndexAccessExpr(var, varRefs.get(ref));
             var.type = types.get(ref);
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
@@ -2351,9 +2351,9 @@ public class Desugar extends BLangNodeVisitor {
                         indexBasedAccess.indexExpr.type, indexBasedAccess.indexExpr, compoundAssignment.pos);
                 BLangSimpleVarRef tempVarRef = ASTBuilderUtil.createVariableRef(tempIndexVarDef.pos,
                         tempIndexVarDef.var.symbol);
-                statements.add(tempIndexVarDef);
-                varRef.add(tempVarRef);
-                type.add(indexBasedAccess.type);
+                statements.add(0, tempIndexVarDef);
+                varRef.add(0, tempVarRef);
+                type.add(0, indexBasedAccess.type);
 
                 if (indexBasedAccess.expr.getKind() == NodeKind.INDEX_BASED_ACCESS_EXPR) {
                     indexBasedAccess = (BLangIndexBasedAccess) indexBasedAccess.expr;
@@ -2364,7 +2364,7 @@ public class Desugar extends BLangNodeVisitor {
 
             // Create the index access expression. ex: c[$temp3$][$temp2$][$temp1$]
             BLangVariableReference var = (BLangVariableReference) indexBasedAccess.expr;
-            for (int ref = varRef.size() - 1; ref >= 0; ref--) {
+            for (int ref = 0; ref < varRef.size(); ref++) {
                 var = ASTBuilderUtil.createIndexAccessExpr(var, varRef.get(ref));
                 var.type = type.get(ref);
             }
@@ -2374,8 +2374,6 @@ public class Desugar extends BLangNodeVisitor {
             BLangExpression rhsExpression = ASTBuilderUtil.createBinaryExpr(compoundAssignment.pos, var,
                     compoundAssignment.expr, compoundAssignment.type, compoundAssignment.opKind, null);
             rhsExpression.type = compoundAssignment.modifiedExpr.type;
-
-            Collections.reverse(statements);
 
             // Create assignment statement. ex: a[$temp3$][$temp2$][$temp1$] = a[$temp3$][$temp2$][$temp1$] + y;
             BLangAssignment assignStmt = ASTBuilderUtil.createAssignmentStmt(compoundAssignment.pos, var,
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
@@ -2330,9 +2330,9 @@ public class Desugar extends BLangNodeVisitor {
     public void visit(BLangCompoundAssignment compoundAssignment) {
         // If compound Assignment is an index based expression such as a[f(1, foo)][3][2] += y,
         // should return a block statement which is equivalent to
-        // var $temp1$ = 2;
-        // var $temp2$ = 3;
         // var $temp3$ = a[f(1, foo)];
+        // var $temp2$ = 3;
+        // var $temp1$ = 2;
         // a[$temp3$][$temp2$][$temp1$] = a[$temp3$][$temp2$][$temp1$] + y;
         if (compoundAssignment.varRef.getKind() == NodeKind.INDEX_BASED_ACCESS_EXPR) {
             int indexExprNumber = 0;
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
@@ -286,7 +286,7 @@ public class Desugar extends BLangNodeVisitor {
     private int errorCount = 0;
     private int annonVarCount = 0;
     private int initFuncIndex = 0;
-    private int indexExprNumber = 0;
+    private int indexExprCount = 0;
 
     // Safe navigation related variables
     private Stack<BLangMatch> matchStmtStack = new Stack<>();
@@ -2355,12 +2355,11 @@ public class Desugar extends BLangNodeVisitor {
         List<BType> types = new ArrayList<>();
 
         // Extract the index Expressions from compound assignment and create variable definitions. ex:
-        // var $temp1$ = 2;
-        // var $temp2$ = 3;
         // var $temp3$ = a[f(1, foo)];
+        // var $temp2$ = 3;
+        // var $temp1$ = 2;
         do {
-            indexExprNumber += 1;
-            BLangSimpleVariableDef tempIndexVarDef = createVarDef("$temp" + indexExprNumber + "$",
+            BLangSimpleVariableDef tempIndexVarDef = createVarDef("$temp" + ++indexExprCount + "$",
                     ((BLangIndexBasedAccess) varRef).indexExpr.type, ((BLangIndexBasedAccess) varRef).indexExpr,
                     compoundAssignment.pos);
             BLangSimpleVarRef tempVarRef = ASTBuilderUtil.createVariableRef(tempIndexVarDef.pos,
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
@@ -2347,9 +2347,10 @@ public class Desugar extends BLangNodeVisitor {
             // var $temp3$ = a[f(1, foo)];
             do {
                 indexExprNumber += 1;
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2328,21 +2328,77 @@ public void visit(BLangXMLNS xmlnsNode) {
...
+                statements.add(tempIndexVarDef);
+                varRef.add(tempVarRef);
+                type.add(indexBasedAccess.type);
+
+                if (indexBasedAccess.expr.getKind() == NodeKind.INDEX_BASED_ACCESS_EXPR) {
+                    indexBasedAccess = (BLangIndexBasedAccess) indexBasedAccess.expr;
+                    continue;
+                }
+                break;
+            } while (true);
```
irshadnilam: Is there a workaround for us to not use `while(true)`?
How about something as follows?
```
do {
// do the casting somewhere here

} while(indexBasedAccess.expr.getKind() == NodeKind.INDEX_BASED_ACCESS_EXPR);
```

@MaryamZi  thoughts? is it okay using `while(true)` here?

KavinduZoysa:
> Shall we add a test case (jballerina-unit-tests) for this scenario?

Yes, unit tests are added.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
pubudu91: Yes, the condition to check is whether the current var ref is a index based access expr. No need for a do-while.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2328,21 +2328,78 @@ public void visit(BLangXMLNS xmlnsNode) {
...
-            varRef = ASTBuilderUtil.createVariableRef(compoundAssignment.varRef.pos, varRef.symbol);
-            varRef.lhsVar = true;
-        }
+            // Create the index access expression. ex: c[$temp3$][$temp2$][$temp1$]
+            BLangVariableReference var = (BLangVariableReference) indexBasedAccess.expr;
+            for (int ref = varRef.size() - 1; ref >= 0; ref--) {
+                var = ASTBuilderUtil.createIndexAccessExpr(var, varRef.get(ref));
+                var.type = type.get(ref);
+            }
+            var.type = compoundAssignment.varRef.type;
```
pubudu91: What's the purpose of the above code block? The variable `var` seems to be used as a temporary placeholder. Only the final value assigned to `var` seems to be significant?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2328,21 +2328,78 @@ public void visit(BLangXMLNS xmlnsNode) {
...
+            // var $temp1$ = 2;
+            // var $temp2$ = 3;
+            // var $temp3$ = a[f(1, foo)];
+            do {
+                indexExprNumber += 1;
+                BLangSimpleVariableDef tempIndexVarDef = createVarDef("$temp" + indexExprNumber + "$",
+                        indexBasedAccess.indexExpr.type, indexBasedAccess.indexExpr, compoundAssignment.pos);
+                BLangSimpleVarRef tempVarRef = ASTBuilderUtil.createVariableRef(tempIndexVarDef.pos,
+                        tempIndexVarDef.var.symbol);
+                statements.add(tempIndexVarDef);
```
pubudu91: If we keep adding to the 0-th index, we won't have to reverse the list at the end.

tests/jballerina-unit-test/src/test/resources/test-src/statements/compoundassignment/compound_assignment.bal
```diff
@@ -290,3 +290,65 @@ function testCompoundAssignmentAdditionWithRecordAccess() returns int {
     x += (ibm["count"] + arr[0]);
     return x;
 }
+
+function testCompoundAssignmentAdditionArrayElementFunctionInvocation() returns [int, int, int] {
```
MaryamZi: Can we also add tests where the execution order of the nested indices matter?

For example, for something like
```ballerina
i[foo()][bar()] += 3
```

a test where we validate that foo() is executed first and then bar() ?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2328,21 +2328,78 @@ public void visit(BLangXMLNS xmlnsNode) {
...
+            List<BLangSimpleVarRef> varRef = new ArrayList<>();
+            List<BType> type = new ArrayList<>();
+
+            BLangIndexBasedAccess indexBasedAccess = (BLangIndexBasedAccess) compoundAssignment.varRef;
+            // Extract the index Expressions from compound assignment and create variable definitions. ex:
+            // var $temp1$ = 2;
+            // var $temp2$ = 3;
+            // var $temp3$ = a[f(1, foo)];
+            do {
+                indexExprNumber += 1;
```
MaryamZi: Even though this works atm, shouldn't this be a global variable instead? We might be reusing the same variable names with compound assignments if there are multiple being rewritten within the same block.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: Instead of L2358-L2362 can we do something like
- `BLangExpression expr = compoundAssignment.varRef;` at the very beginning 
- `expr = compoundAssignment.varRef;` instead of L2358-L2362
- update the check in while to `while (expr.getKind() == NodeKind.INDEX_BASED_ACCESS_EXPR);`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
pubudu91: Sorry, ignore the above comment. Misread the code block here :D

tests/jballerina-unit-test/src/test/resources/test-src/statements/compoundassignment/compound_assignment.bal
KavinduZoysa: Fixed

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/compoundassignment/CompoundAssignmentTest.java
```diff
@@ -154,6 +154,59 @@ public void testCompoundAssignmentDivisionArrayElement() {
         Assert.assertEquals(((BInteger) returns[0]).intValue(), 10);
     }
 
+    @Test(description = "Test compound assignment with addition on array element where the index is an invocation.")
+    public void testCompoundAssignmentAdditionArrayElementFunctionInvocation() {
+        BValue[] returns = BRunUtil.invoke(result,
```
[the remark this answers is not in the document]
KavinduZoysa: Fixed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2328,21 +2328,78 @@ public void visit(BLangXMLNS xmlnsNode) {
...
-        BLangAssignment assignStmt = (BLangAssignment) TreeBuilder.createAssignmentNode();
-        assignStmt.pos = compoundAssignment.pos;
-        BLangVariableReference varRef = compoundAssignment.varRef;
+        // If compound Assignment is an index based expression such as a[f(1, foo)][3][2] += y,
+        // should return a block statement which is equivalent to
+        // var $temp1$ = 2;
+        // var $temp2$ = 3;
+        // var $temp3$ = a[f(1, foo)];
+        // a[$temp3$][$temp2$][$temp1$] = a[$temp3$][$temp2$][$temp1$] + y;
+        if (compoundAssignment.varRef.getKind() == NodeKind.INDEX_BASED_ACCESS_EXPR) {
```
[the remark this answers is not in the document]
KavinduZoysa: It is changed according to the suggestion.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2328,21 +2328,78 @@ public void visit(BLangXMLNS xmlnsNode) {
     }
 
     public void visit(BLangCompoundAssignment compoundAssignment) {
-        BLangAssignment assignStmt = (BLangAssignment) TreeBuilder.createAssignmentNode();
-        assignStmt.pos = compoundAssignment.pos;
-        BLangVariableReference varRef = compoundAssignment.varRef;
+        // If compound Assignment is an index based expression such as a[f(1, foo)][3][2] += y,
+        // should return a block statement which is equivalent to
+        // var $temp1$ = 2;
```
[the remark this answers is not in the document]
KavinduZoysa: Fixed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
KavinduZoysa: Fixed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
KavinduZoysa: If we change the code according to the above suggestion, the last expression does not go through the logic.
ex: Let's consider the first expression is `c[f(1, foo)][1][2]`. After two iterations, according to this `expr = compoundAssignment.varRef;` the expression becomes `c[f(1, foo)]`. According to the check inside the **while**, execution exits from the loop and `c[f(1, foo)]` does not execute. But `c[f(1, foo)]` should go through the loop.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2328,21 +2328,78 @@ public void visit(BLangXMLNS xmlnsNode) {
...
+            BLangAssignment assignStmt = ASTBuilderUtil.createAssignmentStmt(compoundAssignment.pos, var,
+                    rhsExpression);
+
+            statements.add(assignStmt);
+            // Create block statement. ex: var $temp3$ = a[f(1, foo)];var $temp2$ = 3;var $temp1$ = 2;
+            // a[$temp3$][$temp2$][$temp1$] = a[$temp3$][$temp2$][$temp1$] + y;
+            BLangBlockStmt bLangBlockStmt = ASTBuilderUtil.createBlockStmt(compoundAssignment.pos, statements);
+            result = rewrite(bLangBlockStmt, env);
+        } else {
+            BLangVariableReference varRef = compoundAssignment.varRef;
```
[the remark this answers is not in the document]
KavinduZoysa: Fixed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
KavinduZoysa: Fixed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
KavinduZoysa: Please ignore the above comment. Misread the suggestion. Changed the code according to the suggestion.

SupunS:
@MaryamZi 
> Wondering if we should avoid doing this, and define a variable only for non-literal and non-var ref scenarios?

+1 if possible. But can be a bit tricky. eg:
```ballerina 
a[index][getSecondIndex()] += 1;

function getSecondIndex() returns int {
    index +=1;
    return index;
}
```

MaryamZi:
> @MaryamZi
> 
> > Wondering if we should avoid doing this, and define a variable only for non-literal and non-var ref scenarios?
> 
> +1 if possible. But can be a bit tricky. eg:
> 
> ```ballerina
> a[index][getSecondIndex()] += 1;
> 
> function getSecondIndex() returns int {
>     index +=1;
>     return index;
> }
> ```

Yeah, this would be an issue. 

I guess we can go ahead with the current implementation, unless we want to do this at least for literals. Not sure how much of a gain that would be.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -286,6 +286,7 @@
     private int errorCount = 0;
     private int annonVarCount = 0;
     private int initFuncIndex = 0;
+    private int indexExprNumber = 0;
```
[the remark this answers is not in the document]
KavinduZoysa: Changed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2328,21 +2329,71 @@ public void visit(BLangXMLNS xmlnsNode) {
...
+        // var $temp3$ = a[f(1, foo)];
+        // var $temp2$ = 3;
+        // var $temp1$ = 2;
+        // a[$temp3$][$temp2$][$temp1$] = a[$temp3$][$temp2$][$temp1$] + y;
+        List<BLangStatement> statements = new ArrayList<>();
+        List<BLangSimpleVarRef> varRefs = new ArrayList<>();
+        List<BType> types = new ArrayList<>();
+
+        // Extract the index Expressions from compound assignment and create variable definitions. ex:
+        // var $temp1$ = 2;
```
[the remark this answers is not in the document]
KavinduZoysa: I think we need a comment here because it explains the purpose of loop and corrected the comment.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2328,21 +2329,71 @@ public void visit(BLangXMLNS xmlnsNode) {
...
+        // var $temp2$ = 3;
+        // var $temp3$ = a[f(1, foo)];
+        do {
+            indexExprNumber += 1;
+            BLangSimpleVariableDef tempIndexVarDef = createVarDef("$temp" + indexExprNumber + "$",
```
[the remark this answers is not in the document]
KavinduZoysa: Changed.

tests/jballerina-unit-test/src/test/resources/test-src/statements/compoundassignment/compound_assignment.bal
```diff
@@ -290,3 +290,90 @@ function testCompoundAssignmentAdditionWithRecordAccess() returns int {
...
+}
+
+function testCompoundAssignmentSubtractionArrayElementFunctionInvocation() returns [int, int, int] {
+    var x = 5;
+    Girl girl = {};
+    Boy boy = {};
+    int[][][] b = [[[1, 2, 3], [3, 2, 1], [3, 2, 1]], [[10, 20, 30], [30, 20, 10], [30, 20, 10]],
+    [[5, 6, 7], [7, 6, 5], [7, 6, 5]]];
+
+    b[f(1, girl)][2][g(0, boy)] -= x;
```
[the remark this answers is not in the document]
KavinduZoysa: Changed.

PR #19111, opened 2019-09-19
Add java compatibility check to crt0

SupunS:
## Purpose
With this change, a warning message is printed to the console, if the java version used to compile and run the ballerina program mismatches. Major and minor versions are considered for the compatibility check, and the patch/update versions are not considered.

If a mismatch is found, it will log the warning and will let the program to run. 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/util/RuntimeUtils.java
```diff
@@ -196,4 +198,66 @@ private static String initBRELogHandler() {
...
+     * <li>Java 8 or below: <i>1.8.x</i>, <i>1.7.x</i>, <i>1.6.x</i>, etc.</li>
+     * <li>Java 9 or above: <i>9.x.y.z</i>, <i>10.x.y.z</i>, <i>11.x.y.z</i>, etc.</li>
+     * </ul>
+     * 
+     * @param compiledVersion java version to check the compatibility
+     */
+    public static void verifyJavaCompatibility(String compiledVersion) {
+        String runtimeVersion = System.getProperty(JAVA_VERSION);
+        if (compiledVersion == null || runtimeVersion == null || compiledVersion.equals(runtimeVersion)) {
+            return;
```
warunalakshitha: Can compiledVersion or runtimeVersion be null ?  then don't we need to log an error?

SupunS: Ideally it cannot be null. But added the check for the null-safety.

PR #19139, opened 2019-09-20
Fix BBE issues

anupama-pathirage:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/examples/functional-iteration/functional_iteration.bal
+++ b/examples/functional-iteration/functional_iteration.bal
@@ -26,6 +26,7 @@ public function main() {
     int[] positive = numbers.filter(function (int i) returns boolean {
         return i >= 0;
     });
+    io:println("Positive numbers: ", positive);
 
     // The `forEach()` function executes a specified function once for each of
     // the elements in the container.
@@ -44,7 +45,7 @@ public function main() {
 
     io:println("\nExecution Order:-");
     // The `forEach()` function applies the given function to each item of the container.
-    map<json> j = { name: "apple", colors: ["red", "green"], price: 5 };
+    map<json> j = {name: "apple", colors: ["red", "green"], price: 5};
     j.map(function (json value) returns string {
         string result = value.toString();
         io:println("- map operation's value: ", result);
--- a/examples/functions-with-required-parameters/functions_with_required_parameters.bal
+++ b/examples/functions-with-required-parameters/functions_with_required_parameters.bal
@@ -17,7 +17,7 @@ public function main() {
 
     // Arguments for required parameters can also be passed as named arguments.
     // All arguments after the first named argument need to be passed by name.
-    // Named arguments do not have to be specified in the order.
+    // Named arguments do not have to be specified in the order in which the parameters are defined.
     result = calculate(5, c = 7, b = 6);
     // Print the result.
     io:println(result);
--- a/examples/functions-with-required-parameters/functions_with_required_parameters.bal
+++ b/examples/functions-with-required-parameters/functions_with_required_parameters.bal
@@ -17,7 +17,7 @@ public function main() {
 
     // Arguments for required parameters can also be passed as named arguments.
     // All arguments after the first named argument need to be passed by name.
-    // Named arguments do not have to be specified in order.
+    // Named arguments do not have to be specified in the order.
     result = calculate(5, c = 7, b = 6);
     // Print the result.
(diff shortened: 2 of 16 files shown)


examples/functional-iteration/functional_iteration.bal
```diff
@@ -44,7 +45,7 @@ public function main() {
 
     io:println("\nExecution Order:-");
     // The `forEach()` function applies the given function to each item of the container.
-    map<json> j = { name: "apple", colors: ["red", "green"], price: 5 };
+    map<json> j = {name: "apple", colors: ["red", "green"], price: 5};
```
MaryamZi: The formatting guide seems to have this approach too, but IMO with spaces looks cleaner. We may also have to update other BBEs too.

examples/functions-with-required-parameters/functions_with_required_parameters.bal
```diff
@@ -17,7 +17,7 @@ public function main() {
 
     // Arguments for required parameters can also be passed as named arguments.
     // All arguments after the first named argument need to be passed by name.
-    // Named arguments do not have to be specified in order.
+    // Named arguments do not have to be specified in the order.
```
MaryamZi:
```suggestion
    // Named arguments do not have to be specified in the order in which the parameters are defined.
```

examples/hello-world-parallel/hello_world_parallel.description
```diff
@@ -1,3 +1,3 @@
 // This example uses Ballerina to print "Hello, World!" in parallel using workers.
-// A worker is a sequence of statements that is executed concurrently with all
+// A worker is a sequence of statements that are executed concurrently with all
```
MaryamZi: IINM "is" the correct form here, since it applies to "a sequence"

examples/hello-world-client/hello_world_client.bal
```diff
@@ -3,7 +3,7 @@ import ballerina/io;
 
 public function main() {
     // Create an HTTP client to interact with a remote endpoint.
-    http:Client clientEP = new("http://www.mocky.io");
+    http:Client clientEP = new ("http://www.mocky.io");
```
pubudu91: Doesn't a space here look odd? This is basically like a function call, and we don't use spaces between the function name and the parenthesis

examples/hello-world-service/hello_world_service.client.out
```diff
@@ -1,3 +1,3 @@
-// Invoke the service using the below "curl" command.
+// Invoke the service using this "curl" command.
```
pubudu91:
```suggestion
// Invoke the service using this cURL command.
```

examples/hello-world-client/hello_world_client.bal
MaryamZi: It looks odd to me too. 

The confusion I guess is with this vs specifying the type explicitly, where the arguments kinda look like they are bound to the type. 

```ballerina
http:Client clientEP = new http:Client("http://www.mocky.io");
```

examples/functional-iteration/functional_iteration.bal
anupama-pathirage: Yes. This is according to the style guide and even our IDE tools formats according to that format.. We need to update all BBEs soon.

examples/hello-world-parallel/hello_world_parallel.description
anupama-pathirage: Fixed

examples/hello-world-client/hello_world_client.bal
anupama-pathirage: It seems the rule here is keywords are separated by a space always according to the style guide. IDE formatting is also done according to that. WDYT @NipunaMarcus?

NipunaMarcus: @anupama-pathirage - Yes, keywords should be separated by single space .. especially in this scenario as it's not a function invocation... there seems to be a lack of information regarding this in the style guide... I will update the document to reflect this clearly ...

Issue #19148, opened 2019-09-23
file:readDir does not function properly


lafernando:
The following code:

```ballerina
import ballerina/io;
import ballerina/file;

public function main() returns error? {
    file:FileInfo[] files = check file:readDir("/home/laf/Desktop/ABC");
    foreach var file in files {
        io:println(file.getName());
    }
}
```

The above directory has two files "1.txt" and "2.txt", but it prints the following:-

$ ballerina run x1.bal 
ABC
1.txt
2.txt

Basically it prints the passed in directory itself. But its suppose to list the files and directories inside the given directory only.

aashikam:
Fixed in https://github.com/ballerina-platform/ballerina-lang/pull/19156

PR #19156, opened 2019-09-23
Fix issue in file:readDir() functionality

aashikam:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/19148

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/file/src/test/java/org/ballerinalang/stdlib/file/FileTest.java
+++ b/stdlib/file/src/test/java/org/ballerinalang/stdlib/file/FileTest.java
@@ -203,7 +203,7 @@ public class FileTest {
     public void testReadDirWithMaxDepth() {
         BValue[] args = {new BString(srcDirPath.toString()), new BInteger(0)};
         BValue[] returns = BRunUtil.invoke(compileResult, "testReadDirWithMaxDepth", args);
-        assertTrue(returns[0].getType() instanceof BObjectType);
+        assertEquals(returns.length, 0);
     }
 
     @Test(description = "Test for retrieving files inside directory specifying the depth level - 1")
--- a/stdlib/file/src/test/java/org/ballerinalang/stdlib/file/FileTest.java
+++ b/stdlib/file/src/test/java/org/ballerinalang/stdlib/file/FileTest.java
@@ -35,6 +35,7 @@ import org.testng.Assert;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


stdlib/file/src/test/java/org/ballerinalang/stdlib/file/FileTest.java
```diff
@@ -225,6 +227,14 @@ public void testReadNonExistDirectory() throws IOException {
         assertTrue(((BMap) error.getDetails()).get("message").stringValue().contains("File not found: "));
     }
 
+    @Test(description = "Test for reading file info from an empty directory")
+    public void testReadEmptyDirectory() throws IOException {
```
lafernando: @wggihan @aashikam Isn't there a normal test for this?. That is, reading a directory with files in it. If there was such a test, the issue fixed here would have been caught earlier. This specific test is again just for just a special case. There should be another general test case as well.

wggihan: @lafernando The previous behaviour was there intentionally without realising that is a bug. That why this didn't treated as an issue in the first place.

lafernando: Okay, even with the earlier test "testReadDirWithMaxDepth", it was testing a special case with max depth 0, and also the new test is also a special case where the directory is empty. You need a test which actually has multiple files, and multiple levels, and at least get the full count of files and directories of that.

aashikam: There are two other tests named "testReadDirWithMaxDepth1" which tests reading the directory with `maxDepth` 1 and "testReadDir" which tests reading the file tree recursively.

lafernando: If there were more tests which tests this, and if they were correct written, you would have to update those tests as well right, with this PR?. That is how I guessed there aren't other (proper) tests. I just checked those also, and they don't do much, rather than simply checking the type of the first element returned as the result. 

This is just some friendly advice. Each time a feature is done, or the functionality of a feature is updated. We need to make sure the tests cover the functionality in a reasonable way. That is, first checking the general cases, in order to make sure, any permutations of similar usages will work also to a larger probability. And also, edge cases also, to make sure we if we have (knowingly or unknowingly) special cased certain scenarios. And even the general algorithm may fail in edge cases. And finally, in code reviews, these have to be verified by the attendees, to make sure the feature is done correctly and completely. Same with who reviews PRs.

wggihan: First of all, this implementation done based on an incorrect assumption. To worsen the situation other tests also not checked the return result's size or content. We have fixed that now.

IIRC, we didn't have any code review for this, only a PR review. Without knowing the behavior of Java Files.walk there is a high chance to miss this during that. But I agreed, we should identify this at least in test review. Unfortunately we missed that as well.

lafernando: Yes, my comments were purely based on the incorrect test writing. Let's get them fixed. Thanks!.

Issue #19026, opened 2019-09-12
BIR tree redesign and optimization at JVM codegen


vinok88:
**Description:**
Improve BIR tree redesign and optimization at JVM codegen. Type read and check etc.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #19197, opened 2019-09-25
Optimize compiler backend

SupunS:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/19026

Following optimizations were done:
- Move the isAnydata() and isPureType() checks to compile time. Persist these in BIR and load during runtime. This avoids evaluating the logic over and over again at runtime.
- Skip visiting imports of the modules coming from the cache.
- Use an instruction-tag to differentiate the instructions rather than performing a `is` check.
- Avoid performing type checking on finite/record types for nil lifting( eg: `T|() x;` avoid `x is T`). - Rather do a `x is ()`. This is much faster since checking whether a type is nil is faster than performing a type-check on a complex type.

Time taken by the compiler backend phase for a project with with 4456 ballerina LOC:
- Before optimization: 24-26 sec.
- After optimization: 10-11 sec.

Along with the improvements done in https://github.com/ballerina-platform/ballerina-lang/pull/19190, the total compilation time:
- Before optimization: ~45 sec.
- After optimization: 21-22 sec.

Note: These times are on MacOS.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BTupleType.java
```diff
@@ -81,4 +83,21 @@ public String getDesc() {
...
+        }
+
+        for (BType memberType : this.tupleTypes) {
+            if (!memberType.isPureType()) {
+                this.isAnyData = Optional.of(false);
+                return false;
+            }
+        }
+
+        this.isAnyData = Optional.of(true);
```
MaryamZi: We need to check restType too right?

stdlib/bir-old/src/main/ballerina/src/bir/type_parser.bal
```diff
@@ -223,11 +224,13 @@ public type TypeParser object {
         return obj;
     }
 
-    function parseRecordType() returns BRecordType {
-        BRecordType obj = { moduleId:self.cp.packages[self.readInt32()], name:{value:self.readStringCpRef()},
-                                sealed:self.readBoolean(),
-                    restFieldType: TYPE_NIL, fields: [],
-                    initFunction: {funcType: {}, flags: PRIVATE } };
+    function parseRecordType(int typeFlags) returns BRecordType {
+        BRecordType obj = { moduleId:self.cp.packages[self.readInt32()],
```
MaryamZi:
```suggestion
        BRecordType obj = { moduleId: self.cp.packages[self.readInt32()], 
```
Shall we fix the other lines in this change too?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BRecordType.java
```diff
@@ -109,4 +112,28 @@ public String toString() {
         }
         return this.tsymbol.toString();
     }
+
+    @Override
+    public final boolean isAnydata() {
+        if (!this.isAnyData.isPresent()) {
```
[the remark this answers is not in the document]
SupunS: Same perf impact right?

compiler/ballerina-backend-jvm/src/main/ballerina/src/compiler_backend_jvm/jvm_package_gen.bal
```diff
@@ -129,11 +129,12 @@ public function generatePackage(bir:ModuleID moduleId, @tainted JarFile jarFile,
     string pkgName = getPackageName(orgName, moduleName);
 
     [bir:Package, boolean] [ module, isFromCache ] = lookupModule(moduleId);
-
-    if (isEntry || !isFromCache) {
-        addBuiltinImports(moduleId, module);
+    if(!isEntry && isFromCache) {
+        return;
```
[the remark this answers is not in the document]
SupunS: Yes this is intended. Here we don't need to visit the imports of any module coming from the cache.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -178,6 +181,40 @@ public void remove(BType type) {
...
+
+    @Override
+    public boolean isPureType() {
+        if (this.isPureType.isPresent()) {
+            return this.isPureType.get();
+        }
+
+        for (BType memberType : this.memberTypes) {
+            if (!memberType.isPureType()) {
+                this.isAnyData = Optional.of(false);
```
MaryamZi: Shouldn't this be 
```suggestion
                this.isPureType = Optional.of(false);
```

langlib/lang.map/src/main/java/org/ballerinalang/langlib/map/util/MapLibUtils.java
```diff
@@ -62,6 +62,7 @@ public static BType getCommonTypeForRecordField(BRecordType recordType) {
             typeSet.add(recordType.restFieldType);
         }
 
-        return typeSet.size() == 1 ? typeSet.iterator().next() : new BUnionType(typeSet.toArray(emptyTypeArr));
+        return typeSet.size() == 1 ? typeSet.iterator().next()
+                : new BUnionType(typeSet.toArray(emptyTypeArr), recordType.typeFlags);
```
MaryamZi: Setting `recordType.typeFlags` could be incorrect here right?

For example
```ballerina
type Foo record {
    string s;
    error e;
};
```

Both TypeFlags.ANYDATA  and TypeFlags.PURETYPE would be set for this record, but setting the same for the union created here (`string|error`) is incorrect, since this not `anydata`?

stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/Message.java
```diff
@@ -381,7 +381,7 @@ private void updateBMapValue(BType bType, MapValue<String, Object> bMapValue,
         Descriptors.OneofDescriptor oneofDescriptor = fieldDescriptor.getContainingOneof();
         String msgType = oneofDescriptor.getContainingType().getName() + "_" + toCamelCase
                 (fieldDescriptor.getName());
-        MapValue<String, Object> bMsg = new MapValueImpl<>(new BRecordType(msgType, bType.getPackage(), 0, true));
+        MapValue<String, Object> bMsg = new MapValueImpl<>(new BRecordType(msgType, bType.getPackage(), 0, true, 0));
```
MaryamZi: Do we know what this is used for? Wondering if we should add TypeFlags.ANYDATA, TypeFlags.PURETYPE?

stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/ServicesBuilderUtils.java
```diff
@@ -256,7 +257,7 @@ static BType getBallerinaValueType(BPackage bPackage, String protoType) {
         } else if (protoType.equalsIgnoreCase(WRAPPER_BYTES_MESSAGE)) {
             return new BArrayType(BTypes.typeByte);
         } else {
-            return new BRecordType(protoType, bPackage, 0, true);
+            return new BRecordType(protoType, bPackage, 0, true, TypeFlags.asMask(TypeFlags.ANYDATA));
```
MaryamZi: Shouldn't we add TypeFlags.PURETYPE too? Since anything that is anydata is pure?

MaryamZi: Applies in other places too - AbstractSQLStatement.java

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/CallStatement.java
```diff
@@ -142,7 +142,7 @@ public Object execute() {
     private ArrayValue constructTablesForResultSets(List<ResultSet> resultSets, TableResourceManager rm,
                                                      ArrayValue structTypes, String databaseProductName)
             throws SQLException, ApplicationException {
-        BRecordType tableConstraint = new BRecordType("$table$anon$constraint$", null, 0, false);
+        BRecordType tableConstraint = new BRecordType("$table$anon$constraint$", null, 0, false, 0);
```
MaryamZi: Again should we set anydata and pure here too?

stdlib/utils/src/main/java/org/ballerinalang/utils/Next.java
```diff
@@ -86,6 +87,8 @@ public static Object next(Strand strand, Object iterator) {
     private static org.ballerinalang.jvm.types.BType getFiniteType(Object value) {
         Set<Object> valueSpace = new HashSet<>();
         valueSpace.add(value);
-        return new BFiniteType(value.toString(), valueSpace);
+        int masks = value == null ? TypeFlags.NILABLE : 0;
+        return new BFiniteType(value.toString(), valueSpace,
+                TypeFlags.addToMask(masks, TypeFlags.ANYDATA, TypeFlags.PURETYPE));
```
MaryamZi: Don't we have to get the type of the value and check if it is anydata/pure?

tests/ballerina-test-utils/src/main/java/org/ballerinalang/test/util/BRunUtil.java
```diff
@@ -712,12 +714,24 @@ private static Object getJVMValue(BType type, BValue value) {
...
+                        getMask(finiteType.isNullable(), finiteType.isAnydata(), finiteType.isPureType()));
             default:
                 throw new RuntimeException("Function argument for type '" + type + "' is not supported");
         }
     }
 
+    private static int getMask(boolean nilable, boolean isAnydata, boolean isPureType) {
+        int mask = nilable ? TypeFlags.NILABLE : 0;
+        if (isAnydata) {
+            mask = TypeFlags.addToMask(mask, TypeFlags.ANYDATA);
```
MaryamZi: a value being anydata implies that it is pure right? should we update this as 
```ballerina
        if (isAnydata) {
            mask = TypeFlags.addToMask(mask, TypeFlags.ANYDATA);
            mask = TypeFlags.addToMask(mask, TypeFlags.PURETYPE);
        } else if (isPureType) {
            mask = TypeFlags.addToMask(mask, TypeFlags.PURETYPE);
        }
```

MaryamZi: We could do this in multiple places, if there could potentially be a gain.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BTupleType.java
```diff
@@ -41,18 +42,36 @@ public BTupleType(List<BType> typeList) {
         super(null, null, Object.class);
         this.tupleTypes = typeList;
         this.restType = null;
+
+        boolean isAllMembersPure = true;
+        for (BType memberType : tupleTypes) {
+            isAllMembersPure &= memberType.isPureType();
+        }
+
+        if (isAllMembersPure) {
```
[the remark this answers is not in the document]
SupunS: Rest-type is not accepted by this constructor (always null)..

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BUnionType.java
```diff
@@ -108,11 +131,22 @@ public boolean equals(Object o) {
...
         }
+
         BUnionType that = (BUnionType) o;
-        return memberTypes.containsAll(that.memberTypes) && that.memberTypes.containsAll(memberTypes);
+        if (this.memberTypes.size() != that.memberTypes.size()) {
+            return false;
+        }
+
+        for (int i = 0; i < memberTypes.size(); i++) {
+            if (!this.memberTypes.get(i).equals(that.memberTypes.get(i))) {
```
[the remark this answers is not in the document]
SupunS: Here we check for whether the two unions are the 'same' rather than the assignability/equality. So the simpler case would be fast. If it fails, then it will always fall back to the assignable check. That would satisfy the equality of  `int|string` and `string|int`

stdlib/utils/src/main/java/org/ballerinalang/utils/Next.java
SupunS: We reach here only if the value is primitive. Hence it is safe to add these.

tests/ballerina-test-utils/src/main/java/org/ballerinalang/test/util/BRunUtil.java
SupunS: I think its safe to depend on the respective flag to make it future proof. If someday we update the compiler type checking (in such a way that puretype and anydata doesn't have a direct relationship), then we don't have to worry about the runtime.

Kishanthan:
Superb stuff @SupunS. This is a great improvement.

PR #19199, opened 2019-09-25
Fix access modifiers of transactions module types

anupama-pathirage:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/transactions/src/main/ballerina/src/transactions/2pc_participant.bal
+++ b/stdlib/transactions/src/main/ballerina/src/transactions/2pc_participant.bal
@@ -28,12 +28,12 @@ type LocalProtocol record {
 # + name - protocol name
 # + url - protocol URL. This URL will have a value only if the participant is remote. If the participant is local,
 #         the `protocolFn` will be called
-public type RemoteProtocol record {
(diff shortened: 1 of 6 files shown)


hasithaa:
Should we add this to the minor release due to API changes? Still this is an internal module.

anupama-pathirage:
> Should we add this to the minor release due to API changes? Still this is an internal module.

I think it is ok to add this to the patch release due to following reasons

- The transactions module is somewhat different from other modules as it contains mostly internal stuff related to the ballerina transaction manager. There is only one function which should be exposed to the external. (`getCurrentTransactionId()`function) So if someone is using other functions/types which are currently public, it will result in either a malfunctioning or a buggy code. Since we are not changing the API of that public function, I think we can change this in a patch release.

- Transactions is a experimental feature in ballerina.

WDYT? @hasithaa @warunalakshitha @rdhananjaya

PR #19272, opened 2019-10-01
Fix closure issues when using transaction blocks

### LINKED ISSUE #17288: Transactions compilation fails when inside if block
**Description:**

Use a transaction block inside if condition and try to access variable define in the method scope. Compilation will fail..

irshadnilam:
## Purpose
$title

Fixes #17288 and multiple other closure issues (ex: using an arrow function inside transaction)

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2799,6 +2805,27 @@ private void markAndRegisterClosureVariable(BSymbol symbol, DiagnosticPos pos) {
...
+                }
+                break;
+            } else {
+                SymbolEnv enclEnv = cEnv.enclEnv;
+                if (enclEnv == null) {
+                    break;
+                }
+                cEnv = enclEnv;
+                node = cEnv.node;
+            }
```
warunalakshitha: Why we need else part here? Also shall we add comments for whole logic.

irshadnilam: @warunalakshitha 
What the logic does is, check if the variable is used inside a transaction function to mark it as closure instead of marking all the variables inside transaction as a closure. 

The `else` part is required to traverse through parent node until a function node is met.

I have added a comment above the logic.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2270,34 +2270,24 @@ public void visit(BLangCatch bLangCatch) {
 
     @Override
     public void visit(BLangTransaction transactionNode) {
-        analyzeStmt(transactionNode.transactionBody, env);
+        SymbolEnv transactionEnv = SymbolEnv.createTransactionEnv(transactionNode, env);
```
rdhananjaya: Just a questions: why do we need to create a new symbol environment here?

irshadnilam: So that we can identify transactionNode type in Typechecker when the variable is visited.

Please feel free to suggest if there is any other better way to do this, that i might be missing.

rdhananjaya: ack

Issue #19216, opened 2019-09-26
json value get consumed after the string conversion


chanukaranaba:
**Description:**
Refer the following code;
```ballerina
            json jsonConversionRet = jsonutils:fromTable(selectRet);
            io:print("JSON: ");
            io:println(jsonConversionRet.toJsonString());
            io:println("--------------- ");
            io:print("JSON 2: ");
            io:println(jsonConversionRet);
```
JSON 2 value is blank, it means jsonConversionRet.toJsonString() consumed the stream and jsonConversionRet doesn't have value anymore. Is this is expected behavior?
From a developer perspective, the json value shouldn't be blank. 

**Steps to reproduce:**

**Affected Versions:**
1.0.0

MaryamZi:
This is a bug with the `.toJsonString()` method. 

As @SupunS pointed out `getJSONString()` has not been implemented for `StreamingJsonValue`. This needs to be implemented adding https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/StreamingJsonValue.java#L125.

rdhananjaya:
Similar issue is present in `copy` function.
Please refer [ballerina-dev mail thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/ballerina-dev/zoPlcNMf_lQ/BAQciuTkCgAJ)

rdhananjaya:
Similar issue is present in `copy` function.
Please refer [ballerina-dev mail thread](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/ballerina-dev/zoPlcNMf_lQ/BAQciuTkCgAJ)

hasithaa:
Fixed with #19285

PR #19285, opened 2019-10-03
Fix unmaterialized StreamingJson issue

rdhananjaya:
## Purpose
There are situations where StreamingJsonValue convert it's value to some other form without materializing the stream, there returning empty value.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/19216

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/StreamingJsonValue.java
```diff
@@ -150,6 +151,20 @@ void appendToCache(Object value) {
         super.add(size, value);
     }
 
+    @Override
+    public String getJSONString() {
+        // Consume and materialize the stream.
+        buildDatasource();
```
MaryamZi: Wouldn't it be better to check if `datasource.hasNext()` and then do `buildDatasource()` like it has been done for the others?

rdhananjaya: buildDatasource does that right.

MaryamZi: Just that we'll be avoiding the try-catch block but in this scenario that shouldn't be much of a concern I guess. 

+1 to keep as is, since with the if check we seem to be duplicating the hasNext() check anyway.

rdhananjaya:
> Can we please add tests? I guess we should be able to come up with some tests using the `ballerinax/java.jdbc` module?

Issue is where to place the tests, unit test module doesn't have any DB tests at the moment and adding json test in java.jdbc module doesn't feel right. WDY think about adding an integration test, in TableDataStreamingTestCase

MaryamZi:
> > Can we please add tests? I guess we should be able to come up with some tests using the `ballerinax/java.jdbc` module?
> 
> Issue is where to place the tests, unit test module doesn't have any DB tests at the moment and adding json test in java.jdbc module doesn't feel right. WDY think about adding an integration test, in TableDataStreamingTestCase

We discussed a similar case some time ago, in relation to WebSub, and I remember @anupama-pathirage mentioning we can add it to the jdbc tests for the moment. @anupama-pathirage thoughts please?

I'm +1 to adding it along with the db temporarily at least, or we might run into regressions.

PR #19346, opened 2019-10-10
Fix int range overflow for integer division

### LINKED ISSUE #17969: [Spec Deviation] Integer division should panic on overflow
**Description:**
The 2019r3 spec says

for int,
" / performs integer division, with any fractional part discarded ie with truncation towards zero; a panic will occur on division by zero or overflow, which happens if the first operand is -2 ^ 63 and the second operand is -1"

**Affected Versions:**
1.0.0-beta-SNAPSHOT - 15th August build

vinok88:
Fixes #17969

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/MathUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/MathUtils.java
@@ -31,7 +31,7 @@ public class MathUtils {
             if (numerator == Math.pow(-2, 63) && denominator == -1) {
                 // a panic will occur on division by zero or overflow,
                 // which happens if the first operand is -2^63 and the second operand is -1
-                throw BallerinaErrors.createError(BallerinaErrorReasons.Number_OVERFLOW, " int range overflow");
+                throw BallerinaErrors.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, " int range overflow");
             }
             return numerator / denominator;
         } catch (ArithmeticException e) {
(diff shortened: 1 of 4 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/MathUtils.java
```diff
@@ -28,6 +28,11 @@
 
     public static long divide(long numerator, long denominator) {
         try {
+            if (numerator == Math.pow(-2, 63) && denominator == -1) {
```
rdhananjaya: Integer.MIN_VALUE

vinok88: Do you mean java int min value? that is -2^31.

vinok88: I think i should use Long.Min value. :)

rdhananjaya: Yes the Long version :D

vinok88: Fixed. Thanks

PR #19401, opened 2019-10-14
Revert "Enable all tests for streams module"

grainier:
Reverts ballerina-platform/ballerina-lang#19261


grainier:
> Shall we create an issue to track enabling these if we don't already have one?

Created https://github.com/ballerina-platform/ballerina-lang/issues/19403 to keep track

PR #19505, opened 2019-10-21
Enable multi-line strings with a warning

### LINKED ISSUE #19509: Disallowing multi-line strings breaks existing user code
**Description:**
Since multi-line strings are not allowed in the spec, it was disallowed in the implementation as well in the 1.0.x branch. However, this causes user code to break. 

**Steps to reproduce:**

**Affected Versions:**

pubudu91:
> Fixes #19509


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1111,6 +1111,10 @@ warning.usage.of.deprecated.function=\
 warning.non.module.qualified.error.reason=\
   error reason ''{0}'' is not module qualified
 
+warning.multi.line.strings.not.allowed=\
+  multi-line strings are not allowed by the Ballerina specification version 2019R3. Currently this is allowed due \
```
```suggestion
  multi-line strings are not allowed by the Ballerina specification version 2019R3. This is currently allowed due \
```

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1111,6 +1111,10 @@ warning.usage.of.deprecated.function=\
...
+  to a bug in the compiler (https://git.io/JeRY5) and therefore, will be removed in a future release.
```
```suggestion
  to a bug in the compiler (https://git.io/JeRY5) and will be removed in a future release.
```

Issue #19610, opened 2019-10-28
Taint flow analysis does not seems to work with 'check' expressions


rdhananjaya:
**Description:**
Consider this: https://github.com/ECLK/Results-Dist/blob/83fddefd27a0d7e939795e7f31388a802fa4cee3/subscriber/src/subscriber/save.bal#L95

In this sample `check` also need to enforce the marking of return tainted just as return statement.

**Steps to reproduce:**

**Affected Versions:**
1.0.2
**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #19630, opened 2019-10-29
Improve taint propagation analysis for 'check' expressions

rdhananjaya:
## Purpose
$subject

As `check` expression may cause a function to return with provided error, we need to consider check expressions when calculating the return value taintedness of a function.

```ballerina
    error? k = foo();
}

    _ = check bar(); // check may return the error, which is tainted
}

function bar() returns @tainted string|error {
    return error("reason", message = "tainted data here");
}
```

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/19610

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/http/src/main/ballerina/src/http/resiliency/http_retry_client.bal
+++ b/stdlib/http/src/main/ballerina/src/http/resiliency/http_retry_client.bal
@@ -69,7 +69,7 @@ public type RetryClient client object {
     # + message - An HTTP outbound request message or any payload of type `string`, `xml`, `json`, `byte[]`,
     #             `io:ReadableByteChannel` or `mime:Entity[]`
     # + return - The HTTP `Response` message, or an error if the invocation fails
-    public remote function post(string path, RequestMessage message) returns Response|ClientError {
+    public remote function post(string path, RequestMessage message) returns @tainted Response|ClientError {
         var result = performRetryAction(path, <Request>message, HTTP_POST, self);
         if (result is HttpFuture) {
             return getInvalidTypeError();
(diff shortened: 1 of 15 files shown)


hasithaa:
@wggihan Pls review changes.

stdlib/http/src/main/ballerina/src/http/resiliency/http_retry_client.bal
```diff
@@ -69,7 +69,7 @@ public type RetryClient client object {
     # + message - An HTTP outbound request message or any payload of type `string`, `xml`, `json`, `byte[]`,
     #             `io:ReadableByteChannel` or `mime:Entity[]`
     # + return - The HTTP `Response` message, or an error if the invocation fails
-    public remote function post(string path, RequestMessage message) returns Response|ClientError {
+    public remote function post(string path, RequestMessage message) returns @tainted Response|ClientError {
         var result = performRetryAction(path, <Request>message, HTTP_POST, self);
```
ldclakmal: Don't we need to sanitize the return of the `performRetryAction` method here and return the sanitized output?
Because IMO the taintedness should not be passed to the end-user who uses the HTTP client actions with retry config.

rdhananjaya: I don't see how we can sanitize the values in a general way given that they can contain arbitrary data

ldclakmal: Yes. Agree with that. IMO, we need to fix https://github.com/ballerina-platform/ballerina-lang/issues/19844.

Issue #19804, opened 2019-11-06
Re-enable support for using mapping constructors with record binding patterns


pubudu91:
Support for using record binding patterns with mapping constructors was removed with https://github.com/ballerina-platform/ballerina-lang/pull/11053/commits/009046794e35f5ea8ab3c986176b2248734be762

Need  to re-enable this. I had a quick look at this, with the check added here to disable the support removed. It seems to be working. Need to test more and re-enable.

dulajdilshan:
https://github.com/ballerina-platform/ballerina-lang/pull/40283/ only enables destructuring assignments.

PR #19861, opened 2019-11-13
Fix usages of "invalid literal for type" error message

### LINKED ISSUE #17360: Cannot figure out actual error from the error message
**Description:**
```ballerina

import ballerina/http;

http:Client circuitBreakerEP = new("http://localhost:9092", {
        circuitBreaker: {
            timeoutMillis: 2000
        }
    });
```

When trying to compile the above code, following error is shown:
```
error: .::Untitled-1.bal:5:25: invalid literal for type 'ballerina/http:CircuitBreakerConfig?'
```

This doesn't say what the error is, or no information on how to resolve the error.

**Affected Versions:**
1.0-alpha

pubudu91:
Fixes #17360

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
@@ -190,7 +190,6 @@ public enum DiagnosticCode {
 
     INVALID_LITERAL_FOR_TYPE("invalid.literal.for.type"),
     INCOMPATIBLE_MAPPING_CONSTRUCTOR("incompatible.mapping.constructor.expression"),
-    MISSING_FIELDS_IN_MAPPING_CONSTRUCTOR("missing.fields.in.mapping.constructor"),
     MAPPING_CONSTRUCTOR_COMPATIBLE_TYPE_NOT_FOUND("mapping.constructor.compatible.type.not.found"),
     CANNOT_INFER_TYPES_FOR_TUPLE_BINDING("cannot.infer.types.for.tuple.binding"),
     INVALID_LITERAL_FOR_MATCH_PATTERN("invalid.literal.for.match.pattern"),
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
@@ -189,6 +189,8 @@ public enum DiagnosticCode {
     UNSAFE_CONVERSION_ATTEMPT("unsafe.conversion.attempt"),
 
     INVALID_LITERAL_FOR_TYPE("invalid.literal.for.type"),
+    INCOMPATIBLE_MAPPING_CONSTRUCTOR("incompatible.mapping.constructor.expression"),
+    MAPPING_CONSTRUCTOR_COMPATIBLE_TYPE_NOT_FOUND("mapping.constructor.compatible.type.not.found"),
     INVALID_LITERAL_FOR_MATCH_PATTERN("invalid.literal.for.match.pattern"),
     INVALID_EXPR_WITH_TYPE_GUARD_FOR_MATCH_PATTERN("invalid.expr.with.type.guard.for.match"),
     ARRAY_LITERAL_NOT_ALLOWED("array.literal.not.allowed"),
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
@@ -190,6 +190,7 @@ public enum DiagnosticCode {
 
     INVALID_LITERAL_FOR_TYPE("invalid.literal.for.type"),
     INCOMPATIBLE_MAPPING_CONSTRUCTOR("incompatible.mapping.constructor.expression"),
+    MISSING_FIELDS_IN_MAPPING_CONSTRUCTOR("missing.fields.in.mapping.constructor"),
     MAPPING_CONSTRUCTOR_COMPATIBLE_TYPE_NOT_FOUND("mapping.constructor.compatible.type.not.found"),
     CANNOT_INFER_TYPES_FOR_TUPLE_BINDING("cannot.infer.types.for.tuple.binding"),
     INVALID_LITERAL_FOR_MATCH_PATTERN("invalid.literal.for.match.pattern"),
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
@@ -191,6 +191,7 @@ public enum DiagnosticCode {
     INVALID_LITERAL_FOR_TYPE("invalid.literal.for.type"),
     INCOMPATIBLE_MAPPING_CONSTRUCTOR("incompatible.mapping.constructor.expression"),
     MAPPING_CONSTRUCTOR_COMPATIBLE_TYPE_NOT_FOUND("mapping.constructor.compatible.type.not.found"),
+    CANNOT_INFER_TYPES_FOR_TUPLE_BINDING("cannot.infer.types.for.tuple.binding"),
     INVALID_LITERAL_FOR_MATCH_PATTERN("invalid.literal.for.match.pattern"),
     INVALID_EXPR_WITH_TYPE_GUARD_FOR_MATCH_PATTERN("invalid.expr.with.type.guard.for.match"),
     ARRAY_LITERAL_NOT_ALLOWED("array.literal.not.allowed"),
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -153,6 +153,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
+
 import javax.xml.XMLConstants;
 
 import static org.wso2.ballerinalang.compiler.tree.BLangInvokableNode.DEFAULT_WORKER_NAME;
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -959,8 +959,23 @@ public class TypeChecker extends BLangNodeVisitor {
         }
 
         BUnionType unionType = (BUnionType) expType;
-        Set<BType> memberTypes = unionType.getMemberTypes();
+        BType[] memberTypes = unionType.getMemberTypes().toArray(new BType[0]);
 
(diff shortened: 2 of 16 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -944,8 +943,31 @@ public void visit(BLangRecordLiteral recordLiteral) {
         }
     }
 
-    private List<BType> getRecordCompatibleType(BType bType, BLangRecordLiteral recordLiteral) {
+    private boolean isMappingConstructorCompatibleType(BType type) {
+        return type.tag == TypeTags.RECORD || type.tag == TypeTags.MAP || type.tag == TypeTags.JSON;
+    }
+
+    private void reportIncompatibleMappingConstructorError(BLangRecordLiteral mappingConstructorExpr, BType expType) {
```
pubudu91: Should we special case the `T?` case where T is a record? If so, can provide an even better error message for the particular issue mentioned in #17360

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -944,8 +944,31 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+            return;
+        }
+
+        BUnionType unionType = (BUnionType) expType;
+        Set<BType> memberTypes = unionType.getMemberTypes();
+
+        for (BType bType : memberTypes) {
+            if (isMappingConstructorCompatibleType(bType)) {
+                dlog.error(mappingConstructorExpr.pos, DiagnosticCode.INCOMPATIBLE_MAPPING_CONSTRUCTOR, unionType);
+                return;
```
gimantha: should we not print all the incompatible member types at once?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/worker/WaitActionsNegativeTest.java
```diff
@@ -79,7 +79,9 @@ public void testNegativeWorkerActions() {
         BAssertUtil.validateError(resultNegative, index++,
                 "missing non-defaultable required record field 'f2'", 86, 25);
         BAssertUtil.validateError(resultNegative, index++,
-                "invalid literal for type 'sealedRec'", 87, 26);
+                                  "incompatible types: expected 'sealedRec', found 'record {| other id; other name; " +
+                                          "other status; |}",
```
hasithaa: Not related to this, but Saying "other" is confusing here.  We need to come up with a better name for `other`.  One option is to make it special. like, `<unknown_type>` Shall we create an issue?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1184,3 +1184,12 @@ error.invalid.package.name.qualifier=\
...
   identifier literal only supports alphanumeric characters
+
+error.incompatible.mapping.constructor.expression=\
+  incompatible mapping constructor expression for type ''{0}''
+
+error.mapping.constructor.compatible.type.not.found =\
+  a type compatible with mapping constructor expressions not found in type ''{0}''
+
+error.cannot.infer.types.for.tuple.binding =\
+  invalid expression: types cannot be inferred for ''{0}''
```
hasithaa: The error message is not related to a tuple, but the error key says tuple binding.

hasithaa: It seems like no test case is added for this particular case.

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
```diff
@@ -189,6 +189,9 @@
     UNSAFE_CONVERSION_ATTEMPT("unsafe.conversion.attempt"),
 
     INVALID_LITERAL_FOR_TYPE("invalid.literal.for.type"),
```
hasithaa: Are we removing all usage of this with this PR?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1184,3 +1184,12 @@ error.invalid.package.name.qualifier=\
 
 error.identifier.literal.only.supports.alphanumerics=\
   identifier literal only supports alphanumeric characters
+
+error.incompatible.mapping.constructor.expression=\
+  incompatible mapping constructor expression for type ''{0}''
```
hasithaa: Shall we add test cases for this one as well?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
hasithaa: +1, `T?` is 80% of the case, when considering unions.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/worker/WaitActionsNegativeTest.java
pubudu91: Created #19876 to track this.

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
pubudu91: There's one usage in record binding patterns. #19804 is to take a look and address this as well.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
pubudu91: This is just to isolate instances where the union has a type with which we can use record literals (i.e., records, maps an JSON), but for this particular case, the given record literal isn't compatible with that type.

compiler/ballerina-lang/src/main/resources/compiler.properties
pubudu91: This error is only for tuples (lists). Will modify the error message to make it clear.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1870,6 +1894,23 @@ public void visit(BLangWaitForAllExpr waitForAllExpr) {
         }
     }
 
+    private BRecordType getWaitForAllExprReturnType(List<BLangWaitForAllExpr.BLangWaitKeyValue> keyVals) {
+        BRecordType retType = new BRecordType(null);
+        
+        for (BLangWaitForAllExpr.BLangWaitKeyValue keyVal : keyVals) {
+            BSymbol symbol = symResolver.lookupSymbol(env, names.fromIdNode(keyVal.key), SymTag.VARIABLE);
```
rdhananjaya: I think what we expect here is to get the constraint type of the `future`
If so I think we need to do something in the line of 
```java
BType t = ((BFutureType) checkExpr(keyVal.valueExpr, env)).constraint
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: How about cases like 

`Foo|Bar x = {};`

where both `Foo` and `Bar` are records but `{}` is not valid for either of them?

That still reaches here right? If it does IMO, we should mention both of them.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -947,8 +947,63 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                       expType);
+            return;
+        }
+
+        BUnionType unionType = (BUnionType) expType;
+        BType[] memberTypes = unionType.getMemberTypes().toArray(new BType[0]);
+
+        // Special case handling for `T?` where T is a record type. This is done to give more user friendly error
+        // messages for this common scenario.
+        if (memberTypes.length == 2) {
```
MaryamZi: Can't we make this a generic check for any union with just a single record type?

MaryamZi: e.g., `Foo|int?`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
pubudu91: Not quite. Consider something like below:
```ballerina
int x = wait {w1, w2};
```
What I want is to create a record type for the RHS so that in the error message I can say expected `int`, but found this type of record. If I call checkExpr(), it'll go in the normal type checking flow, which means if there are any errors there, the compiler will log errors for those as well.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
pubudu91: Can do. But went with the `T?` case (at least for now) since it's the most common one.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
pubudu91: The error message says the mapping constructor is not valid for the union. It doesn't mention individual types. For example, for the above case you mentioned:
```
incompatible mapping constructor expression for type '(Foo|Bar)'
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
rdhananjaya: If I understand correctly your concern is around duplicate error message due to `checkExpr` right,
```ballerina
 int x = wait { w1+1, nonFutureRef };
```
Because of caching I don't think there will be multiple error messages.

Anyway for `wait { someKey: w1, someKey1: w1 }` resolving symbol for the `key` will always result in undefined symbol (other)

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: ah yeah, missed that we were using `unionType`, sorry!

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
pubudu91: Fixed this with 1be935a. Have a look

PR #19926, opened 2019-11-17
Add default values to generated record types

### LINKED ISSUE #19245: [gRPC][Protobuf] Set default values to the generated record fields
**Description:**
Default values are not set in generated record fields. So it is mandatory to set values to all fields of the record. but normally in gRPC communication, we don't need to set values to all fields, is not set default value is applied.

daneshk:
## Purpose
This is to avoid getting an error when dispatching the message which has nil values.

Fixes #19245

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/builder/BallerinaFileBuilder.java
+++ b/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/builder/BallerinaFileBuilder.java
@@ -304,6 +304,24 @@ public class BallerinaFileBuilder {
 
             return result;
         });
+        handlebars.registerHelper("isNotNull", (object, options) -> {
+            CharSequence result;
+            if (object != null) {
+                result = options.fn(options.context);
+            } else {
+                result = null;
+            }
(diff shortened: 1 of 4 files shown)


stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/builder/BallerinaFileBuilder.java
```diff
@@ -304,6 +304,24 @@ private static Template compileTemplate(String defaultTemplateDir, String templa
 
             return result;
         });
+        handlebars.registerHelper("isNotNull", (object, options) -> {
+            CharSequence result;
+            if (object != null) {
```
wggihan: @daneshk If possible, we need to extract this to a separate method and use in both places.

PR #19929, opened 2019-11-17
Check object initialization for non existing init function

### LINKED ISSUE #19566: New expression with args list compiles for an object with implicit __init method
Consider the following program:
```ballerina
public function main() {
    Person? p = new("Pubudu");
}

type Person object {
    string name = "";
};
```
The above program compiles and runs. It should actually give a compile error since there isn't an `__init` method with a string param for the `Person` object.

KavinduZoysa:
## Purpose
$subject when args are passed.

Fixes #19566

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


hasithaa:
This is not a patch compatible fix, as this will break the existing code. isn't it ?

KavinduZoysa:
> This is not a patch compatible fix, as this will break the existing code. isn't it ?

Yes, therefore the changes are pushed to the master branch via PR #19945.

Issue #19769, opened 2019-11-04
Change generated code use BString value


manuranga:
Codegen will use BString but it's translated to java string when calling stdlib or langlib.

PR #19942, opened 2019-11-18
Introduce StringValue to represent ballerina string

manuranga:
completes #19769

### DIFF
--- a/compiler/ballerina-backend-jvm/src/main/ballerina/src/compiler_backend_jvm/jvm_instruction_gen.bal
+++ b/compiler/ballerina-backend-jvm/src/main/ballerina/src/compiler_backend_jvm/jvm_instruction_gen.bal
@@ -17,6 +17,14 @@
 import ballerina/io;
 import ballerina/bir;
 import ballerina/jvm;
+import ballerina/runtime;
+
+boolean IS_BSTRING = runtime:getProperty("ballerina.bstring") != "";
+string BSTRING_VALUE = runtime:getProperty("ballerina.bstring") == "" ? STRING_VALUE : I_STRING_VALUE;
+
+const string I_STRING_VALUE = "org/ballerinalang/jvm/values/StringValue";
+const string BMP_STRING_VALUE = "org/ballerinalang/jvm/values/BmpStringValue";
+const string NON_BMP_STRING_VALUE = "org/ballerinalang/jvm/values/NonBmpStringValue";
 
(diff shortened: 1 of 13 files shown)


manuranga:
This is a thin slice MVP.
Contains a unit test that has string concat using new StringValue.
System prop is used to switch on this behaviour not to break other code.

compiler/ballerina-backend-jvm/src/main/ballerina/src/compiler_backend_jvm/jvm_instruction_gen.bal
```diff
@@ -17,6 +17,22 @@
 import ballerina/io;
 import ballerina/bir;
 import ballerina/jvm;
+import ballerina/runtime;
+
+boolean IS_BSTRING = runtime:getProperty("ballerina.bstring") != "";
+string BSTRING_VALUE = getBStingClassName();
```
lankavitharana: shall we use a ternary operator here instead of a method call? and can use the IS_BSTRING property right?

compiler/ballerina-backend-jvm/src/main/ballerina/src/compiler_backend_jvm/jvm_instruction_gen.bal
```diff
@@ -41,8 +57,28 @@ type InstructionGenerator object {
             any val = loadIns.value;
             self.mv.visitLdcInsn(val);
         } else if (bType is bir:BTypeString) {
-            any val = loadIns.value;
-            self.mv.visitLdcInsn(val);
+            string val = <string> loadIns.value;
+            if (IS_BSTRING) {
+                int[] highSurrogates = listHighSurrogates(val);
```
lankavitharana: this is nice :)

tests/ballerina-test-utils/src/main/java/org/ballerinalang/test/util/BRunUtil.java
```diff
@@ -403,6 +404,7 @@ private static Method getMethod(String functionName, Class<?> funcClass) throws
                     org.ballerinalang.jvm.types.BTypes.typeAny);
             scheduler.start();
             if (futureValue.panic instanceof RuntimeException) {
+                futureValue.panic.printStackTrace();
```
lankavitharana: do we need this?

PR #19949, opened 2019-11-18
[Filepath] Migrate native functions to java interoperability

### LINKED ISSUE #19896: [Filepath] Migrate to java interoperability
**Description:**
$title

aashikam:
Fixes #19896

PR to master #19960

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/filepath/src/main/java/org/ballerinalang/stdlib/filepath/nativeimpl/FilepathUtils.java
```diff
@@ -18,32 +18,56 @@
...
+ * @since 1.0.5
  */
-@BallerinaFunction(
-        orgName = Constants.ORG_NAME,
-        packageName = Constants.PACKAGE_NAME,
-        functionName = "resolve",
-        isPublic = true
-)
-public class Resolve {
+public class FilepathUtils {
```
wggihan: What is the correct name is it FilepathUtils or FilePathUtils?

aashikam: IMO it doesn't matter. This is the class with all the native implementations of the module `ballerina/filepath`. In this context, this class is only used within this module and with the name `FilepathUtils` it is obvious what the class is for.

wggihan: But there is a Java naming convention that needs to follow. We can't come up with random class names. All ballerina modules use lower case names. Are we using Rabbitmq instead of RabbitMQ?

aashikam: It is not a random class name. There is no standard for the word filepath. People have used it in different ways. As for RabbitMQ that is the name of the server. As I said earlier, in a clean code perspective this doesn't add any value or an issue if the name is either `FilePath` or `Filepath`. It is merely an objective argument. I have changed the name anyway. Also ballerina/http module has class names with both `HTTP` prefix and `Http` prefix. So I don't see that there is a java naming convention followed anywhere. ([HttpConstants](https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/http/src/main/java/org/ballerinalang/net/http/HttpConstants.java), [HTTPServicesRegistry](https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/http/src/main/java/org/ballerinalang/net/http/HTTPServicesRegistry.java))

wggihan: This is not about clean code. It about standard. Don't mix two things. When someone do a incorrect thing, it doesn't warrant us to do the same mistake.

aashikam: Is there a reference link to this standard that I can refer to? Because if there is a standard that has to be followed it is never corrected in the entire http module so far. 
https://github.com/ballerina-platform/ballerina-lang/tree/master/stdlib/http/src/main/java/org/ballerinalang/net/http
Also I would like to know why `Filepath` is incorrect?

aashikam: Looking as the existing code base, one can definitely argue that it is the standard of naming (Http) in the ballerina java source files.

wggihan: Did I mention Filepath is incorrect? 

The naming convention guide provided by Oracle https://www.oracle.com/technetwork/java/codeconventions-135099.html

aashikam: If Filepath was not incorrect then what was the issue here? The util class (Utils) for the ballerina/filepath module (Filepath) = FilepathUtils.

wggihan: FYI, ballerina module names is a different standard compare to the Java naming convention. So please use correct Java standard in the Java implementation and correct Ballerina standard in the Ballerina implentation. Don't mix both since those are two different languages and have different constructs.

Issue #19711, opened 2019-11-01
Cleanup current ballerina dist commands


keizer619:
**Description:**
Cleanup current ballerina dist commands from which was implemented from picco cli

PR #20017, opened 2019-11-21
Remove update tool logics from jballerina

keizer619:


keizer619:
Resolves https://github.com/ballerina-platform/ballerina-lang/issues/19711

cli/ballerina-tool/src/main/java/org/ballerinalang/tool/Main.java
```diff
@@ -54,7 +53,6 @@
 
     public static void main(String... args) {
         try {
-            ToolUtil.checkForUpdate(outStream, args);
```
Kishanthan: Can we create an issue to track this removal and where to put it instead? I mean, which command(s) to put this check for upcoming release ?

keizer619: https://github.com/ballerina-platform/ballerina-lang/issues/20021

PR #20018, opened 2019-11-21
Java interop with native string support

manuranga:


compiler/ballerina-backend-jvm/src/main/ballerina/src/compiler_backend_jvm/jvm_cast_gen.bal
```diff
@@ -33,7 +33,10 @@ function generatePlatformCheckCast(jvm:MethodVisitor mv, BalToJVMIndexMap indexM
 }
 
 function generateBToJCheckCast(jvm:MethodVisitor mv, bir:BType sourceType, jvm:JType targetType) {
-    if (targetType is jvm:JByte) {
+    if (targetType is jvm:JRefType && targetType.typeValue == "org/ballerinalang/jvm/values/StringValue") {
```
lankavitharana: BTW shall we move this inside "generateCheckCastBToJRef" method instead of here? that would be more clear IMO

PR #20031, opened 2019-11-22
Add int|bool union support in llvm backend

### LINKED ISSUE #20100: [LLVM] int|bool union support in llvm backend
Create an int|bool union in llvm backend and assign values.
Pass the union to a function where it takes an int|bool union.

dulvinw:
$title. 

In this we have created memory allocations for bool|int union as a tagged struct with the largest member type of the union.
We have used bitcasts and the int64 type tag value to identify the underlying type of union at runtime.

fixes #20100


compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_gen.bal
```diff
@@ -105,13 +111,29 @@ function mapFuncsToNameAndGenrator(llvm:LLVMModuleRef mod, llvm:LLVMBuilderRef b
...
+            }
             FuncGenrator funcGen = new(func, mod, builder);
             genrators[func.name.value] = funcGen;
         }
     }
     return genrators;
 }
 
+function checkForValidFunction(bir:Function func) returns boolean {
+    if (stringutils:contains(func.name.value, "init")) {
```
vinok88: Shouldn't these be exact matches or atleast match with "__init"?

dulvinw: Yes I think they should Me. I'll change these values

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_gen.bal
```diff
@@ -169,3 +203,133 @@ function appendAllTo(any[] toArr, any[] fromArr) {
...
+    }
+    return <llvm:LLVMTypeRef>structMap[typeName];
+}
+
+function createUnion(bir:BUnionType bType) returns llvm:LLVMTypeRef {
+    string largestType = "null";
+    foreach var member in bType.members {
+        if (member is bir:BTypeInt) {
+            largestType = getLargerType(largestType, "int");
+            createTaggedInt();
```
lafernando: This seemingly looks like a random place to create the tagged struct type. Shall we rather move the "createTaggedInt" and "createTaggedBool" calls inside to "getTaggedStructType". So what it does is, it lazily create the types when someone looks up the tagged struct types using "getTaggedStructType". The idea is, now the creation of these struct types are bound to a certain flow in the code. This may not be apparent when this codebase grows, and we actually want to look up the tagged struct types without the "createUnion" call actually not being done. This may not happen practically, but we have to leave room for it to be possible. 

So basically, in any place we do "<llvm:LLVMTypeRef>structMap[largestType]", we should do "getTaggedStructType".

dulvinw: I got the gist of it. I'll add the changes necessary.

dulvinw: Done

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_bb_body_gen.bal
```diff
@@ -79,14 +93,41 @@ type BbBodyGenrator object {
...
 
     function genConstantLoadIns(bir:ConstantLoad constLoad) {
-        if !(constLoad.value is int) {
-            error err = error("NotIntErr", message = "invalid value in constLoad");
-            panic err;
+        if (constLoad.value is int) {
+            self.genIntConstLoadIns(constLoad);
+        } else {
+            self.genBoolConstLoadIns(constLoad);
         }
```
SupunS: It would be better to add a `else if` for `boolean` type followed by a `else` case which will handle all the unsupported types (may be panic with a proper error). This will help for debugging in the long run, as we add more types in future.

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_bb_body_gen.bal
```diff
@@ -79,14 +93,41 @@ type BbBodyGenrator object {
...
+    }
+
+    function genMoveIns(bir:Move moveIns) {
+        if (moveIns.lhsOp.typeValue is bir:BUnionType) {
+            self.castLhsOpForMoveIns(moveIns);
+        }
+        llvm:LLVMValueRef lhsRef = self.parent.getLocalVarRef(moveIns.lhsOp);
+        var rhsVarOp = moveIns.rhsOp;
+        llvm:LLVMValueRef rhsVarOpRef = self.parent.genLoadLocalToTempVar(rhsVarOp);
+        var loaded = <llvm:LLVMValueRef> llvm:llvmBuildStore(self.builder, rhsVarOpRef, lhsRef);
```
SupunS: Do we need to cast to `<llvm:LLVMValueRef>`?

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_bb_body_gen.bal
```diff
@@ -79,14 +93,41 @@ type BbBodyGenrator object {
...
+    function genIntConstLoadIns(bir:ConstantLoad constLoad) {
+        llvm:LLVMValueRef lhsRef = self.parent.getLocalVarRef(constLoad.lhsOp);
         var constRef = llvm:llvmConstInt(llvm:llvmInt64Type(), <int>constLoad.value, 0);
         var loaded = llvm:llvmBuildStore(self.builder, constRef, <llvm:LLVMValueRef>lhsRef);
     }
+
+    function genBoolConstLoadIns(bir:ConstantLoad constLoad) {
+        llvm:LLVMValueRef lhsRef = self.parent.getLocalVarRef(constLoad.lhsOp);
+        var constRef = llvm:llvmConstInt(llvm:llvmInt1Type(), <boolean>constLoad.value?1:0, 0);
+        var loaded = llvm:llvmBuildStore(self.builder, constRef, <llvm:LLVMValueRef>lhsRef);
```
SupunS: Unused var `loaded`. Lets fix in all places.

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_bb_body_gen.bal
```diff
@@ -37,19 +37,33 @@ type BbBodyGenrator object {
...
-        var loaded = <llvm:LLVMValueRef> llvm:llvmBuildStore(self.builder, rhsVarOpRef, lhsRef);
+    function genTypeTest(bir:TypeTest typeTestIns) {
+        bir:BType rhsOpType = typeTestIns.rhsOp.typeValue;
+        boolean isLhsUnionType = isUnionType(rhsOpType);
+        if (isLhsUnionType) {
+            llvm:LLVMValueRef unionTagValue =
+                        self.parent.loadElementFromStruct(typeTestIns.rhsOp, TAGGED_UNION_FLAG_INDEX);
+            llvm:LLVMValueRef valueOfType = getValueRefFromInt(getTag(typeTestIns.typeValue), 0);
+            llvm:LLVMValueRef isEq = llvm:llvmBuildICmp(self.builder, llvm:LLVMIntEQ, unionTagValue, valueOfType, "typeTest");
+            _ = llvm:llvmBuildStore(self.builder, isEq, self.parent.getLocalVarRef(typeTestIns.lhsOp));
```
SupunS: Looks like the return value of the `llvmBuildStore(...)` method is never used at any place. So Im wondering whether this API actually needs a return?

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_func_gen.bal
```diff
@@ -15,14 +15,14 @@
...
 
 import ballerina/llvm;
-//import ballerina/io;
 import ballerina/bir;
 
 type FuncGenrator object {
     bir:Function func;
     llvm:LLVMValueRef? funcRef = ();
     llvm:LLVMModuleRef mod;
     map<llvm:LLVMValueRef>? localVarRefs = {};
```
SupunS: Not related to this PR, but why does this `localVarRefs` needs to be nil-able?

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_bb_body_gen.bal
dulvinw: +1. Will add necessary changes

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_func_gen.bal
dulvinw: I was under the assumption that there could be chances where local variables are non-existing in a function. Is that invalid?

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_bb_body_gen.bal
dulvinw: This api binds bytedecos LLVMBuildStore function. It has a return value hence we have a return value for the extern function as well. WDYT the best way?

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_func_gen.bal
SupunS: an empty map would represent that right?

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_bb_body_gen.bal
SupunS: since only the compiler is using this we can decide on the ballerina API, and it doesn't have to be exactly same as the native API. If we know for sure that the return value is/will **never** used, then its clean to have a API without returns.

compiler/ballerina-backend-llvm/src/main/ballerina/src/compiler_backend_llvm/llvm_func_gen.bal
dulvinw: Yes! I will update.

Issue #19299, opened 2019-10-03
No type checking for check-ed error type vs enclosing invokable's error type


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
Compilation should fail for
```ballerina
const R1 = "first reason";
const R2 = "second reason";

type E1 error<R1>;
type E2 error<R2>;

function foo() returns E1? {
    int i = check bar(); // should fail - check returns E2, but return type is E1?
}

function bar() returns int|E2 {
    return E2();
}

public function main() {
    E1? x = foo();
}
```

Currently fails at runtime:
```cmd
$ ballerina run xyz.bal 
Compiling source
        xyz.bal

Generating executables
Running executables

error: second reason 
        at xyz:bar(xyz.bal:12)
           xyz:foo(xyz.bal:8)
           xyz:main(xyz.bal:16)
```

**Affected Versions:**
```cmd
$ ballerina -v
Ballerina 1.0.1
Language specification 2019R3
```

hasithaa:
Resolved with #19374

MaryamZi:
Re-opening since the fix is being modified with https://github.com/ballerina-platform/ballerina-lang/pull/19529 to make this a warning instead of an error, for the 1.0.2 patch release.

Keeping this issue open to track making it an error.

PR #20110, opened 2019-11-29
Make checked vs enclosing error mismatch an error

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/19299

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -815,8 +815,8 @@ error.checked.expr.invalid.usage.no.error.type.rhs=\
 error.checked.expr.invalid.usage.only.error.types.rhs=\
   invalid usage of the ''{0}'' expression operator: all expression types are equivalent to error type
 
-error.checked.expr.no.error.return.in.encl.invokable=\
-  invalid usage of the ''check'' expression operator: no error type return in enclosing invokable
+error.checked.expr.no.matching.error.return.in.encl.invokable=\
+  invalid usage of the ''check'' expression operator: no matching error return type(s) in the enclosing invokable
```
hasithaa: should we say function here instead of `invokable`? since we only have functions now.

MaryamZi: This is the same error used with worker returns too though. Is it OK to say function there also?

```ballerina
public function main() {
    worker w1 {
        check foo();
    }
}

    return error("err");
}
```

Error with the current changes:
```cmd
$ ballerina run qwe.bal 
Compiling source
        qwe.bal
error: .::qwe.bal:3:9: invalid usage of the 'check' expression operator: no matching error return type(s) in the enclosing invokable
```

hasithaa: Ok, Got it. Let's address this issue separately then.

PR #20180, opened 2019-12-03
Improve type narrowing of var type in foreach

### LINKED ISSUE #19041: The type of the record foreach typed binding pattern seems to be fixed as `any`
**Description:**
$title, where it ideally should be the union of all possible types.

**Steps to reproduce:**
Case I - doesn't mandate error when the record could have error values.
```ballerina
import ballerina/io;

public type Foo record {|
    float f1;
    anydata|error...;    
|};

public function main() {
    Foo a12 = { f1: 1.0, "e1": error("err reason") };
    foreach any item in a12 {
        io:println(item); // works for the error value too
    }
}
```

Case II - requires any even when all anydata:
```ballerina
import ballerina/io;

public type Foo record {
    float f1;
};

public function main() {
    Foo a12 = { f1: 1.0 };
    foreach anydata item in a12 { // incompatible types: expected 'any', found 'anydata'
        io:println(item);
    }
}
```

**Affected Versions:**
```cmd
$ ballerina -v
Ballerina 1.0.0
Language specification 2019R3
```

gimantha:
Fixes #19041

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -1170,7 +1170,7 @@ public class Types {
         List<BField> fields = recordType.fields;
         BUnionType unionType = BUnionType.create(null);
 
-        if (recordType.restFieldType.tag != TypeTags.NONE) {
+        if (!recordType.sealed) {
             unionType.add(recordType.restFieldType);
         }
 
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -21,7 +21,6 @@ import org.ballerinalang.model.Name;
 import org.ballerinalang.model.TreeBuilder;
 import org.ballerinalang.model.elements.Flag;
 import org.ballerinalang.model.types.TypeKind;
-import org.ballerinalang.model.types.UnionType;
 import org.ballerinalang.util.BLangCompilerConstants;
 import org.ballerinalang.util.diagnostic.DiagnosticCode;
 import org.wso2.ballerinalang.compiler.semantics.model.SymbolTable;
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -1175,17 +1175,19 @@ public class Types {
         }
 
         for (BField field : fields) {
-            if (!isAssignable(unionType, field.type) && !isAssignable(field.type, unionType)) {
-                unionType.add(field.type);
+            boolean fieldAssignableToUnionType = isAssignable(field.type, unionType);
+            boolean unionTypeAssignableField = isAssignable(unionType, field.type);
+
+            if (fieldAssignableToUnionType) {
                 continue;
             }
 
-            if (isAssignable(field.type, unionType)) {
-                continue;
-            }
-
-            if (isAssignable(unionType, field.type)) {
+            if (unionTypeAssignableField) {
                 unionType = BUnionType.create(null);
+                unionType.add(field.type);
+            }
+
+            if (!unionTypeAssignableField) {
                 unionType.add(field.type);
             }
         }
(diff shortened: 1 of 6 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1166,26 +1168,36 @@ private BAttachedFunction getNextFunc(BObjectType iteratorType) {
...
-        // If there are no fields in the record, return the rest field type as the inferred type.
-        if (fields.isEmpty()) {
-            return recordType.restFieldType;
+        if (recordType.restFieldType.tag != TypeTags.NONE) {
+            unionType.add(recordType.restFieldType);
         }
 
-        BType inferredType = fields.get(0).type; // If all the fields are the same, doesn't matter which one we pick
+        for (BField field : fields) {
+            if (!isAssignable(unionType, field.type) && !isAssignable(field.type, unionType)) {
```
MaryamZi: Should we extract out `isAssignable(unionType, field.type)` and `isAssignable(field.type, unionType)` to two `boolean` variables since we seem to be checking all of them twice, worst case?

MaryamZi: Should we also swap the checks?

L1183-L1185 first, L1187-L1190 second and if both fail we do this?

langlib/langlib-test/src/test/resources/test-src/statements/foreach/foreach-var-type-negative.bal
```diff
@@ -0,0 +1,62 @@
+public type Foo record {|
```
MaryamZi: Shall we also add a test for an open (inclusive) record, with the default `anydata` rest field type?

```ballerina
type Baz record {
    string s;
    int i;
};
```

gimantha: sure

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
gimantha: fixed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1166,26 +1168,38 @@ private BAttachedFunction getNextFunc(BObjectType iteratorType) {
 
     public BType inferRecordFieldType(BRecordType recordType) {
         List<BField> fields = recordType.fields;
+        BUnionType unionType = BUnionType.create(null);
 
-        // If there are no fields in the record, return the rest field type as the inferred type.
-        if (fields.isEmpty()) {
-            return recordType.restFieldType;
+        if (recordType.restFieldType.tag != TypeTags.NONE) {
```
MaryamZi: Shall we check
```suggestion
        if (!recordType.sealed) {
```
instead?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1166,26 +1168,38 @@ private BAttachedFunction getNextFunc(BObjectType iteratorType) {
...
-        BType inferredType = fields.get(0).type; // If all the fields are the same, doesn't matter which one we pick
+        for (BField field : fields) {
+            boolean fieldAssignableToUnionType = isAssignable(field.type, unionType);
+            boolean unionTypeAssignableField = isAssignable(unionType, field.type);
 
-        // If it's an open record, the rest field type should also be of the same type as the mandatory fields.
-        if (!recordType.sealed && recordType.restFieldType.tag != inferredType.tag) {
-            return symTable.anyType;
-        }
+            if (fieldAssignableToUnionType) {
```
MaryamZi: I guess we don't need a variable for `fieldAssignableToUnionType` since this is the only usage?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1166,26 +1168,38 @@ private BAttachedFunction getNextFunc(BObjectType iteratorType) {
...
-            if (inferredType.tag != fields.get(i).type.tag) {
-                return symTable.anyType;
+            if (unionTypeAssignableField) {
+                unionType = BUnionType.create(null);
+                unionType.add(field.type);
             }
+
+            if (!unionTypeAssignableField) {
+                unionType.add(field.type);
+            }
```
MaryamZi: Should we do 
```java
            if (unionTypeAssignableField) {
                unionType = BUnionType.create(null);                
            } 
            unionType.add(field.type);
```

Again, we won't need the variable since this is the only usage?

Issue #19623, opened 2019-10-28
Compiler Crash for Database Code Build


lafernando:
When compiling the below Ballerina code, the compiler crashes with a "bad, sad" error and the log files contains the following:

```
[2019-10-28 10:58:03,702] SEVERE {b7a.log.crash} - -1 
java.lang.ArrayIndexOutOfBoundsException: -1
        at org.objectweb.asm.Frame.getConcreteOutputType(Frame.java:1139)
        at org.objectweb.asm.Frame.merge(Frame.java:1184)
        at org.objectweb.asm.MethodWriter.computeAllFrames(MethodWriter.java:1607)
        at org.objectweb.asm.MethodWriter.visitMaxs(MethodWriter.java:1543)
        at org.ballerinalang.nativeimpl.jvm.methodvisitor.VisitMaxs.visitMaxs(VisitMaxs.java:49)
        at ballerina.jvm.$value$MethodVisitor.visitMaxs(types.bal:61)
        at ballerina.jvm.$value$MethodVisitor.call(types.bal)
        at ballerina.compiler_backend_jvm.jvm_method_gen.genJMethodForBFunc(jvm_method_gen.bal:325)
        at ballerina.compiler_backend_jvm.jvm_method_gen.generateMethod(jvm_method_gen.bal:42)
        at ballerina.compiler_backend_jvm.jvm_package_gen.generatePackage(jvm_package_gen.bal:215)
        at ballerina.compiler_backend_jvm.main.generateJarBinary(main.bal:72)
        at ballerina.compiler_backend_jvm.main.main(main.bal:47)
        at ballerina.compiler_backend_jvm.___init.$lambda$main$(compiler_backend_jvm)
        at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:493)
        at org.ballerinalang.jvm.scheduling.Scheduler.run(Scheduler.java:238)
        at org.ballerinalang.jvm.scheduling.Scheduler.runSafely(Scheduler.java:208)
        at java.lang.Thread.run(Thread.java:748)
 
[2019-10-28 10:58:03,705] SEVERE {b7a.log.crash} - -1 
java.lang.ArrayIndexOutOfBoundsException: -1
        at org.objectweb.asm.Frame.getConcreteOutputType(Frame.java:1139)
        at org.objectweb.asm.Frame.merge(Frame.java:1184)
[...]

PR #20195, opened 2019-12-04
Disallow check expr within trx block

rdhananjaya:
## Purpose
Similar to `return`, `check` expression is also disallowed within the transaction block as `check` expression has `return` like semantic when the RHS value is an error. 

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/19623

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/statements/transaction/transaction_block_negative.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/statements/transaction/transaction_block_negative.bal
@@ -11,32 +11,6 @@ function testCorrectTransactionBlock() returns (string) {
     return a;
 }
 
-function testTwoAbortedBlocks() returns (string) {
-    string a = "";
-    transaction with retries=2 {
-    } onretry {
-        a = a + " retry";
-    } aborted {
-        a = a + " committed";
-    } aborted {
-        a = a + " aborted";
-    }
-    return a;
-}
-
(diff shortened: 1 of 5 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/statements/transaction/transaction_block_negative.bal
```diff
@@ -11,32 +11,6 @@ function testCorrectTransactionBlock() returns (string) {
     return a;
 }
 
-function testTwoAbortedBlocks() returns (string) {
-    string a = "";
-    transaction with retries=2 {
-    } onretry {
-        a = a + " retry";
-    } aborted {
```
rdhananjaya: With current grammer impl, multiple aborted/committed blocks are prohibited from the grammer level.

hasithaa:
What is the behavior with checkpanic? I assume it is same as panic?

rdhananjaya:
> What is the behavior with checkpanic? I assume it is same as panic?

Yes the panic of checkpanic will be handle by the transaction block as a transaction attempt failure.

Issue #19761, opened 2019-11-04
Issue with error binding patterns with a rest binding pattern


MaryamZi:
**Description:**
$title. 

The type of the rest binding pattern seems to be fixed as `map<anydata>`, even though there could always be an error in the detail.

```ballerina
import ballerina/io;

type Detail record {
    string message?;
    error cause?;
    int? code;
};

function foo(int? i) returns error<string, Detail> {
    return error("err reason", message = "err message", code = i, f = 1.0);
}

public function main() {
    var error(r, message = message, code = code, ...rest) = foo(1);
    
    io:println(rest is map<anydata>); // compilation error - unnecessary condition: expression will always evaluate to 'true'
}
```
When the detail actually has an error, it results in the runtime crashing.

```ballerina
type Detail record {
    string message?;
    error cause?;
    int? code;
};

function foo(int? i) returns error<string, Detail> {
    return error("err reason", message = "err message", code = i, f = 1.0, cause = error("cause error"));
}

public function main() {
    var error(r, message = message, code = code, ...rest) = foo(1);
}
```

ballerina-internal.log
```
[2019-11-04 11:51:50,835] SEVERE {b7a.log.crash} - org.ballerinalang.jvm.values.ErrorValue cannot be cast to org.ballerinalang.jvm.values.MapValue 
java.lang.ClassCastException: org.ballerinalang.jvm.values.ErrorValue cannot be cast to org.ballerinalang.jvm.values.MapValue
	at test.main(test.bal:12)
	at ___init.$lambda$main$(.)
	at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:426)
	at org.ballerinalang.jvm.scheduling.Scheduler.run(Scheduler.java:218)
	at org.ballerinalang.jvm.scheduling.Scheduler.runSafely(Scheduler.java:191)
	at java.lang.Thread.run(Thread.java:748)
 
[2019-11-04 11:51:50,847] SEVERE {b7a.log.crash} - org.ballerinalang.jvm.values.ErrorValue cannot be cast to org.ballerinalang.jvm.values.MapValue 
java.lang.ClassCastException: org.ballerinalang.jvm.values.ErrorValue cannot be cast to org.ballerinalang.jvm.values.MapValue
	at test.main(test.bal:12)
	at ___init.$lambda$main$(.)
	at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:426)
[...]

PR #20208, opened 2019-12-04
Fix rest type of error def binding pattern

rdhananjaya:
## Purpose
Rest type of error variable definition (type binding pattern) is a map of T, where T is the union of types of fields in error detail value which are not extracted in the binding pattern.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/19761

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1373,8 +1375,20 @@ private boolean validateErrorVariable(BLangErrorVariable errorVariable, BErrorTy
...
         if (isRestDetailBindingAvailable(errorVariable)) {
+            // Type of rest pattern is a map type where constraint type is,
+            // union of keys whose values are not matched in error binding/match pattern.
             BTypeSymbol typeSymbol = createTypeSymbol(SymTag.TYPE);
-            BMapType restType = new BMapType(TypeTags.MAP, recordType.restFieldType, typeSymbol);
+            BUnionType restFieldType = BUnionType.create(null, recordType.restFieldType);
+            for (Map.Entry<String, BField> entry : fieldMap.entrySet()) {
+                if (!matchedDetailFields.contains(entry.getKey())) {
+                    BType type = entry.getValue().getType();
+                    if (!type.equals(symTable.nilType)) {
```
MaryamZi: Why are we not adding if the type is nil?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1373,8 +1375,20 @@ private boolean validateErrorVariable(BLangErrorVariable errorVariable, BErrorTy
...
+                        restFieldType.add(type);
+                    }
+                }
+            }
+
+            BMapType restType = new BMapType(TypeTags.MAP, restFieldType, typeSymbol);
```
MaryamZi: Shouldn't we first check if the union has more than one member? If it is just one type we don't need a union?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1373,8 +1375,20 @@ private boolean validateErrorVariable(BLangErrorVariable errorVariable, BErrorTy
...
+            // Type of rest pattern is a map type where constraint type is,
+            // union of keys whose values are not matched in error binding/match pattern.
             BTypeSymbol typeSymbol = createTypeSymbol(SymTag.TYPE);
-            BMapType restType = new BMapType(TypeTags.MAP, recordType.restFieldType, typeSymbol);
+            BUnionType restFieldType = BUnionType.create(null, recordType.restFieldType);
+            for (Map.Entry<String, BField> entry : fieldMap.entrySet()) {
+                if (!matchedDetailFields.contains(entry.getKey())) {
+                    BType type = entry.getValue().getType();
+                    if (!type.equals(symTable.nilType)) {
+                        restFieldType.add(type);
```
MaryamZi: Shouldn't we check if it is assignable to the current collection of types and add only if not? 

@gimantha is doing something similar in https://github.com/ballerina-platform/ballerina-lang/pull/20180/files#diff-782094a79e6831ac57a34ad157af49c1R1177 to identify the narrowest type representing all possible fields.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/variabledef/ErrorVariableDefinitionTest.java
```diff
@@ -238,7 +238,7 @@ public void testNegativeErrorVariables() {
         BAssertUtil.validateError(resultNegative, ++i,
                 "redeclared symbol 'message'", 54, 26);
         BAssertUtil.validateError(resultNegative, ++i,
-                "incompatible types: expected 'int', found 'map<(anydata|error)>'", 56, 18);
+                "incompatible types: expected 'int', found 'map<(anydata|error|string)>'", 56, 18);
```
MaryamZi: Identifying the narrowest type will remove the requirement to add `string` here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
rdhananjaya: Yep we don't need this, will remove

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
rdhananjaya: It feels like a premature optimization, do you see a downside of single-member union here?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/variabledef/ErrorVariableDefinitionTest.java
rdhananjaya: Fixed above

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1373,8 +1375,20 @@ private boolean validateErrorVariable(BLangErrorVariable errorVariable, BErrorTy
         }
 
         if (isRestDetailBindingAvailable(errorVariable)) {
+            // Type of rest pattern is a map type where constraint type is,
+            // union of keys whose values are not matched in error binding/match pattern.
             BTypeSymbol typeSymbol = createTypeSymbol(SymTag.TYPE);
-            BMapType restType = new BMapType(TypeTags.MAP, recordType.restFieldType, typeSymbol);
+            BUnionType restFieldType = BUnionType.create(null, recordType.restFieldType);
```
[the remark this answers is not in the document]
rdhananjaya: Thank you, I completely forgot about sealed detail records.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
rdhananjaya: Fixed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: IINM, we've done this across such similar union type creations in the compiler.

IMO it would be cleaner to have a union only when there is > 1 type, this could potentially be helpful in subsequent type-checking/analysis when accessing elements from the map too (same type checks would pass and we wouldn't have to work with union types)?

rdhananjaya: I didn't know that it's a common practice to lift out single union members. Will change as suggested.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -221,4 +221,11 @@ public boolean isPureType() {
                         ((BUnionType) type).memberTypes.stream() : Stream.of(type))
                 .collect(Collectors.toCollection(LinkedHashSet::new));
     }
+
+    public BType normalize() {
```
MaryamZi: Should we add this method here? By definition, a union type has > 1 member types, IMO this isn't part of the union type?

rdhananjaya: If we are to go by that definition we need to create a union builder. Since it's a too much of a change, and since we have to normalize most of the union instances I think it's a responsibility of the union class.

MaryamZi: IMO, we shouldn't even be creating a union type if we have only one type. What we usually do when creating a union type IINM is creating a set and then adding members to it, then checking the number of members and creating a union type if > 1, else using the only member.

We create a union type in this scenario and in #20180 just because it makes it easier to check assignability right? IMO, we don't need a union builder, but can just do this in line since it won't be that common?

rdhananjaya: The builder is not to support this use case but to support the common case. What build would do it what you said in the first paragraph, but we don't have to manually create sets and stuff, and it will also stop people like me creating probable single-member unions.

I originally created the union not to check assignability, but because I wasn't aware of the optimization we seemingly suppose to do for unions. 

OK will inline this.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1373,17 +1376,42 @@ private boolean validateErrorVariable(BLangErrorVariable errorVariable, BErrorTy
...
+    private BType getRestMapConstraintType(Map<String, BField> errorDetailFields, Set<String> matchedDetailFields,
+                                           BRecordType recordType) {
+        BUnionType restUnionType = BUnionType.create(null);
+        if (!recordType.sealed) {
+            restUnionType.add(recordType.restFieldType);
+        }
+        for (Map.Entry<String, BField> entry : errorDetailFields.entrySet()) {
+            if (!matchedDetailFields.contains(entry.getKey())) {
+                BType type = entry.getValue().getType();
+                if (!types.isAssignable(type, restUnionType)) {
```
MaryamZi: Shouldn't we check this both ways to extract out the narrowest type like in https://github.com/ballerina-platform/ballerina-lang/pull/20180/files#diff-782094a79e6831ac57a34ad157af49c1R1178-R1186?

For example if the rest type is a finite type Foo

```ballerina
type Foo "foo"|"bar";
```

And I have a field of type `string`, the union type will currently be `string|Foo` right? Where just `string` will be enough?

I guess we can do this as an improvement.

MaryamZi: Shall we create an issue to track this?

rdhananjaya: Issue to track this https://github.com/ballerina-platform/ballerina-lang/issues/20252

PR #20318, opened 2019-12-10
Fix grpc bidirectional streaming BBE

Bhashinee:
## Purpose
$subject.


examples/grpc-bidirectional-streaming/grpc_bidirectional_streaming_service.bal
```diff
@@ -14,7 +14,7 @@ service Chat on new grpc:Listener(9090) {
     //This `resource` is triggered when a new caller connection is initialized.
     resource function onOpen(grpc:Caller caller) {
         log:printInfo(string `${caller.getId()} connected to chat`);
-        consMap[caller.getId().toString()] = caller;
+        consMap[caller.getId().toString()] = <@untainted> caller;
```
wggihan: @daneshk @hasithaa is this behaviour correct? Do we need to untained if we didn't mention anything inthe signature?

Issue #20130, opened 2019-12-02
Make string interop available for stdlib


manuranga:

SupunS:
Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/20364

PR #20364, opened 2019-12-12
bootstrap on 1.1.0-alpha

manuranga:


manuranga:
this is to fix #20130

composer/package-lock.json
```diff
@@ -4721,7 +4721,7 @@
     },
     "@types/webpack-env": {
```
warunalakshitha: I think we do not need this file

manuranga: thanks, fixed

PR #20369, opened 2019-12-12
Add a delay to cache cleanup task

### LINKED ISSUE #20366: Cache cleanup task doesn't have an initialDelayInMillis
**Description:**
$title. Therefore, each time cache gets initialized, a cleanup task also gets triggered. That causes concurrent modification issues on the cache functions (put, get, hasKey).

**Steps to reproduce:**
https://ballerina.io/learn/by-example/cache.html will return; 
```
error: {ballerina}TypeCastError message=incompatible types: '()' cannot be cast to 'string'
        at grainierx:main(main.bal:14)
```
**Affected Versions:**
1.1.0-alpha

n/a

grainier:
## Purpose
$title & add locks to where necessary to prevent concurrent modification issues.

Fixes #20366

## Approach
- add a delay to cache cleanup task
- add locks to where necessary to prevent concurrent modification issues.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/cache/src/main/ballerina/src/cache/cache.bal
```diff
@@ -151,15 +151,17 @@ public type Cache object {
...
-        int maxCapacity = self.capacity;
-        float ef = self.evictionFactor;
-        int numberOfKeysToEvict = <int>(maxCapacity * ef);
-        // Get the above number of least recently used cache entry keys from the cache
-        string[] cacheKeys = self.getLRUCacheKeys(numberOfKeysToEvict);
-        // Iterate through the map and remove entries.
-        foreach var c in cacheKeys {
-            // These cache values are ignored. So it is not needed to check the return value for the remove function.
-            var tempVar = self.entries.remove(c);
+        lock {
```
wggihan: Can we add new entries to cache while evict function execute or lock prevent the access to the map?

grainier: It prevents access to the map.

wggihan: This will drastically affect the performance of caching. Can't we only delay the start time and refain from using locking?

grainier: We can, but then again, earlier the map access was anyway concurrent safe. Now that it's not, it's advised to handle it by the user.

grainier: Anyway, revert locks on 2a25a0f.

wggihan: We can alter to use this cache implementation without relying on tasks. A community user already did this and we might able to add that change. So for the moment let remove the locking

Issue #20309, opened 2019-12-10
No way to link external jar files to single bal files


anupama-pathirage:
**Description:**
$subject.

So we cannot use ballerinax/java.jdbc module in a single bal files. Earlier copying to bre/lib works and not working now. 

**Steps to reproduce:**

**Affected Versions:**

1.0.5, 1.1.0-alpha

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

hevayo:
Duplicate of #20281

Issue #20362, opened 2019-12-12
Encrypted configs not working


kaneeldias:
**Description:**
Secret key prompt is passed over when using an encrypted value in ballerina.conf. Doesn't read from secret.txt either (both ballerina build and ballerina run).

**Steps to reproduce:**
"Configure Secrets as Configuration Items" in https://ballerina.io/learn/how-to-deploy-and-run-ballerina-programs/

**Affected Versions:**
1.1.0-rc2

pubudu91:
Tested this and this seems to happen only when trying to run a `.bal` file or a module. Running the JAR file using `ballerina run program.jar` or `java -jar program.jar` is working.

pubudu91:
Closing this since this issue was fixed in #20371

pubudu91:
Reopening this since this problem is still there for the default config file (ballerina.conf).

warunalakshitha:
This happen when you have ballerina.conf file in same directory path. Workaround is to pass the config file with --b7a.config.file=<path> without keeping ballerina.conf file in same root path when you have encrypted values.

BTw original issue needs to be fix. It requires an api change.

warunalakshitha:
This issue is fixed with latest refactor with creating jars with same compiler process. Verified in ballerina 1.2.0. Hence closing issue.

PR #20371, opened 2019-12-12
Fix No way to link external jar files to single bal files

warunalakshitha:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/20309


pubudu91:
Fixes #20362 as well

PR #20398, opened 2019-12-15
Migrate serialize and deserialize methods to interop

gimantha:
## Purpose
$subject


hasithaa:
I think we don't need an interop at all. Shall we just deprecate this API.?

PR #20408, opened 2019-12-16
Deprecate deserialize and serialize methods

gimantha:
## Purpose
$Subject


stdlib/streams/src/main/java/org/ballerinalang/nativeimpl/Deserialize.java
```diff
@@ -21,25 +21,15 @@
...
-        orgName = "ballerina", packageName = "streams",
-        functionName = "deserialize",
-        args = {@Argument(name = "serializedString", type = TypeKind.STRING)},
-        returnType = {@ReturnType(type = TypeKind.MAP)},
-        isPublic = true
-)
 public class Deserialize {
 
+    @Deprecated
     public static MapValue deserialize(Strand strand, Object object, String str) {
```
hasithaa: If we replace Strand with the Object, class we should be able to remove the  JBallerina runtime dependency as well.

stdlib/streams/src/main/ballerina/src/streams/util/snapshot.bal
```diff
@@ -47,14 +47,22 @@ boolean stateLoaded = false;
 map<boolean> loadedStates = {};
 
 # Native function to deserialize a serialized snapshot.
+# This function is deprecated since v1.1.
```
irshadnilam: From my understanding the word `deprecate` in software world means that the function is no longer recommended (there are replaced with newer/better) functions and those will be removed in future. 

It seems like we are panicking an error with `not supported` message. 
Should we maybe completely remove this method without deprecating it? The word deprecate gives a false idea. 

WDYT?

gimantha: Thats a valid point, but these methods were publicly available in 1.0. We cannot simply remove them. If they are going to use these function, we have to anyway convey the user that functionality is not supported, hence the panic. :)

irshadnilam: In that case, shouldn't we keep the functions as it is in working state (marked as deprecated) and remove it in a major version?

gimantha: from stdlib point, java methods are not considered as API and as per further discussion, we decided to remove ballerina APIs and bump streams module version

Issue #20020, opened 2019-11-22
Revisit final jar creation task logic


Kishanthan:
**Description:**
There is a considerable time increase with final jar creation task in 1.0.4 release. This is possibly due to the fix introduced with #19783

**Steps to reproduce:**
Compare the build times between 1.0.3 and 1.0.4

**Affected Versions:**
1.0.4

**OS, DB, other environment details and versions:**
Mac, Windows

**Suggested Assignees (optional):**
@warunalakshitha

NipunaRanasinghe:
Closing since this is fixed with #20797 and #20862.

PR #20438, opened 2019-12-17
Improve uber jar creation speed

manuranga:
Related to https://github.com/ballerina-platform/ballerina-lang/issues/20020


Kishanthan:
Need to revisit this approach as it increases the jar size. Hence closing it.

PR #20495, opened 2019-12-24
fixed table access bug

rahul3:
Fixed a bug in the example code which gave an error as follows:

error: .::table.bal:68:30: invalid operation: type 'record {| anydata...; |}' does not support field access for non-required field 'name'

See https://stackoverflow.com/questions/59472416/issue-accessing-the-table-data-structure-in-ballerina-error-invalid-operation/59472417#59472417

Fixes #<Issue Number>

## Check List 
- [ x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [x ] Ballerina By Examples

### DIFF
--- a/examples/table/table.bal
+++ b/examples/table/table.bal
@@ -65,7 +65,7 @@ public function main() {
     io:println("Using while loop: ");
     while (tb.hasNext()) {
         var ret = tb.getNext();
-        io:println("Name: ", ret.name);
+        io:println("Name: ", ret.get("name"));
     }
 
     // Convert the `table` to JSON format.


MaryamZi:
Hi @rahul3, thank you very much for your contribution. 

As mentioned in https://stackoverflow.com/a/59475324 this has been fixed for 1.1.0 though. So I guess we could continue to use field access, WDYT?

rahul3:
Hello Maryam,

my opinion on this is that the change in this pull request will make the code compatible for earlier versions as well. It is simple enough to merge in my opinion unless you wanted to highlight the functionality of the original syntax in the tutorial.

Thanks,
Rahul Padmanabhan

MaryamZi:
Hi @rahul3, once the bug with the site is fixed, a user would be able to easily access a compatible version of the example.

As for accessing the field though, IMO we should suggest using field access where the field is a required field since we could retrieve the field as a value of the expected type.

```ballerina
type Employee record {|
    string name;
    int age;
|};

public function main() {
    Employee f = {name: "Josh", age: 20};

    string name1 = f.name; // valid

    string name2 = f.get("name"); // error: incompatible types: expected 'string', found '(string|int)'
}
```

rahul3:
Hello,

I agree with your proposal. Feel free to decline the pull request or let me know whether I should cancel it.

Thanks,
Rahul

MaryamZi:
Closing this PR as per https://github.com/ballerina-platform/ballerina-lang/pull/20495#issuecomment-571010617.

@rahul3, thank you once again for your efforts to identify and fix the issue.

PR #20531, opened 2020-01-06
Support using type `byte` as `json`

### LINKED ISSUE #20189: Can't use a `byte` as `json`
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    byte b = 1;
    // json j = b; // fails with 'incompatible types: expected 'json', found 'byte''

    any a = b;
    io:println(a is json); // prints false

    int i = <int> b;
    a = i;
    io:println(a is json); // prints true
}
```

**Affected Versions:**
1.0.4

irshadnilam:
## Purpose
$title

Fixes #20189

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/http/src/main/ballerina/src/http/http_commons.bal
+++ b/stdlib/http/src/main/ballerina/src/http/http_commons.bal
@@ -220,10 +220,10 @@ function buildRequest(RequestMessage message) returns Request {
         request.setTextPayload(message);
     } else if (message is xml) {
         request.setXmlPayload(message);
-    } else if (message is json) {
-        request.setJsonPayload(message);
     } else if (message is byte[]) {
         request.setBinaryPayload(message);
+    } else if (message is json) {
+        request.setJsonPayload(message);
     } else if (message is io:ReadableByteChannel) {
         request.setByteChannel(message);
     } else {
@@ -242,10 +242,10 @@ function buildResponse(ResponseMessage message) returns Response {
         response.setTextPayload(message);
     } else if (message is xml) {
         response.setXmlPayload(message);
-    } else if (message is json) {
-        response.setJsonPayload(message);
     } else if (message is byte[]) {
         response.setBinaryPayload(message);
+    } else if (message is json) {
+        response.setJsonPayload(message);
     } else if (message is io:ReadableByteChannel) {
         response.setByteChannel(message);
     } else {
--- a/stdlib/messaging/nats/src/main/ballerina/src/nats/commons.bal
+++ b/stdlib/messaging/nats/src/main/ballerina/src/nats/commons.bal
@@ -20,7 +20,9 @@ public type Content byte[] | boolean | string | int | float | decimal | xml | js
 
 function convertData(Content data) returns string | byte[] | error {
     string | byte[] | error converted;
-    if (data is boolean) {
+    if (data is byte[]) {
+        converted = data;
+    } else if (data is boolean) {
         converted = data.toString();
     } else if (data is string) {
         converted = data.toString();
@@ -34,15 +36,13 @@ function convertData(Content data) returns string | byte[] | error {
         converted = data.toString();
     } else if (data is json) {
         converted = data.toJsonString();
-    } else if (data is record{}) {
+    } else {
         json | error jsonConverted = typedesc<json>.constructFrom(data);
         if (jsonConverted is json) {
             converted = jsonConverted.toString();
         } else {
             converted = jsonConverted;
         }
-    } else {
-        converted = data;
     }
     return converted;
 }
--- a/tests/jballerina-unit-test/src/test/resources/test-src/types/byte/byte_as_json_test.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/types/byte/byte_as_json_test.bal
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+// Copyright (c) 2020 WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
 //
 // WSO2 Inc. licenses this file to you under the Apache License,
 // Version 2.0 (the "License"); you may not use this file except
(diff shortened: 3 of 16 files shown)


stdlib/http/src/main/ballerina/src/http/http_commons.bal
```diff
@@ -242,10 +242,10 @@ function buildResponse(ResponseMessage message) returns Response {
         response.setTextPayload(message);
     } else if (message is xml) {
         response.setXmlPayload(message);
-    } else if (message is json) {
-        response.setJsonPayload(message);
     } else if (message is byte[]) {
         response.setBinaryPayload(message);
+    } else if (message is json) {
+        response.setJsonPayload(message);
```
KavinduZoysa: What is the reason to interchange this `else if` condition?

irshadnilam: Take the following code snippet

```ballerina
type Person record {
    
};

public function main() {
   int[] | json | Person value = "";
   if (value is json) {
       // do something
   } else if( value is int[]) { // <--- error on this line
       //do something
   }
}
```
The above code snippet prints out the following error. 
`Person' will not be matched to 'int[]'`

This is because the `is json` check also captures `int[]`, since `int` can be used as `json`. 
And when we are in the `else` clause the type is narrowed to `Person`. 

The reason for interchanging `else if` above is because of this but for type `byte[]` 

ps: please correct me if my understanding is wrong here @MaryamZi  @pubudu91 @rdhananjaya  :)

MaryamZi: Yeah, basically a `byte` is `json`, and therefore a `byte[]` is also `json`.

So, when using the type guard, if the `json` type test comes before the `byte[]` type test, the `byte[]` type test will never be evaluated to true since any `byte[]` is a `json`.

irshadnilam:
> Can we add more tests including (up/down-)casting, arrays, etc.?
> 
> We can add some tests similar to https://github.com/ballerina-platform/ballerina-lang/pull/20190/files#diff-25ed9dc50c09611a897394a34a13111bR1

Added with https://github.com/ballerina-platform/ballerina-lang/pull/20531/commits/b54bce93d3e6eaebe9f988dd17dad3d3bfc59047

pubudu91:
@irshadnilam This is an addition right? If this doesn't break anything, shouldn't this be sent to `ballerina-1.1.x` as well?

stdlib/messaging/nats/src/main/ballerina/src/nats/commons.bal
```diff
@@ -20,7 +20,9 @@ public type Content byte[] | boolean | string | int | float | decimal | xml | js
 
 function convertData(Content data) returns string | byte[] | error {
     string | byte[] | error converted;
-    if (data is boolean) {
+    if (data is byte[]) {
```
MaryamZi: Can we check with the std lib team if this order change is OK? Since this used to be the last case, just to make sure we aren't making the least likely usage the first check.

tests/jballerina-unit-test/src/test/resources/test-src/types/jsontype/json-value.bal
```diff
@@ -13,6 +13,12 @@ function testFloatAsJsonVal () returns (json) {
     return j;
 }
 
+function testByteAsJsonVal() returns (json) {
```
MaryamZi:
```suggestion
function testByteAsJsonVal() returns json {
```

tests/jballerina-unit-test/src/test/resources/test-src/types/byte/byte_as_json_test.bal
```diff
@@ -0,0 +1,160 @@
...
+                }
+            }
+            _ => {
+                if val != arr[index] {
+                    return false;
+                }
+            }
+		}
+	}
+	return true;
```
MaryamZi: Shall we reformat this bit? Noticed it's off to the int tests too, can you please reformat that too?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/conversion/NativeConversionNegativeTest.java
```diff
@@ -87,14 +87,6 @@ public void testTupleConversionFail() {
                 + "T2]'");
     }
 
-    @Test(description = "Test converting an unsupported array to json")
```
MaryamZi: Instead of removing this and the tests in https://github.com/ballerina-platform/ballerina-lang/pull/20531/files#diff-17ddc3514c16ada932b765cb849594b1L228, shall we change the tests to use some other non-JSON compatible field?

stdlib/messaging/nats/src/main/ballerina/src/nats/commons.bal
MaryamZi: @daneshk @aashikam FYA please.

daneshk: Order doesn't matter. should be fine

tests/jballerina-unit-test/src/test/resources/test-src/types/jsontype/json-value.bal
irshadnilam: Fixed

tests/jballerina-unit-test/src/test/resources/test-src/types/byte/byte_as_json_test.bal
irshadnilam: Fixed

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/conversion/NativeConversionNegativeTest.java
irshadnilam: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/20531/commits/091f20ee401e21f690630dd8cf20c16f1e9a1b11

irshadnilam:
> @irshadnilam This is an addition right? If this doesn't break anything, shouldn't this be sent to `ballerina-1.1.x` as well?

There are scenarios that can break. I have fixed such cases on stdlib implementation. 
ex: https://github.com/ballerina-platform/ballerina-lang/pull/20531/files#diff-4095ba2af5338e77ae8471ef868b9e21L223-L229

So we cannot sent this to 1.1.x branch.

PR #20548, opened 2020-01-08
Print errors after failer in travis

manuranga:
This pr adds description after a testng failure, so you don't have to search for "FAILED" text.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,11 @@ git:
 services:
   - docker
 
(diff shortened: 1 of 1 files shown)


manuranga:
This pr adds description after a testng failure, so you don't have to search for "FAILED" text.

Kishanthan:
This change will not work on windows as the change is specific with unix cmds.

PR #20646, opened 2020-01-16
Consider the constraint for table type checking

### LINKED ISSUE #17808: JDBC module return a java class cast error
**Description:**
Following code returns following error,
```
error: error org.ballerinax.jdbc.table.BCursorTable cannot be cast to org.ballerinalang.jvm.values.ErrorValue
```

**Steps to reproduce:**
```
@http:ResourceConfig {
        path: "/borrow/{book}"
    }
    resource function borrowBook(http:Caller caller, http:Request request, string book) {
        string selectBook = "SELECT count FROM library where name =?";
        string updateCount = "UPDATE library SET count = ? WHERE name=?";
        var selectBookRes = libryDB->select(selectBook, Count, book);
        boolean available = false;
        int count = 0;
        if (selectBookRes is table<Book>) {
            Book b = <Book>selectBookRes.getNext();
            count = <int>b.count;
            io:println(count);
            if (count > 0) {
               var updateRes = libryDB->update(updateCount, count, book);
               if (updateRes is jdbc:UpdateResult) {
                   var res = caller->respond("Happy reading, " + <@untiant>book + "!");
               } else {
                   io:println("error 1");
                   error e = updateRes;
                   var res = caller->respond("Error borrowing book, " + <string>e.detail()["message"]);
               }
            } else {
               var res = caller->respond("Sorry, " + <@untiant>book + ", not available."); 
            }
        } else {
            io:println();
            error e = selectBookRes;
            var res = caller->respond("Error fetching book info" + <@untiant><string>e.detail()["message"]);
        }
        
    }
```
**Affected Versions:**
1.00-pre-beta
Windows, H2_DB

MaryamZi:
## Purpose
$title.

Fixes #17808

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/jdbc/src/test/resources/test-src/actions/select_test.bal
```diff
@@ -272,7 +272,7 @@ function testCloseConnectionPool(string jdbcURL) returns @tainted (int) {
     return count;
 }
 
-function getTableCountValColumn(table<ResultCount> | error result) returns int {
+function getTableCountValColumn(table<record {}> | error result) returns int {
```
irshadnilam: Why do we need these changes? (and other similar)

MaryamZi: Earlier, with table type-checking the constraint was not considered. Thus, it was possible to assign a value of type `table<record {}>` (or some table broader than `table<Foo>`) to a variable of `table<Foo>` (as mentioned in https://github.com/ballerina-platform/ballerina-lang/issues/17808). 

This PR fixes the type-checking bug, which requires this change to accept a result of a db client remote method.

PR #20714, opened 2020-01-23
Fix the spec deviation in Unicode

### LINKED ISSUE #13180: [Spec Deviation] Support StringNumericEscape
**Description:**

Updated 2020R1 Spec:

The syntax for Unicode escapes in strings has changed from \u[CodePoint] to \u{CodePoint} so as to align with ECMAScript. Although this is an incompatible change, the previous syntax was not implemented.

Related commit - https://github.com/ballerina-platform/ballerina-spec/commit/eb2fa56ac777605dee3a26b1b7c389c1e5f7601d

### **------Old Content-------**
_From 2019R1 Spec :_

As per the spec:
```
...
string-literal :=
DoubleQuotedStringLiteral | symbolic-string-literal
DoubleQuotedStringLiteral := " (StringChar | StringEscape)* "
...
StringEscape := StringSingleEscape | StringNumericEscape
...
StringNumericEscape := \u[ CodePoint ]
CodePoint := HexDigit+
```

"In a StringNumericEscape, CodePoint must valid Unicode code point; more precisely, it
must be a hexadecimal numeral denoting an integer n where 0 <= n < 0xD800 or 0xDFFF <
n <= 0x10FFFF."

Following needs to be supported:
```ballerina
string s1 = "\u0";
string s2 = "\uD799";
string s3 = "\uEFFF";
string s4 = "\u10FFFF";
```

KavinduZoysa:
## Purpose
$subject

Fixes #13180

## Approach
Change the lexer to add the curly braces to Unicode.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/util/Constants.java
```diff
@@ -40,4 +40,5 @@ private Constants() {
 
     public static final int INIT_METHOD_SPLIT_SIZE = 50;
 
+    public static final String UNICODE_REGEX = "\\\\u[{]([a-fA-F0-9]*)[}]";
```
rdhananjaya: do we need `[{]` and `[}]` instead we should be able to use `\\{` and `\\}`

rdhananjaya: And please check what happens if an empty string is matched here and too large number matches here.
I think it will be a number format exception when parsing the int.
Shall we therefore use something like `\\\\u\\{([a-fA-F0-9]{1,6})\\}`

KavinduZoysa: > do we need `[{]` and `[}]` instead we should be able to use `\\{` and `\\}`

+1

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/string/UniCodeTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.ballerinalang.test.util.BCompileUtil;
+import org.ballerinalang.test.util.BRunUtil;
+import org.ballerinalang.test.util.CompileResult;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * Negative test class to check the Unicode patterns.
+ */
+public class UniCodeTest {
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1143,3 +1143,6 @@ error.mapping.constructor.compatible.type.not.found =\
 
 error.cannot.infer.types.for.tuple.binding=\
   invalid list constructor expression: types cannot be inferred for ''{0}''
+
+error.unsupported.unicode=\
+  unsupported unicode ''{0}''
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangParserListener.java
```diff
@@ -2648,6 +2651,22 @@ public void exitSimpleLiteral(BallerinaParser.SimpleLiteralContext ctx) {
...
+                if ((hexDecimalVal >= 0xD800 && hexDecimalVal <= 0xDFFF) || hexDecimalVal > 0x10FFFF) {
+                    String hexStringWithBraces = matcher.group(0);
+                    int offset = originalText.indexOf(hexStringWithBraces) + 1;
+                    dlog.error(new DiagnosticPos(diagnosticSrc, pos.sLine, pos.eLine, pos.sCol + offset,
+                                    pos.sCol + offset + hexStringWithBraces.length()),
+                            DiagnosticCode.INVALID_UNICODE, hexStringWithBraces);
+                }
+                text = matcher.replaceFirst("\\\\u" + hexStringVal);
+                matcher = pattern.matcher(text);
+            }
```
KRVPerera: Instead of discarding the previous match data and starting from the beginning we can start from the previous `matchers` end index.

```
int idx = 0;
while (matcher.find(idx)) {
    ....
    text = matcher.replaceFirst("\\\\u" + hexStringVal);
    idx = matcher.end() - 2;  // 2 assuming we remove two characters only by the replace
    matcher = pattern.matcher(text);
}
```

KavinduZoysa: +1 and done

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/util/Constants.java
KavinduZoysa: > And please check what happens if an empty string is matched here and too large number matches here.
> I think it will be a number format exception when parsing the int.
> Shall we therefore use something like `\\\\u\\{([a-fA-F0-9]{1,6})\\}`

According to the spec, there should be at least one hex value. So if the user enters an empty string, parse throws an error. Also, the Unicodes which are out of range(greater than 0x10FFFF) are handled [here](https://github.com/ballerina-platform/ballerina-lang/pull/20714/files#diff-87627a1d9b0a3201b74c0067bc4bdcbbR2661), because we need to throw a compile error.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangParserListener.java
```diff
@@ -2648,6 +2651,24 @@ public void exitSimpleLiteral(BallerinaParser.SimpleLiteralContext ctx) {
         } else if ((node = ctx.QuotedStringLiteral()) != null) {
             String text = node.getText();
             text = text.substring(1, text.length() - 1);
+            String originalText = text; // to log the errors
+            Pattern pattern = Pattern.compile(Constants.UNICODE_REGEX);
```
Shan1024: Since this will be called more than once (in a general scenario when compiling a Ballerina program), I think it is better to compile this Regex once (declare as a top level variable) to reduce the overhead. WDYT?

compiler/ballerina-lang/src/main/resources/grammar/BallerinaLexer.g4
```diff
@@ -341,7 +341,7 @@ EscapeSequence
 
 fragment
 UnicodeEscape
-    :   '\\' 'u' HexDigit HexDigit HexDigit HexDigit
+    :   '\\' 'u' LEFT_BRACE HexDigit+ RIGHT_BRACE
```
Shan1024: Adding braces around the value looks bit weird. Is there any specific reason to do this? 😅

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/util/Constants.java
```diff
@@ -40,4 +40,5 @@ private Constants() {
 
     public static final int INIT_METHOD_SPLIT_SIZE = 50;
 
+    public static final String UNICODE_REGEX = "\\\\u\\{([a-fA-F0-9]*)\\}";
```
Shan1024: Lexer rule is defined as `HexDigit+`. But here, the regex is defined as `[a-fA-F0-9]*`. I think this can be changed to `[a-fA-F0-9]+`. WDYT?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangParserListener.java
```diff
@@ -2648,6 +2651,24 @@ public void exitSimpleLiteral(BallerinaParser.SimpleLiteralContext ctx) {
         } else if ((node = ctx.QuotedStringLiteral()) != null) {
             String text = node.getText();
             text = text.substring(1, text.length() - 1);
```
Shan1024: I think it is better to merge these two lines in this PR as well - 

`String text = node.getText().substring(1, text.length() - 1);`

compiler/ballerina-lang/src/main/resources/grammar/BallerinaLexer.g4
hasithaa: It seems like, In the JS world, it is a norm. :) Here is the Spec change related to this. https://github.com/ballerina-platform/ballerina-spec/commit/eb2fa56ac777605dee3a26b1b7c389c1e5f7601d

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangParserListener.java
```diff
@@ -2648,6 +2651,24 @@ public void exitSimpleLiteral(BallerinaParser.SimpleLiteralContext ctx) {
...
+            String originalText = text; // to log the errors
+            Pattern pattern = Pattern.compile(Constants.UNICODE_REGEX);
+            Matcher matcher = pattern.matcher(text);
+            int position = 0;
+            while (matcher.find(position)) {
+                String hexStringVal = matcher.group(1);
+                int hexDecimalVal = Integer.parseInt(hexStringVal, 16);
+                if ((hexDecimalVal >= 0xD800 && hexDecimalVal <= 0xDFFF) || hexDecimalVal > 0x10FFFF) {
```
hasithaa: Better to move these values to constants.

compiler/ballerina-lang/src/main/resources/grammar/BallerinaLexer.g4
hasithaa: Should we change HexDigit* here to avoid a syntax error?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangParserListener.java
KavinduZoysa: Done

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/util/Constants.java
KavinduZoysa: Done

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangParserListener.java
KavinduZoysa: We cannot do that because we need to get `text.length()`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangParserListener.java
KavinduZoysa: Done

compiler/ballerina-lang/src/main/resources/grammar/BallerinaLexer.g4
KavinduZoysa: According to the spec, Unicode is defined as,
```
StringNumericEscape := \u{ CodePoint }
CodePoint := HexDigit+
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/antlr4/BallerinaLexer.java
```diff
@@ -1,4 +1,4 @@
-// Generated from BallerinaLexer.g4 by ANTLR 4.5.3
+// Generated from /home/kavindu/WSO2-GIT/test/ballerina-lang/compiler/ballerina-lang/src/main/resources/grammar/BallerinaLexer.g4 by ANTLR 4.5.3
```
rdhananjaya: Shall we generate the parser from `compiler/ballerina-lang/src/main/resources/grammar/` directory

```bash
Navigate to grammar directory:
java -jar ~/Downloads/antlr-4.5.3-complete.jar *.g4 -package org.wso2.ballerinalang.compiler.parser.antlr4 -o ../../java/org/wso2/ballerinalang/compiler/parser/antlr4/
```

hasithaa: This will get fixed, Once I combine All PRs. But it is a good practice to generate parser as @rdhananjaya mentioned.

KavinduZoysa: ACK.

PR #20747, opened 2020-01-27
Prevent compiling of test bal files when not required

### LINKED ISSUE #19229: Using --skip-tests flag does not skip compiling the tests
Shouldn't the test compilation also be skipped when the --skip-tests flag is used? Currently when I use --skip-tests, I still get the compile errors of the test files.

IrushiL:
## Purpose
Currently, the _--skip-tests_ flag in the Ballerina build command compiles the tests even though they are not executed. This PR changes this behavior to ensure that the tests are not compiled when the     _--skip-tests_ flag is used.

The existing implementation compiles the tests in the run command as well. This is also removed through this PR.

An integration test is added to check if the compilation of tests is skipped when the _--skip-tests_ flag is used in the build command.

Fixes #19229

## Approach
The TEST_ENABLED parameter of the BuildCommand and the RunCommand are changed to reflect the following requirements.
- Build Command - compiles test cases when _--skip-tests_ flag is not used
- Build Command - does not compile test cases when _--skip-tests_ flag is used
- Run Command - does not compile test cases

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [X] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/BuildCommandTest.java
+++ b/cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/BuildCommandTest.java
@@ -744,6 +744,23 @@ public class BuildCommandTest extends CommandTest {
         Path execJar = tmpDir.resolve("sample.jar");
         Assert.assertTrue(Files.exists(execJar), "Check if jar gets created");
     }
+
+    @Test(description = "Test the --skip-tests flag in the build command to ensure it avoids compiling tests")
+    public void testBuildWithSkipTests() throws IOException {
+        // valid source root path where the project contains test bal files with compilation errors
+        Path projectWithTestErrors = this.testResources.resolve("project-with-test-errors");
+        BuildCommand buildCommand = new BuildCommand(projectWithTestErrors, printStream, printStream, false, true);
+        new CommandLine(buildCommand).parse("--skip-tests", "-a");
+        buildCommand.execute();
+
+        String buildLog = readOutput(true);
+        Assert.assertEquals(buildLog.replaceAll("\r", ""), "Compiling source\n" +
+                "\ttestOrg/module1:0.1.0\n" +
+                "\nCreating balos\n" +
+                "\ttarget/balo/module1-2019r3-any-0.1.0.balo\n" +
+                "\nGenerating executables\n" +
+                "\ttarget/bin/module1.jar\n");
+    }
     
     // Check compile command inside a module directory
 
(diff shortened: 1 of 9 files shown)


cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/BuildCommandTest.java
```diff
@@ -744,6 +744,23 @@ public void testBuildCommandSingleFileWithOutput() throws IOException {
         Path execJar = tmpDir.resolve("sample.jar");
         Assert.assertTrue(Files.exists(execJar), "Check if jar gets created");
     }
+
+    @Test(description = "Test the --skip-tests flag in the build command to ensure it avoids compiling tests")
+    public void testBuildWithSkipTests() throws IOException {
```
anupama-pathirage: Why we are throwing IOException from here?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/packaging/converters/FilterSearch.java
```diff
@@ -47,7 +47,7 @@ public FileVisitResult visitFile(Object file, BasicFileAttributes attrs) throws
      */
     private boolean isExcluded(Path path) throws IOException {
```
anupama-pathirage: Do we need to throw IOException from here?

cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/BuildCommandTest.java
IrushiL: Since the output log generated from the build command is read at line 756 and the `readOutput` method used for this throws an IOException.

anupama-pathirage: Sorry my bad yor are correct., This comment was for  line 732 method which was already there.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/packaging/converters/FilterSearch.java
IrushiL: This is not needed, removed it.

PR #20749, opened 2020-01-27
Fill-in missing values by list constructor

### LINKED ISSUE #20721: Error in spec of fill-in of missing values by list constructor
Implementation issue to track https://github.com/ballerina-platform/ballerina-spec/issues/349

KRVPerera:
## Purpose
Fixes #20721

## Approach

Changed explicit array size and list constructor size mismatch check by loosening it to only to error out scenarios where the list constructor is trying to create an array larger than given explicit array length.

Added fill value check to test if the array is created using a type which has a valid fill value

## Samples
Code Snippet
```ballerina
    int[4] arr = [3, 1];
    io:println("arr : ", arr);
    arr[3] = 30;
    io:println("arr : ", arr)
```
Output:
```
    arr : 3 1 0 0
    arr : 3 1 0 30
```
#### ToDo
- [x] boolean / int / float / decimal / string/ byte arrays
- [x] any
- [x] anydata
- [x] unions
- [x] tuples
- [x] records
- [x] Singleton array
- [ ] Object array  - Issue #20983 
- [x] json array
- [x] xml arrays

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [x] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
@@ -813,7 +813,9 @@ public class ArrayValueImpl extends AbstractArrayValue {
             case TypeTags.BOOLEAN_TAG:
                 return;
             default:
-                Arrays.fill(refValues, size, index, elementType.getZeroValue());
+                if (arrayType.hasFillerValue()) {
+                    Arrays.fill(refValues, size, index, elementType.getZeroValue());
+                }
         }
     }
 
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
@@ -848,8 +848,6 @@ public class ArrayValueImpl extends AbstractArrayValue {
 
     @Override
     protected void fillerValueCheck(int index, int size) {
-        // if there has been values added beyond the current index, that means filler values
-        // has already been checked. Therefore no need to check again.
         if (!hasFillerValue.isPresent()) {
             hasFillerValue = Optional.of(TypeChecker.hasFillerValue(this.elementType));
         }
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
@@ -40,7 +40,6 @@ import java.io.OutputStream;
 import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.Map;
-import java.util.Optional;
 import java.util.StringJoiner;
 import java.util.stream.IntStream;
 
@@ -71,8 +70,6 @@ public class ArrayValueImpl extends AbstractArrayValue {
     private double[] floatValues;
     private String[] stringValues;
     private BString[] bStringValues;
-
-    private Optional<Boolean> hasFillerValue = Optional.empty();
 
     // ------------------------ Constructors -------------------------------------------------------------------
 
@@ -817,7 +814,6 @@ public class ArrayValueImpl extends AbstractArrayValue {
                 return;
             default:
                 Arrays.fill(refValues, size, index, elementType.getZeroValue());
-
         }
     }
 
@@ -848,17 +844,13 @@ public class ArrayValueImpl extends AbstractArrayValue {
 
     @Override
     protected void fillerValueCheck(int index, int size) {
-        if (!hasFillerValue.isPresent()) {
-            hasFillerValue = Optional.of(TypeChecker.hasFillerValue(this.elementType));
-        }
-
         // if the elementType doesn't have an implicit initial value & if the insertion is not a consecutive append
         // to the array, then an exception will be thrown.
-        if (hasFillerValue.get()) {
+        if (arrayType.hasFillerValue()) {
             return;
         } else if (index > size) {
             throw BLangExceptionHelper.getRuntimeException(BallerinaErrorReasons.ILLEGAL_LIST_INSERTION_ERROR,
-                    RuntimeErrors.ILLEGAL_ARRAY_INSERTION, size, index + 1);
+                                                           RuntimeErrors.ILLEGAL_ARRAY_INSERTION, size, index + 1);
         }
     }
 
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
@@ -40,6 +40,7 @@ import java.io.OutputStream;
 import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.Optional;
 import java.util.StringJoiner;
 import java.util.stream.IntStream;
 
@@ -70,6 +71,8 @@ public class ArrayValueImpl extends AbstractArrayValue {
     private double[] floatValues;
     private String[] stringValues;
     private BString[] bStringValues;
+
+    private Optional<Boolean> hasFillerValue = Optional.empty();;
 
     // ------------------------ Constructors -------------------------------------------------------------------
 
@@ -130,35 +133,40 @@ public class ArrayValueImpl extends AbstractArrayValue {
         this.arrayType = type;
         BArrayType arrayType = (BArrayType) type;
         this.elementType = arrayType.getElementType();
+        initArrayValues(elementType, false);
         if (arrayType.getState() == ArrayState.CLOSED_SEALED) {
             this.size = maxSize = arrayType.getSize();
         }
-        initArrayValues(this.elementType, false);
     }
 
     private void initArrayValues(BType elementType, boolean useBString) {
+        int initialArraySize = Math.max(DEFAULT_ARRAY_SIZE, arrayType.getSize());
         switch (elementType.getTag()) {
             case TypeTags.INT_TAG:
-                this.intValues = new long[DEFAULT_ARRAY_SIZE];
+                this.intValues = new long[initialArraySize];
                 break;
             case TypeTags.FLOAT_TAG:
-                this.floatValues = new double[DEFAULT_ARRAY_SIZE];
+                this.floatValues = new double[initialArraySize];
                 break;
             case TypeTags.STRING_TAG:
                 if (useBString) {
-                    this.bStringValues = new BString[DEFAULT_ARRAY_SIZE];
+                    this.bStringValues = new BString[initialArraySize];
                 } else {
-                    this.stringValues = new String[DEFAULT_ARRAY_SIZE];
+                    this.stringValues = new String[initialArraySize];
                 }
                 break;
             case TypeTags.BOOLEAN_TAG:
-                this.booleanValues = new boolean[DEFAULT_ARRAY_SIZE];
+                this.booleanValues = new boolean[initialArraySize];
                 break;
             case TypeTags.BYTE_TAG:
-                this.byteValues = new byte[DEFAULT_ARRAY_SIZE];
+                this.byteValues = new byte[initialArraySize];
                 break;
             default:
-                this.refValues = new Object[DEFAULT_ARRAY_SIZE];
+                this.refValues = new Object[initialArraySize];
+                if (arrayType.getState() == ArrayState.CLOSED_SEALED) {
+                    fillerValueCheck(initialArraySize, initialArraySize);
+                    fillValues(arrayType.getSize());
+                }
         }
     }
 
@@ -166,20 +174,20 @@ public class ArrayValueImpl extends AbstractArrayValue {
     public ArrayValueImpl(BArrayType type, long size) {
         this.arrayType = type;
         this.elementType = type.getElementType();
+        initArrayValues(this.elementType, false);
         if (size != -1) {
             this.size = this.maxSize = (int) size;
         }
-        initArrayValues(this.elementType, false);
     }
 
     @Deprecated
     public ArrayValueImpl(BArrayType type, long size, boolean useBString) {
         this.arrayType = type;
         this.elementType = type.getElementType();
+        initArrayValues(this.elementType, useBString);
         if (size != -1) {
             this.size = this.maxSize = (int) size;
         }
-        initArrayValues(this.elementType, useBString);
     }
 
     // ----------------------- get methods ----------------------------------------------------
@@ -842,13 +850,15 @@ public class ArrayValueImpl extends AbstractArrayValue {
     protected void fillerValueCheck(int index, int size) {
         // if there has been values added beyond the current index, that means filler values
         // has already been checked. Therefore no need to check again.
-        if (this.size >= index) {
-            return;
+        if (!hasFillerValue.isPresent()) {
+            hasFillerValue = Optional.of(TypeChecker.hasFillerValue(this.elementType));
         }
 
         // if the elementType doesn't have an implicit initial value & if the insertion is not a consecutive append
         // to the array, then an exception will be thrown.
-        if (!TypeChecker.hasFillerValue(this.elementType) && (index > size)) {
+        if (hasFillerValue.get()) {
+            return;
+        } else if (index > size) {
             throw BLangExceptionHelper.getRuntimeException(BallerinaErrorReasons.ILLEGAL_LIST_INSERTION_ERROR,
                     RuntimeErrors.ILLEGAL_ARRAY_INSERTION, size, index + 1);
         }
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
@@ -72,7 +72,7 @@ public class ArrayValueImpl extends AbstractArrayValue {
     private String[] stringValues;
     private BString[] bStringValues;
 
-    private Optional<Boolean> hasFillerValue = Optional.empty();;
+    private Optional<Boolean> hasFillerValue = Optional.empty();
 
     // ------------------------ Constructors -------------------------------------------------------------------
 
@@ -140,7 +140,7 @@ public class ArrayValueImpl extends AbstractArrayValue {
     }
 
     private void initArrayValues(BType elementType, boolean useBString) {
-        int initialArraySize = Math.max(DEFAULT_ARRAY_SIZE, arrayType.getSize());
+        int initialArraySize = (arrayType.getSize() != -1) ? arrayType.getSize() : DEFAULT_ARRAY_SIZE;
         switch (elementType.getTag()) {
             case TypeTags.INT_TAG:
                 this.intValues = new long[initialArraySize];
@@ -165,7 +165,7 @@ public class ArrayValueImpl extends AbstractArrayValue {
                 this.refValues = new Object[initialArraySize];
                 if (arrayType.getState() == ArrayState.CLOSED_SEALED) {
                     fillerValueCheck(initialArraySize, initialArraySize);
-                    fillValues(arrayType.getSize());
+                    fillValues(initialArraySize);
                 }
         }
     }
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -613,13 +613,13 @@ public class TypeChecker extends BLangNodeVisitor {
         BType actualType = symTable.semanticError;
         resultType = symTable.semanticError;
 
-        if ((expType.tag == TypeTags.ANY || expType.tag == TypeTags.ANYDATA || expType.tag == TypeTags.NONE)
+        int expTypeTag = expType.tag;
+        if ((expTypeTag == TypeTags.ANY || expTypeTag == TypeTags.ANYDATA || expTypeTag == TypeTags.NONE)
                 && listConstructor.exprs.isEmpty()) {
             dlog.error(listConstructor.pos, DiagnosticCode.INVALID_LIST_CONSTRUCTOR, expType);
             return;
         }
 
-        int expTypeTag = expType.tag;
         if (expTypeTag == TypeTags.JSON) {
             checkExprs(listConstructor.exprs, this.env, expType);
             actualType = expType;
@@ -628,10 +628,17 @@ public class TypeChecker extends BLangNodeVisitor {
             if (arrayType.state == BArrayState.OPEN_SEALED) {
                 arrayType.size = listConstructor.exprs.size();
                 arrayType.state = BArrayState.CLOSED_SEALED;
-            } else if (arrayType.state != BArrayState.UNSEALED && arrayType.size != listConstructor.exprs.size()) {
-                dlog.error(listConstructor.pos,
-                        DiagnosticCode.MISMATCHING_ARRAY_LITERAL_VALUES, arrayType.size, listConstructor.exprs.size());
-                return;
+            } else if ((arrayType.state != BArrayState.UNSEALED) && (arrayType.size != listConstructor.exprs.size())) {
+                if (arrayType.size < listConstructor.exprs.size()) {
+                    dlog.error(listConstructor.pos,
+                               DiagnosticCode.MISMATCHING_ARRAY_LITERAL_VALUES, arrayType.size,
+                               listConstructor.exprs.size());
+                    return;
+                }
+                if (!types.hasFillerValue(arrayType.eType)) {
+                    dlog.error(listConstructor.pos, DiagnosticCode.INVALID_LIST_CONSTRUCTOR, expType);
+                    return;
+                }
             }
             checkExprs(listConstructor.exprs, this.env, arrayType.eType);
             actualType = arrayType;
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -476,7 +476,7 @@ public class TypeChecker extends BLangNodeVisitor {
 
     private boolean literalAssignableToFiniteType(BLangLiteral literalExpr, BFiniteType finiteType,
                                                   int targetMemberTypeTag) {
-        return finiteType.valueSpace.stream()
+        return finiteType.getValueSpace().stream()
                 .anyMatch(valueExpr -> valueExpr.type.tag == targetMemberTypeTag &&
                         types.checkLiteralAssignabilityBasedOnType((BLangLiteral) valueExpr, literalExpr));
     }
@@ -527,7 +527,7 @@ public class TypeChecker extends BLangNodeVisitor {
         Set<BLangExpression> matchedValueSpace = new LinkedHashSet<>();
 
         for (BFiniteType finiteType : finiteTypeMembers) {
-            matchedValueSpace.addAll(finiteType.valueSpace.stream()
+            matchedValueSpace.addAll(finiteType.getValueSpace().stream()
                                              .filter(expression -> expression.type.tag == tag)
                                              .collect(Collectors.toSet()));
         }
@@ -636,7 +636,7 @@ public class TypeChecker extends BLangNodeVisitor {
                     return;
                 }
                 if (!types.hasFillerValue(arrayType.eType)) {
-                    dlog.error(listConstructor.pos, DiagnosticCode.INVALID_LIST_CONSTRUCTOR, expType);
+                    dlog.error(listConstructor.pos, DiagnosticCode.INVALID_LIST_CONSTRUCTOR_ELEMENT_TYPE, expType);
                     return;
                 }
             }
@@ -3182,7 +3182,7 @@ public class TypeChecker extends BLangNodeVisitor {
                 return false;
             } else {
                 BFiniteType reasonType = (BFiniteType) ctorType.reasonType;
-                if (reasonType.valueSpace.size() > 1) {
+                if (reasonType.getValueSpace().size() > 1) {
                     dlog.error(iExpr.pos, DiagnosticCode.INDIRECT_ERROR_CTOR_NOT_ALLOWED_ON_NON_CONST_REASON,
                             iExpr.type);
                     return false;
@@ -3211,7 +3211,7 @@ public class TypeChecker extends BLangNodeVisitor {
             return true;
         } else {
             BFiniteType finiteType = (BFiniteType) errorType.reasonType;
-            if (finiteType.valueSpace.size() != 1) {
+            if (finiteType.getValueSpace().size() != 1) {
                 if (errorType == symTable.errorType) {
                     dlog.error(pos, DiagnosticCode.CANNOT_INFER_ERROR_TYPE, expType.tsymbol.name);
                 } else {
@@ -3248,7 +3248,7 @@ public class TypeChecker extends BLangNodeVisitor {
     private void setErrorReasonParam(BLangInvocation iExpr, boolean reasonArgGiven, BErrorType ctorType) {
         if (!reasonArgGiven && ctorType.reasonType.getKind() == TypeKind.FINITE) {
             BFiniteType finiteType = (BFiniteType) ctorType.reasonType;
-            BLangExpression reasonExpr = (BLangExpression) finiteType.valueSpace.toArray()[0];
+            BLangExpression reasonExpr = (BLangExpression) finiteType.getValueSpace().toArray()[0];
             iExpr.requiredArgs.add(reasonExpr);
             return;
         }
@@ -4418,7 +4418,7 @@ public class TypeChecker extends BLangNodeVisitor {
             case TypeTags.FINITE:
                 BFiniteType finiteIndexExpr = (BFiniteType) indexExprType;
                 boolean validIndexExists = false;
-                for (BLangExpression finiteMember : finiteIndexExpr.valueSpace) {
+                for (BLangExpression finiteMember : finiteIndexExpr.getValueSpace()) {
                     int indexValue = ((Long) ((BLangLiteral) finiteMember).value).intValue();
                     if (indexValue >= 0 &&
                             (arrayType.state == BArrayState.UNSEALED || indexValue < arrayType.size)) {
@@ -4443,7 +4443,7 @@ public class TypeChecker extends BLangNodeVisitor {
                     finiteType = finiteTypes.get(0);
                 } else {
                     Set<BLangExpression> valueSpace = new LinkedHashSet<>();
-                    finiteTypes.forEach(constituent -> valueSpace.addAll(constituent.valueSpace));
+                    finiteTypes.forEach(constituent -> valueSpace.addAll(constituent.getValueSpace()));
                     finiteType = new BFiniteType(null, valueSpace);
                 }
 
@@ -4504,7 +4504,7 @@ public class TypeChecker extends BLangNodeVisitor {
             case TypeTags.FINITE:
                 BFiniteType finiteIndexExpr = (BFiniteType) currentType;
                 LinkedHashSet<BType> possibleTypes = new LinkedHashSet<>();
-                for (BLangExpression finiteMember : finiteIndexExpr.valueSpace) {
+                for (BLangExpression finiteMember : finiteIndexExpr.getValueSpace()) {
                     int indexValue = ((Long) ((BLangLiteral) finiteMember).value).intValue();
                     BType fieldType = checkTupleFieldType(tuple, indexValue);
                     if (fieldType.tag != TypeTags.SEMANTIC_ERROR) {
@@ -4539,7 +4539,7 @@ public class TypeChecker extends BLangNodeVisitor {
                     finiteType = finiteTypes.get(0);
                 } else {
                     Set<BLangExpression> valueSpace = new LinkedHashSet<>();
-                    finiteTypes.forEach(constituent -> valueSpace.addAll(constituent.valueSpace));
+                    finiteTypes.forEach(constituent -> valueSpace.addAll(constituent.getValueSpace()));
                     finiteType = new BFiniteType(null, valueSpace);
                 }
 
@@ -4655,7 +4655,7 @@ public class TypeChecker extends BLangNodeVisitor {
             case TypeTags.FINITE:
                 BFiniteType finiteIndexExpr = (BFiniteType) currentType;
                 LinkedHashSet<BType> possibleTypes = new LinkedHashSet<>();
-                for (BLangExpression finiteMember : finiteIndexExpr.valueSpace) {
+                for (BLangExpression finiteMember : finiteIndexExpr.getValueSpace()) {
                     String fieldName = (String) ((BLangLiteral) finiteMember).value;
                     BType fieldType = checkRecordRequiredFieldAccess(accessExpr, names.fromString(fieldName), record);
                     if (fieldType == symTable.semanticError) {
@@ -4707,7 +4707,7 @@ public class TypeChecker extends BLangNodeVisitor {
                     finiteType = finiteTypes.get(0);
                 } else {
                     Set<BLangExpression> valueSpace = new LinkedHashSet<>();
-                    finiteTypes.forEach(constituent -> valueSpace.addAll(constituent.valueSpace));
+                    finiteTypes.forEach(constituent -> valueSpace.addAll(constituent.getValueSpace()));
                     finiteType = new BFiniteType(null, valueSpace);
                 }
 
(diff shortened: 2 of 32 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/sealed-array.bal
```diff
@@ -257,3 +327,13 @@ function testSealedArrayConstrainedMapInvalidIndex (int[3] x1, int index) {
...
+
+// xml arrays
+
+function createXMLAutoFilledSealedArray() returns int {
+    xml a = xml `<name>Ballerina</name>`;
+    xml[5] sealedArray = [a, a];
+    sealedArray[3] = a;
+    xml test = sealedArray[4];
+    return sealedArray.length();
+}
```
MaryamZi:
```suggestion
}

```

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/sealed-array-semantics-negative.bal
```diff
@@ -27,7 +27,7 @@ function accessInvalidIndexOfSealedArrayUsingKeyword() returns string {
 }
 
 function initializeInvalidSizedSealedArray() {
-    boolean[4] sealedArray1 = [true, true, false];
+    //boolean[4] sealedArray1 = [true, true, false]; // this is valid now
```
MaryamZi: Why are we commenting these out instead of removing them?

KRVPerera: Let me remove. I wanted to avoid rewriting every line number of the java side asset on message and line number.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/SealedArrayTest.java
```diff
@@ -53,6 +53,11 @@ public void testCreateIntegerSealedArray() {
                 returnValues == null || returnValues.length == 0 || returnValues[0] == null, "Invalid Return Values.");
         Assert.assertEquals(((BInteger) returnValues[0]).intValue(), 5, "Length didn't match");
 
+        returnValues = BRunUtil.invoke(compileResult, "createIntAutoFilledSealedArray");
```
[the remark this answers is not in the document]
KRVPerera: Using testerina asserts ? or with If conditions and panics? 
What is more suitable in the long term?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -609,10 +609,14 @@ public void visit(BLangListConstructorExpr listConstructor) {
...
                 arrayType.state = BArrayState.CLOSED_SEALED;
-            } else if (arrayType.state != BArrayState.UNSEALED && arrayType.size != listConstructor.exprs.size()) {
-                dlog.error(listConstructor.pos,
-                        DiagnosticCode.MISMATCHING_ARRAY_LITERAL_VALUES, arrayType.size, listConstructor.exprs.size());
-                return;
+            } else if ((arrayType.state != BArrayState.UNSEALED) && (arrayType.size != listConstructor.exprs.size())) {
+                if (arrayType.size < listConstructor.exprs.size()) {
+                    dlog.error(listConstructor.pos,
+                            DiagnosticCode.MISMATCHING_ARRAY_LITERAL_VALUES, arrayType.size, listConstructor.exprs.size
+                            ());
```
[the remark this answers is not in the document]
KRVPerera: Reaching 120 character limit. Check now.

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/sealed-array.bal
```diff
@@ -257,3 +334,36 @@ function testSealedArrayConstrainedMapInvalidIndex (int[3] x1, int index) {
...
+        panic error(message + " Expected : " + expected.toString() + " Actual : " + actual.toString());
+    }
+}
+
+function isEqualPanic(string expected, any[] arr, string message = "Not equal") {
+    string actual = arr.toString().trim();
+    if (expected != actual) {
+        panic error(message + " Expected : " + expected + " Actual : " + actual);
+    }
+}
```
kalaiyarasiganeshalingam: Add new line at EOF.

KRVPerera: done

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArrayFillTestRuntime.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
kalaiyarasiganeshalingam:
```suggestion
 * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2563,196 @@ public int hashCode() {
...
+            if (!initFunction.symbol.getParameters().stream()
+                    .allMatch(bVarSymbol -> bVarSymbol.defaultableParam == true)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    /**
+     * This will handle two types
```
kalaiyarasiganeshalingam:
```suggestion
     * This will handle two types.
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -4766,4 +4775,6 @@ private boolean isConst(BLangExpression expression) {
     private Name getCurrentCompUnit(BLangNode node) {
         return names.fromString(node.pos.getSource().getCompilationUnitName());
     }
+
+
```
kalaiyarasiganeshalingam: Remove extra lines

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -619,10 +619,18 @@ public void visit(BLangListConstructorExpr listConstructor) {
             if (arrayType.state == BArrayState.OPEN_SEALED) {
                 arrayType.size = listConstructor.exprs.size();
                 arrayType.state = BArrayState.CLOSED_SEALED;
-            } else if (arrayType.state != BArrayState.UNSEALED && arrayType.size != listConstructor.exprs.size()) {
-                dlog.error(listConstructor.pos,
-                        DiagnosticCode.MISMATCHING_ARRAY_LITERAL_VALUES, arrayType.size, listConstructor.exprs.size());
-                return;
+            } else if ((arrayType.state != BArrayState.UNSEALED) && (arrayType.size != listConstructor.exprs.size())) {
```
dulvinw: If we move this logic to a function, we can improve the readability of the code. WDYT?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KRVPerera: done

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KRVPerera: It will be like two functions ..
```if (isSealedArray(arrayType) && sizeMismatch(arrayType, listConstructor.exprs))```. Will this help ?

File: compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java [hunk 10]
```diff
@@ -2537,4 +2563,196 @@ public int hashCode() {
...
+
+    private boolean checkFillerValue(BUnionType type) {
+        if (type.isNullable()) {
+            return true;
+        }
+
+        Iterator<BType> iterator = type.getMemberTypes().iterator();
+        BType firstMember = iterator.next();
+        // is first value is a valid fill value
+        if (firstMember.getKind() == TypeKind.NIL) {
```
MaryamZi: If this was true we would have already returned at https://github.com/ballerina-platform/ballerina-lang/pull/20749/files#diff-782094a79e6831ac57a34ad157af49c1R2678 right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2563,196 @@ public int hashCode() {
...
+        // is first value is a valid fill value
+        String defaultFillValue = getDefaultFillValue(firstMember);
+        if (firstMember.toString().equals(defaultFillValue)) {
+            defaultFillValuePresent = true;
+        }
+
+        while (iterator.hasNext()) {
+
+            Object value =  iterator.next();
+            if (value == null) {
```
MaryamZi: Again, can these be null?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2563,196 @@ public int hashCode() {
...
+
+        if (!allMembersHaveSameType) {
+            return false;
+        }
+        return defaultFillValuePresent;
+    }
+
+    private boolean checkFillerValue(BRecordType type) {
+        for (BField field : type.fields) {
+            if (field.symbol.defaultableParam) {
```
MaryamZi: Is this set for record fields too? IINM, we usually check only a flag.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2563,196 @@ public int hashCode() {
...
+                continue;
+            }
+            if (!hasFillerValue(field.type)) {
```
MaryamZi: IINM, this is not correct. According to the spec, fill member should be done only for lists.

```ballerina
type Foo record {
   string s;
};

Foo[2] arr = [];
```

This should still fail IMO.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: Let me remove. I think this should not be null

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2563,196 @@ public int hashCode() {
...
+            }
+
+            BType valueType = ((BLangExpression) value).type;
+            if (valueType.getKind() == TypeKind.NIL) {
+                return true;
+            }
+
+            if (allMembersHaveSameType && !isSameType(valueType, firstElement.type)) {
+                allMembersHaveSameType = false;
+            }
```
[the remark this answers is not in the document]
KRVPerera: What is at the end of the union there is a ```nil``` value. Therefore I let the loop continue. If there is ```nil``` it is the fill value.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
```diff
@@ -130,56 +133,60 @@ public ArrayValueImpl(BArrayType type) {
...
         this.elementType = arrayType.getElementType();
+        initArrayValues(elementType, false);
         if (arrayType.getState() == ArrayState.CLOSED_SEALED) {
             this.size = maxSize = arrayType.getSize();
         }
-        initArrayValues(this.elementType, false);
     }
 
     private void initArrayValues(BType elementType, boolean useBString) {
+        int initialArraySize = Math.max(DEFAULT_ARRAY_SIZE, arrayType.getSize());
```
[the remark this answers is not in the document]
KRVPerera: Yes, I agree. For a fixed-length array (explicit size), there will be an unused allocated capacity due to this logic.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2563,196 @@ public int hashCode() {
...
+            if ((type.tsymbol.flags & Flags.ABSTRACT) == Flags.ABSTRACT) {
+                return false;
+            }
+        } else {
+            if (initFunction.symbol.getReturnType().getKind() == TypeKind.ERROR) {
+                return false;
+            }
+            if (!hasFillerValue(initFunction.symbol.getReturnType())) {
+                return false;
+            }
```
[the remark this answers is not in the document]
KRVPerera: yes

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2563,196 @@ public int hashCode() {
...
+    private boolean checkFillerValue(BFiniteType type) {
+        // For singleton types, that value is the implicit initial value
+        if (type.valueSpace.size() == 1) {
+            return true;
+        }
+
+        // is first value null
+        Iterator iterator = type.valueSpace.iterator();
+        BLangExpression firstElement = (BLangExpression) iterator.next();
+        if ((firstElement == null) || (firstElement.type.getKind() == TypeKind.NIL)) {
```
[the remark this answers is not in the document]
KRVPerera: let me check the null value.
IsNullable for "BFiniteType" actually goes through every item in value space.
```Java
public boolean isNullable() {
        return this.valueSpace.stream().anyMatch(v -> v.type.tag == TypeTags.NIL);
}
```
May be we can remove nullable check in union type code inside loop and for first item. Let me check.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -109,6 +110,31 @@
     private Names names;
     private int finiteTypeCount = 0;
 
+    /**
+     * Keep default values for basic types in String format.
+     *
```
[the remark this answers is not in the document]
KRVPerera: added 1.1.2

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -619,16 +620,25 @@ public void visit(BLangListConstructorExpr listConstructor) {
...
+                if (arrayType.size < listConstructor.exprs.size()) {
+                    dlog.error(listConstructor.pos,
+                               DiagnosticCode.MISMATCHING_ARRAY_LITERAL_VALUES, arrayType.size,
+                               listConstructor.exprs.size());
+                    return;
+                }
+                //TODO: check fillable values
+                if (!types.hasFillerValue(arrayType.eType)) {
+                    dlog.error(listConstructor.pos, DiagnosticCode.INVALID_LIST_CONSTRUCTOR, expType);
```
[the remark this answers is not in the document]
KRVPerera: Check the error message on 281133

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: Commit https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/f21bca4d1e537732576a0558486343c55dd1def9
should address this.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BFiniteType.java
```diff
@@ -37,9 +37,11 @@
  */
 public class BFiniteType extends BType implements FiniteType {
 
-    public Set<BLangExpression> valueSpace;
+    private Set<BLangExpression> valueSpace;
```
MaryamZi: Shouldn't the getter return an unmodifiable set too?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -109,6 +110,32 @@
     private Names names;
     private int finiteTypeCount = 0;
 
+    /**
+     * Keep filler value for basic types in String format.
+     *
+     * @since 1.1.2
```
MaryamZi:
```suggestion
     * @since 1.2.0
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2564,166 @@ public int hashCode() {
...
+                return false;
+        }
+    }
+
+    private boolean checkFillerValue(BObjectType type) {
+        BAttachedFunction initFunction = ((BObjectTypeSymbol) type.tsymbol).initializerFunc;
+        if (initFunction == null) {
+            if ((type.tsymbol.flags & Flags.ABSTRACT) == Flags.ABSTRACT) {
+                return false;
+            }
```
MaryamZi: This can be

```java
    return (type.tsymbol.flags & Flags.ABSTRACT) != Flags.ABSTRACT;
```

right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2564,166 @@ public int hashCode() {
...
+        } else {
```
MaryamZi: We can then remove this else and move out this block?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2564,166 @@ public int hashCode() {
...
+            if (initFunction.symbol.getReturnType().getKind() == TypeKind.ERROR) {
```
MaryamZi: IINM, this can never be just error. It's either nil or a union with error. Just checking `initFunction.symbol.getReturnType().getKind() != TypeKind.NIL` should be enough?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2564,166 @@ public int hashCode() {
...
+
+        // For singleton types, that value is the implicit initial value
+        if (type.getValueSpace().size() == 1) {
+            return true;
+        }
+
+        Iterator iterator = type.getValueSpace().iterator();
+        if (!iterator.hasNext()) { // sanity check this cannot be
+            return false;
+        }
```
MaryamZi: IMO, we can remove this.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2564,166 @@ public int hashCode() {
...
+        BLangExpression firstElement = (BLangExpression) iterator.next();
+        String defaultFillValue = getDefaultFillValue(firstElement);
+        if (firstElement.toString().equals(defaultFillValue)) {
+            defaultFillValuePresent = true;
+        }
+
+        while (iterator.hasNext()) {
+            Object value =  iterator.next();
+            BType valueType = ((BLangExpression) value).type;
+            if (!isSameType(valueType, firstElement.type)) {
```
MaryamZi: Can we extract out `firstElement.type` to a variable?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: Even if the object is not an abstract object we still need to do some checks or parameters in the initializer function. But I changed the logic giving importance to abstraction check. Can you check the new logic whether I got your point correctly? https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/ff0acc05501340c2064c3bb1b3211752955782ad

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2564,160 @@ public int hashCode() {
...
+        if (type.isNullable()) {
+            return true;
+        }
+        Iterator<BType> iterator = type.getMemberTypes().iterator();
+        BType firstMember = iterator.next();
+        boolean defaultFillValuePresent = false;
+
+        // is first value is a valid fill value
+        String defaultFillValue = getDefaultFillValue(firstMember);
+        if (firstMember.toString().equals(defaultFillValue)) {
```
MaryamZi: Can we validate if this works for constants?

```ballerina
const ZERO = 0;

type Foo ZERO|1|2;
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
```diff
@@ -840,15 +848,15 @@ protected void rangeCheck(long index, int size) {
 
     @Override
     protected void fillerValueCheck(int index, int size) {
-        // if there has been values added beyond the current index, that means filler values
-        // has already been checked. Therefore no need to check again.
-        if (this.size >= index) {
-            return;
+        if (!hasFillerValue.isPresent()) {
```
SupunS: This `fillerValueCheck()` method is called for each and every add operation. So could this be an overhead?

Since `hasFillerValue` needs to be calculated only once for a given type, can we move this flag to the array-type, and have it calculated only once during the initialisation of the type?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2537,4 +2564,160 @@ public int hashCode() {
...
+
+    private boolean checkFillerValue(BArrayType type) {
+        return hasFillerValue(type.eType);
+    }
+
+    private String getDefaultFillValue(BLangExpression expr) {
+        return getDefaultFillValue(expr.type);
+    }
+
+    private String getDefaultFillValue(BType type) {
```
MaryamZi: Wondering if keeping this in the String format is the best approach? Can't we use the value itself and compare, based on the assumption that they will only be used for comparison with either literals or constant refs?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/SealedArrayTest.java
```diff
@@ -36,177 +36,117 @@
...
-        Assert.assertEquals(((BInteger) returnValues[0]).intValue(), 5, "Length didn't match");
+        BRunUtil.invoke(compileResult, "createFloatAutoFilledSealedArray");
 
-        returnValues = BRunUtil.invoke(compileResult, "createFloatDefaultSealedArray");
-        Assert.assertFalse(
-                returnValues == null || returnValues.length == 0 || returnValues[0] == null, "Invalid Return Values.");
-        Assert.assertEquals(returnValues[0].stringValue(), "[0.0, 0.0, 0.0, 0.0, 0.0]", "Length didn't match");
-        Assert.assertEquals(((BInteger) returnValues[1]).intValue(), 5, "Length didn't match");
+        BRunUtil.invoke(compileResult, "createFloatSealedArrayWithLabel");
+
```
MaryamZi: Shall we remove the extra lines?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/SealedArrayTest.java
```diff
@@ -282,51 +222,99 @@ public void testNegativeSealedArrays() {
         BAssertUtil.validateError(resultNegative, 0, "variable 'sealedArray1' is not initialized", 19, 5);
     }
 
+    // TODO : uncomment
```
MaryamZi: Invalid comment?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/SealedArrayTest.java
```diff
@@ -282,51 +222,99 @@ public void testNegativeSealedArrays() {
...
+    @Test()
+    public void testNegativeAutoFillSealedArray() {
+        Assert.assertEquals(listExprNegative.getErrorCount(), 11);
+        BAssertUtil.validateError(listExprNegative, 0,
+                                  "invalid usage of list constructor: type 'Person[5]' does not have a filler value",
+                                  24,
+                                  19);
```
MaryamZi: Can we have these in a single line?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArrayFillTestRuntime.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.ballerinalang.test.util.BCompileUtil;
+import org.ballerinalang.test.util.CompileResult;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * Test cases for filling the elements of the array with its type's implicit initial value.
+ *
+ * @since 1.1.1
```
MaryamZi: Incorrect version?

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-fill-test.bal
```diff
@@ -150,6 +157,14 @@ function testAnyArrayFill(int index) returns any[] {
     return ar;
 }
 
+function testAnySealedArrayFill(int index) returns any[] {
+    Person p = {name:"John", age:25};
```
MaryamZi:
```suggestion
    Person p = {name: "John", age: 25};
```

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-test.bal
```diff
@@ -153,3 +153,23 @@ function testGetFromFrozenArray() returns int {
...
+    }
+};
+
+function testObjectDynamicArrayFilling() {
+    Age[] y = [];
+    y[0] = new(5);
+    y[1] = new(5);
+}
+
+function assertArrayLengthPanic(int expected, any[] arr, string message = "Array length did not match") {
```
MaryamZi: This is not used?

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-test.bal
```diff
@@ -153,3 +153,23 @@ function testGetFromFrozenArray() returns int {
...
 }
+
+type Age object {
+    public int age;
+    public function __init(int age) {
+    	 self.age = age;
+    }
+};
+
+function testObjectDynamicArrayFilling() {
```
MaryamZi: Is this test complete/called?

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/sealed_array_listexpr_negative.bal
```diff
@@ -0,0 +1,113 @@
...
+
+// [0] cannot initialize abstract objects
+function createAbstractObjectArray() {
+    Person[5] x = [];
+    Person p1 = x[2];
+    p1.age = 20;
+}
+
+// [1] cannot create multi dimensional arrays of abstract objects
+function createAbstractObjectEmptyArray() {
```
MaryamZi: The names are not quite clear IMO. :)

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/sealed_array_listexpr_negative.bal
```diff
@@ -0,0 +1,113 @@
...
+    Person[5][] y = [];
```
MaryamZi: This shouldn't fail right?

`Person[5][]` means 5 arrays of open Person arrays (`Person[]`)? So for the members, `[]` is a valid constructor?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
KRVPerera: fixed by https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/c0998bf94dc83e767795d8135dc411e1605c6405

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-fill-test.bal
```diff
@@ -70,6 +70,13 @@ function testTupleArrayFill(int index) returns [string, int][] {
     return ar;
 }
 
+function testTupleSealedArrayFill(int index) returns [string, int][] {
```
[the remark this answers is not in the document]
KRVPerera: They are written in a way to pass an index value from and call each method from java. But yes we can refactor the whole file.  Shall we do it in a different PR.

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-fill-test.bal
KRVPerera: fixed https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/b76a0ffddd265c6a2226339c76801d5aff39ffb9

tests/jballerina-unit-test/src/test/resources/test-src/expressions/listconstructor/list_constructor.bal
```diff
@@ -35,3 +35,18 @@ function testListConstructorExpr() returns boolean {
         && fooArr[0] == "DDD"
         && fooArr[1] == 444;
 }
+
+function testListConstructorAutoFillExpr() {
+    int [8] arrOfEightInts = [1, 2, 3];
```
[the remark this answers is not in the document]
KRVPerera: Fixed https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/b76a0ffddd265c6a2226339c76801d5aff39ffb9

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArrayFillTestRuntime.java
KRVPerera: fixed https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/b76a0ffddd265c6a2226339c76801d5aff39ffb9

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/SealedArrayTest.java
KRVPerera: Fixed https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/b76a0ffddd265c6a2226339c76801d5aff39ffb9

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/SealedArrayTest.java
KRVPerera: Fixed https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/b76a0ffddd265c6a2226339c76801d5aff39ffb9

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/SealedArrayTest.java
KRVPerera: Fixed https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/ab2afb7227959b618fcd4d99c0854f2ca5b1dd40

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: fixed test case added 
https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/7d525fc1a7d35deb2cffeed2a9d4e3c5022d5b8b

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/sealed_array_listexpr_negative.bal
KRVPerera: Fixed by https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/a40950b3d01179e69e9b8985cfb7bfcf277b393a

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/sealed_array_listexpr_negative.bal
KRVPerera: Fixed

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-test.bal
KRVPerera: Fixed https://github.com/ballerina-platform/ballerina-lang/pull/20749/commits/6469ab1a9367249f3dae2e65642ec135bcc29c26#diff-73b31fb7a29b74dcbbff2be6c403b505R216

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: If the integer is '5' the valid fill value should be '0' for an int. Need to get the default value or implicit default value for int type which is zero. 
I have added a negative test case for this.
Test Case:  https://github.com/ballerina-platform/ballerina-lang/pull/20749/files#diff-65f99a2264d6173185afec0d71e151c6R63
Error Message: https://github.com/ballerina-platform/ballerina-lang/pull/20749/files#diff-7319ccb85e364b3233fb96f11c424f29R212

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
```diff
@@ -808,8 +813,9 @@ protected void fillValues(int index) {
             case TypeTags.BOOLEAN_TAG:
                 return;
             default:
-                Arrays.fill(refValues, size, index, elementType.getZeroValue());
-
+                if (arrayType.hasFillerValue()) {
```
MaryamZi: Do we need to check this here?

Isn't `fillValues` always called after `fillerValueCheck`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -109,6 +110,35 @@
     private Names names;
     private int finiteTypeCount = 0;
 
+    private boolean checkFillerValue(BUnionType type) {
```
MaryamZi: We seem to have methods with different names`checkFillerValue`, `hasFillerValue`, etc. across the class for similar checks. Shall we revisit these and group them together?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2553,4 +2583,182 @@ public int hashCode() {
...
+    private String getDefaultFillValue(BFiniteType finiteType) {
+        if (finiteType.getValueSpace().size() == 1) {
+            for (BLangExpression valueLiteral : finiteType.getValueSpace()) {
+                if (!(valueLiteral instanceof BLangLiteral)) {
+                    return DefaultValues.UNKNOWN.getValue();
+                }
+                return valueLiteral.toString();
+            }
+        }
+        return DefaultValues.UNKNOWN.getValue();
```
MaryamZi: What about when the value space has more than one value but are all of the same type?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
KRVPerera: This is not true for code coming from desugar.
As an example.

```
     AbstractPersonObject[5][] y = [];
     AbstractPersonObject e1 = new Employee("John", "Doe");
     y[0] = [e1];
     AbstractPersonObject[][5] r = [];
     r[0] = [e1, e1, e1, e1, e1];
```
```r[0] = [e1, e1, e1, e1, e1]``` expands by first creating a sealed array. Then adding values.
```r[0] => 
AbstractPersonObject[5] = [] 
```
 This is not allowed we cannot fill this by 5 AbstractPersonObject s. If user write this code it will give a compiler error . But since this is tool generated implementation we cannot error out this. 
```fillerValueCheck``` error out only if there is a need for filling. Consecutive inserts should not fail too which is guranteed by ```index > size``` condition. Since this array size is already 5 as generated by desugar ```fillerValueCheck``` at runtime will not give an error.

File: compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java (hunk 42 above)
Comment (KRVPerera) [reply compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java]: hasFillerValue is the public method which calls checkfillerValue helper methods sometime recursively. I will try to refactor.

File: compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java (hunk 29 above)
Comment (KRVPerera) [reply compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java]: Note: The code is written similar to run-time check. https://github.com/ballerina-platform/ballerina-lang/blob/7e0d4677ed81725d69b1a01c52df2a47333589dd/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java#L1787

The plan is to replace string enum with objects and also in accordance with spec. This may require calling initialiser methods to create zero value objects for every type.

File: compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BFiniteType.java [hunk 44]
```diff
@@ -111,4 +114,11 @@ public boolean isAnydata() {
         this.isAnyData = Optional.of(true);
         return true;
     }
+
+    public void addValue(BLangExpression value) {
+        this.valueSpace.add(value);
+        if (!nullable && (value.type.tag == TypeTags.NIL)) {
```
KRVPerera: isNullable call

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2553,4 +2565,150 @@ public int hashCode() {
...
+            return true;
+        }
+        Iterator<BType> iterator = type.getMemberTypes().iterator();
+        BType firstMember = iterator.next();
+        while (iterator.hasNext()) {
+            if (!isSameType(firstMember, iterator.next())) {
+                return false;
+            }
+        }
+        return isValueType(firstMember) && hasFillerValue(firstMember);
```
MaryamZi: The following should work right?

```ballerina
const FOO = "foo";
const BAR = "bar";

type FooBar FOO|BAR;

public function main() {
    (map<FooBar>|map<string>)[2] x = [];
}
```

I guess it won't work with the `isValueType` check?

hasithaa: Shall we create an issue for this?

MaryamZi: Created https://github.com/ballerina-platform/ballerina-lang/issues/21251

PR #20770, opened 2020-01-28
Align the function and object method related grammar with the spec

### LINKED ISSUE #20784: The grammar for functions in the implementation has deviated from the spec
How functions and methods are defined in the grammar of the implementation is different from how it is defined in the spec. Need to align the grammar in the implementation with the spec. Otherwise it'll get harder to maintain and add new features related to functions in the future.

pubudu91:
Fixes #20784

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangPackageBuilder.java
```diff
@@ -1666,9 +1656,11 @@ void createTrapExpr(DiagnosticPos pos, Set<Whitespace> ws) {
         addExpressionNode(trapExpr);
     }
 
-    void endFunctionDef(DiagnosticPos pos, Set<Whitespace> ws, boolean publicFunc, boolean remoteFunc,
-                        boolean nativeFunc, boolean privateFunc, boolean bodyExists, boolean isLambda) {
+    void endFunctionDefinition(DiagnosticPos pos, Set<Whitespace> ws, String funcName, DiagnosticPos funcNamePos,
+                               boolean publicFunc, boolean remoteFunc, boolean nativeFunc, boolean privateFunc,
+                               boolean bodyExists, boolean isLambda) {
```
irshadnilam: From what i know requiring this many parameters in a function can be considered as a bad practice. IMO it is better if we can think of an alternative approach.

WDYT?

irshadnilam: Maybe not for the immediate release, but in the long term :) 
I feel like there is an object hiding there (FunctionType?) instead of those booleans.

pubudu91: Yeah, I agree. It doesn't look right. Each time some new property gets added, you most likely have to add a new param to the relevant function the package builder as well. Since we'll be moving to a new parser, not sure whether this package builder will be relevant when that happens. So don't see much point in putting a lot of effort to refactor this.

pubudu91:
Closing this in favour of #20910

PR #20797, opened 2020-01-30
Improve final executable jar creation task

NipunaRanasinghe:
## Purpose
This PR adds improvements related to executable jar creation process, which reduces the final uber jar creation time by more than ~80% in average. 

Possibly fixes #20438.

## Approach
This PR uses `apache.commons.compress.archivers.zip` library APIs which is capable of copying the compressed data as it is, when constructing final executable uber jar by merging the thin jar with all the dependency jars.

## Remarks
Average uber jar creation execution time for time for few known samples are listed below.

| Sample Scenario | before (ms) | after (ms) |
| --------- |:---------:|:------------:|
| Hello World | ~1600 | ~350 |
| Hello World Service | ~4600 | ~650 |

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CreateExecutableTask.java
+++ b/cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CreateExecutableTask.java
@@ -29,10 +29,10 @@ import org.ballerinalang.packerina.buildcontext.sourcecontext.SingleModuleContex
 import org.wso2.ballerinalang.compiler.tree.BLangPackage;
 import org.wso2.ballerinalang.util.Lists;
 
+import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CreateExecutableTask.java
```diff
@@ -111,77 +90,88 @@ public void execute(BuildContext buildContext) {
...
+            try {
+                String entryName = entry.getName();
                 if (entryName.startsWith("META-INF/services")) {
                     StringBuilder s = services.get(entryName);
                     if (s == null) {
                         s = new StringBuilder();
                         services.put(entryName, s);
                     }
                     char c = '\n';
+                    InputStream inStream = zipFile.getInputStream(entry);
```
SupunS: shall we wrap this with a BufferedInputStream as well?

NipunaRanasinghe: I have already tried wrapping input streams with the buffered ones, but couldn't had any noticeable gain. The reason might be we have to access input streams only for very few entries (only for the entries under `META-INF/services`). Anyway let have this included. :)

PR #20846, opened 2020-02-03
Migrate codegen to use BString

vinok88:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/string/StringValueBasicsTest.java
```diff
@@ -69,6 +70,13 @@ public void testError() {
         Assert.assertEquals(((BInteger) returns[0]).intValue(), 5);
     }
 
+    @Test
+    public void testArrayStore() {
+        BValue[] returns = BRunUtil.invoke(result, "testArrayStore");
+        Assert.assertEquals(returns[0].getClass(), BValueArray.class);
+        Assert.assertEquals(((BValueArray) returns[0]).getStringArray()[0], "h😀llo");
```
riyafa: We assert the length because that is the main difference from the old String. If we assert by value both old and new cases pass. Shall we return the length and assert?

tests/jballerina-unit-test/src/test/resources/test-src/types/string/string-value-test.bal
```diff
@@ -20,3 +20,9 @@ function testError() returns int {
     error err = error(smiley);
     return err.reason().length();
 }
+
+function testArrayStore() returns string[] {
+    string[] arr = [];
+    arr[0] = "h😀llo";
+    return arr;
```
riyafa: Shall we return the length?

tests/jballerina-unit-test/src/test/resources/test-src/types/string/string-value-test.bal
```diff
@@ -20,3 +20,9 @@ function testError() returns int {
     error err = error(smiley);
     return err.reason().length();
 }
+
+function testArrayStore() returns string[] {
```
riyafa: Shall we also add test for a multi element array also, just to be on the safe side?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/StringUtils.java
```diff
@@ -133,4 +133,8 @@ public static StringValue fromString(String s) {
         return new NonBmpStringValue(s, highSurrogatesArr);
     }
 
+    public static BString fromString(BString s) {
```
manuranga: `fromString` is used to convert java string to ballerina. do we need a way to convert BString to BString?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/BmpStringValue.java
```diff
@@ -46,9 +48,9 @@ public int length() {
      }
 
      @Override
-     public StringValue concat(StringValue str) {
+     public BString concat(BString str) {
          if (str instanceof BmpStringValue) {
-             return new BmpStringValue(this.value + ((BmpStringValue) str).value);
+             return new BmpStringValue((this.value + ((BmpStringValue) str).value));
```
manuranga: extra paran

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/javainterop/RefTypeTests.java
```diff
@@ -388,7 +387,7 @@ public static HandleValue getHandle() {
         return new HandleValue(m);
     }
 
-    public static StringValue useHandle(HandleValue h) {
+    public static org.ballerinalang.jvm.values.api.BString useHandle(HandleValue h) {
```
manuranga: need the long name ?

vinok88: yes. org.ballerinalang.model.values.BString also imported here for assertions.

manuranga: got it.

PR #20847, opened 2020-02-03
Handle Ballereina reserved literal name when generating code from proto

### LINKED ISSUE #20837: [gRPC] Invalid field name in generated stub file.
**Description:**
If we have proto file definition as below,
```proto
message test {
  string json = 1;
}
```
This will generate the ballerina record as follows,
```ballerina
type test record {
    string json;
};
```
Since json is special type in ballerina we need to generate the code like below,
```ballerina
type test record {
    string 'json;
};
```

daneshk:
## Purpose
Fix the issue when generating the code from the proto definition which has ballerina reserved literal name as the field names.

Fixes #20837

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/builder/components/Field.java
+++ b/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/builder/components/Field.java
@@ -85,7 +85,7 @@ public class Field {
                 fieldDefaultValue = fieldLabel;
             }
             String fieldName = fieldDescriptor.getName();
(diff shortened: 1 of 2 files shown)


daneshk:
There is no java API to retrieve Ballerina reserved literal names. So I hardcoded all words for now. Since those are subjected to change, this is not the ideal solution. I created issue #20856, requesting an API.

PR #20854, opened 2020-02-05
Handle Ballerina reserved literal name when generating code from proto(1.1.x)

### LINKED ISSUE #20837: [gRPC] Invalid field name in generated stub file.

daneshk:
## Purpose
Fix the issue when generating the code from the proto definition which has ballerina reserved literal name as the field names.

Fixes #20837

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/builder/components/Field.java
+++ b/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/builder/components/Field.java
@@ -85,7 +85,7 @@ public class Field {
                 fieldDefaultValue = fieldLabel;
             }
             String fieldName = fieldDescriptor.getName();
-            if (Arrays.stream(BallerinaLexer.ruleNames).anyMatch(fieldName::equalsIgnoreCase) || Names.ERROR.value
+            if (Arrays.stream(RESERVED_LITERAL_NAMES).anyMatch(fieldName::equalsIgnoreCase) || Names.ERROR.value
                     .equalsIgnoreCase(fieldName)) {
                 fieldName = "'" + fieldName;
             }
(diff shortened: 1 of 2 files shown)


daneshk:
There is no java API to retrieve Ballerina reserved literal names. So I hardcoded all words for now. Since those are subjected to change, this is not the ideal solution. I created issue #20856, requesting an API.

stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/builder/components/Field.java
```diff
@@ -141,4 +140,17 @@ private Builder(DescriptorProtos.FieldDescriptorProto fieldDescriptor) {
...
+            "record", "annotation", "parameter", "transformer", "worker", "listener", "remote", "xmlns", "returns",
+            "version", "channel", "abstract", "client", "const", "typeof", "source", "from", "on", "group", "by",
+            "having", "order", "where", "followed", "for", "window", "every", "within", "snapshot", "inner", "outer",
+            "right", "left", "full", "unidirectional", "forever", "limit", "ascending", "descending", "int", "byte",
+            "float", "decimal", "boolean", "string", "error", "map", "json", "xml", "table", "stream", "any",
+            "typedesc", "type", "future", "anydata", "handle", "var", "new", "__init", "if", "match", "else",
+            "foreach", "while", "continue", "break", "fork", "join", "some", "all", "try", "catch", "finally", "throw",
+            "panic", "trap", "return", "transaction", "abort", "retry", "onretry", "retries", "committed", "aborted",
+            "with", "in", "lock", "untaint", "start", "but", "check", "checkpanic", "primarykey", "is", "flush",
+            "wait", "default"};
```
anupama-pathirage: Isn't it difficult to maintain this in the future as we have to change this array whenever new reserved word is adding to the language? There is a possibility to miss this.

anupama-pathirage: Sorry for the noise. Just noticed your comment on requeesting an API.

Issue #20727, opened 2020-01-24
Implement functions to stream type


mohanvive:
Below functions needs to be implemented,
- [x] map
- [x] foreach
- [x] filter
- [ ] reduce
- [ ] iterator 
- [x] next

PR #20889, opened 2020-02-06
Add streamlib reduce, foreach and iterator functionalities

gimantha:
With this PR, reduce, iterator and forEach functions will be introduced to stream type as a part of the solution to #20727.


gimantha:
@hasithaa @mohanvive Shall we merge this first and then create separate issues to implement these functions in ballerina. Then we can work on them in the next sprint. @pcnfernando's tasks depends on these functions

mohanvive:
@gimantha +1 to proceed from me.

PR #20959, opened 2020-02-10
Fix type checker to resolve expected type from select clause

mohanvive:


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2614,7 +2614,14 @@ public void visit(BLangQueryExpr queryExpr) {
         for (WhereClauseNode whereClauseNode : whereClauseList) {
             whereEnv = typeCheckWhereClause((BLangWhereClause) whereClauseNode, selectClause, parentEnv);
         }
-        checkExpr(selectClause.expression, whereEnv);
+
+        if (queryExpr.expectedType.tag != symTable.noType.tag) {
```
MaryamZi: I guess we can use the `expType` instead.
```suggestion
        if (expType.tag != symTable.noType.tag) {
```

MaryamZi: Shall we check 
```suggestion
        if (expType.tag == TypeTags.ARRAY) {
```
instead?

Issue #20987, opened 2020-02-11
[gRPC] Optional fields are not deserialization properly


daneshk:
**Description:**
$subject.
In proto3 specification, we don't need to set values to all the fields in the message. when we don't set the values, default values should set when deserializing.

**Steps to reproduce:**

**Affected Versions:**
1.1.1

Issue #20888, opened 2020-02-06
Failed to publish metrics to Ballerina gRPC server


nadundesilva:
**Description:**
gRPC server implemented using the following message and rpc call fails to receive data.

```proto
service MetricsServer {
    rpc publishMetrics (MetricsPublishRequest)
    		returns (google.protobuf.Empty);
}

message MetricsPublishRequest {
    string userId = 1;
    repeated Metric metrics = 2;
}

message Metric {
    int64 timestamp = 1;
    string name = 2;
    double value = 3;
    map<string, string> tags = 4;
}
```
The following error was observed in the gRPC client. (The client was implemented using Java)
```
UNKNOWN: HTTP status code 499
invalid content-type: text/plain
headers: Metadata(:status=499,content-type=text/plain,grpc-status=CANCELLED,grpc-message=Failed to dispatch inbound message. null,server=ballerina,date=Thu, 6 Feb 2020 12:30:07 +0530)
DATA-----------------------------
failed to dispatch inbound message. null
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:240)
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:221)
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:140)
	at org.ballerinalang.observe.trace.extension.choreo.gen.NegotiatorGrpc$NegotiatorBlockingStub.publishMetrics(NegotiatorGrpc.java:282)
	at org.ballerinalang.observe.trace.extension.choreo.client.ChoreoClient.publishMetrics(ChoreoClient.java:53)
	at org.ballerinalang.observe.trace.extension.choreo.MetricsReporterExtension$Task.run(MetricsReporterExtension.java:120)
	at java.util.TimerThread.mainLoop(Timer.java:555)
	at java.util.TimerThread.run(Timer.java:505)
```

**Steps to reproduce:**
1. Generate a ballerina gRPC service using the above proto.
2. Start the server and publish data using any gRPC client.

**Affected Versions:**
Ballerina 1.1.1

**Suggested Assignees (optional):**
@daneshk

PR #20989, opened 2020-02-11
Fix grpc map field serialization issue and optional field deserialization issue(master)

daneshk:
## Purpose
Fix grpc map field serialization issue and an optional field deserialization issue

Resolves: #20888
Resolves: #20987

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-integration-test/src/test/resources/grpc/src/clients/19_grpc_map_field_client.bal
```diff
@@ -0,0 +1,168 @@
...
+    io:println(testMapFields());
+    io:println(testOptionalFields());
+}
+
+public function testMapFields() returns string {
+    MetricsPublishRequest request = {id: "xxxxx", metrics: [{timestamp:1580966325916,
+    name:"ballerina/http/Caller_3XX_requests_total_count",value:0.0,tags:[{key:"action", value:"respond"}]}]};
+    grpc:Headers|error publishMetrics = blockingEp->publishMetrics(request);
+    if (publishMetrics is error) {
+        return io:sprintf("Metrics publish failed: %s - %s", publishMetrics.reason(), <string> publishMetrics.detail()["message"]);
```
anupama-pathirage: Shall we stick to the 120 max line width? Check all the places.

tests/jballerina-integration-test/src/test/resources/grpc/src/grpcservices/19_grpc_map_service.bal
```diff
@@ -0,0 +1,143 @@
...
+public type MetricsPublishRequest record {|
+    string id = "";
+    Metric[] metrics = [];
+|};
+
+public type Metric record {|
+    int timestamp = 0;
+    string name = "";
+    float value = 0.0;
+    record {| string key; string value; |}[] tags = [];
```
anupama-pathirage:
```suggestion
    record {|string key; string value;|}[] tags = [];
```

anupama-pathirage: Shall we reformat the .bal files?

stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/nativeimpl/client/FunctionUtils.java
```diff
@@ -248,18 +248,22 @@ public static Object externBlockingExecute(ObjectValue clientEndpoint, String me
...
+                    dataContext = new DataContext(Scheduler.getStrand(),
                             new NonBlockingCallback(Scheduler.getStrand()));
                     blockingStub.executeUnary(requestMsg, methodDescriptors.get(methodName), dataContext);
                 } else {
                     return notifyErrorReply(INTERNAL, "Error while executing the client call. Method type " +
                             methodType.name() + " not supported");
                 }
             } catch (Exception e) {
+                if (dataContext != null) {
+                    dataContext.getCallback().notifyFailure(MessageUtils.getConnectorError(e));
```
chamil321: What is the special reason to notify the error to callback and return error subsequently?
If the exception has occurred before the non-blocking call, then call back should not be called ideally. We can simply unblock strand[1] and return error. Problem is if we notify the callback when the non-blocking task is not processed, that can cause strand state issues.

[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/http/src/main/java/org/ballerinalang/net/http/nativeimpl/connection/Respond.java#L122

daneshk: We need to release the strand when there is an error before doing the network call. Thanks for the suggestion

daneshk: Fixed in 25dc241

PR #21043, opened 2020-02-13
Fix ballerina test failing due to missing tests directory 

warunalakshitha:
## Purpose
$Subject


anupama-pathirage:
@warunalakshitha Don't we have test cases for this scenario? How can we avoid such regressions in future?

warunalakshitha:
@anupama-pathirage We already have a testcase which was disabled. we are working on enabling them.

PR #21335, opened 2020-02-27
Send vscode client configuration during initialization

### LINKED ISSUE #21324: Initial IDE Config Sync Call not working in LS
**Description:**
Initial Config Sync Call not working in LS. Expected behavior is soon after Language Server gets the initialization request, initial configurations need to be sync using `workspace/didChangeConfiguration` notification.

**Affected Versions:**
1.2.0-SNAPSHOT

rasika:
Fixes #21324

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tool-plugins/vscode/src/extension.ts
+++ b/tool-plugins/vscode/src/extension.ts
@@ -17,7 +17,7 @@
  * under the License.
  *
  */
-import { ExtensionContext, commands, window } from 'vscode';
+import { ExtensionContext, commands, window, workspace, Location, Uri } from 'vscode';
 import { ballerinaExtInstance } from './core';
 import { activate as activateAPIEditor } from './api-editor';
 // import { activate as activateDiagram } from './diagram'; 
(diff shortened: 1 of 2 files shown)


tool-plugins/vscode/src/extension.ts
```diff
@@ -91,6 +91,21 @@ export function activate(context: ExtensionContext): Promise<any> {
...
         // Enable Ballerina Syntax Highlighter
         activateSyntaxHighlighter(ballerinaExtInstance);
+
+        ballerinaExtInstance.onReady().then(() => {
+            const langClient = <ExtendedLangClient>ballerinaExtInstance.langClient;
+            // Send initial configuration 
+            const args: DidChangeConfigurationParams = {
+                settings: workspace.getConfiguration('ballerina'),
+            };
+            langClient.sendNotification("workspace/didChangeConfiguration", args);
```
nadeeshaan: I think this is to capture the initial configuration at the server-side? It seems that we are violating the LSP by sending the didChangeConfiguration notification through the extension bypassing the protocol. I think what we should do to fetch the initial configuration is to use the configuration request at the server-side [1].
[1] https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#workspace_configuration

rasika: +1, initially I was reluctant to add this `workspace/configuration` request since some clients might not have implemented it. If we stressing down on the protocol let's strict with it, I'll add a fix.

nadeeshaan: > +1, initially I was reluctant to add this `workspace/configuration` request since some clients might not have implemented it. If we stressing down on the protocol let's strict with it, I'll add a fix.

Awesome :D

rasika: As discussed since we are having troubles in getting the result of the CompletableFuture, will do this as an improvement. Created https://github.com/ballerina-platform/ballerina-lang/issues/21390 to track this.

PR #21371, opened 2020-02-29
Fix MainFunctionsTest

SupunS:


Kishanthan:
Looks like @manuranga has pushed the same change

PR #21375, opened 2020-02-29
Add toArray() method to map lang library

pubudu91:
## Samples
```ballerina
// Getting the values as an array
map<int> ints = {"one": 1, "two": 2, "three": 3, "four": 4, "five": 5};
int[] intArr = ints.toArray();

// Getting record field values as an array
type Foo record {|
    string name;
    int age;
    float weight;
    decimal height;
    boolean isStudent;
|};

Foo foo = {
        name: "John Doe",
        age: 25,
        weight: 65.5,
        height: 172.3,
        isStudent: true
    };

(string|int|float|decimal|boolean)[] arr = foo.toArray();
```
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/langlib/lang.map/src/main/java/org/ballerinalang/langlib/map/ToArray.java
+++ b/langlib/lang.map/src/main/java/org/ballerinalang/langlib/map/ToArray.java
@@ -1,0 +1,107 @@
+/*
+ *  Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  WSO2 Inc. licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package org.ballerinalang.langlib.map;
+
+import org.ballerinalang.jvm.scheduling.Strand;
+import org.ballerinalang.jvm.types.BArrayType;
+import org.ballerinalang.jvm.types.BMapType;
+import org.ballerinalang.jvm.types.BRecordType;
+import org.ballerinalang.jvm.types.BType;
+import org.ballerinalang.jvm.types.TypeTags;
+import org.ballerinalang.jvm.values.ArrayValue;
+import org.ballerinalang.jvm.values.ArrayValueImpl;
+import org.ballerinalang.jvm.values.MapValue;
+import org.ballerinalang.langlib.map.util.MapLibUtils;
+import org.ballerinalang.model.types.TypeKind;
+import org.ballerinalang.natives.annotations.Argument;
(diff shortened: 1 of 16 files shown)


langlib/lang.map/src/main/java/org/ballerinalang/langlib/map/ToArray.java
```diff
@@ -0,0 +1,107 @@
...
+@BallerinaFunction(
+        orgName = "ballerina", packageName = "lang.map",
+        functionName = "toArray",
+        args = {@Argument(name = "m", type = TypeKind.MAP)},
+        returnType = {@ReturnType(type = TypeKind.ARRAY, elementType = TypeKind.ANY)},
+        isPublic = true
+)
+public class ToArray {
+
+    public static ArrayValue toArray(Strand strand, MapValue<?, ?> m) {
```
pramodya1994: Can we add a test for this method?

pubudu91: Added with 11bc310

langlib/langlib-test/src/test/resources/test-src/maplib_test.bal
```diff
@@ -73,3 +73,140 @@ function testReduce() returns float {
...
+function testMapOfUnionToArray() {
+    map<int|string|Person> m = {"i": 10, "s": "foo", "p": new Person("Pubudu")};
+    (int|string|Person)[] arr = m.toArray();
+
+    assert(10, <anydata>arr[0]);
+    assert("foo", <anydata>arr[1]);
+    assertSameRef(m["p"], arr[2]);
+}
+
+type Foo record {|
```
MaryamZi: Shall we also add a test for records where the record fields are all of a single simple basic type?

```ballerina
type Bar record {|
    byte a;
    byte b;
    byte...;
|};
```

This array can also be updated right? Shall we add a test or two to cover updating a value at an index and maybe adding new elements at the end?

langlib/langlib-test/src/test/resources/test-src/maplib_test.bal
```diff
@@ -73,3 +73,140 @@ function testReduce() returns float {
...
+    var arr = foo.toArray();
+
+    assert(<(string|int|float|decimal|boolean)[]>["John Doe", 25, 65.5, 172.3d, true, "Sri Lanka", 12500], arr);
+}
+
+
+// Util functions
+
+function assert(anydata expected, anydata actual) {
+    if (expected != actual) {
```
MaryamZi: Shall we return here after checking `==` instead?

langlib/langlib-test/src/test/resources/test-src/maplib_test.bal
```diff
@@ -73,3 +73,140 @@ function testReduce() returns float {
...
+        typedesc<anydata> expT = typeof expected;
+        typedesc<anydata> actT = typeof actual;
+        string reason = "expected [" + expected.toString() + "] of type [" + expT.toString()
+                            + "], but found [" + actual.toString() + "] of type [" + actT.toString() + "]";
+        error e = error(reason);
+        panic e;
```
MaryamZi:
```suggestion
        panic error(reason);
```

Just to be consistent with our usage of reasons, shall we use a simple string as the reason (e.g., "AssertionError") and then add this current reason as the message field of the detail?

Issue #14148, opened 2019-03-06
Write compiler plugin for resources without input parameters


ThisaruGuruge:
**Description:**
> Currently, compiler plugin identifies a service using the input parameters of the resource. But we need a way to write a compiler plugin to a service which does not have an input parameter for the resource.
> A service cannot be identified by the compiler plugin if it does not have an input parameter, as of now.

__UPDATE:__

This acts as a bug in Ballerina task library. 

Currently ballerina task services have only one resource, `onTrigger()` which has no parameters, by default. But a user can pass parameters as an attachment as follows.

```ballerina
import ballerina/kafka;
import ballerina/task;

kafka:ConsumerConfig consumerConfig = {
    bootstrapServers: "localhost:9100",
    groupId: "test-group",
    clientId: "basic-consumer",
    offsetReset: "earliest",
    topics: ["test"]
};

public function main() {
    kafka:Consumer kafkaConsumer = new(consumerConfig);
    task:Scheduler timer = new({ intervalInMillis: 1000 });
    var attachResult = timer.attach(timerService, attachment = kafkaConsumer);
    var startResult = timer.start(); 
}

service timerService = service {
    resource function onTrigger(kafka:Consumer kafkaConsumer) {
        // implementation
    }
}
```

This will throw a compilation error, as the compiler identifies this as Kafka service, because of the parameter of the resource function, and it does not comply with the Kafka resource signature. We should have a way to identify the resources without looking at the resource function parameters.

riyafa:
I think we need service typing for compile time validation: https://groups.google.com/forum/#!msg/ballerina-dev/AicMMCXL8kA/w9l1vNz1CAAJ

hasithaa:
This related new service typing proposal in SL. With that, we don't need this feature anymore.

PR #21410, opened 2020-03-02
Support multiple attachments in Ballerina task scheduler

### LINKED ISSUE #21268: Ballerina task scheduler does not support multiple attachments
**Description:**
Currently, task scheduler `attach` API supports only a single attachment [1]. It should support for multiple attachments.

[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/stdlib/task/src/main/ballerina/src/task/scheduler.bal#L30

ThisaruGuruge:
> Fixes #21268 

### Providing multiple arguments:
```ballerina
import ballerina/task;

public type Person record {
    string name;
    int age;
};

public type Account record {
    int number;
    float balance;
};

task:Scheduler timer = new({ intervalInMillis: 1000 });

public function main() {
    Person sam = { name: "Sam", age: 29 };
    Account acc = { number: 188008, balance: 1233.02 };
    task:TimerConfiguration timerConfiguration = { intervalInMillis: 1000 };
    task:Scheduler timer = new(timerConfiguration);
    var attachResult = timer.attach(TimerService, sam, acc);
    var startResult = timer.start();
}

service TimerService = service {
    resource function onTrigger(Person p, Account a) {
        // Handle trigger
    }
};
```

```ballerina
public type Person record {
    string name;
    int age;
};

task:Scheduler timer = new({ intervalInMillis: 1000 });

public function main() {
    Person p = { name: "Sam", age: 29 };
    var attachResult = timer.attach(TimerService, attachment = p); // This will break
    var startResult = timer.start();
}

service TimerService = service {
    resource function onTrigger(Person p) {
        // Handle trigger
    } 
};
```

To fix this, just remove the named argument. 

```ballerina
public type Person record {
    string name;
    int age;
};

task:Scheduler timer = new({ intervalInMillis: 1000 });

public function main() {
    Person p = { name: "Sam", age: 29 };
    var attachResult = timer.attach(TimerService, p); // Removed named argument
    var startResult = timer.start();
}

service TimerService = service {
    resource function onTrigger(Person p) {
        // Handle trigger
    } 
};
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/task/src/main/ballerina/src/task/scheduler.bal
+++ b/stdlib/task/src/main/ballerina/src/task/scheduler.bal
@@ -28,11 +28,10 @@ public type Scheduler object {
     # + attachments - Set of optional parameters which needs to passed inside the resources.
     # + return - Returns `task:SchedulerError` if the process failed due to any reason, nil otherwise.
     public function attach(service serviceToAttach, any... attachments) returns SchedulerError? {
-        string message = "Failed to attach the service to the scheduler";
         var result = attachExternal(self.taskListener, serviceToAttach, ...attachments);
         if (result is ListenerError) {
-            SchedulerError err = error(SCHEDULER_ERROR_REASON, message = message, cause = result);
-            return err;
+            string message = "Failed to attach the service to the scheduler";
+            return error(SCHEDULER_ERROR_REASON, message = message, cause = result);
         }
     }
 
@@ -44,8 +43,7 @@ public type Scheduler object {
         var result = detachExternal(self.taskListener, attachedService);
         if (result is ListenerError) {
             string message = "Scheduler failed to detach the service";
-            SchedulerError err = error(SCHEDULER_ERROR_REASON, message = message, cause = result);
-            return err;
+            return error(SCHEDULER_ERROR_REASON, message = message, cause = result);
         }
     }
 
@@ -56,8 +54,7 @@ public type Scheduler object {
         var result = startExternal(self.taskListener);
         if (result is ListenerError) {
             string message = "Scheduler failed to start";
-            SchedulerError err = error(SCHEDULER_ERROR_REASON, message = message, cause = result);
-            return err;
+            return error(SCHEDULER_ERROR_REASON, message = message, cause = result);
         }
     }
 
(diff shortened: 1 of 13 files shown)


stdlib/task/src/main/ballerina/src/task/scheduler.bal
```diff
@@ -25,23 +25,14 @@ public type Scheduler object {
...
-            }
-        } else {
-            var result = attachExternal(self.taskListener, serviceToAttach, {});
-            if (result is ListenerError) {
-                SchedulerError err = error(SCHEDULER_ERROR_REASON, message = message, cause = result);
-                return err;
-            }
+        var result = attachExternal(self.taskListener, serviceToAttach, ...attachments);
+        if (result is ListenerError) {
+            SchedulerError err = error(SCHEDULER_ERROR_REASON, message = message, cause = result);
```
wggihan: Shall we try to merge this to a single line?

ThisaruGuruge: Done

stdlib/task/src/main/java/org/ballerinalang/stdlib/task/utils/TaskExecutor.java
```diff
@@ -40,9 +38,17 @@ public static void executeFunction(ServiceInformation serviceInformation) {
     }
 
     private static Object[] getParameterList(AttachedFunction function, ServiceInformation serviceInformation) {
-        if (function.type.paramTypes.length > 0 && Objects.nonNull(serviceInformation.getAttachment())) {
-            return new Object[]{serviceInformation.getAttachment(), Boolean.TRUE};
+        Object[] attachments = serviceInformation.getAttachment();
+        int numberOfParameters = function.type.paramTypes.length;
+        Object[] parameters = null;
+        if (numberOfParameters == attachments.length) {
```
chamil321: How does this behave in the error cases? If the # of onTrigger resource param count and the # of attach() function params are different, null will be returned. 

If the above is validated during the compile-time, then this condition is not necessary. WDYT?

ThisaruGuruge: Ideally, this should be validated in the compile time, but unfortunately, we can't write compiler plugin for this, just yet. (See #14148)

Therefore this is added. But since this is invoked from a resource function, we cannot return any error here. Is there a way to reflect this error scenario?

PR #21476, opened 2020-03-04
Implement get generated keys only if supported by JDBC

### LINKED ISSUE #20795: Improve java.jdbc API to allow user to explicitly request to return generated keys
### Description:
According to JDBC standard, 
"The methods execute, executeUpdate, or Connection.prepareStatement accept an optional parameter, which can be used to indicate that any auto-generated values should be returned when the statement is executed or prepared." [[JDBC 4.1](https://download.oracle.com/otndocs/jcp/jdbc-4_1-mrel-spec/index.html), [JDBC 3.0](http://java.cnam.fr/iagl/biblio/spec/jdbc-3_0-fr-spec.pdf)]

**JDBC standard update methods**
```
execute(String sql, Statement.RETURN_GENERATED_KEYS) 
execute(String sql, int[] columnIndexes)
execute(String sql, String[] columnNames)

executeUpdate(String SQL, Statement.RETURN_GENERATED_KEYS)
executeUpdate(String SQL, int[] column indexes)
executeUpdate(String SQL, String[] column names)
```
Therefore to be fully JDBC compliant we should change below Ballerina java.jdbc API too. This way, Ballerina can let the user check if the driver and database he uses to support key generations or not and use appropriate APIs.

**Ballerina standard update methods**
```
testDB->batchUpdate
testDB->update
```

#### supportsGetGeneratedKeys
JDBC standard does not have an API that guarantees the user to get relevant information on what statements support retrieving generated keys or not. "[supportsGetGeneratedKeys](https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#supportsGetGeneratedKeys())" only ensures that a driver should at lease support key retrieval in the insert statement.

#### generatedKeyAlwaysReturned
In [JDBC 4.1](https://download.oracle.com/otndocs/jcp/jdbc-4_1-mrel-spec/index.html) they have introduced a new DatabaseMetaData method.
**generatedKeyAlwaysReturned** — Method used to indicate whether a generated key will always be returned if the column name(s) or index(es) specified for the auto-generated key column(s) are valid and the statement succeeds."

This will help drivers that support key retrieval for all the statements like [IBM](https://www.ibm.com/support/knowledgecenter/SSEPGG_10.5.0/com.ibm.db2.luw.apdv.java.doc/src/tpc/imjcc_t0057054.html). Currently, we have _INSERT, UPDATE, DELETE, MERGE_ like a statement but in two flavors. Regular single statement flavor and batch flavors to be worried. There will be new statements, and the adaptation of new API by the vendors will be different. Therefore the new JDBC API will also not solve this issue for Ballerina like intermediates. 

### Steps to reproduce
Since the master branch does not have guard by 'supportsGetGeneratedKeys' branch can be tested with HANA DB with any query. #20709 
We can also experiment with an alternate JDBC driver, which returns true for "supportsGetGeneratedKeys" API but does not support the batch statement. Currently, we are filtering Oracle and MSSQL by checking database names.

**Affected Versions:**
all version

OracleDB, MSSQL, HANA DB

#20673

Area/StandardLibs Type/Improvement Type/Proposal

kalaiyarasiganeshalingam:
## Purpose
$Subject

Fixes #20795

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/jdbc/src/main/ballerina/src/java.jdbc/types.bal
```diff
@@ -25,12 +25,15 @@ import ballerinax/java;
...
 #               provided in poolOptions
+# + getGeneratedKeys - Allow to retrieves auto-generated keys after a statement has been executed if value is `true`.
+#                      It will return auto-generated keys only if supported by JDBC
 public type ClientConfiguration record {|
     string url = "";
     string username = "";
     string password = "";
     PoolOptions poolOptions?;
     map<anydata> dbOptions = {};
+    boolean getGeneratedKeys = false;
```
anupama-pathirage: Why this is a property in ClientConfiguration level? Can't we make it a optional parameter in update action only?

anupama-pathirage:
@kalaiyarasiganeshalingam Shall we add some negative tests as well where the new parameter is false and user trying to read the generated keys from the return value etc?

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,16 @@ public type Client client object {
     #
     # + sqlQuery - SQL statement to execute
     # + parameters - The parameters to be passed to the update query
+    # + getGeneratedKeys - Allow to retrieves auto-generated keys after a statement has been executed if value is
```
anupama-pathirage:
```suggestion
    # + getGeneratedKeys - Allow to retrieves auto-generated keys after a statement has been executed if the value is
```

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,16 @@ public type Client client object {
...
+    #                      `true`.It will return auto-generated keys only if supported by JDBC
```
anupama-pathirage:
```suggestion
    #                      `true`. If it is not supported by the used JDBC driver, still it will return `nil`
```

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +80,21 @@ public type Client client object {
     #           may or may not continue to process the remaining commands in the batch. This property can be
     #           used to override this behavior. When it is set to true, if there is a failure in a few commands and
     #           the JDBC driver continues with the remaining commands, the successfully executed commands in the batch
-    #           also will get rolled back.
+    #           also will get rolled back
+    # + getGeneratedKeys - Allow to retrieves auto-generated keys after a statement has been executed if value is
```
anupama-pathirage: please change similar to above.

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +80,21 @@ public type Client client object {
...
+    #                      `true`.It will return auto-generated keys only if supported by JDBC
     # + return - A `BatchUpdateResult` with the updated row count and returned error if any. If all the commands
     #            in the batch have executed successfully, the error will be `nil`. If one or more commands have failed,
     #            the `returnedError` field will give the corresponding `Error` along with the int[] which
-    #            contains updated row count or the status returned from each command in the batch.
+    #            contains updated row count or the status returned from each command in the batch
     public remote function batchUpdate(@untainted string sqlQuery, boolean rollbackAllInFailure,
-                                       Param?[]... parameters)
+                                       boolean? getGeneratedKeys = (), Param?[]... parameters)
```
anupama-pathirage: It seems the parameters are not consistent in this remote function now. The `rollbackAllInFailure` is a `boolean` while the next similar parameter `getGeneratedKeys` is `boolean?`

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,16 @@ public type Client client object {
     #
     # + sqlQuery - SQL statement to execute
     # + parameters - The parameters to be passed to the update query
+    # + getGeneratedKeys - Allow to retrieves auto-generated keys after a statement has been executed if value is
+    #                      `true`.It will return auto-generated keys only if supported by JDBC
     # + return - `UpdateResult` with the updated row count and key column values,
     #             else `Error` will be returned if there is an error
-    public remote function update(@untainted string sqlQuery, Param... parameters)
-                                  returns UpdateResult|Error {
+    public remote function update(@untainted string sqlQuery, boolean? getGeneratedKeys = (),
```
anupama-pathirage: IMO, instead of `boolean?` we should use `boolean` with default value. So for simple update statements which doesn't have parameters, we can specify the `sqlQuery` only

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,16 @@ public type Client client object {
...
+                                  Param... parameters) returns UpdateResult|Error {
         if (!self.clientActive) {
             return self.handleStoppedClientInvocation();
         }
-        return nativeUpdate(self, java:fromString(sqlQuery), parameters);
+        return nativeUpdate(self, java:fromString(sqlQuery), parameters, getGeneratedKeys);
```
anupama-pathirage: Shall we swap `parameters` and `getGeneratedKeys`? Bascially have the same order in the remote function it self.

stdlib/jdbc/src/main/ballerina/src/java.jdbc/types.bal
```diff
@@ -232,7 +232,7 @@ public type UpdateResult record {|
 # + returnedError - The `Error` returned from the remote function in case of a failure
 public type BatchUpdateResult record {|
     int[] updatedRowCount;
-    map<anydata[]> generatedKeys;
+    map<anydata[]>? generatedKeys;
```
anupama-pathirage: Why only this map is mar as nillable? Why we are not being consistent with `UpdateResult` record?

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/methods/ExternFunctions.java
```diff
@@ -72,9 +74,9 @@ public static void createClient(ObjectValue client, MapValue<String, Object> con
         String dbType = url.split(":")[1].toUpperCase(Locale.getDefault());
 
         SQLDatasource.SQLDatasourceParamsBuilder builder = new SQLDatasource.SQLDatasourceParamsBuilder(dbType);
-        SQLDatasource.SQLDatasourceParams sqlDatasourceParams = builder.withPoolOptions(poolOptionsWrapper)
-                .withJdbcUrl(url).withUsername(username).withPassword(password).withDbOptionsMap(dbOptions)
-                .withIsGlobalDatasource(userProvidedPoolOptionsNotPresent).build();
+        SQLDatasource.SQLDatasourceParams sqlDatasourceParams = builder.withPoolOptions(poolOptionsWrapper).
+                withJdbcUrl(url).withUsername(username).withPassword(password).withDbOptionsMap(dbOptions).
+                withIsGlobalDatasource(userProvidedPoolOptionsNotPresent).build();
```
anupama-pathirage: What is the reason for appending `.` at the end instead of the beginning of the line?

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
kalaiyarasiganeshalingam: We  can't swap as required parameter not allowed after default-able parameters.
ERROR: ballerinax/java.jdbc:0.0.0::client.bal:132:24:: required parameter not allowed after defaultable parameters

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,16 @@ public type Client client object {
     #
     # + sqlQuery - SQL statement to execute
     # + parameters - The parameters to be passed to the update query
+    # + getGeneratedKeys - Allow to retrieves auto-generated keys after a statement has been executed if the value is
+    #                      `true`. If it is not supported by the used JDBC driver, still it will return `nil`.
     # + return - `UpdateResult` with the updated row count and key column values,
     #             else `Error` will be returned if there is an error
-    public remote function update(@untainted string sqlQuery, Param... parameters)
-                                  returns UpdateResult|Error {
+    public remote function update(@untainted string sqlQuery, boolean getGeneratedKeys = false, Param... parameters)
```
daneshk: IMO, a better name for this variable is `returnGeneratedKeys`

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
daneshk: We don't need to make the parameter defaultable in internal function. because we always have a value for that parameter.

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +80,21 @@ public type Client client object {
...
+    #                      `true`. If it is not supported by the used JDBC driver, still it will return `nil`.
     # + return - A `BatchUpdateResult` with the updated row count and returned error if any. If all the commands
     #            in the batch have executed successfully, the error will be `nil`. If one or more commands have failed,
     #            the `returnedError` field will give the corresponding `Error` along with the int[] which
-    #            contains updated row count or the status returned from each command in the batch.
-    public remote function batchUpdate(@untainted string sqlQuery, boolean rollbackAllInFailure,
-                                       Param?[]... parameters)
+    #            contains updated row count or the status returned from each command in the batch
+    public remote function batchUpdate(@untainted string sqlQuery, boolean rollbackAllInFailure = false,
+                                       boolean getGeneratedKeys = false, Param?[]... parameters)
```
daneshk: Same as above. I think a better name for this variable is `returnGeneratedKeys`

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +80,21 @@ public type Client client object {
     #           may or may not continue to process the remaining commands in the batch. This property can be
     #           used to override this behavior. When it is set to true, if there is a failure in a few commands and
     #           the JDBC driver continues with the remaining commands, the successfully executed commands in the batch
-    #           also will get rolled back.
+    #           also will get rolled back
+    # + getGeneratedKeys - Allow to retrieves auto-generated keys after a statement has been executed if the value is
```
daneshk:
```suggestion
    # + getGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is set to true, auto-generated keys are made available to retrieve when the statement is executed. If the underline JDBC driver does not support it, still it will return `()` value.
```

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +80,21 @@ public type Client client object {
...
+    #                      `true`. If it is not supported by the used JDBC driver, still it will return `nil`.
```
daneshk: # + getGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is set to true, auto-generated keys are made available to retrieve when the statement is executed. If the underline JDBC driver does not support it, still it will return `()` value.

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/datasource/SQLDatasource.java
```diff
@@ -50,6 +51,7 @@
     private AtomicInteger clientCounter = new AtomicInteger(0);
     private Lock mutex = new ReentrantLock();
     private boolean poolShutdown = false;
+    private boolean supportsGetGeneratedKeys = false;
```
daneshk:
```suggestion
    private boolean supportGeneratedKeys = false;
```

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/UpdateStatement.java
```diff
@@ -74,24 +78,26 @@ public Object execute() {
             ArrayValue generatedParams = constructParameters(parameters);
             conn = getDatabaseConnection(strand, client, datasource);
             String processedQuery = createProcessedQueryString(query, generatedParams);
-            stmt = conn.prepareStatement(processedQuery, Statement.RETURN_GENERATED_KEYS);
+
+            boolean keyRetrievalSupportedStatement = isKeyRetrievalSupportedStatement();
```
anupama-pathirage: Shall we change this to something like below and use that boolean for other two places? That will avoid taking `&&` twice with the same booleans.

`boolean retrieveGeneratedKeys = isKeyRetrievalSupportedStatement() && getGeneratedKey`

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/UpdateStatement.java
```diff
@@ -133,7 +134,25 @@ private boolean isDdlStatement() {
...
+            return Arrays.stream(GenKeyStmt.values()).anyMatch(stmt -> this.query.trim().toUpperCase(Locale.ENGLISH).
+                    startsWith(stmt.name()));
+        } else {
+            ErrorHandlerUtils.printError("ERROR: Unable to get keys as JDBC is not supported to retrieve " +
+                    "auto-generated keys from " + datasource.getDatabaseProductName());
+            return false;
+        }
+    }
+
+    public static boolean getGeneratedKey(Object getGeneratedKey) {
```
anupama-pathirage: What is the purpose of this method?

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/BatchUpdateStatement.java
```diff
@@ -83,8 +87,8 @@ public BatchUpdateStatement(ObjectValue client, SQLDatasource datasource, String
         String errorMessagePrefix = "failed to execute batch update";
         try {
             conn = getDatabaseConnection(strand, client, datasource);
-            boolean generatedKeyReturningSupported = isGeneratedKeyReturningSupported();
-            if (generatedKeyReturningSupported) {
+            boolean keyRetrievalSupportedStatement = datasource.isKeyRetrievalSupported();
```
daneshk:
```suggestion
            boolean keyRetrievalSupported = datasource.isKeyRetrievalSupported();
```

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/datasource/SQLDatasource.java
```diff
@@ -119,6 +123,10 @@ public boolean isPoolShutdown() {
         return poolShutdown;
     }
 
+    public boolean isKeyRetrievalSupported() {
+        return this.supportsGetGeneratedKeys;
```
anupama-pathirage: Normal practice is boolean returned function name = `is` + `boolean` variable name. Can we rename the method name or variable name to follow the same practice?

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/Constants.java
```diff
@@ -165,6 +165,9 @@
         public static final String MSSQL_SERVER = "microsoft sql server";
     }
 
+    // Warning suppression
+    public static final String UNCHECKED = "unchecked";
```
anupama-pathirage: Why do we need to define this constant like this? This is not a jdbc related constant. @daneshk / @wggihan Are we following similar pattern in other modules?

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/datasource/SQLDatasource.java
```diff
@@ -61,7 +63,9 @@ public SQLDatasource init(SQLDatasourceParams sqlDatasourceParams) {
             throw ErrorGenerator.getSQLDatabaseError(e);
         }
         try (Connection con = getSQLConnection()) {
-            databaseProductName = con.getMetaData().getDatabaseProductName().toLowerCase(Locale.ENGLISH);
+            DatabaseMetaData metaData = con.getMetaData();
+            databaseProductName = metaData.getDatabaseProductName().toLowerCase(Locale.ENGLISH);
+            this.supportsGetGeneratedKeys = metaData.supportsGetGeneratedKeys();
```
daneshk: We cannot rely on this API to check whether DB driver supports `get generated keys`. Because there is no guarantee that all DB driver implements this function. IMO we should not use this for our internal logic

stdlib/jdbc/src/test/java/org/ballerinax/jdbc/actions/UpdateTest.java
```diff
@@ -229,12 +229,24 @@ public void testInsertBoolDataAsIntsAndReturnInts() {
         Assert.assertEquals(((BInteger) returns[2]).intValue(), 1);
     }
 
+    /**
+     * Test for getting auto-generated key when getGeneratedKey is true and the table doesn't have auto generated key.
```
anupama-pathirage: Instead of this comment, shall we use descriptive name for the function itself. This is the pattern followed for other test cases as well.

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/BatchUpdateStatement.java
```diff
@@ -83,8 +87,8 @@ public BatchUpdateStatement(ObjectValue client, SQLDatasource datasource, String
...
+            if (getGeneratedKey && keyRetrievalSupportedStatement) {
```
daneshk: Shall we remove checking `keyRetrievalSupportedStatement` and only rely on value set by the user? 
and pass all the errors we encounter while performing the operation to the user.

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/Constants.java
wggihan: AFAIK, directly using annotation without using this kind of constant in other modules.

stdlib/jdbc/src/main/ballerina/src/java.jdbc/types.bal
kalaiyarasiganeshalingam: Fixed.

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,18 @@ public type Client client object {
     #
     # + sqlQuery - SQL statement to execute
     # + parameters - The parameters to be passed to the update query
+    # + returnGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is
+    #                         set to true, auto-generated keys are made available to retrieve when the statement
+    #                         is executed. If the underline JDBC driver does not support it, still it will return `()`
```
anupama-pathirage: Shouldn't it be an error if the underline JDBC driver does not support it?

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +82,23 @@ public type Client client object {
     #           may or may not continue to process the remaining commands in the batch. This property can be
     #           used to override this behavior. When it is set to true, if there is a failure in a few commands and
     #           the JDBC driver continues with the remaining commands, the successfully executed commands in the batch
-    #           also will get rolled back.
+    #           also will get rolled back
+    # + returnGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is
+    #                         set to true, auto-generated keys are made available to retrieve when the statement
+    #                         is executed. If the underline JDBC driver does not support it, still it will return `()`
+    #                         value
```
anupama-pathirage: Same as above?

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
daneshk: +1. we need to send an error if underline JDBC not supported

kalaiyarasiganeshalingam: Yes, It returns an error  but I forgot to update here. 
I have fixed it.

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,18 @@ public type Client client object {
     #
     # + sqlQuery - SQL statement to execute
     # + parameters - The parameters to be passed to the update query
+    # + returnGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is
+    #                         set to true, auto-generated keys are made available to retrieve when the statement
+    #                         is executed. If the underline JDBC driver does not support it, still it will return
```
daneshk:
```suggestion
    #                         is executed. If the underline JDBC driver does not support it, it will return
```

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +82,23 @@ public type Client client object {
     #           may or may not continue to process the remaining commands in the batch. This property can be
     #           used to override this behavior. When it is set to true, if there is a failure in a few commands and
     #           the JDBC driver continues with the remaining commands, the successfully executed commands in the batch
-    #           also will get rolled back.
+    #           also will get rolled back
+    # + returnGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is
+    #                         set to true, auto-generated keys are made available to retrieve when the statement
+    #                         is executed. If the underline JDBC driver does not support it, still it will return
```
daneshk:
```suggestion
    #                         is executed. If the underline JDBC driver does not support it, it will return
```

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,18 @@ public type Client client object {
     #
     # + sqlQuery - SQL statement to execute
     # + parameters - The parameters to be passed to the update query
+    # + returnGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is
+    #                         set to true, auto-generated keys are made available to retrieve when the statement
```
anupama-pathirage:
```suggestion
    #                         set to `true`, auto-generated keys are made available to retrieve when the statement
```

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,18 @@ public type Client client object {
...
+    #                         is executed. If the underline JDBC driver does not support it, still it will return
+    #                         an `Error`
```
anupama-pathirage:
```suggestion
    #                         an `Error`.
```

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +82,23 @@ public type Client client object {
     #           may or may not continue to process the remaining commands in the batch. This property can be
     #           used to override this behavior. When it is set to true, if there is a failure in a few commands and
     #           the JDBC driver continues with the remaining commands, the successfully executed commands in the batch
-    #           also will get rolled back.
+    #           also will get rolled back
+    # + returnGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is
+    #                         set to true, auto-generated keys are made available to retrieve when the statement
```
anupama-pathirage:
```suggestion
    #                         set to `true`, auto-generated keys are made available to retrieve when the statement
```

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +82,23 @@ public type Client client object {
...
+    #                         is executed. If the underline JDBC driver does not support it, still it will return
+    #                         an `Error`
```
anupama-pathirage:
```suggestion
    #                         an `Error`.
```

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +82,23 @@ public type Client client object {
...
     # + return - A `BatchUpdateResult` with the updated row count and returned error if any. If all the commands
     #            in the batch have executed successfully, the error will be `nil`. If one or more commands have failed,
     #            the `returnedError` field will give the corresponding `Error` along with the int[] which
-    #            contains updated row count or the status returned from each command in the batch.
+    #            contains updated row count or the status returned from each command in the batch
```
anupama-pathirage:
```suggestion
    #            contains updated row count or the status returned from each command in the batch.
```

anupama-pathirage: The convention is to use the full stop if we have multiple sentences.

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/UpdateStatement.java
```diff
@@ -70,26 +71,36 @@ public Object execute() {
...
+                    if (rs.next()) {
+                        generatedKeys = getGeneratedKeys(rs);
+                    }
                 } else {
-                    generatedKeys = new MapValueImpl<>();
+                    handleErrorOnTransaction(this.strand);
+                    String message = "The returnGeneratedKeys only support INSERT, UPDATE, DELETE and MERGE " +
+                            "operations.";
+                    checkAndObserveSQLError(strand, "execute update failed: " + message);
+                    return ErrorGenerator.getSQLApplicationError(message);
```
anupama-pathirage: Can we add a test case for this scenario as well.

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/UpdateStatement.java
```diff
@@ -70,26 +71,36 @@ public Object execute() {
...
             stmt = processedStatement.prepare();
             int count = stmt.executeUpdate();
-            MapValue<String, Object> generatedKeys;
-            if (!isDdlStatement()) {
-                rs = stmt.getGeneratedKeys();
-                //This result set contains the auto generated keys.
-                if (rs.next()) {
-                    generatedKeys = getGeneratedKeys(rs);
+            if (returnGeneratedKeys) {
+                if (isKeyRetrievalSupported()) {
```
daneshk: Why do we need this extra check?

kalaiyarasiganeshalingam: The returnGeneratedKeys only support INSERT, UPDATE, DELETE and MERGE operations So we need to check if the statement has one of these.

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/UpdateStatement.java
kalaiyarasiganeshalingam: Added.

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/UpdateStatement.java
daneshk: Do we need specifically handle this? what if we not handled and user set return generated keys for other operation. do we get a meaningful error from the driver?

kalaiyarasiganeshalingam: We haven't got a meaningful error from the driver but got an empty ResultSet. So, I specifically handled this.

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -59,14 +59,17 @@ public type Client client object {
...
     # + sqlQuery - SQL statement to execute
     # + parameters - The parameters to be passed to the update query
+    # + returnGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is
+    #                         set to `true`, auto-generated keys are made available to retrieve when the statement
+    #                         is executed. If the underline JDBC driver does not support it, it will return an `Error`.
     # + return - `UpdateResult` with the updated row count and key column values,
     #             else `Error` will be returned if there is an error
-    public remote function update(@untainted string sqlQuery, Param... parameters)
-                                  returns UpdateResult|Error {
+    public remote function update(@untainted string sqlQuery, boolean returnGeneratedKeys, Param... parameters)
```
anupama-pathirage: Can't we make this a defaultable parameter with default value is `false`. Then we don't need to append `false` for simple update statements which doesn't have parameters

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +81,22 @@ public type Client client object {
...
+    # + returnGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is
+    #                         set to `true`, auto-generated keys are made available to retrieve when the statement
+    #                         is executed. If the underline JDBC driver does not support it, it will return an `Error`.
     # + return - A `BatchUpdateResult` with the updated row count and returned error if any. If all the commands
     #            in the batch have executed successfully, the error will be `nil`. If one or more commands have failed,
     #            the `returnedError` field will give the corresponding `Error` along with the int[] which
     #            contains updated row count or the status returned from each command in the batch.
     public remote function batchUpdate(@untainted string sqlQuery, boolean rollbackAllInFailure,
-                                       Param?[]... parameters)
+                                       boolean returnGeneratedKeys, Param?[]... parameters)
```
anupama-pathirage: Same as above. Can't we change the boolean arugments to defaultable parameters?

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/UpdateStatement.java
anupama-pathirage: @daneshk I too think we don't need to explicitly check this. Isn't it a driver or DB dependent thing whether throwing an exception or returning empty result set. Why we are limiting only to the `INSERT, DELETE, UPDATE, MERGE` since the developer now has the full control to retrieve the generated keys or not.

anupama-pathirage:
@kalaiyarasiganeshalingam Shall we remove the `false` parameter in BBEs which doesn't have parameters?

stdlib/jdbc/src/main/ballerina/src/java.jdbc/client.bal
```diff
@@ -78,18 +81,22 @@ public type Client client object {
...
+    # + returnGeneratedKeys - Indicate that generated keys should be made available for retrieval. If the value is
+    #                         set to `true`, auto-generated keys are made available to retrieve when the statement
+    #                         is executed. If the underline JDBC driver does not support it, it will return an `Error`.
     # + return - A `BatchUpdateResult` with the updated row count and returned error if any. If all the commands
     #            in the batch have executed successfully, the error will be `nil`. If one or more commands have failed,
     #            the `returnedError` field will give the corresponding `Error` along with the int[] which
     #            contains updated row count or the status returned from each command in the batch.
     public remote function batchUpdate(@untainted string sqlQuery, boolean rollbackAllInFailure,
-                                       Param?[]... parameters)
+                                       boolean returnGeneratedKeys = false, Param?[]... parameters)
```
anupama-pathirage: IMO we should make `rollbackAllInFailure` also a defaultable parameter.

stdlib/jdbc/src/main/java/org/ballerinax/jdbc/statement/UpdateStatement.java
kalaiyarasiganeshalingam: Removed it.

examples/jdbc-client-crud-operations/jdbc_client_crud_operations.bal
```diff
@@ -45,7 +45,7 @@ public function main() {
     // of the `java.jdbc` module.
     int age = 24;
     string name = "Anne";
-    ret = testDB->update("INSERT INTO student(age, name) values (?, ?)",
+    ret = testDB->update("INSERT INTO student(age, name) values (?, ?)", false,
```
daneshk: Why we can't remove passing `false`? we already have set `false` as default value

kalaiyarasiganeshalingam: If update statement which doesn't have parameters, no need to append false but In this case, update statements have parameters So we need to set that value.

Issue #20933, opened 2020-02-09
Build fails on Windows with Toml test failures


jclark:
I followed the steps from the repo's README.md to build on Windows. I'm using latest version of JDK 8. I get three test failures all in ballerina-toml-parser-test-suite:

1. ManifestProcessorTest. testMultipleDependencies
2. ManifestProcessorTest. testSingleDependencies
3. TomlFileToManifestTest. testTomlFile

Stack traces as follows:

1:
```
java.lang.NumberFormatException: For input string: "sers"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at com.moandjiezana.toml.StringValueReaderWriter.replaceUnicodeCharacters(StringValueReaderWriter.java:67)
	at com.moandjiezana.toml.StringValueReaderWriter.read(StringValueReaderWriter.java:51)
	at com.moandjiezana.toml.ValueReaders.convert(ValueReaders.java:23)
	at com.moandjiezana.toml.InlineTableValueReader.read(InlineTableValueReader.java:33)
	at com.moandjiezana.toml.ValueReaders.convert(ValueReaders.java:23)
	at com.moandjiezana.toml.TomlParser.run(TomlParser.java:49)
	at com.moandjiezana.toml.Toml.read(Toml.java:138)
	at org.ballerinalang.toml.parser.ManifestProcessor.parseTomlContentFromString(ManifestProcessor.java:107)
	at org.ballerinalang.toml.ManifestProcessorTest.testMultipleDependencies(ManifestProcessorTest.java:158)
```
2:
```
java.lang.NumberFormatException: For input string: "sers"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at com.moandjiezana.toml.StringValueReaderWriter.replaceUnicodeCharacters(StringValueReaderWriter.java:67)
	at com.moandjiezana.toml.StringValueReaderWriter.read(StringValueReaderWriter.java:51)
	at com.moandjiezana.toml.ValueReaders.convert(ValueReaders.java:23)
	at com.moandjiezana.toml.InlineTableValueReader.read(InlineTableValueReader.java:33)
	at com.moandjiezana.toml.ValueReaders.convert(ValueReaders.java:23)
	at com.moandjiezana.toml.TomlParser.run(TomlParser.java:49)
[...]

PR #21491, opened 2020-03-05
Fix Toml parser test cases in windows

### LINKED ISSUE #20044: Ballerina TOML parser test cases are failing in Windows
**Description:**
The following test cases are failing in Windows. 

Failing with error  `java.lang.NumberFormatException: For input string: "sers"`
- org.ballerinalang.toml.ManifestProcessorTest > testMultipleDependencies
- org.ballerinalang.toml.ManifestProcessorTest > testSingleDependencies

These two are failing because generated TOML contains  `\` (backslash) (from windows path) and those have to be escaped according to the TOML spec. 

 Failing with error ` expected [path/to/github.balo] but found [path\to\github.balo]`
-  ballerina-toml-parser-test-suite > org.ballerinalang.toml.TomlFileToManifestTest > testTomlFile

parkavi11:
Fixes #20044, #20933

## Suggested labels
Area/BuildTools

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/test/java/org/ballerinalang/toml/ManifestProcessorTest.java
+++ b/compiler/ballerina-lang/src/test/java/org/ballerinalang/toml/ManifestProcessorTest.java
@@ -132,8 +132,8 @@ public class ManifestProcessorTest {
         Path baloPath = tmpDir.resolve("string_utils.balo");
         Files.createFile(baloPath);
 
-        String tomlData = this.validProjectBlock + "[dependencies] \n " +
-                "string-utils = { path = \"" + baloPath.toString().replace("\\", "/") + "\", version = \"1.1.5\" } \n ";
+        String tomlData = this.validProjectBlock + "[dependencies] \n string-utils = { path = \""
+                + baloPath.toString().replace("\\", "/") + "\", version = \"1.1.5\" } \n ";
 
         Manifest manifest = ManifestProcessor.parseTomlContentFromString(tomlData);
         Assert.assertEquals(manifest.getDependencies().get(0).getModuleID(), "string-utils");
@@ -157,9 +157,9 @@ public class ManifestProcessorTest {
         Path baloPath = tmpDir.resolve("string_utils.balo");
         Files.createFile(baloPath);
 
-        String tomlData = this.validProjectBlock + "[dependencies] \n " +
-                "string-utils = { path = \"" + baloPath.toString().replace("\\", "/") + "\", version = \"1.0.5\" } \n " +
-                "jquery = { version = \"2.2.3\" } \n";
+        String tomlData = this.validProjectBlock + "[dependencies] \n string-utils = { path = \""
+                + baloPath.toString().replace("\\", "/") + "\", version = \"1.0.5\" } \n "
+                + "jquery = { version = \"2.2.3\" } \n";
 
         Manifest manifest = ManifestProcessor.parseTomlContentFromString(tomlData);
         Assert.assertEquals(manifest.getDependencies().get(0).getModuleID(), "string-utils");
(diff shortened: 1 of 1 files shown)


compiler/ballerina-lang/src/test/java/org/ballerinalang/toml/ManifestProcessorTest.java
```diff
@@ -131,9 +131,20 @@ public void testSingleDependencies() throws TomlException, IOException {
...
         Path baloPath = tmpDir.resolve("string_utils.balo");
         Files.createFile(baloPath);
-        
-        Manifest manifest = ManifestProcessor.parseTomlContentFromString(this.validProjectBlock + "[dependencies] \n " +
-                "string-utils = {path = \"" + baloPath + "\", version = \"1.1.5\"} \n");
+
+        String baloPathStr = baloPath.toString();
+        String tomlData;
+
+        if (baloPathStr.contains("/")) {
```
hevayo: Shall we use Java Path object to construct the path.

parkavi11: The path gets modified in windows when Java Path object is used. That's why I used directly the string after replacing with forward slash.

pubudu91:
@hevayo @hemikak Do we need these test cases, since we're using a library for TOML now? If we need these tests, shouldn't it be using the TOML library, instead of using the old, custom written parser?

hevayo:
@pubudu91  ManifestProcessor internally use the Toml library to parse the config. And we can consider theses tests as unit tests for ManifestProcessor. But you are write we can reconsider the test cases. BTW we did find an issue because of these tests.

compiler/ballerina-lang/src/test/java/org/ballerinalang/toml/ManifestProcessorTest.java
```diff
@@ -131,9 +131,20 @@ public void testSingleDependencies() throws TomlException, IOException {
...
+            tomlData = this.validProjectBlock + "[dependencies] \n " +
```
hasithaa: We can simplify this logic. This line is redundant.

compiler/ballerina-lang/src/test/java/org/ballerinalang/toml/ManifestProcessorTest.java
```diff
@@ -154,10 +165,23 @@ public void testMultipleDependencies() throws TomlException, IOException {
...
-        
-        Manifest manifest = ManifestProcessor.parseTomlContentFromString(this.validProjectBlock + "[dependencies] \n " +
-                "string-utils = { path = \"" + baloPath + "\", version = \"1.0.5\" } \n " +
-                "jquery = { version = \"2.2.3\" } \n");
+
+        String baloPathStr = baloPath.toString();
+        String tomlData;
+
+        if (baloPathStr.contains("/")) {
+            tomlData = this.validProjectBlock + "[dependencies] \n " +
```
hasithaa: Same here, We can simplify this logic. This line is redundant in both cases.

compiler/ballerina-lang/src/test/java/org/ballerinalang/toml/ManifestProcessorTest.java
parkavi11: simplified

compiler/ballerina-lang/src/test/java/org/ballerinalang/toml/ManifestProcessorTest.java
parkavi11: done

PR #21501, opened 2020-03-05
Add filling-read operation for lvalue arrays

pubudu91:
## Samples
```ballerina
import ballerina/io;

    int[][][] arr = [];
    arr[0][0][2] = 102;
    arr[0][1][1] = 111;
    
    arr[1][0][2] = 202;
    arr[1][1][4] = 214;

    arr[4][0][1] = 501;

    foreach var x in arr {
        io:print("[");
        foreach var y in x {
            io:print("[");
            foreach var z in y {
                io:print(z, ", ");
            }
            io:print("], ");
        }
        io:println("]");
    }
}
```
The runs without any issues and prints the following:
```
[[0, 0, 102, ], [0, 111, ], ]
[[0, 0, 202, ], [0, 0, 0, 0, 214, ], ]
[]
[]
[[0, 501, ], ]
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
```diff
@@ -229,9 +229,21 @@ public Object getRefValue(long index) {
         return get(index);
     }
 
+    @Override
+    public Object fillAndGetRefValue(long index) {
+        if (refValues != null) {
+            boolean needsFilling = index >= size;
+            if (needsFilling && this.elementType.getTag() == TypeTags.ARRAY_TAG) {
```
MaryamZi:
```suggestion
            if (index >= size && this.elementType.getTag() == TypeTags.ARRAY_TAG) {
```

MaryamZi: Why are we specifically checking for array-typed elements though? This is valid for any array with an element type for which fill member is valid right?

The following seems to crash.

```ballerina
public function main() {
    map<int>[] arr = [];

    arr[0]["i"] = 1;
}
```

MaryamZi: I think we should also check if fill member is valid for this particular element type before calling `.getZeroValue()`.

The following also crashes
```ballerina
import ballerina/io;

public function main() {
    Foo[][2] arr = [];

    arr[0][1] = {s: ""}; // should fail here since Foo doesn't have a filler value

    io:println(arr[0][0].s); // fails since this particular `Foo` doesn't have a field `s`
}

type Foo record {
    string s;
};
```

ballerina-internal.log
```
[2020-03-10 11:46:37,334] SEVERE {b7a.log.crash} - null 
	at abc.main(abc.bal:8)
	at ___init.$lambda$main$(.)
	at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:426)
	at org.ballerinalang.jvm.scheduling.Scheduler.run(Scheduler.java:218)
	at org.ballerinalang.jvm.scheduling.Scheduler.runSafely(Scheduler.java:191)
 
[2020-03-10 11:46:37,339] SEVERE {b7a.log.crash} - null 
	at abc.main(abc.bal:8)
	at ___init.$lambda$main$(.)
	at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:426)
	at org.ballerinalang.jvm.scheduling.Scheduler.run(Scheduler.java:218)
	at org.ballerinalang.jvm.scheduling.Scheduler.runSafely(Scheduler.java:191)
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/TupleValueImpl.java
```diff
@@ -163,6 +163,11 @@ public Object getRefValue(long index) {
         return get(index);
     }
 
+    @Override
+    public Object fillAndGetRefValue(long index) {
+        return get(index);
```
MaryamZi: Shouldn't we implement the same for tuples too?

The following fails with an index out of range?

```ballerina
public function main() {
    [boolean, int[]...] x = [true];

    x[1][1] = 1;
}
```

The following also
```ballerina
public function main() {
    [boolean, [float, int]...] x = [true];

    x[1][1] = 1;
}
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArrayLValueFillTest.java
```diff
@@ -0,0 +1,140 @@
...
+    @Test
+    public void test2DRecordArrays() {
+        BRunUtil.invoke(compileResult, "test2DRecordArrays");
+    }
+
+    @Test
+    public void test2DObjectArrays() {
+        BValueArray arr = (BValueArray) BRunUtil.invokeFunction(compileResult, "test2DObjectArrays")[0];
+
+        assertEquals(arr.size(), 3);
```
MaryamZi: Can't we write these tests in Ballerina? Maybe in a separate PR?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
pubudu91: The `add()` method has the logic for checking if a particular type has a filler value. Will check and see what's going on here

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArrayLValueFillTest.java
pubudu91: For non-anydata types, there's no direct way to compare values for equality right? That's why I wrote them in Java for objects.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/TupleValueImpl.java
pubudu91: This was fixed by adding a fillRead() method for tuple as well.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ArrayValueImpl.java
pubudu91: Fixed this by adding a separate fillRead() method instead of relying on the add() method. Some of the logic in the add() method wasn't really applicable to the fill-read case.

Issue #21580, opened 2020-03-10
Testerina report should be generated in a readable format


azinneera:
**Description:**
testerina report should be generated in a readable format. i.e. html
Depends on #21084 and #14605

PR #21581, opened 2020-03-10
Implement testerina code coverage

### LINKED ISSUE #14605: Testerina does not generate reports as in surefire or any other
At the moment when we run Ballerina tests, it only shows the test results in the console. But ideally it should generate some sort of report so that one can do true CI/CD.

### LINKED ISSUE #21084: Porting current code coverage and report generation code

azinneera:
To make the report presentable an HTML report will be generated which will be tracked by #21580

Fixes #21084 
Fixes #14605

## Approach

Code Coverage
* Jacoco has been used for analyzing the code coverage of Ballerina source files. The implementation is broken into two parts.
  1.  instrumentation of files and generating code coverage information in binary format
  2. Converting the report in to readable format
* The first part is done by injecting the jacoco agent.jar in to the process that runs the test suite. This will create a .exec binary file in the target/coverage directory. Required filters are used to avoid instrumenting unnecessary classes.
* The second part is implemented using jacoco java APIs since the jacoco reports are not meaningful for Ballerina tests. For this, the generated .exec file is given and based on the source class files, the coverage of Ballerina source files (.bal) is generated. 
* Report generation runs as another separate process after running the test suite since instrumentation and analysis cannot be done in a single process.
* The mapping from java class file to the corresponding ballerina source file happens automatically in Jacoco using the `lineNumberTable`. 

Test Report 
* As of this PR, the test report is a json file. The file has three main elements, namely, the `projectName`, the `moduleStatus` and the `modulCoverage`(empty if code coverage is not enabled).
* Since there are separate processes run for executing tests and generating the code coverage report, separate json files are generated in these processes and later aggregated in to a single report in `RunTestsTask#execute` metthod. 
* The final json report is located at **_target/test_results.json_**

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/RunTestsTask.java
```diff
@@ -166,23 +221,95 @@ private static void writeToJson(TestSuite testSuite, Path jsonPath) {
...
+        String orgName = String.valueOf(bLangPackage.packageID.orgName);
+        String packageName = String.valueOf(bLangPackage.packageID.name);
+
+        String jacocoAgentJarPath = Paths.get(System.getProperty(BALLERINA_HOME)).resolve(BALLERINA_HOME_BRE)
+                .resolve(BALLERINA_HOME_LIB).resolve(TesterinaConstants.AGENT_FILE_NAME).toString();
         try {
+            if (coverage) {
+                String agentCommand = "-javaagent:"
+                        + jacocoAgentJarPath
+                        + "=destfile="
```
warunalakshitha: Shall we reduce the no of new lines here

PR #21653, opened 2020-03-11
Add Protobuf Definitions for Linking A Program

ramith:


misc/tracing-extensions/modules/ballerina-choreo-extension/src/main/proto/handshake.proto
```diff
@@ -30,3 +33,7 @@ message PublishProgramRequest {
     string programJson = 2;
 }
 
+message LinkApplicationRequest {
+    string id = 1;
```
a5anka: Shall we rename this field before merging?

Issue #21735, opened 2020-03-13
XML attribute access does not return an error when attribute is not available


rdhananjaya:
**Description:**
XML attribute access is lax

```ballerina
public function main() {
   xml x = xml `<abc><def>DEF</def><ghi>1</ghi></abc>`;
   string|error attr = x.k;
   io:println(attr is error); // should print true.
}
```

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #21736, opened 2020-03-13
fix xml attribute access not returning error

rdhananjaya:
## Purpose
```ballerina
public function main() {
   xml x = xml `<abc><def>DEF</def><ghi>1</ghi></abc>`;
   string|error attr = x.k;
   io:println(attr is error); // print true after this PR
}
```

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/21735
Also fix multiple xml related issues

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/langlib/lang.__internal/src/main/java/org/ballerinalang/langlib/internal/GetAttribute.java
+++ b/langlib/lang.__internal/src/main/java/org/ballerinalang/langlib/internal/GetAttribute.java
@@ -37,7 +37,7 @@ import static org.ballerinalang.jvm.util.exceptions.BallerinaErrorReasons.XML_OP
  */
 @BallerinaFunction(
         orgName = "ballerina", packageName = "lang.__internal",
-        functionName = "getAttribute",
+        functionName = "__getAttribute",
         args = {@Argument(name = "xmlValue", type = TypeKind.XML),
                 @Argument(name = "attrName", type = TypeKind.STRING)},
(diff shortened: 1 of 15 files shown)


langlib/lang.__internal/src/main/java/org/ballerinalang/langlib/internal/GetAttribute.java
```diff
@@ -54,6 +54,10 @@ public static Object getAttribute(Strand strand, XMLValue xmlVal, String attrNam
                     "Invalid xml attribute access on xml " + xmlVal.getNodeType().value());
         }
         XMLQName qname = new XMLQName(attrName);
-        return xmlVal.getAttribute(qname.getLocalName(), qname.getUri());
+        String attrVal = xmlVal.getAttribute(qname.getLocalName(), qname.getUri());
+        if (attrVal == null) {
+            return createError(XML_OPERATION_ERROR, "attribute '" + attrName + "' not found");
```
MaryamZi: As discussed offline, we would have to consider optional access here right?

Shall we also add some tests?

rdhananjaya: +1

PR #21761, opened 2020-03-13
[Observability] Remove connector name from the metric id

madushajg:

### DIFF
--- a/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/ObservableClientConnectorListener.java
+++ b/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/ObservableClientConnectorListener.java
@@ -66,6 +66,6 @@ public class ObservableClientConnectorListener extends ClientConnectorListener {
     private void addHttpStatusCode(int statusCode) {
         Optional<ObserverContext> observerContext = ObserveUtils.getObserverContextOfCurrentFrame(context.getStrand());
         observerContext.ifPresent(ctx -> ctx.addTag(ObservabilityConstants.TAG_KEY_HTTP_STATUS_CODE_GROUP,
-                String.valueOf(statusCode).charAt(0) + STATUS_CODE_GROUP_SUFFIX));
+                statusCode / 100 + STATUS_CODE_GROUP_SUFFIX));
     }
 }
--- a/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/ObservableClientConnectorListener.java
+++ b/stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/ObservableClientConnectorListener.java
@@ -24,6 +24,7 @@ import org.wso2.transport.http.netty.message.HttpCarbonMessage;
 
 import java.util.Optional;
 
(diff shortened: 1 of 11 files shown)


stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/ObservableClientConnectorListener.java
```diff
@@ -64,7 +65,7 @@ public void onError(Throwable throwable) {
 
     private void addHttpStatusCode(int statusCode) {
         Optional<ObserverContext> observerContext = ObserveUtils.getObserverContextOfCurrentFrame(context.getStrand());
-        observerContext.ifPresent(ctx -> ctx.addTag(ObservabilityConstants.TAG_KEY_HTTP_STATUS_CODE,
-                String.valueOf(statusCode)));
+        observerContext.ifPresent(ctx -> ctx.addTag(ObservabilityConstants.TAG_KEY_HTTP_STATUS_CODE_GROUP,
+                String.valueOf(statusCode).charAt(0) + STATUS_CODE_GROUP_SUFFIX));
```
nadundesilva: Can we do an integer division here instead of `String.valueOf(statusCode).charAt(0)` ? I think it might be better.

stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/callback/StreamingCallableUnitCallBack.java
```diff
@@ -40,20 +41,21 @@ public StreamingCallableUnitCallBack(StreamObserver responseSender, ObserverCont
...
     public void notifyFailure(ErrorValue error) {
         if (responseSender != null) {
             handleFailure(responseSender, error);
         }
         if (observerContext != null) {
-            observerContext.addTag(TAG_KEY_HTTP_STATUS_CODE,
-                    String.valueOf(getMappingHttpStatusCode(Status.Code.INTERNAL.value())));
+            observerContext.addTag(TAG_KEY_HTTP_STATUS_CODE_GROUP,
+                    String.valueOf(getMappingHttpStatusCode(Status.Code.INTERNAL.value())).charAt(0) +
+                            STATUS_CODE_GROUP_SUFFIX);
```
nadundesilva: Same as in the other place we can use integer division. It might be even better to just hardcode "5xx"

Please do so in all similar places

stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/ObservableClientConnectorListener.java
madushajg: Fixed with f22240bfddcb1eaaa1c7c232136fd264c35b9b78

stdlib/grpc/src/main/java/org/ballerinalang/net/grpc/callback/StreamingCallableUnitCallBack.java
madushajg: Fixed with f22240bfddcb1eaaa1c7c232136fd264c35b9b78

PR #21771, opened 2020-03-15
Fix Connector Name Issue and Test Case Failures

nadundesilva:


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmDesugarPhase.java
```diff
@@ -301,7 +301,7 @@ public static void rewriteObservableFunctionInvocations(List<BIRBasicBlock> basi
                     if (callIns.isVirtual) {
                         BIRVariableDcl selfArg = getVariableDcl(callIns.args.get(0).variableDcl);
                         connectorName = getPackageName(selfArg.type.tsymbol.pkgID.orgName,
-                                selfArg.type.tsymbol.pkgID.name) + "/" + selfArg.type.tsymbol.name.value;
```
tharinduwijewardane: i am still using the additional slash. hope merging this will not be an issue

PR #21774, opened 2020-03-16
Fix error logs of cache module

ldclakmal:
## Purpose
This PR remove the logging of the errors due to cache related operations.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/cache/src/main/ballerina/src/cache/cache_errors.bal
```diff
@@ -47,3 +47,19 @@ function prepareError(string message, error? err = ()) returns Error {
...
     return cacheError;
 }
+
+# Prepare the `error` as an `Error`.
+#
+# + message - Error message
+# + err - `error` instance
+# + return - Prepared `Error` instance
+function prepareErrorWithDebugLog(string message, error? err = ()) returns Error {
+    log:printDebug(message, err);
```
chamil321: Don't we need to check whether the debug has enabled or not?

stdlib/cache/src/main/ballerina/src/cache/cache_errors.bal
```diff
@@ -47,3 +47,19 @@ function prepareError(string message, error? err = ()) returns Error {
...
+    Error cacheError;
+    if (err is error) {
+        cacheError = error(CACHE_ERROR, message = message, cause = err);
+    } else {
+        cacheError = error(CACHE_ERROR, message = message);
+    }
+    return cacheError;
+}
```
chamil321:
```suggestion
}

```

stdlib/cache/src/main/ballerina/src/cache/cache_errors.bal
ldclakmal: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/21774/commits/d318ea7f0a283342354a9095c6d44c18fdbc1753

stdlib/cache/src/main/ballerina/src/cache/cache_errors.bal
ldclakmal: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/21774/commits/d318ea7f0a283342354a9095c6d44c18fdbc1753

stdlib/cache/src/main/ballerina/src/cache/cache_errors.bal
```diff
@@ -47,3 +47,21 @@ function prepareError(string message, error? err = ()) returns Error {
...
+
+# Prepare the `error` as an `Error`.
+#
+# + message - Error message
+# + err - `error` instance
+# + return - Prepared `Error` instance
+function prepareErrorWithDebugLog(string message, error? err = ()) returns Error {
+    log:printDebug(function() returns string {
+        return message;
+    });
```
pubudu91: Do we need to make the log statement a part of this function? Doing so defeats the purpose of using the log function which takes a function pointer. The purpose of it is to delay the evaluation of any parameters passed to the log function. Using it like this doesn't achieve that objective. You might as well use the log function which takes a string param.

ldclakmal: Added this function in order to separate out the logging and creating the particular error type of the module. That was the convention we followed most of the stdlib modules. Added the log statement as a part of this function, in order to remove code duplicates all over the places where we need to create an error.
I just wanted to evaluate the debug function iff the debug log level is enabled. That's why I used the log function which takes a function pointer. Otherwise there is no any alternative as per the BBE [1]. Also, there is nothing to evaluate since the `message` parameter is already constructed, since I need to create the error message despite of the log level is enabled or not.
Please suggest the if there is any alternative or suggestion to change.

[1] https://ballerina.io/v1-1/learn/by-example/log-api.html

pubudu91: Then you can simply use the string param instead of the function pointer. You only need to use the function pointer if you want to avoid evaluating any expression you may want to pass as a param to the log function. The log level check is done irrespective of whether it is a function pointer param or a string param.

ldclakmal: Got it. Then the @chamil321 's concern [1] become invalid. Actually, I also had the same concern. :)

[1] https://github.com/ballerina-platform/ballerina-lang/pull/21774#discussion_r392768882

ldclakmal: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/21774/commits/1a43013d6df6f6aa478d0e2e1f9b0b54438beceb

chamil321: My bad, I saw it as an expression inside the log function. :man_facepalming:

PR #21786, opened 2020-03-16
Fix BBE tests to enable BBE test automation

ramithaYO:

### DIFF
--- a/examples/abstract-objects/tests/abstract_objects_test.bal
+++ b/examples/abstract-objects/tests/abstract_objects_test.bal
@@ -1,5 +1,4 @@
 import ballerina/test;
-import ballerina/io;
 
 any[] outputs = [];
 int counter = 0;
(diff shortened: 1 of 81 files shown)


examples/abstract-objects/tests/abstract_objects_test.bal
```diff
@@ -10,14 +9,14 @@ int counter = 0;
     functionName: "println"
 }
 public function mockPrint(any... s) {
-    outputs[counter] = string.convert(s[0]);
+    outputs[counter] = s[0].toString();
```
MaryamZi: @ramithaYO @keizer619 not introduced by this PR, but a significant number of tests seem to be comparing `.toString()` representations. Can we check the actual value instead? 

Basically add all to outputs here.
```suggestion
    outputs.push(s);
```

And when testing test against the actual values.

```ballerina
    test:assertEquals(outputs[0], "John Doe");
    test:assertEquals(outputs[1], 50); // `50` instead of `"50"`
```

If there are time constraints, I guess we can go ahead with these changes for the time-being. But let's create an issue to track improving these. WDYT?

examples/anonymous-records/tests/anonymous_records_test.bal
```diff
@@ -9,14 +8,14 @@ int count = 0;
     functionName: "println"
 }
 public function mockPrint(any... s) {
-    outputs[count] = string.convert(s[0]);
+    outputs[count] = s[0].toString();
```
MaryamZi: Shall we create an issue to improve these tests without using `.toString()`, if we haven't already?

MaryamZi: Please also mention replacing `outputs[count] = x` with `outputs.push(x)`

PR #21795, opened 2020-03-16
Fix empty space print for error with empty details

KRVPerera:
## Purpose
Fix empty space print for error with empty details. With this change, we don't have to change union_type.bal BBE output to have extra spaces.

```
< error key '' not found
---
> error key '' not found<space>
```
BBE : https://ballerina.io/v1-1/learn/by-example/union-type.html

Trim the string and check if it is empty.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ErrorValue.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ErrorValue.java
@@ -88,31 +88,20 @@ public class ErrorValue extends BError implements RefValue {
     @Override
     @Deprecated
     public String stringValue() {
-        String detailsString;
         if (details == null) {
-            detailsString = "";
-        } else {
-            detailsString = org.ballerinalang.jvm.values.utils.StringUtils.getStringValue(details);
-        }
-        if (detailsString.trim().isEmpty()) {
             return "error " + reason.getValue();
         }
-        return "error " + reason.getValue() + " " + detailsString;
+        return "error " + reason.getValue() + " " + org.ballerinalang.jvm.values.utils.StringUtils.getStringValue(
+                details);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ErrorValue.java
```diff
@@ -88,16 +88,31 @@ public ErrorValue(BType type, BString reason, Object details) {
...
-                        details -> " " + org.ballerinalang.jvm.values.utils.StringUtils.getStringValue(details)).orElse(
-                        "");
+        String detailsString;
+        if (details == null) {
+            detailsString = "";
+        } else {
+            detailsString = org.ballerinalang.jvm.values.utils.StringUtils.getStringValue(details);
+        }
+        if (detailsString.trim().isEmpty()) {
+            return "error " + reason.getValue();
```
pubudu91: Do we need this conditional check? The `details` will be `null` if the user hasn't specified any details right?

KRVPerera: Yes, please check the new logic.

pubudu91:
Shall we rebase this to `stage` since this is not a BBE change?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ErrorValue.java
KRVPerera: We need to check because sometimes this is an empty MapValueImpl object

PR #21837, opened 2020-03-16
Fix taint checking BBE

pubudu91:


examples/taint-checking/taint_checking.bal
```diff
@@ -31,14 +31,13 @@ public function main(string... args) {
     var result = customerDBEP->
     select("SELECT firstname FROM student WHERE registration_id = " +
             args[0], ());
```
MaryamZi: Shall we indent these lines?

PR #21872, opened 2020-03-17
Flush output stream upon xml serialization

rdhananjaya:
## Purpose
$title

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/21873

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #21883, opened 2020-03-17
Apply fix #21872 to all xml items

rdhananjaya:
## Purpose
Generalize fix #21872  so that it applies to all xml types (element, sequence, pi, comment, text)

Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLValue.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLValue.java
@@ -16,6 +16,7 @@
 
 package org.ballerinalang.jvm.values;
 
+import org.ballerinalang.jvm.BallerinaXMLSerializer;
 import org.ballerinalang.jvm.XMLNodeType;
 import org.ballerinalang.jvm.types.BType;
 import org.ballerinalang.jvm.types.BTypes;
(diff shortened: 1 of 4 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLValue.java
```diff
@@ -195,4 +197,20 @@ public void setChildren(XMLValue seq) {
     public abstract XMLValue children(String qname);
 
     public abstract XMLValue getItem(int index);
+
+    @Override
+    public void serialize(OutputStream outputStream) {
+        try {
+            if (outputStream instanceof BallerinaXMLSerializer) {
+                ((BallerinaXMLSerializer) outputStream).write(this);
+            } else {
```
MaryamZi: What would `outputStream` be if it comes to the else clause?

rdhananjaya: https://github.com/ballerina-platform/ballerina-lang/blob/5a44cf7e3e4a992d7e2ae5fc68e74c85894457f3/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/BallerinaXMLSerializer.java#L153 here the outputStream is BallerinaXMLSerializer, which means we just want to append serialized xml to this outputStream.

Else it's a generic output stream that doesn't know how to write an xml value, so we decorate it with BallerinaXMLSerializer which knows how to write xml values.

PR #21932, opened 2020-03-18
Fix stringValue for errors when details is an empty map

KRVPerera:
## Purpose
Fix empty space print for error with empty details. With this change, we don't have to change union_type.bal BBE output to have extra spaces.
```
< error key '' not found
---
> error key '' not found<space>
```

BBE : https://ballerina.io/v1-1/learn/by-example/union-type.html'

## Approach
Check if details empty and generate string value accordingly

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ErrorValue.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ErrorValue.java
@@ -88,7 +88,7 @@ public class ErrorValue extends BError implements RefValue {
     @Override
     @Deprecated
     public String stringValue() {
-        if (details == null) {
+        if (isEmptyDetail()) {
             return "error " + reason.getValue();
         }
         return "error " + reason.getValue() + " " + org.ballerinalang.jvm.values.utils.StringUtils.getStringValue(
@@ -97,7 +97,7 @@ public class ErrorValue extends BError implements RefValue {
 
     @Override
     public BString bStringValue() {
-        if (details == null) {
+        if (isEmptyDetail()) {
             return StringUtils.fromString("error ").concat(reason);
         }
         return StringUtils.fromString("error ").concat(reason).concat(StringUtils.fromString(" ")).concat(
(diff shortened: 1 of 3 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/ErrorValue.java
```diff
@@ -254,6 +254,16 @@ private String getErrorMessage() {
         return errorMsg;
     }
 
+    private boolean isEmptyDetail() {
+        if (details == null) {
+            return true;
+        }
+        if ((details instanceof MapValue) && ((MapValue) details).isEmpty()) {
```
MaryamZi: Not changed by this PR but `details` is either a map or a record right? 

@SupunS, @warunalakshitha, can't details be a MapValue instead of Object?

KRVPerera: Yes, some generic empty check is needed.

SupunS: Yes, it should be a MapValue. This constructor as well as the API in `BValueCreator` are wrong i think.

MaryamZi: Yeah. 

@KRVPerera, shall we create an issue to track fixing the type of `details` in `ErrorValue`?

We can remove this instanceof check when `details` is defined as a `MapValue` (I guess we can do that now also, but maybe we should keep it just to be on the safer side).

KRVPerera: Create an issue : #21957

PR #21975, opened 2020-03-19
Build with test

m36dot:
This pull request makes the repo build with "./gradlew build --no-build-cache --rerun-tasks -x :jballerina-unit-test:test". This build was tested in 2 separate machine and this command builds it one machine but fails in the other.

if in the case it doesn't build (due to unidentified intermittent test cases in integration tests) use "./gradlew build --no-build-cache --rerun-tasks -x :jballerina-unit-test:test -x :jballerina-integration-test:test" to build it.

The test cases skipped are documented here -> https://docs.google.com/document/d/1GkT46RZKsfA0Usvjv1IaaBDkx6sfdxFU-dhmOFiajx4/edit

if any test cases are not documented in the doc then you can find them commented in their respective testng.xml files. All test tasks that have failure intermittent or not are documented in the above doc, only certain test cases within test tasks might be missing from the doc.

### DIFF
--- a/composer/package-lock.json
+++ b/composer/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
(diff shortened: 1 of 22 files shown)


SupunS:
@m36dot Whats the issue this PR addresses?

If this is something no longer valid, can we close this?

PR #22058, opened 2020-03-22
Introduce annotations for Path/Query/Databinding params

### LINKED ISSUE #17749: Incorrect line numbers for resource signature erros
**Description:**
See the attached image:

image

There is an error in the resource signature (L#15). But the line number is pointing to the resource annotation(L#10):
```
error: supun/http2grpc:0.1.0::proxy.bal:10:5: Invalid parameter(s) in the resource signature
```

### LINKED ISSUE #21303: Introduce separate annotations for Path/Query/Databinding params
**Description:**
Add @http:bodyparam annotation in the resource signature when binding payload instead of specifying the body field in the resource config.
Add @http:queryparam annotation in the resource signature instead of having req.getQueryParamValue(string key) function.
Evaluate the possibility of having @http:pathparam annotation and making it work without path field in the resource config.

**Steps to reproduce:**

**Affected Versions:**

chamil321:
Fixes #21303
Fixes #17749

## Samples
```ballerina
   @http:ResourceConfig {
        path: "/{foo}"
    }
    resource function testResource(http:Caller caller, http:Request req, 
                             @http:PathParam string foo,
                             @http:BodyParam string payload, 
                             @http:QueryParam string[] bar) {
        json responseJson = { path1 : foo, body : payload, query1 : bar[0]};
        checkpanic caller->respond(<@untainted> responseJson);
    }
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/http/src/test/resources/test-src/services/dispatching/uri-template.bal
```diff
@@ -12,7 +12,8 @@ service Ecommerce on testEP {
         methods:["GET"],
         path:"/products/{productId}/{regId}"
     }
-    resource function productsInfo1 (http:Caller caller, http:Request req, string productId, string regId) {
+    resource function productsInfo1 (http:Caller caller, http:Request req, @http:PathParam string productId,
```
kalaiyarasiganeshalingam:
```suggestion
    resource function productsInfo1(http:Caller caller, http:Request req, @http:PathParam string productId,
```
Can you check other places?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BFunctionType.java
```diff
@@ -119,4 +121,9 @@ public String toString() {
     public String getAnnotationKey() {
         return this.typeName;
     }
+
+    public List<String> getParamAnnotationKeys() {
+        return Arrays.stream(this.annotations.getKeys()).filter(key -> key.startsWith("$param$")).collect(
```
pubudu91: Convert this to a loop. We don't use streams in the runtime since it has a performance impact.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ServiceDesugar.java
```diff
@@ -201,7 +201,8 @@ void engageCustomServiceDesugar(BLangService service, SymbolEnv env) {
     }
 
     private void engageCustomResourceDesugar(BLangFunction functionNode, SymbolEnv env) {
-        httpFiltersDesugar.addHttpFilterStatementsToResource(functionNode, env);
-        httpFiltersDesugar.addCustomAnnotationToResource(functionNode, env);
+        if (HttpFiltersDesugar.isHttpResource(functionNode)) {
```
pubudu91: Do we really need this check? This method is called for resource functions only.

stdlib/http/src/main/java/org/ballerinalang/net/http/compiler/ResourceValidator.java
```diff
@@ -2,68 +2,172 @@
...
+    }
+
+    private static void validateQueryParam(BLangSimpleVariable param, DiagnosticLog dlog) {
+        BType paramType = param.type;
+        int varTag = paramType.tag;
+
+        if (varTag != TypeTags.STRING && (varTag != TypeTags.ARRAY ||
+                ((BArrayType) paramType).getElementType().tag != TypeTags.STRING)) {
+            dlog.logDiagnostic(Diagnostic.Kind.ERROR, param.pos, "incompatible query param type: `" + param.type +
+                    "`, expected `string`, `string[]`");
```
pubudu91: Shall we change this to the same format as the standard incompatible types error message? 
```java
dlog.logDiagnostic(Diagnostic.Kind.ERROR, param.pos, "incompatible query param type: expected 'string|string[]', found '" + param.type + "'";
```
Also, the expected type can be stated as `string|string[]` right? Also, how would the user know what type(s) to use for query params?
Another point, in error messages, we use ' around type names, not `

stdlib/http/src/main/java/org/ballerinalang/net/http/compiler/ResourceValidator.java
```diff
@@ -2,68 +2,172 @@
...
+    }
+
+    private static void validateBodyParam(BLangSimpleVariable param, DiagnosticLog dlog) {
+        BType paramType = param.type;
+        int type = paramType.tag;
+
+        if (type != TypeTags.RECORD && type != TypeTags.JSON && type != TypeTags.XML &&
+                type != TypeTags.STRING && (type != TypeTags.ARRAY || !validArrayType(paramType))) {
+            dlog.logDiagnostic(Diagnostic.Kind.ERROR, param.pos, "incompatible entity-body param type: `"
+                    + param.type + "`, expected `string`, `json`, `xml`, `byte[]`, `{}`, `{}[]`");
```
pubudu91: Same comment as above here, regarding error message format, use of unions and formatting

stdlib/http/src/main/java/org/ballerinalang/net/http/compiler/ResourceValidator.java
```diff
@@ -2,68 +2,172 @@
...
+        return null;
+    }
+
+    private static void validatePathParam(BLangSimpleVariable param, List<String> pathSegments, DiagnosticLog dlog) {
+        BType paramType = param.type;
+        int varTag = paramType.tag;
+
+        if (varTag != TypeTags.STRING && varTag != TypeTags.INT && varTag != TypeTags.BOOLEAN &&
+                varTag != TypeTags.FLOAT) {
+            dlog.logDiagnostic(Diagnostic.Kind.ERROR, param.pos, "incompatible path param type: `" + param.type +
```
pubudu91: Check https://github.com/ballerina-platform/ballerina-lang/pull/22058/files#r396224360

stdlib/http/src/main/java/org/ballerinalang/net/http/compiler/ResourceValidator.java
```diff
@@ -2,68 +2,172 @@
...
-            dlog.logDiagnostic(Diagnostic.Kind.ERROR, pos, "first parameter should be of type " + CALLER_TYPE);
+        if (isInvalidResourceParam(signatureParams.get(0), CALLER_TYPE)) {
+            dlog.logDiagnostic(Diagnostic.Kind.ERROR, pos, "first parameter should be of type `" + CALLER_TYPE + "`");
             return;
         }
 
-        if (!isValidResourceParam(signatureParams.get(1), HTTP_REQUEST_TYPE)) {
-            dlog.logDiagnostic(Diagnostic.Kind.ERROR, pos, "second parameter should be of type " + HTTP_REQUEST_TYPE);
+        if (isInvalidResourceParam(signatureParams.get(1), HTTP_REQUEST_TYPE)) {
+            dlog.logDiagnostic(Diagnostic.Kind.ERROR, pos, "second parameter should be of type `" +
```
pubudu91: Use ' instead of ` to quote types

pubudu91: Check in other places as well

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ServiceDesugar.java
chamil321: Yes, but this is called for both HTTP and WebSocket resource functions. Since the service typing support is not available yet, this method is used to check the first function parameter to distinguish it.

stdlib/http/src/main/java/org/ballerinalang/net/http/compiler/ResourceValidator.java
chamil321: Changed the error as suggested.
If we stated the type as `string|string[]`, user can misinterpret is as a union type as ballerina has similar syntax. 
The types of each param are mentioned in the module.md. Also, we can add it to the annotation docs if suitable

MaryamZi: Yeah assuming we only allow either `string` or `string[]` I also think `string|string[]` could be misleading.

How about
`"incompatible query param: expected a param of type 'string' or 'string[]', found '" + param.type + "'"`

On  a separate note, does "query **parameter**" sound odd? `QueryParam` for the annotation is OK, but I was wondering if we should use "parameter" for the error message.

chamil321: Changed the error as suggested

PR #22119, opened 2020-03-23
Fixes issue in handling error for streams with query expressions

### LINKED ISSUE #21757: [Query Expressions] Issue in handling error for streams
**Description:**
As per the query specification, when querying a stream there is a guideline (below) to handle the errors. 

If i is a stream
call i.next() resulting in a value r;
if r is an error, then complete execution of the from-clause early with error r;

But, there is a spec violation with the current query implementation for streams. As per the current implementation, `from-clause` gets de-sugared to a foreach statement. As per the recent discussion, it seems foreach statements will not work for iterators which could through error (i.e, error). Then, we need to improve/change the de-sugar implementation with `foreach` function of stream lang-lib.

pcnfernando:
Fixes #21757

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
@@ -131,7 +131,7 @@ public class QueryDesugar extends BLangNodeVisitor {
         BLangLiteral nillLiteral = ASTBuilderUtil.createLiteral(fromClause.pos, symTable.nilType,
                 null);
         BVarSymbol outputArrayVarSymbol = new BVarSymbol(0, new Name("$outputDataArray$"),
-                env.scope.owner.pkgID, outputArrayType, env.scope.owner);
+                env.scope.owner.pkgID, outputUnionType, env.scope.owner);
         BLangSimpleVariable outputArrayVariable =
                 ASTBuilderUtil.createVariable(pos, "$outputDataArray$", outputUnionType,
                         nillLiteral, outputArrayVarSymbol);
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
@@ -28,7 +28,7 @@ import org.wso2.ballerinalang.compiler.semantics.model.symbols.BInvokableSymbol;
 import org.wso2.ballerinalang.compiler.semantics.model.symbols.BVarSymbol;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BArrayType;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BType;
-import org.wso2.ballerinalang.compiler.semantics.model.types.BUnionType;
+import org.wso2.ballerinalang.compiler.tree.BLangIdentifier;
 import org.wso2.ballerinalang.compiler.tree.BLangNodeVisitor;
 import org.wso2.ballerinalang.compiler.tree.BLangSimpleVariable;
 import org.wso2.ballerinalang.compiler.tree.clauses.BLangDoClause;
@@ -50,6 +50,7 @@ import org.wso2.ballerinalang.compiler.tree.expressions.BLangStatementExpression
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangTypeTestExpr;
 import org.wso2.ballerinalang.compiler.tree.statements.BLangAssignment;
 import org.wso2.ballerinalang.compiler.tree.statements.BLangBlockStmt;
+import org.wso2.ballerinalang.compiler.tree.statements.BLangExpressionStmt;
 import org.wso2.ballerinalang.compiler.tree.statements.BLangIf;
 import org.wso2.ballerinalang.compiler.tree.statements.BLangSimpleVariableDef;
 import org.wso2.ballerinalang.compiler.tree.statements.BLangStatement;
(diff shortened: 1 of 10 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
```diff
@@ -96,78 +106,123 @@ public static QueryDesugar getInstance(CompilerContext context) {
...
-        //      x[x.length()] = {
+        //      $tempDataArray$[$tempDataArray$.length()] = {
         //         firstName: person.firstName,
         //         lastName: person.lastName
         //      };
-
-        BLangInvocation lengthInvocation = createLengthInvocation(selectClause.pos, outputArrayVariable.symbol);
-        lengthInvocation.expr = outputVarRef;
-        BLangIndexBasedAccess indexAccessExpr = ASTBuilderUtil.createIndexAccessExpr(outputVarRef, lengthInvocation);
+        BLangInvocation lengthInvocation = createLengthInvocation(selectClause.pos, tempArrayVariable.symbol);
```
gimantha: Shall we replace this with push()?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2824,10 +2826,32 @@ public void visit(BLangQueryExpr queryExpr) {
...
+            expSelectType = nilRemovedSet.size() == 1 ? nilRemovedSet.iterator().next() :
+                    BUnionType.create(null, nilRemovedSet);
+            if (expSelectType.tag == TypeTags.ARRAY) {
+                expSelectType = ((BArrayType) expSelectType).eType;
+            }
+        }
         BType selectType = checkExpr(selectClause.expression, whereEnv, expSelectType);
-        resultType = selectType == symTable.semanticError ? selectType : new BArrayType(selectType);
+        BType actualType = new BArrayType(selectType);
+        if ((collectionNode instanceof BLangSimpleVarRef) &&
```
gimantha: check the tag instead of using instance of

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2824,10 +2826,32 @@ public void visit(BLangQueryExpr queryExpr) {
...
+                (((BLangSimpleVarRef) collectionNode).type instanceof BStreamType)) {
```
gimantha: check the type tag instead of instanceof

tests/jballerina-unit-test/src/test/resources/test-src/query/simple-query-with-defined-type.bal
```diff
@@ -151,16 +174,16 @@ function testFilteringNullElements() returns Person[] {
     return outputPersonList;
 }
 
-function testMapWithArity () returns string[] {
-    map<any> m = {a:"1A", b:"2B", c:"3C", d:"4D"};
+function testMapWithArity() returns string[] {
+    map<any> m = {a: "1A", b: "2B", c: "3C", d: "4D"};
     string[] val = from var v in m
                    where <string> v == "1A"
```
gimantha: What will happen if we remove the cast <string> ?

tests/jballerina-unit-test/src/test/resources/test-src/query/simple-query-with-var-type.bal
```diff
@@ -217,5 +228,15 @@ function testQueryExpressionWithSpreadOperatorV2() returns Teacher[]{
                    teacherId: "TER1200"
             };
 
-    return  outputPersonList;
+    return outputPersonList;
+}
+
+public function testQueryWithStream() returns int[]|error {
+    NumberGenerator numGen = new;
+    var numberStream = new stream<int, error>(numGen);
```
gimantha: What will happen if the iterator object next method does not return an error but the stream has the errorType as the completionType?

tests/jballerina-unit-test/src/test/resources/test-src/query/query-exp-iterable-objects.bal
```diff
@@ -0,0 +1,61 @@
+type Iterable object {
```
gimantha: Shall we add another test for iterable which returns an error as well?

tests/jballerina-unit-test/src/test/resources/test-src/query/simple-query-with-defined-type.bal
pcnfernando: it will throw error since map is any type map. Isn't that expected?

mohanvive: It is a compile time error. (i.e, operator '==' not defined for 'any' and 'string')

mohanvive: sorry, just now saw. That is an expected case.

gimantha: ah yes.. sorry didnt see that m is a map<any>

tests/jballerina-unit-test/src/test/resources/test-src/query/simple-query-with-var-type.bal
pcnfernando: It passes without throwing any error

Issue #22076, opened 2020-03-23
`.isReadOnly()` evaluates to false for XML text singletons


MaryamZi:
**Description:**
The spec says "A text singleton is inherently immutable, since it does not have storage identity.", but $title.

**Steps to reproduce:**
```ballerina
import ballerina/io;
import ballerina/lang.'xml;

public function main() {
    'xml:Text a = <'xml:Text> xml `foo`;
    io:println(a.isReadOnly()); // prints false
}
```

**Affected Versions:**
1.2.0

PR #22126, opened 2020-03-24
Improve xml readOnly operation

rdhananjaya:
## Purpose
Improve readOnly operation for xml:Text, xml:ProcessingInstruction, xml:Comment

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/22076 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLNonElementItem.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLNonElementItem.java
@@ -23,6 +23,7 @@ import org.ballerinalang.jvm.BallerinaXMLSerializer;
 import org.ballerinalang.jvm.XMLNodeType;
 import org.ballerinalang.jvm.values.api.BMap;
 import org.ballerinalang.jvm.values.api.BXML;
+import org.ballerinalang.jvm.values.freeze.FreezeUtils;
 import org.ballerinalang.jvm.values.freeze.Status;
 
 import java.io.ByteArrayOutputStream;
@@ -137,7 +138,6 @@ public abstract class XMLNonElementItem extends XMLValue {
 
     @Override
     public Object copy(Map<Object, Object> refs) {
-        // XMLContentHolderItem is immutable
         return this;
     }
 
@@ -208,16 +208,19 @@ public abstract class XMLNonElementItem extends XMLValue {
 
     @Override
     public void attemptFreeze(Status freezeStatus) {
-
+        if (FreezeUtils.isOpenForFreeze(this.freezeStatus, freezeStatus)) {
+            this.freezeStatus = freezeStatus;
+        }
     }
 
     @Override
     public void freezeDirect() {
-
+        this.freezeStatus.setFrozen();
     }
 
     @Override
     public Object freeze() {
+        freezeDirect();
         return this;
     }
 
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLNonElementItem.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLNonElementItem.java
@@ -29,7 +29,6 @@ import org.ballerinalang.jvm.values.freeze.Status;
 import java.io.ByteArrayOutputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
-import java.util.Map;
 import java.util.NoSuchElementException;
 
 import static org.ballerinalang.jvm.util.BLangConstants.STRING_NULL_VALUE;
@@ -134,16 +133,6 @@ public abstract class XMLNonElementItem extends XMLValue {
             return this;
         }
         return new XMLSequence();
-    }
-
-    @Override
-    public Object copy(Map<Object, Object> refs) {
-        return this;
-    }
-
-    @Override
-    public Object frozenCopy(Map<Object, Object> refs) {
-        return this;
     }
 
     @Override
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLSequence.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLSequence.java
@@ -553,11 +553,15 @@ public final class XMLSequence extends XMLValue {
 
     @Override
     public synchronized boolean isFrozen() {
+        if (freezeStatus.isFrozen()) {
+            return true;
+        }
         for (BXML child : this.children) {
             if (!child.isFrozen()) {
                 return false;
             }
         }
+        freezeStatus.setFrozen();
         return true;
     }
 
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLSequence.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLSequence.java
@@ -552,6 +552,16 @@ public final class XMLSequence extends XMLValue {
     }
 
     @Override
+    public synchronized boolean isFrozen() {
+        for (BXML child : this.children) {
+            if (!child.isFrozen()) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    @Override
     public IteratorValue getIterator() {
         return new IteratorValue() {
             @Override
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLText.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLText.java
@@ -26,6 +26,7 @@ import org.ballerinalang.jvm.types.BTypes;
 import org.ballerinalang.jvm.values.api.BString;
 import org.ballerinalang.jvm.values.api.BXML;
 
+import java.util.Map;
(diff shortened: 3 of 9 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLNonElementItem.java
```diff
@@ -137,7 +138,6 @@ public XMLValue getItem(int index) {
 
     @Override
     public Object copy(Map<Object, Object> refs) {
-        // XMLContentHolderItem is immutable
         return this;
     }
```
MaryamZi: This logic only applies to XMLText.java now right? Shall we remove this implementation and implement `copy()` XMLText.java instead?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/builtinfunctions/FreezeAndIsFrozenTest.java
```diff
@@ -436,6 +436,16 @@ public void testFrozenValueUpdatePanicWithCheckTrap() {
...
     }
 
+    @Test
+    public void testXMLItemsCloneReadOnly() {
+        BValue[] returns = BRunUtil.invoke(result, "testXMLItemsCloneReadOnly", new BValue[0]);
+        Assert.assertTrue(((BBoolean) returns[0]).booleanValue());
+        Assert.assertTrue(((BBoolean) returns[1]).booleanValue());
+        Assert.assertTrue(((BBoolean) returns[2]).booleanValue());
+        Assert.assertTrue(((BBoolean) returns[3]).booleanValue());
+        Assert.assertTrue(((BBoolean) returns[4]).booleanValue());
```
MaryamZi: Shall we move these assertions to the bal file instead?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLNonElementItem.java
rdhananjaya: +1

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/builtinfunctions/FreezeAndIsFrozenTest.java
rdhananjaya: Moved

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLSequence.java
```diff
@@ -551,6 +551,16 @@ public void freezeDirect() {
         }
     }
 
+    @Override
+    public synchronized boolean isFrozen() {
```
MaryamZi: IIUC, a sequence cannot be updated. So if all the children are read-only that indirectly makes the sequence also read-only?

If that is the case, if the following check evaluates to true, can't we set the frozen flag to true here? That way in the read-only scenario, we can avoid going through all the children over and over again?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLText.java
```diff
@@ -57,6 +58,21 @@ public String getItemType() {
         return getNodeType().value();
     }
 
+    @Override
```
MaryamZi: Shouldn't we override `attemptFreeze()`, `freezeDirect()` and `freeze()` too?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/freeze-and-isfrozen.bal
```diff
@@ -251,6 +251,20 @@ function testFrozenXmlSetChildrenDeep() {
...
+                        xml `<!-- comment text -->`,
+                        xml `<?PIT data?>`,
+                        xml `<item><child>String Content <sub></sub>More Str</child><child></child></item>`);
+
+    xml x1 = x0.cloneReadOnly();
+    assertTrue((x1.<hello>).isReadOnly());
+    assertTrue((x1.<hello>/*).isReadOnly());
+    assertTrue(x1[1].isReadOnly());
+    assertTrue(x1[2].isReadOnly());
+    assertTrue((x1/**/<child>/*)[0].isReadOnly());
```
MaryamZi: Can we add some false tests too?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/freeze-and-isfrozen.bal
```diff
@@ -489,3 +503,10 @@ type FreezeAllowedDepartment record {|
     string head;
     (int)...;
 |};
+
+function assertTrue(boolean value) {
+    if !(value) {
+        error e = error("Not True", message = "expected: true, found: " + value.toString());
+        panic e;
```
MaryamZi: Shall we use a reason without spaces?

```suggestion
        error e = panic error("AssertionError", message = "expected: true, found: " + value.toString());
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLSequence.java
rdhananjaya: Membership of the sequence is unmodifiable but each individual xml item is not.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLText.java
rdhananjaya: We can re-use the super class methods without any ill efect IMO

tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/freeze-and-isfrozen.bal
rdhananjaya: Will do

tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/freeze-and-isfrozen.bal
rdhananjaya: +1, I blindly copied it from some other code will change

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLSequence.java
MaryamZi: Yeah, but if this evaluates to true once, it will always be true after that right?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLText.java
MaryamZi: Aren't they unnecessary though?

rdhananjaya: They are sitting high in the hierarchy, providing expected behavior.
In my view, overriding them with empty methods are unnecessary.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLSequence.java
rdhananjaya: Ah yes, I miss that on your first comment.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/XMLText.java
MaryamZi: `freezeDirect()` is OK. 

But for `attemptFreeze()` if (by the slightest chance) there is an attempt to freeze the same value concurrently there is a chance it'll panic. https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/freeze/FreezeUtils.java#L45 unless we set the frozen flag when creating the value itself.

This IMO is unnecessary since it's making it read-only is always possible.

And for `freeze()` I think we can remove `freezeDirect()` call.

```
    @Override
    public Object freeze() {
        // freezeDirect(); we can remove this
        return this;
    }
```

rdhananjaya: If I'm not mistaken freeze action could only be initiated from a `cloneReadOnly` operation, which means there couldn't be any concurrent invocations

MaryamZi: I guess it'll work since we seem to be checking `isFrozen()` on ` copy()` results before attempting freezing and  `.cloneReadOnly()` seems to be calling `freezeDirect()`.

PR #22140, opened 2020-03-24
Add AST node generator implementation 

tharindulak:


sameerajayasoma:
I've merged this to my local environment. I will send a new PR after updating the tool to generate syntax tree with the updated format.

Thanks, @tharindulak for the contribution

PR #22145, opened 2020-03-24
Add Avro key serialization / deserialization support for Kafka module

### LINKED ISSUE #21965: Add key serialization / deserialization support to Kafka Avro
**Description:**
Kafka Avro only supports Value serialization / deserialization now. It should support key serialization too.

### LINKED ISSUE #22022: Kafka AvroRecord missing an API documentation
> kafka `AvroRecord` does not include API documentation.

ThisaruGuruge:
> Fixes #22022 
> Fixes #21965 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/messaging/kafka/src/main/ballerina/src/kafka/producer.bal
```diff
@@ -270,210 +271,246 @@ public type Producer client object {
...
     string keySerializerType) returns ProducerError? {
     if (key is ()) {
-        return producerSendString(producer, value, topic, partition, timestamp);
+        return sendString(producer, value, topic, partition, timestamp);
     }
     if (keySerializerType == SER_STRING) {
         if (key is string) {
             handle keyHandle = java:fromString(key);
-            return producerSendStringString(producer, value, topic, keyHandle, partition, timestamp);
+            return sendStringString(producer, value, topic, keyHandle, partition, timestamp);
```
wggihan: Shall we revisit this function names? It bit difficult to get the idea that what this does by looking at the code?

stdlib/messaging/kafka/src/main/java/org/ballerinalang/messaging/kafka/utils/AvroUtils.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    protected static MapValue<String, Object> handleAvroConsumer(Object value) {
+        if (value instanceof GenericRecord) {
+            MapValue<String, Object> genericAvroRecord = getAvroGenericRecord();
+            populateBallerinaGenericAvroRecord(genericAvroRecord, (GenericRecord) value);
+            return genericAvroRecord;
+        } else {
+            throw new BLangRuntimeException("Invalid type - expected: AvroGenericRecord");
```
wggihan: Shall we try to return ErrorValue here?

ThisaruGuruge: This method is to just populate the consumer record. Ideally, this line of code shouldn't be reached. Also, this is called inside a resource invocation. Hence, I think throwing this is appropriate.

wggihan: Both are extending from RuntimeException. But behaviour is different.  What is the behaviour getting if this throws this?

ThisaruGuruge: This method is called from both the `poll()` function and the resource. If this line is hit when the `poll` is called, this will panic. If this is hit through the resource, this will log the error, and will continue.

stdlib/messaging/kafka/src/main/ballerina/src/kafka/producer.bal
wggihan: IMO,  one of the most important factor is the readability of the code. We should able to read this even in the non-IDE situation. My suggestion is to go further and change to something like
`sendByteArrayValueAndByteArrayKey()` if possible.

ThisaruGuruge: Agreed. will update.

stdlib/messaging/kafka/src/main/java/org/ballerinalang/messaging/kafka/utils/AvroUtils.java
wggihan: Sometimes back there was a discussion about this throwing error whether it ErrorValue or BRuntime*. I couldn't remember now. @hasithaa @pubudu91 Any comment about what to use here?

ThisaruGuruge: Updated

PR #22302, opened 2020-03-30
Add support to report function execution errors in Observations

nadundesilva:

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/observability/metrics/BallerinaMetricsObserver.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/observability/metrics/BallerinaMetricsObserver.java
@@ -25,7 +25,6 @@ import java.time.Duration;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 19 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/observability/metrics/BallerinaMetricsObserver.java
```diff
@@ -115,10 +114,6 @@ private void stopObservation(ObserverContext observerContext) {
 
         Set<Tag> allTagsForCounts = new HashSet<>(allTags.size() + 1);
         Tags.tags(allTagsForCounts, allTags);
-        Boolean error = (Boolean) observerContext.getProperty(PROPERTY_ERROR);
-        if (error != null && error) {
```
tharinduwijewardane, line 119: can't this be simplified into `if (error) {`

tharinduwijewardane: ah this is a removed line. my bad

PR #22316, opened 2020-03-30
[master] Add support for record -> json assignment

### LINKED ISSUE #18697: JSON compatible record to JSON assignment not allowed
Consider the following:
```ballerina
public function main() {
    Person p = {};
    json j = p; // Currently not allowed. Compile error
}

type Person record {|
    string name = "";
    int age = 0;
|};
```
The above case should be allowed. Attempts to add additional fields to the `json` should result in panics.

### LINKED ISSUE #22306: Field access and optional field access not allowed for records assigned to a JSON variable
Consider the following:
```ballerina
import ballerina/io;

public function main() {
    testNestedRecordModification();
}

function testNestedRecordModification() {
    record {|
        string name;
        record {|
            record {|
                int x;
                float...;
            |} nestedL2;
        |} nestedL1;
    |} rec = {
        name: "John",
        nestedL1: {
            nestedL2: {
                x: 100
            }
        }
    };

    json jR = rec;

    json nestedL2 = checkpanic jR.nestedL1.nestedL2;
    map<json> l2M = <map<json>>nestedL2;
    l2M["restF1"] = 12.34;
}
```
Currently the above panics.
```
error: {ballerina}JSONOperationError message=JSON value is not a mapping
	at 18697:testNestedRecordModification(18697.bal:48)
	   18697:main(18697.bal:6)
```
The reason is since the value referred to by `jR` is actually a record, it's not considered a JSON object.

pubudu91:
## Samples
```ballerina
public function main() {
    Person p = {};
    json j = p; // Used to be a compile error
}

type Person record {|
    string name = "";
    int age = 0;
|};
```
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/JSONUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/JSONUtils.java
@@ -298,7 +298,7 @@ public class JSONUtils {
 
         BType type = ((RefValue) json).getType();
         int typeTag = type.getTag();
-        return typeTag == TypeTags.JSON_TAG || typeTag == TypeTags.MAP_TAG || typeTag == TypeTags.RECORD_TYPE_TAG;
+        return typeTag == TypeTags.MAP_TAG || typeTag == TypeTags.RECORD_TYPE_TAG;
     }
 
     /**
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/JSONUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/JSONUtils.java
@@ -297,7 +297,8 @@ public class JSONUtils {
         }
 
         BType type = ((RefValue) json).getType();
-        return type.getTag() == TypeTags.JSON_TAG || type.getTag() == TypeTags.MAP_TAG;
+        int typeTag = type.getTag();
+        return typeTag == TypeTags.JSON_TAG || typeTag == TypeTags.MAP_TAG || typeTag == TypeTags.RECORD_TYPE_TAG;
     }
 
     /**
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -518,7 +518,7 @@ public class Types {
 
         if (targetTag == TypeTags.MAP && sourceTag == TypeTags.RECORD) {
             BRecordType recordType = (BRecordType) source;
-            return isAssignableRecordType(recordType, (BMapType) target);
+            return isAssignableRecordType(recordType, target);
         }
 
         if (target.getKind() == TypeKind.SERVICE && source.getKind() == TypeKind.SERVICE) {
@@ -564,6 +564,10 @@ public class Types {
 
             if (sourceTag == TypeTags.MAP) {
                 return isAssignable(((BMapType) source).constraint, target, unresolvedTypes);
+            }
+
+            if (sourceTag == TypeTags.RECORD) {
+                return isAssignableRecordType((BRecordType) source, target);
             }
         }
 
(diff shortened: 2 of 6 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/JSONUtils.java
```diff
@@ -297,7 +297,8 @@ public static boolean isJSONObject(Object json) {
         }
 
         BType type = ((RefValue) json).getType();
-        return type.getTag() == TypeTags.JSON_TAG || type.getTag() == TypeTags.MAP_TAG;
+        int typeTag = type.getTag();
+        return typeTag == TypeTags.JSON_TAG || typeTag == TypeTags.MAP_TAG || typeTag == TypeTags.RECORD_TYPE_TAG;
```
MaryamZi: Not related to this PR, but returning true for `isJSONObject` if ` typeTag == TypeTags.JSON_TAG` is wrong right? And we need to check the map's constraint too assuming this is a generic method.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -618,17 +622,34 @@ private boolean isAssignable(BType source, BType target, Set<TypePair> unresolve
...
+        BType targetType;
+        switch (type.tag) {
+            case TypeTags.MAP:
+                targetType = ((BMapType) type).constraint;
+                break;
+            case TypeTags.JSON:
+                targetType = type;
+                break;
+            default:
+                throw new IllegalArgumentException("Incompatible target type: " + type.toString());
```
MaryamZi: Shouldn't we just return false?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -618,17 +622,34 @@ private boolean isAssignable(BType source, BType target, Set<TypePair> unresolve
...
+        }
+
         if (recordType.sealed) {
-            return recordFieldsAssignableToMap(recordType, targetMapType);
+            return recordFieldsAssignableToType(recordType, targetType);
         } else {
-            return isAssignable(recordType.restFieldType, targetMapType.constraint)
-                    && recordFieldsAssignableToMap(recordType, targetMapType);
+            return recordFieldsAssignableToType(recordType, targetType) &&
+                    isAssignable(recordType.restFieldType, targetType);
```
MaryamZi: Shall we move this logic to `recordFieldsAssignableToType`. Rest fields are also record fields?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/json/RecordToJSONTest.java
```diff
@@ -0,0 +1,81 @@
+/*
+ *   Copyright (c) 2016, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
MaryamZi:
```suggestion
 *   Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
pubudu91: Added this default case since I changed the method signature to accept `BType` and wanted to ensure this method is used only in contexts with the supported types (`map` and `json`).

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
pubudu91: +1.. That'd actually simplify things a bit

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/JSONUtils.java
pubudu91: Just checked this. It seems simple literals are in their original type, so it doesn't count as a JSON tag. For example, the following works as expected:
```ballerina
import ballerina/io;

public function main() {
    json j1 = 10;
    io:println(j1.name);
}
```
`j1` is a `Long` when you check at this point.

MaryamZi: Yeah but `typeTag == TypeTags.JSON_TAG` will never be true right?

pubudu91: Ah just checked it.. yeah, you're right.. for JSON objects, it's `map<json>`.. I'll remove this condition and see..

pubudu91: Removed it in 4b4eaeab6e39b5cf11c6e0dbd24855b4643cf582

PR #22346, opened 2020-03-31
Narrow type bound to union type parameters

### LINKED ISSUE #16828: Type bound to type param needs to be narrowed down for union types
Consider the following:
```ballerina
public function test() returns int {
    map<int>|map<Person> m = <map<int>>{"1": 1};
    int x = m.get("1"); // compile error
    return x;
}

type Person record {|
    string name;
    int age;
|};
```
In the above, `m.get("1")` invocation's return type is `any|error`. But the type can be narrowed to `Person|int`.

### LINKED ISSUE #22137: Incorrect type-checking for push on union-typed list
**Description:**
$title.

Compilation doesn't fail when the value pushed is not compatible with any of the members of the union.

**Steps to reproduce:**
```ballerina
import ballerina/lang.array;

public function main() {
    // int[] x = [1];
    // Fails for non-union type.
    // x.push(true); // incompatible types: expected 'int', found 'boolean'

    int[]|string[] arr = [1, 2];
    arr.push(true);
    array:unshift(arr, 13.2);
}
```

Results in a bad, sad error at runtime similar to https://github.com/ballerina-platform/ballerina-lang/issues/22129

**Affected Versions:**
1.2.0

KRVPerera:
Fixes #16828
Fixes #22137

## Approach
Add logic to TypeParamAnalyser. Union and union with arrays will be flattened and merged as a union again.

#### Case 1
```Ballerina
int[]|float[] arr = <int[]>[1, 2];
[int, int][] y = arr.enumerate();
```
##### Previous output
```Logtalk
error: .::t3.bal:5:30: incompatible types: expected '[int,int][]', found '[int,(any|error)][]'
```
##### output after PR
```Logtalk
error: .::t3_bad.bal:5:22: incompatible types: expected '[int,int][]', found '[int,(int|float)][]'
```
Correct code must be `[int, (int | float)][] y = arr.enumerate();`

#### Case 2
```Ballerina
map<int>|map<Person> m = <map<int>>{"1": 1};
int|Person x = m.get("1");
```
##### Previous output
```Logtalk
error: .::t1.bal:7:20: incompatible types: expected '(int|Person)', found '(any|error)'
```
##### output after PR
compiles correctly

#### Case 3
```Ballerina
int[]|string[] arr = [1, 2];
arr.push(true);
```
##### Previous output
```Logtalk
ballerina: Oh no, something really went wrong.
```
##### output after PR
```Logtalks
error: .::tt2.bal:4:14: incompatible types: expected '(int|string)', found 'boolean'
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
@@ -414,8 +414,6 @@ public class TypeParamAnalyzer {
         for (BType type : actualType.getMemberTypes()) {
             if (type.tag == TypeTags.ARRAY) {
                 tupleTypes.add(((BArrayType) type).eType);
-            } else {
-                tupleTypes.add(type);
             }
         }
         BUnionType tupleElementType = BUnionType.create(null, tupleTypes);
@@ -428,8 +426,6 @@ public class TypeParamAnalyzer {
         for (BType type : actualType.getMemberTypes()) {
             if (type.tag == TypeTags.MAP) {
                 tupleTypes.add(((BMapType) type).constraint);
-            } else {
-                tupleTypes.add(type);
             }
         }
         BUnionType tupleElementType = BUnionType.create(null, tupleTypes);
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
@@ -278,6 +278,7 @@ public class TypeParamAnalyzer {
             updateTypeParamAndBoundType(pos, env, expType, actualType);
 
             // If type param discovered before, now type check with actual type. It has to be matched.
+
             if (checkContravariance) {
                 types.checkType(pos, getMatchingBoundType(expType, env, new HashSet<>()), actualType,
                                 DiagnosticCode.INCOMPATIBLE_TYPES);
@@ -311,10 +312,6 @@ public class TypeParamAnalyzer {
                 if (actualType.tag == TypeTags.RECORD) {
                     findTypeParamInMapForRecord(pos, (BMapType) expType, (BRecordType) actualType, env, resolvedTypes,
                                                 result);
-                }
-                if (actualType.tag == TypeTags.UNION) {
-                    findTypeParamInUnion(pos, ((BMapType) expType).constraint, (BUnionType) actualType, env, resolvedTypes,
-                                               result);
                 }
                 return;
             case TypeTags.STREAM:
@@ -422,6 +419,35 @@ public class TypeParamAnalyzer {
         BUnionType tupleElementType = BUnionType.create(null, members);
         findTypeParam(pos, expType, tupleElementType, env, resolvedTypes, result);
     }
+
+    private void findTypeParamInUnionForArray(DiagnosticPos pos, BArrayType expType, BUnionType actualType,
+    SymbolEnv env, HashSet<BType> resolvedTypes, FindTypeParamResult result) {
+        LinkedHashSet<BType> tupleTypes = new LinkedHashSet<>();
+        for (BType type : actualType.getMemberTypes()) {
+            if (type.tag == TypeTags.ARRAY) {
+                tupleTypes.add(((BArrayType) type).eType);
+            } else {
+                tupleTypes.add(type);
+            }
+        }
+        BUnionType tupleElementType = BUnionType.create(null, tupleTypes);
+        findTypeParam(pos, expType.eType, tupleElementType, env, resolvedTypes, result);
+    }
+
+    private void findTypeParamInUnionForMap(DiagnosticPos pos, BMapType expType, BUnionType actualType,
+                                            SymbolEnv env, HashSet<BType> resolvedTypes, FindTypeParamResult result) {
+        LinkedHashSet<BType> tupleTypes = new LinkedHashSet<>();
+        for (BType type : actualType.getMemberTypes()) {
+            if (type.tag == TypeTags.MAP) {
+                tupleTypes.add(((BMapType) type).constraint);
+            } else {
+                tupleTypes.add(type);
+            }
+        }
+        BUnionType tupleElementType = BUnionType.create(null, tupleTypes);
+        findTypeParam(pos, expType.constraint, tupleElementType, env, resolvedTypes, result);
+    }
+
 
     private void findTypeParamInRecord(DiagnosticPos pos, BRecordType expType, BRecordType actualType, SymbolEnv env,
                                        HashSet<BType> resolvedTypes, FindTypeParamResult result) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
@@ -314,8 +314,8 @@ public class TypeParamAnalyzer {
                                                 result);
                 }
                 if (actualType.tag == TypeTags.UNION) {
-                    findTypeParamInUnion(pos, ((BMapType) expType).constraint, (BUnionType) actualType, env, resolvedTypes,
-                                         result);
+                    findTypeParamInUnion(pos, ((BMapType) expType).constraint, (BUnionType) actualType, env,
+                                         resolvedTypes, result);
                 }
                 return;
             case TypeTags.STREAM:
@@ -323,6 +323,7 @@ public class TypeParamAnalyzer {
                     findTypeParamInStream(pos, ((BStreamType) expType), ((BStreamType) actualType), env, resolvedTypes,
                                           result);
                 }
+                // TODO : Handle unions
                 return;
             case TypeTags.TUPLE:
                 if (actualType.tag == TypeTags.TUPLE) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
@@ -330,6 +330,9 @@ public class TypeParamAnalyzer {
                     findTypeParamInTuple(pos, (BTupleType) expType, (BTupleType) actualType, env, resolvedTypes,
                                          result);
                 }
+                if (actualType.tag == TypeTags.UNION) {
+                    findTypeParamInUnion(pos, expType, (BUnionType) actualType, env, resolvedTypes, result);
+                }
                 return;
             case TypeTags.RECORD:
                 if (actualType.tag == TypeTags.RECORD) {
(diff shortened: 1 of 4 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -398,6 +405,34 @@ private void findTypeParamInTupleForArray(DiagnosticPos pos, BArrayType expType,
         findTypeParam(pos, expType.eType, tupleElementType, env, resolvedTypes, result);
     }
 
+    private void findTypeParamInUnionForArray(DiagnosticPos pos, BArrayType expType, BUnionType actualType,
+                                              SymbolEnv env, HashSet<BType> resolvedTypes, FindTypeParamResult result) {
+        LinkedHashSet<BType> tupleTypes = new LinkedHashSet<>();
+        for (BType type : actualType.getMemberTypes()) {
+            if (type.tag == TypeTags.ARRAY) {
```
rdhananjaya, line 412: I don't think this else branch will ever get hit when all members of the union are arrays. The same goes for the map case below.

KRVPerera: Removed the else parts : 41a8b883de93e15548d1f90cbc4abeaa63428c69

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/TypeParamTest.java
```diff
@@ -67,13 +67,9 @@ public void testTypeParamNegative() {
...
-
-        // Disabled due to https://github.com/ballerina-platform/ballerina-lang/issues/22137
-        // BAssertUtil.validateError(result, err++, "incompatible types: expected '(int|string)', found 'int'", 130,
-        // 14);
-        // BAssertUtil.validateError(result, err++, "incompatible types: expected '(int|string)', found 'float'", 131,
-        //                           24);
-
+        BAssertUtil.validateError(result, err++, "incompatible types: expected '(int|string)', found 'boolean'", 130,
+                                  14);
+        BAssertUtil.validateError(result, err++, "incompatible types: expected '(int|string)', found 'float'", 131, 24);
```
rdhananjaya, line 72: Shall we add a few more negative test cases with incompatible unions?
Such as on (int[]|string[]|float), (int[]|map<int>)

KRVPerera: Don't think code visit this logic for negative cases? It gives an `undefined function`.
I added some different test cases ec685a4fd8a1d34e0a93b9eb724c0c95f3d51a9c
Any suggestions?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -306,6 +309,10 @@ private void findTypeParam(DiagnosticPos pos, BType expType, BType actualType, S
                     findTypeParamInMapForRecord(pos, (BMapType) expType, (BRecordType) actualType, env, resolvedTypes,
                                                 result);
                 }
+                if (actualType.tag == TypeTags.UNION) {
+                    findTypeParamInUnionForMap(pos, (BMapType) expType, (BUnionType) actualType, env, resolvedTypes,
+                                               result);
+                }
```
MaryamZi, line 315: The issue we are trying to solve is not only for lists and maps. I think you have to handle the union scenarios for pretty much each case here.

For example, for streams also it fails atm.
```ballerina
type Foo record {
    string id;
};

type Bar record {
    int id;
};


public function main() { 
    
    Foo[] fooList = [{id: "1234"}, {id: "5678"}];
        
    stream<Foo> fooStream = fooList.toStream();

    stream<Foo>|stream<Bar> fooBarStream = fooStream;

    record {|Foo|Bar value;|}|error? res = fooBarStream.next(); 
}
```

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/TypeParamTest.java
```diff
@@ -111,4 +109,14 @@ public void testLangLibImports() {
...
 
+    @Test(description = "Tests for type narrowing for union return parameters")
+    public void testTypeNarrowingForUnionReturnParameters() {
+        CompileResult result = BCompileUtil.compile("test-src/type-param/type_param_narrowing_for_union_return.bal");
+        BRunUtil.invoke(result, "testSimpleUnionReturnParameterNarrowing");
+        BRunUtil.invoke(result, "testUnionOfMapsReturnParameterNarrowing");
+        BRunUtil.invoke(result, "testStringIntFloatSimpleAndArrayUnionReturnParameterNarrowing");
+        BRunUtil.invoke(result, "testIntFloatSimpleAndMapUnionReturnParameterNarrowing");
+        BRunUtil.invoke(result, "testIntFloatSimpleArrayMapUnionReturnParameterNarrowing");
+        Assert.assertEquals(result.getErrorCount(), 0);
```
MaryamZi, line 120: I don't think we need this, we won't get to this point if there are errors.

langlib/langlib-test/src/test/resources/test-src/type-param/type_param_test_negative.bal
```diff
@@ -126,7 +126,14 @@ function testInvalidArgForBoundRequiredParam() {
 }
 
 function testInvalidArgOnUnionTypedValue() {
-    int[]|string[] arr = [1, 2];
+    int[] | string[] arr = [1, 2];
```
MaryamZi, line 129: We don't add spaces around the `|` in unions now. Please fix everywhere.

```suggestion
    int[]|string[] arr = [1, 2];
```

langlib/langlib-test/src/test/resources/test-src/type-param/type_param_narrowing_for_union_return.bal
```diff
@@ -0,0 +1,48 @@
+function testSimpleUnionReturnParameterNarrowing() {
```
MaryamZi, line 1: Please add the licence.

langlib/langlib-test/src/test/resources/test-src/type-param/type_param_narrowing_for_union_return.bal
```diff
@@ -0,0 +1,48 @@
...
+    int[]|float[] arr = <int[]>[1, 2];
+    [int, (int|float)][] y = arr.enumerate();
+    assertTrue(true);
```
MaryamZi, line 4: While these are not tests for the langlib methods themselves (`enumerate`, `get`, etc.) IMO it'll still be useful to assert something from the result to verify correct runtime behaviour.

For example here we can assert
```ballerina
assertTrue(y[0][1] is int);
assertEquals(1, y[0][1]);
assertTrue(y[1][1] is int);
assertEquals(2, y[1][1]);
```

Shall we replace the `assertTrue(true)`s with proper assertions for type and value?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -398,6 +405,30 @@ private void findTypeParamInTupleForArray(DiagnosticPos pos, BArrayType expType,
         findTypeParam(pos, expType.eType, tupleElementType, env, resolvedTypes, result);
     }
 
+    private void findTypeParamInUnionForArray(DiagnosticPos pos, BArrayType expType, BUnionType actualType,
+                                              SymbolEnv env, HashSet<BType> resolvedTypes, FindTypeParamResult result) {
+        LinkedHashSet<BType> tupleTypes = new LinkedHashSet<>();
```
MaryamZi, line 410: Can we call this something like `memberTypes` instead? L422 too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -398,6 +405,30 @@ private void findTypeParamInTupleForArray(DiagnosticPos pos, BArrayType expType,
         findTypeParam(pos, expType.eType, tupleElementType, env, resolvedTypes, result);
     }
 
+    private void findTypeParamInUnionForArray(DiagnosticPos pos, BArrayType expType, BUnionType actualType,
```
MaryamZi, line 408: I think we can generalize this and `findTypeParamInUnionForMap`.

```java
    private void findTypeParamInUnion(DiagnosticPos pos, BType expType, BUnionType actualType,
                                      SymbolEnv env, HashSet<BType> resolvedTypes, FindTypeParamResult result) {
        LinkedHashSet<BType> memTypes = new LinkedHashSet<>();
        for (BType type : actualType.getMemberTypes()) {
            switch (type.tag) {
                case TypeTags.ARRAY:
                    memTypes.add(((BArrayType) type).eType);
                    break;
                case TypeTags.MAP:
                    memTypes.add(((BMapType) type).constraint);
                    break;
            }
        }
        BUnionType unionType = BUnionType.create(null, memTypes);
        findTypeParam(pos, expType, unionType, env, resolvedTypes, result);
    }
```

To this we can pass as `expType` the array element type or map constraint type.
e.g.,
```java
findTypeParamInUnion(pos, ((BArrayType) expType).eType, (BUnionType) actualType, env, resolvedTypes, 
                     result);
```

I think this will be the cleaner approach when we're adding checks for the rest of the cases in `findTypeParam` too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -398,6 +405,30 @@ private void findTypeParamInTupleForArray(DiagnosticPos pos, BArrayType expType,
...
+                                              SymbolEnv env, HashSet<BType> resolvedTypes, FindTypeParamResult result) {
+        LinkedHashSet<BType> tupleTypes = new LinkedHashSet<>();
+        for (BType type : actualType.getMemberTypes()) {
+            if (type.tag == TypeTags.ARRAY) {
```
MaryamZi, line 412: What about tuples, and similarly records for maps?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
KRVPerera: I also so some duplicates codes, creating an opportunity for refactoring. Let me think of a way to reduce it. But if we pass `eType` we cannot get member types right ?

```
findTypeParamInUnion(pos, ((BArrayType) expType).eType, (BUnionType) actualType, env, resolvedTypes, 
                     result);
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
KRVPerera: Related to : https://github.com/ballerina-platform/ballerina-lang/issues/22570

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
KRVPerera: I have fixed this please help review.
Records: https://github.com/ballerina-platform/ballerina-lang/pull/22346/commits/d617fd7cae953123c47176d6b7bbb38f2d3fecae
Tuples: https://github.com/ballerina-platform/ballerina-lang/pull/22346/commits/12fc9860180fb75f5c7ff449ac5779cd1a6b1a27

PR #22348, opened 2020-03-31
Add Object method mocking functionality

### LINKED ISSUE #22095: Function mocking for Object functions
**Description:**
Check the compatibility of Object functions with the function mocking implementation

Ibaqu:
## Purpose
Object method mocking functionality is implemented in this PR

Fixes #22095 

## Approach
Iterate through type definitions to verify if the function exists in the Object and it the return types are compatible

During BIR generation, the `callTerminator` of the object method we are mocking is modified. 
`isVirtual` is set to false and the Object reference is removed from the arguments

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
@@ -302,7 +302,7 @@ public class BIRGen extends BLangNodeVisitor {
                     String functionKey = null;
 
                     // Generate the function key
-                    if (callTerminator.isVirtual && callTerminator.args.size() == 1) {
+                    if (callTerminator.isVirtual) {
                         // Function key for object methods
                         String objectPkg = callTerminator.args.get(0).variableDcl.type.toString();
                         functionKey = objectPkg + MOCK_ANNOTATION_DELIMITER + callTerminator.name.toString();
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
@@ -203,6 +203,7 @@ public class BIRGen extends BLangNodeVisitor {
 
     // Required variables for Mock function implementation
     private static final String MOCK_ANNOTATION_DELIMITER = "#";
+    private static final String MOCK_OBJECT_DELIMITER = ":";
 
     public static BIRGen getInstance(CompilerContext context) {
         BIRGen birGen = context.get(BIR_GEN);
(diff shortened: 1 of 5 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -299,14 +300,31 @@ private void replaceMockedFunctions(BIRPackage birPkg, Map<String, String> mockF
...
-                    String functionKey = callTerminator.calleePkg.toString() + MOCK_ANNOTATION_DELIMITER
-                            + callTerminator.name.toString();
-                    if (mockFunctionMap.get(functionKey) != null) {
-                        // Just "get" the reference. If this doesnt work then it doesnt exist
-                        String mockfunctionName = mockFunctionMap.get(functionKey);
-                        callTerminator.name = new Name(mockfunctionName);
-                        callTerminator.calleePkg = function.pos.src.pkgID;
+
+                    // This will be an object function
+                    if (callTerminator.isVirtual && callTerminator.args.size() == 1) {
```
Ibaqu: Is this a viable way of identifying an object function? Are there better alternatives?

Either way, if it is not an object function, the generated `objectKey` will not be able to pull the relevant mock function from the mockFunctionMap and the program flow will continue without any changes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -299,14 +300,31 @@ private void replaceMockedFunctions(BIRPackage birPkg, Map<String, String> mockF
...
+                        String objectPkg = callTerminator.args.get(0).variableDcl.type.toString();
+                        String objectKey = objectPkg + MOCK_ANNOTATION_DELIMITER + callTerminator.name.toString();
```
Ibaqu: The generated object key will look something like this : 

`org/<module_name>:<version>:<objectName>#<functionName>`.

This is the same format as the key added to the `mockFunctionMap` in the testAnnotationProcessor class

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -297,16 +298,32 @@ private void replaceMockedFunctions(BIRPackage birPkg, Map<String, String> mockF
...
+                    // Get the mock function from the Mock function map and replace where necessary
                     if (mockFunctionMap.get(functionKey) != null) {
-                        // Just "get" the reference. If this doesnt work then it doesnt exist
-                        String mockfunctionName = mockFunctionMap.get(functionKey);
-                        callTerminator.name = new Name(mockfunctionName);
+                        String mockFunctionName = mockFunctionMap.get(functionKey);
+                        callTerminator.name = new Name(mockFunctionName);
                         callTerminator.calleePkg = function.pos.src.pkgID;
+
+                        if (callTerminator.isVirtual) {
```
warunalakshitha, line 321: Do we need this check. I think line 305 covers it.

Ibaqu: Actually this is required since this block of code is shared for both types of methods. If this check is not there, then it will process all functions like it is an object method.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -297,16 +298,32 @@ private void replaceMockedFunctions(BIRPackage birPkg, Map<String, String> mockF
...
                 BIRTerminator bbTerminator = functionBasicBlock.terminator;
                 if (bbTerminator.kind.equals(InstructionKind.CALL)) {
-                    //We get the callee and the name and generate 'calleepackage#name'
                     BIRTerminator.Call callTerminator = (BIRTerminator.Call) bbTerminator;
-                    String functionKey = callTerminator.calleePkg.toString() + MOCK_ANNOTATION_DELIMITER
-                            + callTerminator.name.toString();
+                    String functionKey = null;
+
+                    // Generate the function key
+                    if (callTerminator.isVirtual && callTerminator.args.size() == 1) {
```
SupunS, line 305: Do we need this second check? What if the object-method call has more than one argument?

misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestAnnotationProcessor.java
```diff
@@ -312,16 +329,41 @@ public void process(FunctionNode functionNode, List<AnnotationAttachmentNode> an
...
+
+                    // Look thru all the existing definitions for objectName
+                    for (BLangTypeDefinition objectBlangType : typeDefinitionList) {
+                        // If the correct type definition is found, we need to look thru its list of function
+                        if (objectBlangType.getName().toString().equals(objectName)) {
+                            List<BLangFunction> objectMethods =
+                                    ((BLangObjectTypeNode) objectBlangType.typeNode).getFunctions();
+                            // Check every method for the matching function name
+                            for (BLangFunction objectMethod : objectMethods) {
+                                if (objectMethod.getName().toString().equals(functionName)) {
```
SupunS, line 356: Can we reduce the depth? 

Also can we cache the funcNames, without iterating through everything every-time?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
Ibaqu: The argument considered in this case is the object reference we are using to call the object function. ie. p1.getName() for an object Person p1, the argument, in this case, is p1. 

I can remove the second check for size, but there are other virtual functions like `__init` that would get caught in this block then. When I add the size check, only the object methods are considered.

Ibaqu: You are correct. The size check is not working when arguments are considered

Ibaqu: Removed the check for arguments. It looks only for isVirtual

misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestAnnotationProcessor.java
Ibaqu: Im trying to reduce the program depth by generating a map with all of the object types and associated functions. 
This way I can just pass the Object name and the function name to a function and it can return the required type. Would this implementation suffice?

Ibaqu: I have refactored the code to introduce the concept I have mentioned above.

PR #22365, opened 2020-04-01
[1.2.x] Narrow type bound to union type parameters

### LINKED ISSUE #16828: Type bound to type param needs to be narrowed down for union types

### LINKED ISSUE #22137: Incorrect type-checking for push on union-typed list

KRVPerera:
Fixes #16828
Fixes #22137

## Approach
Add logic to TypeParamAnalyser. Union and union with arrays will be flattened and merged as a union again.

#### Case 1
```Ballerina
int[]|Person[] arr = <int[]>[1, 2];
[int, int][] y = arr.enumerate();
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
@@ -278,7 +278,6 @@ public class TypeParamAnalyzer {
             updateTypeParamAndBoundType(pos, env, expType, actualType);
 
             // If type param discovered before, now type check with actual type. It has to be matched.
-
             if (checkContravariance) {
                 types.checkType(pos, getMatchingBoundType(expType, env, new HashSet<>()), actualType,
                                 DiagnosticCode.INCOMPATIBLE_TYPES);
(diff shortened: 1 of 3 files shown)


rdhananjaya:
@KRVPerera this change is not patch compatible 
Consider this:
```ballerinia
    int[]|Person[] arr = <int[]>[1, 2];
    var y = arr.enumerate();
    if (y is [int, int|Person][]) { //
        io:println("true");
    }
```

This code will break after your change because after your change the type test is always true. Shall we close this PR on the patch branch?

KRVPerera:
Closing on patch as backward incompatible

Issue #22350, opened 2020-03-31
`.slice()` on a tuple with a rest-descriptor fails at runtime


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
public function main() {
    [int, string...] x = [1, "hello", "world"];

    (int|string)[] a = x.slice(1);
}
```

```cmd
$ ballerina run xyz.bal 
Compiling source
        xyz.bal
Running executables

error: {ballerina/lang.array}InherentTypeViolation message=incompatible types: expected 'int', found 'string'
        at ballerina.lang_array:slice(array.bal:106)
           xyz:main(xyz.bal:4)
```

This is because the type of the sliced array is only computed based on the members, and does not consider the rest-descriptor.

https://github.com/ballerina-platform/ballerina-lang/blob/v1.2.0/langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Slice.java#L82

So, in this scenario, the sliced array is created as an `int[]`, to which we then try to add strings.

**Affected Versions:**
1.2.0

pubudu91:
Closing since this was fixed with #22373

PR #22373, opened 2020-04-01
Fix slice on tuple with rest descriptor

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/22350

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Slice.java
```diff
@@ -70,18 +73,24 @@ public static ArrayValue slice(Strand strand, ArrayValue arr, long startIndex, l
...
-                BUnionType unionType = new BUnionType(tupleType.getTupleTypes(), tupleType.getTypeFlags());
+
+                List<BType> memTypes = new ArrayList<>(tupleType.getTupleTypes());
+
+                BType restType = tupleType.getRestType();
+                if (restType != null) {
+                    memTypes.add(restType);
+                }
+
+                BUnionType unionType = new BUnionType(memTypes);
```
pubudu91, line 91: Shouldn't the new type have the same flags as the original type?

MaryamZi: Not really. For the array type maybe we could use the same but not for the union type.

The flags are `NILABLE`, `ANYDATA`, `PURE`.

Consider the tuple type `[int, string?, error...]`. The is `PURE` and `ANYDATA`.

The union created for this is  `int|string|error?`. This is `NILABLE` and `PURE`.

This `BUnionType` constructor computes the new flags - https://github.com/ballerina-platform/ballerina-lang/blob/ballerina-1.2.x/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BUnionType.java#L56.

PR #22439, opened 2020-04-03
Add 'never' type descriptor

### LINKED ISSUE #22099: Add never-type-descriptor
Copied from Spec.
```
never-type-descriptor := never
```
The type descriptor any describes the type that does not contain any shapes. No value ever belongs to the never.

This can be useful to describe for the return type of a function, if the function never returns. It can also be useful as a type parameter. For example, xml<never> describes the an xml type that has no constituents, i.e. the empty xml value.

Note that for anytype T, the type T|never is the same as T.

dulajdilshan:
## Purpose
Add `never` type descriptor

> `never-type-descriptor := never`
> 
> The type descriptor never describes the type that does not contain any shapes. No value ever belongs to the never.
> 
> This can be useful to describe for the return type of a function, if the function never returns. It can also be useful as a type parameter. For example, xml<never> describes the an xml type that has no constituents, i.e. the empty xml value.
> 
> Note that for anytype T, the type T|never is the same as T.

Fixes #22099 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BType.java
```diff
@@ -25,15 +25,7 @@
...
-import static org.wso2.ballerinalang.compiler.util.TypeTags.BOOLEAN;
-import static org.wso2.ballerinalang.compiler.util.TypeTags.BYTE;
-import static org.wso2.ballerinalang.compiler.util.TypeTags.DECIMAL;
-import static org.wso2.ballerinalang.compiler.util.TypeTags.ERROR;
-import static org.wso2.ballerinalang.compiler.util.TypeTags.FLOAT;
-import static org.wso2.ballerinalang.compiler.util.TypeTags.INT;
-import static org.wso2.ballerinalang.compiler.util.TypeTags.NIL;
-import static org.wso2.ballerinalang.compiler.util.TypeTags.STRING;
-import static org.wso2.ballerinalang.compiler.util.TypeTags.TYPEDESC;
+import static org.wso2.ballerinalang.compiler.util.TypeTags.*;
```
dulvinw, line 28: We dont import * right? I think this should be caught by checkstyle

compiler/ballerina-lang/src/main/resources/grammar/BallerinaParser.g4
```diff
@@ -72,7 +72,7 @@ functionDefinitionBody
     |   externalFunctionBody SEMICOLON
     ;
 
-functionDefinition
+functionD   efinition
```
dulvinw, line 75: This is a typo right?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1121,6 +1121,9 @@ warning.usage.of.deprecated.construct=\
 warning.non.module.qualified.error.reason=\
   error reason ''{0}'' is not module qualified
 
+warning.function.does.not.return=\
+  This function doesn't return, hence use 'returns never' here
```
dulvinw, line 1125: I think a better mesaage would be missing ‘returns never’ in the definition of function ‘’{0}’’. WDYT?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNeverType.java
```diff
@@ -0,0 +1,23 @@
+package org.wso2.ballerinalang.compiler.semantics.model.types;
+
+import org.wso2.ballerinalang.compiler.util.Names;
+import org.wso2.ballerinalang.compiler.util.TypeTags;
+
+/**
+ * {@code BNeverType} represents the singleton type when functions don't have return value.
```
dulvinw, line 7: Function don’t have ‘a’ return value.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolTable.java
```diff
@@ -28,31 +28,7 @@
...
-import org.wso2.ballerinalang.compiler.semantics.model.types.BStreamType;
-import org.wso2.ballerinalang.compiler.semantics.model.types.BStringSubType;
-import org.wso2.ballerinalang.compiler.semantics.model.types.BTableType;
-import org.wso2.ballerinalang.compiler.semantics.model.types.BTupleType;
-import org.wso2.ballerinalang.compiler.semantics.model.types.BType;
-import org.wso2.ballerinalang.compiler.semantics.model.types.BTypedescType;
-import org.wso2.ballerinalang.compiler.semantics.model.types.BUnionType;
-import org.wso2.ballerinalang.compiler.semantics.model.types.BXMLSubType;
-import org.wso2.ballerinalang.compiler.semantics.model.types.BXMLType;
+import org.wso2.ballerinalang.compiler.semantics.model.types.*;
```
dulvinw, line 31: Import of *

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/antlr4/BallerinaParserListener.java
```diff
@@ -1,4 +1,4 @@
-// Generated from BallerinaParser.g4 by ANTLR 4.5.3
+// Generated from /home/dulaj/projects/ballerina-lang/compiler/ballerina-lang/src/main/resources/grammar/BallerinaParser.g4 by ANTLR 4.5.3
```
dulvinw, line 1: Seems like this is not the standard

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/antlr4/BallerinaParserBaseListener.java
```diff
@@ -1,4 +1,4 @@
-// Generated from BallerinaParser.g4 by ANTLR 4.5.3
+// Generated from /home/dulaj/projects/ballerina-lang/compiler/ballerina-lang/src/main/resources/grammar/BallerinaParser.g4 by ANTLR 4.5.3
```
dulvinw, line 1: Here as well 😊

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BType.java
dulajdilshan: i'll change it

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/antlr4/BallerinaParserListener.java
dulajdilshan: oops

MaryamZi:
The scenario I tried to mention was

```ballerina
type Foo record {
    int i;
    never j;
};

type Bar record {
    int i;
    string s;
};

public function main() {
    Foo f = {i: 1};

    Bar b = {s: "str", j: "j-val", ...f}; // this should not fail with a duplicate key `j` error
}
```

MaryamZi:
I think the following should also fail, probably at runtime though.

```ballerina
type Foo record {
    int i;
    never j;
};

type Bar record {
    int i;
};

public function main() {
    Bar b = {j: "j-val", i: 1};

    // `j` via spread field
    Foo f1 = {...b};

    string s = "j";
    // `j` via computed key
    Foo f2 = {i: 1, [s]: "j-val"};
}
```

Looping in @hasithaa too.

pubudu91:
I see that there are some unnecessary files committed. Refrain from committing the files in the ` compiler/ballerina-lang/src/main/gen` directory. Since you only have one commit here, should be easy to get rid of that and keep it out of the history

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -1574,6 +1574,12 @@ public void visit(BLangThrow throwNode) {
 
     public void visit(BLangPanic panicNode) {
         this.checkStatementExecutionValidity(panicNode);
+
+        // Check whether the enclosed function's return type is 'nil'
+        if (this.env.enclInvokable.returnTypeNode.type.tag == TypeTags.NIL) {
+            dlog.warning(this.env.enclInvokable.returnTypeNode.pos, DiagnosticCode.FUNCTION_DOES_NOT_RETURN);
```
pubudu91, line 1580: Is this only applicable for functions returning `()`? I was under the impression that this applies to any function in which a panic is used. e.g.,
```ballerina
function foo() returns string|never {
    // some code
    
    if (true) {
        panic error("ERROR");
    }

    // more code

    return "foo";
}
```

pubudu91: @hasithaa thoughts?

compiler/ballerina-lang/src/main/resources/compiler.properties
KRVPerera: +adding
- I think it is a practice not to use short forms like "doesn't"
- does not start with a capital letter
- being a warning rather than an error type, the message looks a bit forceful. I believe the idea of giving a warning is maybe because the user forgot to add `returns never`. It can also be the case that this is not intentional. So the warning should guide the user to find the actual solution

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
dulajdilshan: Removed this part as discussed

tests/jballerina-unit-test/src/test/resources/testng.xml
```diff
@@ -20,7 +20,7 @@
 
 <!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
 
-<suite name="ballerina-root-test-suite" time-out="120000">
+<suite name="ballerina-test-suite" time-out="120000">
```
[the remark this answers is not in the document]
dulajdilshan: Ack. Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNeverType.java
```diff
@@ -0,0 +1,45 @@
...
+import org.wso2.ballerinalang.compiler.util.TypeTags;
+
+/**
+ * {@code BNeverType} represents the singleton type when functions don't have a return value.
+ * The value of the {@code BNeverType} is written as 'never'
+ *
+ * @since 1.2.3
+ */
+
+public class BNeverType extends BNilType {
```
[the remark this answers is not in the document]
dulajdilshan: Never is a subtype of null. So, shouldn't it be `extends BNilType implements NeverType`?
* Here I have ignored NeverType interface. I'll add it later

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNilType.java
```diff
@@ -35,6 +35,10 @@ public BNilType() {
         super(TypeTags.NIL, null, Flags.READONLY);
     }
 
+    protected BNilType(int tag) {
```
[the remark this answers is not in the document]
dulajdilshan: But It is a `protected` constructor. So only accessible by the classes within the package or sub classes

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -1037,6 +1037,10 @@ public void visit(BLangSimpleVariable varNode) {
             symbol.params = tsymbol.params;
             symbol.restParam = tsymbol.restParam;
             symbol.retType = tsymbol.returnType;
+        } else if (varSymbol.type.tag == TypeTags.NEVER && ((env.scope.owner.tag & SymTag.RECORD) != SymTag.RECORD)) {
```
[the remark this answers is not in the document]
dulajdilshan: This is to check if the variable is defined as a `never-type` and is not a field of a record

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2432,6 +2433,15 @@ private void validateDefaultable(BLangRecordTypeNode recordTypeNode) {
         }
     }
 
+    private void validateOptionalNeverTypedField(BLangRecordTypeNode recordTypeNode) {
+        for (BLangSimpleVariable field : recordTypeNode.fields) {
+            if (field.type.tag == TypeTags.NEVER && !field.flagSet.contains(Flag.OPTIONAL)) {
+                dlog.error(field.pos, DiagnosticCode.NEVER_TYPE_NOT_ALLOWED_FOR_REQUIRED_FIELDS, field.name.value);
+//            }
```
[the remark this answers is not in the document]
dulajdilshan: Ack

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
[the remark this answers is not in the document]
KRVPerera: it can be a separate if block right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNeverType.java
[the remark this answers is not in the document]
KRVPerera: I believe subtyping rules are defined in assignability rules. I don't think we need to map subtyping between `never` and `nil` in a class hierarchy when we model. If we don't extend `BNilType` we don't have to create an extra constructor for `BNilType`. I don't think we need a `NeverType` interface too. Only if in future we need to implement the behaviour of a NeverType to another type. Check with others as well.

This is my personal opinion:
If `never` is a subtype of `nil`, the shape of the `never` type should be smaller than `nil`. Extending `nil` model does not agree with that. `nil` has a value but `never` cannot have a value. I also believe when we create a class it should only have minimum things to help behave like self. Any other extra things just increase the class size without any value addition. For two functions with `never` return type, we need to create two objects of these during compilation. Isn't it better if it is as small as possible?

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
```diff
@@ -356,6 +356,10 @@
     USAGE_OF_DEPRECATED_CONSTRUCT("usage.of.deprecated.construct"),
     OPERATOR_NOT_SUPPORTED("operator.not.supported"),
     OPERATOR_NOT_ALLOWED_VARIABLE("operator.not.allowed.variable"),
+    FUNCTION_DOES_NOT_RETURN("function.does.not.return"),
```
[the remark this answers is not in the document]
dulajdilshan: Ack

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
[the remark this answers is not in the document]
dulajdilshan: Yeah. Thanks

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNeverType.java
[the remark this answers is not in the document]
dulajdilshan: @hasithaa  @pubudu91 @MaryamZi Any thoughts ?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNilType.java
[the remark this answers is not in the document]
KRVPerera: okay

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -884,6 +884,7 @@ public void visit(BLangUnionTypeNode unionTypeNode) {
                         memBType.tag == TypeTags.UNION && !Symbols.isFlagOn(memBType.tsymbol.flags, Flags.TYPE_PARAM) ?
                                 ((BUnionType) memBType).getMemberTypes().stream() :
                                 Stream.of(memBType))
+                .filter(memBType -> memBType.tag != TypeTags.NEVER)
```
pubudu91, line 887: Instead of handling this here, I think it'd be better to add this logic to the API for creating unions in `BUnionType` itself. That way, this behaviour will be guaranteed for any union created, not just the ones created through the resolver.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1184,6 +1190,12 @@ warning.usage.of.deprecated.construct=\
 warning.non.module.qualified.error.reason=\
   error reason ''{0}'' is not module qualified
 
+warning.function.does.not.return=\
```
KRVPerera, line 1193: Can we remove this ```warning.function.does.not.return```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/util/TypeTags.java
```diff
@@ -116,6 +117,7 @@ public static boolean isXMLTypeTag(int tag) {
             case XML_PI:
             case XML_COMMENT:
             case XML_TEXT:
+            case NEVER:
```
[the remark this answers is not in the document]
dulajdilshan: Reverted

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNeverType.java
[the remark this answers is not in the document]
dulajdilshan: Resolved

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNilType.java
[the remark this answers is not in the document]
dulajdilshan: Resolved

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNeverType.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public BNeverType() {
+        super(TypeTags.NEVER);
+    }
+
+    @Override
+    public boolean isNullable() {
```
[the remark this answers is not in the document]
dulajdilshan: Never return typed functions doesn't return anything. So we need to pass `true` for `isNullable()`. Don't we ?

compiler/ballerina-lang/src/main/resources/compiler.properties
dulajdilshan: sure

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
dulajdilshan: Ack

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNeverType.java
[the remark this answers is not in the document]
KRVPerera: Yes, this is needed now, as we extend BType

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -108,7 +108,9 @@ public void setNullable(boolean nullable) {
      * @return The created union type.
      */
     public static BUnionType create(BTypeSymbol tsymbol, LinkedHashSet<BType> types) {
-        LinkedHashSet<BType> memberTypes = toFlatTypeSet(types);
+        LinkedHashSet<BType> memberTypes = toFlatTypeSet(types).stream()
+                .filter(memBType -> memBType.tag != TypeTags.NEVER)
+                .collect(Collectors.toCollection(LinkedHashSet::new));
```
pubudu91, line 113: Not exactly related to this PR. But can you convert the stream API usages in this class to loops as well? Should be straight forward with IntelliJ

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BNeverType.java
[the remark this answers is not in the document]
dulajdilshan: Thanks

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
dulajdilshan: Okay

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BNeverType.java
```diff
@@ -0,0 +1,40 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.ballerinalang.jvm.types;
+
+/**
+ * {@code BNeverType} represents the type of a {@code Never}.
+ *
+ * @since 1.3.0
```
pubudu91, line 23: Let's update this version number across all new files

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never-type-negative.bal
```diff
@@ -0,0 +1,41 @@
+function testNeverTypedVariable() {
```
pubudu91, line 1: Shall we add tests for the following scenarios as well?

- In a function with a `never` return type,
  - explicitly return `()`
  - explicitly return some other value (say, a `string`)

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never-type.bal
```diff
@@ -0,0 +1,71 @@
+import ballerina/lang.'xml;
```
pubudu91, line 1: Shall we add a test case for the following as well: 

>  A table shape satisfies a key-constraint key<K> if and and only if its set of key value shapes are a subset of K. The shape of a keyless table satisfies the key-constraint key<never>.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2432,6 +2433,15 @@ private void validateDefaultable(BLangRecordTypeNode recordTypeNode) {
         }
     }
 
+    private void validateOptionalNeverTypedField(BLangRecordTypeNode recordTypeNode) {
+        // Never type is only allowed in an optional field in a record
+        for (BLangSimpleVariable field : recordTypeNode.fields) {
+            if (field.type.tag == TypeTags.NEVER && !field.flagSet.contains(Flag.OPTIONAL)) {
+                dlog.error(field.pos, DiagnosticCode.NEVER_TYPE_NOT_ALLOWED_FOR_REQUIRED_FIELDS, field.name.value);
```
pubudu91, line 2440: This semantic rule about only optional fields being allowed to be `never` type fields is not included in the spec right? It's something we discussed? If so, will have to bring this up with James and add it to the spec

pubudu91:
I think the following should fail at compile time:
```ballerina
import ballerina/io;

public function main() {
    Person p = {name: "Pubudu"};
    p["foo"] = ();
    io:println(p);
}

type Person record {|
    string name;
    never...;
|};
```
Currently the above fails at runtime with the following message:
```
error: {ballerina/lang.map}InherentTypeViolation message=invalid value for record field 'foo': expected value of type 'never', found '()'
	at test:main(test.bal:5)
```

pubudu91:
I think it's better to add a few balo test cases as well, to ensure that the `never` type is getting written/read back correctly

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never-type.bal
```diff
@@ -0,0 +1,71 @@
...
+//------------- Testing XML<never> -----------------
+
+function testXMLWithNeverType() {
+    xml<never> x = <xml<never>> 'xml:concat();  //generates an exmpty XML sequence and assign it to XML<never>
+}
+
+
+//---------------Test 'never' types with 'union-type' descriptors
+function testNeverWithUnionType1() {
+    string|never j;
```
pubudu91, line 61: Should this be allowed if we are not allowing `never x;`?

pubudu91:
A few more cases to consider:
```ballerina
future<never> f;
    
map<never> m = {};
m["a"] = ();
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BNeverType.java
dulajdilshan: ack

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1181,6 +1187,9 @@ warning.usage.of.deprecated.construct=\
 warning.non.module.qualified.error.reason=\
   error reason ''{0}'' is not module qualified
 
+error.invalid.never.return.typed.function.invocation=\
+  invalid invocation. function ''{0}'' returns never, hence only allowed to invoke in call statements
```
pubudu91, line 1191: How about something like 
```
invalid invocation ''{0}()'': functions/methods returning ''never'' cannot be called in an expression context
```

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never-type.bal
dulajdilshan: But the spec says

> for anytype T, the type T|never is the same as T.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
dulajdilshan: We discussed about it . But, it is not included in the spec.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -371,6 +371,12 @@ error.missing.required.record.field=\
 error.default.values.not.allowed.for.optional.fields=\
   a default value specified for optional field ''{0}''
 
+error.never.type.not.allowed.for.required.fields=\
+  field ''{0}'' is defined as a never-typed required field
+
+error.never.type.not.allowed.as.variable.type=\
+  variable ''{0}'' is not allowed to have type ''never''
```
pubudu91, line 378: How about,
```
''never''-typed variable definitions not allowed
```
or
```
defining variables of type ''never'' not allowed
```
or 
```
variables of type ''never'' cannot be defined
```

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never-type-negative.bal
dulajdilshan: Ack

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never-type.bal
dulajdilshan: Ack

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -371,6 +371,12 @@ error.missing.required.record.field=\
 error.default.values.not.allowed.for.optional.fields=\
   a default value specified for optional field ''{0}''
 
+error.never.type.not.allowed.for.required.fields=\
+  field ''{0}'' is defined as a never-typed required field
```
pubudu91, line 375: Better to define error messages so that it clearly conveys that what's being attempted is not allowed, and if possible, what to use instead.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -3971,6 +3971,11 @@ private void checkFunctionInvocationExpr(BLangInvocation iExpr) {
             // This will enable, type param support, if the function is called directly.
             this.env = SymbolEnv.createInvocationEnv(iExpr, this.env);
         }
+        // Log an error if the function returns never and invoked invalidly.
+        if (funcSymbol.type.getReturnType().tag == TypeTags.NEVER &&
+                env.node.getKind() != NodeKind.EXPRESSION_STATEMENT) {
```
pubudu91, line 3976: Should this be in the code analyzer? Not exactly a type checking logic right?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BNeverType.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public class BNeverType extends BNullType {
+    /**
+     * Create a {@code BNeverType} represents the type of a {@code Never}.
+     *
+     * @param typeName string name of the type
+     * @param pkg      package path
+     */
+    BNeverType(String typeName, BPackage pkg) {
```
pubudu91, line 32: No need to parameterize the name right? Since the name won't change at all

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never-type.bal
pubudu91: That would be useful for cases like the following:
```ballerina
public function main() {
    int x = foo();
}

function foo() returns int|never {
    return 10;
}
```

pubudu91: let's check with the others as well

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
pubudu91: Can you put a comment in the spec issue explaining the scenario and the conclusion we came to?

compiler/ballerina-lang/src/main/resources/compiler.properties
dulajdilshan: Seems like
`''never''-typed variable definitions not allowed`
fits better

compiler/ballerina-lang/src/main/resources/compiler.properties
dulajdilshan: How about this ?
`'never' type is not allowed as a required field, use optional field declaration for ''{0}'' instead`

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BNeverType.java
dulajdilshan: Ack

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
dulajdilshan: Will change this

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -420,10 +420,11 @@ private void visitFunction(BLangFunction funcNode) {
         if (funcNode.body != null) {
             analyzeNode(funcNode.body, invokableEnv);
 
-            boolean isNilableReturn = funcNode.symbol.type.getReturnType().isNullable();
+            boolean isNeverOrNilableReturn = funcNode.symbol.type.getReturnType().tag == TypeTags.NEVER ||
+                    funcNode.symbol.type.getReturnType().isNullable();
             // If the return signature is nil-able, an implicit return will be added in Desugar.
             // Hence this only checks for non-nil-able return signatures and uncertain return in the body.
-            if (!isNilableReturn && !this.statementReturns) {
+            if (!(isNeverOrNilableReturn) && !this.statementReturns) {
```
MaryamZi, line 427:
```suggestion
            if (!isNeverOrNilableReturn && !this.statementReturns) {
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2009,6 +2010,14 @@ public void visit(BLangInvocation invocationExpr) {
...
 
+        // Log an error if the function returns never and invoked invalidly.
+        if (invocationExpr.symbol != null &&
+                ((BInvokableSymbol) invocationExpr.symbol).getReturnType().tag == TypeTags.NEVER &&
+                invocationExpr.parent.getKind() != NodeKind.EXPRESSION_STATEMENT) {
+            dlog.error(invocationExpr.pos, DiagnosticCode.INVALID_NEVER_RETURN_TYPED_FUNCTION_INVOCATION,
+                    invocationExpr.name);
+        }
+
         if ((invocationExpr.symbol != null) && invocationExpr.symbol.kind == SymbolKind.FUNCTION) {
```
MaryamZi, line 2021: We seem to be repeating the `invocationExpr.symbol != null` check. Can we check if it null and return if so before L2014? Then we can check the rest.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2221,14 +2221,21 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
-                        || (s.tag == TypeTags.FINITE  && isAssignable(s, target, unresolvedTypes,
-                                                                      unresolvedReadonlyTypes))
-                        || (s.tag == TypeTags.XML
-                            && isAssignableToUnionType(expandedXMLBuiltinSubtypes, target, unresolvedTypes,
-                                                       unresolvedReadonlyTypes)));
+        for (BType s : sourceTypes) {
+            if (s.tag == TypeTags.NEVER) {
+                continue;
+            }
+            if ((targetTypes.stream().noneMatch(t -> isAssignable(s, t, unresolvedTypes,
```
MaryamZi, line 2228: We can replace this stream too right? Formatting seems off btw.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -371,6 +371,12 @@ error.missing.required.record.field=\
 error.default.values.not.allowed.for.optional.fields=\
   a default value specified for optional field ''{0}''
 
+error.never.type.not.allowed.for.required.fields=\
+  'never' typed required fields not allowed, define ''{0}'' as an optional field instead
```
MaryamZi, line 375: How about

```suggestion
  a required field cannot be of type ''never'', define ''{0}'' as an optional field instead
```

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -371,6 +371,12 @@ error.missing.required.record.field=\
...
+
+error.never.typed.var.def.not.allowed=\
+  ''never''-typed variable definitions not allowed
```
MaryamZi, line 378:
```suggestion
  cannot define a variable of type ''never''
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/never/NeverTypeTest.java
```diff
@@ -0,0 +1,145 @@
...
+        BAssertUtil.validateError(negativeCompileResult, i++,
+                "incompatible types: expected 'never', found 'string'", 25, 12);
+        BAssertUtil.validateError(negativeCompileResult, i++,
+                "incompatible types: expected 'never', found 'string'", 31, 16);
+        BAssertUtil.validateError(negativeCompileResult, i++,
+                "incompatible types: expected 'never', found '()'", 36, 12);
+        BAssertUtil.validateError(negativeCompileResult, i++,
+                "incompatible types: expected 'never', found '()'", 41, 20);
+        BAssertUtil.validateError(negativeCompileResult, i++,
+                "never typed required fields not allowed, define 'y' as an optional field instead", 49, 5);
```
MaryamZi, line 121: The error message should've been 
```suggestion
                "'never' typed required fields not allowed, define 'y' as an optional field instead", 49, 5);
```
right?

dulajdilshan: Yeah. :grimacing: My bad

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
dulajdilshan: There was an error occurred in the integration tests. I changed the behaviour

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BNeverType.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @since 2.0.0-preview1
+ */
+public class BNeverType extends BNullType {
+    /**
+     * Create a {@code BNeverType} represents the type of a {@code Never}.
+     *
+     * @param pkg package path
+     */
+    BNeverType(BPackage pkg) {
```
pubudu91, line 31: Why is this `BPackage` param required? `BNullType`'s constructor accepts two `String` params`

pubudu91: Ah sorry, nvm. I was looking at the wrong class.

PR #22494, opened 2020-04-06
Fix type checking against `error[]` and `any[]`

### LINKED ISSUE #22410: Array of error is allowed to assigned to array of any
**Description:**
As `error` is not in `any` we should not allow this.

```ballerina
public function main() {
    var f = foo();
}

function foo () returns any[] {
    int[]|string[]|error[] k = <error[]>[error("reason")];
    error[] ea = [];
    any[] j = ea;
    return k;
}
```

**Steps to reproduce:**

**Affected Versions:**
1.2.0

KRVPerera:
## Purpose
Proper type checking logic is not happening when trying to assign `error[]` to `any[]`.

Fixes #22410

## Approach
Type checking logic is added to filter out error type. Existing code let any type or array to be assigned to `any` type array.

## Samples
```Ballerina
error[] ea = [];
any[] j = ea;
```

```LogTalk
error: .::error_test_negative.bal:127:15: incompatible types: expected 'any[]', found 'error[]'
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -570,10 +570,6 @@ public class Types {
         }
 
         if (targetTag == TypeTags.JSON) {
-            if (sourceTag == TypeTags.JSON) {
-                return true;
-            }
-
             if (sourceTag == TypeTags.ARRAY) {
                 return isArrayTypesAssignable((BArrayType) source, target, unresolvedTypes);
             }
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -832,21 +832,22 @@ public class Types {
     }
 
     public boolean isArrayTypesAssignable(BArrayType source, BType target, Set<TypePair> unresolvedTypes) {
+        BType sourceElementType = source.getElementType();
         if (target.tag == TypeTags.ARRAY) {
             // Both types are array types
             BArrayType lhsArrayType = (BArrayType) target;
-            if (source.state == BArrayState.UNSEALED) {
-                return isAssignable(source.eType, lhsArrayType.eType, unresolvedTypes);
+            if (lhsArrayType.state == BArrayState.UNSEALED) {
+                return isAssignable(sourceElementType, lhsArrayType.eType, unresolvedTypes);
             }
             return checkSealedArraySizeEquality(source, lhsArrayType)
-                    && isAssignable(source.eType, lhsArrayType.eType, unresolvedTypes);
+                    && isAssignable(sourceElementType, lhsArrayType.eType, unresolvedTypes);
         }
 
         if (target.tag == TypeTags.UNION) {
-            return isAssignable(source.eType, target, unresolvedTypes);
+            return isAssignable(sourceElementType, target, unresolvedTypes);
         }
 
-        BType sourceElementType = source.getElementType();
+
         // If the target type is a JSON, then element type of the rhs array
         // should only be a JSON supported type.
         if (target.tag == TypeTags.JSON) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -845,18 +845,19 @@ public class Types {
         } else if (source.tag == TypeTags.ARRAY) {
             // Only the right-hand side is an array type
 
+            if (target.tag == TypeTags.UNION) {
+                return isAssignable(source, target, unresolvedTypes);
+            }
+
+            BType sourceElementType = ((BArrayType) source).getElementType();
             // If the target type is a JSON, then element type of the rhs array
             // should only be a JSON supported type.
             if (target.tag == TypeTags.JSON) {
-                return isAssignable(((BArrayType) source).getElementType(), target, unresolvedTypes);
-            }
-
-            if (target.tag == TypeTags.UNION) {
-                return isAssignable(source, target);
+                return isAssignable(sourceElementType, target, unresolvedTypes);
             }
 
             // Then lhs type should 'any' type
-            return target.tag == TypeTags.ANY;
+            return (target.tag == TypeTags.ANY) && (sourceElementType.tag != TypeTags.ERROR);
 
         } else if (target.tag == TypeTags.ARRAY) {
             // Only the left-hand side is an array type
@@ -873,7 +874,7 @@ public class Types {
         }
 
         // In this case, lhs type should be of type 'any' and the rhs type cannot be a value type
-        return target.tag == TypeTags.ANY && !isValueType(source);
+        return (target.tag == TypeTags.ANY) && !isValueType(source) && (source.tag != TypeTags.ERROR);
     }
 
     private boolean isFunctionTypeAssignable(BInvokableType source, BInvokableType target,
@@ -994,13 +995,12 @@ public class Types {
 
         BArrayType lhsArrayType = (BArrayType) target;
         BArrayType rhsArrayType = (BArrayType) source;
+        boolean hasSameTypeElements = isSameType(lhsArrayType.eType, rhsArrayType.eType, unresolvedTypes);
         if (lhsArrayType.state == BArrayState.UNSEALED) {
-            return rhsArrayType.state == BArrayState.UNSEALED &&
-                    isSameType(lhsArrayType.eType, rhsArrayType.eType, unresolvedTypes);
+            return (rhsArrayType.state == BArrayState.UNSEALED) && hasSameTypeElements;
         }
 
-        return checkSealedArraySizeEquality(rhsArrayType, lhsArrayType)
-                && isSameType(lhsArrayType.eType, rhsArrayType.eType, unresolvedTypes);
+        return checkSealedArraySizeEquality(rhsArrayType, lhsArrayType) && hasSameTypeElements;
     }
 
     public boolean checkSealedArraySizeEquality(BArrayType rhsArrayType, BArrayType lhsArrayType) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -837,9 +837,11 @@ public class Types {
                 return isAssignable(sourceElementType, targetElementType, unresolvedTypes);
             }
 
-            if (targetArrayType.size == source.size) {
-                return isAssignable(sourceElementType, targetElementType, unresolvedTypes);
+            if (targetArrayType.size != source.size) {
+                return false;
             }
+            
+            return isAssignable(sourceElementType, targetElementType, unresolvedTypes);
         } else if (target.tag == TypeTags.JSON) {
             return isAssignable(sourceElementType, target, unresolvedTypes);
         }
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -844,14 +844,17 @@ public class Types {
             if (targetArrayType.size == source.size) {
                 return isAssignable(sourceElementType, targetElementType, unresolvedTypes);
             }
-            return false;
-        }
-
-        if ((target.tag == TypeTags.UNION) || (target.tag == TypeTags.JSON)) {
+        } else if (target.tag == TypeTags.JSON) {
             return isAssignable(sourceElementType, target, unresolvedTypes);
+        } else if (target.tag == TypeTags.UNION) {
+            return isAssignable(sourceElementType, target, unresolvedTypes);
+//            for (BType memberType : ((BUnionType) target).getMemberTypes()) {
+//                if (isArrayTypesAssignable(source, memberType, unresolvedTypes)) {
+//                    return true;
+//                }
+//            }
         }
-
-        return (target.tag == TypeTags.ANY) && (sourceElementType.tag != TypeTags.ERROR);
+        return false;
     }
 
     private boolean isFunctionTypeAssignable(BInvokableType source, BInvokableType target,
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -837,10 +837,10 @@ public class Types {
             BArrayType lhsArrayType = (BArrayType) target;
             BArrayType rhsArrayType = (BArrayType) source;
             if (lhsArrayType.state == BArrayState.UNSEALED) {
-                return isArrayTypesAssignable(rhsArrayType.eType, lhsArrayType.eType, unresolvedTypes);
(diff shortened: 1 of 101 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -784,18 +784,19 @@ public boolean isArrayTypesAssignable(BType source, BType target, Set<TypePair>
         } else if (source.tag == TypeTags.ARRAY) {
```
MaryamZi, line 784: I think we can refactor this method a bit.

In L779 and L782 IMO we can call `isAssignable` instead of `isArrayTypesAssignable`.

AFAICT we can then change the first parameter of this method to a `BArrayType` (we'll probably have to deprecate this and introduce a new method). This'll reduce the number of specific checks and reuse the `isAssignable` logic, which has been tested quite a lot, more.

KRVPerera: Sorry, I think I don't get this clearly?

The `isArrayTypesAssignable` is getting called by `isAssignable` logic anyway [link](https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java#L618).
Is there a point going for all the checks in `isAssignable` when we know eventually it will call `isArrayTypesAssignable` function anyway?

And changing the first parameter to BArrayType makes this specific right?

MaryamZi: > Is there a point going for all the checks in `isAssignable` when we know eventually it will call `isArrayTypesAssignable` function anyway?
> 

L779 and L782 check assignability for the element types of the first level of arrays. And the element types may or may not be array types, so we can't say for sure it'll be calling `isArrayTypesAssignable`.

Consider 

```ballerina
string[] s = [];
(string|int)[] t = s;
```

For `(string|int)[] t = s;`, `isArrayTypesAssignable` gets invoked first, which then at L779 calls `isArrayTypesAssignable` again with source `string` and target `(string|int)` which eventually ends up calling `isAssignable` like in L807. Instead, we can directly call  `isAssignable` to check assignability of element types.



> And changing the first parameter to BArrayType makes this specific right?

IMO, the name `isArrayTypesAssignable` itself has already made it specific. :) This IS a specific check right?

Basically we call this method only when the source type is an array, to check if it is assignable to a specific type. 

So rhsType is always an array.

- if lhsType is also an array - return `isAssignable(rhsElemType, lhsElemType)`
- if lhsType is a union - return true if for at least one member `isArrayTypesAssignable` is true
- if lhsType is `json` - return `isAssignable(rhsElemType, lhsType)` 
- if lhsType is `anydata` - return `rhsType.isAnydata()` 
- if lhsType is `any` - return `rhsElemType.tag != TypeTags.ERROR`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -775,49 +776,26 @@ private boolean isArrayTypeAssignableToTupleType(BArrayType source, BTupleType t
...
         }
 
-        // Now both types are not array types and they have to be assignable
-        if (isAssignable(source, target, unresolvedTypes)) {
-            return true;
-        }
-
-        if (target.tag == TypeTags.UNION) {
-            return isAssignable(source, target, unresolvedTypes);
+        if ((target.tag == TypeTags.UNION) || (target.tag == TypeTags.JSON)) {
```
MaryamZi, line 794: For unions, consider something like,

```ballerina
   int[] x = [1, 2];
   int|string[] y = x;
```

The current check will cause this assignment to pass right?

I think for unions we have to check `isArrayTypesAssignable` individually for each member and return true if at least one returns true.

```java
        if (target.tag == TypeTags.UNION) {
            for (BType memberType : ((BUnionType) target).getMemberTypes()) {
                if (isArrayTypesAssignable(source, memberType, unresolvedTypes) {
                    return true;
                }
            }
            return false;
        }
```

MaryamZi: Please add a couple of tests too?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -775,49 +776,26 @@ private boolean isArrayTypeAssignableToTupleType(BArrayType source, BTupleType t
...
+            return isAssignable(sourceElementType, target, unresolvedTypes);
         }
 
-        // In this case, lhs type should be of type 'any' and the rhs type cannot be a value type
-        return target.tag == TypeTags.ANY && !isValueType(source);
+        return (target.tag == TypeTags.ANY) && (sourceElementType.tag != TypeTags.ERROR);
```
MaryamZi, line 798:
```ballerina
error[] x = [e1, e2];
any y = x;
```

This is valid. So we don't need this check.

stdlib/http/src/main/ballerina/src/http/resiliency/failover_client_endpoint.bal
```diff
@@ -308,7 +308,7 @@ function performFailoverAction (string path, Request request, HttpOperation requ
     Response inResponse = new;
     HttpFuture inFuture = new;
     Request failoverRequest = request;
-    error?[] failoverActionErrData = [];
```
MaryamZi, line 311: @ThisaruGuruge can you verify these changes please? The previous behaviour (passing an `error?[]` where a `ClientError?[]` is expected in `populateFailoverErrorHttpStatusCodes`) was allowed due to a bug.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
MaryamZi: The last check is incorrect, sorry, my bad.

See https://github.com/ballerina-platform/ballerina-lang/pull/22494#discussion_r412675176

stdlib/http/src/main/ballerina/src/http/resiliency/failover_client_endpoint.bal
ThisaruGuruge: This should be fine.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: Looks like unions are handled here.
https://github.com/ballerina-platform/ballerina-lang/pull/22494/files#diff-782094a79e6831ac57a34ad157af49c1L562

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -775,49 +772,28 @@ private boolean isArrayTypeAssignableToTupleType(BArrayType source, BTupleType t
...
-        } else if (source.tag == TypeTags.ARRAY) {
-            // Only the right-hand side is an array type
 
-            // If the target type is a JSON, then element type of the rhs array
-            // should only be a JSON supported type.
-            if (target.tag == TypeTags.JSON) {
-                return isAssignable(((BArrayType) source).getElementType(), target, unresolvedTypes);
+            if (targetArrayType.size == source.size) {
+                return isAssignable(sourceElementType, targetElementType, unresolvedTypes);
             }
```
MaryamZi, line 786: Shall we do
```suggestion
            if (targetArrayType.size != source.size) {
                return false;
            }
            
            return isAssignable(sourceElementType, targetElementType, unresolvedTypes);
```

instead?

Improves readabilty IMO.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/assign/AssignStmtTest.java
```diff
@@ -207,6 +207,20 @@ public void testAssignmentNegativeCases() {
...
+                                  "incompatible types: expected 'error[]', found 'any[]'", 100, 26);
+        BAssertUtil.validateError(resultNegative, i++,
+                                  "incompatible types: expected '(CError|LError)?[]', found 'error?[]'", 118, 19);
+        BAssertUtil.validateError(resultNegative, i++,
+                                  "incompatible types: expected '(CError|LError)?[]', found 'error?[]'", 119, 11);
+        BAssertUtil.validateError(resultNegative, i++,
+                                  "incompatible types: expected '(error|int[])', found 'error[]'", 127, 21);
+        BAssertUtil.validateError(resultNegative, i++,
+                                  "incompatible types: expected '(int|error[])', found 'error'", 132, 21);
+//        BAssertUtil.validateError(resultNegative, i++,
```
MaryamZi, line 222: This can be removed?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -564,12 +565,8 @@ private boolean isAssignable(BType source, BType target, Set<TypePair> unresolve
         }
 
         if (targetTag == TypeTags.JSON) {
-            if (sourceTag == TypeTags.JSON) {
```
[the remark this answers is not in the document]
KRVPerera: Already checked here right?
https://github.com/ballerina-platform/ballerina-lang/pull/22494/files/9ba99d6bf7cf3860b98b291a7e4a2d9f2827138f#diff-782094a79e6831ac57a34ad157af49c1L473

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/assign/AssignStmtTest.java
KRVPerera: yes

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: I think this is subjective. Also, I will be adding more false statements to the method for this change. Think code is obvious anyway. I prefer to keep the code as it is if there is no objection.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/assign/AssignStmtTest.java
KRVPerera: yes, duplicate

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
MaryamZi: As a practice, we usually try to return early as and when possible, and minimize else/else-if blocks. 

With this change we can update the method itself to something like
```java
    private boolean isArrayTypesAssignable(BArrayType source, BType target, Set<TypePair> unresolvedTypes) {
        BType sourceElementType = source.getElementType();
        if (target.tag == TypeTags.ARRAY) {
            BArrayType targetArrayType = (BArrayType) target;
            BType targetElementType = targetArrayType.getElementType();
            if (targetArrayType.state == BArrayState.UNSEALED) {
                return isAssignable(sourceElementType, targetElementType, unresolvedTypes);
            }

            if (targetArrayType.size != source.size) {
                return false;
            }
            
            return isAssignable(sourceElementType, targetElementType, unresolvedTypes);
        }
        
        if (target.tag == TypeTags.JSON) {
            return isAssignable(sourceElementType, target, unresolvedTypes);
        }
        
        return false;
    }
```

And it's just one additional `return false` anyway right?

MaryamZi: [This blog](https://medium.com/better-programming/why-you-need-to-stop-using-else-statements-5b1fd09dea9e) was shared sometime back.

tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
```diff
@@ -86,3 +86,41 @@ public function restActionResultAssignment() returns [int, int, string, string,
...
+    error testError = error("Test Error");
+    error[] errorArray = [testError];
+    any anyArray = errorArray;
+    error[] errorArrayBack = <error[]>anyArray;
+    assertEquality(errorMessage, errorArrayBack[0].reason());
+}
+
+function testAssignIntArrayToJson() {
+    int[*] intArray = [1, 2];
+    json jsonObjectOfIntArray = intArray;
```
MaryamZi, line 101: This variable name does not make sense IMO.

Shall we rename it to `jsonVar` or `jsonArray`?

tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
```diff
@@ -86,3 +86,41 @@ public function restActionResultAssignment() returns [int, int, string, string,
...
+    json jsonToJson = jsonObjectOfIntArray;
```
MaryamZi, line 102: Why do we need another assignment to a variable of the same type? What are we trying to test here?

tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
```diff
@@ -86,3 +86,41 @@ public function restActionResultAssignment() returns [int, int, string, string,
...
+    assertEquality("1 2", jsonToJson.toString());
```
MaryamZi, line 103: Shall we assert the actual values instead of using `.toString()`?

How about
```
assertTrue(jsonVar is int[2]);

int[2] arr = <int[2]> jsonVar;
assertEquality(1, arr[0]);
assertEquality(2, arr[1]);
```

tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
```diff
@@ -86,3 +86,41 @@ public function restActionResultAssignment() returns [int, int, string, string,
...
+}
+
+function testAssignIntOrStringArrayIntOrFloatOrStringUnionArray() {
+    int[]|string[] intOrStringArray = <int[]>[1, 2];
+    (int|float)[]|string[] intOrFloatArray = intOrStringArray;
```
MaryamZi, line 108: Again variable name doesn't reflect the actual type does it.

Given that these are small tests I'd rather go with simple names, `a` and `b`, `arr1` and `arr2`, etc.

tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
KRVPerera: extra line removed

tests/jballerina-bstring-unit-test/src/test/java/org/ballerinalang/test/error/ErrorTest.java
```diff
@@ -270,10 +270,11 @@ public void testErrorNegative() {
                 "error reason is mandatory for direct error constructor", 112, 28);
         BAssertUtil.validateError(negativeCompileResult, i++,
                 "incompatible types: expected 'error', found '(error|int)'", 118, 11);
-        BAssertUtil.validateError(negativeCompileResult, i,
+        BAssertUtil.validateError(negativeCompileResult, i++,
```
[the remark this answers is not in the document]
KRVPerera: removed

KRVPerera:
Fixed code review suggestions

PR #22507, opened 2020-04-08
Syntax tree modifer functionality

### LINKED ISSUE #22441: Implement syntax tree rewriter
This is required for the formatter

sameerajayasoma:
Fixes #22441 

## Samples

```java
    /**
     * An implementation of {@code SyntaxTreeModifier}.
     */
    private static class VariableDeclModifier extends SyntaxTreeModifier {

        @Override
        public Node transform(VariableDeclaration varDeclStmt) {
            String oldVarName = varDeclStmt.variableName().getText();
            Identifier newVarName = NodeFactory.createIdentifier(oldVarName + "new");
            return NodeFactory.createVariableDeclaration(varDeclStmt.finalKeyword(), varDeclStmt.typeName(),
                    newVarName, varDeclStmt.equalsToken(), varDeclStmt.initializer(), varDeclStmt.semicolonToken());
        }
    }
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/syntax/tree/NodeFactory.java
```diff
@@ -0,0 +1,120 @@
...
+                                                                      Node returnTypeDesc,
+                                                                      BlockStatement functionBody) {
+        Objects.requireNonNull(visibilityQualifier, "visibilityQualifier must not be null");
+        Objects.requireNonNull(functionKeyword, "functionKeyword must not be null");
+        Objects.requireNonNull(functionName, "functionName must not be null");
+        Objects.requireNonNull(openParenToken, "openParenToken must not be null");
+        Objects.requireNonNull(parameters, "parameters must not be null");
+        Objects.requireNonNull(closeParenToken, "closeParenToken must not be null");
+        Objects.requireNonNull(returnTypeDesc, "returnTypeDesc must not be null");
+        Objects.requireNonNull(functionBody, "functionBody must not be null");
```
SupunS, line 64: Can we delegate this to a common method? something like: `validateNodes(Node... nodes)` that loops through and validates

sameerajayasoma: Can do that Supun? I’ve done that in each “modify” methods in nodes. 

But, if we do that, we won’t be able to provide precise error messages. That is why I wanted to do a null check for each parameter.

sameerajayasoma: On a different thought, I think we can do what you suggested if we store field names such as "visibilityQualifier", "functionKeyword", "functionName" etc. As we discussed during the call, Syntax Tree to diagram generator needs field names.

PR #22511, opened 2020-04-08
Fix direct return of byte binary operation in a lambda

### LINKED ISSUE #22168: JVM Bytecode Generation Failure with "Bad type on operand stack"
In Ballerina v1.2.0, the attached program crashes at application startup. I'm attaching here the Ballerina source code, and also the "ballerina-internal.log" that is generated. The error starts with "Bad type on operand stack" in the log. 
[code-and-logs.zip](https://github.com/ballerina-platform/ballerina-lang/files/4379059/code-and-logs.zip)

Kishanthan:
## Purpose
This PR fixes the issue of direct return of byte binary operation in a lambda.

An example :
```ballerina
public function main() {
    var fn = function () returns int {
        byte a = 255;
        return a - a;
    };
}
```

Fixes #22168

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/bytetype/BByteValueTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/bytetype/BByteValueTest.java
@@ -792,4 +792,34 @@ public class BByteValueTest {
         BInteger bInteger = (BInteger) returns[0];
         Assert.assertEquals(bInteger.intValue(), expected, "Invalid byte value returned.");
     }
+
+    @Test(description = "Test byte value return as int in lambda 1")
+    public void testByteReturnAsIntInLambda1() {
+        BValue[] returns = BRunUtil.invoke(result, "testByteReturnAsIntInLambda1");
+        Assert.assertEquals(returns.length, 1);
+        Assert.assertSame(returns[0].getClass(), BInteger.class);
+        BInteger intValue = (BInteger) returns[0];
+        Assert.assertEquals(intValue.intValue(), 0, "Invalid int value returned.");
+    }
+
+    @Test(description = "Test byte value return as int in lambda 2")
+    public void testByteReturnAsIntInLambda2() {
+        BValue[] returns = BRunUtil.invoke(result, "testByteReturnAsIntInLambda2");
+        Assert.assertEquals(returns.length, 1);
+        Assert.assertSame(returns[0].getClass(), BBoolean.class);
+        BBoolean booleanVal = (BBoolean) returns[0];
+        Assert.assertTrue(booleanVal.booleanValue());
+    }
+
+    @Test(description = "Test byte value return as int in lambda 3")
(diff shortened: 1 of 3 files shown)


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/bytetype/BByteValueTest.java
```diff
@@ -792,4 +792,34 @@ private void invokeBitwisePrecedenceTestFunctionForInt(int a, int b, int c, Stri
         BInteger bInteger = (BInteger) returns[0];
         Assert.assertEquals(bInteger.intValue(), expected, "Invalid byte value returned.");
     }
+
+    @Test(description = "Test byte value return as int in lambda 1")
+    public void testByteReturnAsIntInLambda1() {
+        BValue[] returns = BRunUtil.invoke(result, "testByteReturnAsIntInLambda1");
```
MaryamZi, line 798: With new lang tests we've tried writing the tests in Ballerina itself as much as possible.

For example https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/jballerina-unit-test/src/test/resources/test-src/expressions/listconstructor/list_constructor_infer_type.bal#L55

We can rewrite these tests similarly right?

Kishanthan: looks like you have to repeat the assertEquality method for each new test ?

Kishanthan: And yes, we can write that in the new style. The thing is all the other methods are written in the old way. May be we have to add a new bal test file with these new tests.

MaryamZi: For the time-being yes, the `assertEquality` function has to be repeated in each test file. But once the compiler test suite is done we can directly switch to calling a similar (most-likely the same) function in that particular module.

MaryamZi: Yeah, there is some inconsistency when adding tests to an existing file. But even then, IMO it's kinda OK given that we usually add new tests at the end?

I'm OK with either approach.

Kishanthan: Updated the tests :)

MaryamZi: Great, thanks!

Issue #22430, opened 2020-04-02
Add bare readonly type for inherently immutable basic types


MaryamZi:
**Description:**
See https://github.com/ballerina-platform/ballerina-spec/issues/462

MaryamZi:
Initial implementation was introduced with https://github.com/ballerina-platform/ballerina-lang/pull/22519.

The following issues track the rest of the tasks to be completed:
- Add support for `readonly & T` #22515
- Support readonly record fields #22600
- Allow using the type-test for `readonly` with selectively immutable basic types #22785

PR #22519, opened 2020-04-08
Add readonly type - Phase I

MaryamZi:
## Purpose
$title. 

Initial changes to fix https://github.com/ballerina-platform/ballerina-lang/issues/22430
Adds support for readonly record fields in the typedesc, partial fix for https://github.com/ballerina-platform/ballerina-lang/issues/22600

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/22601

Working on improvements, adding tests, etc. Sending this PR since several tasks are blocked on this.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1054,22 +1054,107 @@ private boolean validateSpecifiedFields(BLangRecordLiteral mappingConstructor, B
...
-            // Check if `field` is explicitly assigned a value in the record literal
-            // If a required field is missing, it's a compile error
-            if (!specFieldNames.contains(field.name.value) && Symbols.isFlagOn(field.symbol.flags, Flags.REQUIRED)) {
+            String fieldName = field.name.value;
+            BType fieldType = symTable.semanticError;
+            DiagnosticPos fieldPos;
+
+            if (specFieldNames.contains(fieldName)) {
+                RecordLiteralNode.RecordField recordField = getRecordFieldByName(specifiedFields, fieldName);
+                if (recordField == null) {
```
pubudu91, line 1071: Can this be `null` here, if the mapping constructor contains the field?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1054,22 +1054,107 @@ private boolean validateSpecifiedFields(BLangRecordLiteral mappingConstructor, B
...
+                        BLangRecordLiteral.BLangRecordSpreadOperatorField spreadOpField =
+                                (BLangRecordLiteral.BLangRecordSpreadOperatorField) recordField;
+                        BRecordType recordType = (BRecordType) spreadOpField.expr.type;
+
+                        for (BField bField : recordType.fields) {
+                            if (fieldName.equals(bField.name.value)) {
+                                fieldType = bField.type;
+                                break;
+                            }
+                        }
```
pubudu91, line 1097: Isn't this part duplicate of the logic in `getRecordFieldByName()`? Do we need to do the same here, after calling that method? Should be able to directly take the type of the spread field right?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -347,6 +347,9 @@ error.invalid.record.rest.descriptor=\
 error.missing.required.record.field=\
   missing non-defaultable required record field ''{0}''
 
+error.invalid.field.for.readonly.record.field=\
+  invalid field; ''readonly'' field expected for ''{0}''
```
pubudu91, line 351: Shouldn't this be a colon?

MaryamZi: Fixed, but we seem to have used a mix of `:` and `;` in error messages.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Yeah not in this context, I've removed the null check.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Yeah, noticed the duplication, but the issue was that we basically needed the type and pos, and for key-value and variable name fields we can retrieve a `RecordField` and get the type and pos from it. But for a spread-field the type is from a `BField` and the pos is still from the spread field. 

I've refactored this a bit with an inner class.

tests/jballerina-unit-test/src/test/resources/test-src/record/readonly_record_fields.bal
```diff
@@ -0,0 +1,159 @@
...
+};
+
+function testValidUpdateOfPossiblyReadonlyFieldInUnion() {
+    Details d = {
+        name: "Jo",
+        id: 1234
+    };
+
+    Student|Details sd = d;
+    sd.name = "May";
```
hasithaa, line 94: Need to clarify this behavior. `Student.name` is a readonly field, So `sd.name` assignment can panic at runtime. 
We have similar behavior on the inherent type violation. So this is OK IMO.

MaryamZi: Created https://github.com/ballerina-platform/ballerina-spec/issues/495 to clarify.

PR #22686, opened 2020-04-16
 Add nil literal support

### LINKED ISSUE #22359: Add nil literal to the new parser
**Description:**
Refer: https://ballerina.io/spec/lang/2020R1/#nil-literal

lochana-chathura:
Fixes #22359 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -5403,4 +5414,28 @@ private STNode parseLocalTypeDefinitionStatement(STNode annots) {
...
+            case OPEN_PAREN_TOKEN:
+                startContext(ParserRuleContext.NIL_LITERAL);
+                STNode openParenthesisToken = parseOpenParenthesis();
+                STNode closeParenthesisToken = parseCloseParenthesis();
+                endContext();
+                return STNodeFactory.createNilLiteral(openParenthesisToken, closeParenthesisToken);
+            default:
+                Solution sol = recover(token, ParserRuleContext.NIL_LITERAL);
+                return sol.recoveredNode;
+        }
```
lochana-chathura, line 5439: Is this approach ok?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/syntax/tree/SyntaxKind.java
```diff
@@ -184,6 +185,7 @@
     SERVICE_BODY(3023),
     ANNOTATION(3024),
     METADATA(3025),
+    NIL_LITERAL(3026),
```
lochana-chathura, line 188: Couldn't place this under section where other literals are declared as `parserTestutils.java` has the following method.
```public static boolean isTerminalNode(SyntaxKind syntaxKind) {
        return SyntaxKind.IMPORT_DECLARATION.compareTo(syntaxKind) > 0 || syntaxKind == SyntaxKind.EOF_TOKEN;
    }```

File: compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java [hunk 3]
```diff
@@ -814,7 +817,8 @@ private Result seekMatch(ParserRuleContext currentCtx, int lookahead, int curren
                 case ACCESS_EXPRESSION:
                     return seekInAccessExpression(currentCtx, lookahead, currentDepth, matchingRulesCount);
                 case BASIC_LITERAL:
-                    hasMatch = isBasicLiteral(nextToken.kind);
+                    STToken nextNextToken = this.tokenReader.peek(lookahead + 1);
+                    hasMatch = isBasicLiteral(nextToken.kind, nextNextToken.kind);
```
SupunS, line 821: Don't need to handle `()` in basic literals.. It was added to deal with literals with a single token. So may be we could only handle `null` here.

For `()` lets use `NIL_LITERAL` and add it to the `EXPRESSIONS` alternatives array.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -1358,10 +1367,14 @@ private ParserRuleContext getNextRule(ParserRuleContext currentCtx, int nextLook
...
-                endContext();
-                // After parsing nil type descriptor all the other parsing is same as next rule of simple type
-                return getNextRuleForTypeDescriptor();
-            }
+                    endContext();
+                    // After parsing nil type descriptor all the other parsing is same as next rule of simple type
+                    return getNextRuleForTypeDescriptor();
+                }
+                if (parentCtx == ParserRuleContext.NIL_LITERAL) {
+                    endContext();
```
SupunS, line 1375: Eventhough the `endContext()` is called here, `startContext()` is not called at the first switch-case of `getNextRule()` method.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -1644,6 +1657,9 @@ private ParserRuleContext getNextRule(ParserRuleContext currentCtx, int nextLook
                 return ParserRuleContext.EXPRESSION_RHS;
             case LOCAL_TYPE_DEFINITION_STMT:
                 return ParserRuleContext.TYPE_KEYWORD;
+            case NULL_KEYWORD:
+                endContext(); // end nil-literal
```
SupunS, line 1661: A context is not started for null keyword.. So ending the context here is not correct. 

We might not need any context for null keyword at all. It is very similar to string literal, int literal etc. A context is needed when a token can occur in different contexts/places  and if we can't determine whats next rule should be.. Then we make use of the parent context.

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -1425,6 +1425,21 @@
           "type": "Node"
         }
       ]
+    },
+    {
+      "name": "NilLiteral",
```
sameerajayasoma, line 1430: Use NilLiteralNode as per the new convention

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -1425,6 +1425,21 @@
...
+      "base": "Node",
+      "kind": "NIL_LITERAL",
+      "attributes": [
+        {
+          "name": "firstToken",
```
sameerajayasoma, line 1435: Should this be openParenToken?

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -1425,6 +1425,21 @@
...
+          "type": "Token"
+        },
+        {
+          "name": "secondToken",
```
sameerajayasoma, line 1439: Should this be closeParanToken

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
lochana-chathura: @SupunS No switch case means, `( )` is considered to be `NIL_LITERAL` kind and `NULL` is considered to be `NULL_KEYWORD` kind, and will have separate `pareseNilLiteral( )`, `pareseNullKeyword( )` methods ?

SupunS: yes. For the null keyword, we can do the same way as, string literal

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
lochana-chathura: Adding `NIL_LITERAL` to `EXPRESSIONS` array cause 
```public function foo hello (
}``` to be generated as ```public function foo  MISSING[(]hello  MISSING[] MISSING[=](
 MISSING[] MISSING[)] MISSING[)] MISSING[{]}```
I will have a look on what's happening there.

Comment (lochana-chathura) [reply compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java]: Adding `NIL_LITERAL` to `EXPRESSIONS` array cause func definition with extra func name;  `public function foo hello (
}` to be generated as

`public function foo  MISSING[(]hello  MISSING[] MISSING[=](
 MISSING[] MISSING[)] MISSING[)] MISSING[{]}`

I will have a look on what's happening there

File: compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/expressions/NilLiteralTest.java [hunk 9]
```diff
@@ -0,0 +1,39 @@
...
+    public void testSimpleNilLiteral() {
+        test("()", "basic-literals/nil_literal_assert_01.json");
+        test("null", "basic-literals/nil_literal_assert_02.json");
+    }
+
+    @Test
+    public void testNilLiteral() {
+        test("() != null + a", "basic-literals/nil_literal_assert_03.json");
+    }
+}
```
SupunS, line 39: Shall we add some more tests, to validate the usages of nil-literal (`()`), such as var-decl, defaultable-params, in object/record fields, etc?

lochana-chathura:
Plz hold the PR. I'll add IsExpressionNode name change to this.

lochana-chathura:
Change added!

PR #22734, opened 2020-04-17
Refactor record, tuple and error var desugar logic

pubudu91:

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
@@ -1185,25 +1185,25 @@ public class Desugar extends BLangNodeVisitor {
     @Override
     public void visit(BLangTupleVariable varNode) {
         //  case 1:
-        //  (string, int) (a, b) = (tuple)
+        //  [string, int] (a, b) = (tuple)
         //
         //  any[] x = (tuple);
         //  string a = x[0];
         //  int b = x[1];
         //
         //  case 2:
-        //  ((string, float) int)) ((a, b), c)) = (tuple)
+        //  [[string, float], int] [[a, b], c] = (tuple)
         //
         //  any[] x = (tuple);
         //  string a = x[0][0];
         //  float b = x[0][1];
         //  int c = x[1];
 
-        //create tuple destruct block stmt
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -1184,17 +1184,76 @@ public void visit(BLangLetExpression letExpression) {
...
+        //  any[] x = (tuple);
+        //  string a = x[0][0];
+        //  float b = x[0][1];
+        //  int c = x[1];
+
+        //create tuple destruct block stmt
+        final BLangBlockStmt blockStmt = ASTBuilderUtil.createBlockStmt(varNode.pos);
+
+        //create a simple var for the array 'any[] x = (tuple)' based on the dimension for x
+        String name = "tuple";
```
MaryamZi, line 1206: Not introduced by this PR, but shouldn't we be using different names, starting with `$`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -1184,17 +1184,76 @@ public void visit(BLangLetExpression letExpression) {
...
+        //
+        //  case 2:
+        //  ((string, float) int)) ((a, b), c)) = (tuple)
+        //
+        //  any[] x = (tuple);
+        //  string a = x[0][0];
+        //  float b = x[0][1];
+        //  int c = x[1];
+
+        //create tuple destruct block stmt
```
MaryamZi, line 1202: Shall we refactor in all the moved code? 

```suggestion
        // Create tuple destruct block stmt
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
pubudu91: you mean unique names? yeah, have noticed the use of same name in some places in desugar. will change and see if it breaks anything

pubudu91: Changed to `$tuple$`

Issue #22639, opened 2020-04-15
Tables don't give a compile-time error if two rows have the same key


gimantha:
The table should give a compile time error if the table constructor have two rows with the same key. Currently it replaces the older row with the new row

```ballerina
type Person record {
    map<string> name;
    int age;
};

type GlobalTable table<Person> key(name);

GlobalTable tab = table [
  { name: {"AAA":"DDDD"}, age: 31 },
  { name: {"AAA":"DDDD"}, age: 34 }
];

function testGlobalTableConstructExpr() returns string {
    return tab.toString();
}
```

jclark:
The spec says this is a compile-time error

https://github.com/ballerina-platform/ballerina-spec/commit/f46738eb86c075107881338317e449d699848b5d

This was decided in https://github.com/ballerina-platform/ballerina-spec/issues/477#issuecomment-610125045

mohanvive:
Fixed in https://github.com/ballerina-platform/ballerina-lang/pull/22878 ....

PR #22739, opened 2020-04-17
Fix Key hashing issue in Table implementation

gimantha:
## Purpose

Fix par of #22639


mohanvive:
@gimantha It seems Travis is not running for this PR. Better to check the build locally with tests.

PR #22776, opened 2020-04-20
Populate caller remote IP and introduce function to get remote hostname[master]

### LINKED ISSUE #22283: Http : Caller's remote-address.host property has never been set
**Description:**
http:Caller's remoteAddress.host property is not set at all. 

Previously, it was doing a reverse dns lookup and that caused a significant delay (~15s) when the call is from a remote network. But IMO, Ip can be added instead of the hostname. 
https://github.com/ballerina-platform/ballerina-lang/issues/20085

**Steps to reproduce:**
Make a call from remote network and observe the property.

**Affected Versions:**
1.1.4

MacOS Mojave

chamil321:
Fixes #22283

### DIFF
--- a/stdlib/http/src/main/ballerina/src/http/http_connection.bal
+++ b/stdlib/http/src/main/ballerina/src/http/http_connection.bal
@@ -172,11 +172,24 @@ public type Caller client object {
         response.statusCode = STATUS_ACCEPTED;
         return self->respond(response);
     }
+
+    # Gets the hostname from the remote address. This method may trigger a DNS reverse lookup if the address was created
+    # with a literal IP address.
+    #
+    # + return - The hostname of the address or `nil` if it is unresolved
+    public function getRemoteHostName() returns string? {
+        return java:toString(nativeGetRemoteHostName(self));
+    }
 };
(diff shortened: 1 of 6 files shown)


stdlib/http/src/main/ballerina/src/http/http_connection.bal
```diff
@@ -172,13 +172,26 @@ public type Caller client object {
         response.statusCode = STATUS_ACCEPTED;
         return self->respond(response);
     }
+
+    # Gets the hostname from the remote address. This method may trigger a DNS reverse lookup if the address was created
+    # with a literal IP address.
+    #
+    # + return - The hostname of the address or `nil` if it is unresolved
```
daneshk, line 179:
```suggestion
    # + return - The hostname of the address or else `()` if it is unresolved
```

chamil321: It is fixed in https://github.com/ballerina-platform/ballerina-lang/pull/22779/files#diff-c90efb208c957133fd64332cd26e00f1R241

PR #22816, opened 2020-04-21
Add type cast action and separate function/method calls and actions

### LINKED ISSUE #17993: [Spec Deviation] Actions nested inside other actions
**Description:**
The 2019r3 spec says

"Actions are an intermediate syntactic category between expressions and statements. Actions are similar to expressions, in that they yield a value. However, an action cannot be nested inside an expression; **it can only occur as part of a statement or nested inside other actions**. This is because actions are shown in the sequence diagram in the graphical syntax."

**Steps to reproduce:**
The following is not allowed though.
```ballerina
type Foo client object {
    
    remote function abc() returns int {
        return 1;
    }

    remote function def(int i) returns string {
        return "str";
    }
};

public function main() {
    Foo f = new;
    string s = f->def(f->abc());
}
```

```cmd
$ ballerina run test.bal 
error: .::test.bal:14:23: action invocation as an expression not allowed here
```

**Affected Versions:**
1.0.0-beta-SNAPSHOT - 15th August build

pubudu91:
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/model/tree/ActionNode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/model/tree/ActionNode.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *  Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
  *
  *  WSO2 Inc. licenses this file to you under the Apache License,
  *  Version 2.0 (the "License"); you may not use this file except
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/model/tree/ActionNode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/model/tree/ActionNode.java
@@ -1,0 +1,26 @@
+/*
+ *  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  WSO2 Inc. licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.ballerinalang.model.tree;
+
+/**
+ * Interface used for marking nodes as actions.
+ *
+ * @since 1.3.0
+ */
+public interface ActionNode extends Node {
+}
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ConstantPropagation.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ConstantPropagation.java
@@ -275,6 +275,14 @@ public class ConstantPropagation extends BLangNodeVisitor {
     }
 
     @Override
+    public void visit(BLangInvocation.BLangActionInvocation remoteMethodCallNode) {
+        remoteMethodCallNode.expr = rewrite(remoteMethodCallNode.expr);
+        rewrite(remoteMethodCallNode.requiredArgs);
+        rewrite(remoteMethodCallNode.restArgs);
+        result = remoteMethodCallNode;
+    }
+
+    @Override
     public void visit(BLangNamedArgsExpression namedArgsExpression) {
         namedArgsExpression.expr = rewrite(namedArgsExpression.expr);
         result = namedArgsExpression;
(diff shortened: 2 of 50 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1816,6 +1811,63 @@ public void visit(BLangInvocation iExpr) {
...
+            return;
+        }
+
+        // Find the variable reference expression type
+        BType exprType = checkExpr(aInv.expr, this.env, symTable.noType);
+        BType actualType = symTable.semanticError;
+        BLangVariableReference varRef = (BLangVariableReference) aInv.expr;
+
+        if (exprType == symTable.semanticError || exprType.tag != TypeTags.OBJECT ||
+                ((varRef.symbol.tag & SymTag.ENDPOINT) != SymTag.ENDPOINT) && !aInv.async) {
```
dulvinw, line 1833: WDYT of using a method instead of all these. Looks a bit too verbose for this method for me :)

compiler/ballerina-lang/src/main/java/org/ballerinalang/model/tree/ActionNode.java
```diff
@@ -0,0 +1,26 @@
+/*
+ *  Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
```
irshadnilam, line 3: Wrong year?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
pubudu91: Just the `(varRef.symbol.tag & SymTag.ENDPOINT) != SymTag.ENDPOINT)` part or the whole thing? The expression as a whole, checks whether it's a remote method call or an async method call.

pubudu91: On second thoughts, that's probably not the correct way to check for a remote method call

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ConstantPropagation.java
```diff
@@ -274,6 +274,14 @@ public void visit(BLangInvocation bLangInvocationNode) {
         result = bLangInvocationNode;
     }
 
+    @Override
+    public void visit(BLangInvocation.BLangActionInvocation remoteMethodCallNode) {
```
MaryamZi, line 278: Shall we use 
```suggestion
    public void visit(BLangInvocation.BLangActionInvocation actionInvocation) {
```
instead?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -3615,6 +3614,58 @@ public void visit(BLangInvocation iExpr) {
         fixTypeCastInTypeParamInvocation(iExpr, genIExpr);
     }
 
+    @Override
+    public void visit(BLangInvocation.BLangActionInvocation actionInvocation) {
```
MaryamZi, line 3618: Aren't we duplicating most of the logic from `visit(BLangInvocation iExpr)` here? Can't we use a common method instead?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangPackageBuilder.java
```diff
@@ -1654,8 +1655,16 @@ private void createSimpleVariableReference(DiagnosticPos pos, Set<Whitespace> ws
         this.exprNodeStack.push(varRef);
     }
 
-    void createFunctionInvocation(DiagnosticPos pos, Set<Whitespace> ws, boolean argsAvailable) {
-        BLangInvocation invocationNode = (BLangInvocation) TreeBuilder.createInvocationNode();
+    void createFunctionInvocation(DiagnosticPos pos, Set<Whitespace> ws, boolean argsAvailable,
+                                  boolean remoteMethodCall) {
```
MaryamZi, line 1659:
```suggestion
                                  boolean actionInvocation) {
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangPackageBuilder.java
```diff
@@ -3194,6 +3213,7 @@ void addWorkerReceiveExpr(DiagnosticPos pos, Set<Whitespace> ws, String workerNa
         workerReceiveExpr.pos = pos;
         workerReceiveExpr.addWS(ws);
         //if there are two expressions, this is a channel receive and the top expression is the key
+        // TODO: Not needed?
```
MaryamZi, line 3216: Not sure if the changes to support multiple-receive-action will remove the key from normal worker receive. But shall we remove this and create an issue anyway?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
pubudu91: Yeah, it's duplicated. Will see about moving it to a common place

pubudu91: Addressed with ccf1fe7

PR #22817, opened 2020-04-21
Populate dependencies based on 'scope' parameter

### LINKED ISSUE #19271: [Packerina Spec] Dependency jars getting packed to balo
**Description:**
When a jar file is used as a dependency in a Ballerina module (via platform.libraries in the toml), that jar is getting packed with the balo. If the jar is proprietary (eg : SAP JCO jars), it's illegal to re-distribute like this. If we don't add this under platform.libraries in toml, the building of the module fails with CLASS_NOT_FOUND, since there're inter-op methods. 

It's better if we can have an attribute like 'scope' under platform.libraries, so that the jars are used only in the compile time and not packed with the balo.

suganyasuven:
Fixes #19271

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CopyNativeLibTask.java
```diff
@@ -120,6 +122,26 @@ private void copyImportedJarsForModules(BuildContext buildContext, List<BLangPac
         }
     }
 
+    private void copyPlatformLibsForModules(PackageID packageID, ExecutableJar executableJar, Path project) {
+        List<Library> libraries = manifest.getPlatform().libraries;
+        if (null == libraries) {
```
praveennadarajah, line 127:
```suggestion
        if (libraries == null) {
```

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CopyNativeLibTask.java
```diff
@@ -120,6 +122,26 @@ private void copyImportedJarsForModules(BuildContext buildContext, List<BLangPac
...
+            return;
+        }
+        List<Library> libs = libraries.stream().filter(lib -> lib.getModules() == null ||
+                Arrays.asList(lib.getModules()).contains(packageID.name.value)).collect(Collectors.toList());
+        for (Library lib : libs) {
+            String libFilePath = lib.getPath();
+            if (null != libFilePath) {
```
praveennadarajah, line 134:
```suggestion
            if (libFilePath != null) {
```

cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/DependencyScopeTest.java
```diff
@@ -0,0 +1,141 @@
...
+    public void setup() throws IOException {
+        super.setup();
+        try {
+            this.testResources = super.tmpDir.resolve("scope-test-resources");
+            testResourcesURI = getClass().getClassLoader().getResource("test-resources").toURI();
+            Path storedJarDependencyProject = Paths.get(testResourcesURI).resolve("stored-jar-dependency-project");
+            Files.walkFileTree(storedJarDependencyProject, new BuildCommandTest.Copy(storedJarDependencyProject,
+                    this.testResources));
+        } catch (URISyntaxException e) {
+            Assert.fail("error loading resources");
```
praveennadarajah, line 62:
```suggestion
            Assert.fail("Error loading resources");
```

suganyasuven: Error messages start with simple letters.

cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/DependencyScopeTest.java
```diff
@@ -0,0 +1,141 @@
...
+        Assert.assertFalse(isJarExists(baloZipFile, storedJarPath));
+    }
+
+    private boolean renameFile(File oldName, File newName) {
+        return oldName.renameTo(newName);
+    }
+
+    private boolean isJarExists (File file, String jarEntry) throws IOException {
+        try (ZipFile zipFile = new ZipFile(file)) {
+            Enumeration<? extends ZipEntry> entries = zipFile.entries();
```
praveennadarajah, line 128: No need to iterate all the zip entries and find whether a particular entry is there or not. The time complexity and performance hit would be high by doing this way. Instead, you can directly check a particular entry's presence in the following way.

`return new ZipFile(file).getEntry(jarEntry) != null;`

cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/DependencyScopeTest.java
```diff
@@ -0,0 +1,133 @@
...
+        copy(scopeToml.resolve("case2").resolve(ProjectDirConstants.MANIFEST_FILE_NAME),
+                this.testResources.resolve(ProjectDirConstants.MANIFEST_FILE_NAME));
+        buildCommand.execute();
+        Assert.assertTrue(Files.exists(balo), "Check if balo directory exists");
+
+        // Check whether dependency jars getting packed to balo
+        Assert.assertTrue(renameFile(baloFile, baloZipFile));
+        Assert.assertFalse(isJarExists(baloZipFile, storedJarPath));
+
+        // compile scope
```
IrushiL, line 109: When checking the Ballerina.toml file, it looks like "case3" should be the "test" scope.

cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/DependencyScopeTest.java
```diff
@@ -0,0 +1,133 @@
...
+        copy(scopeToml.resolve("case1").resolve(ProjectDirConstants.MANIFEST_FILE_NAME),
+                this.testResources.resolve(ProjectDirConstants.MANIFEST_FILE_NAME));
+        buildCommand.execute();
+        Assert.assertTrue(Files.exists(balo), "Check if balo directory exists");
+
+        // Check whether dependency jars getting packed to balo
+        Assert.assertTrue(renameFile(baloFile, baloZipFile));
+        Assert.assertTrue(isJarExists(baloZipFile, storedJarPath));
+
+        // package scope
```
IrushiL, line 99: When checking the Ballerina.toml file, it looks like "case2" should be the "compile" scope.

cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/DependencyScopeTest.java
```diff
@@ -0,0 +1,133 @@
...
+        copy(scopeToml.resolve("case1").resolve(ProjectDirConstants.MANIFEST_FILE_NAME),
+                this.testResources.resolve(ProjectDirConstants.MANIFEST_FILE_NAME));
+        buildCommand.execute();
+        Assert.assertTrue(Files.exists(balo), "Check if balo directory exists");
+
+        // Check whether dependency jars getting packed to balo
+        Assert.assertTrue(renameFile(baloFile, baloZipFile));
+        Assert.assertTrue(isJarExists(baloZipFile, storedJarPath));
```
IrushiL, line 97: This code segment seems to be repeating for all 3 scopes. Is it possible to separate it out and avoid code duplication?

compiler/ballerina-lang/src/main/java/org/ballerinalang/toml/model/Library.java
```diff
@@ -68,4 +69,12 @@ public String getArtifactId() {
     public void setArtifactId(String artifactId) {
         this.artifactId = artifactId;
     }
+
+    public String getScope() {
+        return scope;
+    }
+
+    public void setScope(String scope) {
```
IrushiL, line 77: This method does not seem to be used.

compiler/ballerina-lang/src/main/java/org/ballerinalang/toml/model/Library.java
```diff
@@ -28,6 +28,7 @@
     public String path;
     public String groupId;
     public String[] modules;
+    public String scope;
```
IrushiL, line 31: Can we make this a private field?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CopyNativeLibTask.java
```diff
@@ -120,6 +122,26 @@ private void copyImportedJarsForModules(BuildContext buildContext, List<BLangPac
...
+        if (libraries == null) {
+            return;
+        }
+        List<Library> libs = libraries.stream().filter(lib -> lib.getModules() == null ||
+                Arrays.asList(lib.getModules()).contains(packageID.name.value)).collect(Collectors.toList());
+        for (Library lib : libs) {
+            String libFilePath = lib.getPath();
+            if (libFilePath != null) {
+                Path nativeFile = project.resolve(Paths.get(libFilePath));
+                if (lib.getScope() != null && lib.getScope().equals("test")) {
```
warunalakshitha, line 136: if lib.getScope() == null means, default version? based on the logic scope null values are put in to moduleLibs.

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/writer/BaloFileWriter.java
```diff
@@ -298,15 +295,15 @@ private void addModuleDoc(Path root, Path moduleSourceDir) throws IOException {
...
-                                 HashSet<Path> moduleDependencyList) throws IOException {
+    private void addPlatformLibs(Path root, Path projectDirectory, String moduleName) throws IOException {
         //If platform libs are defined add them to balo
         if (null != manifest.getPlatform().libraries) {
             Path platformLibsDir = root.resolve(ProjectDirConstants.BALO_PLATFORM_LIB_DIR_NAME);
             Files.createDirectory(platformLibsDir);
 
             List<Path> libs = manifest.getPlatform().libraries.stream()
-                    .filter(lib -> lib.getModules() == null || Arrays.asList(lib.getModules()).contains(moduleName))
+                    .filter(lib -> (lib.getModules() == null || Arrays.asList(lib.getModules()).contains(moduleName))
```
warunalakshitha, line 305: Shall we replace with for loop without using java streams.

cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/DependencyScopeTest.java
warunalakshitha: May be we can have three test cases for each scope which will be more clear

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CopyNativeLibTask.java
suganyasuven: yes. We don't need that non-null check here.

compiler/ballerina-lang/src/main/java/org/ballerinalang/toml/model/Library.java
suganyasuven: Yes we can remove the setter

cli/ballerina-packerina/src/test/java/org/ballerinalang/packerina/cmd/DependencyScopeTest.java
```diff
@@ -0,0 +1,146 @@
...
+import java.nio.file.Paths;
+import java.util.zip.ZipFile;
+
+import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
+import static org.wso2.ballerinalang.compiler.util.ProjectDirConstants.BLANG_COMPILED_PKG_BINARY_EXT;
+
+/**
+ * Dependency scope tests.
+ *
+ * @since 1.2.x
```
warunalakshitha, line 43: Should be 1.3.0

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CopyNativeLibTask.java
```diff
@@ -239,6 +243,42 @@ private void copyLibsFromBalo(Path baloFilePath, Set<Path> moduleDependencySet)
         }
     }
 
+    private void copyAndValidateBaloDependencies(BuildContext buildContext, Path importDependencyPath,
```
warunalakshitha, line 246: If I import 2 modules (mod1, mod2) and mod1 has compiler dependency and same compiler dependency add as default in mod2. But that dependency will be packed in mod2 balo and at runtime everything will work fine. But still warning message will be shown during compilation as per below logic. Shall we fix that case as well

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CopyNativeLibTask.java
```diff
@@ -239,6 +243,42 @@ private void copyLibsFromBalo(Path baloFilePath, Set<Path> moduleDependencySet)
...
+        Manifest manifestFromBalo = RepoUtils.getManifestFromBalo(importDependencyPath);
+        List<Library> baloDependencies = manifestFromBalo.getPlatform().libraries;
+
+        if (baloDependencies == null) {
+            return;
+        }
+
+        // If platform libs are defined, copy them to target
+        List<Library> libraries = manifest.getPlatform().libraries;
+        for (Library baloTomlLib : baloDependencies) {
```
warunalakshitha, line 257: This logic has performace impact due to loop inside loop. Can we collect all compiler libs of balos in ahash set( key can be orgname+ packageid+ artfactID+ filename) and default libs two another set. Then we can compare those two hashsets two check first one is subset of default(packing) libs.

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CopyNativeLibTask.java
suganyasuven: Still, we need that warning message as dependencies of mod2 will be copied to the  mod2 and the current module only. For compiling mod1, we need to copy that dependency to mod1. As we are copying the dependencies based on the modules, the user have to define that dependency for mod1. Please correct me if I'm wrong.

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/CopyNativeLibTask.java
```diff
@@ -239,6 +244,46 @@ private void copyLibsFromBalo(Path baloFilePath, Set<Path> moduleDependencySet)
...
+            }
+        }
+
+        if (platformLibs.containsAll(baloCompileScopeDependencies)) {
+            return;
+        }
+
+        baloCompileScopeDependencies.removeAll(platformLibs);
+        for (Path baloTomlLib : baloCompileScopeDependencies) {
+            buildContext.out().println("warning: " + packageID + " is missing a native library dependency - " +
```
warunalakshitha, line 282: can we check !platformLibs.contains(baloTomlLib) and then log warning. so we can get rid of 276 to 280 lines

PR #22827, opened 2020-04-22
Add floating point literal support

### LINKED ISSUE #22358: Add floating point literal
**Description:**
Refer: https://ballerina.io/spec/lang/2020R1/#floating-point-literal

lochana-chathura:
Fixes #22358  

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -414,15 +414,19 @@ private STNode processEndOfLine() {
...
-    private STToken parseDotOrEllipsis() {
+    private STToken processDot() {
         if (reader.peek() == LexerTerminals.DOT && reader.peek(1) == LexerTerminals.DOT) {
             reader.advance(2);
             return getSyntaxToken(SyntaxKind.ELLIPSIS_TOKEN);
         }
+        if (this.mode != ParserMode.IMPORT && isDigit(reader.peek())) {
+            reader.retreat();
+            return processDecimalFloatLiteral();
+        }
```
lochana-chathura, line 429: I added `retreat()` method to the `charReader.java`. If it is not acceptable I can mange to do above without calling `reader.retreat();`. But that will add few more lines, if blocks and a new variables to the processNumericLiteral() and processDecimalFloatLiteral(); methods

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -543,9 +555,133 @@ private STToken processNumericLiteral(int startChar) {
...
+                    if (!exponent) { // Check whether an exponent has been passed previously
+                        if (processExponent()) {
+                            exponent = true;
+                            continue;
+                        }
+                    }
+
+                    processInvalidToken();
+                    return readToken();
+                case '.':
```
SupunS, line 601: lets use `LexerTerminals.DOT` everywhere posible

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
SupunS: I feel the lexer shouldn't reach here for floating-points.

Here we start parsing a token from a dot. But a floating-point numeric can never start from a dot right?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -543,9 +555,133 @@ private STToken processNumericLiteral(int startChar) {
...
+     * DecimalTypeSuffix := d | D
+     * <br/>
+     * FloatTypeSuffix :=  f | F
+     * </code>
+     *
+     * @return The decimal floating point literal.
+     */
+    private STToken processDecimalFloatLiteral() {
+        boolean exponent = false;
+        boolean dot = false;
```
SupunS, line 579: Lets rename:
  exponent -> hasExponent
  dot -> hasDot

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -543,9 +555,133 @@ private STToken processNumericLiteral(int startChar) {
...
+                case 'f':
+                case 'F':
+                case 'd':
+                case 'D':
+                    reader.advance();
+                    break;
+                case 'e':
+                case 'E':
+                    if (!exponent) { // Check whether an exponent has been passed previously
+                        if (processExponent()) {
```
SupunS, line 593: Can we avoid these if-conditions, and the flags?

Looking at the grammar, there are three main components the decimalNumber,  the exponent, and the suffix. These always exists in order. For example: decimalNumber component always the first component. suffix always comes after the decimalNumber, and always before the suffix.

So, what I feel is, suppose we have three separate functions to process these three separate components, (say `a()`, `b()` and `c()` ) if we chain these three, then the complexity would be reduced.

for e.g: we start from `a()` and continue till we reach the exponent, and once we reach the exponent, call `b()`. But never return to a() and process again. Simply finish the processing and exit from b(). So `a()` never have to worry whether `b()` was successfully completed or not..

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
lochana-chathura: according to spec, `DottedDecimalNumber := DecimalNumber . Digit*  | . Digit+ `
so starting from a `.` is possible.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
lochana-chathura: I'll make that change.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
lochana-chathura: Removed retreat() and refactored. :)

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -536,9 +547,162 @@ private STToken processNumericLiteral(int startChar) {
...
+                switch (nextChar) {
+                    case 'e':
+                    case 'E':
+                        return processExponent(false);
+                    case 'f':
+                    case 'F':
+                    case 'd':
+                    case 'D':
+                        return parseFloatingPointTypeSuffix();
+                }
```
SupunS, line 601: This code block looks like a duplicate of L#602-L#610. Can we use a single switch-case for both scenarios?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -536,9 +547,162 @@ private STToken processNumericLiteral(int startChar) {
...
+        }
+
+        // Make sure at least one digit is present after the indicator
+        if (!isDigit(nextChar)) {
+            processInvalidToken();
+            return readToken();
+        }
+
+        while (true) {
+            if (isDigit(nextChar)) {
```
SupunS, line 657: while(isDigit(nextChar))

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -536,9 +547,162 @@ private STToken processNumericLiteral(int startChar) {
...
+
+        // Capture if there is a sign
+        if (nextChar == LexerTerminals.PLUS || nextChar == LexerTerminals.MINUS) {
+            reader.advance();
+            nextChar = peek();
+        }
+
+        // Make sure at least one digit is present after the indicator
+        if (!isDigit(nextChar)) {
+            processInvalidToken();
```
SupunS, line 652: Should we log an error here?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -494,24 +499,31 @@ private STToken processEqualOperator() {
...
-                    reader.advance();
-                    processInvalidToken();
-                    return readToken();
+                    // Integer part of the float cannot have a leading zero
+                    if (startChar == '0' && len > 1) {
+                        processInvalidToken();
+                        return readToken();
+                    }
+
+                    return processDecimalFloatLiteral(false);
```
SupunS, line 526: Add a comment saying that we wouldn't be reaching here, if the floating-point starts with a dot.

Otherwise, anyone who will check this code later, might think this was a mistake :)

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -536,9 +547,162 @@ private STToken processNumericLiteral(int startChar) {
...
+     * @param isDotStart Whether process starts from a dot
+     * @return The decimal floating point literal.
+     */
+    private STToken processDecimalFloatLiteral(boolean isDotStart) {
+        int nextChar = peek();
+
+        // Direct process to the dot switch case
+        if (isDotStart) {
+            nextChar = LexerTerminals.DOT;
+        }
```
SupunS, line 576: can we get rid of this `isDotStart` flag? Why does this method needs to know whether the numeric started with a dot?

SupunS: What if this method always accepts the starting char as the input? Will that solve this complexity?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -547,22 +711,64 @@ private STToken processNumericLiteral(int startChar) {
...
         }
+        nextChar = peek();
 
-        return getLiteral(SyntaxKind.HEX_INTEGER_LITERAL);
+        switch (nextChar) {
+            case LexerTerminals.DOT:
+                reader.advance();
+                nextChar = peek();
+                while (true) {
+                    if (isHexDigit(nextChar)) {
```
SupunS, line 746:
```suggestion
                while (isHexDigit(nextChar)) {
```

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
lochana-chathura: the reason that I needed a `isDotStart` flag is that when `processDecimalFloatLiteral()`  is calling from `processDot()`, reader is already advanced for the `DOT`. But when the  `processDecimalFloatLiteral()` is calling from `processNumericalLiteral()`, reader is not advanced for the `DOT`. If we can fix either places we don't need a flag at all. Can't advance reader for `DOT` inside `processNumericalLiteral()` then it will cause problem for `ParserMode.IMPORT`.  Passing `starting char` won't help as what we want to know is not the starting char. we want want to know whether starting char is already advanced or not in the reader. 

As I said previously calling retreat() method inside processDot() would solve the problem. Then the `processDecimalFloatLiteral()` doesn't need to know whether it is calling from.

Any suggestion?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -3047,6 +3051,7 @@ private boolean isValidLVExpr(STNode expression) {
         switch (expression.kind) {
             case IDENTIFIER_TOKEN:
             case QUALIFIED_NAME_REFERENCE:
+            case SIMPLE_NAME_REFERENCE:
```
lochana-chathura, line 3054: Saw a bug here. `SIMPLE_NAME_REFERENCE` added and fixed.

Issue #20851, opened 2020-02-05
Move default values to the runtime 'type'


SupunS:
**Description:**

Currently default values are part of the 'value' and are populated during the initialisation of the value, by generating code. (e.g: records, objects) This design has several issues:
- There are situations where one can create a value using a type (e.g.: constructFrom, etc). In such cases the default value info is not available at the runtime, in the current model.
- Types can have closures. So the default value (expressions) should be evaluated at the point where the type is initialised, but not the value.

The solution is to make the default value info a part of the runtime type. 

This may also require to separate the type in to two parts as 'shape' and the 'type-descriptor'. Here the 'shape' will only have the type information (such as field names and their types) where as the 'type-descriptor' would have other information such as annotations, default values, etc. Every time a new value is created, a new type-descriptor will also be created. But the 'shape' is common across all instances of that type.
This way, we can make the runtime type checking faster, as it will only have to check for the equality of the 'shape'.

Currently this is a blocker for implementing the record/object default values with expressions (spec deviation).

manuranga:
```
public function main() {
   int x = 20;

   record {
       string name;
       int age = x;
   } person = { name: "Pubudu" };
}
```

Above sample should work after the proper fix. Got it form @pubudu91

manuranga:
# Sample Code:
```ballerina
public function main() returns (error?){
   int x = 20;
   record {
       string name;
       int age = x; 
   } person1 = { name: "Pubudu" };

   var personType = typeof person1;
  
   x = 2;

   var person2 = check personType.constructFrom({ name: "Manu" });


   io:println(person2.age);
}
```

# Solution:
convert `x` to a map representation like we do for lambdas
pass the created map to `NewMap` instruction just like we do for `NewFP`
```
...
        %2 = NewMap map<any | error{string, $anonType$builtin$0}>;
        %6 = ConstLoad 20;
        %5 = <any | error> %6;
        %8 = ConstLoad x;
        %2[%8] = %5;
        %12 = NewMap $anonType$0(%2);
...
```
Values(eg: `person1`) has to keep a reference to the map, because it now needs that value to create type desc when called `typeof`
`TypedescValue` will have an addition value to keep default value map
`BallerinaValues.createRecordValue` should accept additional map to get default values.

SupunS:
I prefer the solution 2, with some minor adjustment. 

[...]

PR #22906, opened 2020-04-27
Split `newMap` instruction into `newType` and `newMap`

### LINKED ISSUE #22772: Change NewMap instructions to accept a typedesc as the argument
Sub task of #20851

change 
```
        %12 = NewMap $anonType$0;
```
to
```
        %9 = NewTypeDesc $anonType$0;
        %12 = NewMap %9;
```

manuranga:
Fixes #22772
Sub task of #20851


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/api/BTypedesc.java
```diff
@@ -38,4 +39,10 @@
      * @return describing type
      */
     BType getDescribingType();
+
+    /**
+     * @param strand strand to be used to run the user-defined-type initialization code.
+     * @return instantiated object
+     */
+    Object instantiate(Strand strand);
```
Kishanthan, line 47: Do we have to pass in the strand? This is going to be generated method at runtime right?

manuranga: this method calls the `__init` of the newly created object, for that we need a strand, if not `start` keyword inside `__init` will not work.

Issue #22680, opened 2020-04-16
Improve enum support


irshadnilam:
**Description:**

Lang issue to track  https://github.com/ballerina-platform/ballerina-spec/issues/276
Changes introduced in https://github.com/ballerina-platform/ballerina-spec/commit/6e13cb39db7c70ddc7b61d38e6c9ddc63528697f

**Suggested Labels (optional):**
Area/Language

**Suggested Assignees (optional):**
@irshadnilam

PR #22907, opened 2020-04-27
Add support for Enums

irshadnilam:
## Purpose
This PR adds support for `enum` syntax. 

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/22680

## Samples
Adds support for the following
```ballerina
public enum Color {
  RED,
  GREEN,
  BLUE
}
```
is equivalent to 

```ballerina
public const RED = "RED";
public const GREEN = "GREEN";
public const BLUE = "BLUE";
public type Color RED|GREEN|BLUE;
```

Enum members can also have expressions of type string.
```ballerina
public enum Color {
  RED = "red",
  GREEN,
  BLUE
}
```
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


whizsid:
Can I use tuples with enum? This feature is already in Rust language and I like it.

```ballerina
enum {
   TWO_DIMENSIONAL(int,int);
   THREE_DIMENTIONAL(int,int,int);
}
```

irshadnilam:
> Can I use tuples with enum? This feature is already in Rust language and I like it.
> 
> ```ballerina
> enum {
>    TWO_DIMENSIONAL(int,int);
>    THREE_DIMENTIONAL(int,int,int);
> }
> ```

@whizsid Thank you for bringing this suggestion. 

Would you able to open an issue with this suggestion on ballerina-spec repository?
https://github.com/ballerina-platform/ballerina-spec/

hasithaa:
@irshadnilam We need more test cases, when enum value is used as function parameters, etc. This is already covered by the constants. But better to have a separate set of test cases for this.

compiler/ballerina-lang/src/main/resources/grammar/BallerinaParser.g4
```diff
@@ -156,6 +157,15 @@ constantDefinition
     :   PUBLIC? CONST typeName? Identifier ASSIGN constantExpression SEMICOLON
     ;
 
+enumDefinition
+    :   documentationString? annotationAttachment* PUBLIC? ENUM Identifier LEFT_BRACE
+            (enumMember (COMMA enumMember)*)? RIGHT_BRACE
```
MaryamZi, line 162: The grammar for module-enum-decl in the spec is 

```
module-enum-decl :=
   metadata
   [public] enum identifier { enum-member (, enum-member)* }
```
At least one enum-member is required.

So shouldn't this be

```suggestion
            enumMember (COMMA enumMember)* RIGHT_BRACE
```

MaryamZi: The following results in a bad, sad error atm.

```ballerina
public enum Color {

}
```

compiler/ballerina-lang/src/main/resources/grammar/BallerinaParser.g4
```diff
@@ -156,6 +157,15 @@ constantDefinition
     :   PUBLIC? CONST typeName? Identifier ASSIGN constantExpression SEMICOLON
     ;
 
+enumDefinition
+    :   documentationString? annotationAttachment* PUBLIC? ENUM Identifier LEFT_BRACE
```
MaryamZi, line 161: The spec says the attachment point for `module-enum-decl` is `type`.

But this seems to be matching for consts now.

```ballerina
public annotation X on type;
public const annotation Y on source const;

@X // should work, doesn't
@Y // should fail, doesn't
public enum Color {
    RED,
    BLUE
}
```

MaryamZi: https://github.com/ballerina-platform/ballerina-lang/issues/23308 tracks fixing this,

compiler/ballerina-lang/src/main/resources/grammar/BallerinaParser.g4
MaryamZi: https://github.com/ballerina-platform/ballerina-lang/issues/23308 tracks fixing this,

PR #22914, opened 2020-04-27
[1.2.x] Add start and complete methods to Compiler Plugins

### LINKED ISSUE #11205: Need to re-enable compilerplugin runner tests
**Description:**
As of now the compiler plugin runner tests are commented out since they are not updated with the latest syntax changes. But we should update them and re-enable the tests again.

### LINKED ISSUE #22898: Cannot Detect the Completion of Traversing the AST
**Description:**
Currently, there is no way of detecting whether a visitor has completed traversing the AST. If we have created a custom annotation and we want to know that we have completed visiting all the occurrences of that annotation, currently there is no means of doing this. It would be better if we could get a new method added to AbstractCompilerPlugin class so that we can detect this situation. Note that we cannot use `public void codeGenerated(PackageID packageID, Path binaryPath)` method here because it get called only when the Java code generation phase executes. In my use case this method does not get called because I use the `-c` flag.

**Steps to reproduce:**
N/A

**Affected Versions:**
Ballerina 1.2.x

N/A

hasithaa:
Fix #11205
Fix #22898

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/ballerina-compiler-plugin-test/src/test/java/io/ballerina/test/compiler/plugins/CompilerPluginTest.java
```diff
@@ -37,97 +38,71 @@
...
                 }});
 
         assertData(TestEvent.Kind.FUNC_ANN,
-                xyzEventMap,
+                funcEvents,
                 new ArrayList<TestEvent>() {{
                     add(new TestEvent(TestEvent.Kind.FUNC_ANN, "routerFunc", 1));
                 }});
 
-        // Test variable events
```
hasithaa, line 109: These cases are not supported in the current annotation design.

PR #22916, opened 2020-04-27
[1.2.x] Add start and complete methods to Compiler Plugins

### LINKED ISSUE #11205: Need to re-enable compilerplugin runner tests

### LINKED ISSUE #22898: Cannot Detect the Completion of Traversing the AST

hasithaa:
Fix #11205
Fix #22898

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/ballerina-compiler-plugin-test/src/test/resources/META-INF/services/org.ballerinalang.compiler.plugins.CompilerPlugin
```diff
@@ -1,2 +1,2 @@
-io.ballerina.test.compiler.plugins.ABCCompilerPlugin
-io.ballerina.test.compiler.plugins.XYZCompilerPlugin
\ No newline at end of file
+io.ballerina.test.compiler.plugins.TestCompilerPlugin
+io.ballerina.test.compiler.plugins.FunctionsTestCompilerPlugin
```
rdhananjaya, line 2: Is this not ending with a new line expected here?

compiler/ballerina-lang/src/main/java/org/ballerinalang/compiler/plugins/CompilerPlugin.java
```diff
@@ -116,8 +116,25 @@ default void setCompilerContext(CompilerContext context) {
...
      */
     void codeGenerated(PackageID packageID, Path binaryPath);
+
+
+    /**
+     * Notifies when the compiler starts executing compiler plugins for a particular module.
+     *
+     * @param packageID PackageID of the module
+     */
+    default void pluginExecutionStarted(PackageID packageID) {
```
MaryamZi, line 130: Do we need the "plugin" prefix, given that these are methods in `CompilerPlugin`? Wouldn't just `executionStarted` and `executionCompleted` be enough?

tests/ballerina-compiler-plugin-test/src/test/resources/META-INF/services/org.ballerinalang.compiler.plugins.CompilerPlugin
hasithaa: It is an auto-generated file.

compiler/ballerina-lang/src/main/java/org/ballerinalang/compiler/plugins/CompilerPlugin.java
hasithaa: I thought the same, But we have a method called codeGenerated, which is executing in a later phase. So executionCompleted is not the right word, So that is why I have added plugin prefix.

PR #22962, opened 2020-04-29
Add query-expr clause to handle duplicate table keys

### LINKED ISSUE #22641: Add query-expr clause to control handling of duplicate table keys
Refer to https://github.com/ballerina-platform/ballerina-spec/issues/481
The query clause might look like this (only makes sense when creating a table):

```
table key(id) from r in stm
  select r
  duplicate var prev io:println(`duplicate for ${r.id}: ${r.name} vs ${prev.name}`)
```
or

```
table key(id) from r in stm
  select r
  duplicate var prev DuplicateError("duplicate detected", this=r, prev=prev)
```

i.e. syntax is

```
duplicate-clause := "duplicate" typed-binding-pattern expression
```

Semantics is:

- `duplicate` clause gets executed when select emits a row that has the same key as a row that it emitted earlier
- static type of expression must be subtype of `error?`
- bind typed-binding-pattern to the previous row that has the same key
- evaluate expression resulting in v; evaluation of expression is allowed to have side-effects such as logging
- if v is an error, then the result of the query-expr is v
- otherwise (v is nil), replace earlier row with new row and continue without giving an error
 
Issues:
* Not sure about the right keyword. Maybe should be something with `key` in it.
* Is providing the previous value sufficiently useful to justify the complexity?
* We could get more complex and allow user
     * to select which row they wanted or
     * to merge the two rows with the same key

_Originally posted by @jclark in https://github.com/ballerina-platform/ballerina-spec/issues/477#issuecomment-609531425_

lasinicl:
Fixes #22641

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


lasinicl:
Changes included in a different PR

Issue #22515, opened 2020-04-08
Add support for `readonly & T` 


MaryamZi:
**Description:**
$title.

See https://github.com/ballerina-platform/ballerina-spec/issues/476.

MaryamZi:
https://github.com/ballerina-platform/ballerina-lang/pull/22965 introduced support for `T & readonly`.

Closing this issue since the remaining sub-tasks are being tracked with other issues.

- Support `T & readonly` for table types #23032
- Disallow langlib methods that modify the value on immutable values #22964 
- Allow constructing values with their read-only bit on with a type-cast-expr `<readonly>` #22954
- Allow using the type-test for `readonly` with selectively immutable basic types #22785
- Support readonly record fields #22600
- The type for a `readonly T x;` record field needs to be set as `T & readonly` #23257

Issue #22964, opened 2020-04-29
Disallow langlib methods that modify the value on immutable values 


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
The following should result in a compilation error.
```ballerina
public function main() {
    int[] & readonly arr = [1, 2];

    arr.push(3);
}
```

MaryamZi:
Sub-task of https://github.com/ballerina-platform/ballerina-lang/issues/22515.

Issue #22954, opened 2020-04-29
Allow constructing values with their read-only bit on with a type-cast-expr `<readonly>`


MaryamZi:
**Description:**
$title.

The spec says

> Note that a type-cast-expr of <readonly> can be used both to cause constructors withing the expression to construct values with the read-only bit on and to verify that the value resulting from the evaluation of expression has its read-only bit on.

Sub-task of https://github.com/ballerina-platform/ballerina-lang/issues/22515

MaryamZi:
Moreover, the static type of `<readonly> T` should be `T & readonly`.

e.g.,
```ballerina
function foo(map<int> m) {
    map<int> & readonly n = <readonly> m; // valid
}
```

MaryamZi:
The original issue will be fixed with https://github.com/ballerina-platform/ballerina-lang/pull/24484.

> Moreover, the static type of `<readonly> T` should be `T & readonly`.
> 
> e.g.,
> 
> ```ballerina
> function foo(map<int> m) {
>     map<int> & readonly n = <readonly> m; // valid
> }
> ```

This will be tracked with https://github.com/ballerina-platform/ballerina-lang/issues/24559.

PR #22965, opened 2020-04-29
Add support for `T & readonly`

### LINKED ISSUE #22950: Setting members to a tuple assigned to a JSON array-typed variable does not work
**Description:**
$title. Doesn't fail, but doesn't add either.

This is due to the check at https://github.com/ballerina-platform/ballerina-lang/blob/v1.2.2/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/JSONUtils.java#L251

IMO, we don't have to special case JSON arrays anymore, since member access can be used for assignment only when the type is guaranteed to be a list (or map).

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    [json...] a = [];
    json[] b = a;

    b[0] = {hello: "world"};
    b[1] = 2;

    io:println(b.length()); // prints `0`
}
```

**Affected Versions:**
1.2.2

### LINKED ISSUE #23050: Rewrite mapping/list value creation
**Description:**
$title.

Right now, what we do is create the initial mapping/list and then populate the initial values separately (i.e., desugared as container creation + index-based assignments).

But with the introduction of `T & readonly` types, and values being created as immutable values this leads to issues since setting of the initial values are also considered invalid updates.

Instead we should allow specifying initial values when creating the mapping/list itself.

See the [related discussion on slack](https://ballerina-platform.slack.com/archives/CAXS256RX/p1588314320197000).

Prerequisite for https://github.com/ballerina-platform/ballerina-lang/issues/22515

MaryamZi:
## Purpose
$title, this allows creating values known to be immutable at compile-time.
Main changes to fix https://github.com/ballerina-platform/ballerina-lang/issues/22515.

Fixes #23050 - Rewrite mapping/list value creation

This PR also fixes #22950 - Setting members to a tuple assigned to a JSON array-typed variable does not work

## Samples
```ballerina
import ballerina/io;

type Employee record {|
    PersonalDetails details;
    string department;
|};

type PersonalDetails record {|
    string name;
    int id;
|};

public function main() {
    Employee e1 = {
        details: {
            name: "Jo",
            id: 1234	
        },
        department: "finance"
    };

    // Compilation Error: 
    // incompatible types: expected 'Employee & readonly', found 'Employee'
    // Employee & readonly e2 = e1; 
   
    // Immutable `Employee` value.
    Employee & readonly e3 = {
        details: {
            name: "May",
            id: 7890	
        },
        department: "finance"
    };   

    // Compilation Error: 
    // cannot update 'readonly' record field 'department' in 'Employee & readonly'
    // e3.department = "IT";

    io:println(e3.isReadOnly()); // true
}
```

## Remarks
To Do
- Tests (WIP)
- Allow constructing values with their read-only bit on with a type-cast-expr `<readonly>` #22954
- Disallow langlib methods that modify the value on immutable values #22964

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


MaryamZi:
Closing this PR temporarily since some significant changes need to be introduced to improve how the immutable values are created. Will re-open once done.

Kishanthan:
@MaryamZi can you check for alignment issues in this PR?

MaryamZi:
> @MaryamZi can you check for alignment issues in this PR?

@Kishanthan do you mean in Java code or Ballerina? I'm not sure I understood where the issues are. Can you please point out one maybe?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmValueGen.java
```diff
@@ -680,8 +688,12 @@ private void createRecordInitWrapper(ClassWriter cw, String className, BIRNode.B
                 String refTypeClassName = getTypeValueClassName(typeRef.tsymbol.pkgID,
                         toNameString(typeRef));
                 mv.visitInsn(DUP2);
+
+                mv.visitInsn(ICONST_0);
+                mv.visitTypeInsn(ANEWARRAY, MAPPING_INITIAL_VALUE_ENTRY);
+
                 mv.visitMethodInsn(INVOKESTATIC, refTypeClassName, "$init",
-                        String.format("(L%s;L%s;)V", STRAND, MAP_VALUE), false);
+                        String.format("(L%s;L%s;[L%s;)V", STRAND, MAP_VALUE, BINITIAL_VALUE_ENTRY), false);
```
manuranga, line 696: I am in the process of removing `createRecordInitWrapper` method, can we please move (or duplicate) this in `createInstantiateMethod`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
```diff
@@ -1336,33 +1390,27 @@ void generateArrayStoreIns(BIRNonTerminator.FieldAccess inst) {
...
-                    String.format("(JL%s;)V", JvmPackageGen.IS_BSTRING ? JvmConstants.B_STRING_VALUE :
-                            STRING_VALUE), true);
-        } else if (valueType.tag == TypeTags.BOOLEAN) {
-            this.mv.visitMethodInsn(INVOKEINTERFACE, ARRAY_VALUE, "add", "(JZ)V", true);
-        } else if (valueType.tag == TypeTags.BYTE) {
-            this.mv.visitInsn(I2B);
-            this.mv.visitMethodInsn(INVOKEINTERFACE, ARRAY_VALUE, "add", "(JB)V", true);
-        } else {
-            this.mv.visitMethodInsn(INVOKEINTERFACE, ARRAY_VALUE, "add", String.format("(JL%s;)V", OBJECT), true);
+            BType valueType = inst.rhsOp.variableDcl.type;
```
Kishanthan, line 1393: looks like there is an alignment issue here. probably due to merge conflicts with jvm codegen package classes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
```diff
@@ -1336,33 +1390,27 @@ void generateArrayStoreIns(BIRNonTerminator.FieldAccess inst) {
...
+                                        String.format("(JL%s;)V", JvmPackageGen.IS_BSTRING ?
+                                                JvmConstants.B_STRING_VALUE : STRING_VALUE), true);
+            } else if (valueType.tag == TypeTags.BOOLEAN) {
+                this.mv.visitMethodInsn(INVOKEINTERFACE, ARRAY_VALUE, method, "(JZ)V", true);
+            } else if (valueType.tag == TypeTags.BYTE) {
+                this.mv.visitInsn(I2B);
+                this.mv.visitMethodInsn(INVOKEINTERFACE, ARRAY_VALUE, method, "(JB)V", true);
+            } else {
+                this.mv.visitMethodInsn(INVOKEINTERFACE, ARRAY_VALUE, method, String.format("(JL%s;)V", OBJECT), true);
+            }
```
Kishanthan, line 1412: looks like there is an alignment issue here. probably due to merge conflicts with jvm codegen package classes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
```diff
@@ -1583,14 +1631,17 @@ void generateFPLoadIns(BIRNonTerminator.FPLoad inst, LambdaMetadata lambdaMetada
...
+            this.mv.visitTypeInsn(CHECKCAST, XML_QNAME);
+            this.loadVar(newXMLElement.defaultNsURIOp.variableDcl);
+            loadType(this.mv, newXMLElement.type);
+
+            this.mv.visitMethodInsn(INVOKESTATIC, XML_FACTORY, "createXMLElement",
+                                    String.format("(L%s;L%s;L%s;)L%s;", XML_QNAME,
+                                                  JvmPackageGen.IS_BSTRING ? JvmConstants.B_STRING_VALUE :
+                                                          STRING_VALUE, BTYPE, XML_VALUE), false);
+            this.storeToVar(newXMLElement.lhsOp.variableDcl);
+        }
```
Kishanthan, line 1644: looks like there is an alignment issue here. probably due to merge conflicts with jvm codegen package classes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
```diff
@@ -1627,23 +1678,28 @@ void generateNewXMLTextIns(BIRNonTerminator.NewXMLText newXMLText) {
...
-    }
+            this.loadVar(newXMLPI.targetOp.variableDcl);
+            this.loadVar(newXMLPI.dataOp.variableDcl);
+            loadType(this.mv, newXMLPI.type);
+
+            String consVal = JvmPackageGen.IS_BSTRING ? JvmConstants.B_STRING_VALUE : STRING_VALUE;
+            this.mv.visitMethodInsn(INVOKESTATIC, XML_FACTORY, "createXMLProcessingInstruction",
+                                    String.format("(L%s;L%s;L%s;)L%s;", consVal, consVal, BTYPE, XML_VALUE), false);
+            this.storeToVar(newXMLPI.lhsOp.variableDcl);
+        }
```
Kishanthan, line 1702: looks like there is an alignment issue here. probably due to merge conflicts with jvm codegen package classes.

Kishanthan:
> > @MaryamZi can you check for alignment issues in this PR?
> 
> @Kishanthan do you mean in Java code or Ballerina? I'm not sure I understood where the issues are. Can you please point out one maybe?

Pl check now. I have added few comments. you have to check the merge conflicts classes with jvm codegen package for this issue.

MaryamZi:
> > > @MaryamZi can you check for alignment issues in this PR?
> > 
> > 
> > @Kishanthan do you mean in Java code or Ballerina? I'm not sure I understood where the issues are. Can you please point out one maybe?
> 
> Pl check now. I have added few comments. you have to check the merge conflicts classes with jvm codegen package for this issue.

Thank you, missed this. Fixed in 6f89a47.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmValueGen.java
MaryamZi: We don't need initial values for referenced record types, so having the initial value population logic in "instantiate" should be enough. Moved to `createInstantiateMethod` in 6f89a47.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2106,6 +2112,9 @@ public void visit(BLangFieldBasedAccess fieldAccessExpr) {
                     if (isAllReadonlyTypes(varRefType)) {
                         dlog.error(fieldAccessExpr.pos, DiagnosticCode.CANNOT_UPDATE_READONLY_VALUE_OF_TYPE,
                                    varRefType);
+                        types.checkType(fieldAccessExpr, actualType, this.expType);
```
KRVPerera, line 2115: Why do we call `checkType` here?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -1334,7 +1334,8 @@ private void defineErrorDetails(List<BLangTypeDefinition> typeDefNodes, SymbolEn
     }
 
     private void defineFields(List<BLangTypeDefinition> typeDefNodes, SymbolEnv pkgEnv) {
-        for (BLangTypeDefinition typeDef : typeDefNodes) {
+        for (int i = 0; i < typeDefNodes.size(); i++) {
```
KRVPerera, line 1337: Is this a performance improvement? I think this is unnecessary.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1154,6 +1166,87 @@ private static boolean isReadonlyType(BType sourceType) {
         return false;
     }
 
+    boolean isSelectivelyImmutableType(BType type) {
+        return isSelectivelyImmutableType(type, new HashSet<>());
+    }
+
+    public static boolean isSelectivelyImmutableType(BType type, Set<BType> unresolvedTypes) {
```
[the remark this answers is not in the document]
MaryamZi: No, this does not check for the read-only bit.

Even though mappings in general are selectively immutable, some mappings can never be immutable. e.g., `map<future>`. This method checks if the specified type can be immutable. 

If it's still unclear maybe we can call this something like `isPossiblyImmutableSelectivelyImmutableType`, but given that the logic is quite obvious I'm not sure we need it.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
MaryamZi: This is to avoid potential concurrent modification errors since we may have to define new record type defs while iterating through the current list.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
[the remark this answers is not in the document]
KRVPerera: Okay. I was thinking if someone wanted to write a method in future to check the read-only bit; then we may need to change the name at that time.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2168,14 +2208,19 @@ public void visit(BLangIndexBasedAccess indexBasedAccessExpr) {
...
         BLangExpression indexExpr = indexBasedAccessExpr.indexExpr;
+
         if (actualType != symTable.semanticError &&
-                types.isSubTypeOfBaseType(exprType, TypeTags.RECORD) &&
-                (indexBasedAccessExpr.lhsVar || indexBasedAccessExpr.compoundAssignmentLhsVar) &&
-                (indexExpr.getKind() == NodeKind.LITERAL || isConst(indexExpr))) {
-            String name = getConstFieldName(indexExpr);
-            if (allRecordsHaveRequiredReadonlyField(exprType, name)) {
+                (indexBasedAccessExpr.lhsVar || indexBasedAccessExpr.compoundAssignmentLhsVar)) {
+            if (isAllReadonlyTypes(exprType)) {
```
MaryamZi, line 2214: Check for optional fields.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/XMLFactory.java
```diff
@@ -345,6 +386,31 @@ public static XMLValue createXMLProcessingInstruction(BString tartget, BString d
...
+
+    /**
+     * Create a processing instruction type XMLValue, specifying the type which will indicate mutability.
+     *
+     * @param target    PI target
+     * @param data      PI data
+     * @param type      The type for the PI
+     * @return XMLValue Processing instruction type XMLValue
+     */
+    public static XMLValue createXMLProcessingInstruction(BString target, BString data, BType type) {
```
MaryamZi, line 410: Accept a flag instead and set relevant type internally.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1481,7 +1497,12 @@ private boolean validateRequiredAndReadonlyFields(BRecordType type,
                         types.isAssignable(typePosPair.type, symTable.readonlyType)) {
                     continue;
                 }
-                dlog.error(typePosPair.pos, DiagnosticCode.INVALID_FIELD_FOR_READONLY_RECORD_FIELD, fieldName);
+
+                if (mutableRecordType && !Symbols.isFlagOn(field.type.flags, Flags.READONLY)) {
```
pubudu91, line 1501: The `!Symbols.isFlagOn(field.type.flags, Flags.READONLY)` check is not relevant here right? IIUC, this checks if there's a `readonly` field in a mutable record type? Plus, isn't it the same check as in the if block above? Is the field being `readonly` and the field's type being `readonly` two different things?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: This is not required since what we are checking here is something like

```
(Foo & readonly)|(Bar & readonly) x = {};
x["val"] = 1;
```

This is always invalid since the value itself is immutable and cannot be updated.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Yeah they're different.

So there are three possible scenarios.

1. The expected type is `T & readonly` where T is the record type. Meaning, the value itself will be immutable and whatever fields that are specified should also be immutable. This is what is checked with `mutableRecordType`.

```ballerina
Foo & readonly f = {};
```

2. A field's type is `T & readonly`. This means that whatever value set for that field will have to be immutable. That's the second check.

```ballerina
type Foo record {|
    Bar & readonly b;
|};
``` 
And you can also update the `b` field in a `Foo` value after creating  the record value, but the `Bar` value must always be immutable.

3. Then there's `readonly` fields where in addition to 2) there's also the constraint that the value cannot be updated once it is set.

```ballerina
type Foo record {|
    readonly Bar b;
|};
```

We only want to log an error here for 3.

PR #22983, opened 2020-04-29
Add query-expr clause to handle duplicate table keys

### LINKED ISSUE #22641: Add query-expr clause to control handling of duplicate table keys

lasinicl:
Fixes #22641

```Person[]|error outputPersonList =
            from var person in personList
            select {
                   firstName: person.firstName,
                   lastName: person.lastName,
                   age: person.age
            }
            on conflict error("PersonNotFound");
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/DataflowAnalyzer.java
```diff
@@ -818,6 +819,8 @@ public void visit(BLangQueryExpr queryExpr) {
         }
 
         analyzeNode(queryExpr.selectClause, env);
+
```
mohanvive, line 822: No need line space here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -3403,6 +3408,15 @@ private SymbolEnv typeCheckWhereClause(BLangWhereClause whereClause, BLangSelect
         return typeNarrower.evaluateTruth(whereClause.expression, selectClause, parentEnv);
     }
 
+    private SymbolEnv typeCheckOnConflictClause(BLangOnConflictClause onConflictClause, SymbolEnv parentEnv) {
```
mohanvive, line 3411: No need to return SymbolEnv, this could be a void method.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/query/OnConflictClauseTest.java
```diff
@@ -0,0 +1,58 @@
+/*
```
mohanvive, line 1: Write negative testcase for type checker validation that you added.

tests/jballerina-unit-test/src/test/resources/test-src/query/on-clause.bal
```diff
@@ -0,0 +1,31 @@
...
+
+    Person p1 = {firstName: "Alex", lastName: "George", age: 23};
+    Person p2 = {firstName: "Ranjan", lastName: "Fonseka", age: 30};
+    Person p3 = {firstName: "John", lastName: "David", age: 33};
+
+    Person[] personList = [p1, p2, p3];
+
+    Person[] outputPersonList =
+            from var person in personList
+            on printHello()
```
grainier, line 19: Better to use a proper example. This shouldn't be allowed and type checker should only allow following [1];

`join-on-condition := "on" expression "equals" expression`

[1] https://github.com/ballerina-platform/ballerina-spec/issues/435

tests/jballerina-unit-test/src/test/resources/test-src/query/on-conflict-clause-negative.bal
```diff
@@ -0,0 +1,31 @@
...
+    Person[] personList = [p1, p2, p3];
+
+    Person[]|error outputPersonList =
+            from var person in personList
+            select {
+                   firstName: person.firstName,
+                   lastName: person.lastName,
+                   age: person.age
+            }
+            on conflict printHello();
```
grainier, line 24: Same as previous; better to use a proper example.

grainier:
merged in : https://github.com/grainier/ballerina/pull/29

Issue #22987, opened 2020-04-30
Enable the following disabled tests


dulvinw:
**Description:**
The following lock tests were disabled due to being intermittent. Please fix them and enable.

LocksInMainTest > simpleLock 
LocksInMainTest > testThrowErrorInsideLock
LocksInMainTest > testReturnInsideLock
LocksInMainTest > testThrowErrorInsideTryFinallyInsideLock
LocksInMainTest > testParallelRunUsingLocks
**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #22991, opened 2020-04-30
Disable intermittent lock tests

dulvinw:
Purpose
Few lock tests are failing intermittently. Hence disabling for the moment and to be fixed later.

#22987 is created to track the disabled test fixes


shafreenAnfar:
Please don't merge this PR until #22993 #22995 pass.

PR #23148, opened 2020-05-07
Introduce add and put function for table type

pcnfernando:

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/TableValueImpl.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/TableValueImpl.java
@@ -172,6 +172,7 @@ public class TableValueImpl<K, V> implements TableValue<K, V> {
 
     //Generates the key from the given data
     public V put(V value) {
+        handleFrozenTableValue();
         return valueHolder.putData(value);
     }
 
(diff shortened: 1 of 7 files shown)


mohanvive:
@pcnfernando Please send this PR directly to master.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/TableValueImpl.java
```diff
@@ -376,19 +389,13 @@ public V getData(K key) {
             return values.get(TableUtils.hash(key, new ArrayList<>()));
         }
 
-        public V putData(K key, V data) {
```
gimantha, line 379: Have you checked the case where the key is something we give intentionally, but the actual key is something else.
For example,
table<Person> key(name) t = table[];
t["Chiran"] = {name: "NotChiran"};
How does the above case work?

pcnfernando: This works.
But when I use an Int as key and use it for index based access I get an ArrayOutOfBound.

Below is the sample I tried.

```ballerina
type Customer record {
  readonly int id;
  string firstName;
  string lastName;
};
type CustomerTable table<Customer> key(id);

function testTableIndexAccess() returns boolean {
boolean testPassed = true;
CustomerTable custTbl = table [
{ id: 1, firstName: "Sanjiva", lastName: "Weerawarana" },
{ id: 5, firstName: "Gimantha", lastName: "Bandara" }
];

custTbl[5] = { id: 5, firstName: "Gimantha", lastName: "Bandaraaa" };

return testPassed;
}
```

gimantha: @pcnfernando can you plz get an update from the master and see if this is fixed.. I fixed a similar issue

pcnfernando: Yes, this works as expected now. Thanks

pcnfernando:
Let's not merge this until the inconsistent issue is figured out.

PR #23197, opened 2020-05-13
Change `constructFrom` to use typedesc for user defined types.

### LINKED ISSUE #23012: Refactor `constructFrom` to use generated typedesc classes
Sub task of #20851
```ballerina
public function main() returns (error?){
   int x = 20;
   record {
       string name;
       int age = x; 
   } person1 = { name: "Pubudu" };

   var personType = typeof person1;
  
   x = 2;

   var person2 = check personType.constructFrom({ name: "Manu" });


   io:println(person2.age);
}
```
above test should work and print 2

manuranga:
Fixes #23012 for user defined types.


manuranga:
This fixes #20851

Issue #23276, opened 2020-05-14
Identify and fix issues in `xml<T> & readonly`


MaryamZi:
**Description:**
$title. Need more in depth tests to identify any issues.

The following issues are present for `xml:Element & readonly` with attributes.

Case I - the type of the attribute map is not set as readonly
```ballerina
import ballerina/lang.'xml as xmllib;

public function main() {
    xmllib:Element & readonly x = xml `<book status="available" count="5"/>`;

    // The following fails with 
    // `incompatible types: expected 'map<string> & readonly', found 'map<string>'`
    map<string> & readonly attributeMap = x.getAttributes();
}
```

Case II - `setAttribute` fails at runtime since the value is considered immutable - need to have a mechanism to distinguish initalization updates vs subsequent updates

```ballerina
import ballerina/lang.'xml as xmllib;

public function main() {
    xmllib:Element & readonly x = xml `<book status="available" count="5"/>`;

    map<string> attributeMap = x.getAttributes();
}
```

This **crashes** at runtime - ballerina-internal.log

```
[2020-05-15 00:02:43,513] SEVERE {b7a.log.crash} - {ballerina/lang.xml}InvalidUpdate 
org.ballerinalang.jvm.util.exceptions.BLangFreezeException: {ballerina/lang.xml}InvalidUpdate
	at org.ballerinalang.jvm.values.ReadOnlyUtils.handleInvalidUpdate(ReadOnlyUtils.java:61)
	at org.ballerinalang.jvm.values.XMLItem.setAttribute(XMLItem.java:202)
	at org.ballerinalang.jvm.values.XMLValue.setAttribute(XMLValue.java:75)
	at qwe.main(qwe.bal:4)
	at ___init.$lambda$main$(.)
	at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:486)
	at org.ballerinalang.jvm.scheduling.Scheduler.run(Scheduler.java:267)
	at org.ballerinalang.jvm.scheduling.Scheduler.runSafely(Scheduler.java:235)
	at java.lang.Thread.run(Thread.java:745)
```

Sub-task of #22515

MaryamZi:
Created https://github.com/ballerina-platform/ballerina-spec/issues/520 regarding Case I.

The following should work regardless.

```ballerina
import ballerina/lang.'xml as xmllib;

public function main() {
    xmllib:Element & readonly x = xml `<book status="available" count="5"/>`;

[...]

Issue #23032, opened 2020-04-30
Support `T & readonly` for table types


MaryamZi:
**Description:**
Now that the table implementation is merged, we can $title.

Sub-task of https://github.com/ballerina-platform/ballerina-lang/issues/22515

PR #23302, opened 2020-05-16
Implement `table<T> & readonly` and fix several issues with `xml<T> & readonly` 

MaryamZi:
## Purpose
Support `T & readonly` for table types.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23032

Fix the following issues with immutable XML elements.
- Fixes attribute setting failing on initialization for immutable elements.
- Fixes the type of an immutable element's attribute map not being marked immutable.
- Fixes children of an immutable element not being marked immutable.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23276

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -3443,6 +3443,12 @@ public void visit(BLangXMLElementLiteral bLangXMLElementLiteral) {
 
         resultType = checkXmlSubTypeLiteralCompatibility(bLangXMLElementLiteral.pos, symTable.xmlElementType,
                                                          this.expType);
+
+        if (!Symbols.isFlagOn(resultType.flags, Flags.READONLY)) {
```
KRVPerera, line 3447: Shouldn't this code be
```Java
if (Symbols.isFlagOn(resultType.flags, Flags.READONLY)) {
   markChildrenAsImmutable(bLangXMLElementLiteral);
}
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/XMLFactory.java
```diff
@@ -284,13 +286,9 @@ public static XMLValue createXMLElement(XMLQName startTagName, String defaultNsU
...
             return new XMLItem(new QName(defaultNsUri, startTagName.getLocalName(), prefix), readonly);
-        } else {
-            XMLItem xmlItem = new XMLItem(new QName(nsUri, startTagName.getLocalName(), prefix), readonly);
-            if (defaultNsUri != null && !defaultNsUri.isEmpty()) {
-                xmlItem.setAttribute(XMLConstants.XMLNS_ATTRIBUTE, null, null, defaultNsUri);
-            }
-            return xmlItem;
         }
+        return createXMLItemWithDefaultNSAttribute(new QName(nsUri, startTagName.getLocalName(), prefix), readonly,
+                                                   defaultNsUri);
```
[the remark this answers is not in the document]
MaryamZi: Yeah will change and check.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Would it make a difference? And I think the common case would be the current check evaluating to true (i.e., mutable result type).

KRVPerera: My opinion is that if lines are separated here, we lose the actual logic we need. Logic is not obvious anymore. Because these two lines are separated someone may put some lines in between in future and decide to return. At that time for some elements, we won't be calling `markChildrenAsImmutable`.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/XMLFactory.java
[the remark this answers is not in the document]
MaryamZi: Updated with 8e8807b.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: I've updated the logic with 8e8807b since this is a single invocation. But regarding adding lines in between, IMO the logic here is quite straightforward and it is reasonable to expect whoever doing the change to understand the area they are changing when doing the changes.  :) Especially given that we've adopted a similar approach with a significant number of functions.

PR #23372, opened 2020-05-19
Add optional field access and conditional expression parsing

### LINKED ISSUE #23267: Add conditional expression parsing
**Description:**
Refer: https://ballerina.io/ballerina-spec/spec.html#conditional-expr

**Suggested Labels:**
Component/Parser, Type/NewFeature

**Suggested Assignees**
@lochana-chathura

### LINKED ISSUE #23268: Add optional field access expression parsing
**Description:**
Refer: https://ballerina.io/ballerina-spec/spec.html#optional-field-access-expr

**Suggested Labels:**
Component/Parser, Type/NewFeature

**Suggested Assignees:**
@lochana-chathura

lochana-chathura:
Fixes #23268 
Fixes #23267  

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -9077,4 +9097,45 @@ private STNode parseIntersectionTypeDescriptor(STNode leftTypeDesc, ParserRuleCo
...
+     * @param lhsExpr Preceding expression of the question mark
+     * @return <code>conditional-expr</code>.
+     */
+    private STNode parseConditionalExpression(STNode lhsExpr) {
+        STNode questionMark = parseQuestionMark();
+        STNode middleExpr = parseExpression(OperatorPrecedence.ELVIS_CONDITIONAL, true, false);
+        STNode colon = parseColon();
+        STNode endExpr = parseExpression(OperatorPrecedence.ELVIS_CONDITIONAL, true, false);
+        return STNodeFactory.createConditionalExpressionNode(lhsExpr, questionMark, middleExpr, colon, endExpr);
+    }
```
lochana-chathura, line 9140: Assuming `qualified-identifier` gets the higher precedence `a ? b:c : d` passes as anticipated. 
But, If we consider simple `a ? b : c`, it cannot be parsed as `b:c` is parsed as qualified identifier. 
That's problem I'm facing now

SupunS: I see. May be we would have to special case that. i.e: in `expr1 ? expr2 : expr3` if colon is missing, we check whether `expr2` is a qualified identifier or not. If yes, we can split it back and treat it as the `expr2:expr3`. Not the cleanest way, but that would work i think.
WDYT?

lochana-chathura: Well, `expr2` is not always a qualified identifier, when we have this problem. 
For example in `a ? b : c+d`, we have a binary expression. 
So depend on the situation we will have to check their children or children of children or .... for a qualified identifier. 

Sometimes there cloud be more than one qualified identifier. Splitting and re-parsing is not possible in such complex cases.

lochana-chathura: Since we have an LL(k) parser I was thinking something like this.
When parsing a qualified identifier, we can check whether there's a spare colon ahead until lower precedence operator or end of expression token is reached. If there's no spare colon ahead, we will not parse the qualified identifier. 

We will have to use a flag to identify whether we are in a `conditional-expr` when doing the above thing. 

Another problem in this approach is we could have `mapping-constructors` when checking over the tokens ahead. So we have to make sure that we are looking for a spare colon which is not surrounded by `{`  `}`  tokens. 

WDYT?

SupunS: Ok.
I think we have to treat `a ? b : c` as an invalid code. Because according to https://github.com/ballerina-platform/ballerina-spec/issues/523, `b:c` is a var ref.

If someone want to make it working, they have to use parenthesis like:  `a ? (b) : c`.
That will simplify many things.

lochana-chathura: Ok great. Then no trouble at all.

lochana-chathura:
Took a pull from https://github.com/ballerina-platform/ballerina-lang/pull/23407
So please merge this after above PR if no further changes required.

lochana-chathura:
`AnnotationsTest > testRecoveryInAnnotValue` has following code.
```
@pkg1:annot1 {
    "name": "foo"
    description: "some function",
    "info": { "returns" "none", status: code}
```
With conditional-expr support new recovery is,
```
@pkg1:annot1 {
    "name": "foo"
 MISSING[?]    description MISSING[:]"some function",
    "info": { "returns"  MISSING[:]"none", status: code}
 MISSING[}]
```
That is the only change to previous test cases.
Added in https://github.com/ballerina-platform/ballerina-lang/pull/23372/commits/ad2e8fd4e5d2f4e31e1f31394e289efda23c40c6

lochana-chathura:
In the last commit https://github.com/ballerina-platform/ballerina-lang/pull/23372/commits/04e1cfaccbb17ff41f02974699f781f11ce76656 no changes to the test cases. Only wrong indentation and missing new line at the end in json files corrected.

PR #23478, opened 2020-05-25
Fix union types with finite types in transformer

manuranga:


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
```diff
@@ -466,10 +470,91 @@ public BLangNode transform(TypeDefinitionNode typeDefNode) {
...
+        // Generate a name for the anonymous object
+        String genName = anonymousModelHelper.getNextAnonymousTypeKey(diagnosticSource.pkgID);
+        IdentifierNode anonTypeGenName = createIdentifier(pos, genName);
+        bLTypeDef.setName(anonTypeGenName);
+        bLTypeDef.flagSet.add(Flag.PUBLIC);
+        bLTypeDef.flagSet.add(Flag.ANONYMOUS);
+
+        bLTypeDef.typeNode = toIndirect;
+        bLTypeDef.pos = pos;
+        currentCompilationUnit.addTopLevelNode(bLTypeDef);
```
rasika, line 550: If you wan't to add another top-level node can achieve with `additionalTopLevelNodes.add(bLTypeDef)`, these nodes will be added to current compilation unit in [here](https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java#L340) or [here](https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java#L294).

manuranga: Ya I tried that, but the I got the types added in wrong oder. I wanted to get them in exact order as the old parser.

I'll improve this to keep the topNode, but have new private method `addToTop`, that will ignore it if it started form an inner node.

rasika: How about adding them first? I noticed they are always on top of the list. How about mentioning the [index](https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html#add(int,%20E)) with `additionalTopLevelNodes.add(<index>, bLTypeDef)` and maintain an index?

manuranga: I'll fix this with my next pr

PR #23492, opened 2020-05-26
Include local repository by default

keizer619:

### DIFF
--- a/misc/maven-resolver/src/main/java/org/ballerinalang/maven/MavenResolver.java
+++ b/misc/maven-resolver/src/main/java/org/ballerinalang/maven/MavenResolver.java
@@ -83,8 +83,7 @@ public class MavenResolver {
         String localRepoPath = System.getProperty("user.home") + File.separator + ".m2"
                 + File.separator + "repository";
         if (new File(localRepoPath).exists()) {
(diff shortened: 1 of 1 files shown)


misc/maven-resolver/src/main/java/org/ballerinalang/maven/MavenResolver.java
```diff
@@ -78,6 +79,13 @@ public MavenResolver(String targetLocation) {
         repositories = new ArrayList<>();
         repositories.add(new RemoteRepository.Builder(
                 "central", "default", "https://repo.maven.apache.org/maven2/").build());
+
+        String localRepoPath = System.getProperty("user.home") + File.separator + ".m2"
+                + File.separator + "repository";
+        if (new File(localRepoPath).exists()) {
+            repositories.add(new RemoteRepository.Builder(
+                    "local", "default", "file:" + localRepoPath).build());
```
IrushiL, line 87: Can we join lines 86, 87 and 83, 84 unless it is intentional?

keizer619: 83, 84 cannot be merged as it exceeds the limit. 
86, 87 can be merged

PR #23504, opened 2020-05-26
Migrate ballerina/java as a langlib module

### LINKED ISSUE #23503: Migrate ballerina/java as a langlib module
**Description:**
This is required with jdk 11 migration due an issue with annotation preocssing. Our old style extern based on java class level annotations now have to be migrated to use the new ballerina based interop annotation. Hence, the solution is the move ballerina/java which is defined in the stdlib to langlib and make it a dependent module for all the others that require interop to work.

This also will make ballerina/java to compile and codegen using the ballerina based java annotation.

**Steps to reproduce:**

**Affected Versions:**

Kishanthan:
## Purpose
Fixes #23503


langlib/lang.__internal/src/main/ballerina/src/lang.__internal/stream-utils.bal
```diff
@@ -14,6 +14,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
+import ballerina/lang.java;
```
hasithaa, line 17: Not sure this is the correct thing to do. I think the java module should remain as `ballerina/java` @sameerajayasoma @Kishanthan  WDYT?

Kishanthan: It can still remain as `ballerina/java`. Since this module is moved inside the langlib dir, I made the module name as such. But it is not the final thing and we need to decide on that.

sameerajayasoma: Yes, `ballerina/lang.*` is reserved for langlib modules. Let's continue to use `ballerina/java`

PR #23518, opened 2020-05-26
Add Observability Improvement Changes to Master Branch

nadundesilva:
## Purpose
Add the Observability Improvement changes to master branch. This includes the following changes.
   * Add support to instrument using `@observe:Observable` annotation (defined in `ballerina/observe` module)
     * Add `object_name` and `function` tags for the Observations done in normal objects
     * FP Calls will not be observed even if the annotation had been added
   * Move instrumentation to function callee side (this is done to separately aggregate metrics for each invocation)
     * Refactor Observability instrumentation to do it by modifying the BIR (Previously this was done at the Java byte code generation phase)
     * Synchronously Calls are instrumented by adding start, stop and reportError calls accordingly
     * Async Calls are instrumented by generating a wrapper function which calls the actual function synchronously and adding the start, stop and reportError calls accordingly
     * Entry points (resource functions & main method) and workers are instrumented by adding start, stop and reportError calls in the function body itself
     * BIR Error Entries had been added in relevant places to handle and report errors accordingly had been added in relevant places to handle and report errors accordingly
   * Add tags for properly mapping metrics to the source code. These are passed in the startObservation Java Interop call as string constants.
     * `src.module` - The module ID (`<ORG_NAME>/<MODULE_NAME>:<MODULE_VERSION>`) of the function call (eg:- `nadundesilva/testmodule:0.1.0`
     * `src.position` - The position ID (`<SOURCE_CODE_FILE>:<START_LINE>:<START_COLUMN>`) of the function call (eg:- `testservice.bal:12:5`)
     * `src.entry_point.main` - The value `true` is added when the observation is done for the main method or tag is not present otherwise
     * `src.entry_point.resource` - The value `true` is added when the observation is done for a resource function or tag is not present otherwise
     * `src.worker` - The value `true` is added when the observation is done for a worker or tag is not present otherwise
     * `src.remote` - The value `true` is added when the observation is done for a remote function or tag is not present otherwise
   * Add `service` & `resource` tags to all the out of the box metrics and traces to identify proper resource for `caller-respond` calls
   * Remove Connector Name prefix from metric name and add a `connector_name` tag instead
   * Add new tags in `ballerina/http/Client` connector calls
     * `http.method`
     * `http.status_code_group`
     * `http.base_url`
   * Minor changes in observability in connectors; grpc, http, kafka, nats, rabbitmq
   * Remove the `http.status_code` tag and add `http.status_code_group` tag (eg:- `2xx` instead of `200`, `201`, `202`) instead. (Only for traces the actual status code had been added)
   * Improvements to Observability Instrumentation
     * Update tags to not create Tag objects every time a metric is created by keeping a `Map<String, Tag>` instead of `Map<String, String>` for tags.
   * Packaging AST, AST hash in the Jar file at Compile time
   * Re-enable Tracing Test case after fixing the test case which was using old Ballerina Table implementation


nadundesilva:
Sent PR #24060 for Perf improvements when Observability is disabled

nadundesilva:
#24152 to add a build time flag and Ballerina.toml section for packaging extension. This allows the user to reduce size if required. However, by default it is packaged.

misc/tracing-extensions/modules/ballerina-choreo-extension/spotbugs-exclude.xml
```diff
@@ -0,0 +1,27 @@
+<!--
+  ~ Copyright (c) 2018, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
warunalakshitha, line 2: Licence header should be 2020.

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/BuildCommand.java
```diff
@@ -413,8 +424,10 @@ public void execute() {
                 .addTask(new CreateBaloTask(), isSingleFileBuild)   // create the BALOs for modules (projects only)
                 .addTask(new CreateBirTask())   // create the bir
                 .addTask(new CopyNativeLibTask())    // copy the native libs(projects only)
+                .addTask(new CopyChoreoExtensionTask(), isChoreoExtensionSkipped)   // copy the Choreo extension
                 .addTask(new CreateJarTask(skipCopyLibsFromDist))   // create the jar
                 .addTask(new CopyResourcesTask(), isSingleFileBuild)
+                .addTask(new CopyObservabilitySymbolsTask(), isSingleFileBuild)
```
warunalakshitha, line 430: There are cases like running unit tests, where we skip copying libs from distribution using 'skipCopyLibsFromDist' flag. You might need to use 'skipCopyLibsFromDist' in CopyChoreoExtensionTask. Lets see  if unit tests failed.

nadundesilva: I think in all our test cases, this will be skipped. Because none of them has the build time flag or the `Ballerina.toml` table. (Except maybe in any test case which uses `ballerina new`). Will confirm once the other `lang-server` related test case issue had been fixed.

misc/tracing-extensions/modules/ballerina-choreo-extension/spotbugs-exclude.xml
nadundesilva: Fixed in 16163af

nadundesilva:
Created an issue for the concern related to Choreo Extension packaging in Docker Scenarios
https://github.com/ballerina-platform/ballerina-lang/issues/24188

nadundesilva:
Created an issue #24192 for reducing the number of async call wrapper functions generated as per the discussions.
@Kishanthan @warunalakshitha

nadundesilva:
Sent #24194 regarding the weird compilation issue related to `org.ballerinalang.langserver.extensions.VisibleEndpointVisitor.visit(packageNode)` failing when called at compile time (for collecting the AST json to be stored in the Jar)

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/BuildCommand.java
nadundesilva: Resolving as all the integrations are passing.

daneshk:
Reviewed gRPC related changes. LGTM

grainier:
Reviewed observe lib/runtime changes. LGTM.

Issue #23461, opened 2020-05-25
Optimize log api performance


MadhukaHarith92:
**Description:**
Currently, each module can be assigned its own log level. This is not a frequently used feature. However, when executing an API function (eg:printDebug) there is a validation that checks whether the log level (in this case DEBUG) is enabled. During this check, it is first checked whether the log level is enabled for that particular module. If not the global log level is considered. This check causes latency and impacts the performance of log-API.

**Affected Versions:**
Ballerina 1.2.4

PR #23530, opened 2020-05-27
Improve log-API performance

MadhukaHarith92:
Fixes #<Issue Number>
https://github.com/ballerina-platform/ballerina-lang/issues/23461

### DIFF
--- a/stdlib/log-api/src/main/java/org/ballerinalang/stdlib/log/AbstractLogFunction.java
+++ b/stdlib/log-api/src/main/java/org/ballerinalang/stdlib/log/AbstractLogFunction.java
@@ -56,11 +56,9 @@ public abstract class AbstractLogFunction {
      * @param strand strand
      * @param message  log message
      * @param logLevel log level
-     * @param pckg package
      * @param consumer log message consumer
      */
-    static void logMessage(Strand strand, Object message, BLogLevel logLevel, String pckg,
-            BiConsumer<String, String> consumer) {
+    static void logMessage(Strand strand, Object message, BLogLevel logLevel, BiConsumer<String, String> consumer) {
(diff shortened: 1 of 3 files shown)


stdlib/log-api/src/main/java/org/ballerinalang/stdlib/log/AbstractLogFunction.java
```diff
@@ -77,7 +75,15 @@ public String get() {
...
-        if (LOG_MANAGER.getPackageLogLevel(pckg).value() <= logLevel.value()) {
+        String pckg = ".";
+        BLogLevel enabledLogLevel;
+        if (LOG_MANAGER.isModuleLogLevelEnabled()) {
+            pckg = getPackagePath();
+            enabledLogLevel = LOG_MANAGER.getPackageLogLevel(pckg);
+        } else {
+            enabledLogLevel = LOG_MANAGER.getGlobalLogLevel();
+        }
+        if (enabledLogLevel.value() <= logLevel.value()) {
```
pubudu91, line 86: Doesn't this logic give the wrong module name in the log when a module specific log level is not set? For example, consider there's a log call in a function in a module `foo` like `log:printInfo("Hello World!");` and this module is used in my project's module `bar`. The printed log should look like something like,
```
2020-05-27 15:30:27,539 INFO  [pubudu/foo] - Hello World! 
```
But with this change, it would print it like,
```
2020-05-27 15:30:27,539 INFO  [] - Hello World! 
```

pubudu91: The module name printed with the log should always be the point of origin of that particular log call.

Issue #23525, opened 2020-05-26
[Bindgen Tool] Add a flag to generate public Java binding objects


IrushiL:
**Description:**
As of the current implementation all Ballerina binding objects generated from the Bindgen tool are public and therefore will be exposed to the outside from the module. However, the requirement sometimes could be to hide these binding objects and expose functionality build on top of it. 

This could be supported by taking into account the following points.
* Making the default access modifier private.
* Defining a flag that could be used to set the access modifier of generated bindings as public.
E.g. `[--public]`
* The above flag will be applicable for all the Java classes generated through a single execution. Therefore, it would be convenient to separate out private and public bindings and generate them separately or generate the bindings iteratively.
* If a binding is generated as public, all its dependent classes (ones that are used inside the object as parameters or return types) will be made public. This would be recursive until all the dependencies are turned public.
* Once a binding is already generated as public it cannot be changed to private. If this is required, the existing bindings will have to be deleted and regenerated from scratch. This could be implemented if a metadata file is maintained to keep details of bindings generated.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #23544, opened 2020-05-27
Improve array util functions and bindings in the Bindgen tool

IrushiL:
## Purpose
- Make the default access modifier of generated bindings private instead of public.
- Improve the array util functions and modify the bindings template accordingly.
- Refactor the output stream related code.

Fixes (partially): https://github.com/ballerina-platform/ballerina-lang/issues/23525

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/command/BindingsGenerator.java
+++ b/misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/command/BindingsGenerator.java
@@ -20,6 +20,7 @@ package org.ballerinalang.bindgen.command;
 import org.ballerinalang.bindgen.exceptions.BindgenException;
 import org.ballerinalang.bindgen.model.JClass;
 import org.ballerinalang.bindgen.model.JError;
+import org.ballerinalang.bindgen.utils.BindgenMvnResolver;
 import org.ballerinalang.toml.model.Library;
 import org.ballerinalang.toml.model.Manifest;
 import org.wso2.ballerinalang.util.TomlParserUtils;
@@ -57,7 +58,6 @@ import static org.ballerinalang.bindgen.utils.BindgenUtils.getUpdatedConstantsLi
 import static org.ballerinalang.bindgen.utils.BindgenUtils.isPublicClass;
 import static org.ballerinalang.bindgen.utils.BindgenUtils.notifyExistingDependencies;
 import static org.ballerinalang.bindgen.utils.BindgenUtils.writeOutputFile;
-import static org.ballerinalang.bindgen.utils.MvnResolverUtils.mavenResolver;
 
 /**
  * Class for generating Ballerina bindings for Java APIs.
@@ -88,7 +88,7 @@ public class BindingsGenerator {
     private static Set<JError> exceptionList = new HashSet<>();
     private static Map<String, String> failedClassGens = new HashMap<>();
 
-    public BindingsGenerator(PrintStream out, PrintStream err) {
+    BindingsGenerator(PrintStream out, PrintStream err) {
         this.outStream = out;
         this.errStream = err;
     }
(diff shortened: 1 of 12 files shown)


misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/command/BindingsGenerator.java
```diff
@@ -88,9 +88,9 @@
     private static Set<JError> exceptionList = new HashSet<>();
     private static Map<String, String> failedClassGens = new HashMap<>();
 
-    public BindingsGenerator(PrintStream stream) {
-        this.outStream = stream;
-        this.errStream = stream;
+    public BindingsGenerator(PrintStream out, PrintStream err) {
```
pramodya1994, line 91: Can't we make this `package-private` since it is only used in one place inside package.

misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/model/JClass.java
```diff
@@ -177,4 +178,8 @@ public String getShortClassName() {
     public String getPackageName() {
         return packageName;
     }
+
+    public void setAccessModifier(String accessModifier) {
```
pramodya1994, line 182: Are we using this method anywhere?

misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/utils/MvnResolverUtils.java
```diff
@@ -173,4 +173,8 @@ private static String getModuleName(Path projectRoot, String outputPath) {
         }
         return null;
     }
+
+    public static void setOutStream(PrintStream outStream) {
+        MvnResolverUtils.outStream = outStream;
```
keizer619, line 178: @IrushiL Its good to use setters for a non static property, since this is a static property can't we just pass outStream to relevant static methods as a param?

misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/utils/BindgenUtils.java
```diff
@@ -90,8 +90,8 @@
     private BindgenUtils() {
```
pramodya1994, line 90: Seems `getClassNamesInJar` and `getModuleName` methods is never used, Shall we remove it if there is no usage?

misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/utils/MvnResolverUtils.java
IrushiL: Refactored the code and made it a private variable instantiated through the constructor.

misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/command/BindingsGenerator.java
IrushiL: Done

misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/model/JClass.java
IrushiL: This is added to set the access modifier of a binding object that is to be generated. It is not used at the moment, but we intend on providing a flag to set it during the command execution. This will be implemented in the near future.

misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/utils/BindgenUtils.java
IrushiL: These 2 methods were kept as util methods for future use. But I removed them for now and will add them when it's actually being used.

IrushiL:
> I added a few comments

Addressed them.

PR #23607, opened 2020-05-29
Add version info to telemetry publish requests

madushajg:


misc/tracing-extensions/modules/ballerina-choreo-extension/src/main/proto/telemetry.proto
```diff
@@ -31,6 +32,7 @@ message TracesPublishRequest {
     string observabilityId = 2;  // ID received in the handshake
     string nodeId = 3;
     string appId = 4;
```
nadundesilva, line 34: We can now remove appId right ? In both metrics publish request and traces publish request

madushajg: Fixed with 6b26993ca503ae61076c580777c5bef728b0af56

PR #23640, opened 2020-06-01
Fix failure when iterating over empty array

riyafa:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
@@ -978,6 +978,9 @@ public class TypeChecker {
             if (sourceTupleType.getRestType() != null) {
                 tupleTypes.add(sourceTupleType.getRestType());
             }
+            if (tupleTypes.isEmpty()) {
+                return targetType.getState() == ArrayState.UNSEALED || targetType.getSize() == 0;
+            }
             sourceArrayType =
                     new BArrayType(new BUnionType(new ArrayList<>(tupleTypes), sourceTupleType.getTypeFlags()));
         }
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/jvm/ForeachArrayTests.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/jvm/ForeachArrayTests.java
@@ -218,4 +218,9 @@ public class ForeachArrayTests {
         Assert.assertEquals(returns.length, 1);
         Assert.assertEquals(returns[0].stringValue(), "0:d0 1: 2:d2 3: ");
     }
+
+    @Test
+    public void testEmptyArray() {
+        BRunUtil.invoke(program, "testEmptyArray");
+    }
 }
--- a/tests/jballerina-unit-test/src/test/resources/test-src/jvm/types.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/jvm/types.bal
@@ -157,6 +157,29 @@ function tupleTest() returns int {
    return 10;
 }
 
+public function testRestType() {
+    [int...] x = [1, 2];
+    any y = x;
+    assertEquality(y is string[], false);
+}
+
+public function testEmptyArrayType() {
+    var x = [];
+    any a = x;
+    assertEquality(a is int[2], false);
+    assertEquality(a is int[], true);
+
+    string[] sa = [];
+    any arr = sa;
+    assertEquality(arr is string[], true);
+    assertEquality(arr is int[], false);
+
+    int[0] ia = [];
+    any iarr = ia;
+    assertEquality(iarr is int[0], true);
+    assertEquality(iarr is int[], true);
+}
+
 function divideBy([int,int] d) returns [int, int] {
    //  int q = d[0] / d[1];
    // int r = d[0] % d[1];
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/jvm/TypesTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/jvm/TypesTest.java
@@ -219,6 +219,16 @@ public class TypesTest {
     }
 
     @Test
+    public void testRestType() {
+        BRunUtil.invoke(compileResult, "testRestType");
+    }
+
+    @Test
+    public void testEmptyArrayType() {
+        BRunUtil.invoke(compileResult, "testEmptyArrayType");
+    }
+
+    @Test
     public void testRecords() {
         BValue[] result = BRunUtil.invoke(compileResult, "recordsTest");
         Assert.assertEquals((result[0]).stringValue(), "JBallerina");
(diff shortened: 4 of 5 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -2193,6 +2193,9 @@ private static boolean checkFillerValue(BUnionType type) {
         if (type.isNullable()) {
             return true;
         }
+        if (type.getMemberTypes().isEmpty()) {
```
MaryamZi, line 2196: I'm not sure if this is the correct fix. IMO, we should only be creating a union type if there's at least two members. So ideally we wouldn't have a union with no members.

This is the approach we usually take in the compiler too. See https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java#L3831

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/jvm/ForeachArrayTests.java
```diff
@@ -218,4 +218,11 @@ public void testArrayWithNullElements() {
         Assert.assertEquals(returns.length, 1);
         Assert.assertEquals(returns[0].stringValue(), "0:d0 1: 2:d2 3: ");
     }
+
+    @Test
+    public void testEmptyArray() {
```
MaryamZi, line 223: Shall we move the assertion logic to the bal source?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -974,6 +974,9 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
             sourceArrayType = (BArrayType) sourceType;
         } else {
             BTupleType sourceTupleType = (BTupleType) sourceType;
+            if (sourceTupleType.getTupleTypes().isEmpty()) {
```
MaryamZi, line 977: Wouldn't this cause issues when the source type has just a rest-descriptor?

I think the following will produce an incorrect result with this change.

```ballerina
import ballerina/io;

public function main() {
    [int...] x = [1, 2];
    any y = x;
    io:println(y is string[]);
}
```

We should ideally have tests for the same too.

riyafa: Thanks fixed

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -978,6 +978,9 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
             if (sourceTupleType.getRestType() != null) {
                 tupleTypes.add(sourceTupleType.getRestType());
             }
+            if (tupleTypes.isEmpty()) {
```
MaryamZi, line 981: This will not work for something like

```ballerina
import ballerina/io;

public function main() {
    var x = [];

    any a = x;
    io:println(x is int[2]); // should be false
}
```

I think if `tupleTypes.isEmpty()` we should return true only if the target array state is `ArrayState.UNSEALED` or the target size is also 0.

riyafa: Fixed thanks.

tests/jballerina-unit-test/src/test/resources/test-src/jvm/types.bal
```diff
@@ -157,6 +157,31 @@ function tupleTest() returns int {
    return 10;
 }
 
+public function testRestType() {
+    [int...] x = [1, 2];
+    any y = x;
+    assertEquality(y is string[], false);
+    assertEquality(y is [int], true);
```
MaryamZi, line 164: Not due to the changes you introduced, but this should also fail. :( Seems like there are issues in the `checkIsTupleType` logic too. 

The `is` check looks for the belongs-to relationship

Should we fix that check in this PR too?

tests/jballerina-unit-test/src/test/resources/test-src/jvm/types.bal
```diff
@@ -157,6 +157,31 @@ function tupleTest() returns int {
...
+    string[] sa = [];
+    any arr = sa;
+    assertEquality(arr is string[], true);
+    assertEquality(arr is int[], false);
+
+    int[0] ia = [];
+    any iarr = ia;
+    assertEquality(iarr is int[0], true);
+    assertEquality(iarr is int[], true);
+
```
MaryamZi, line 182: Extra new line.

tests/jballerina-unit-test/src/test/resources/test-src/jvm/types.bal
riyafa: so only this needs to be true? `y is [int...]`

MaryamZi: And `y is int[]`

riyafa: both `y is int[]` and `y is int[0]` return false now. Are rest types considered arrays as well as tuples?

MaryamZi: `y is int[]` should be true, `y is int[0]` should be false.

Arrays and tuples both represent lists. The difference being in an array all the members are of the same type whereas for tuples they may be of different types. Arrays and tuples can be used interchangeably given that they fulfill "belongs-to" criteria.  

The types have to match and when it comes to the size,

`int[2]` is the same as `[int, int]` - list of two int members

`int[2]` is assignable to `[int, int, int...]` because `[int, int, int...]` represents a list of at least two int members, and `int[2]` is always that.

But you can't assign a `[int, int, int...]` to `int[2]` since it may have more than 2 members which will violate the inherent type of `int[2]`.

riyafa: Shall I create a different issue to fix this and remove the above test case?

MaryamZi: Okay I guess, since it's a different path. Shall we create one with high priority though, since this would lead to unexpected errors?

riyafa: Created issue https://github.com/ballerina-platform/ballerina-lang/issues/23676 . Please verify if it captures the requirements 🙂

MaryamZi: LGTM, thanks.

PR #23656, opened 2020-06-01
Fix system and xslt module errors

chamil321:

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/BallerinaErrors.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/BallerinaErrors.java
@@ -23,6 +23,7 @@ import org.ballerinalang.jvm.types.BPackage;
 import org.ballerinalang.jvm.types.BType;
 import org.ballerinalang.jvm.types.BTypeIdSet;
 import org.ballerinalang.jvm.types.BTypes;
+import org.ballerinalang.jvm.types.TypeConstants;
 import org.ballerinalang.jvm.util.exceptions.BLangExceptionHelper;
 import org.ballerinalang.jvm.util.exceptions.BallerinaErrorReasons;
 import org.ballerinalang.jvm.util.exceptions.RuntimeErrors;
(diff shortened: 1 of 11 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/BallerinaErrors.java
```diff
@@ -99,15 +100,32 @@ public static ErrorValue createError(BType type, BString message, BString detail
...
+
+    public static ErrorValue createDistinctError(String typeIdName, BPackage typeIdPkg, String message,
+                                                 ErrorValue cause) {
+        MapValueImpl<Object, Object> details = new MapValueImpl<>(BTypes.typeErrorDetail);
+        ErrorValue error = new ErrorValue(new BErrorType(TypeConstants.ERROR, BTypes.typeError.getPackage(),
+                                                         TypeChecker.getType(details)),
+                                          StringUtils.fromString(message), cause, details);
+        setTypeId(typeIdName, typeIdPkg, error);
+        return error;
+    }
```
rdhananjaya, line 122: Do we need another constructor taking:
- typeIdName
- pkgId
- message
- cause
- error detail type

Maybe we can add it when there is a requirement for it.

stdlib/system/src/main/java/org/ballerinalang/stdlib/system/utils/SystemUtils.java
```diff
@@ -49,41 +47,31 @@
...
 
     /**
-     * Returns error object for input reason.
-     * Error type is generic ballerina error type. This utility to construct error object from message.
+     * Returns error object  with message. Error type is generic ballerina error type. This utility to construct
+     * error object from message.
      *
-     * @param error Reason for creating the error object. If the reason is null, "UNKNOWN" sets by
-     *              default.
+     * @param typeId The string type ID of the particular error object. If the typeId is  null, "UNKNOWN" sets by
```
rdhananjaya, line 53: It's is required that the typeId to be same name as the defined type for distinct types.
For example `type A distinct error`, here typeId of `A` is `A`. 
Please refer the spec, in the spec the value we call typeId is referred to as `local id`.
[https://ballerina.io/ballerina-spec/spec.html#section_5.1.3](https://ballerina.io/ballerina-spec/spec.html#section_5.1.3)

rdhananjaya: The issue is we don't have a way to enforce this when we are creating error values programmatically in Java.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/BallerinaErrors.java
chamil321: +1 We'll add if it is required. Still, we did not find any

stdlib/system/src/main/java/org/ballerinalang/stdlib/system/utils/SystemUtils.java
chamil321: I think that param doc is misleading. I removed it. The original author must have meant to set the error message as "UNKNOWN" if the error reason(previously we had reason for errors) or the throwable is null. I fixed the logic to check only the throwable as the typeId cannot be null and throughout the module, a null has NOT been passed as a parameter to the typeId.

PR #23658, opened 2020-06-01
Send app secret to Telemetry service along with telemetry data

madushajg:


misc/tracing-extensions/modules/ballerina-choreo-extension/src/main/proto/telemetry.proto
```diff
@@ -17,6 +17,7 @@ message MetricsPublishRequest {
     string observabilityId = 2;  // ID received in the handshake
     string version = 3;
     string nodeId = 4;
+    string appSecret = 5;
```
nadundesilva, line 20: Let's stick with "projectSecret" in all the places

madushajg: Fixed with 1afd8a94da2693806aad2bfb2d45a086a22b5169

PR #23670, opened 2020-06-02
Check if scope is testOnly when copying libs from distribution

### LINKED ISSUE #19271: [Packerina Spec] Dependency jars getting packed to balo

suganyasuven:
Fixes #19271 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


suganyasuven:
> Do we have test to cover this scenari

PR #23703, opened 2020-06-03
Fix unnecessary <any|error> type passing for map store access. 

### LINKED ISSUE #23551: Bad, sad error at runtime with updates on an object union via an invocation of a function pointer
**Description:**
$title.

**Steps to reproduce:**
The following works as expected.
```ballerina
import ballerina/io;

type Foo object {
    int i = 1;
};

type Bar object {
    int i = 1;
};

public function main() {
    Foo f = new;
    Foo|Bar fb = f;
    io:println(fb.i);

    fb.i = 2;
    io:println(fb.i);
}
```
```cmd
$ ballerina run test.bal 
Compiling source
        test.bal
Running executables

1
2
```

But when we attempt updating the union via a function pointer, it results in a bad, sad error and even `2` is not printed!

```ballerina
import ballerina/io;

type Foo object {
    int i = 1;
};

type Bar object {
    int i = 1;
};

public function main() {
    Foo f = new;
    Foo|Bar fb = f;
    io:println(fb.i);

    fb.i = 2;
    io:println(fb.i);

    // Addition
    var fn = function () {
        fb.i = 3;
    };
    fn();
    io:println(fb.i);
}
```

```cmd
$ ballerina run test.bal 
Compiling source
        test.bal
Running executables

1
ballerina: Oh no, something really went wrong.
```

ballerina-internal.log
```
[2020-05-28 00:24:31,873] SEVERE {b7a.log.crash} - null 
java.lang.IncompatibleClassChangeError
	at org.ballerinalang.jvm.MapUtils.handleMapStore(MapUtils.java:99)
	at test.main(test.bal:16)
	at ___init.$lambda$main$(.)
	at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:482)
	at org.ballerinalang.jvm.scheduling.Scheduler.run(Scheduler.java:263)
	at org.ballerinalang.jvm.scheduling.Scheduler.runSafely(Scheduler.java:231)
 
```

**Affected Versions:**
1.2.3

HindujaB:
Fixes #23551 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


HindujaB:
> Lets add more test scenarios as discussed during the standup.

Modified the test case in the last commit

tests/jballerina-unit-test/src/test/resources/test-src/types/uniontypes/union_types_basic.bal
```diff
@@ -166,3 +166,70 @@ function testUnionLhsWithDiscriminatedFloatDecimalLiterals() returns [(float|dec
...
+function assertEquality(any|error expected, any|error actual) {
+    if expected is anydata && actual is anydata && expected == actual {
+        return;
+    }
+    if expected === actual {
+        return;
+    }
+    panic error(ASSERTION_ERR_REASON,
+                message = "expected '" + expected.toString() + "', found '" + actual.toString () + "'");
+}
```
Kishanthan, line 235: Missing EOF line here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
```diff
@@ -1122,28 +1118,6 @@ private void updateClosureVars(BLangSimpleVarRef varRefExpr, BVarSymbol mapSymbo
         result = rewriteExpr(desugar.addConversionExprIfRequired(accessExpr, varRefExpr.type));
```
MaryamZi, line 1118: Thinking out loud, the closure map is a `map<any|error>` right?

Looking at the comments and the logic, it seems like there' some legacy code (based on simple basic types vs ref types in the BVM) which may not be valid now.

So basically what I believe we've tried to do here is if it is a simple assignment like `x = 1;` where the ref we're working is on the LHS, we use `any|error` as the type of index-based access for simple basic types with an eventual cast on the RHS, whereas for ref types we're using the original types (which were compatible with the ref type impl). IIRC, we used to need this in the BVM based implementation (boxing), I'm not sure if we still need to do it in the front-end.

Where we have the closure map `map<any|error> $innerMap$1`.

```ballerina
x = 1;
```
becomes
```ballerina
$innerMap$1["x"] = <any> 1;
```

Then we have the access on the RHS, where in L1118 we've added a cast to the expected type. Basically, we access the relevant value from the `map<any|error>` and cast to the expected type.
```ballerina
int i = x;
```
becomes
```ballerina
int i = <int> $innerMap$1["x"] ;
```

So coming back to the problem we're trying to solve,

```ballerina
    var fn = function () {
        fb.i = 3;
    };
```
it seems like even for this, it goes in the lhsVar path, without the cast. We seem to be returning `$innerMap$1["fb"]` instead of `<Foo|Bar> $innerMap$1["fb"]`.

IMO we can do one of the following.

1. Semantically, a map (`map<T>`) access on the RHS returns `T?` so if someone writes this code in Ballerina, the type of access expr will be `T` on the LHS, but `T?` on the RHS. With the current changes, we're changing this for both LHS and RHS to `T`, but I guess i's OK since we know that value is always present in the map. If we are going ahead with this though let's check
a) if this works as expected for simple basic types (int, byte, float, etc.)
b) if we need the `desugar.addConversionExprIfRequired(accessExpr, varRefExpr.type)` anymore since the access expression's type is now varRefExpr.type

or

2. we need to figure out if we can add the cast on the LHS side also if the current var ref is part of a compound expression.

tests/jballerina-unit-test/src/test/resources/test-src/types/uniontypes/union_types_basic.bal
```diff
@@ -166,3 +166,70 @@ function testUnionLhsWithDiscriminatedFloatDecimalLiterals() returns [(float|dec
...
+        self.lastName = lastName;
+    }
+};
+
+function testUnionTypeWithFunctionPointerAccess() {
+    Engineer engineer = new Engineer(20, "John", "Doe");
+    Employee employee = new Engineer(20, "Jane", "Doe");
+
+    Engineer|Manager person1 = engineer;
+    Employee|Manager person2 = employee;
```
Kishanthan, line 212: Are we not testing Manager type here?

HindujaB: It works same as the engineer type. We can remove it and use only the 'employee' and 'engineer' types. Is it the concern?

HindujaB: Removed the unnecessary object type in the test case

PR #23704, opened 2020-06-03
Fix encoding and math module errors

Bhashinee:
## Purpose
$subject.

### DIFF
--- a/stdlib/encoding/src/main/java/org/ballerinalang/stdlib/encoding/nativeimpl/Decode.java
+++ b/stdlib/encoding/src/main/java/org/ballerinalang/stdlib/encoding/nativeimpl/Decode.java
@@ -25,6 +25,8 @@ import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.util.Base64;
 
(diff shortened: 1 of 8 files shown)


stdlib/encoding/src/main/java/org/ballerinalang/stdlib/encoding/nativeimpl/Decode.java
```diff
@@ -37,15 +39,16 @@ public static Object decodeBase64Url(String input) {
             byte[] output = Base64.getUrlDecoder().decode(input);
             return BValueCreator.createArrayValue(output);
         } catch (IllegalArgumentException e) {
-            return EncodingUtil.createError("Input is not a valid Base64 URL encoded value");
+            return EncodingUtil.createError("Input is not a valid Base64 URL encoded value", DECODING_ERROR);
```
chamil321, line 42: @Bhashinee I think we need to define this distinct error DECODING_ERROR. the only it will get generated properly at ballerina level with typeIds. 

It's is required that the typeId to be same name as the defined type for distinct types.
For example type A distinct error, here typeId of A is A.

Issue #23376, opened 2020-05-19
Ballerina build fail for large modules


Rajith90:
**Description:**
I have attached a ballerina module [gen.zip](https://github.com/ballerina-platform/ballerina-lang/files/4651488/gen.zip). When I try to build it with command [1], it gives me the error [2].
Seems like there is an limitation when generating java byte code from ballerina, when the ballerina module is too large.

Please find the attached libs required 
[lib.zip](https://github.com/ballerina-platform/ballerina-lang/files/4681410/lib.zip)
. Please edit the ballerina.toml with correct paths

[1] - 
```
ballerina build --experimental EndpointReferenceProject
```

[2] - 
```
Compiling source
	wso2/EndpointReferenceProject:3.1.0

Creating balos
	target/balo/EndpointReferenceProject-2019r3-java8-3.1.0.balo
error: wso2:EndpointReferenceProject:EndpointReferenceProject:1:1: method is too large: 'wso2/EndpointReferenceProject:3.1.0.<init>'
error: error: wso2:EndpointReferenceProject:EndpointReferenceProject:1:1: method is too large: 'wso2/EndpointReferenceProject:3.1.0.<init>'
```
**Steps to reproduce:**

**Affected Versions:**
1.1.5

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Kishanthan:
Can you also share the dependencies that is required to build the project here?

Rajith90:
Hi @Kishanthan ,
I have attached the lib zip file with necessary jars and balos. Please edit the ballerina.toml file with correct paths.

riyafa:
@Rajith90 What is the version of the ballerina that is used in your project?

riyafa:
We have done some major improvements in later versions of ballerina. If your project is currently on version 1.1.x version of Ballerina can you please update to 1.2.x version of ballerina and update if the issue persists. Also please upload the modified project files for the syntax changes.

riyafa:
This particular issue has been fixed in https://github.com/ballerina-platform/ballerina-lang/pull/23919 . Further improvements are planned through bir optimization. Some relevant issues are https://github.com/ballerina-platform/ballerina-lang/issues/23914 and  https://github.com/ballerina-platform/ballerina-lang/issues/24619

menakaj:
Hi @Kishanthan, @riyafa The same issue is observed for the main method as well.

PR #23717, opened 2020-06-04
Move assignments with init or record to a separate method

riyafa:
## Purpose

Related issue: https://github.com/ballerina-platform/ballerina-lang/issues/23376

## Approach
To prevent the method too large issue, this fix moves the assignment statements with init or record expressions to a separate method, thus making the init method smaller

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


riyafa:
I thought it would be done manually. But I can try to create a module that will make the init function bigger which fails during build in the previous versions and passes in this?

Kishanthan:
> I thought it would be done manually. But I can try to create a module that will make the init function bigger which fails during build in the previous versions and passes in this?

Yes, lets add a test case that verifies this fix.

riyafa:
The difficult part of adding the test is getting the init method to be too large within the 50 statements limit. One thing we can do to write a test is to make this 50 statements limit configurable. @Kishanthan WDYT?

Kishanthan:
> The difficult part of adding the test is getting the init method to be too large within the 50 statements limit. One thing we can do to write a test is to make this 50 statements limit configurable. @Kishanthan WDYT?

Yes, that is an option.
> 
> We cannot use the project from #23376 directly because it requires so many dependencies and hence will make this repo larger if we use it.
> 
> Between this PR is for the 1.1.x branch while the plan is to work on a proper implementation for the master branch

But we can add this fix to master as well and work on the optimization improvement. Shall we push this fix to both 1.2.x and master branches?

tests/jballerina-integration-test/src/test/resources/testng.xml
```diff
@@ -294,12 +294,12 @@
...
-            <class name="org.ballerinalang.test.packaging.ModulePushTestCase"/>
+<!--            <class name="org.ballerinalang.test.packaging.ModulePushTestCase"/>-->
             <!--            <class name="org.ballerinalang.test.packaging.PackagingNegativeTestCase"/>-->
-            <class name="org.ballerinalang.test.packaging.PackagingTestCase"/>
-            <class name="org.ballerinalang.test.packaging.LockFileTestCase"/>
+<!--            <class name="org.ballerinalang.test.packaging.PackagingTestCase"/>-->
+<!--            <class name="org.ballerinalang.test.packaging.LockFileTestCase"/>-->
             <!--            <class name="org.ballerinalang.test.packaging.NativePackagingTestCase"/>-->
-            <class name="org.ballerinalang.test.packaging.PathDependencyTestCase"/>
+<!--            <class name="org.ballerinalang.test.packaging.PathDependencyTestCase"/>-->
```
Kishanthan, line 302: Are these packaging tests already failing in this 1.1.x branch? @hevayo @keizer619

riyafa:
To see if changing the size affects the performance. Did a small perf test with the  project in issue #23376 and got the following results:
With method size 50:     53.75s    
With method size 2:        49.85s

Tested with another file that has large records initialized and results are as follows:
With method size 50:     10.11s
With method size 2:        10.01s

Runtime for the same:
With method size 50:     347ms
With method size 2:        354ms

tests/jballerina-integration-test/src/test/resources/testng.xml
riyafa: These tests fail when pulling or pushing to central.

Kishanthan: @keizer619 @hevayo , is this expected with 1.1.x branch? Shall we go ahead with disabling the tests?

Kishanthan: This is ok as per @hevayo 
"tests are failing probably due to lets encrypt migration of the staging server. And prior 1.2 clients are not compatible with that."

PR #23732, opened 2020-06-04
Fix expressions are supported as function-call lhs

### LINKED ISSUE #23632: Expressions are supported for function-call lhs
**Description:**
Below code correctly parses by the new parser
```ballerina
int a = f(1)(2);
```

lochana-chathura:
## Purpose
$subject.

Fixes #23632 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


lochana-chathura:
over loaded the existing `seekMatchInExpressionRhs()` method by adding `allowFuncCall`  argument to method signature.
No change to previous syntax validation and recovery tests. :) New 3 recovery test cases added.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -1765,82 +1768,97 @@ private Result seekInAccessExpression(ParserRuleContext currentCtx, int lookahea
...
         switch (parentCtx) {
             case ARG_LIST:
-                next = new ParserRuleContext[] { ParserRuleContext.BINARY_OPERATOR, ParserRuleContext.DOT,
+                alternatives = new ParserRuleContext[] { ParserRuleContext.BINARY_OPERATOR, ParserRuleContext.DOT,
                         ParserRuleContext.ANNOT_CHAINING_TOKEN, ParserRuleContext.OPTIONAL_CHAINING_TOKEN,
                         ParserRuleContext.CONDITIONAL_EXPRESSION, ParserRuleContext.XML_NAVIGATE_EXPR,
                         ParserRuleContext.MEMBER_ACCESS_KEY_EXPR, ParserRuleContext.COMMA,
                         ParserRuleContext.ARG_LIST_START, ParserRuleContext.ARG_LIST_END };
-                return seekInAlternativesPaths(lookahead, currentDepth, currentMatches, next, isEntryPoint);
+                return seekInAlternativesPaths(lookahead, currentDepth, currentMatches, alternatives, isEntryPoint);
```
lochana-chathura, line 1790: Under `ARG_LIST` context, `ARG_LIST_START` is always there as an alternative.
In other contexts, `ARG_LIST_START` is added as an alternative only if `allowFuncCall` is `true`.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -4299,6 +4299,15 @@ private boolean isValidExprRhsStart(SyntaxKind tokenKind) {
         }
     }
 
+    private boolean isInvalidFunctionCallStart(SyntaxKind tokenKind, STNode lhsExpr) {
```
SupunS, line 4302: Shouldn't this be merged with `isValidExprRhsStart`?
Because there are other places where we use `isValidExprRhsStart` for the same purpose, and those should behave the same way right?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -1875,16 +1893,24 @@ private Result seekMatchInExpressionRhs(int lookahead, int currentDepth, int cur
...
+        if (allowFuncCall) {
+            alternatives = addNewAlternative(alternatives, ParserRuleContext.ARG_LIST_START);
+        }
         return seekInAlternativesPaths(lookahead, currentDepth, currentMatches, alternatives, isEntryPoint);
     }
 
+    private ParserRuleContext[] addNewAlternative(ParserRuleContext[] alternatives, ParserRuleContext newAlternative) {
+        List<ParserRuleContext> alternativesList = new ArrayList<>(Arrays.asList(alternatives));
+        alternativesList.add(newAlternative);
+        return alternativesList.toArray(alternatives);
```
SupunS, line 1911: This operation is a bit expensive. Can we avoid this?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
SupunS: Why is that needed?
What if is write `expr()` inside func-args?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -3522,6 +3548,22 @@ private ParserRuleContext getNextRuleForExpr() {
...
+        if (parentCtx == ParserRuleContext.CONSTANT_EXPRESSION) {
+            endContext();
+            return getNextRuleForConstExpr();
+        }
+        return getNextRuleForExprWithFuncCallEnabled();
+    }
+
+    private ParserRuleContext getNextRuleForExprWithFuncCallEnabled() {
+        // Here we start a new context so that expr-rhs will allow function calls
+        startContext(ParserRuleContext.FUNCTION_CALL_START);
```
SupunS, line 3563: Can we do it without starting a context? because it is also an expensive operation. Should use that only if theres no other way.

What if we give an alias, say "VARIABLE_REF_RHS", and that will call the `seekMatchInExpressionRhs` with the flag on/off

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -1770,82 +1773,97 @@ private Result seekInAccessExpression(ParserRuleContext currentCtx, int lookahea
...
                         ParserRuleContext.CONDITIONAL_EXPRESSION, ParserRuleContext.XML_NAVIGATE_EXPR,
-                        ParserRuleContext.MEMBER_ACCESS_KEY_EXPR, ParserRuleContext.QUERY_PIPELINE_RHS,
-                        ParserRuleContext.ARG_LIST_START };
-                return seekInAlternativesPaths(lookahead, currentDepth, currentMatches, next, isEntryPoint);
+                        ParserRuleContext.MEMBER_ACCESS_KEY_EXPR, ParserRuleContext.QUERY_PIPELINE_RHS };
+                break;
             default:
                 if (isParameter(parentCtx)) {
-                    next = new ParserRuleContext[] { ParserRuleContext.CLOSE_PARENTHESIS,
+                    alternatives = new ParserRuleContext[] { ParserRuleContext.CLOSE_PARENTHESIS,
```
SupunS, line 1851: Will the above logic be simplified, if we take the common things out, similar to the second switch-case?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
lochana-chathura: Used `System.arraycopy()` instead of arrayLists. HBT?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
lochana-chathura: Common alternatives are located in the middle of array. No of start and end alternatives are different for each ctx. So are we going to add the common ones at the end?

lochana-chathura: btw `MEMBER_ACCESS_KEY_EXPR` is missing as an alternative under `COMPUTED_FIELD_NAME` ctx. I think it's a mistake

SupunS:
Changes looks OK. 
Let's resolve the conflicts.

PR #23754, opened 2020-06-05
Add cause to the BError constructor

chamil321:
Fixes #23634

### DIFF
--- a/bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BError.java
+++ b/bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BError.java
@@ -89,4 +89,24 @@ public class BError implements BRefType {
     public BError getCause() {
         return cause;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (obj instanceof  BError) {
+            BError that = (BError) obj;
+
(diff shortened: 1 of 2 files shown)


bvm/ballerina-core/src/main/java/org/ballerinalang/model/values/BError.java
```diff
@@ -84,4 +85,8 @@ public Object value() {
     public boolean isFrozen() {
         return true;
     }
+
+    public BError getCause() {
+        return cause;
+    }
```
rdhananjaya, line 91: Shall we override the equals method. Without it an unit test case is failing.
Following worked in my local machine.
```java
    @Override
    public boolean equals(Object obj) {
        if (this == obj) {
            return true;
        }

        if (obj instanceof  BError) {
            BError that = (BError) obj;

            boolean isCauseSame = false;
            if (this.cause != null) {
                isCauseSame = this.cause.equals(that.cause);
            } else if (this.cause == null && that.cause == null) {
                isCauseSame = true;
            }
            return this.message.equals(that.message) && this.details.equals(that.details) && isCauseSame;
        }
        return false;
    }
```

chamil321: Thanks, Fixed via edc2257a72440de5a34243906a648875a446eef5

PR #23765, opened 2020-06-07
Update file module to new error syntax

Bhashinee:
## Purpose
$subject.

### DIFF
--- a/stdlib/file/src/main/ballerina/src/file/file_errors.bal
+++ b/stdlib/file/src/main/ballerina/src/file/file_errors.bal
@@ -14,15 +14,6 @@
 // specific language governing permissions and limitations
 // under the License.
 
-# Record type to hold the details of an error.
-#
-# + message - Specific error message of the error.
-# + cause - Any other error, which causes this error.
-public type Detail record {
-    string message;
-    error cause?;
-};
-
 # Represents an error that occurs when a file system operation is denied due to invalidity.
 public type InvalidOperationError distinct error;
 
--- a/stdlib/file/src/main/ballerina/src/file/file_errors.bal
+++ b/stdlib/file/src/main/ballerina/src/file/file_errors.bal
@@ -23,26 +23,17 @@ public type Detail record {
     error cause?;
 };
 
-# Identifies invalid operation error.
-public const INVALID_OPERATION_ERROR = "{ballerina/file}InvalidOperationError";
-
 # Represents an error that occurs when a file system operation is denied due to invalidity.
-public type InvalidOperationError error<INVALID_OPERATION_ERROR, Detail>;
+public type InvalidOperationError distinct error;
 
-# Identifies permission error.
-public const PERMISSION_ERROR = "{ballerina/file}PermissionError";
 # Represents an error that occurs when a file system operation is denied, due to the absence of file permission.
-public type PermissionError error<PERMISSION_ERROR, Detail>;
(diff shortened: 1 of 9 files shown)


stdlib/file/src/main/ballerina/src/file/file_errors.bal
```diff
@@ -23,26 +23,17 @@ public type Detail record {
     error cause?;
```
chamil321, line 23: Do we need this record?

stdlib/file/src/main/java/org/ballerinalang/stdlib/file/utils/FileUtils.java
```diff
@@ -71,7 +71,7 @@ public static ErrorValue getBallerinaError(String error, Throwable ex) {
      * @return Ballerina error object.
      */
     public static ErrorValue getBallerinaError(String error, String details) {
-        return BallerinaErrors.createError(error, populateSystemErrorRecord(details));
+        return BallerinaErrors.createDistinctError(error, FILE_PACKAGE_ID, error, populateSystemErrorRecord(details));
```
chamil321, line 74: I think we can get rid of `populateSystemErrorRecord()` method as well. We only need to pass a record to createDistinctError method only if the detail record has are fields other than `message` and `cause`. So since we pass the `message` field to the `populateSystemErrorRecord()`, we can simply pass only the message.

```suggestion
        return BallerinaErrors.createDistinctError(error, FILE_PACKAGE_ID, details != null ? details : UNKNOWN_MESSAGE);
```

chamil321: Also, shall we change the param name from `details` to `message`. And the docs as well

stdlib/file/src/test/java/org/ballerinalang/stdlib/file/FileTest.java
```diff
@@ -121,7 +121,7 @@ public void testFileRenameWithInvalidPath() throws IOException {
             BValue[] returns = BRunUtil.invoke(compileResult, "testRename", args);
             assertTrue(returns[0] instanceof BError);
             BError error = (BError) returns[0];
-            assertEquals(error.getReason(), FileConstants.INVALID_OPERATION_ERROR);
+            assertEquals(error.getMessage(), FileConstants.INVALID_OPERATION_ERROR);
```
chamil321, line 124: Ideally FileConstants.INVALID_OPERATION_ERROR should not be retrieved as the error message. The message should be `File already exists in the new path `

stdlib/ldap/src/main/java/org/ballerinalang/stdlib/ldap/util/LdapUtils.java
```diff
@@ -36,6 +36,8 @@
 import javax.naming.directory.SearchControls;
 import javax.naming.directory.SearchResult;
 
+import static org.ballerinalang.stdlib.ldap.LdapConstants.LDAP_PACKAGE_ID;
```
kalaiyarasiganeshalingam, line 39: Can we remove the static import?

chamil321: These static imports do not damage the code and it improves the readability. Hence resolving

Issue #23548, opened 2020-05-27
Issue in the untainted annotation


kalaiyarasiganeshalingam:
**Description:**
Untainted annotation supports any string value instead of  **untainted**. See below images:

**Image without annotation**

<img width="677" alt="Screen Shot 2020-05-27 at 4 05 27 PM" src="https://user-images.githubusercontent.com/12771261/83033431-b9048680-a054-11ea-982b-2e43ac0d44ab.png">

**Image with annotation**

<img width="739" alt="Screen Shot 2020-05-27 at 4 09 07 PM" src="https://user-images.githubusercontent.com/12771261/83034543-01707400-a056-11ea-9753-9177431c5c43.png">

Issue #23793, opened 2020-06-08
Error when using `xml:getContent()` with valid-typed XML variables holding values accessed from an XML value


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/io;
import ballerina/lang.'xml;

public function main() {
    xml x = xml `<foo>Foo</foo>`;

    xml v = x.<foo>/*;

    io:println(v is 'xml:Text);
    io:println(v);

    'xml:Text txtVal = <'xml:Text> v;
    string strVal = txtVal.getContent();
    io:println(strVal);
}
```

```cmd
$ ballerina run qwe.bal 
Compiling source
        qwe.bal
Running executables

true
Foo
error: lang.xml function 'getContent' only accept xml type 'text|processing instruction|comment' 
        at ballerina.lang_xml.0_8_0:getContent(xml.bal:141)
           qwe:main(qwe.bal:13)
```

**Affected Versions:**
1.2.4
Current master

rdhananjaya:
This general issue is fixed with #29047

xml:getContent does not support xml:Text now, but when we modify the given code to be a comment as bellow it works now.

```ballerina
import ballerina/io;
import ballerina/lang.'xml;

public function main() {
    xml x = xml `<foo><!-- Foo --></foo>`;

    xml v = x.<foo>/*;

    io:println(v is 'xml:Text);
    io:println(v);

    'xml:Comment txtVal = <'xml:Comment> v;
    string strVal = txtVal.getContent();
    io:println(strVal);
```

PR #23829, opened 2020-06-09
Add annotation validation to typeconversion expr

rdhananjaya:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23548

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -3224,10 +3229,113 @@ public void visit(BLangUnaryExpr unaryExpr) {
         resultType = types.checkType(unaryExpr, actualType, expType);
     }
 
+    public void visit(BLangAnnotationAttachment annAttachmentNode) {
```
MaryamZi, line 3232: I'm not sure if this belongs here. Why are we moving this from SemanticAnalyzer, given that `BLangAnnotationAttachment` itself is not an expression, and only the `expr` part is?

We already seem to be doing the relevant semantic/constant analyzer based analysis in SemanticAnalyzer, and only seem to be calling into TypeChecker for the expression https://github.com/ballerina-platform/ballerina-lang/pull/23829/files#diff-7123076514bbb67aa85076709d29ee0fL2880.

rdhananjaya: What I want to do is execute annotation validation logic from `typecheker` since that's where we see the annotation on a `typeconvertion` expression.

If not moving to typechecker we will have to move that logic to its own class, but I don't see any compelling advantage of moving to a separate class that moving to typecheker.

rdhananjaya: It seems we do call back to `SemanticAnalyzer` from `typechecker`.

MaryamZi: Yeah, there was a similar requirement with streaming and it was decided to have this circular dependency for the time-being instead of moving the logic.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6478,6 +6485,19 @@ private void logUndefinedSymbolError(DiagnosticPos pos, String name) {
         }
     }
 
+    public void checkAnnot(BLangAnnotationAttachment annAttachmentNode, SymbolEnv env) {
```
MaryamZi, line 6488: Unused?

misc/testerina/modules/report-tools/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@ballerina/testerina-report",
-    "version": "2.0.0-Preview1-M3",
+    "version": "2.0.0-SNAPSHOT",
```
MaryamZi, line 3: Do we need this change?

MaryamZi:
Is there a PR for 1.2.x? IIRC, we decided to include these changes in the patch versions too right?

tests/jballerina-integration-test/src/test/resources/http2/src/http2services/10_http_2.0_forwarded_header.bal
```diff
@@ -37,6 +37,6 @@ service forwardedBackend on new http:Listener(9106, {httpVersion: "2.0"}) {
         http:Response response = new();
         response.setHeader("forwarded", header);
         response.setPayload("forward is working");
-        var resultSentToInitClient = caller->respond(<@untianted> response);
+        var resultSentToInitClient = caller->respond(<@untaianted> response);
```
MaryamZi, line 40: Shouldn't this fail?

```suggestion
        var resultSentToInitClient = caller->respond(<@untainted> response);
```

misc/testerina/modules/report-tools/package.json
rdhananjaya: Looks like a merge issue, ill remove and go though the changed file list again

rdhananjaya:
> Is there a PR for 1.2.x? IIRC, we decided to include these changes in the patch versions too right?

This broke a few invalid use cases in stdlib, who we include this in patch release as this is a breaking change?

MaryamZi:
> > Is there a PR for 1.2.x? IIRC, we decided to include these changes in the patch versions too right?
> 
> This broke a few invalid use cases in stdlib, who we include this in patch release as this is a breaking change?

I think we considered this and decided to do so even though it is a breaking change because the behaviour is incorrect right? Basically using an annotation that is not `untainted` also untaints the value?

```ballerina
public function main(int i) {
    foo(<@taint> i);
}

function foo(@untainted int i) {
    
}
``` 

Looping in @hasithaa too.

langlib/lang.annotations/src/main/ballerina/src/lang.annotations/annotations.bal
```diff
@@ -15,7 +15,7 @@
 // under the License.
 
 # Denote that the return value is tainted.
-public const annotation tainted on parameter, return, source listener, source var;
+public const annotation tainted on parameter, return, source listener, source var, source type;
```
MaryamZi, line 18: This will allow using `<@tainted>` too right? I hope that's intentional? Do we have tests for that flow?

Something like
```ballerina
public function main() {
    int x = 1;
    foo(<@tainted> x);
}

function foo(@untainted int i) {
    
}
```

rdhananjaya: We already had logic [1] in taint analyzer to untaint for `<@untaint>`, looks like we don't have any test for it, I'll add.

[1] - https://github.com/ballerina-platform/ballerina-lang/issues/23793

rdhananjaya: Test case added

PR #23896, opened 2020-06-10
Fix webSocket test errrors

kalaiyarasiganeshalingam:
## Purpose
$Subject

Fixes #23634

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/http/src/main/java/org/ballerinalang/net/http/websocket/WebSocketUtil.java
```diff
@@ -575,6 +576,28 @@ public static void countDownForHandshake(ObjectValue webSocketClient) {
...
+        } else {
+            exception = new WebSocketException(message, cause);
+        }
+        BallerinaErrors.setTypeId(errorCode, WebSocketConstants.PROTOCOL_HTTP_PKG_ID, exception);
+        return exception;
+    }
+
+    public static void setNotifyFailure(String msg, NonBlockingCallback callback) {
+        WebSocketException exception = new WebSocketException(msg);
+        BallerinaErrors.setTypeId(WebSocketConstants.ErrorCode.WsInvalidHandshakeError.toString(),
```
rdhananjaya, line 596: Unless we trap this error in Ballerina and convert to a error value, I don't think we need to do this here.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/BallerinaErrors.java
```diff
@@ -121,7 +121,7 @@ public static ErrorValue createDistinctError(String typeIdName, BPackage typeIdP
         return error;
     }
 
-    private static void setTypeId(String typeIdName, BPackage typeIdPkg, ErrorValue error) {
+    public static void setTypeId(String typeIdName, BPackage typeIdPkg, ErrorValue error) {
```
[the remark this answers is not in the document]
kalaiyarasiganeshalingam: No, I didn't check with them.
WebSocketException class extends Errorvalue. It returns ErrorValue without setting typeId. Therefore, We had to set the typeId in the module.

stdlib/http/src/main/java/org/ballerinalang/net/http/websocket/WebSocketUtil.java
rdhananjaya: Had a discussion with @kalaiyarasiganeshalingam and it turned out this exception is mapped to Ballerina distinct error defined here https://github.com/ballerina-platform/ballerina-lang/pull/23896/files#diff-0dc26ba9e7a9ac8f581ee802d2a2338dR21
Hence this change is needed.

Issue #22329, opened 2020-03-31
Fix table usages in HTTP persistence cookie implementation


daneshk:
**Description:**
The previous table type is removed in PR #22284 and the new proposed design can be found in doc[1]. Along with this effort, the usage of previous table type in CSV-IO implementation is removed/commented. The CSV-IO APIs are used in csv_persistent_cookie_handler.bal and it also removed. This task is to fix this once the new table design is available.

https://github.com/ballerina-platform/ballerina-spec/blob/master/lang/proposals/tablenoclone/tablenoclone.md

Issue #22328, opened 2020-03-31
Fix table usages in CSV-IO channel implementation with new table design


daneshk:
**Description:**
The previous table type is removed in PR #22284 and the new proposed design can be found in doc[1]. Along with this effort, the usage of previous table type in CSV-IO implementation is removed/commented. This task is to fix this once the new table design is available. 

1. https://github.com/ballerina-platform/ballerina-spec/blob/master/lang/proposals/tablenoclone/tablenoclone.md

PR #23900, opened 2020-06-10
Migrate stdlib modules to new table syntax

MadhukaHarith92:
## Purpose
Change io:csv, http:cookie modules to the new table syntax and enable test cases.

Fixes #<Issue Number>
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/22328
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/22329


stdlib/http/src/test/resources/test-src/services/nativeimpl/cookie/cookie-native-function-negative.bal
```diff
@@ -248,7 +248,6 @@ function testAddPersistentCookieWithoutPersistentStore() returns @tainted http:C
     }
     return cookies;
 }
-
 //TODO Table remove - Fix
 //function testRemovePersistentCookieFromCookieStore_1() returns @tainted http:Cookie[] {
```
daneshk, line 252: Why this test case is commented out?

stdlib/http/src/test/java/org/ballerinalang/stdlib/services/nativeimpl/cookie/CookieNativeFunctionNegativeTest.java
```diff
@@ -157,6 +157,7 @@ public void testAddPersistentCookieWithoutPersistentStore() {
                           "Cookie objects are in the Return Values");
     }
 
+    //TODO Table remove - Fix
```
daneshk, line 160: Why TODO? Can we remove it

stdlib/io/src/main/ballerina/src/io/readable_csv_channel.bal
```diff
@@ -94,21 +96,23 @@ public type ReadableCSVChannel object {
...
+# Returns a table, which corresponds to the CSV records.
+# ```ballerina
+# var tblResult1 = readableCSVChannel.getTable(Employee);
+# var tblResult2 = readableCSVChannel.getTable(Employee, ["id", "name"]);
+# ```
+#
+# + structType - The object in which the CSV records should be deserialized
+# + tableKey - The key of the table
+# + return - Table, which represents the CSV records or else an `io:Error`
+    public function getTable(typedesc<record {}> structType, public string[] tableKey = [])
```
daneshk, line 108:
```suggestion
    public function getTable(typedesc<record {}> structType, public string[] fieldNames = [])
```

daneshk: Shall we name it as `fieldNames`?

stdlib/io/src/main/ballerina/src/io/readable_csv_channel.bal
```diff
@@ -94,21 +96,23 @@ public type ReadableCSVChannel object {
...
-    //    return getTableExtern(self, structType);
-    //}
+# Returns a table, which corresponds to the CSV records.
+# ```ballerina
+# var tblResult1 = readableCSVChannel.getTable(Employee);
+# var tblResult2 = readableCSVChannel.getTable(Employee, ["id", "name"]);
+# ```
+#
+# + structType - The object in which the CSV records should be deserialized
+# + tableKey - The key of the table
```
daneshk, line 106:
```suggestion
# + fieldNames - The names of the fields used as the (composite)key of the table.
```

tests/jballerina-unit-test/src/test/resources/test-src/types/string/bstring-table-test.bal
```diff
@@ -20,14 +20,11 @@ type Employee record {
...
-           [
-               {1, "Mary🤒", 300.5},
-               {2, "John💉", 200.5},
-               {3, "Jim", 330.5}
-           ]
-       };
+   table<Employee> tbEmployee = table [
+               {id: 1, name: "Mary", salary: 300.5},
+               {id: 2, name: "John", salary: 200.5},
+               {id: 3, name: "Jim", salary: 330.5}
```
daneshk, line 26: Why are we changing these test cases?

MadhukaHarith92: Test cases are the same. Syntax of initializing the table got changed.

stdlib/http/src/test/java/org/ballerinalang/stdlib/services/nativeimpl/cookie/CookieNativeFunctionNegativeTest.java
MadhukaHarith92: Updated in 9a986a0

stdlib/http/src/test/resources/test-src/services/nativeimpl/cookie/cookie-native-function-negative.bal
MadhukaHarith92: This test case checked the error message when removing a non-existing cookie. However, with the new table, it panics if we try to remove an entry from the table if the key does not exist hence the test case failed. Updated in 9a986a0 to assert the runtime exception.

stdlib/io/src/main/java/org/ballerinalang/stdlib/io/nativeimpl/GetTable.java
```diff
@@ -35,69 +58,107 @@
...
+                return IOUtils.createEoFError();
+            }
+            List<String[]> records = new ArrayList<>();
+            while (delimitedChannel.hasNext()) {
+                records.add(delimitedChannel.read());
+            }
+            return getTable(typedescValue, key, records);
+        } catch (BallerinaIOException | BallerinaException e) {
+            String msg = "failed to process the delimited file: " + e.getMessage();
+            log.error(msg, e);
```
Bhashinee, line 77: Do we need to log this error here? As we are returning the error, anyway user will have to handle that.

MadhukaHarith92: Updated in 4312c4f

tests/jballerina-unit-test/src/test/resources/test-src/types/string/bstring-table-test.bal
daneshk: "🤒", "💉" are missing in the updated one

MadhukaHarith92: Didn't know they were intentionally added hence removed. Added back in 70e13a6 .

Issue #24023, opened 2020-06-12
ListenerImplReceiveTest hangs intermittently in Email Connector


Maninda:
**Description:**
`org.ballerinalang.stdlib.email.ListenerImapReceiveTest` hangs for about an hour with the log `INFO: User hascode is listening to remote server at 127.0.0.1...` intermittently.

**Steps to reproduce:**
Build with the unit tests in Email Connector. The build will hang with the above log.

PR #24025, opened 2020-06-12
Add the maximum upper bound to the number of polls in listener of Email Connector

Maninda:
## Purpose
`org.ballerinalang.stdlib.email.ListenerImapReceiveTest` and `org.ballerinalang.stdlib.email.ListenerPopReceiveTest` may hang indefinitely when the email is not correctly received by the email server. In that case the test should fail after a finite number of polls.

Fixes #https://github.com/ballerina-platform/ballerina-lang/issues/24023

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples


stdlib/email/src/main/ballerina/src/email/listener_endpoint.bal
```diff
@@ -164,6 +169,7 @@ public type ListenerConfig record {|
     string protocol = "IMAP";
     PopConfig|ImapConfig? protocolConfig = ();
     int pollingInterval = 60000;
+    int noOfPolls?;
```
daneshk, line 172: what is the use of setting `noOfPolls` from the user perspective?

daneshk: Is it possible to call the listener `stop()` function if we need to stop the polling.

Maninda: Suppose the user wants to handle the error path when there are no emails arriving within a certain time period, he can set that time period as `pollingInterval * noOfPolls`.

Maninda: Yes he can use the`stop()` function too but with this new field it can automatically exit the listening task like a timeout.

Maninda: Reverted the field.

PR #24057, opened 2020-06-13
Add lang-lib methods to improve JSON compatibility

### LINKED ISSUE #23335: Improve JSON compatibility
2020R2 Spec Add following functions
- [ ] fromJsonWithType(json v, typedesc<anydata> t) returns t|error
- [ ] fromJsonStringWithType(string str, typedesc<anydata> t) returns t|error
- [ ] toJson(anydata v) returns json
- [ ] toJsonString(anydata v) returns string 

Spec issue : https://github.com/ballerina-platform/ballerina-spec/issues/353

dulajdilshan:
## Purpose
Add following functions 

1.  fromJsonWithType(json v, typedesc t) returns t|error
2. fromJsonStringWithType(string str, typedesc t) returns t|error
3. toJson(anydata v) returns json
4. toJsonString(anydata v) returns string

Fixes #23335, #24247, #24625 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.value/src/main/ballerina/src/lang.value/value.bal
```diff
@@ -117,9 +117,26 @@ public function toString((any|error) v) returns string = external;
...
-# + v - json value
+# + v - anydata value
 # + return - string representation of json
-public function toJsonString(json v) returns string = external;
+public function toJsonString(anydata v) returns string = external;
+
+# Converts a value of type `anydata` to `json`.
+# This does a deep copy of `v` converting values that do
+# not belong to json into values that do.
+# A value of type `xml` is converted into a string as if
```
MaryamZi, line 127: XML conversion doesn't seem to work when its in a map?

```ballerina
public function main() {
    xml x = xml `<foo><bar/>Hello World</foo>`;
    json j1 = x.toJson(); // WORKS

    map<anydata> m1 = {a: 1, b: "str"};
    json j2 = m1.toJson(); // WORKS

    map<anydata> m2 = {a: 1, b: x};
    json j3 = m2.toJson(); // ERROR - error: {ballerina/lang.typedesc}ConversionError 
                        // message='map<anydata>' value cannot be converted to 'json'
}
```

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonStringWithType.java
```diff
@@ -0,0 +1,72 @@
...
+        String str = value.getValue();
+        if (str.equals("null")) {
+            return null;
+        }
+        Reader reader = new StringReader(str);
+        try {
+            Object jsonFromString = JSONParser.parse(reader);
+            return FromJsonWithType.fromJsonWithType(strand, jsonFromString, t);
+        } catch (BallerinaException e) {
+            return BallerinaErrors.createError("{ballerina}FromJsonStringWithTypeError", e.getMessage());
```
pubudu91, line 69: Double check this error message format. AFAIK, the convention is to prefix the reason with the module. But I can see that some of the `lang.value` functions have used this format. Let's check with the others and verify.

langlib/lang.value/src/main/ballerina/src/lang.value/value.bal
```diff
@@ -117,9 +117,26 @@ public function toString((any|error) v) returns string = external;
 
 # Returns the string that represents `v` in JSON format.
 #
-# + v - json value
```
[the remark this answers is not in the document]
dulajdilshan: In here the spec is wrong right ? I'll correct the rest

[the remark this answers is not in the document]
MaryamZi: Shall we create an issue in the spec repo for this?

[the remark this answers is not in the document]
dulajdilshan: Sure I'll create

langlib/lang.value/src/main/ballerina/src/lang.value/value.bal
dulajdilshan: Solved

langlib/lang.value/src/main/ballerina/src/lang.value/value.bal
```diff
@@ -117,9 +117,26 @@ public function toString((any|error) v) returns string = external;
...
-public function toJsonString(json v) returns string = external;
+public function toJsonString(anydata v) returns string = external;
+
+# Converts a value of type `anydata` to `json`.
+# This does a deep copy of `v` converting values that do
+# not belong to json into values that do.
+# A value of type `xml` is converted into a string as if
+# by the `toString` function.
+# A value of type `table` is converted into a list of
+# mappings one for each row.
```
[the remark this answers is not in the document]
dulajdilshan: Solved

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/JSONParser.java
```diff
@@ -687,7 +687,7 @@ public State transition(StateMachine sm, char[] buff, int i, int count) throws J
                     ch = buff[i];
                     sm.processLocation(ch);
                     if (ch == sm.currentQuoteChar) {
-                        ((ArrayValue) sm.currentJsonNode).append(sm.value());
+                        ((ArrayValue) sm.currentJsonNode).append(changeForBString(sm.value()));
```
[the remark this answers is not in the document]
dulajdilshan: Got it. Solved

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,290 @@
...
+        List<BType> convertibleTypes;
+        convertibleTypes = TypeConverter.getConvertibleTypes(value, targetType);
+        if (convertibleTypes.isEmpty()) {
+            if (TypeChecker.getType(value).getTag() == TypeTags.STRING_TAG) {
+                try {
+                    return XMLFactory.parse(((StringValue) value).getValue());
+                } catch (Throwable e) {
+                    throw createConversionError(value, targetType, e.getMessage());
+                }
+            }
```
pubudu91, line 136: Why are we trying to convert it to an XML value if there aren't any convertible types? Consider the following example,
```ballerina
import ballerina/io;

public function main() {
    string foo = "Foo";
    var v = foo.fromJsonWithType(int);
    io:println(v);
}
```
The above fails with the following panic:
```
error {ballerina/lang.typedesc}ConversionError message='string' value cannot be converted to 'int': failed to parse xml: Unexpected character 'F' (code 70) in prolog; expected '<'
 at [row,col {unknown-source}]: [1,1]
```
It says `string` value cannot be converted to `int` but then has an XML parsing error as well.

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonStringWithType.java
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,227 @@
...
+        switch (targetType.getTag()) {
+            case TypeTags.MAP_TAG:
+                MapValueImpl<BString, Object> newMap = new MapValueImpl<>(targetType);
+                for (Map.Entry entry : map.entrySet()) {
+                    BType constraintType = ((BMapType) targetType).getConstrainedType();
+                    putToMap(newMap, entry, constraintType, unresolvedValues, strand);
+                }
+                return newMap;
+            case TypeTags.JSON_TAG:
+                BType matchingType = TypeConverter.resolveMatchingTypeForUnion(map, targetType);
```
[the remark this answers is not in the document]
dulajdilshan: Solved

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: The spec says that `fromJsonWithType` should do the inverse of the `toJson` which converts `xml` into a  string. 
So, `fromJsonWithType` should be able to convert an appropriate `string` to an `xml`. 

Changing the behaviour of `TypeConverter.getConvertibleTypes()` to support this would also work, but will affect `cloneWithType` method and future methods to be implemented (if there are)

pubudu91: Yeah, it doesn't make sense to try and convert to XML if I didn't specify an XML typedesc

dulajdilshan: Yeah. Thanks!

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,291 @@
...
+                return null;
+            }
+            return createError(CONSTRUCT_FROM_CONVERSION_ERROR,
+                    BLangExceptionHelper.getErrorMessage(RuntimeErrors.CANNOT_CONVERT_NIL, targetType));
+        }
+        List<BType> convertibleTypes;
+        convertibleTypes = TypeConverter.getConvertibleTypes(value, targetType);
+        if (convertibleTypes.isEmpty()) {
+            if (TypeChecker.getType(value).getTag() == TypeTags.STRING_TAG &&
+                    (targetType.getTag() >= TypeTags.XML_ELEMENT_TAG || targetType.getTag() == TypeTags.XML_TAG)) {
```
pubudu91, line 131: Shall we add an upper bound for the subtypes as well for the following? If we add another `anydata` type, this will break.
```
targetType.getTag() >= TypeTags.XML_ELEMENT_TAG
```

dulajdilshan: okay

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,291 @@
...
+                TupleValueImpl newTuple = new TupleValueImpl(tupleType);
+                int minLen = tupleType.getTupleTypes().size();
+                for (int i = 0; i < array.size(); i++) {
+                    BType elementType = (i < minLen) ? tupleType.getTupleTypes().get(i) : tupleType.getRestType();
+                    Object newValue = convert(array.get(i), elementType, unresolvedValues, t, strand);
+                    newTuple.add(i, newValue);
+                }
+                return newTuple;
+            case TypeTags.JSON_TAG:
+                newArray = new ArrayValueImpl(new BArrayType(BTypes.typeJSON));
```
pubudu91, line 260: Since this array type is not going to change, how about adding it to `BTypes`, instead of creating a new JSON array type each time?

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,344 @@
...
+            case TypeTags.XML_COMMENT_TAG:
+            case TypeTags.XML_PI_TAG:
+            case TypeTags.XML_TEXT_TAG:
+                newValue = ToString.toString(strand, value);
+                break;
+            case TypeTags.TABLE_TAG:
+                try {
+                    newValue = JSONUtils.toJSON((TableValueImpl) value);
+                } catch (Exception e) {
+                    return createConversionError(value, targetType, e.getMessage());
```
pubudu91, line 142: This should be thrown right? Since the signature of `toJson()` is `json`

misc/testerina/modules/report-tools/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ballerina/testerina-report",
-  "version": "2.0.0-Preview1-M3",
```
MaryamZi, line 3: Why are we committing this and the next file?

dulajdilshan: I'll correct it

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
dulajdilshan: Ack. Thanks

pubudu91:
The following fails: 
```ballerina
    table<map<anydata>> t = table [
        {name: "John Doe", age: 25},
        {name: "Jane Doe", age: 20}
    ];

    j = t.toJson();
    io:println(j);
```
Not a bug in this PR though, created #24231

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,340 @@
...
+            if (targetType.isNilable()) {
+                return null;
+            }
+            return createError(CONSTRUCT_FROM_CONVERSION_ERROR,
+                    BLangExceptionHelper.getErrorMessage(RuntimeErrors.CANNOT_CONVERT_NIL, targetType));
+        }
+        List<BType> convertibleTypes;
+        convertibleTypes = getConvertibleTypesFromJson(value, targetType);
+        if (convertibleTypes.isEmpty()) {
+            throw createConversionError(value, targetType);
```
dulajdilshan, line 128: Should return

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,340 @@
...
+        } else if (!allowAmbiguity && convertibleTypes.size() > 1) {
```
dulajdilshan, line 129: no need isambiguity

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,340 @@
...
+
+        BType sourceType = TypeChecker.getType(value);
+        BType matchingType = convertibleTypes.get(0);
+
+        // handle string (xml as string) to xml conversion
+        if (TypeChecker.getType(value).getTag() == TypeTags.STRING_TAG &&
+                ((matchingType.getTag() >= TypeTags.XML_ELEMENT_TAG && matchingType.getTag() <= TypeTags.XML_TEXT_TAG)
+                        || matchingType.getTag() == TypeTags.XML_TAG)) {
+            try {
+                return XMLFactory.parse(((StringValue) value).getValue());
```
dulajdilshan, line 141: Error put a a cause

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibValueTest.java
```diff
@@ -41,7 +41,7 @@
  *
  * @since 1.0
  */
-@Test(enabled = false)
+@Test(enabled = true)
```
MaryamZi, line 44:
```suggestion
@Test
```

MaryamZi: We can just remove `(enabled = false)`. Please fix everywhere.

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,268 @@
...
+                        // Has to be a numeric conversion.
+                        newValue = TypeConverter.convertValues(matchingType, value);
+                    }
+                    break;
+                }
+                // should never reach here
+                throw BallerinaErrors.createConversionError(value, targetType);
+        }
+
+        unresolvedValues.remove(typeValuePair);
```
MaryamZi, line 157: Do we have to remove since we'll be creating a new list for each call to the method? Can you check others too.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeConverter.java
```diff
@@ -263,7 +264,44 @@ static boolean isConvertibleToFloatingPointTypes(Object value) {
...
+                                                          List<TypeValuePair> unresolvedValues) {
+        List<BType> convertibleTypes = new ArrayList<>();
+
+        int targetTypeTag = targetType.getTag();
+
+        convertibleTypes.addAll(TypeConverter.getConvertibleTypes(value, targetType));
+
+        if (convertibleTypes.size() == 0) {
+            switch (targetTypeTag) {
+                case TypeTags.RECORD_TYPE_TAG:
```
MaryamZi, line 277: We need to consider maps also right? And arrays/tuples?

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -50,6 +50,33 @@ function testToJsonString() returns map<string> {
     return result;
 }
 
+function testToJsonStringForNonJsonTypes() {
```
pubudu91, line 53: Can you also include a couple of tests where,
- the variable type is `anydata`
- the variable type is a record with an `xml` field

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TableJSONDataSource.java
```diff
@@ -159,7 +159,11 @@ private static void constructJsonData(MapValueImpl record, MapValue<BString, Obj
                         getStructData(record.getMapValue(key), structFields, index, key));
                 break;
             case TypeTags.XML_TAG:
-                jsonObject.put(StringUtils.fromString(name), record.getStringValue(key));
+                Object val = record.get(key);
+                BString strVal = org.ballerinalang.jvm.StringUtils.fromString(
+                        org.ballerinalang.jvm.values.utils.StringUtils.getStringValue(val));
+                jsonObject.put(StringUtils.fromString(name), strVal);
+//                jsonObject.put(StringUtils.fromString(name), record.getStringValue(key));
```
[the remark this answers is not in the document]
dulajdilshan: Ack

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -343,6 +343,56 @@ public static boolean isSameType(BType sourceType, BType targetType) {
...
+        BType sourceType = TypeChecker.getType(sourceValue);
+
+        if (TypeChecker.checkIsLikeType(sourceValue, BTypes.typeJSON, true)) {
+            return true;
+        }
+
+        if (sourceType.getTag() == TypeTags.ARRAY_TAG) {
+            ArrayValue source = (ArrayValue) sourceValue;
+            BType elementType = ((BArrayType) source.getType()).getElementType();
+            if (BTypes.isValueType(elementType)) {
```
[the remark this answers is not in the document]
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,198 @@
...
+        isPublic = true
+)
+public class ToJson {
+
+    private static final String AMBIGUOUS_TARGET = "ambiguous target type";
+
+    public static Object toJson(Strand strand, Object value) {
+        BType targetJsonType = BTypes.typeJSON;
+        try {
+            return convert(value, targetJsonType, new ArrayList<>(), strand);
```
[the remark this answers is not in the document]
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,198 @@
...
+        } catch (Exception e) {
+            return e;
+        }
+    }
+
+    private static Object convert(Object value, BType targetType, List<TypeValuePair> unresolvedValues, Strand strand) {
+        if (value == null) {
+            if (targetType.isNilable()) {
```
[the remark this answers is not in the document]
dulajdilshan: Ack

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeConverter.java
dulajdilshan: They were caught in this method call in line: 310
`TypeConverter.getConvertibleTypes(value, targetType)`

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
dulajdilshan: Sure.

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,198 @@
...
+                                             Strand strand) {
+        ArrayValueImpl newArray = new ArrayValueImpl(new BArrayType(BTypes.typeJSON));
+        for (int i = 0; i < array.size(); i++) {
+            Object newValue = convert(array.get(i), BTypes.typeJSON, unresolvedValues, strand);
+            newArray.add(i, newValue);
+        }
+        return newArray;
+    }
+
+    private static void putToMap(MapValue<BString, Object> map, Map.Entry entry, BType fieldType,
```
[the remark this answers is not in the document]
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,198 @@
...
+        for (Map.Entry entry : map.entrySet()) {
+            BType constraintType = ((BMapType) targetType).getConstrainedType();
+            putToMap(newMap, entry, constraintType, unresolvedValues, strand);
+        }
+        return newMap;
+    }
+
+    private static Object convertArrayToJson(ArrayValue array, List<TypeValuePair> unresolvedValues,
+                                             Strand strand) {
+        ArrayValueImpl newArray = new ArrayValueImpl(new BArrayType(BTypes.typeJSON));
```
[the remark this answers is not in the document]
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,198 @@
...
+
+        unresolvedValues.remove(typeValuePair);
+        return newValue;
+    }
+
+    private static Object convertMapToJson(MapValue<?, ?> map, BType targetType, List<TypeValuePair> unresolvedValues,
+                                           Strand strand) {
+        MapValueImpl<BString, Object> newMap = new MapValueImpl<>(targetType);
+        for (Map.Entry entry : map.entrySet()) {
+            BType constraintType = ((BMapType) targetType).getConstrainedType();
```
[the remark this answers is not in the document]
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,198 @@
...
+
+        if (unresolvedValues.contains(typeValuePair)) {
+            throw new BallerinaException(VALUE_LANG_LIB_CYCLIC_VALUE_REFERENCE_ERROR,
+                    BLangExceptionHelper.getErrorMessage(RuntimeErrors.CYCLIC_VALUE_REFERENCE,
+                            ((RefValue) value).getType()));
+        }
+
+        unresolvedValues.add(typeValuePair);
+
+        if (!TypeChecker.isConvertibleToJson(value, new ArrayList<>())) {
```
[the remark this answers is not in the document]
dulajdilshan: Ack. Thanks

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,198 @@
...
+                    throw createConversionError(value, targetType);
+                }
+                break;
+        }
+
+        unresolvedValues.remove(typeValuePair);
+        return newValue;
+    }
+
+    private static Object convertMapToJson(MapValue<?, ?> map, BType targetType, List<TypeValuePair> unresolvedValues,
```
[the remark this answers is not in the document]
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: But, There can be situations like
```ballerina
public function main() {
    json anObject = { name : "anObject", value : 10, sub : { subName : "subObject", subValue : 10 }};
    string k = anObject.toJsonString();
}
```

if we remove this line :`unresolvedValues.remove(typeValuePair);` ,  `10` is duplicated isn't it ?

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibValueTest.java
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,198 @@
...
+                } catch (Exception e) {
+                    throw createConversionError(value, targetType, e.getMessage());
+                }
+                break;
+            case TypeTags.RECORD_TYPE_TAG:
+            case TypeTags.MAP_TAG:
+                newValue = convertMapToJson((MapValue<?, ?>) value, new BMapType(BTypes.typeJSON),
+                        unresolvedValues, strand);
+                break;
+            case TypeTags.ERROR_TAG:
```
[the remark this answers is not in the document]
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,198 @@
...
+                newValue = ((RefValue) value).copy(new HashMap<>());
+                break;
+            default:
```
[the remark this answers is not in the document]
dulajdilshan: Chnaged.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeConverter.java
```diff
@@ -263,7 +264,44 @@ static boolean isConvertibleToFloatingPointTypes(Object value) {
...
-    private static boolean isConvertibleToRecordType(Object sourceValue, BRecordType targetType,
+    public static List<BType> getConvertibleTypesFromJson(Object value, BType targetType,
+                                                          List<TypeValuePair> unresolvedValues) {
+        List<BType> convertibleTypes = new ArrayList<>();
+
+        int targetTypeTag = targetType.getTag();
+
+        convertibleTypes.addAll(TypeConverter.getConvertibleTypes(value, targetType));
+
+        if (convertibleTypes.size() == 0) {
```
MaryamZi, line 275: `TypeConverter.getConvertibleTypes(value, targetType)` calls `isConvertibleToRecordType` with `isFromJson` set to `false` right?

What if `targetType` is a union of two records, say `Foo|Bar`, and the value is convertible to `Foo` even when  `isFromJson` is `false` in `isConvertibleToRecordType`, but it is only convertible to `Bar` when `isFromJson` is true?

So IINM, from L273 when we reach here `convertibleTypes` will only have `Foo` and we return that in L301. But if we actually check `isConvertibleToRecordType(value, (BRecordType) targetType, true, unresolvedValues)` it'll evaluate to `true` for `Bar`, and there'll actually be 2 convertible types?

MaryamZi: For example

```ballerina
import ballerina/io;

type Foo record {|
    int i;
    xml x;
|};

type Bar record {|
    int i;
    string x;
|};

type FooBar Foo|Bar;

public function main() {
    json j = {i: 1, x: "<abc>ABC</abc>"};

    // No error.
    io:println(j.fromJsonWithType(Foo)); // i=1 x=<abc>ABC</abc>
    // No error.
    io:println(j.fromJsonWithType(Bar)); // i=1 x=<abc>ABC</abc>

    // Then this should be ambiguous?
    io:println(j.fromJsonWithType(FooBar)); // i=1 x=<abc>ABC</abc>
}
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeConverter.java
```diff
@@ -263,7 +264,44 @@ static boolean isConvertibleToFloatingPointTypes(Object value) {
...
+            switch (targetTypeTag) {
+                case TypeTags.RECORD_TYPE_TAG:
+                    if (isConvertibleToRecordType(value, (BRecordType) targetType, true, unresolvedValues)) {
+                        convertibleTypes.add(targetType);
+                    }
+                    break;
+                case TypeTags.TABLE_TAG:
+                    if (((BTableType) targetType).getConstrainedType().getTag() == TypeTags.RECORD_TYPE_TAG ||
+                            ((BTableType) targetType).getConstrainedType().getTag() == TypeTags.MAP_TAG) {
```
MaryamZi, line 284: Re: the constraint, the spec says

> The type specified by a row-type-parameter must be a subtype of map<any|error>

So it is always a record or a map right? Do we need this check?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeConverter.java
```diff
@@ -299,17 +337,33 @@ private static boolean isConvertibleToRecordType(Object sourceValue, BRecordType
             Map.Entry valueEntry = (Map.Entry) object;
             String fieldName = valueEntry.getKey().toString();
 
-            if (targetFieldTypes.containsKey(fieldName)) {
-                if (getConvertibleTypes(valueEntry.getValue(), targetFieldTypes.get(fieldName),
-                                        unresolvedValues).size() != 1) {
+            if (isFromJson) {
+                if (targetFieldTypes.containsKey(fieldName)) {
+                    if (getConvertibleTypesFromJson(valueEntry.getValue(), targetFieldTypes.get(fieldName),
```
MaryamZi, line 342: The only difference in the if and else blocks is the method called right? The arguments are all the same? Instead of the `isFromJson` parameter can we not use something like a functional interface, which I believe is what @rdhananjaya also suggested once.

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonStringWithType.java
```diff
@@ -0,0 +1,71 @@
...
+        },
+        isPublic = true
+)
+public class FromJsonStringWithType {
+
+    public static Object fromJsonStringWithType(Strand strand, BString value, TypedescValue t) {
+
+        String str = value.getValue();
+        if (str.equals("null")) {
+            return null;
```
MaryamZi, line 61: We need to check if the target type is nilable.

The folowing crashes.

```ballerina
type IntMap map<int>;

public function main() {
    string s = "null";

    map<int> m = checkpanic s.fromJsonStringWithType(IntMap); // doesn't panic

    int i = m.get("first");
}
```

The get fails with an NPE in the internal log.
```log
[2020-07-16 00:57:59,786] SEVERE {b7a.log.crash} - null 
	at org.ballerinalang.langlib.map.Get.get(Get.java:45)
	at ballerina.lang_map.1_1_0.map.get(map.bal:51)
	at qwe.main(qwe.bal:8)
	at ___init.$lambda$main$(.)
	at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:500)
	at org.ballerinalang.jvm.scheduling.Scheduler.run(Scheduler.java:276)
	at org.ballerinalang.jvm.scheduling.Scheduler.runSafely(Scheduler.java:244)
```

MaryamZi: Do we need to special case "null" anyway? The `JSONParser` and `FromJsonWithType.fromJsonWithType` will handle this properly right?

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,268 @@
...
+            case TypeTags.ARRAY_TAG:
+            case TypeTags.TUPLE_TAG:
+                newValue = convertArray((ArrayValue) value, matchingType, unresolvedValues, t, strand);
+                break;
+            case TypeTags.ERROR_TAG:
+                newValue = ((RefValue) value).copy(new HashMap<>());
+                break;
+            default:
+                if (((matchingType.getTag() >= TypeTags.XML_ELEMENT_TAG &&
+                        matchingType.getTag() <= TypeTags.XML_TEXT_TAG) || matchingType.getTag() == TypeTags.XML_TAG)) {
```
MaryamZi, line 134: Can't we use `org.ballerinalang.jvm.types.TypeTags#isXMLTypeTag` here instead?

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,268 @@
...
+        return newValue;
+    }
+
+    private static Object convertMap(MapValue<?, ?> map, BType targetType, List<TypeValuePair> unresolvedValues,
+                                     TypedescValue t, Strand strand) {
+        switch (targetType.getTag()) {
+            case TypeTags.MAP_TAG:
+                MapValueImpl<BString, Object> newMap = new MapValueImpl<>(targetType);
+                for (Map.Entry entry : map.entrySet()) {
+                    BType constraintType = ((BMapType) targetType).getConstrainedType();
```
MaryamZi, line 167: We should move this out of the for loop, or we'll be extracting the constraint per entry.

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,268 @@
...
+
+                for (Map.Entry entry : map.entrySet()) {
+                    BType fieldType = targetTypeField.getOrDefault(entry.getKey().toString(), restFieldType);
+                    putToMap(newRecord, entry, fieldType, unresolvedValues, t, strand);
+                }
+                return newRecord;
+            case TypeTags.JSON_TAG:
+                BType matchingType = TypeConverter.resolveMatchingTypeForUnion(map, targetType);
+                return convert(map, matchingType, unresolvedValues, t, strand);
+            default:
```
MaryamZi, line 195: This also we don't need right?

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,268 @@
...
+                for (int i = 0; i < array.size(); i++) {
+                    BType elementType = (i < minLen) ? tupleType.getTupleTypes().get(i) : tupleType.getRestType();
+                    Object newValue = convert(array.get(i), elementType, unresolvedValues, t, strand);
+                    newTuple.add(i, newValue);
+                }
+                return newTuple;
+            case TypeTags.JSON_TAG:
+                newArray = new ArrayValueImpl((BArrayType) BTypes.typeJsonArray);
+                for (int i = 0; i < array.size(); i++) {
+                    Object newValue = convert(array.get(i), BTypes.typeJSON, unresolvedValues, t, strand);
```
MaryamZi, line 227: Better to pass the `targetType` itself instead of `BTypes.typeJSON` since it could be `json & readonly`.

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,268 @@
...
+                    newArray.add(i, newValue);
+                }
+                return newArray;
+            case TypeTags.TABLE_TAG:
+                BTableType tableType = (BTableType) targetType;
+
+                BTableType newTableType = new BTableType(tableType.getConstrainedType(),
```
MaryamZi, line 234: Why do we create a new type?

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -0,0 +1,169 @@
...
+            return null;
+        }
+
+        BType sourceType = TypeChecker.getType(value);
+
+        if (sourceType.getTag() <= TypeTags.BOOLEAN_TAG) {
+            if (TypeChecker.checkIsType(value, jsonType)) {
+                return value;
+            } else {
+                // Has to be a numeric conversion.
```
MaryamZi, line 89: Do we ever reach here? All the numeric basic types are both `anydata` and `json`. So, I don't think you'll ever have to do a numeric conversion?

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonStringWithType.java
dulajdilshan: Ack.

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,268 @@
...
+            case TypeTags.MAP_TAG:
+            case TypeTags.RECORD_TYPE_TAG:
+                newValue = convertMap((MapValue<?, ?>) value, matchingType, unresolvedValues, t, strand);
+                break;
+            case TypeTags.ARRAY_TAG:
+            case TypeTags.TUPLE_TAG:
+                newValue = convertArray((ArrayValue) value, matchingType, unresolvedValues, t, strand);
+                break;
+            case TypeTags.ERROR_TAG:
+                newValue = ((RefValue) value).copy(new HashMap<>());
```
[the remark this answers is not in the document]
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: Sure

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: Ack

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeConverter.java
dulajdilshan: Shouldn't this be okay according to [this](https://github.com/ballerina-platform/ballerina-spec/issues/556) ?

Let's take this
```ballerina
 io:println(j.fromJsonWithType(FooBar));
```
When we reach L273, `convertibleTypes` will have only have `Bar`.  isn't it?

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,268 @@
...
+                MapValueImpl<BString, Object> newMap = new MapValueImpl<>(targetType);
+                for (Map.Entry entry : map.entrySet()) {
+                    BType constraintType = ((BMapType) targetType).getConstrainedType();
+                    putToMap(newMap, entry, constraintType, unresolvedValues, t, strand);
+                }
+                return newMap;
+            case TypeTags.RECORD_TYPE_TAG:
+                BRecordType recordType = (BRecordType) targetType;
+                MapValueImpl<BString, Object> newRecord;
+                if (t != null && t.getDescribingType() == targetType) {
```
pubudu91, line 174: Do we need this null check here? `t` can never be `null` right?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeConverter.java
dulajdilshan: I think yes. because  If I remove the above-mentioned check, 
```ballerina
public function main() {
    json j = [
        "cake",
        "buscuit"
    ];
    table<string>|error tabString = j.fromJsonWithType(TableString);
}
```
This works without giving a prior error for `table<string>`

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeConverter.java
```diff
@@ -263,7 +264,44 @@ static boolean isConvertibleToFloatingPointTypes(Object value) {
...
+                case TypeTags.XML_ELEMENT_TAG:
+                case TypeTags.XML_COMMENT_TAG:
+                case TypeTags.XML_PI_TAG:
+                case TypeTags.XML_TEXT_TAG:
+                    if (TypeChecker.getType(value).getTag() == TypeTags.STRING_TAG) {
+                        convertibleTypes.add(targetType);
+                    }
+                    break;
+                default:
+                    break;
```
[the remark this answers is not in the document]
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
```diff
@@ -0,0 +1,268 @@
...
+        functionName = "fromJsonWithType",
+        isPublic = true
+)
+public class FromJsonWithType {
+    private static final String AMBIGUOUS_TARGET = "ambiguous target type";
+
+    public static Object fromJsonWithType(Strand strand, Object v, TypedescValue t) {
+        BType describingType = t.getDescribingType();
+        try {
+            return convert(v, describingType, new ArrayList<>(), t, strand);
```
pubudu91, line 81: Is there a need to pass both `t` and `describingType`?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeConverter.java
MaryamZi: Ah, didn't know of the "fromJsonWithType should only do something different from cloneWithType if cloneWithType would produce an error." part.
 
Yeah, this behaviour is correct then. Thanks!

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: Removed

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: ok

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
dulajdilshan: Got it

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: Ack

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
dulajdilshan: I followed the same implementation as in the `cloneWithType`. 

`convert` method is get called later. Sometimes, `t` and `targetType` both are  needed

pubudu91: My point was if you have `t`, you can always get the type from it. Anyways, not a major issue. Just gives the impression that target type can be some other type than what's described by `t`

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonStringWithType.java
```diff
@@ -0,0 +1,72 @@
...
+        isPublic = true
+)
+public class FromJsonStringWithType {
+
+    public static Object fromJsonStringWithType(Strand strand, BString value, TypedescValue t) {
+
+        String str = value.getValue();
+        try {
+            if (str.equals("null")) {
+                return FromJsonWithType.fromJsonWithType(strand, null, t);
```
pubudu91, line 62: I think what @MaryamZi meant was to simply pass `str` to the JSON parser without special casing "null".

PR #24093, opened 2020-06-14
Fix type aliasing issue for typedescs

pubudu91:


SupunS:
Thanks @pubudu91  for the quick fix!

PR #24106, opened 2020-06-14
Add support for cyclic dependency in union type definitions

### LINKED ISSUE #23690: Allow cyclic reference in unions typedef
**Description:**
Currently the following snippet
```ballerina
type SomeName string|SomeName[];
```
produces the following compilation error:
`cyclic type reference in '[SomeName, SomeName]`

But we need to support cyclic reference in unions.

irshadnilam:
## Purpose
This PR adds support for cyclic dependency in union type definitions

Fixes #23690

Note: Opening as a Draft PR, more tests needs to be added. 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
@@ -709,6 +709,14 @@ public class TypeChecker {
     }
 
     private static boolean checkIsUnionType(BType sourceType, BUnionType targetType, List<TypePair> unresolvedTypes) {
+        // If we encounter two types that we are still resolving, then skip it.
+        // This is done to avoid recursive checking of the same type.
+        TypePair pair = new TypePair(sourceType, targetType);
+        if (unresolvedTypes.contains(pair)) {
+            return true;
+        }
+        unresolvedTypes.add(pair);
+
         switch (sourceType.getTag()) {
             case TypeTags.UNION_TAG:
                 return isUnionTypeMatch((BUnionType) sourceType, targetType, unresolvedTypes);
@@ -2283,42 +2291,46 @@ public class TypeChecker {
                 BTupleType tupleType = (BTupleType) type;
                 return tupleType.getTupleTypes().stream().allMatch(TypeChecker::hasFillerValue);
             case TypeTags.UNION_TAG:
-                return checkFillerValue((BUnionType) type);
+                return checkFillerValue((BUnionType) type, unanalyzedTypes);
             default:
                 return false;
         }
     }
 
-    private static boolean checkFillerValue(BUnionType type) {
-        // NIL is a member.
-        if (type.isNullable()) {
-            return true;
-        }
-        // All members are of same type.
-        Iterator<BType> iterator = type.getMemberTypes().iterator();
-        BType firstMember;
-        for (firstMember = iterator.next(); iterator.hasNext(); ) {
-            if (!isSameType(firstMember, iterator.next())) {
-                return false;
-            }
-        }
-        // Control reaching this point means there is only one type in the union.
-        return BTypes.isValueType(firstMember) && hasFillerValue(firstMember);
-    }
-
-    private static boolean checkFillerValue(BRecordType type, List<BType> unAnalyzedTypes) {
+    private static boolean checkFillerValue(BType type, List<BType> unAnalyzedTypes) {
         if (unAnalyzedTypes.contains(type)) {
             return true;
         }
         unAnalyzedTypes.add(type);
-        for (BField field : type.getFields().values()) {
-            if (Flags.isFlagOn(field.flags, Flags.OPTIONAL)) {
-                continue;
-            }
-            if (!Flags.isFlagOn(field.flags, Flags.REQUIRED)) {
-                continue;
-            }
-            return false;
+        switch (type.getTag()) {
+            case TypeTags.RECORD_TYPE_TAG:
+                BRecordType recordType = (BRecordType) type;
+                for (BField field : recordType.getFields().values()) {
+                    if (Flags.isFlagOn(field.flags, Flags.OPTIONAL)) {
+                        continue;
+                    }
+                    if (!Flags.isFlagOn(field.flags, Flags.REQUIRED)) {
+                        continue;
+                    }
+                    return false;
+                }
+                break;
+            case TypeTags.UNION_TAG:
+                BUnionType unionType = (BUnionType) type;
+                // NIL is a member.
+                if (unionType.isNullable()) {
+                    return true;
+                }
+                // All members are of same type.
+                Iterator<BType> iterator = unionType.getMemberTypes().iterator();
+                BType firstMember;
+                for (firstMember = iterator.next(); iterator.hasNext(); ) {
+                    if (!isSameType(firstMember, iterator.next())) {
+                        return false;
+                    }
+                }
+                // Control reaching this point means there is only one type in the union.
+                return BTypes.isValueType(firstMember) && hasFillerValue(firstMember);
         }
         return true;
     }
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BUnionType.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BUnionType.java
@@ -35,6 +35,7 @@ public class BUnionType extends BType {
     private int typeFlags;
     private final boolean readonly;
     private BIntersectionType immutableType;
+    private boolean resolving = false;
 
     /**
      * Create a {@code BUnionType} which represents the union type.
@@ -68,22 +69,19 @@ public class BUnionType extends BType {
 
     public BUnionType(List<BType> memberTypes, boolean readonly) {
         this(memberTypes, 0, readonly);
-        boolean nilable = false, isAnydata = true, isPureType = true;
-        for (BType memberType : memberTypes) {
-            nilable |= memberType.isNilable();
-            isAnydata &= memberType.isAnydata();
-            isPureType &= memberType.isPureType();
-        }
+        setFlagsBasedOnMembers();
+    }
 
-        if (nilable) {
-            this.typeFlags = TypeFlags.addToMask(this.typeFlags, TypeFlags.NILABLE);
-        }
-        if (isAnydata) {
-            this.typeFlags = TypeFlags.addToMask(this.typeFlags, TypeFlags.ANYDATA);
-        }
-        if (isPureType) {
-            this.typeFlags = TypeFlags.addToMask(this.typeFlags, TypeFlags.PURETYPE);
-        }
+    /**
+     * Constructor used when defining union type defs where cyclic reference is possible.
+     *
+     * @param typeFlags flags associated with the type
+     * @param readonly boolean represents if the type is readonly
+     */
+    public BUnionType(int typeFlags, boolean readonly) {
+        super(null, null, Object.class);
+        this.typeFlags = typeFlags;
+        this.readonly = readonly;
     }
 
     public BUnionType(BType[] memberTypes, int typeFlags, boolean readonly) {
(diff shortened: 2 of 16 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BUnionType.java
```diff
@@ -127,6 +127,10 @@ public int getTag() {
 
     @Override
     public String toString() {
+        if (resolving) {
+            return BLangConstants.CYCLIC_UNION;
+        }
+        resolving = true;
```
irshadnilam, line 133: Needs to be discussed. Printing error messages for cyclic types (`stackoverflow`) , what would be the best approach?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
```diff
@@ -1088,7 +1147,12 @@ static void loadType(MethodVisitor mv, BType bType) {
             loadErrorType(mv, (BErrorType) bType);
             return;
         } else if (bType.tag == TypeTags.UNION) {
-            loadUnionType(mv, (BUnionType) bType);
+            BUnionType unionType = (BUnionType) bType;
+            if (unionType.isTypeDef) {
```
MaryamZi, line 1151: This `isTypeDef` is what we'll be moving to a flag right? 

Looping in @pubudu91 for his input too. Basically what we want to do is to have some way to know if this particular union is the type of a type-definition. 

```ballerina
type Cyclic int|Cyclic[];
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BUnionType.java
```diff
@@ -189,11 +205,15 @@ public boolean isNilable() {
     }
 
     private boolean checkNillable(List<BType> memberTypes) {
+        this.resolving = true;
         for (BType t : memberTypes) {
             if (t.isNilable()) {
+                this.resolving = false;
                 return true;
             }
+            this.resolving = false;
```
MaryamZi, line 214: Is this correct?  Doesn't this need to be true until all members are iterated over?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -605,70 +609,82 @@ private void defineTypeNodes(List<? extends TypeDefinition> typeDefs, SymbolEnv
...
-                    // Eg - A -> B -> C -> B // Last B is what we are currently checking
-                    //
-                    // In such case, we create a new list with relevant type names.
-                    int i = visitedNodes.indexOf(currentTypeNodeName);
-                    List<String> dependencyList = new ArrayList<>(visitedNodes.size() - i);
-                    for (; i < visitedNodes.size(); i++) {
-                        dependencyList.add(visitedNodes.get(i));
+                boolean sameTypeNode = unresolvedTypeNodeName.equals(currentTypeNodeName);
+                boolean isVisitedNode = visitedNodes.contains(currentTypeNodeName);
+                boolean typeDef =  unresolvedType.getKind() == NodeKind.TYPE_DEFINITION;
```
MaryamZi, line 650: `unresolvedType` is a `TypeDefinition`. Does this ever become false?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -605,70 +609,82 @@ private void defineTypeNodes(List<? extends TypeDefinition> typeDefs, SymbolEnv
...
+
+                if (typeDef && (sameTypeNode || isVisitedNode)) {
+                    BLangTypeDefinition typeDefinition = (BLangTypeDefinition) unresolvedType;
+                    if (fromStructuredType && typeDefinition.getTypeNode().getKind() == NodeKind.UNION_TYPE_NODE) {
```
MaryamZi, line 654: Why do we need this `fromStructuredType` check?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -2255,42 +2263,46 @@ private static boolean hasFillerValue(BType type, List<BType> unanalyzedTypes) {
...
+                    return false;
+                }
+                break;
+            case TypeTags.UNION_TAG:
+                BUnionType unionType = (BUnionType) type;
+                // NIL is a member.
+                if (unionType.isNullable()) {
+                    return true;
+                }
+                // All members are of same type.
```
pubudu91, line 2296: Will this happen at runtime though? IINM, at compile time we get rid of duplicate types. e.g.,
```ballerina
int|int x = 10;
```
In the above union, we'll only have `int` once.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -2255,42 +2263,46 @@ private static boolean hasFillerValue(BType type, List<BType> unanalyzedTypes) {
...
+                Iterator<BType> iterator = unionType.getMemberTypes().iterator();
+                BType firstMember;
+                for (firstMember = iterator.next(); iterator.hasNext(); ) {
+                    if (!isSameType(firstMember, iterator.next())) {
+                        return false;
+                    }
+                }
```
pubudu91, line 2303: Will this happen at runtime though? IINM, at compile time we get rid of duplicate types. e.g.,
```ballerina
int|int x = 10;
```
In the above union, we'll only have `int` once.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
pubudu91: You mean the sym tag? That fix was sort of a temporary fix. Basically now every union type symbol will have the `TYPE_DEF` tag on. Don't feel like it's the right solution. That should only be set for unions which are defined like you've illustrated above IMO.

MaryamZi: No no, what I meant is if we should have a new flag for the type that says this is for a type definition. Instead of a new field in the union type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
irshadnilam: Yes, I came across few cases for Constant Definition.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
irshadnilam: To avoid scenarios, such as the following
```ballerina
type A int|A;
```
and to only allow scenarios where the cyclic is coming from a structural type.
```ballerina
type A int|map<A>
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
pubudu91: Yes, but isn't that what the sym tag of the type symbol do (though it seems to be messed up atm)?

shafreenAnfar:
Closed and re-opened because github workflow has been updated.

shafreenAnfar:
Closed and re-opened because github workflow has been updated.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
MaryamZi: I guess it could happen with unions of finite types with value spaces of the same basic type.

e.g.,
```ballerina
type Foo 1|2;
type Bar 0|4;

public function main() {
    (Foo|Bar)[] x = [];
    x[1] = 1;
}
```

The filler value check for `Foo|Bar` will reach here.

MaryamZi: Came across a separate issue when trying this out btw https://github.com/ballerina-platform/ballerina-lang/issues/24211

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BUnionType.java
```diff
@@ -180,7 +183,20 @@ public int hashCode() {
         return Objects.hash(super.hashCode(), memberTypes);
     }
 
+    public void setMemberTypes(BType[] memberTypes) {
+        this.memberTypes = Arrays.asList(memberTypes);
+        setFlagsBasedOnMembers();
+    }
+
     public boolean isNilable() {
+        if (memberTypes == null) {
```
MaryamZi, line 192: Why do we need this check? Does `isNilable` get called before the union type creation is completed?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
MaryamZi: Shall we use the `Flags.ANONYMOUS` flag for inline unions and not use it for type-definitions, and use that flag to do this differentiation instead of introducing a new field in `BUnionType`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -263,14 +261,18 @@ public boolean isAnydata() {
         if (this.isAnyData.isPresent()) {
             return this.isAnyData.get();
         }
-
+        if (this.resolvingAnyData) {
+            return false;
```
MaryamZi, line 265: Shouldn't we be returning `true` if we are resolving?

Can you check if `type Foo int|map<Foo>` correctly resolves to `anydata`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
irshadnilam: +1, using  Flags.ANONYMOUS

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/types/BUnionType.java
irshadnilam: Yes, `hasFillerValue` is called on `populate` methods, resulting in this call. Having this check prevents the null pointer issue since we are setting the members to null when creation.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
irshadnilam: Yes, it should be true. Fixed.

irshadnilam:
> This example crashes.

This is resolved.

hasithaa:
Closing this PR, since @KRVPerera is working on a fix on top of this PR.

Issue #23468, opened 2020-05-25
Record sub-type checks need to consider the readonly flag


MaryamZi:
**Description:**
$title.

The assignment of `u` to `s` should fail.

```ballerina
type Student record {|
    readonly Details details;
    int id;
|};

type Undergraduate record {|
    readonly & Details details;
    int id;
|};

type Details record {|
    string name;
|};

public function main() {
    
    Undergraduate u = {
        details: {
            name: "Jo"
        },
        id: 1234
    };
    Student s = u;
}
```

MaryamZi:
Needs runtime checks too.

PR #24107, opened 2020-06-14
Consider the requirement of a field to be readonly for record/object subtyping

MaryamZi:
## Purpose
$title - if the field in the target type is a `readonly` field, the relevant field in the source type should also be `readonly` either explicitly or due to the source type itself having its read-only bit set.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23468

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -754,8 +754,7 @@ public class Types {
                 return false;
             }
 
-            if (Symbols.isFlagOn(field.symbol.flags, Flags.READONLY) &&
-                    !Symbols.isFlagOn(sourceMapType.flags, Flags.READONLY)) {
+            if (hasIncompatibleReadOnlyFlags(field.symbol.flags, sourceMapType.flags)) {
                 return false;
             }
 
@@ -765,6 +764,10 @@ public class Types {
         }
 
         return isAssignable(sourceMapType.constraint, targetRecType.restFieldType);
+    }
+
+    private boolean hasIncompatibleReadOnlyFlags(int targetFlags, int sourceFlags) {
+        return Symbols.isFlagOn(targetFlags, Flags.READONLY) && !Symbols.isFlagOn(sourceFlags, Flags.READONLY);
     }
 
     private boolean isErrorTypeAssignable(BErrorType source, BErrorType target, Set<TypePair> unresolvedTypes) {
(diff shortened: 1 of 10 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -750,8 +750,16 @@ private boolean isAssignableMapType(BMapType sourceMapType, BRecordType targetRe
...
 
         for (BField field : targetRecType.fields.values()) {
-            if (!(Symbols.isFlagOn(field.symbol.flags, Flags.OPTIONAL) &&
-                    isAssignable(sourceMapType.constraint, field.type))) {
+            if (!Symbols.isFlagOn(field.symbol.flags, Flags.OPTIONAL)) {
+                return false;
+            }
+
+            if (Symbols.isFlagOn(field.symbol.flags, Flags.READONLY) &&
+                    !Symbols.isFlagOn(sourceMapType.flags, Flags.READONLY)) {
```
pubudu91, line 758: Should we consider extracting this out as a method? Seeing this repeated a couple of times + would probably be easier to read when used in more complex checks

pubudu91:
Is the following disallowed by design?
```ballerina
type Person object {
    string name = "John";
    readonly int age;

    function __init(int age) {
        self.age = age;
    }
};
```
Gives the error
```
error: .::test.bal:13:9: cannot update 'readonly' object field 'age' in 'Person'
```

MaryamZi:
> Is the following disallowed by design?
> 
> ```ballerina
> type Person object {
>     string name = "John";
>     readonly int age;
> 
>     function __init(int age) {
>         self.age = age;
>     }
> };
> ```
> 
> Gives the error
> 
> ```
> error: .::test.bal:13:9: cannot update 'readonly' object field 'age' in 'Person'
> ```

I assume you mean to set the value in the initializer method? If so it has to be `init` right?

The following doesn't give any errors.

```ballerina
type Person object {
    string name = "John";
    readonly int age;

    function init(int age) {
        self.age = age;
    }
};
```

The error in your sample is expected, because we don't allow setting values for object fields in methods other than in the `init`.

pubudu91:
Argh.. Forgot we changed it to `init` :sweat_smile: Even in the `ini()`, should only be allowed to set it once right?
```ballerina
type Person object {
    string name = "John";
    readonly int age;

    function init(int age) {
        self.age = age;
        self.name = "foo";
        self.age = age + 20;
    }
};
```
This is probably not related to this PR though

MaryamZi:
> Argh.. Forgot we changed it to `init`  Even in the `ini()`, should only be allowed to set it once right?
> 
> ```ballerina
> type Person object {
>     string name = "John";
>     readonly int age;
> 
>     function init(int age) {
>         self.age = age;
>         self.name = "foo";
>         self.age = age + 20;
>     }
> };
> ```
> 

Yeah, I believe the plan is to have this be an error. This is not yet implemented. https://github.com/ballerina-platform/ballerina-spec/issues/532 is the related spec issue and I just created https://github.com/ballerina-platform/ballerina-lang/issues/24119 to track fixing the implementation.

PR #23473, opened 2020-05-25
Restrict annotation value to true, map<anydata|readonly> , map<anydata|readonly>[]

KRVPerera:
## Purpose
Fixes #15533

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #24179, opened 2020-06-16
Fix code suggestion for PR #23473

### LINKED ISSUE #15533: [Annotations] Restrict constraints of maps/fields of records to be anydata and freeze the values
**Description:**
According to the v2019R1 spec, 

```
annotation-decl :=
   metadata
   [public] [const] annotation [type-descriptor] annot-tag 
   [on annot-attach-points]
annot-tag := identifier

The type-descriptor specifies the type of the annotation tag. The type must be 
a subtype of one of the following three types: true, map<anydata>, map<anydata>[].

If the annotation-decl for a tag specifies const, then a mapping-constructor-expr 
in annotations with that tag must be a const-expr and is evaluated at compile-time 
with the semantics of a const-expr. Otherwise, the mapping-constructor-expr is 
evaluated when the annotation is evaluated and the resulting value is automatically 
frozen.
```

In the initial iteration to introduce annotation changes the following is not done:
- restricting the constraints of maps/fields of records to be `anydata` - cannot be done at the moment since annotations currently have fields of types `service` (websocket), `function` (transactions), etc.
- freezing the annot value - cannot be done since the value may have non-anydata values

**Related Issues:**
https://github.com/ballerina-platform/ballerina-spec/issues/70

KRVPerera:
Fixes #15533
https://github.com/ballerina-platform/ballerina-lang/pull/23473#discussion_r439937146

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -626,11 +627,16 @@ public void visit(BLangAnnotationAttachment astAnnotAttach) {
...
 
     private boolean isCompileTimeAnnotationValue(BLangExpression expr) {
-        // TODO Compile time literal constants
         switch (expr.getKind()) {
             case LITERAL:
             case NUMERIC_LITERAL:
                 return true;
+            case INVOCATION:
+                BLangInvocation invocation = (BLangInvocation) expr;
+                if (invocation.name.getValue().equals(CLONE_READ_ONLY)) {
```
MaryamZi, line 636: Thinking about this a little more, currently an invocation is not a compile-time-known value right?

So I think we should differentiate between the `.cloneReadOnly` we add, and what the user may have added for a field.

Shall we allow this only when this is the attachment itself, and return false for recursive checks?

KRVPerera:
Closing as will be introduced changes by original PR:https://github.com/ballerina-platform/ballerina-lang/pull/23473

PR #24189, opened 2020-06-16
Extend JWT validator configuration [master]

ldclakmal:
## Purpose
This PR updates the signature validation fields of `JwtValidatorConfg`.

The current record has `JwtTrustStoreConfig|JwksConfig signatureConfig?` field which validates the signature either from one of these. But with the new config, user can configure either one of these or both. Then the signature can be validated depending on the `kid` property of the provided/inbound JWT. With this change, we only need to update the field name from `signatureConfig` to `jwksConfig` if using JWK for signature validation. Also, this will remove the breaking changes from 1.2.x to latest.

This is an extension of PR https://github.com/ballerina-platform/ballerina-lang/pull/22982 which provided the JWT signature validation support with JWKs.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/24190

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


stdlib/jwt/src/main/ballerina/src/jwt/jwt_validator.bal
```diff
@@ -270,18 +272,21 @@ function validateJwtRecords(string jwt, JwtHeader jwtHeader, JwtPayload jwtPaylo
...
+            _ = check validateSignatureByJwks(jwt, kid, alg, jwksConfig);
         } else {
             return prepareError("Key ID (kid) is not provided in JOSE header.");
         }
     }
 
+    JwtTrustStoreConfig? trustStoreConfig = config?.trustStoreConfig;
+    if (trustStoreConfig is JwtTrustStoreConfig) {
+        _ = check validateSignatureByTrustStore(jwt, alg, trustStoreConfig);
+    }
```
daneshk, line 288: I think we need to move this as else if statement

ldclakmal: Yes. Otherwise the signature can be validated twice. Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/24189/commits/dba0b304c31a8f674e20a08c8736638925d219c8

shafreenAnfar:
Closed and re-opened because github workflow has been updated.

stdlib/jwt/src/main/ballerina/src/jwt/jwt_validator.bal
daneshk: Shall we check the kid first, code looks like, 
```ballerina
if (kid is string) {
  if (jwksConfig is JwksConfig) {
    // validate
  } else if (trustStoreConfig is JwtTrustStoreConfig) {
   // validate
  } else {
    // error 
 }
} else {
 if (trustStoreConfig is JwtTrustStoreConfig) {
   // validate
  } else {
    // error 
 }
}
```
current logic will some cases, I think

ldclakmal: Actually, there is a use case to pass the signature validation if, none of the `JwksConfig` or `JwtTrustStoreConfig` are provided. And there is another use case where kid is not given, but only `JwksConfig` is provided, so that we have to return an error message. Considering all, I updated the logic as
```ballerina
if (jwksConfig is JwksConfig) {
    if (kid is string) {
        // validate
    } else if (trustStoreConfig is JwtTrustStoreConfig) {
       // validate
    } else {
        // return error
    }
} else if (trustStoreConfig is JwtTrustStoreConfig) {
    // validate
}
// continue without signature validation
```
WDYT?

ldclakmal: Added following test cases which covers all the possibilities.
``` 
  * Test cases for JWT signature validation with TrustStore and JWK based authentication/authorization scenarios.
  *
  * The following 8 scenarios are tested here.
  * T (TRUE) - Parameter/Config is available.
  * F (FALSE) - Parameter/Config is not available.
  *
  * | Scenario |  kid  | JwksConfig | TrustStoreConfig | Test Status |
  * |----------|-------|------------|------------------|-------------|
  * |   1      |   T   |     T      |         T        |    PASS     |
  * |   2      |   T   |     T      |         F        |    PASS     |
  * |   3      |   T   |     F      |         T        |    PASS     |
  * |   4      |   T   |     F      |         F        |    PASS     |
  * |   5      |   F   |     T      |         T        |    PASS     |
  * |   6      |   F   |     T      |         F        |    FAIL     |
  * |   7      |   F   |     F      |         T        |    PASS     |
  * |   8      |   F   |     F      |         F        |    PASS     |
  *
  * Additionally, there are 4 more scenarios where kid is available, but it is invalid.
  * |   9      |   T   |     T      |         T        |    FAIL     |
  * |  10      |   T   |     T      |         F        |    FAIL     |
  * |  11      |   T   |     F      |         T        |    PASS     |
  * |  12      |   T   |     F      |         F        |    PASS     |
```

Issue #24341, opened 2020-06-18
Warnings are getting printed for some Query BBEs


mohanvive:
**Description:**
Below warning is getting printed for query expression related below BBEs.

- Query Expression
- Stream From Query Expression
- String From Query Expression
- XML From Query Expression
- Join Iterables

```ballerina
warning: .::query_expression_to_stream.bal:1:1: Error while generating json AST for query_expression_to_stream.bal. Error occurred while generating JSON
```

Observed, ```ballerina: Oh no, something really went wrong. Bad. Sad.``` for below query expression BBE.

- Table From Query Expression

mohanvive:
@grainier Please followup.

sameerajayasoma:
Do you have the full stack trace? Please attach the ballerina-internal.log file as well.

mohanvive:
There is no much stack trace for cases which generated the warning (or no any logs in ballerina-internal.log). Just the below one.

```
Compiling source
	query_expression.bal
warning: .::query_expression.bal:1:1: Error while generating json AST for query_expression.bal. Error occurred while generating JSON
Running executables

name=Alex degree=Bachelor of Medicine expectedGradYear=2025
name=John degree=Bachelor of Medicine expectedGradYear=2027
```

For `ballerina: Oh no, something really went wrong. Bad. Sad.` error case, it generated below logs in ballerina-internal.log. 

```
[2020-06-19 02:02:08,154] SEVERE {b7a.log.crash} - io.ballerinalang.compiler.syntax.tree.SimpleNameReferenceNode cannot be cast to io.ballerinalang.compiler.syntax.tree.Token 
java.lang.ClassCastException: io.ballerinalang.compiler.syntax.tree.SimpleNameReferenceNode cannot be cast to io.ballerinalang.compiler.syntax.tree.Token
	at io.ballerinalang.compiler.syntax.tree.RecordFieldNode.fieldName(RecordFieldNode.java:49)
	at org.wso2.ballerinalang.compiler.parser.BLangNodeTransformer.transform(BLangNodeTransformer.java:1063)
	at org.wso2.ballerinalang.compiler.parser.BLangNodeTransformer.transform(BLangNodeTransformer.java:405)
	at io.ballerinalang.compiler.syntax.tree.RecordFieldNode.apply(RecordFieldNode.java:67)
	at org.wso2.ballerinalang.compiler.parser.BLangNodeTransformer.transform(BLangNodeTransformer.java:1011)
[...]

PR #24346, opened 2020-06-19
Fix NPE thrown when generating AST json fails

nadundesilva:

### DIFF
--- a/misc/observerability-symbol-collector/src/main/java/org/ballerinalang/observability/anaylze/DefaultObservabilitySymbolCollector.java
+++ b/misc/observerability-symbol-collector/src/main/java/org/ballerinalang/observability/anaylze/DefaultObservabilitySymbolCollector.java
@@ -150,7 +150,9 @@ public class DefaultObservabilitySymbolCollector implements ObservabilitySymbolC
         List<BLangCompilationUnit> compilationUnits = module.getCompilationUnits();
         for (BLangCompilationUnit cUnit : compilationUnits) {
             CUnitASTHolder jsonCUnit = getCUnitASTHolder(visibleEPsByNode, cUnit);
-            pkgASTHolder.addCompilationUnit(cUnit.name, jsonCUnit);
+            if (jsonCUnit.getAst() != null) {
+                pkgASTHolder.addCompilationUnit(cUnit.name, jsonCUnit);
+            }
         }
         return pkgASTHolder;
     }


misc/observerability-symbol-collector/src/main/java/org/ballerinalang/observability/anaylze/DefaultObservabilitySymbolCollector.java
```diff
@@ -150,7 +150,9 @@ private PkgASTHolder getModuleASTHolder(BLangPackage module) {
         List<BLangCompilationUnit> compilationUnits = module.getCompilationUnits();
         for (BLangCompilationUnit cUnit : compilationUnits) {
             CUnitASTHolder jsonCUnit = getCUnitASTHolder(visibleEPsByNode, cUnit);
-            pkgASTHolder.addCompilationUnit(cUnit.name, jsonCUnit);
+            if (jsonCUnit.getAst() != null) {
```
a5anka, line 153: I am bit skeptical  on this null check. When can `jsonCUnit.getAst()` be null?

nadundesilva: When an exception is thrown while generating the AST JSON (while calling `TextDocumentFormatUtil.generateJSON`), jsonCUnit.getAst() can be null. We observed this in several scenarios. The warning in https://github.com/ballerina-platform/ballerina-lang/blob/master/misc/observerability-symbol-collector/src/main/java/org/ballerinalang/observability/anaylze/DefaultObservabilitySymbolCollector.java#L181 gets printed and AST is null.

Please refer  #24341

nadundesilva: Ideally Formatter shouldn't throw an error, but this is just to be sure that this doesn't break the compilation itself.

The cases in #24341 were throwing a Stackoverflow exception due to a bug (This bug had been fixed)

a5anka: It is better to show errors than hide it. IMO what we should do is provide a way to disable AST generation.

nadundesilva:
Closing as per the above comment.

PR #24360, opened 2020-06-19
Add compiler plugin crash reseliency

rdhananjaya:
## Purpose
$subject.

Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/resources/compiler.properties
+++ b/compiler/ballerina-lang/src/main/resources/compiler.properties
@@ -1406,3 +1406,6 @@ error.methods.not.allowed=\
 
 error.distinct.typing.only.support.objects.and.errors=\
   distinct typing is only supported for object type and error type
+
+warning.compiler.plugin.crashed=\
+  compiler plugin crashed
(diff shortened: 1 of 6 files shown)


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1406,3 +1406,6 @@ error.methods.not.allowed=\
 
 error.distinct.typing.only.support.objects.and.errors=\
   distinct typing is only supported for object type and error type
+
+warning.compiler.plugin.crashed=\
+  compiler plugin crashed
```
irshadnilam, line 1411: Is there a way to identify a plugin (a name perhaps)?
If so, can we print? 
`compiler plugin crashed: xyz plugin`

rdhananjaya: I'm afraid compiler plugins at the moment doesn't have anything like that.
The only way to see which plugin failed it to check the ballerina-internal-log.

We can parse the stack trace to detect which compiler plugin crashed. I think we can do something like that.

PR #24407, opened 2020-06-22
 Fix conditional expressions in all places

### LINKED ISSUE #24294: Fix conditional expressions in all places
When there is `expr1 ? ident : expr2`, parser needs to termination without treating `ident:` as a qualified identifier.

Currently this scenario works. However, this middle-expr can be many other things as well such as `check ident`, `<T> ident`, etc. So we need to fix this for any expression that can ends with a n identifier.

Some of the issues were fixed in https://github.com/ballerina-platform/ballerina-lang/pull/24279/commits/63c5f08e8bfb1264361cf96f78769034492de40d. But need to find and fix remaining ones.

lochana-chathura:
## Purpose
$subject.

Fixes #24294 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
+++ b/compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
@@ -1964,7 +1964,7 @@ public class BallerinaParser extends AbstractParser {
 
     private STNode parseParamGivenAnnotsAndQualifier(SyntaxKind prevParamKind, STNode leadingComma, STNode annots,
                                                      STNode qualifier, boolean isParamNameOptional) {
-        STNode type = parseTypeDescriptor(ParserRuleContext.TYPE_DESC_BEFORE_IDENTIFIER, false);
+        STNode type = parseTypeDescriptor(ParserRuleContext.TYPE_DESC_BEFORE_IDENTIFIER);
         STNode param = parseAfterParamType(prevParamKind, leadingComma, annots, qualifier, type, isParamNameOptional);
         endContext();
         return param;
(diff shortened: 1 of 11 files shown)


lochana-chathura:
Most of the changes are from last commit (https://github.com/ballerina-platform/ballerina-lang/pull/24407/commits/a76b8aa955d8f4a5a188a7aded72b5eec3ef5b05) as type-descriptor is referred in many places.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -2319,7 +2327,7 @@ private STNode parseTypeDescriptorInternal(SyntaxKind tokenKind, ParserRuleConte
                     return solution.recoveredNode;
                 }
 
-                return parseTypeDescriptorInternal(solution.tokenKind, context);
+                return parseTypeDescriptorInternal(solution.tokenKind, context, false);
```
SupunS, line 2330: Shouldn't we pass `isInConditionalExpr` as the last arg?

lochana-chathura: yh It should be. thanks. Fixed with bbc37f6. 
btw just found out that parser crashes for
```
function main() {
    int x = a?b:c
}
```
missing semicolon in conditional-expr. I'll have a look

lochana-chathura: Latter issue fixed with aecf63b

Issue #23803, opened 2020-06-08
Error cannot be trap when panic inside foreach built in function


warunalakshitha:
**Description:**
Following code snippet can be used to reproduce the issue.

```ballerina

import ballerina/io;

public function main() {
    var e = trap foo();
    io:println("***********");
    io:println(e);
}


function foo() {
    int[] values = [5, -1, 4];
    values.forEach(function (int val) {
        if( val < 0) {
            panic error("value: " + val.toString() + " is negative.");
        }
    });
}
```

program exit when error happen. Ideally panic should be trapped and program should continue.

PR #24413, opened 2020-06-22
[Runtime] Fix errors cannot be trap when panic inside foreach built in function

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23803

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/scheduling/Strand.java
```diff
@@ -52,6 +53,7 @@
     public Object[] frames;
     public int resumeIndex;
     public Object returnValue;
```
riyafa, line 55: why do we have it as public? I think it is better to have it as private and access it via a method call. It will hide the fields in the object and future modification would be easier

warunalakshitha: We use these fields in codegen as well. So we need to have them public. If not codegen side need to change with private methods but that will be a considerable change.

Issue #23702, opened 2020-06-03
Re-enable tests disabled with `brokenOnErrorChange`


rdhananjaya:
**Description:**
These are mostly negative test cases.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #24448, opened 2020-06-25
Remove `brokenOnErrorChange` test group

rdhananjaya:
## Purpose
$subject

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23702

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/checkedexpr/CheckedExpressionOperatorTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/checkedexpr/CheckedExpressionOperatorTest.java
@@ -249,7 +249,7 @@ public class CheckedExpressionOperatorTest {
     }
 
     @Test(description = "Test service resource that returns an error containing check expression",
-            groups = "brokenOnErrorChange")
+            groups = {"disableOnOldParser", "brokenOnNewParser"})
     public void testSemanticErrorsWithResources() {
         CompileResult compile = BCompileUtil.compile(
                 "test-src/expressions/checkedexpr/checked_expr_within_resource.bal");
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/matchstmt/MatchStructuredErrorPatternsTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/matchstmt/MatchStructuredErrorPatternsTest.java
@@ -34,7 +34,7 @@ import org.testng.annotations.Test;
  *
  * @since 0.990.4
  */
-@Test(groups = "brokenOnErrorChange")
+@Test(groups = {"disableOnOldParser", "brokenOnNewParser"})
 public class MatchStructuredErrorPatternsTest {
     private CompileResult result, resultNegative;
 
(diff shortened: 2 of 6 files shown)


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/checkedexpr/CheckedExpressionOperatorTest.java
```diff
@@ -249,7 +249,7 @@ public void testCheckInBinaryLTExpression() {
     }
 
     @Test(description = "Test service resource that returns an error containing check expression",
-            groups = "brokenOnErrorChange")
+            groups = {"disableOnOldParser", "brokenOnNewParser"})
```
MaryamZi, line 252: Is this test still broken? The expectation is to not have compilation errors right?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/matchstmt/MatchStructuredErrorPatternsTest.java
```diff
@@ -34,7 +34,7 @@
  *
  * @since 0.990.4
  */
-@Test(groups = "brokenOnErrorChange")
+@Test(groups = {"disableOnOldParser", "brokenOnNewParser"})
```
MaryamZi, line 37: Wouldn't continuing to have the "brokenOnErrorChange" group be useful in identifying when/why exactly the test was disabled, and follow up on enabling the test? 

The "disableOnOldParser" group will be irrelevant when the old parser is removed, and just the fixes from the new parser wouldn't fix these tests right?

rdhananjaya: It won't work by just fixing the parser, but the parser team don't have a way to know that parser is broken for error match pattern

MaryamZi: That is being tracked separately by the parser team right? 

Can we maybe fix these tests so that they will work once the error match support is added in the new parser (if there are no deviations)? If that is possible, I guess we can even run these tests with the old parser?

If that is not possible, can we maybe have three groups
```java
@Test(groups = {"disableOnOldParser", "brokenOnNewParser", "brokenOnErrorChange"})
```

My concern is because once the "disableOnOldParser" group is removed, it would seem like this is broken due to parser incompatibilities even though there are other issues.

rdhananjaya: IMO effort required to fix this in old parser doesn't worth it.

The sole purpose of this PR is to remove `brokenOnErrorChange` flag so that there is no confusion for the parser team. If they are separately tracking what to fix then I'm completly OK closing this PR.

MaryamZi: > IMO effort required to fix this in old parser doesn't worth it.
> 

Oh okay, I thought it was just changes related to the error type descriptor that was required. Yeah, wouldn't be worth the effort if the changes we do to fix them become obsolete with the new parser.

> The sole purpose of this PR is to remove `brokenOnErrorChange` flag so that there is no confusion for the parser team. If they are separately tracking what to fix then I'm completly OK closing this PR.

Can we not keep all 3 groups? That way it would be straight-forward for both the parser team and the core team to identify tests that need to be fixed/upon which there's an impact?

rdhananjaya:
Closing this PR as we no longer need to change the test groups given the progress of new parser

PR #24484, opened 2020-06-29
Add support for `readonly` fields in a mapping constructor expression and allow using `readonly` as the CET for mapping/list constructor expressions

### LINKED ISSUE #22954: Allow constructing values with their read-only bit on with a type-cast-expr `<readonly>`

### LINKED ISSUE #23663: Support specifying readonly fields in the mapping constructor expression
**Description:**
$title.

See https://github.com/ballerina-platform/ballerina-spec/issues/493

```
mapping-constructor-expr := { [field (, field)*] }
field := 
  specific-field
  | computed-name-field
  | spread-field
specific-field :=
   [readonly] (field-name | string-literal) : value-expr
   | [readonly] [variable-name]
value-expr := expression
computed-name-field := [ field-name-expr ] : value-expr
field-name-expr := expression
spread-field := ... expression
```

Grammar changes have already been added in https://github.com/ballerina-platform/ballerina-lang/pull/22519/files#diff-0e72b59e6b830c29e9c6161f3bb4177aR358

### LINKED ISSUE #24437: Validate type-checking checks with intersection types
**Description:**
With the introduction of the `BIntersectionType` some checks that specifically check for a type or a union type may fail.

For example, the `null` literal cannot be used with `json & readonly` atm.

```ballerina
public function main() {
    readonly & json x = null;
}
```

Need to fix all usages and maybe check for the possibility of improving the same.

**Affected Versions:**
slp1

### LINKED ISSUE #24466: Issue in error messages with `readonly` fields in anonymous record/objects
**Description:**
$title. The `readonly` modifier is not printed.

**Steps to reproduce:**
```ballerina
public function main() {
    record {|
        int i;
        string s;
    |} a = {i: 1, s: "hello"};

    record {|
        readonly int i;
        string s;
    |} x = a; // error

    object {
        int i = 2;
        string s = "world";
    } b = new;

    object {
        readonly int i = 2;
        string s = "world";
    } y = b; // error
}
```

```cmd
$ ballerina run qwe.bal 
Compiling source
        qwe.bal
error: .::qwe.bal:11:12: incompatible types: expected 'record {| int i; string s; |}', found 'record {| int i; string s; |}'
error: .::qwe.bal:21:11: incompatible types: expected 'object { int i; string s; }', found 'object { int i; string s; }'
```

**Affected Versions:**
slp1

### LINKED ISSUE #24472: Support `readonly` as the contextually expected type with a constructor expression
**Description:**
See https://github.com/ballerina-platform/ballerina-spec/issues/554

### LINKED ISSUE #24474: Bad, sad error at runtime when an anonymous record has a rest field which again is of an anonymous type
**Description:**
$title. Seems to be due to a type definition not getting added for the inner anonymous type (which we now do at desugar).

**Steps to reproduce:**
```ballerina
public function main() {
    any a = 1;
    boolean b = a is record {| record {| int i; |}...; |};
}
```

ballerina-internal.log
```log
[2020-06-28 12:35:50,914] SEVERE {b7a.log.crash} - $type$_anonType_1 
java.lang.NoSuchFieldError: $type$_anonType_1
	at ___init.$populate$type$_anonType_0(.)
	at ___init.$createTypes(.)
	at ___init.$currentModuleInit(.)
	at ___init.___init_(.)
	at ___init.$moduleInit(.)
	at ___init.$lambda$$moduleInit$(.)
	at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:486)
	at org.ballerinalang.jvm.scheduling.Scheduler.run(Scheduler.java:267)
	at org.ballerinalang.jvm.scheduling.Scheduler.runSafely(Scheduler.java:235)
```

**Affected Versions:**
1.2.4, slp1

MaryamZi:
## Purpose

This PR
- Adds support for `readonly` fields in a mapping-constructor expression. Fixes #23663

Such a mapping constructor will create a value of a type that belongs to the applicable contextually expected type, but additionally, the fields with which `readonly` were used will be `readonly` fields.

The constraints this enforce on the value are the same as with `readonly` fields in the type-descriptor. i.e.,
1. the values specified for such fields have to be immutable
2. the field cannot be updated once the value is created
3. if the constructor creates a closed record and all the fields are now `readonly` the resultant type will also be an readonly type

- Allows using `readonly` as the CET for mapping/list constructor expressions. Fixes #24472

The type for such a value is inferred as done with `var`, but in addition the value created will have its read-only bit set.

- Allows constructing values with their read-only bit on with a type-cast-expr `<readonly>` Fixes #22954

This PR also 
fixes #24466 - Issue in error messages with `readonly` fields in anonymous record/objects
fixes #24474 - Bad, sad error at runtime when an anonymous record has a rest field which again is of an anonymous type
fixes #24437 - Improves type-checking checks with intersection types by using the effective type when visiting an expression

## Samples

```ballerina
type Employee record {|
    Details details;
    string department;
|};

type Details record {|
    string name;
    int id;
|};

public function main() {
    Employee emp = {
        readonly details: {
            name: "May",
            id: 112233
        },
        department: "IT"
    };
}
```

Since `readonly` is specified with `details` when creating `emp`, the particular `Details` value is created as an immutable value.
```ballerina
io:println(emp.isReadOnly()); // false
io:println(emp.details.isReadOnly()); // true
```

The effective type of `emp` is now a subtype of `Employee` that mandates the `details` field to be `readonly`.
```ballerina
io:println(emp is record {|readonly Details details; string department;|}); // true
```

Since the field itself will also be marked as a `readonly` field, attempting to change the `details` field itself will panic.

```ballerina
emp.details = { // {ballerina/lang.map}InherentTypeViolation message=cannot update 'readonly' field 'details' in record of type '$anonType$0'
    name: "Jo",
    id: 4324
};
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -1998,7 +1998,14 @@ public void visit(BLangRecordLiteral recordLiteral) {
 
         Set<Object> names = new HashSet<>();
         BType type = recordLiteral.type;
-        boolean isOpenRecord = type != null && type.tag == TypeTags.RECORD && !((BRecordType) type).sealed;
+        boolean isRecord = type != null && type.tag == TypeTags.RECORD;
+        boolean isOpenRecord = isRecord && !((BRecordType) type).sealed;
+
+        // A record type is inferred for a record literal even if the CET is a map, if the mapping
```
KRVPerera, line 2004:
```suggestion
        // A record type is inferred for a record literal even if the CET (contextually expected type) is a map, if 
        // the mapping constructor expression has `readonly` fields.
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -1998,7 +1998,14 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+        // constructor expression has `readonly` fields.
```
KRVPerera, line 2005:
```suggestion
        // A record type is inferred for a record literal even if the CET (contextually expected type) is a map, if 
        // the mapping constructor expression has `readonly` fields.
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -360,16 +348,35 @@ public BType checkExpr(BLangExpression expr, SymbolEnv env, BType expType, Diagn
...
         this.expType = preExpType;
         this.diagCode = preDiagCode;
-        if (resultType.tag != TypeTags.SEMANTIC_ERROR) {
+
+        validateAndSetExprExpectedType(expr, expType);
+
+        return resultType;
+    }
+
+    private void validateAndSetExprExpectedType(BLangExpression expr, BType origExpType) {
```
KRVPerera, line 366: `origExpType` type is not used.

MaryamZi: Thanks, fixed in 290649c.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
MaryamZi: Fixed in 290649c.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2020,7 +2020,14 @@ public void visit(BLangRecordLiteral recordLiteral) {
 
         Set<Object> names = new HashSet<>();
         BType type = recordLiteral.type;
-        boolean isOpenRecord = type != null && type.tag == TypeTags.RECORD && !((BRecordType) type).sealed;
+        boolean isRecord = type != null && type.tag == TypeTags.RECORD;
```
pubudu91, line 2023: Not related to this PR, but there shouldn't be a case where the type of a record literal is `null` in code analyzer right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2084,6 +2091,10 @@ public void visit(BLangRecordLiteral recordLiteral) {
                 }
             }
         }
+
+        if (isInferredRecordForMapCET) {
+            recordLiteral.expectedType = type;
```
pubudu91, line 2096: Why is the expected type changed here?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
MaryamZi: Yeah, ideally should never be null.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -360,16 +348,35 @@ public BType checkExpr(BLangExpression expr, SymbolEnv env, BType expType, Diagn
...
+        if (expr.getKind() != NodeKind.RECORD_LITERAL_EXPR ||
+                expr.expectedType == null ||
+                expr.expectedType.tag != TypeTags.MAP ||
+                expr.type.tag != TypeTags.RECORD) {
             expr.expectedType = resultType;
         }
-        return resultType;
+
+        // If the expected type is a map, but a record type is inferred due to the presence of `readonly` fields in
+        // the mapping constructor expression, we don't override the expected type.
```
pubudu91, line 379: Shall we move this above line 371? When placed here, it looks like a dangling comment

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
MaryamZi: We don't log the error at L2079 if a record type is inferred for a mapping constructor expression when the contextually expected type is a map type.

e.g.,
```ballerina
map<int> m = {
   readonly a: 1,
   b: 2
};
```

Here the actual type of the constructed value is `record {| readonly int a; int...; |}`. Usually you'll have to use a literal key to provide for the rest field (`b` in the sample), but that is not required when a record type is inferred for a map ACET.

To do this, we need to know here what the original applicable expected type was.

Usually, the `expectedType` field of an expression is overridden in `checkExpr`, but for this particular scenario we don't do it because we need to know the exact ACET - see https://github.com/ballerina-platform/ballerina-lang/blob/ed729d2aa3f90d27522bb8d1f867a4ee61da6cd8/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java#L378

I don't think this is required, but just for uniformity we set the `expectedType` as the normal path here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -360,16 +348,35 @@ public BType checkExpr(BLangExpression expr, SymbolEnv env, BType expType, Diagn
...
+
+    private void validateAndSetExprExpectedType(BLangExpression expr) {
+        if (resultType.tag == TypeTags.SEMANTIC_ERROR) {
+            return;
+        }
+
+        if (expr.getKind() != NodeKind.RECORD_LITERAL_EXPR ||
+                expr.expectedType == null ||
+                expr.expectedType.tag != TypeTags.MAP ||
+                expr.type.tag != TypeTags.RECORD) {
```
pubudu91, line 374: Wouldn't the inverse of this check read better?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1481,51 +1497,193 @@ private boolean exprIncompatible(BType eType, BLangExpression expr) {
...
+                    name = (String) ((BLangLiteral) keyExpr).value;
+                }
+            } else {
+                BLangRecordVarNameField varNameField = (BLangRecordVarNameField) field;
+
+                if (!varNameField.readonly) {
+                    continue;
+                }
+                name = varNameField.variableName.value;
+            }
```
pubudu91, line 1591: I think we should consider providing common info such as the name, type of a field through an API. Would simplify code like this. Again, I mean as a future improvement.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1481,51 +1497,193 @@ private boolean exprIncompatible(BType eType, BLangExpression expr) {
...
+            String key = readOnlyEntry.getKey();
+            Name fieldName = names.fromString(key);
+
+            BType readOnlyFieldType;
+            if (field.isKeyValueField()) {
+                readOnlyFieldType = ((BLangRecordKeyValueField) field).valueExpr.type;
+            } else {
+                // Has to be a varname field.
+                readOnlyFieldType = ((BLangRecordVarNameField) field).type;
+            }
```
pubudu91, line 1622: Another example of what I mentioned above.

pubudu91:
Shouldn't the following be allowed?
```ballerina
map<anydata> m = {foo: <readonly>"bar"};
```
Gives,
```
error: .::test2.bal:14:28: incompatible types: expected 'anydata', found 'readonly'
```
28 is the start of the type cast expr

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: I actually found this more readable :smiley: But when swapping the comment, I changed it to return early if it matches the inverse, so it's kinda swapped now.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Yeah, +1.

MaryamZi:
> Shouldn't the following be allowed?
> 
> ```ballerina
> map<anydata> m = {foo: <readonly>"bar"};
> ```
> 
> Gives,
> 
> ```
> error: .::test2.bal:14:28: incompatible types: expected 'anydata', found 'readonly'
> ```
> 
> 28 is the start of the type cast expr
> 
> The following works:
> 
> ```ballerina
> map<anydata> m = {foo: <readonly>{"bar":"baz"}};
> ```

The spec says

> Note that a type-cast-expr of <readonly\> can be used both to cause constructors within the expression to construct values with the read-only bit on and to verify that the value resulting from the evaluation of expression has its read-only bit on.

In your second example, when `<readonly>` is used with a constructor expression, the applicable contextually expected type for the constructor expression thus becomes `anydata & readonly`.

In the first example, it's just a cast. The static type of `<readonly>"bar"` should be the intersection of `readonly` and `string`, which is `string`, and your example should be allowed. But there's a deviation with the type-cast-expr where the static type is not the intersection - see https://github.com/ballerina-platform/ballerina-lang/issues/24559. Once that is fixed, this would work.

PR #24499, opened 2020-06-30
Add project api to compiler context

### LINKED ISSUE #24441: Need a flag to indicate a packageNode is from the source project
**Description:**
In the following code snippet from the DataMapper compiler plugin we need a way of identifying whether the package nodes we encounter while traversing the AST are from the same source project. The package nodes can be either from the bal files located on the src folder of a particular Ballerina project or from the balos imported by the project.

One solution for the above problem would be to create a flag/function API call in the packageNode to decide whether it is from the source project or from an imported balo file. 

```
    @Override
    public void process(PackageNode packageNode) {
        for (TopLevelNode node : ((BLangPackage) packageNode).topLevelNodes) {
            if (node instanceof BLangTypeDefinition) {
                if ((((BLangTypeDefinition) node).getTypeNode().type.flags & Flags.CLIENT) == Flags.CLIENT) {
                    clientFlag = true;
                    break;
                }
            }
        }

        if (clientFlag) {
            String pkgRoot = ((BLangPackage) packageNode).repos.resolve(((BLangPackage) packageNode).packageID).inputs.
                    get(0).toString();
            projectSourceFolder = pkgRoot.substring(0, pkgRoot.lastIndexOf("/"));
```

**Steps to reproduce:**
N/A

**Affected Versions:**
Ballerina Swan Lake Preview 1

N/A

N/A

N/A

N/A

hevayo:
## Purpose
Add a Project API to compiler context. This can be use by compiler plugins and other components to retrieve information about the project.

Fixes #24441

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/project/Project.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/project/Project.java
@@ -17,7 +17,6 @@
  */
 package org.ballerinalang.project;
 
-import org.ballerinalang.compiler.JarResolver;
 import org.ballerinalang.model.elements.PackageID;
 import org.wso2.ballerinalang.compiler.util.CompilerContext;
 
@@ -33,7 +32,7 @@ public interface Project {
     CompilerContext.Key<Project> PROJECT_KEY = new CompilerContext.Key<>();
 
     /**
-     * Return true is module exists in the project.
+     * Returns true if the module exists in the project.
      *
      * @param moduleId Module Id
      * @return
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/project/Project.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/project/Project.java
@@ -1,0 +1,50 @@
+/*
+ *  Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  WSO2 Inc. licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.ballerinalang.project;
+
+import org.ballerinalang.compiler.JarResolver;
+import org.ballerinalang.model.elements.PackageID;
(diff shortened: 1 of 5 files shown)


compiler/ballerina-lang/src/main/java/org/ballerinalang/project/Project.java
```diff
@@ -0,0 +1,50 @@
...
+ * Defines functionality of the Project.
+ *
+ * @since 2.0.0
+ */
+public interface Project {
+
+    CompilerContext.Key<Project> PROJECT_KEY = new CompilerContext.Key<>();
+
+    /**
+     * Return true is module exists in the project.
```
miyurud, line 36: "Return true is module exists in the project." has to be changed as "Returns true if the module exists in the project."

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/SourceDirectoryManager.java
```diff
@@ -62,10 +68,15 @@ public static SourceDirectoryManager getInstance(CompilerContext context) {
...
         context.put(PROJECT_DIR_KEY, this);
+        // We will add the source directory manager as the project API to the context
+        // This has to be further refactored with ProjectAPI Implementation
+        context.put(PROJECT_KEY, this);
 
         this.names = Names.getInstance(context);
         this.options = CompilerOptions.getInstance(context);
         this.sourceDirectory = initializeAndGetSourceDirectory(context);
+        // load manifest only once
+        this.manifest = getManifest();
```
hasithaa, line 79: getManifest() logic can be improved to load manifest only once.

compiler/ballerina-lang/src/main/java/org/ballerinalang/project/Project.java
```diff
@@ -0,0 +1,50 @@
...
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.ballerinalang.project;
+
+import org.ballerinalang.compiler.JarResolver;
```
miyurud, line 20: It seems even though "org.ballerinalang.compiler.JarResolver" is imported within this class its never used.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/SourceDirectoryManager.java
hevayo: Created #24516 to track the issue. 

I will be merging this PR without this refactor since that change will touch many places. Will address it with the new Project API

Issue #24529, opened 2020-07-01
Interop doesn't work with intersection types


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/java;

function getInstance() returns map<string> & readonly = @java:Method {
    class: "org.maryam.Test"
} external;
```

```java
public static MapValueImpl<BString, BString> getInstance() {
    return new MapValueImpl<BString, BString>(new BMapType(BTypes.typeString, true));
}
```

Fails with
```cmd
$ ballerina build foo

Compiling source
        maryam/foo:0.1.0
error: maryam/foo:0.1.0::main.bal:3:1: {ballerina/java}METHOD_SIGNATURE_DOES_NOT_MATCH 'Incompatible return type for method 'getInstance' in class 'org.maryam.Test': Java type 'org.ballerinalang.jvm.values.MapValueImpl' will not be matched to ballerina type '(map<string> & readonly)''
```

**Affected Versions:**
slp1

PR #24531, opened 2020-07-01
Fix interop for intersection types

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/24529

## Approach
Since the value that will be returned or passed as an argument for an intersection type will be a value of the effective type, we directly use the effective type in `JMethodResolver`.

## Samples
```ballerina
function getValues(map<int> & readonly intMap, map<string> & readonly stringMap) 
   returns (int|string)[] & readonly = @java:Method {
        class: "org/ballerinalang/nativeimpl/jvm/tests/StaticMethods"
} external;
```

```java
public static ArrayValue getValues(MapValue<BString, Long> intMap, MapValue<BString, BString> stringMap) {
   // ....
}
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/java/org/ballerinalang/nativeimpl/jvm/tests/StaticMethods.java
```diff
@@ -392,4 +395,36 @@ public static BFuture getFutureOnly(BFuture future) {
...
+
+        for (Map.Entry<BString, Long> intEntry : intMap.entrySet()) {
+            entries[index++] = new ListInitialValueEntry.ExpressionEntry(intEntry.getValue());
+        }
+
+        for (Map.Entry<BString, BString> stringEntry : stringMap.entrySet()) {
+            entries[index++] = new ListInitialValueEntry.ExpressionEntry(stringEntry.getValue());
+        }
+
+        return new ArrayValueImpl(new BArrayType(new BUnionType(new ArrayList(2) {{
```
pubudu91, line 413: Can't we avoid creating this array type each time this is called?

MaryamZi: The size which is passed to the `BArrayType` constructor depends on the actual map value arguments, so we can't create the array type just once. We could maybe extract out the union type, but this method is just called twice via a single test, so I'm not sure if there's much to gain by doing so. WDYT?

pubudu91: Ah just now I noticed that this is a test file. No need then

PR #24554, opened 2020-07-02
Create methods for finding specific Ancestors and all Ancestors

### LINKED ISSUE #23709: API to retrieve the parent node by querying
**Description:**
Currently, I'm implementing the new formatter with the new parser and faced the following issues. Since all the children nodes should be formatted based on their parent nodes, there should be an API to get the parent node. Currently, the parent() method returns the immediate parent that won't help for the formatter in most of the cases. As an example, a variable inside a function returns the function body node as the parent. But what the formatter requires is function node as the parent. Ideally, there should be an API to retrieve the parent node by querying. 

**Steps to reproduce:**

**Affected Versions:**

m36dot:
## Purpose
$subject

Fixes #23709

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/syntax/tree/Node.java
+++ b/compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/syntax/tree/Node.java
@@ -25,6 +25,10 @@ import io.ballerinalang.compiler.text.LineRange;
 import io.ballerinalang.compiler.text.TextDocument;
 import io.ballerinalang.compiler.text.TextRange;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Predicate;
+
 /**
  * This class represents a node in the syntax tree.
  *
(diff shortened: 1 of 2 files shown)


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/syntax/tree/Node.java
```diff
@@ -59,6 +63,33 @@ public NonTerminalNode parent() {
         return parent;
     }
 
+    public NonTerminalNode ancestor(Predicate<Node> filter) {
```
sameerajayasoma, line 66: Shall we make the return type `Optional< NonTerminalNode >`?

m36dot: sure

Issue #24471, opened 2020-06-27
Check on the possibility of using `Flags.ANONYMOUS` to identify anonymous types when deciding to use the shape for `toString()` instead of the auto-generated name


MaryamZi:
**Description:**
Currently, to identify if a type is anonymous when building the `toString()` representation we check if the name starts with `$`. Need to check on the possibility of using `Flags.ANONYMOUS` to identify anonymous types instead.

We would especially need this when we allow quoted identifiers to have non-alphanumeric characters which will allow a user to define a type with a name starting with `$`.

hasithaa:
This issue has been marked as "Deferred" and will be revisited in the future when resources are available.

PR #24652, opened 2020-07-08
Fix few issues in diagnostics related to `readonly` and mapping/list constructor expressions

### LINKED ISSUE #23816: Issues in diagnostics for `T & readonly`
**Description:**
$title.

**Issue I**
Attempting to update an immutable value via member/field access lvexprs result in an additional error.

```ballerina
public function main() {
    readonly & string[] x = ["int"];
    x[2] = ["abc"]; // cannot update 'readonly' value of type 'string[] & readonly'
                          // incompatible types: expected 'other', found '[string]'
}  
```

**Issue II**
Errors related to anonymous records/objects print `$anonType$n`.

```ballerina
public function main() {
    readonly & record {int i;} x = {i: 1}; 
    x.i = 2; // cannot update 'readonly' value of type '($anonType$0 & readonly)'
} 
```

### LINKED ISSUE #24564: Additional error when there are issues with the applicable CET for a constructor expression
**Description:**
$title.

**Steps to reproduce:**
```ballerina
public function main() {
    Foo f = {
        i: 1
    };
}
```

```cmd
$ ballerina run qwe.bal 
Compiling source
        qwe.bal
error: .::qwe.bal:2:5: unknown type 'Foo'
error: .::qwe.bal:2:13: a type compatible with mapping constructor expressions not found in type 'other'
```

**Affected Versions:**
slp1

MaryamZi:
## Purpose
$title.

These changes have been introduced on top of the changes in https://github.com/ballerina-platform/ballerina-lang/pull/24484. Need to merge that first.

Fixes #24564 - Additional error when there are issues with the applicable CET for a constructor expression
Fixes #23816 - Issues in diagnostics for `T & readonly`

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BStructureType.java
```diff
@@ -49,4 +54,22 @@ public BStructureType(int tag, BTypeSymbol tSymbol, int flags) {
...
+
+        if (value.isEmpty() || value.startsWith(DOLLAR)) {
+            return true;
+        }
+
+        if (!(Symbols.isFlagOn(this.tsymbol.flags, Flags.READONLY))) {
+            return false;
+        }
+
+        return value.startsWith("($");
```
KRVPerera, line 73: We have already checked part of this condition above L65. Should we return `false` here?

KRVPerera, line 73: Just curious what this check does?

MaryamZi: For immutable record/object types the name is something like `(T & readonly)`. So in the anonymous case it will be something like `($anonType$n & readonly)`.

We should clean this up in general though, to look at a flag maybe, instead of the name. https://github.com/ballerina-platform/ballerina-lang/issues/24471 tracks that.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1280,7 +1280,16 @@ private BType checkListConstructorCompatibility(BType bType, BLangListConstructo
             this.dlog.setNonConsoleDLog();
 
             List<BType> compatibleTypes = new ArrayList<>();
+            boolean erroredExpType = false;
             for (BType memberType : ((BUnionType) bType).getMemberTypes()) {
+                if (memberType == symTable.semanticError) {
```
pubudu91, line 1285: How can there be a semantic error in a union type?

MaryamZi: One example is when there's a member type that is not defined.

```ballerina
public function main() {
    Foo|int[] x = [1, 2];
}
```

We seem to be logging an `unknown type 'Foo'` error, and using semantic error as the member type.

PR #24515, opened 2020-06-30
Add BFM support in new parser

lochana-chathura:
## Purpose
$subject.

Fixes #22030 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #24684, opened 2020-07-09
Enable documentaion tests for new parser

lochana-chathura:
## Purpose
* Enable documentation tests for new parser
* Revamp markdown documentation parsing added by #24515

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -1719,6 +1750,7 @@ private boolean processDocumentationReference(int nextChar) {
             case LexerTerminals.MODULE:
             case LexerTerminals.FUNCTION:
             case LexerTerminals.PARAMETER:
+            case LexerTerminals.CONST:
```
lochana-chathura, line 1753: Support for the special keyword `const` was added, as it is going to be included in the spec. (https://github.com/ballerina-platform/ballerina-spec/issues/336)
Keyword is already supported in the DocumentationAnalyzer and used in the unit tests as well.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -12815,7 +12815,7 @@ private STNode parseDocumentationLine(STNode hashToken) {
             case 1:
                 STNode docElement = docElements.get(0);
                 if (docElement.kind == SyntaxKind.DOCUMENTATION_DESCRIPTION) {
-                    if (((STToken) docElement).text().startsWith("# Deprecated")) {
+                    if (((STToken) docElement).text().startsWith(" # Deprecated")) {
```
SupunS, line 12818: IMO amount of leading whitespaces shouldn't matter.
Is it possible to send this from the lexer itself, as a token?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -1578,13 +1598,60 @@ private STToken readTokenInBracedContentInInterpolation() {
...
+        lookAheadCount++;
+        int lookAheadChar = reader.peek(lookAheadCount);
+        while (lookAheadChar == LexerTerminals.SPACE || lookAheadChar == LexerTerminals.TAB) {
+            lookAheadCount++;
+            lookAheadChar = reader.peek(lookAheadCount);
+        }
+
+        // Look ahead for a "Deprecated" word match.
+        for (int i = 0; i < 10; i++) {
+            if ((char) lookAheadChar != LexerTerminals.DEPRECATED.charAt(i)) {
```
SupunS, line 1640: we can do this in a better-performance way:
```java
char[] deprecatedChars = {'D', 'E', 'P', 'R', 'E', 'C', 'A', 'T', 'E', 'D' };

if ((char) lookAheadChar != deprecatedChars[i]) {
}
```

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -1578,13 +1598,60 @@ private STToken readTokenInBracedContentInInterpolation() {
...
+                return readDocumentationInternalToken();
+            }
+            lookAheadCount++;
+            lookAheadChar = reader.peek(lookAheadCount);
+        }
+
+        // There is a match. Hence return a deprecation literal.
+        processLeadingTrivia();
+        reader.mark();
+        while (!getLexeme().endsWith(LexerTerminals.DEPRECATED)) {
```
SupunS, line 1651: why do we need this while loop?
is it possible to avoid `getLexeme().endsWith(LexerTerminals.DEPRECATED)` check? Coz regex-related operations on top of java Strings (such as contains(), endsWith(), startsWith(), etc ) can be really expensive..

lochana-chathura: Fixed with 2331d55. Thanks for pointing out expensive operation. :)

Reason for the while loop:
By code line 1648, lexer knows there's a `WS # WS Deprecated` match ahead. After calling `processLeadingTrivia()` lexer doesn't know, for how many chars, the reader should be advanced.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaLexer.java
```diff
@@ -1362,6 +1306,70 @@ private STToken processSlashToken() {
...
+                    reader.advance();
+                    nextChar = peek();
+                    continue;
+            }
+            break;
+        }
+
+        STNode leadingTrivia = STNodeFactory.createNodeList(this.leadingTriviaList);
+        String lexeme = getLexeme();
+        STNode trailingTrivia = STNodeFactory.createNodeList(new ArrayList<>(0)); // No trailing trivia
```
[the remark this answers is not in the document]
lochana-chathura: New line is already included in the documentation string. 
e.g. 
```
#  line1
function foo() { }
```
documentation string would be `"# line1\n"`
Then in the documentation lexer it will be parsed as the trailing trivia of whatever the last token.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -12994,242 +12994,70 @@ private DiagnosticErrorCode validateArgMatchPatternOrder(SyntaxKind prevArgKind,
...
-     */
-    private STNode parseMinusToken() {
-        STToken token = peek();
-        if (token.kind == SyntaxKind.MINUS_TOKEN) {
-            return consume();
-        } else {
-            return STNodeFactory.createMissingToken(SyntaxKind.MINUS_TOKEN);
-        }
-    }
+    private STNode rearrangeMarkdownDocumentationLines(List<STNode> markdownDocLineList) {
```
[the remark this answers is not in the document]
lochana-chathura: yea flattening.
yes we can. will add that fix.

PR #24761, opened 2020-07-15
Fix issue in parsing invalid mapping constructure expression

### LINKED ISSUE #24760: Issue in parsing invalid mapping constructure expression
**Description:**
consider the following example of an invalid expression statement.
``` 
function foo() {
    {var1: fooVar1, var2: fooVar2};
}
```
bellow is the target section of the generated tree.
```
                          {
                              "kind": "LIST",
                              "children": [
                                {
                                  "kind": "QUALIFIED_NAME_REFERENCE",
                                  "children": [
                                    {
                                      "kind": "IDENTIFIER_TOKEN",
                                      "value": "var1"
                                    },
                                    {
                                      "kind": "COLON_TOKEN",
                                      "trailingMinutiae": [
                                        {
                                          "kind": "WHITESPACE_MINUTIAE",
                                          "value": " "
                                        }
                                      ]
                                    },
                                    {
                                      "kind": "SIMPLE_NAME_REFERENCE",
                                      "children": [
                                        {
                                          "kind": "IDENTIFIER_TOKEN",
                                          "value": "fooVar1"
                                        }
                                      ]
                                    }
                                  ]
                                },
```
First field is parsed as `QUALIFIED_NAME_REFERENCE` which should be corrected as `SPECIFIC_FIELD`

- Component/Parser
- Type/Bug

rdulmina:
## Purpose
$subject

Fixes #24760

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -15446,7 +15480,7 @@ private STNode parseStmtAsMappingBPOrMappingConsStart(STNode openBrace, STNode m
 
                 // else treat as expression.
                 switchContext(ParserRuleContext.EXPRESSION_STATEMENT);
-                expr = getExpression(bpOrConstructor);
+                expr = getMappingConstructureExpr(bpOrConstructor);
```
SupunS, line 15483: Would it be possible to incorporate the above change to the existing method?
There are other places that uses the same `getExpression` method, and the issue may also be there for those as well.

rdulmina: This is very specific case, we can not use the existing method since `getExpression` is used in other places. If we r to change `QUALIFIED_NAME_REFERENCE` case in `getExpression` It will affect the others

SupunS: Yeah thats true. I was thinking, we could change the way we construct the `MAPPING_BP_OR_MAPPING_CONSTRUCTOR` content, inside `getExpression`

rdulmina: Changes done

PR #24791, opened 2020-07-16
Add integration test for resolving `ballerina` modules from central

pramodya1994:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-integration-test/src/test/java/org/ballerinalang/test/packaging/PackagingTestCase.java
+++ b/tests/jballerina-integration-test/src/test/java/org/ballerinalang/test/packaging/PackagingTestCase.java
@@ -55,6 +55,7 @@ import static org.ballerinalang.cli.module.util.Utils.createHttpUrlConnection;
 import static org.ballerinalang.cli.module.util.Utils.initializeSsl;
 import static org.ballerinalang.cli.module.util.Utils.setRequestMethod;
 import static org.ballerinalang.test.packaging.ModulePushTestCase.REPO_TO_CENTRAL_SUCCESS_MSG;
+import static org.ballerinalang.test.packaging.PackerinaTestUtils.copyFolder;
 import static org.wso2.ballerinalang.compiler.util.ProjectDirConstants.BLANG_COMPILED_PKG_BINARY_EXT;
 import static org.wso2.ballerinalang.util.RepoUtils.BALLERINA_STAGE_CENTRAL;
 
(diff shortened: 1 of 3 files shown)


tests/jballerina-integration-test/src/test/java/org/ballerinalang/test/packaging/PackagingTestCase.java
```diff
@@ -297,6 +298,27 @@ public void testBuildAndRunModuleWithPeriod() throws BallerinaTestException {
...
+    @Test(description = "Test modules with org-name `ballerina` resolving from central, if new version is available")
+    public void testResolveBallerinaModulesFromCentral() throws BallerinaTestException, IOException {
+        // copy `ballerina-central-module` project from resources to a temp
+        Path originalTestProject = Paths
+                .get("src", "test", "resources", "packaging", "scope", "ballerina-central-module").toAbsolutePath();
+        Path projectPath = tempProjectDirectory.resolve("ballerinaCentralModuleProject");
+        copyFolder(originalTestProject, projectPath);
+
+        // Build module
+        LogLeecher buildLeecher = new LogLeecher("ballerina/socket:9.0.0 pulled from central successfully");
```
NipunaRanasinghe, line 310: If I'm not mistaken, this test will be failing once a newer version of the socket module is available, as the module version is hard-coded here. Is it the expected behaviour or am I missing something?

hevayo: This particular module version is pushed to staging central server we used a high version number so that newer version will not break the test.  I doubt socket module will reach 9.0.0. The test will check if newer modules in central will get pulled even though module is found in bit cache.

pramodya1994: The current `ballerina/socket` version is 0.6.5, So we decided to pull a higher version (9.0.0) than that to stg-central. In this test case that pulled version to stg-central is pulling when building this project. Normally we are using stg-central only for testing purposes and newer versions will to add to it. So we will not have a need to update this version in the near future. Do we have any good way of doing this? So I can put another PR for that.

NipunaRanasinghe: @hevayo @pramodya1994 thanks for the explanations. I also think this would okay as we are using an unreachable version. +1

PR #24822, opened 2020-07-17
Add transactional keyword to resource methods

mohanvive:


mohanvive:
> Shall we add a parser test too?

@SupunS Added a parser test.

PR #24835, opened 2020-07-19
Support executing a single test function

### LINKED ISSUE #24628: Support executing a single test function
**Description:**

Running single test cases in the current test framework is not possible. It may be brought about by including groups or disabling tests but this is a cumbersome task.

Dilhasha:
Fixes #24628 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [x] Module documentation in Module.md files
  - [ ] Ballerina By Examples


tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/SelectedFunctionTest.java
```diff
@@ -0,0 +1,97 @@
...
+        balClient.runMain("test", new String[]{"--tests", "testFunc2", "single-test-execution.bal"},
+                null, new String[]{}, new LogLeecher[]{clientLeecher}, projectPath);
+        clientLeecher.waitForText(20000);
+    }
+
+    @Test
+    public void testMultipleFunctionExecution() throws BallerinaTestException {
+        String msg = "2 passing";
+        LogLeecher clientLeecher = new LogLeecher(msg);
+        balClient.runMain("test", new String[]{"--tests", "testFunc,testFunc2",
```
Ibaqu, line 64: Have we considered having tests for test cases that fail as well?

Suppose we have a test case named `testFunc3` that is set to fail. If we have the command `--tests testFunc2, testFunc3 `, then will the result be `2 passing, 1 failing`? 

Also does the execution continue if the testFunc3 is provided at the start. Something like `--tests testFunc3, testFunc2`?

Dilhasha: @Ibaqu The test failure Is similar to how failed tests are handled in our test suites. 

The order of execution is similar to the default test execution. The tests are not run in the order we pass them to the --tests flag.

tests/testerina-integration-test/src/test/resources/single-file-tests/single-test-execution/single-test-execution.bal
```diff
@@ -0,0 +1,51 @@
+import ballerina/test;
```
anupama-pathirage, line 1: Shall we add the license header for this file?

Issue #718, opened 2017-01-11
[Tooling] Resource name does not generate in source view correctly


maheeka:
When switching between source view and drag-drop view name of the resource is different. Rename name with appending characters or removing characters there is always a single character additional or less in the source view.

PR #24838, opened 2020-07-20
Remove ANTLR dependencies from Documentation Analyzer

### LINKED ISSUE #24730: Remove ANTLR dependencies from Documentation Analyzer
**Description:**
* Currently  single-backtick content in markdown documentation is parsed as a single string chunk and subsequently in the documentation analyzer it is re-parsed using ANTLR.

* Parsing of single-backtick content as syntax nodes should be supported from parser itself.

**Related Issues:**
#22030

**Suggested Labels:**
Component/Parser, Component/Compiler

**Suggested Assignees:**
@lochana-chathura

lochana-chathura:
## Purpose
* Add support for parsing of single-backtick content as syntax nodes
* Remove ANTLR dependencies from Documentation Analyzer

Fixes #24730 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/DocumentationLexer.java
```diff
@@ -669,23 +702,79 @@ private STToken readDocumentationBacktickContentToken() {
...
+        }
+
+        reader.advance(lookAheadCount);
+        resetSpecialKeywordInPlace();
+
+        boolean isIdentifier = backtickStr.matches(identifierRegex);
+        if (isIdentifier) {
+            return getIdentifierToken();
+        } else {
+            reportLexerError(DiagnosticWarningCode.WARNING_INVALID_DOCUMENTATION_IDENTIFIER);
```
lochana-chathura, line 746: do we inject parser diagnostic to the compiler diagnostic somewhere? or two APIs act independently?
right now I log a warning for this in the node transformer as well.

SupunS: Yes, we add all the parser errors to the compiler-diagnostics at the [Parser.java](https://github.com/ballerina-platform/ballerina-lang/blob/74c3396849e62f2b13b587bab6af3d1cd8cf7be9/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/Parser.java#L265) class

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/DocumentationLexer.java
```diff
@@ -669,23 +702,79 @@ private STToken readDocumentationBacktickContentToken() {
...
+                    nextToken = reader.peek(lookAheadCount);
                     continue;
             }
             break;
         }
 
-        return getDocumentationLiteral(SyntaxKind.BACKTICK_CONTENT);
+        if (!isSpecialKeywordInPlace) {
+            // Look for a x(), m:x(), T.y(), m:T.y() match
+            boolean hasMatch = backtickStr.matches(backtickExprRegex);
```
SupunS, line 728: Shall we use our own method to visit the characters and do this validation? 
Regex-matching can be expensive. I think if we remove that regex, way we can also avoid Line #718 (string concat) which is also a bit expensive.
Usually numeric operations are much faster compared to string operations. eg: (int/char comparison vs string compare, array-access vs map-access)

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
```diff
@@ -456,6 +459,24 @@ private DiagnosticPos getPosition(Node node) {
...
+        ChildNodeList children = nonTerminalNode.children();
+        // If there's metadata it will be the first child.
+        // Hence set start position from next immediate child.
+        if (children.get(0).kind() == SyntaxKind.METADATA) {
+            LinePosition startPos = children.get(1).lineRange().startLine();
+            LinePosition endPos = node.lineRange().endLine();
+            return new DiagnosticPos(diagnosticSource, startPos.line() + 1, endPos.line() + 1, startPos.offset() + 1,
+                    endPos.offset() + 1);
+        } else {
+            throw new IllegalArgumentException("Node should have metadata");
```
SupunS, line 476: Shall we return the normal-positions if there are no metadata?
Reason is, theres a plan for making metadata optional in the syntax tree.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
```diff
@@ -4258,26 +4303,69 @@ private String addReferencesAndReturnDocumentationText(LinkedList<BLangMarkdownR
...
+                    // reaching here means, backtick content is not in one of x(), m:x(), T.y(), m:T.y() formats
+                    // no warning is logged for this case
+                    bLangRefDoc.hasParserWarnings = true;
+                }
 
-                Token backtickContent = docReferenceNode.backtickContent();
-                String contentString = backtickContent.isMissing() ? "" : backtickContent.text();
-                bLangRefDoc.referenceName = contentString;
-                docText.append(contentString);
+                if (!referenceType.isPresent() && backtickContent.kind() != SyntaxKind.BACKTICK_CONTENT) {
```
SupunS, line 4338: can take `!referenceType.isPresent()` condition out, as its common for both conditions.

Also would be good to extract this content to a separate method?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
```diff
@@ -4258,26 +4303,69 @@ private String addReferencesAndReturnDocumentationText(LinkedList<BLangMarkdownR
...
                 bLangRefDoc.type = DocumentationReferenceType.BACKTICK_CONTENT;
-                docReferenceNode.referenceType().ifPresent(
+                Optional<Token> referenceType = docReferenceNode.referenceType();
+                referenceType.ifPresent(
                         refType -> {
                             bLangRefDoc.type = stringToRefType(refType.text());
+                            if (backtickContent.kind() == SyntaxKind.BACKTICK_CONTENT) {
+                                // backtick content preceded by a special keyword, should be an identifier
+                                bLangRefDoc.hasParserWarnings = true;
+                                dlog.warning(pos, DiagnosticCode.INVALID_DOCUMENTATION_IDENTIFIER, contentString);
```
SupunS, line 4324: I think we can remove this warning, since all parser errors (should) get added to compiler diagnostics.

stdlib/mime/src/main/ballerina/src/mime/natives.bal
```diff
@@ -620,7 +620,7 @@ function externHasHeader(Entity entity, string headerName, HeaderPosition positi
 # + contentToBeEncoded - Content that needs to be encoded can be of type `string`, `byte[]` or `io:ReadableByteChannel`
 # + charset - Charset to be used. This is used only with the string input
 # + return - An encoded `string` if the given input is of type string, an encoded `byte[]` if the given input is of
-#            type byte[], an encoded `io:ReadableByteChannel` if the given input is of type `io:ReadableByteChannel`,
+#            type byte[], an encoded `io:ReadableByteChannel` if the given input is of type ``io:ReadableByteChannel``,
```
SupunS, line 623: shouldn't this be allowed?
i.e:
```
# some text followed by type `M:X`
```

Spec says "if it is a qualified name M:X, then the source file must have imported M, and X must refer to a public name of an appropriate type in M"

lochana-chathura: this fix was added with fac6140.

seems we have to make an exception for `function` keyword.

Spec says, 
*BFM also recognizes `` `f()` `` as an alternative to `` function `f` ``. **In both cases**, f can have any of the following forms
`x(), m:x(), T.y(), m:T.y()`*

so for `function` keyword, all together 6 formats(`x, m:x, x(), m:x(), T.y(), m:T.y()`) should be allowed right?

SupunS: yea, thats correct.
We can add some parser test - that would be sufficient.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/documentation/MarkdownDocumentationTest.java
```diff
@@ -340,7 +340,7 @@ public void testDocumentationNegative() {
         BAssertUtil.validateWarning(compileResult, index++, "no documentable return parameter", 11, 1);
         BAssertUtil.validateWarning(compileResult, index++, "undocumented field 'cd'", 15, 5);
         BAssertUtil.validateWarning(compileResult, index++,
-                "invalid identifier in documentation reference '9invalidServ'", 21, 3);
+                "invalid identifier in documentation reference '9invalidServ'", 21, 12);
```
lochana-chathura, line 343: In case of invalid backtick content, earlier position was given as the start position of doc ref. Now it is directly pointed at the backtick content. 
e.g.
For, ``# service `9invalidServ` ``
* earlier position: warn column 3
* new position: warn column 12

SupunS: thats great!

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/Parser.java
```diff
@@ -268,8 +269,16 @@ private void reportSyntaxDiagnostics(BDiagnosticSource diagnosticSource, SyntaxT
 
             // TODO This is the temporary mechanism
             // We need to merge the diagnostic reporting mechanisms of the new parser and the semantic analyzer
-            DiagnosticCode code = DiagnosticCode.SYNTAX_ERROR;
-            dlog.error(pos, code, syntaxDiagnostic.message());
+            DiagnosticCode code;
+
+            DiagnosticSeverity severity = syntaxDiagnostic.diagnosticInfo().severity();
+            if (DiagnosticSeverity.WARNING.equals(severity)) {
```
SupunS, line 275: can doo `==` for equal check

PR #24862, opened 2020-07-21
Fix `is type` check for `readonly` fields

### LINKED ISSUE #24495: Runtime `is` checks should change for `readonly` fields
**Description:**
$title.

Currently when doing the `is` check (`x is T`) at runtime we do the "is-like" check only if the value (`x`) is immutable.

But, with the introduction of `readonly` fields in records/objects, it is possible for the `is` check to evaluate to true via the `is-like` check even if the entire value is not immutable.

**Steps to reproduce:**
```ballerina
import ballerina/io;

type Foo record {
    int i;
};

type Bar record {
    readonly string|int i;
};

public function main() {
    Bar b = {i: 100};

    io:println(<any> b is Foo); // currently prints `false`, but should print `true` since the shape is guaranteed to be 
                                // `record { int i; }`
}
```

**Affected Versions:**
slp1

### LINKED ISSUE #24739: Issue in object sub-typing checks with finite-typed fields
**Description:**
$title.

**Steps to reproduce:**

Module `foo`

```ballerina
type HUNDRED 100;

public type MyObj object {
    HUNDRED b = 100;
};
```

Module `bar`

```ballerina
import foo;
import ballerina/io;

type HUNDRED 100;

public type MyObj object {
    HUNDRED b = 100;
};

public function main() {
    MyObj f = new;
    io:println(<any>f is foo:MyObj);
}
```

Running `bar` should print `false` due to the fields of module-level visibility, but this currently prints `true`. Seems to be due to the module details not being available for finite types.

The following prints `false` as expected.

Module `foo`

```ballerina
type MyRec record {
    int i;
};

public type MyObj object {
    MyRec m = {i: 100};
};
```

Module `bar`

```ballerina
import foo;
import ballerina/io;

type MyRec record {
    int i;
};

public type MyObj object {
    MyRec m = {i: 100};
};

public function main() {
    MyObj f = new;
    io:println(<any>f is foo:MyObj);
}
```

**Affected Versions:**
slp1

### LINKED ISSUE #24752: Object sub-typing checks produce incorrect results
**Description:**
Object sub-typing runtime checks with built-in typed fields that are of module-level visibility, result in incorrect results. 

**Steps to reproduce:**
Module `foo`

```ballerina
public type Foo object {
    int i = 1;
};
```

Module `bar`

```ballerina
import foo;
import ballerina/io;

public type Bar object {
    int i = 1;
};

public function main() {
    Bar b = new;
    // io:println(b is foo:Foo); // incompatible types: 'maryam/bar:0.1.0:Bar' will not be matched to 'maryam/foo:0.1.0:Foo'
    io:println(<any> b is foo:Foo);
}
```

This prints true atm, but should print false.

Btw, this seems to be correct at compile-time (note the commented out compilation error).

Prints false for user-defined type-d fields.

**Affected Versions:**
slp1

**Related Issue(s):**
https://github.com/ballerina-platform/ballerina-lang/issues/24739

### LINKED ISSUE #24858: Stack overflow error with `checkIsType` for objects with fields of the same type
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type Foo object {
    Foo? fn;

    public function init(Foo? fn = ()) {
        self.fn = fn;
    }
};

type Bar object {
    Bar? fn;

    public function init(Bar? fn = ()) {
        self.fn = fn;
    }
};

public function main() {
    Foo f1 = new;
    Foo f2 = new (f1);

    boolean b = <any> f1 is Bar;
}
```

**Affected Versions:**
slp2

### LINKED ISSUE #25199: `org.ballerinalang.jvm.types.BType`'s `equals()` method only looks at the module name and not the org-name/version
**Description:**
$title. Results in two completely different types being considered as equal.

**Affected Versions:**
slp2

MaryamZi:
## Purpose
$title. Fixes #24495 - Runtime `is` checks should change for `readonly` fields

This PR also
Fixes #24858 - Stack overflow error with `checkIsType` for objects with fields of the same type
Fixes #24752 - Object sub-typing checks produce incorrect results
Fixes #24739 - Issue in object sub-typing checks with finite-typed fields
Fixes #25199 - `org.ballerinalang.jvm.types.BType`'s `equals()` method only looks at the module name and not the org-name/version

## Approach
When the `is` check is used with records or objects, we pass the value also, to do the `is like` check for `readonly` fields since such fields cannot change.

## Samples
```ballerina
import ballerina/io;

const HUNDRED = 100;

type Foo record {
    HUNDRED i;
};

type Bar record {
    readonly string|int i;
};

public function main() {
    Bar b1 = {i: 123};
    io:println(<any> b1 is Foo); // prints `false`

    Bar b2 = {i: 100};
    io:println(<any> b2 is Foo); // prints `true` now, since the `i` field of `b2` is and will continue to be `100`
}
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -975,6 +1070,137 @@ private static boolean checkIsRecordType(BType sourceType, BRecordType targetTyp
         return true;
     }
 
+    private static boolean checkRecordBelongsToAnydataType(MapValue sourceVal, BRecordType recordType,
+                                                           List<TypePair> unresolvedTypes) {
+        BType targetType = BTypes.typeAnydata;
+        TypePair pair = new TypePair(recordType, targetType);
+        if (unresolvedTypes.contains(pair)) {
```
KRVPerera, line 1077: I think `HashSet` is a better choice if we are searching in the `unresolvedTypes`.

MaryamZi: This is passed on to `checkIsType` and in general we seem to be using lists in `checkIsType` and `checkIsLikeType`. Shall we do this change separately? Can you please create an issue with the suggestion?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -741,6 +796,46 @@ private static boolean checkIsMapType(BType sourceType, BMapType targetType, Lis
...
+                        continue;
+                    }
+
+                    BString name = StringUtils.fromString(field.name);
+
+                    if (Flags.isFlagOn(field.flags, Flags.OPTIONAL) && !sourceMapValue.containsKey(name)) {
+                        continue;
+                    }
+
+                    if (!checkIsLikeType(sourceMapValue.get(name), targetConstrainedType)) {
```
pubudu91, line 824: What's the reason to call `checkIslikeType()` for readonly fields and `checkIsType()` for non-readonly fields?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1132,26 +1363,70 @@ private static boolean checkObjectEquivalency(BType sourceType, BObjectType targ
             return false;
         }
 
+
+        String targetTypeModule = Optional.ofNullable(targetType.getPackage()).map(BPackage::getName).orElse("");
+        String sourceTypeModule = Optional.ofNullable(sourceObjectType.getPackage()).map(BPackage::getName).orElse("");
```
pubudu91, line 1368: We should consider the org name too right? Not sure whether the version should be considered though

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
MaryamZi: A `readonly` field cannot be updated, and whatever the value that was initially set is guaranteed to be immutable. 

For an immutable value, is type is the same as is like type (as though the "inherent type" changed). 

So if there is a `readonly` field, we know that it can never change (the reference or the value). So if the value specified for that field looks like the target type it would mean that it belongs to the target type.

This is similar to how we call check-is-like-type when check-is-type is used with immutable values.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
MaryamZi: We should consider all 3 right? Two different versions of the same module can be considered as two different modules IMO, when considering visibility regions.

Will fix, thanks.

MaryamZi: Came across https://github.com/ballerina-platform/ballerina-lang/issues/25199 when trying to fix this. Both fixed in 8d23769.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -641,6 +642,60 @@ public static boolean checkIsType(BType sourceType, BType targetType, List<TypeP
...
+    private static boolean checkIsType(Object sourceVal, BType sourceType, BType targetType,
+                                      List<TypePair> unresolvedTypes) {
+        int sourceTypeTag = sourceType.getTag();
+        int targetTypeTag = targetType.getTag();
+
+        if (sourceTypeTag != TypeTags.RECORD_TYPE_TAG && sourceTypeTag != TypeTags.OBJECT_TYPE_TAG) {
+            return checkIsType(sourceType, targetType);
+        }
+
+        if (targetType.getTag() == TypeTags.INTERSECTION_TAG) {
```
Kishanthan, line 654: targetTypeTag ? as you have already acquired it

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -975,6 +1070,137 @@ private static boolean checkIsRecordType(BType sourceType, BRecordType targetTyp
...
+            return true;
+        }
+        unresolvedTypes.add(pair);
+
+        Map<String, BField> fields = recordType.getFields();
+
+        for (BField field : fields.values()) {
+            String fieldName = field.getFieldName();
+
+            boolean optionalField = Flags.isFlagOn(field.flags, Flags.OPTIONAL);
```
Kishanthan, line 1087: this flag can move within the if body? as it is used only there.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -975,6 +1070,137 @@ private static boolean checkIsRecordType(BType sourceType, BRecordType targetTyp
...
+
+            if (Flags.isFlagOn(field.flags, Flags.READONLY)) {
+                BString fieldNameBString = StringUtils.fromString(fieldName);
+
+                if (optionalField) {
+                    if (!sourceVal.containsKey(fieldNameBString)) {
```
Kishanthan, line 1093: merge these two ifs

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -975,6 +1070,137 @@ private static boolean checkIsRecordType(BType sourceType, BRecordType targetTyp
...
+            }
+
+            boolean optionalTargetField = Flags.isFlagOn(targetField.flags, Flags.OPTIONAL);
+            boolean optionalSourceField = Flags.isFlagOn(sourceField.flags, Flags.OPTIONAL);
+
+            if (Flags.isFlagOn(sourceField.flags, Flags.READONLY)) {
+                BString fieldNameBString = StringUtils.fromString(fieldName);
+
+                if (optionalSourceField) {
+                    if (!sourceRecordValue.containsKey(fieldNameBString)) {
```
Kishanthan, line 1161: nested ifs. can it be merged?

tests/jballerina-unit-test/src/test/resources/test-src/balo/test_projects/test_project_two/Ballerina.toml
```diff
@@ -0,0 +1,5 @@
+[project]
+org-name= "testorgtwo"
+version= "2.0.0"
+
+[dependencies]
```
Kishanthan, line 5: EOF line?

MaryamZi:
@Kishanthan, addressed the suggestions. Thank you!

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
KRVPerera: Help add labels
https://github.com/ballerina-platform/ballerina-lang/issues/25293

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -641,6 +642,60 @@ public static boolean checkIsType(BType sourceType, BType targetType, List<TypeP
         }
     }
 
+    private static boolean checkIsType(Object sourceVal, BType sourceType, BType targetType,
```
Kishanthan, line 645: why we need another checkIsType method ? whats the purpose of this newly added method?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -741,6 +796,46 @@ private static boolean checkIsMapType(BType sourceType, BMapType targetType, Lis
...
     }
 
+    private static boolean checkIsMapType(Object sourceVal, BType sourceType, BMapType targetType,
+                                          List<TypePair> unresolvedTypes) {
+        BType targetConstrainedType = targetType.getConstrainedType();
+        switch (sourceType.getTag()) {
+            case TypeTags.MAP_TAG:
+                return checkConstraints(((BMapType) sourceType).getConstrainedType(), targetConstrainedType,
+                                        unresolvedTypes);
+            case TypeTags.RECORD_TYPE_TAG:
```
Kishanthan, line 806: This case seems to have a big logic. shall we move it to a separate method for better readability?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
MaryamZi: The current `checkIsType` method is based solely on the source and target types. We check `checkIsLikeType` only if the entire source value is immutable.

But now with `readonly` fields, even though the entire source value is not immutable, due to the presence of `readonly` fields, an `is` check that fails for types, may pass for values. 

For example, from the sample in the description:

```ballerina
import ballerina/io;

const HUNDRED = 100;

type Foo record {
    HUNDRED i;
};

type Bar record {
    readonly string|int i;
};

public function main() {
    Bar b1 = {i: 123};
    io:println(<any> b1 is Foo); // prints `false`

    Bar b2 = {i: 100};
    io:println(<any> b2 is Foo); // prints `true` now, since the `i` field of `b2` is and will continue to be `100`
}
```

Here if we just use the current `checkIsType` with `Bar` and `Foo`, it will evaluate to false, since the source has `string|int` as the type for `i`, while the target has `HUNDRED` as the type.

But since the `i` field in the source is a `readonly` field, we know that the value for this field cannot change, so we look at the actual value to see if it belongs to the target type. So, in the case of `b1`, where `i` is `123`, it is still not a match, but for `b2` since `i` is `100`, and it can never change, the `is` check evaluates to `true`.

Will add a comment too.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
MaryamZi: Done in 24985b7.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
MaryamZi: Added a comment in 24985b7.

MaryamZi: Added a comment in 24985b7.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
KRVPerera: Issue: https://github.com/ballerina-platform/ballerina-lang/issues/25293

hasithaa (approved):

Issue #24818, opened 2020-07-17
Enable tests in SignatureHelpTest


m36dot:
**Description:**
Now we are returning funcdefnode position from func keyword or visibility qualifier keyword to end of funcbody, this is causing the following asserts in signatureHelpTest in language-server-core to fail.
{"exprServiceConstructor1.json", "functionsAndTypeDefs.bal"}
{"signatureWithinAnnotation.json", "signatureWithinAnnotations.bal"}

the PR that is making the change = https://github.com/ballerina-platform/ballerina-lang/pull/24868

PR #24868, opened 2020-07-21
Enable JBallerinaUnitTests

m36dot:
## Purpose
Enables JBallerinaUnitTests picked by Saif in [sheet](https://docs.google.com/spreadsheets/d/10tVbEoUPqTGr5pJiuaW6OPGRfpSQQz2E_AYo-BaUehY/edit?ts=5eaa6c81#gid=1116022599)

## Approach
Fixing bugs in parser, Node transformer or fixing the test asserts.

## Remarks
https://github.com/ballerina-platform/ballerina-lang/issues/24818

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/stdlib/http/src/test/java/org/ballerinalang/stdlib/services/configuration/ResourceConfigPathTest.java
+++ b/stdlib/http/src/test/java/org/ballerinalang/stdlib/services/configuration/ResourceConfigPathTest.java
@@ -61,14 +61,14 @@ public class ResourceConfigPathTest {
                 .compile("test-src/services/configuration/resource-arg--pathparam-match.bal");
         Diagnostic[] diag = compileResult.getDiagnostics();
         Assert.assertEquals(diag.length, 8);
-        assertResponse(diag[0], INVALID_RESOURCE_PARAMETERS, 10);
-        assertResponse(diag[1], INVALID_RESOURCE_PARAMETERS, 18);
-        assertResponse(diag[2], INVALID_RESOURCE_PARAMETERS, 43);
-        assertResponse(diag[3], "Empty data binding param value", 46);
+        assertResponse(diag[0], INVALID_RESOURCE_PARAMETERS, 14);
+        assertResponse(diag[1], INVALID_RESOURCE_PARAMETERS, 22);
+        assertResponse(diag[2], "Empty data binding param value", 46);
+        assertResponse(diag[3], INVALID_RESOURCE_PARAMETERS, 48);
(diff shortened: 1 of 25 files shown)


stdlib/http/src/test/java/org/ballerinalang/stdlib/services/configuration/ResourceConfigPathTest.java
```diff
@@ -61,14 +61,14 @@ public void testPathParamAndSignatureParamMatch() {
...
+        assertResponse(diag[1], INVALID_RESOURCE_PARAMETERS, 22);
+        assertResponse(diag[2], "Empty data binding param value", 46);
+        assertResponse(diag[3], INVALID_RESOURCE_PARAMETERS, 48);
         assertResponse(diag[4], "Invalid data binding param in the signature : expected 'naMe', but found 'name'", 55);
-        assertResponse(diag[5], INVALID_RESOURCE_PARAMETERS, 61);
-        assertResponse(diag[6], "Invalid data binding param in the signature : expected 'naMe', but found 'name'", 63);
-        assertResponse(diag[7], INVALID_RESOURCE_PARAMETERS, 69);
+        assertResponse(diag[5], "Invalid data binding param in the signature : expected 'naMe', but found 'name'", 63);
+        assertResponse(diag[6], INVALID_RESOURCE_PARAMETERS, 65);
+        assertResponse(diag[7], INVALID_RESOURCE_PARAMETERS, 73);
```
lochana-chathura, line 71: @m36dot aren't the old positions correct?
In L34 we have the error in details,
```
private static final String INVALID_RESOURCE_PARAMETERS =
            "invalid resource parameter(s): cannot specify > 2 parameters without specifying path config and/or body " +
                    "config in the resource annotation";
```
So I feel position should be the position of annotation. Not function def start. 
I fixed this to get annotation positions.

m36dot: @SupunS what do you think about this?

SupunS: Hmm interesting case. Yea looks like the old positions are correct.
I think we should first check which diag-position is being used to log the error.
If they are logging an error related to the annotation, but uses the pos of the function, then may be we should fix it there.

Issue #24839, opened 2020-07-20
Bad Sad returned for error() constructor RHS expr


rasika:
**Description:**
Consider following code snippet. When compiling it, a `Bad Sad` error is returned along with the proper diagnostics.

**Code:**
```ballerina
function rightHandExpr() {
	 error();
}
```

**Error:**
```
Compiling source
	main.bal
error: .::main.bal:2:3: missing mandatory error message argument in call to error constructor
ballerina: Oh no, something really went wrong. Bad. Sad.
```

**Affected Versions:**
2.0.0-P3-SNAPSHOT

SupunS:
looks like the error is thrown from the TypeChecker:
```
java.lang.NullPointerException
	at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.checkExpr(TypeChecker.java:352)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.checkExpr(TypeChecker.java:329)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.visit(SemanticAnalyzer.java:2161)
	at org.wso2.ballerinalang.compiler.tree.statements.BLangExpressionStmt.accept(BLangExpressionStmt.java:47)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.analyzeNode(SemanticAnalyzer.java:2591)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.analyzeNode(SemanticAnalyzer.java:2559)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.analyzeStmt(SemanticAnalyzer.java:2555)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.visit(SemanticAnalyzer.java:361)
	at org.wso2.ballerinalang.compiler.tree.BLangBlockFunctionBody.accept(BLangBlockFunctionBody.java:57)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.analyzeNode(SemanticAnalyzer.java:2591)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.visit(SemanticAnalyzer.java:333)
	at org.wso2.ballerinalang.compiler.tree.BLangFunction.accept(BLangFunction.java:73)
[...]

PR #24895, opened 2020-07-22
Refactor the semantic checking logic for expression statements

pubudu91:
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -2184,8 +2184,10 @@ public class SemanticAnalyzer extends BLangNodeVisitor {
         boolean isChecked = exprKind == NodeKind.CHECK_EXPR || exprKind == NodeKind.CHECK_PANIC_EXPR;
         BType expExprType = getExpectedTypeOfExpr(exprStmtNode.expr, isChecked);
 
+        // use of check foo(); or checkpanic bar(); as statements is allowed provided that the static type of the
+        // expression is a subtype of error?
         boolean returnsNilOrNever = expExprType == symTable.nilType || expExprType == symTable.neverType
-                || (isChecked && types.isOptionalErrorType(expExprType));
+                || (isChecked && (expExprType.tag == TypeTags.ERROR || types.isOptionalErrorType(expExprType)));
 
         if (!returnsNilOrNever && exprStmtNode.expr.getKind() != NodeKind.FAIL) {
             dlog.error(exprStmtNode.pos, DiagnosticCode.ASSIGNMENT_REQUIRED);
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -81,6 +81,7 @@ import org.wso2.ballerinalang.compiler.tree.BLangWorker;
 import org.wso2.ballerinalang.compiler.tree.BLangXMLNS;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangAccessExpression;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangBinaryExpr;
+import org.wso2.ballerinalang.compiler.tree.expressions.BLangCheckedExpr;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangConstant;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangErrorVarRef;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangExpression;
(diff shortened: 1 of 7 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2177,14 +2178,44 @@ public void visit(BLangExpressionStmt exprStmtNode) {
...
+        if (isCheckedExpr) {
+            tempExpr = ((BLangCheckedExpr) expr).expr;
+            exprKind = expr.getKind();
+        }
+
+        if (exprKind == NodeKind.INVOCATION) {
+            BInvokableSymbol symbol = (BInvokableSymbol) ((BLangInvocation) tempExpr).symbol;
+            // `symbol` is null for invocations of undefined functions. Consider the return type of undefined
+            // functions as `()`.
+            return symbol != null ? symbol.retType : symTable.nilType;
```
MaryamZi, line 2213: Instead of returning nil, what if we return semanticError, and then avoid logging the "ASSIGNMENT_REQUIRED" error for semanticError too? That way, if by any chance we get there for non-invocation expressions with semantic error as the type, it would also be handled?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3266,6 +3266,20 @@ public boolean hasFillerValue(BType type) {
...
+            BUnionType unionType = (BUnionType) type;
+            for (BType t : unionType.getMemberTypes()) {
+                if (t.tag != TypeTags.ERROR && t.tag != TypeTags.NIL) {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        return type.tag == TypeTags.ERROR;
```
MaryamZi, line 3280: If this is the case, it isn't quite an optional error right? What if we remove this from here and instead in SemanticAnalyzer.java L2188 do 

`(type.tag == TypeTags.ERROR || types.isOptionalErrorType(expExprType))`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
pubudu91: We can't do that right? If it's not an effectively nil returning function/method call expression, we have log the assignment required error, irrespective of the type of the expression.

MaryamZi: What about something like `check x;`? If it is a semantic error, I don't think the "assignment required" error adds much value, and there will already be an error for `x` anyway?

pubudu91: If we don't log it, once we fix the error in `x`, we'll again get the assignment required error the next time we try to run the program, which happened in the reported issue (#24839). Skipping semantic errors was what was done previously as well. The downside to it is you don't get all the errors in one go.

MaryamZi: But in this case, once we fix `x`, it may or may not result in the "assignment required" error, right? I'm all for showing as many errors as possible, but I'm just wondering if logging errors based on possible paths could lead to unnecessary noise? We should of course set the type properly on error.

Also, if we are going to log the "assignment required" error for such scenarios, shouldn't we also log the error when it is an undefined function?

pubudu91: The general case is that the assignment is required right? Considering that this is a case where the user has attempted to use an expression as a statement. Assignments are not required only for specific cases. For example,
```ballerina
public function main() {
    int x = 10;
    checkpanic foo() + x;
}

function foo() returns error? => error("Invalid Value");
```
In the above you won't need the assignment. You just have to fix the error in the expr of the checkpanic expr.

pubudu91: For undefined functions it's not getting logged because here, the return type of undefined functions is taken as `()`. That was done just for simplicity, since I felt there's no point going further than saying that the function is not defined.

MaryamZi: > The general case is that the assignment is required right? Considering that this is a case where the user has attempted to use an expression as a statement. Assignments are not required only for specific cases. For example,
> 
> ```ballerina
> public function main() {
>     int x = 10;
>     checkpanic foo() + x;
> }
> 
> function foo() returns error? => error("Invalid Value");
> ```
> 
> In the above you won't need the assignment. You just have to fix the error in the expr of the checkpanic expr.

I also think the more common case would be to require the assignment, but do we have to make that assumption? I'm not sure I got the example, here the error would be that you can't add nil and int right? And if the user changes `foo` to return `int|error` an assignment will be required, but if they remove `+ x` no assignment is required?



> For undefined functions it's not getting logged because here, the return type of undefined functions is taken as `()`. That was done just for simplicity, since I felt there's no point going further than saying that the function is not defined.

This was the point I was trying to make re: any expression too. If we've already logged appropriate errors, it feels like there's no point giving an error adjusting for code that will definitely have to be changed? What I'm suggesting is to kinda not special case the undefined function scenario, but do the same for all scenarios where the checking-expr's expr itself has errors.

pubudu91: > I also think the more common case would be to require the assignment, but do we have to make that assumption? I'm not sure I got the example, here the error would be that you can't add nil and int right? And if the user changes foo to return int|error an assignment will be required, but if they remove + x no assignment is required?

Sorry, that should've been `checkpanic (foo() + x);`. My point was, the type of the expr here would be semantic error. So can't really figure out anything from that to decide whether there should be an assignment or not. 

> This was the point I was trying to make re: any expression too. If we've already logged appropriate errors, it feels like there's no point giving an error adjusting for code that will definitely have to be changed? What I'm suggesting is to kinda not special case the undefined function scenario, but do the same for all scenarios where the checking-expr's expr itself has errors.

The problem with this is then we get assignment expected errors for nil returning lang lib functions as well. And the same with the example above. So special cased the call-stmt scenario. 
```ballerina
function test1(){
    int x = 0;
    x.forEach(function (int i) { count = count + i;});
}
```

MaryamZi: We discussed this offline, and decided to revisit this if required, once we validate the following is valid according to the spec.

```ballerina
check x;
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/action/start/StartActionTest.java
```diff
@@ -45,6 +45,7 @@ public void testStartActionNegative() {
         BAssertUtil.validateError(result, indx++, "action invocation as an expression not allowed here", 38, 32);
         BAssertUtil.validateError(result, indx++, "action invocation as an expression not allowed here", 39, 32);
         BAssertUtil.validateError(result, indx++, "action invocation as an expression not allowed here", 39, 37);
+        BAssertUtil.validateError(result, indx++, "variable assignment is required", 53, 9);
```
MaryamZi, line 48: This wouldn't get logged if we avoid logging an error if the type is semanticError right? I still think we should continue to not log the error if the type is semanticError. 

Even if we look at this example, we have 3 errors for the same line now, and they aren't quite independent, as in it is not guaranteed the user has to fix them separately.

pubudu91: That again goes to my original point: If we do that, it will take the user two iterations to fix the code. In the first step, the user has to fix the semantic error in the expression itself, and then, has to fix the assignment required error. I think it's better to have all the possible errors in one pass, rather than having to fix them in multiple passes. Basically, for the given piece of code, all possible errors are given, considering each of the components in the line separately:
- `start` can only be used with a function/method/remote method call
- An action cannot be used with wait
- Wait is an expression and needs to be assigned

In this case I think the user does have to address all 3 to make it work. But I agree, there can be instances where the user doesn't have to fix each error individually to make it work. The point is, the user would have all the info at hand to make a decision on which path to take to fix the error.

tests/jballerina-unit-test/src/test/resources/test-src/statements/expression/expression-stmt-negative.bal
```diff
@@ -15,3 +15,18 @@ function test2(){
     bar();
     x;
 }
+
+function baz() returns error|int|() => ();
```
KRVPerera, line 19: Having trouble reproducing this issue. Which ballerina version gives the NPE for this test case?

KRVPerera, line 19: :thinking: Let's change the test case to reflect and NPE.
:pushpin: Line number 19 in the test case masks the NPE.

pubudu91:
Closing this for now. Will resend a PR later with the issue addressed as discussed offline.

Issue #23096, opened 2020-05-05
Service debug points are skipped in ballerina module files inside nested directories


NipunaRanasinghe:
**Description:**
service debug points are skipped for the for module files inside nested directories.

**Steps to reproduce:**
1. Create a ballerina project and add a module
2. Add another directory inside the module and add a ballerina file having an HTTP service.
3. Add breakpoint(s) and try to debug using vscode/intellij IDEA

**Affected Versions:**
1.2.3 and below.

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #24901, opened 2020-07-22
Fix service debbug hit issue

praveennadarajah:
## Purpose
This PR will fix service debug hit issue when a service file is located inside a directory.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23096

### DIFF
--- a/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
+++ b/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
@@ -103,7 +103,7 @@ public class PackageUtils {
                                                   String fileName) {
         // When a bal file is inside a directory, the source path is constructed appropriately for services.
         // But for non service bal files, [directory/file] path occurring twice in source path.
-        // We are handling this by using service regex. If patternMatcher does not contain service regex,
+        // We are handling this by using a service regex. If reference does not contain service regex,
         // appropriate adjustment is made for source path else not.
         Pattern pattern = Pattern.compile(SERVICE_REGEX);
         Matcher matcher = pattern.matcher(reference);
--- a/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
+++ b/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
@@ -93,20 +93,20 @@ public class PackageUtils {
     /**
      * Get relative path when a bal file is inside a directory.
      *
-     * @param sourcePath     source path
-     * @param patternMatcher pattern matcher string
-     * @param sourceName     source name
-     * @param fileName       file name
+     * @param sourcePath    source path
+     * @param reference     reference string
+     * @param sourceName    source name
+     * @param fileName      file name
      * @return relative path
      */
-    public static String getDirectoryRelativePath(String sourcePath, String patternMatcher, String sourceName,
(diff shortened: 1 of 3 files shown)


misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
```diff
@@ -86,4 +89,24 @@ public static boolean isBlank(String str) {
...
+    public static String getDirectoryRelativePath(String sourcePath, String patternMatcher, String sourceName,
+                                                  String fileName) {
+        Pattern pattern = Pattern.compile(SERVICE_REGEX);
+        Matcher matcher = pattern.matcher(patternMatcher);
+        if (matcher.find()) {
+            return sourcePath;
+        } else {
+            return sourcePath.replace(sourceName, fileName);
+        }
+    }
```
NipunaRanasinghe, line 111: @praveennadarajah Shall we add a comment explaining that in here we have handled the service files specifically and the reason behind that?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
```diff
@@ -86,4 +89,24 @@ public static boolean isBlank(String str) {
...
+    /**
+     * Get relative path when a bal file is inside a directory.
+     *
+     * @param sourcePath     source path
+     * @param patternMatcher pattern matcher string
+     * @param sourceName     source name
+     * @param fileName       file name
+     * @return relative path
+     */
+    public static String getDirectoryRelativePath(String sourcePath, String patternMatcher, String sourceName,
```
NipunaRanasinghe, line 102: What do you mean by `patternMatcher` here? AFAIR this the reference location of the source that we are rectifying, right? we can think of using an appropriate arg name instead.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
praveennadarajah: Added comments.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
praveennadarajah: Yes, patternMatcher refers to "reference location" in one instance and "reference type" in another instance. I think renaming patternMatcher to "**reference**" would work fine. WDYT?

NipunaRanasinghe: +1 for `reference`/`referencePath`

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
NipunaRanasinghe: Looks good!

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/utils/PackageUtils.java
praveennadarajah: Renamed to `reference`.

PR #24927, opened 2020-07-24
 Fix test execution logic when before/after functions fail 

### LINKED ISSUE #21091: When there is a failure in after function of a test, the dependant tests are running instead of being skipped
When there is a failure in after function of a test, the dependent tests are running instead of being skipped.

E.g. Executing the below file expects  the result as `1 passing` and `2 skipped`. But the actual output is `3 passing` 

```ballerina
import ballerina/test;
import ballerina/io;

// Tests skipping of dependsOn functions when before func fails.
public function afterFunc() {
    // This will throw an exception and fail the before function
    int i = 12/0;
    io:println("After");
}

@test:Config {
    after: "afterFunc"
}
public function test1() {
    io:println("test1");
}

// This should be skipped
@test:Config {
    dependsOn:["test1"]
}
public function test2() {
    io:println("test2");
}

// This test should pass
@test:Config {}
public function test3() {
    io:println("test3");
}

```

azinneera:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


cli/ballerina-packerina/src/test/resources/test-resources/valid-project/src/mytemplate/tests/main_test.bal
```diff
@@ -33,7 +33,7 @@ function afterFunc () {
 
 # After Suite Function
 
-@test:AfterSuite
+@test:AfterSuite {}
 function afterSuiteFunc () {
     io:println("I'm the after suite function!");
 }
```
hevayo, line 39: Better to add a new line at the end there are some more files you need to fix.

azinneera: Added to necessary files in f6db6be

misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestAnnotationProcessor.java
```diff
@@ -148,7 +149,22 @@ public void process(FunctionNode functionNode, List<AnnotationAttachmentNode> an
             if (BEFORE_SUITE_ANNOTATION_NAME.equals(annotationName)) {
                 suite.addBeforeSuiteFunction(functionName);
             } else if (AFTER_SUITE_ANNOTATION_NAME.equals(annotationName)) {
-                suite.addAfterSuiteFunction(functionName);
+                AtomicBoolean alwaysRun = new AtomicBoolean(false);
+                if (attachmentNode.getExpression() instanceof BLangRecordLiteral) {
```
anupama-pathirage, line 153: Instead of `instanceof` can we check the node kind? i.e. `attachmentNode.getExpression().getKind() == NodeKind.RECORD_LITERAL_EXP`

misc/testerina/modules/testerina-runtime/src/main/java/org/ballerinalang/test/runtime/BTestRunner.java
```diff
@@ -406,41 +411,53 @@ private void executeFunction(Test test, TestSuite suite, String packageName, Cla
...
-                                                  " [after each test function for the test %s] :\n\t    %s",
-                                          test, formatErrorMessage(e));
-                errStream.println(errorMsg2);
-            }
-        });
+    private void executeAfterEachFunction(Test test, TestSuite suite, ClassLoader classLoader, Scheduler scheduler,
+                                          AtomicBoolean shouldSkip, AtomicBoolean shouldSkipTest) {
+        if (!shouldSkip.get() && !shouldSkipTest.get()) {
+            suite.getAfterEachFunctionNames().forEach(afterEachTest -> {
+                String errorMsg2;
```
anupama-pathirage, line 435: Shall we have a more meaningful variable name here?

azinneera: Fixed in f9d41bc

misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestAnnotationProcessor.java
azinneera: Fixed in f9d41bc

PR #24931, opened 2020-07-24
Fix parser doesn't support empy list binding pattern

### LINKED ISSUE #24648: Parser doesn't support empy list binding pattern.
**Description:**
$Subject

- Component/Parser
- Type/Bug

rdulmina:
## Purpose

- $Subject
- Enable ExprBodiedFunctionTest.testSyntaxErrors 

Fixes #24648

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -14253,6 +14253,14 @@ private STNode parseListBindingPattern() {
     }
 
     private STNode parseListBindingPattern(STNode openBracket, List<STNode> bindingPatternsList) {
+        if (isEndOfListBindingPattern(peek().kind) && bindingPatternsList.size() == 0) {
```
SupunS, line 14256: Do we need to check whether the size is zero?

rdulmina: Yes, this block is only to handle empty case. and there are some cases in which the list is not empty and the next token is `close_bracket' which will produce the wrong results if we don't use that check.

PR #24933, opened 2020-07-24
Document BIR spec and add BIR dump file flag

manuranga:


cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/BuildCommand.java
```diff
@@ -149,6 +150,9 @@ public BuildCommand(Path userDir, PrintStream outStream, PrintStream errStream,
     @CommandLine.Option(names = "--dump-bir", hidden = true)
     private boolean dumpBIR;
 
+    @CommandLine.Option(names = "--dump-bir-file", hidden = true)
```
sameerajayasoma, line 153: How about --dump-bir-binary?

settings.gradle
```diff
@@ -130,6 +130,7 @@ include(':build-config:checkstyle')
 include(':debug-adapter:debug-adapter-core')
 include(':debug-adapter:debug-adapter-cli')
 include(':ballerina-libs')
+include(':bir-spec')
```
sameerajayasoma, line 133: Do we need to build this every time?

cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/cmd/BuildCommand.java
Kishanthan: @sameerajayasoma , looks like with the option `--dump-bir-file` you have to give output file name/location. it is not a boolean flag apparently

manuranga:
will be addressed in nballerina branch and merged to master later.

PR #24934, opened 2020-07-24
Fix actions being allowed in expressions

### LINKED ISSUE #23037: Actions are allowed inside expressions
**Description:**

Below code parsers without any errors:

```ballerina
function foo() {
    x = a -> b() + c;
}
```

m36dot:
## Purpose
To allow only expression where expressions are expected and add a diagnostic if an action is provided

Fixes #23037

## Approach
Added a check in parseExpression to see if the parsed node is an action or expression and added a diagnostic if action is returned where it isn't supposed to be. Modified existing test asserts to go with the new change.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -4101,9 +4098,34 @@ private STNode parseExpression(OperatorPrecedence precedenceLevel, boolean isRhs
...
+                parseExpressionRhs(precedenceLevel, expr, isRhsExpr, allowActions, isInMatchGuard, isInConditionalExpr);
+        //braced actions are just paranthesis enclosing actions, no need to add a diagnostic there when we have added
+        //diagnostics to its children
+        if (!allowActions && isAction(actionOrExpression) && actionOrExpression.kind != SyntaxKind.BRACED_ACTION) {
+            actionOrExpression = attachErrorExpectedActionFoundDiagnostic(actionOrExpression);
+        }
+        return actionOrExpression;
     }
 
+    private boolean containsDiagnostics(STNode node, DiagnosticErrorCode diagnosticErrorCode) {
```
SupunS, line 4111: Why do we need this check?
Can the same error be logged in two places? If so, Would removing the error from the other place solve the problem?

m36dot: no I dont think that will work, because there are loops and multiple entry points for the parseExpression method.

m36dot: I ll try to get it working like that

Issue #24598, opened 2020-07-06
Add jBallerina debugger tests for Ballerina services


praveennadarajah:
**Description:**
Need to add Ballerina debugger tests for Ballerina services.

PR #24947, opened 2020-07-27
Add debug engage test for services

praveennadarajah:
## Purpose
This PR will add debug engage test for ballerina service.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/24598


tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/ServiceDebugTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+        runDebuggeeProgram(projectPath, port);
+        addBreakPoint(new BallerinaTestDebugPoint(filePath, 36));
+        initDebugSession(null, port);
+
+        // Test for service debug engage
+        Pair<BallerinaTestDebugPoint, StoppedEventArguments> debugHitInfo = waitForDebugHit(20000);
+        Assert.assertEquals(debugHitInfo.getLeft(), testBreakpoints.get(0));
+
+        terminateDebugSession();
```
azinneera, line 66: Shall we add this to an `@AfterClass` method? If the test throws an error before executing this line, the termination would not take place. When we define the annotation we should define it as `@AfterClass(alwaysRun=true)`.

tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/ServiceDebugTest.java
```diff
@@ -0,0 +1,68 @@
...
+    private String projectPath;
+
+    @BeforeClass
+    public void setup() {
+        testProjectName = "breakpoint-tests";
+        testModuleName = "myService";
+        projectPath = testProjectBaseDir + File.separator + testProjectName;
+
+        testModuleFileName = "tests" + File.separator + "hello_service_test.bal";
+        testProjectPath = testProjectBaseDir.toString() + File.separator + testProjectName;
```
azinneera, line 48: Shall we change the above 2 lines to use `Paths.get` since `File.sperator` is a pre-1.7 JDK way

tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/ServiceDebugTest.java
NipunaRanasinghe: @azinneera I think here we have a requirement to terminate (and re-initiate) the debug session for each test case. Therefore shall we consider using `@AfterMethod` instead?

azinneera: +1

tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/ServiceDebugTest.java
praveennadarajah: Fixed.

tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/ServiceDebugTest.java
praveennadarajah: Fixed.

Issue #25016, opened 2020-07-29
Cannot define a normal `transactional` method


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type Foo client object {
    remote transactional function baz() {  // no error
    }

    transactional function qux() { // error: .::zxa.bal:5:19: invalid token 'transactional'
    }
};
```

**Affected Versions:**
Current master (slp3-SNAPSHOT)

SupunS:
Needs to be fixed along with #24221

SupunS:
Fixed with #25078

Issue #24992, opened 2020-07-28
REMOTE flag not set for remote method declarations 


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
Compilation fails for the following due to the `REMOTE` flag mismatch.

```ballerina
import ballerina/io;

type Foo abstract client object {
    remote function foo();
};

type Bar client object {
    remote function foo() {
        io:println("Bar->foo()");
    }
};

public function main() {
    Foo f = new Bar(); // error: .::qwe.bal:14:2: incompatible types: expected 'Foo', found 'Bar'
    f->foo();
}
```

Works with the old parser.

**Affected Versions:**
slp2

SupunS:
Fixed with #25078

PR #25078, opened 2020-07-31
Improve 'transactional' functions parsing

### LINKED ISSUE #24221: Improve 'transactional' functions parsing
**Description:**

$subject.

rdulmina:
## Purpose

- Improve 'transactional' functions parsing.
- Use a list to contain function qualifiers.
- Add a single method to parse function qualifiers.

Fixes #24221, #25016 #24992

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -1460,28 +1461,21 @@ private STNode parseFuncDefinition(STNode metadata, boolean isObjectMethod, STNo
...
      *
      * @param metadata Metadata
-     * @param visibilityQualifier Visibility qualifier
+     * @param qualifiers qualifier list
      * @return Parsed node
      */
-    private STNode parseFuncDefOrFuncTypeDesc(STNode metadata, boolean isObjectMethod, STNode... qualifiers) {
-        parseTransactionalQUalifier(qualifiers);
+    private STNode parseFuncDefOrFuncTypeDesc(STNode metadata, boolean isObjectMethod, List<STNode> qualifiers) {
+        if (peek().kind == SyntaxKind.TRANSACTIONAL_KEYWORD) {
```
SupunS, line 1468: If the token is neither `transactional` or `function`, we'll have to add recovery here

rdulmina: Added.

misc/testerina/modules/testerina-core/src/main/ballerina/Ballerina.toml
```diff
@@ -8,5 +8,5 @@ target = "java8"
     [[platform.libraries]]
         artifactId = "mock"
         version = "0.0.0"
-        path = "./lib/testerina-core-2.0.0-Preview2.jar"
+        path = "./lib/testerina-core-2.0.0-Preview3-SNAPSHOT.jar"
```
lochana-chathura, line 11: Do we need this change?

tests/jballerina-unit-test/ballerina-internal.log.1
```diff
@@ -0,0 +1,720 @@
+[2020-08-07 12:25:46,714] SEVERE {b7a.log.crash} - ballerina/lang_annotations/1_0_0/___init
```
lochana-chathura, line 1: this file also?

compiler/ballerina-parser/src/test/resources/declarations/object-type-def/object_type_def_source_41.bal
```diff
@@ -0,0 +1,14 @@
...
+     remote function foo0(){
+
+     }
+     transactional function foo1() {
+
+     }
+     function foo2() {
+
+     }
+};
```
lochana-chathura, line 14: Add new line :)

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -5636,6 +5662,97 @@ private STNode parseObjectMethodOrField(SyntaxKind nextTokenKind, SyntaxKind nex
...
+                    qualifier = parseRemoteKeyword();
+                    break;
+                case TRANSACTIONAL_KEYWORD:
+                    qualifier = parseTransactionalKeyword();
+                    break;
+                case RESOURCE_KEYWORD:
+                    qualifier = parseResourceKeyword();
+                    break;
+                default:
+                    Solution solution = recover(peek(), context, context, qualifierList);
```
lochana-chathura, line 5690: these qualifiers are always optional right? Do we need to go to the recovery?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -5636,6 +5662,97 @@ private STNode parseObjectMethodOrField(SyntaxKind nextTokenKind, SyntaxKind nex
...
+    }
+
+    private boolean isEndOfFunctionQualifiers(SyntaxKind tokenKind) {
+        switch (tokenKind) {
+            case FUNCTION_KEYWORD:
+            case EOF_TOKEN:
+                return true;
+            default:
+                return false;
+        }
```
lochana-chathura, line 5726: what if the function keyword missing? we'll end up in an infinite loop right as function keyword is not handled inside `parseFunctionQualifiers`?

I think we may check for `REMOTE, TRANSACTIONAL, RESOURCE` keywords and else return true. WDYT?

compiler/ballerina-parser/src/test/resources/declarations/object-type-def/object_type_def_source_41.bal
lochana-chathura, line 14: add a new line :)

misc/testerina/modules/testerina-core/src/main/ballerina/Ballerina.toml
rdulmina: We don't need that. Thanx for mentioning.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -5550,19 +5573,19 @@ private STNode parseObjectMember(SyntaxKind nextTokenKind, STNode metadata) {
...
+                member = parseObjectMethod(metadata, new ArrayList<>());
                 break;
             default:
                 if (isTypeStartingToken(nextTokenKind)) {
                     member = parseObjectField(metadata, STNodeFactory.createEmptyNode());
                     break;
                 }
 
-                Solution solution = recover(peek(), ParserRuleContext.OBJECT_MEMBER_WITHOUT_METADATA);
+                Solution solution = recover(peek(), ParserRuleContext.OBJECT_MEMBER_WITHOUT_METADATA, metadata);
```
lochana-chathura, line 5588: Shall we also modify ```return parseObjectMember(solution.tokenKind);``` beneath. Otherwise we'll be re-parsing with the wrong method I think.

tests/jballerina-unit-test/ballerina-internal.log.1
rdulmina: thanx

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
rdulmina: Yes there can be `invalid tokens` before qualifier or missing `function` keyword. Those things will recover here.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
rdulmina: Missing `function` keyword will recover inside the recovery of `parseFunctionQualifiers()` method.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
rdulmina: U mean to rename the functions?

rdulmina: renamed the functions

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
lochana-chathura: I see. btw In case there's a missing function key we only skip the while loop. And do the qualifier validation again. 
This will be a problem, specially when there's an error in the qualifier. 
e.g  For transactional  transactional <missing function>, updateLastNodeInListWithInvalidNode will call twice for same error ryt?
I think we can simply return if the solution.kind is function keyword. WDYT?

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
lochana-chathura: not exactly the renaming of method. Earlier we had called the wrong overloaded method (with missing metadata arg) :)

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
rdulmina: Validation is inside the while loop, so we do simply return if the function keyword recovered.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
rdulmina: Yes, thanx for mentioning I have changed that.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -5636,6 +5662,97 @@ private STNode parseObjectMethodOrField(SyntaxKind nextTokenKind, SyntaxKind nex
...
+                    qualifier = parseResourceKeyword();
+                    break;
+                default:
+                    Solution solution = recover(peek(), context, context, qualifierList);
+
+                    // If the parser recovered by inserting a token, then try to re-parse the same
+                    // rule with the inserted token. This is done to pick the correct branch
+                    // to continue the parsing.
+                    if (solution.action == Action.REMOVE) {
+                        return solution.recoveredNode;
```
SupunS, line 5696: Can returning immediately be a problem?
eg:
```ballerina
transactional xyz resource function foo() {
}
```
Here `xyz` is an invalid token. After recovering from it, shouldn't the loop continue?

rdulmina: No, In the case of invalid token after recovery it goes to resume parsing. Then it parses the rest of the qualifiers and return here so we return immediately right?

SupunS: ah ok. Have any tests covering that?

rdulmina: Yes already added recovery test cases for fuction-def, object-method and service resource method. Those tests contains your example as well.

PR #25118, opened 2020-08-04
Improve query-expr and query-action parsing

### LINKED ISSUE #24923: Improve query expressions parsing
**Description:**

- There are few TODOs to be address
- Improve order by clause parsing (there are some redundant checks in places like parsing ascending/descending keywords)
- Improve order by clause recovery

lochana-chathura:
## Purpose
$subject.

Fixes #24923 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -10918,6 +10818,8 @@ private STNode parseJoinClause(boolean isRhsExpr) {
      */
     private STNode parseOnClause(boolean isRhsExpr) {
         STNode onKeyword = parseOnKeyword();
+        // note that parsing expression includes following.
+        // equals-expr := expression `equals` expression
         STNode expression = parseExpression(OperatorPrecedence.QUERY, isRhsExpr, false);
```
lochana-chathura, line 10823: * `join-on-condition := on expression equals expression` from spec is parsed as `on-clause := on expression`.
i.e. we parse the `equals` keyword implicitly.  

* `equals` operator was added to error handler as well since it is already supported in the parser. 
There's no such operator in the spec though.

lochana-chathura:
Also noticed a change coming from upstream in which the invalid node in leading minutiae get flattened.  
Assertion updated for that change with 6ed3200

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -10693,13 +10630,10 @@ private STNode parseOrderKeyList(boolean isRhsExpr) {
 
         // Parse the remaining order keys
         nextToken = peek();
-        STNode orderKeyListMemberEnd;
+        STNode leadingComma;
         while (!isEndOfOrderKeys(nextToken.kind)) {
-            orderKeyListMemberEnd = parseOrderKeyListMemberEnd();
-            if (orderKeyListMemberEnd == null) {
-                break;
-            }
```
SupunS, line 10701: Wouldn't we need this?
For eg: In case when theres an additional after the order-keys, but not one of the "EndOfOrderKeys" tokens ..?

lochana-chathura:
```java    
private boolean isEndOfOrderKeys(SyntaxKind tokenKind) {
        switch (tokenKind) {
            case COMMA_TOKEN:
            case ASCENDING_KEYWORD:
            case DESCENDING_KEYWORD:
                return false;
            default:
                return !isTypeStartingToken(tokenKind);
        }
    }
```
If there's an additional token and not one of end of order keys it should be one of following.
* comma token - recover with missing identifer
* ascending/ descending token - recover with missing ident 
* type start token - it is considered as an expr

Can there be any other case?

lochana-chathura:
An integration test failed due to a time out. Hence re-triggering builds

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
SupunS: I think not all type-starting-tokens are expressions..
I was thinking, we can terminate if the next token is an end of order keys for sure. If not, try to parse the "parseOrderKeyListMemberEnd" and let the error recovery handle whether to terminate or continue.. WDYT?

lochana-chathura: yea. that's better. Fixed in a434d8b

PR #25124, opened 2020-08-04
Align mapping constructor spread-expr analysis with the spec

### LINKED ISSUE #22504: Align mapping constructor spread-expr analysis with the spec
**Description:**
The spec now includes the spread-expr - see https://ballerina.io/spec/lang/2020R1/#mapping-constructor-expr.

The spec says
```
It is a compile-time error if the static type of the expression in a spread-field 
allows a field that duplicates a specific-field or that could also occur in 
another spread-field. Note that a spread-field with an inclusive record type 
of record { never x?; } cannot duplicate a specific field for x.
```

There are some changes required in the implementation to make it fully spec compliant as discussed in https://github.com/ballerina-platform/ballerina-spec/issues/479.

**Case I**
The following should be **disallowed** since even though field `i` is optional in `Foo` it could still occur.

```ballerina
type Foo record {|
    int i?;
|};

public function main() {
    Foo f = {i: 1};

    map<int> m = {i: 0, ...f};
}
```

**Case II**
As spread-fields we cannot use values of inclusive-record types such as `Bar` in the following example, since `b` could have a field with the key `i`.

```ballerina
type Bar record {
    int j;
};

public function main() {
    Bar b = {j: 1, "i": 2};

    map<anydata> m = {i: 0, ...b};
}
```

**Case III**
But a value of an inclusive-record type can be used if there are `never x?;` fields for each `x` that has already occurred/could have occurred in the mapping constructor expression.
```ballerina
type Bar record {
    int j;
    never i?;
};

public function main() {
    Bar b = {j: 2, "k": 3};

    map<anydata> m = {i: 0, ...b}; // allowed now
}
```

**Case IV**
This also means that in general you cannot have more than one spread-field of a type that has an inclusive-type-descriptor.

```ballerina
type Foo record {
    int i;
    never j?;
};

type Bar record {
    never i?;
    int j;
};

public function main() {
    Foo f = {i: 1, "k": 4};
    Bar b = {j: 2, "k": 4};

    map<anydata> m = {...f, ...b}; // `k` duplicated
}
```

**Affected Versions:**
1.2.1

dulajdilshan:
Fixes #22504 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2043,17 +2045,39 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                LinkedHashMap<String, BField> fieldsInRecordLiteral = ((BRecordType) spreadOpExpr.type).fields;
+                for (Object fieldName : names) {
+                    if (fieldsInRecordLiteral.containsKey(fieldName)) {
+                        if (fieldsInRecordLiteral.get(fieldName).type.tag != TypeTags.NEVER) {
+                            // An error is logged if the same field is occurred, and it is not defined as a never-type
+                            this.dlog.error(spreadOpExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL_SPREAD_OP,
+                                    recordLiteral.expectedType.getKind().typeName(), fieldName, spreadOpField);
+                        }
+                    } else {
+                        if (!((BRecordType) spreadOpExpr.type).sealed) {
```
MaryamZi, line 2067: Can extract `((BRecordType) spreadOpExpr.type).sealed` out to a variable, right? Instead of accessing the `sealed` field in each iteration?

MaryamZi: Shall we add a comment here to say we only get here if there are other fields.

And if we don't already have one, please add a test for something like

```ballerina
public function main() {
    record {
        int x;
    } c = {x: 3, "y": 2.0};
    map<anydata> n = {...c};
}
```

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -956,6 +956,12 @@ error.duplicate.key.in.table.literal=\
 error.duplicate.key.in.record.literal.spread.op=\
   invalid usage of {0} literal: duplicate key ''{1}'' via spread operator ''{2}''
 
+error.inclusive.record.literal.with.spread.op=\
+  invalid usage of {0} literal: inclusive record ''{1}'' with spread operator is not allowed
```
MaryamZi, line 960: Let's fix this also for mapping constructor expression and maps. And IMO we should say "spread field."

tests/jballerina-unit-test/src/test/resources/test-src/expressions/mappingconstructor/mapping_constructor_infer_record.bal
```diff
@@ -231,30 +231,26 @@ type Rec1 record {
...
-type Rec2 record {
-    string i?;
+type Rec2 record {|
+    string i;
     float f?;
-};
+|};
 
 function testInferredRecordTypeWithOptionalTypeFieldViaSpreadOp() {
-    Rec1 rec1 = {i: 1, b: true};
```
MaryamZi, line 240: Shall we remove the `Rec1` type definition if it is no longer used? And rename `Rec2` to `Rec1` maybe?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/mappingconstructor/MappingConstructorExprTest.java
```diff
@@ -212,7 +212,7 @@ public void testSpreadOpField(String test) {
                 { "testSpreadOpInGlobalMap" },
                 { "testMappingConstrExprAsSpreadExpr" },
                 { "testOrderWithSpreadOp" },
-                { "testInherentTypeViolationViaSpreadOp" }
+//                { "testInherentTypeViolationViaSpreadOp" }
```
MaryamZi, line 215: Why are we commenting this out instead of removing/updating?

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
```diff
@@ -290,6 +290,8 @@
     DUPLICATE_KEY_IN_RECORD_LITERAL("duplicate.key.in.record.literal"),
     DUPLICATE_KEY_IN_TABLE_LITERAL("duplicate.key.in.table.literal"),
     DUPLICATE_KEY_IN_RECORD_LITERAL_SPREAD_OP("duplicate.key.in.record.literal.spread.op"),
+    INCLUSIVE_RECORD_LITERAL_WITH_SPREAD_OP("inclusive.record.literal.with.spread.op"),
```
MaryamZi, line 293: Saying "inclusive record literal" is not quite right, right? I think it should be "inclusive record type" or with maps "inclusive mapping type".

compiler/ballerina-lang/src/main/resources/compiler.properties
KRVPerera: I think this error is not descriptive enough. We should also say why not allowed or a suggestion to use `never`. 
Currently, this may give an idea that in general, we cannot use inclusive records with spread operator.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -956,6 +956,12 @@ error.duplicate.key.in.table.literal=\
...
+
+error.more.than.one.inclusive.types=\
+  invalid usage of {0} literal: using more than one inclusive type is not allowed.
```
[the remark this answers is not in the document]
KRVPerera: Shouldn't this be plural `inclusive mapping types are not allowed`, we can also replace `using more than one spread field` to `multiple spread fields`.
So in overall.
`multiple spread fields of inclusive mapping types are not allowed`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2043,17 +2045,39 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                            this.dlog.error(spreadOpExpr.pos, DiagnosticCode.INCLUSIVE_RECORD_LITERAL_WITH_SPREAD_OP,
+                                    recordLiteral.expectedType.getKind().typeName(), spreadOpExpr);
+                        }
                     }
+                }
 
+                for (BField bField : ((BRecordType) spreadOpExpr.type).fields.values()) {
```
KRVPerera, line 2074: 1. Do we need two for loops here?
2. I believe logic assumes we can analyze this iteratively. But I think we should think about final resultant mapping constructor expression. So we may need two passes and maybe three to give errors as well.
#### Case A: Errors and duplicate errors
```Ballerina
import ballerina/io;

type Foo record {
    int i;
};

type Bar record {|
    int j;
    int k;
|};

public function main() {
    Foo f = {i: 1, "k": 7};
    Bar b = {j: 2, "k": 4};

    map<anydata> m = {...b, ...f};    // b first
    io:println("m : ", m);
}
```
```LogTalk
        case5.bal
error: .::case5.bal:16:32: invalid usage of map literal: inclusive record 'f' with spread operator is not allowed
error: .::case5.bal:16:32: invalid usage of map literal: inclusive record 'f' with spread operator is not allowed
```

#### Case B: No errors (Same code with a different order in spread operator fields)

```Ballerina
import ballerina/io;

type Foo record {
    int i;
};

type Bar record {|
    int j;
    int k;
|};

public function main() {
    Foo f = {i: 1, "k": 7};
    Bar b = {j: 2, "k": 4};

    map<anydata> m = {...f, ...b};  // a first
    io:println("m : ", m);
}
```
```LogTalk
m : i=1 k=4 j=2
```

KRVPerera, line 2074: 1. Do we need two for loops here?
2. I believe logic assumes we can analyze this iteratively. But I think we should think about final resultant mapping constructor expression. So we may need two passes and maybe three to give errors as well.
#### Case A: Errors and duplicate errors
```Ballerina
import ballerina/io;

type Foo record {
    int i;
};

type Bar record {|
    int j;
    int k;
|};

public function main() {
    Foo f = {i: 1, "k": 7};
    Bar b = {j: 2, "k": 4};

    map<anydata> m = {...b, ...f};    // b first
    io:println("m : ", m);
}
```
```LogTalk
        case5.bal
error: .::case5.bal:16:32: invalid usage of map literal: inclusive record 'f' with spread operator is not allowed
error: .::case5.bal:16:32: invalid usage of map literal: inclusive record 'f' with spread operator is not allowed
```

#### Case B: No errors (Same code with a different order in spread operator fields)

```Ballerina
import ballerina/io;

type Foo record {
    int i;
};

type Bar record {|
    int j;
    int k;
|};

public function main() {
    Foo f = {i: 1, "k": 7};
    Bar b = {j: 2, "k": 4};

    map<anydata> m = {...f, ...b};  // a first
    io:println("m : ", m);
}
```
```LogTalk
m : i=1 k=4 j=2
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2070,6 +2126,9 @@ public void visit(BLangRecordLiteral recordLiteral) {
                     if (names.contains(name)) {
                         this.dlog.error(keyExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL,
                                         recordLiteral.expectedType.getKind().typeName(), name);
+                    } else if (noOfInclusiveTypes > 0 && !neverTypedKeys.contains(name)) {
+                        this.dlog.error(keyExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL,
```
MaryamZi, line 2130: Isn't this a possible duplication? Not definite, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2035,21 +2038,74 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                    if (!names.contains(name)) {
+                        if (bField.type.tag == TypeTags.NEVER) {
+                            if (!neverTypedKeys.contains(name)) {
+                                neverTypedKeys.add(name);
+                            }
+                        } else {
+                            if (neverTypedKeys.contains(name)) {
+                                neverTypedKeys.remove(name);
+                            } else {
+                                if (noOfInclusiveTypes > 0 && spreadExprRecordType.sealed) {
```
MaryamZi, line 2100: Why do we need to check `spreadExprRecordType.sealed`? Is it to avoid multiple errors?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2035,21 +2038,74 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
 
                 analyzeExpr(spreadOpExpr);
+
+                BType spreadOpExprType = spreadOpExpr.type;
+
+                if (spreadOpExprType.tag == TypeTags.MAP) {
+
+                    if (noOfInclusiveTypes > 0) {
+                        this.dlog.error(spreadOpExpr.pos, DiagnosticCode.MULTIPLE_INCLUSIVE_TYPES,
+                                recordLiteral.expectedType.getKind().typeName());
```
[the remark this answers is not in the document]
dulajdilshan: Yeah. See the below example

```ballerina
type Foo record {
    int i;
    never j?;
};

type Bar record {
    never i?;
    int j;
};

public function main() {
    Foo f = {i: 1, "k": 4};
    Bar b = {j: 2, "k": 4};

    map<anydata> m = {...f, ...b}; // `k` duplicated
}
```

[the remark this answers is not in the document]
dulajdilshan: No.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2035,21 +2038,74 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                    // More than one inclusive-type-descriptors with spread-fields are not allowed.
+                    if (noOfInclusiveTypes > 0) {
+                        this.dlog.error(spreadOpExpr.pos, DiagnosticCode.MULTIPLE_INCLUSIVE_TYPES,
+                                recordLiteral.expectedType.getKind().typeName());
+                    } else {
+                        noOfInclusiveTypes = noOfInclusiveTypes + 1;
                     }
+                }
 
+                LinkedHashMap<String, BField> fieldsInRecordLiteral = ((BRecordType) spreadOpExpr.type).fields;
```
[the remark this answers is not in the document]
dulajdilshan: What about `fieldsInRecordType`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
dulajdilshan: How about a scenario like this?
```ballerina
type Foo record {|
    int i;
    int...;
|};

public function main() {
    Foo f = {i: 3, "k": 1};
    map<int> m = {...f, k: 2};
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2035,21 +2038,74 @@ public void visit(BLangRecordLiteral recordLiteral) {
                 BLangExpression spreadOpExpr = spreadOpField.expr;
 
                 analyzeExpr(spreadOpExpr);
+
+                BType spreadOpExprType = spreadOpExpr.type;
+
+                if (spreadOpExprType.tag == TypeTags.MAP) {
```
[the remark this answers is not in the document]
dulajdilshan: We only have two places where we use `spreadOpExprType.tag`. Do we really need to do that ?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
dulajdilshan: Yeah.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
MaryamZi: The error we currently have is "invalid usage of {0} literal: duplicate key ''{1}''".

What I meant is, `f` here may or may not have a field with key `k`. So saying duplicate key here may not always be correct, right?

Shall we say something along the lines of "possible duplicate"?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/mappingconstructor/spread_op_field_code_analysis_negative.bal
```diff
@@ -31,3 +31,41 @@ function testDulpicateKeyWithSpreadOpField() {
...
+}
+
+function testInclusiveRecordWithSpreadOp2() {
+    Beta b = {j: 1, "i": 2};
+    map<anydata> m = {...b, i:3};  // invalid usage of map literal: duplicate key 'i'.
+}
+
+function testMappWithSpreadOp() {
+    map<string> m1 = {x:"aa", y:"bb"};
+    map<string> m2 = {...m1, y:"cc"};   // mapping constructor expression: inclusive mapping 'm1' may consist of already-occurred keys.
```
MaryamZi, line 60:
```suggestion
    map<string> m1 = {x: "aa", y: "bb"};
    map<string> m2 = {...m1, y: "cc"};   // mapping constructor expression: inclusive mapping 'm1' may consist of already-occurred keys.
```

Please check and fix everywhere.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/mappingconstructor/spread_op_field.bal
```diff
@@ -60,6 +58,19 @@ public function testRecordRefAsSpreadOp() {
     assertEquality(123.4, f["f"]);
 }
 
+type Bla record {
```
MaryamZi, line 61: Shall we use something like `Quux`? :)

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
MaryamZi: Updating your other error a bit, how about something like

`invalid usage of mapping constructor expression: spread field ''{0}'' may have already specified key ''{1}''`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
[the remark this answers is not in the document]
MaryamZi: Not an issue, but do you see a drawback in doing so? I personally usually draw the line at > 1.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/mappingconstructor/spread_op_field.bal
dulajdilshan: okay :smile:

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
dulajdilshan: I have changed to `invalid usage of mapping constructor expression: spread field ''{0}'' may have already specified keys`. Seems like it is appropriate.

MaryamZi (commented):
Please consider the following scenarios.

**Case I**

The following doesn't fail.

```ballerina
type Foo record {
    int i?; // doesn't fail even when this is a required field
};

public function main() {
    map<int> m = {
        i: 2
    };

    Foo f = {i: 1, ...m}; // `m` duplicates `i`
}
```

**Case II**

The following results in two errors. IMO just the first one is enough.

```ballerina
type Foo record {
    int i;
};

type Bar record {
    int i?;
};

public function main() {
    Foo f = {i: 1};
    Bar b = {};

    map<anydata> m = {...f, ...b};
}
```

```cmd
error: .::qwe.bal:13:32: invalid usage of mapping constructor expression: multiple spread fields of inclusive mapping types are not allowed
error: .::qwe.bal:13:32: invalid usage of map literal: duplicate key 'i' via spread operator '...b'
```

**Case III**
Issue with the error when non-spread fields are specified after the spread field.

```ballerina
public function main() {
    map<int> f = {};
    
    int i = 1;
    map<int> m = {...f, i};
}
```

```cmd
error: .::qwe.bal:7:29: invalid usage of mapping constructor expression: spread field 'f' may have already specified keys
```

IMO, the error here should be for `i`, not for the spread field, since it is `i` that may duplicate what is already specified via the spread field.

**Case IV**

The error with optional fields is also "duplicate key 'x'". IMO, we can improve this by saying "possible duplicate" for optional fields. This is an improvement, we can even do this separately.

```ballerina
type Foo record {
    int i?;
};

public function main() {
    Foo f = {};

    map<anydata> m = {i: 1, ...f};
}
```

```cmd
error: .::qwe.bal:8:32: invalid usage of map literal: duplicate key 'i' via spread operator '...f'
```

Same for something like 

```ballerina
type Foo record {
    int i?;
};

public function main() {
    Foo f = {};

    map<anydata> m = {...f, i: 2};
}
```
where the error is currently

```cmd
error: .::qwe.bal:8:29: invalid usage of map literal: duplicate key 'i'
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2249,21 +2252,76 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                    // More than one inclusive-type-descriptors with spread-fields are not allowed.
+                    if (inclusiveTypeSpreadField != null) {
+                        this.dlog.error(spreadOpExpr.pos, DiagnosticCode.MULTIPLE_INCLUSIVE_TYPES);
+                        break;
+                    } else {
+                        inclusiveTypeSpreadField = spreadOpField;
                     }
+                }
 
+                LinkedHashMap<String, BField> fieldsInRecordType = ((BRecordType) spreadOpExpr.type).fields;
```
MaryamZi, line 2292:
```suggestion
                LinkedHashMap<String, BField> fieldsInRecordType = spreadExprRecordType.fields;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2249,21 +2252,76 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
-                for (BField bField : ((BRecordType) spreadOpExpr.type).fields.values()) {
-                    if (Symbols.isOptional(bField.symbol)) {
-                        continue;
+                BRecordType spreadExprRecordType = (BRecordType) spreadOpExprType;
+                boolean isSpreadExprRecordTypeSealed = spreadExprRecordType.sealed;
+                if (!isSpreadExprRecordTypeSealed) {
+                    // More than one inclusive-type-descriptors with spread-fields are not allowed.
+                    if (inclusiveTypeSpreadField != null) {
+                        this.dlog.error(spreadOpExpr.pos, DiagnosticCode.MULTIPLE_INCLUSIVE_TYPES);
+                        break;
```
MaryamZi, line 2286: Shouldn't we let the rest of the checks run here, without breaking? 

For example, in the following scenario we can give the 
`invalid usage of record literal: duplicate key 'i' via spread operator '...r2'` error too, right?

```ballerina
public function main() {
    record {
        int i;
    } r1 = {i: 1};

    record {
        int i;
    } r2 = {i: 1};  

    var x = {...r1, ...r2};
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2249,21 +2252,76 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                for (Object fieldName : names) {
+                    if (fieldsInRecordType.containsKey(fieldName)) {
+                        if (fieldsInRecordType.get(fieldName).type.tag != TypeTags.NEVER) {
+                            // An error is logged if the same field is occurred, and it is not defined as a never-type
+                            this.dlog.error(spreadOpExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL_SPREAD_OP,
+                                    recordLiteral.expectedType.getKind().typeName(), fieldName, spreadOpField);
+                        }
+                    } else {
+                        if (!isSpreadExprRecordTypeSealed) {
+                            // Log an error if the spread-field has a inclusive type
```
MaryamZi, line 2302: Do we need this comment now that the check is kinda self-explanatory?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2249,21 +2252,76 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                    } else {
+                        inclusiveTypeSpreadField = spreadOpField;
                     }
+                }
 
+                LinkedHashMap<String, BField> fieldsInRecordType = ((BRecordType) spreadOpExpr.type).fields;
+                for (Object fieldName : names) {
+                    if (fieldsInRecordType.containsKey(fieldName)) {
+                        if (fieldsInRecordType.get(fieldName).type.tag != TypeTags.NEVER) {
+                            // An error is logged if the same field is occurred, and it is not defined as a never-type
```
MaryamZi, line 2296: IMO we don't need this comment either.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2249,21 +2252,76 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                            this.dlog.error(spreadOpExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL_SPREAD_OP,
+                                    recordLiteral.expectedType.getKind().typeName(), fieldName, spreadOpField);
```
MaryamZi, line 2298: Why are we using `recordLiteral.expectedType` instead of `recordLiteral.type`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2283,7 +2341,10 @@ public void visit(BLangRecordLiteral recordLiteral) {
 
                     if (names.contains(name)) {
                         this.dlog.error(keyExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL,
-                                        recordLiteral.expectedType.getKind().typeName(), name);
+                                recordLiteral.expectedType.getKind().typeName(), name);
+                    } else if (inclusiveTypeSpreadField != null && !neverTypedKeys.contains(name)) {
+                        this.dlog.error(keyExpr.pos, DiagnosticCode.POSSIBLE_DUPLICATE_KEY_VIA_INCLUSIVE_TYPE_SPREAD_OP,
```
MaryamZi, line 2346: For me, this DiagnosticCode (and key) imply that the duplicate is via the spread field, while the actual error is that this field can duplicate a field specified via the spread field.

Shall we change these to something like `POSSIBLE_DUPLICATE_OF_FIELD_SPECIFIED_VIA_SPREAD_OP`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2295,8 +2356,11 @@ public void visit(BLangRecordLiteral recordLiteral) {
                     Object name = ((BLangLiteral) keyExpr).value;
                     if (names.contains(name)) {
                         this.dlog.error(keyExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL,
-                                        recordLiteral.parent.type.getKind().typeName(),
-                                        name);
+                                recordLiteral.parent.type.getKind().typeName(),
```
MaryamZi, line 2359: IMO, it looks cleaner when the arguments are aligned.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
dulajdilshan: Removed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
dulajdilshan: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2284,21 +2287,77 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+
+                BType spreadOpExprType = spreadOpExpr.type;
+                int spreadFieldTypeTag = spreadOpExprType.tag;
+                if (spreadFieldTypeTag == TypeTags.MAP) {
+                    if (inclusiveTypeSpreadField != null) {
+                        this.dlog.error(spreadOpExpr.pos, DiagnosticCode.MULTIPLE_INCLUSIVE_TYPES);
+                        break;
+                    } else {
+                        inclusiveTypeSpreadField = spreadOpField;
+                    }
```
MaryamZi, line 2299:
```suggestion
                    }
                    inclusiveTypeSpreadField = spreadOpField;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2284,21 +2287,77 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
                     continue;
                 }
 
-                for (BField bField : ((BRecordType) spreadOpExpr.type).fields.values()) {
-                    if (Symbols.isOptional(bField.symbol)) {
-                        continue;
+                BRecordType spreadExprRecordType = (BRecordType) spreadOpExprType;
+                boolean isSpreadExprRecordTypeSealed = spreadExprRecordType.sealed;
+                if (!isSpreadExprRecordTypeSealed) {
+                    // More than one inclusive-type-descriptors with spread-fields are not allowed.
```
MaryamZi, line 2318:
```suggestion
                    // More than one spread-field with inclusive-type-descriptors are not allowed.
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2284,21 +2287,77 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                for (Object fieldName : names) {
+                    if (fieldsInRecordType.containsKey(fieldName)) {
+                        if (fieldsInRecordType.get(fieldName).type.tag != TypeTags.NEVER) {
+                            this.dlog.error(spreadOpExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL_SPREAD_OP,
+                                            recordLiteral.type.getKind().typeName(),
+                                            fieldName,
+                                            spreadOpField);
+                        }
+                    } else {
+                        if (!isSpreadExprRecordTypeSealed) {
```
MaryamZi, line 2336: Can we merge these two as an `else if`?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1038,6 +1038,15 @@ error.duplicate.key.in.table.literal=\
 error.duplicate.key.in.record.literal.spread.op=\
   invalid usage of {0} literal: duplicate key ''{1}'' via spread operator ''{2}''
 
+error.possible.duplicate.of.field.specified.via.spread.op=\
+  invalid usage of mapping constructor expression: key ''{0}'' may duplicate a key specified via spread field ''{1}''
+
+error.already.specified.keys.in.spread.field=\
```
MaryamZi, line 1044: Shall we rename this and the DiagnosticCode to reflect that it is a possible duplicate?

e.g.,
```suggestion
error.spread.field.may.duplicate.already.specified.keys=\
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2284,21 +2287,77 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                                                DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL_SPREAD_OP,
+                                                recordLiteral.expectedType.getKind().typeName(),
+                                                name,
+                                                spreadOpField);
+                            }
+                            names.add(name);
+                        }
                     }
-                    names.add(name);
                 }
```
MaryamZi, line 2361: Shall we reorganize this to reduce indentation?

```ballerina
                for (BField bField : fieldsInRecordType.values()) {
                    String name = bField.name.value;
                    if (names.contains(name)) {
                        continue;
                    }

                    if (bField.type.tag == TypeTags.NEVER) {
                        neverTypedKeys.add(name);
                        continue;
                    }

                    if (!neverTypedKeys.remove(name) &&
                            inclusiveTypeSpreadField != null && isSpreadExprRecordTypeSealed) {
                        this.dlog.error(spreadOpExpr.pos,
                                        DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL_SPREAD_OP,
                                        recordLiteral.expectedType.getKind().typeName(),
                                        name,
                                        spreadOpField);
                    }
                    names.add(name);
                }
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2318,7 +2377,11 @@ public void visit(BLangRecordLiteral recordLiteral) {
 
                     if (names.contains(name)) {
                         this.dlog.error(keyExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL,
-                                        recordLiteral.expectedType.getKind().typeName(), name);
+                                        recordLiteral.expectedType.getKind().typeName(),
+                                        name);
```
MaryamZi, line 2381: Can't these continue to be
```suggestion
                                        recordLiteral.expectedType.getKind().typeName(), name);
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -2284,21 +2287,77 @@ public void visit(BLangRecordLiteral recordLiteral) {
...
+                }
+
+                LinkedHashMap<String, BField> fieldsInRecordType = spreadExprRecordType.fields;
+                for (Object fieldName : names) {
+                    if (fieldsInRecordType.containsKey(fieldName)) {
+                        if (fieldsInRecordType.get(fieldName).type.tag != TypeTags.NEVER) {
+                            this.dlog.error(spreadOpExpr.pos, DiagnosticCode.DUPLICATE_KEY_IN_RECORD_LITERAL_SPREAD_OP,
+                                            recordLiteral.type.getKind().typeName(),
+                                            fieldName,
+                                            spreadOpField);
```
MaryamZi, line 2333: Can we have these on a single line? Please check other places too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
dulajdilshan: Thanks

dulajdilshan:
New Pull Request: #26396

PR #25127, opened 2020-08-04
Fix failing workerWithFutureTest1

riyafa:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/workers/workers.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/workers/workers.bal
@@ -543,6 +543,7 @@ function waitFor() {
 }
 
 function add2(int i, int j) returns int {
+   runtime:sleep(1000);
    int l = 0;
    while (l < 99999) {
      l = singleAdd(l);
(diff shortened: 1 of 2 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/workers/workers.bal
```diff
@@ -543,6 +543,7 @@ function waitFor() {
 }
 
 function add2(int i, int j) returns int {
+   runtime:sleep(1000);
```
Kishanthan, line 546: why we have to add a sleep here? do we know the root cause of this test failure ?

riyafa: It is failing because all three code blocks are running parallely. At one point w2 can start and go to sleep then the future can start and run to completion in 1s and then w2 can resume and the result is returned before w1 can run. Adding a sleep will delay the future also so w1 will for sure run to completion before w2 and the add2 method.
I think the test is to test the scenario of `First cancel the future and then wait `. So this change will make sure the orders are correct. `cancel the future and wait`

Kishanthan: after this sleep you have intriduced, can you try and run the this partcular test in a loop and verify?

riyafa: I have verified it by running it a 1000 times

PR #25168, opened 2020-08-07
Remove old context resolvers and the ANTLR based resolving

nadeeshaan:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


JobaDiniz:
I came from [Build your own programming language with ANTLR](https://shalithasuranga.medium.com/build-your-own-programming-language-with-antlr-5201955537a5) which mentions `Ballerina` as a project that uses ANTLR. However it seems this project is using something else, I'm just curious if there is a discussion of why you guys made such decision.

nadeeshaan:
@JobaDiniz Instead of using ANTLR as the parser, we used our own implementation of the parser, for Ballerina. One of the main reasons behind the decision was that the ANTLR parser was not context sensitive enough for the Language Server implementation. When there are multiple errors in a given source, identifying the contexts for providing language features such as code completions were difficult.

Also, the with our implementation we got the flexibility to manage the syntax tree by introducing cleaner parser APIs as well.

JobaDiniz:
So it’s not possible to use your parser for other languages as it’s implemented only for Ballerina?

I’ve seen the 1.3 version in comparison with the 2.0 and it’s way cleaner the classes and parser analyzer

nadeeshaan:
@JobaDiniz yes, this implementation is specific to Ballerina. But, the concept can be implemented for other languages as well. We also have a toml parser implemented reusing this concept.

PR #25203, opened 2020-08-11
[Testerina] Improvements to assertion equality

### LINKED ISSUE #22048: [Testerina] AssertEquals function fails when checking array equality
**Description:**
Please refer to the following sample codes. These test cases are failing with the misleading error messages.

```ballerina
@test:Config {}
function test1() {
    string[] keys = ["A"];
    test:assertEquals(keys, ["A"]);
}

@test:Config {}
function test2() {
    int[] arr = foo();
    test:assertEquals(arr, [1]);
}

function foo() returns int[] {
    int[] arr = [1];
    return arr;
}
```

_Output_
```shell
[fail] test1:
    Assertion Failed!: expected 'A' but found 'A' 
        at ballerina.test:createBallerinaError(assert.bal:41)
        ballerina.test:assertEquals(assert.bal:82)

[fail] test2:
    Assertion Failed!: expected '1' but found '1' 
        at ballerina.test:createBallerinaError(assert.bal:41)
        ballerina.test:assertEquals(assert.bal:82)
```

**Affected Versions:**
Checked with Ballerina 1.2.0

Dilhasha:
## Purpose

Based on the current language specification, deep value equality is not supported for behavioral type values. So, using "assertEquals" to compare any|error type values is misleading. 
Thus, introducing a new function "assertExactEquals" that can be used for exact equality in "any|error" type and changing the "assertEquals" function signature to support value equality only for "anydata|error" type.

Fixes #22048 

## Approach
1. Implement "assertExactEquals", "assertExactNotEquals" and change function signature for "assertEquals", "assertNotEquals" based on expected behavior.
2. Write assertion test cases to cover different data types

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/testerina/modules/testerina-core/src/main/ballerina/src/test/assert.bal
```diff
@@ -67,14 +67,9 @@ public function assertFalse(boolean condition, public string msg = "Assertion Fa
 # + actual - Actual value
 # + expected - Expected value
 # + msg - Assertion error message
-public function assertEquals(any|error actual, any|error expected, public string msg = "Assertion Failed!") {
+public function assertEquals(anydata|error actual, anydata|error expected, public string msg = "Assertion Failed!") {
     boolean isEqual = false;
```
Ibaqu, line 71: Is the assignment required? The very next line is `isEqual = (actual == expected)`. 
Im assuming changing it to `boolean isEqual = (actual == expected)` should work. 

Same for line 87

tests/jballerina-debugger-integration-test/src/test/resources/project-based-tests/basic-project/src/hello-world/tests/hello_world_test.bal
```diff
@@ -33,5 +33,9 @@ public function mockPrint((any|error)... s) {
 function testFunc() {
     // Invoking the main function
     main();
-    test:assertEquals(outputs[0], "Hello, World!");
+    if (outputs[0] is anydata) {
```
azinneera, line 36: Should we ask the user to handle this? Is there a possibility to hide this detail from the user and handle it internally maybe within the `assertEquals` function?

Dilhasha: That is like going back to the previous implementation right?

Dilhasha: Did you mean to allow user to pass "any" type into the function and check it if it qualifies as "anydata" and do a value check?
The problem is, this check happens at runtime. If user passes a value not qualifying as "anydata", still the function would allow it and user would expect value equality. But, he would get notified about the failure only at runtime.

PR #25226, opened 2020-08-12
Handle NPE if incorrect argument is provided

### LINKED ISSUE #24937: NPE when passing incorrect named arg to the main function via the CLI
**Description:**
$title.

**Steps to reproduce:**
```ballerina
public function main() {
    
}
```

When run as `$ ballerina run zxa.bal -i=1`, with an incorrect named arg;

```log
[2020-07-25 12:34:24,781] SEVERE {b7a.log.crash} - null 
	at org.ballerinalang.jvm.util.ArgumentParser.getEntryFuncArgs(ArgumentParser.java:117)
	at org.ballerinalang.jvm.util.ArgumentParser.extractEntryFuncArgs(ArgumentParser.java:79)
	at ___init.main(.)
```

**Affected Versions:**
slp2

suganyasuven:
Fixes #24937

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/util/ArgumentParser.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/util/ArgumentParser.java
@@ -114,6 +114,9 @@ public class ArgumentParser {
             if (isNameArg) {
                 isNamedArgFound = true;
                 ParamInfo info = namedArgs.get(getParamName(arg));
+                if (info == null) {
+                    throw BallerinaErrors.createError("incorrect arguments to call the 'main' function");
+                }
                 bValueArgs[info.index * 2 + 1] = getBValue(info.type, getValueString(arg));
                 bValueArgs[info.index * 2 + 2] = true;
                 if (!info.hasDefaultable) {
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/util/ArgumentParser.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/util/ArgumentParser.java
@@ -113,9 +113,10 @@ public class ArgumentParser {
             // handle named args
             if (isNameArg) {
                 isNamedArgFound = true;
-                ParamInfo info = namedArgs.get(getParamName(arg));
+                String paramName = getParamName(arg);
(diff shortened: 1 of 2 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/util/ArgumentParser.java
```diff
@@ -114,6 +114,9 @@
             if (isNameArg) {
                 isNamedArgFound = true;
                 ParamInfo info = namedArgs.get(getParamName(arg));
+                if (info == null) {
+                    throw BallerinaErrors.createError("incorrect arguments to call the 'main' function");
```
warunalakshitha, line 118: Shall we include which arg cause the issue. @MaryamZi @praneesha Any thoughts on the error message. BTW fix looks fine for me.

MaryamZi: Yeah, we usually try to keep these errors similar to the compilation errors on function calls, right? Shall we say something like

`undefined parameter 'param_name'`

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/main/function/ArgumentParserNegativeTest.java
```diff
@@ -69,6 +69,20 @@ public void testInsufficientArgs() {
         Assert.fail("insufficient arguments not identified");
     }
 
+    @Test
```
pramodya1994, line 72: Shall we add a `description` for this test case?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/util/ArgumentParser.java
praneesha: +1 Better if we can make it "undefined parameter: 'param_name'" only if it doesn't create any inconsistencies with other similar occurrences.

PR #25242, opened 2020-08-12
Use ListNode for all delimited lists in the parser

### LINKED ISSUE #25032: Use ListNode for all delimited lists in the parser
**Description:**

$subject.
 eg: ListenersList

lochana-chathura:
## Purpose
$subject.

Fixes #25032 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
```diff
@@ -278,7 +258,7 @@ public FunctionSignatureNode transform(FunctionSignatureNode functionSignatureNo
         Token openPara = getToken(functionSignatureNode.openParenToken());
         Token closePara = getToken(functionSignatureNode.closeParenToken());
 
-        NodeList<ParameterNode> parameters = this.modifyNodeList(functionSignatureNode.parameters());
+        SeparatedNodeList<ParameterNode> parameters = this.modifySeparatedNodeList(functionSignatureNode.parameters());
```
lochana-chathura, line 261: Found a bug in TreeModifier.java. https://github.com/ballerina-platform/ballerina-lang/blob/c4f3e100fa1cbd4f003f689ecf90fe560d63689b/compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/syntax/tree/TreeModifier.java#L3100-L3101

* Here we have used `modifyGenericNodeList()` which only iterators through the alternative elements as we are parsing a `SeparatedNodeList` object, not  `NodeList`

* So when there's no modification that method simply returns the original SeparatedNodeList which is fine. 
But when there's a modification it return SeparatedNodeList which does not contain separators as no processing has been added with `getSeparator()`.

* Subsequently, when we are formatting when there's a modification, separators just get vanished. 

@SupunS shall I create an issue for the issue root?

SupunS: +1 please create an issue

SupunS: +1. Please create an issue

PR #25350, opened 2020-08-20
Check LogAPI test failure root cause

MadhukaHarith92:


MadhukaHarith92:
Closing since this was only to check the root cause of the LogApi test failure in the Github build action.

Issue #24907, opened 2020-07-23
Add support for `isolated` functions


MaryamZi:
**Description:**
Spec issue - https://github.com/ballerina-platform/ballerina-spec/issues/145
Proposal - https://github.com/ballerina-platform/ballerina-spec/blob/master/lang/proposals/isolated/isolated.md

PR #25423, opened 2020-08-26
Introduce support for isolated functions

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/24907

## Approach
A new phase is introduced for isolation function validation and inferring. In addition, the type checker now takes into account the `isolated` flag.

## Remarks
Pending tasks - support `isolated` with function type descriptors and anonymous functions. https://github.com/ballerina-platform/ballerina-lang/issues/25565

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -4694,6 +4701,9 @@ private STNode parseFunctionQualifiers(ParserRuleContext context, List<STNode> q
                 case RESOURCE_KEYWORD:
                     qualifier = parseResourceKeyword();
                     break;
+                case ISOLATED_KEYWORD:
+                    qualifier = parseIsolatedKeyword();
```
rdulmina, line 4705: Since we r seeing isolated keyword we can directly consume it. So we can use `consume()` here. Other places will also be changed to consume() in future :-)

MaryamZi: Ack, fixed just this for now with becd1b0, thanks!

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/util/diagnotic/DefaultDiagnosticListener.java
```diff
@@ -44,6 +44,7 @@ public void received(Diagnostic diagnostic) {
                 console.println("warning: " + pos + " " + diag.msg);
                 break;
             case NOTE:
+                // console.println("note: " + pos + " " + diag.msg);
```
dulvinw, line 47: Looks like committed by mistake?

MaryamZi: This is actually to help stdlib folks to identify which functions can be marked as isolated. As discussed during the meeting, we can alternatively make this a warning, the issue with that though is that the build fails unless we special case that warning too.

PR #25473, opened 2020-08-27
Fix query join spec deviations

### LINKED ISSUE #25152: Re-visit ballerina query joins for spec-deviations
**Description:**
We need to fix the following:

1. Joins should adhere to the following grammar. However, the current implementation seems to have an optional join-on-condition, where spec says it's mandatory.
```
join-clause := "join" typed-binding-pattern "in" expression join-on-condition
join-on-condition := "on" expression "equals" expression
```
2. Join internals must be based on hashtables. However, the current implementation uses a cartesian product with filtering (by re-using nested from with where clause filtering). 

**Steps to reproduce:**
n/a

**Affected Versions:**
slp2

n/a

- https://github.com/ballerina-platform/ballerina-spec/issues/435
- https://github.com/ballerina-platform/ballerina-spec/issues/581

grainier:
## Purpose
$title. Ref [1] & [2].

Fixes #25152

## Approach
n/a

## Samples
n/a

## Remarks
[1] https://github.com/ballerina-platform/ballerina-lang/issues/25152
[2]https://github.com/ballerina-platform/ballerina-spec/issues/581

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
```diff
@@ -9486,16 +9502,8 @@ private STNode parseJoinClause(boolean isRhsExpr) {
         // allow-actions flag is always false, since there will not be any actions
         // within the from-clause, due to the precedence.
         STNode expression = parseExpression(OperatorPrecedence.QUERY, isRhsExpr, false);
+        STNode onCondition = parseOnClause(isRhsExpr);
         endContext();
```
lochana-chathura, line 9506: Shall we end the join-clause context before parsing the on-clause. like we had earlier?

* Since we already have a expr parsing(L9504) within join-clause context, cannot parse another expr under the same parent context as there would be an ambiguity. 
* In the error handler get next rule for the expression-rhs when the parent ctx is  `join-clause` includes on-clause as an alternative path. 
* Consider recovery for, `from int a in b join int d in e on <missing expr> select f`
after recoverying the missing expr, error handler will then check on-clause as the next rule. 
i.e could have two on-clauses after recovery

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -2730,6 +2730,8 @@ private ParserRuleContext getNextRuleForKeywords(ParserRuleContext currentCtx, i
                     return ParserRuleContext.CONFLICT_KEYWORD;
                 } else if (parentCtx == ParserRuleContext.QUERY_EXPRESSION) {
                     return ParserRuleContext.EXPRESSION;
+                } else if (parentCtx == ParserRuleContext.JOIN_CLAUSE) {
+                    return ParserRuleContext.EXPRESSION;
```
lochana-chathura, line 2734: This is related to above comment. Since we end the join-clause ctx for on-clause we don't need this if condition. 
next rule will be taken as expresssion under `QUERY_EXPRESSION` parent context in L2731-L2732

lochana-chathura:
https://github.com/ballerina-platform/ballerina-lang/blob/166255cf5a1feedc6982ae5c4b43f3f12f5f4a11/compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParserErrorHandler.java#L2106-L2112

This is where we have made the on-clause optional.  We check if the next keyword is on-keyword then give the next rule as `ON_CLAUSE`. In order to make it not optional can remove this if-else check and give next rule as `ON_CLAUSE` directly.

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
lochana-chathura: Pleas update the doc comment above this method indicating optional on-clause as well

lochana-chathura:
Let's move 
* https://github.com/ballerina-platform/ballerina-lang/blob/166255cf5a1feedc6982ae5c4b43f3f12f5f4a11/compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/expressions/QueryExpressionTest.java#L89
* https://github.com/ballerina-platform/ballerina-lang/blob/166255cf5a1feedc6982ae5c4b43f3f12f5f4a11/compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/expressions/QueryExpressionTest.java#L96

above test cases to recovery section as they are not valid syntax anymore!

compiler/ballerina-parser/src/main/java/io/ballerinalang/compiler/internal/parser/BallerinaParser.java
grainier: According to the spec, on-clause is part of the join clause. So I assumed context should be closed ones everything is parsed. I'm confused here on what is `context` and how it works. Is there documentation which I can refer to?

grainier: Spec says follows, and according to that `equals` is only valid within a join. So, If we close the join context before finishing up with the `join-on-condition`, how are we suppose to know, for example `on expression` should followed by an `equals`?

```
join-clause := "join" typed-binding-pattern "in" expression join-on-condition
join-on-condition := "on" expression "equals" expression
```

lochana-chathura:
Had a discussion with @grainier. In this PR we will keep implicit recovery of `equals` keyword in the ballerinaParserErrorhandler and will implement explicit recovery rules for `equals` in a separate PR.

grainier:
@lochana-chathura thank you for helping out with error recovery. As discussed, I've created https://github.com/ballerina-platform/ballerina-lang/issues/25542 to keep track of implementing explicit recovery rules for equals.

mohanvive (approved):

PR #25524, opened 2020-08-28
Improve the symbol lookup method in the semantic API

pubudu91:
## Purpose
This PR introduces a number of changes to the compiler FE in order to make it easier to filter the symbols when using the semantic API. 

## Approach
- Add and persist information about the origin of the symbol. e.g., is it coming from a source or a BIR
- Add and persist position information for public constructs
- In the symbol lookup API implementation, use the above info to filter out the relevant symbols

## Remarks
Opened #25607 to keep track of the issue of excluding the symbol currently being defined when the cursor position is in a var definition statement.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/BallerinaSemanticModel.java
```diff
@@ -71,12 +80,19 @@ public BallerinaSemanticModel(BLangCompilationUnit compilationUnit, BLangPackage
         Map<Name, List<Scope.ScopeEntry>> scopeSymbols =
                 symbolResolver.getAllVisibleInScopeSymbols(lookupVisitor.lookUp(this.compilationUnit));
 
-        for (Entry<Name, List<ScopeEntry>> entry : scopeSymbols.entrySet()) {
+        DiagnosticPos cursorPos = new DiagnosticPos(new BDiagnosticSource(bLangPackage.packageID, compilationUnit.name),
```
nadeeshaan, line 83: is it possible to handle the cursor position without converting to a diagnostic position?

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/BallerinaSemanticModel.java
```diff
@@ -99,4 +115,16 @@ public BallerinaSemanticModel(BLangCompilationUnit compilationUnit, BLangPackage
     public List<Diagnostic> diagnostics(TextRange range) {
         return new ArrayList<>();
     }
+
+    private boolean isSymbolInUserProject(BSymbol symbol, DiagnosticPos cursorPos) {
+        return symbol.origin == SOURCE &&
+                (cursorPos.compareTo(symbol.pos) > SYMBOL_POSITION ||
```
rasika, line 121: Apparently `SYMBOL_POSITION` constant is redundant and the name is misleading here. Shall we just use `0` for the clarity.

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/symbols/BallerinaModule.java
```diff
@@ -71,9 +74,10 @@ protected BallerinaModule(String name, PackageID moduleID, BPackageSymbol packag
         List<FunctionSymbol> functions = new ArrayList<>();
         for (Map.Entry<Name, ScopeEntry> entry : this.packageSymbol.scope.entries.entrySet()) {
             ScopeEntry scopeEntry = entry.getValue();
-            if (scopeEntry.symbol instanceof BInvokableSymbol &&
-                    (scopeEntry.symbol.flags & Flags.PUBLIC) == Flags.PUBLIC &&
-                    scopeEntry.symbol.kind != org.ballerinalang.model.symbols.SymbolKind.ERROR_CONSTRUCTOR) {
+            if (scopeEntry.symbol != null
+                    && scopeEntry.symbol.kind == org.ballerinalang.model.symbols.SymbolKind.FUNCTION
```
rasika, line 78: Any specific reason to use FQCN `org.ballerinalang.model.symbols.SymbolKind.FUNCTION` here? or else shall we just use `SymbolKind.FUNCTION` for better readability

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/BallerinaSemanticModel.java
pubudu91: Will see. Did this conversion to make it easier to compare the user provided line position with the positions in `BSymbol`

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/BallerinaSemanticModel.java
pubudu91: Yeah, I agree the name doesn't reflect the actual position of the symbol. Thought of introducing a constant instead of using `0` since the use of `0` kinda seemed to make it a bit unclear what that comparison is about

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/symbols/BallerinaModule.java
pubudu91: There's `SymbolKind` class in the API package too. So had to use the FQ name to refer to the one in the lang module

compiler/ballerina-compiler-api/build.gradle
```diff
@@ -27,6 +27,11 @@ dependencies {
 
     testCompile 'org.testng:testng'
     testCompile project(path: ':ballerina-test-utils', configuration: 'shadow')
+    testCompile project(':ballerina-io')
```
MaryamZi, line 30: Can we test without `ballerina/io`? Just wondering if we have the requirement for this to be independent of stdlib modules.

compiler/ballerina-lang/src/main/java/org/ballerinalang/model/symbols/SymbolOrigin.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    },
+
+    /**
+     * These are symbols for which there is a corresponding construct in the source file written by the user.
+     */
+    SOURCE {
+        @Override
+        public SymbolOrigin toBIROrigin() {
+            return COMPILED_SOURCE;
```
MaryamZi, line 49: Is this correct or should this be `SOURCE`?

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/BallerinaModuleID.java
```diff
@@ -66,4 +66,24 @@ public String toString() {
...
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (obj == null || getClass() != obj.getClass()) {
+            return false;
+        }
+
+        // TODO: Improve this
```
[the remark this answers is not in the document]
pubudu91: Yes, planning on fixing these in this PR itself.

compiler/ballerina-compiler-api/build.gradle
pubudu91: This is just for the tests. That should be fine right?

compiler/ballerina-lang/src/main/java/org/ballerinalang/model/symbols/SymbolOrigin.java
pubudu91: This is the intended value. The idea was to set a symbol's origin to `COMPILED_SOURCE` if it's coming from the BIR. So any `SOURCE` symbols become `COMPILED_SOURCE` symbols once the compilation is finished.

MaryamZi: Would a comment explaining this help?

compiler/ballerina-compiler-api/build.gradle
MaryamZi: Even that we usually try to avoid, right? Wouldn't it be an issue with the stdlib migration anyway?

pubudu91: Yeah, had a chat with Supun as well. Will replace this with a balo test.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1493,7 +1497,7 @@ private boolean isRestDetailBindingAvailable(BLangErrorVariable errorVariable) {
 
     private BTypeSymbol createTypeSymbol(int type) {
         return new BTypeSymbol(type, Flags.PUBLIC, Names.EMPTY, env.enclPkg.packageID,
-                               null, env.scope.owner, symTable.builtinPos); // TODO: Check with Dhananjaya
+                               null, env.scope.owner, symTable.builtinPos, VIRTUAL); // TODO: Check with Dhananjaya
```
SupunS, line 1500: Shall we convert these TODOs to issues, and remove the comments?
Having people references in a public code-base is a bit weird.

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/BallerinaSemanticModel.java
pubudu91: This basically compares whether the cursor position is before or after the position of a given symbol

compiler/ballerina-lang/src/main/java/org/ballerinalang/model/symbols/SymbolOrigin.java
pubudu91: Ack, will add a comment

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/BallerinaSemanticModel.java
SupunS: Oh okay. So this is an object compare.
I also feel `0` is more natural in that case.

SupunS (approved):

PR #25534, opened 2020-08-31
Add annotations test case to BIR spec

Kishanthan:
## Purpose
This PR adds annotations test case to BIR spec. It also has few refactoring done on ksy file to avoid issues with constants. 

Related to #25535

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
@@ -620,7 +620,7 @@ public class BIRPackageSymbolEnter {
             case TypeTags.DECIMAL:
                 return new BLangConstantValue(getStringCPEntryValue(dataInStream), symTable.decimalType);
             case TypeTags.BOOLEAN:
-                return new BLangConstantValue(dataInStream.readByte() == 1, symTable.booleanType);
+                return new BLangConstantValue(dataInStream.readBoolean(), symTable.booleanType);
             case TypeTags.NIL:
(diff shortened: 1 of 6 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -620,7 +620,7 @@ private BLangConstantValue readConstLiteralValue(DataInputStream dataInStream) t
             case TypeTags.DECIMAL:
                 return new BLangConstantValue(getStringCPEntryValue(dataInStream), symTable.decimalType);
             case TypeTags.BOOLEAN:
-                return new BLangConstantValue(dataInStream.readByte() == 1, symTable.booleanType);
```
manuranga, line 623: Is it ok to change this? We'll have to inform llvm team to change there code.

manuranga (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
Kishanthan: We are writing boolean and reading on CP table. So doing the same in other places also will be consistent which is why the above change. Also, it makes sense to write and read boolean values for boolean constants than writing byte (even though we are doing the same thing writing 1 or 0 byte effectively)

PR #25551, opened 2020-09-01
Add 'const' support for specifying array length

### LINKED ISSUE #13162: [Spec Deviation] Const is not supported when specifying array length
**Description:**
When creating a fixed length array, the size should be able to be specified using a compile-time constant.
```ballerina
const len = 5;
function testFunction() {
    int[len] intArr = [1, 2, 3, 4, 5]; // Currently not supported
}
```

**Affected Versions:**
0.990.2

suleka96:
**Scenario 1**
```
   const len = 5;

public function main() {
    int[len] intArr = [1, 2, 3, 4, 5, 6];
}
```
Expected: size mismatch error
Given: no error

**Scenario 2**
```
   int len = 5;

public function main() {
    int[len] intArr = [1, 2, 3, 4, 5, 6];
}
```
Expected: error on invalid size reference type
Given: no error

**Scenario 3**
```
   const len = "5";

public function main() {
    int[len] intArr = [1, 2, 3, 4, 5,6];
}

```
Expected: error on const array size reference data type
Given: no error

**Scenario 4**
```
public function main() {
    int[len] intArr = [1, 2, 3, 4, 5,6];
}
```
Expected: error on undefined size reference
Given: no error

These spec deviations are fixed in this pull request.

Fixes #13162 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -905,17 +907,61 @@ public void visit(BLangArrayType arrayTypeNode) {
...
+                                arrType = new BArrayType(resultType, arrayTypeSymbol,
+                                        (int) length,
+                                        BArrayState.CLOSED_SEALED);
+                                arrayTypeSymbol.type = arrType;
+                                markParameterizedType(arrType, arrType.eType);
+                                resultType = arrayTypeSymbol.type;
+
+                            } else {
+                                dlog.error(arrayTypeNode.pos,
+                                        DiagnosticCode.INVALID_ARRAY_SIZE_REFERENCE_TYPE,
```
MaryamZi, line 951: I think we can reuse existing error messages for these errors. For example here we can say

`incompatible types: expected 'int', found 'T'` where T is the actual type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -905,17 +907,61 @@ public void visit(BLangArrayType arrayTypeNode) {
...
+                                        ((BConstantSymbol) sizeSymbol).literalType);
+                                resultType = symTable.semanticError;
+                            }
+                        } else {
+                            dlog.error(arrayTypeNode.pos, DiagnosticCode.INVALID_ARRAY_SIZE_REFERENCE, sizeSymbol);
+                            resultType = symTable.semanticError;
+                        }
+                    } else {
+                        dlog.error(arrayTypeNode.pos, DiagnosticCode.ARRAY_SIZE_REFERENCE_NOT_DEFINED, size);
```
MaryamZi, line 960: Here, IMO, we can use the undefined symbol error.

`undefined symbol 'x'`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangArrayType.java
```diff
@@ -62,10 +65,14 @@ public String toString() {
         final StringBuilder[] sb = {new StringBuilder(getTypeName())};
         if (sizes.length == 0) {
             Arrays.stream(sizes).forEach(size -> {
-                if (size == -1) {
-                    sb[0].append("[]");
+                if (size instanceof BLangLiteral) {
```
MaryamZi, line 68: Let's use node kind.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangArrayType.java
```diff
@@ -62,10 +65,14 @@ public String toString() {
...
+                    if ((Integer) (((BLangLiteral) size).getValue()) == -1) {
```
MaryamZi, line 69: Let's extract `(Integer) (((BLangLiteral) size).getValue()` out to a variable.

MaryamZi (commented):
@suleka96, when adding the tests, let's also add a few filler-value-related compile time and runtime tests.

MaryamZi:
Let's also add tests for such arrays defined at module level, within functions, within blocks, within anonymous functions, etc.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -905,17 +907,61 @@ public void visit(BLangArrayType arrayTypeNode) {
...
+                    resultType = arrayTypeSymbol.type;
 
-            markParameterizedType(arrType, arrType.eType);
+                } else {
+                    Name pkgAlias = names.fromIdNode(((BLangSimpleVarRef) size).pkgAlias);
+                    Name typeName = names.fromIdNode(((BLangSimpleVarRef) size).variableName);
+
+                    BSymbol sizeSymbol = lookupMainSpaceSymbolInPackage(size.pos, env, pkgAlias, typeName);
+
+                    if (!(symTable.notFoundSymbol.equals(sizeSymbol))) {
```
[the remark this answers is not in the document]
suleka96: The` visit(BLangArrayType arrayTypeNode)` function does not return anything, so every `if` will get evaluated if I do it this way. So won't having the `else` block be a better approach?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangArrayType.java
suleka96: Is there a specific reason for this? Because if I don't specifically check for the `BLangLiteral` type in the `if` statement, my current logic won't stand as the else block assumes that it is receiving `BLangLiteral` Type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -905,17 +907,61 @@ public void visit(BLangArrayType arrayTypeNode) {
...
-                int size = arrayTypeNode.sizes[i];
-                arrType = (size == UNSEALED_ARRAY_INDICATOR) ?
-                        new BArrayType(resultType, arrayTypeSymbol, size, BArrayState.UNSEALED) :
-                        (size == OPEN_SEALED_ARRAY_INDICATOR) ?
-                                new BArrayType(resultType, arrayTypeSymbol, size, BArrayState.OPEN_SEALED) :
-                                new BArrayType(resultType, arrayTypeSymbol, size, BArrayState.CLOSED_SEALED);
-            }
-            resultType = arrayTypeSymbol.type = arrType;
+                BLangExpression size = arrayTypeNode.sizes[i];
+                if (size instanceof BLangLiteral) {
```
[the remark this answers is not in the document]
suleka96: Is there a specific reason for this? Because if I don't specifically check for the BLangLiteral type in the if statement, my current logic won't stand as the else block assumes that it is receiving BLangLiteral Type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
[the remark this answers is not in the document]
MaryamZi: You can just do `return;`.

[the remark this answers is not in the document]
MaryamZi: Or in this case, because it is within a for loop, and we may want to analyze the rest of the dimensions even if one results in an error, we can do a `continue;` maybe. I guess we'll have to maintain if at least one dimension was erroneous to set the final resultant type.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1044,6 +1044,9 @@ error.mismatching.array.literal.values=\
 error.index.out.of.range=\
   index out of range: index: ''{0}''
 
+error.invalid.array.size.reference=\
+  invalid array size reference ''{0}''
```
MaryamZi, line 1048: Let's make this more detailed.

For example, something like
```suggestion
  invalid reference expression ''{0}'' as array size: expected a constant reference expression
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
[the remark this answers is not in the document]
suleka96: So just to clarify, if we consider an array with 2 dimension. Say the 1st one didn't go through the happy path, so we log the error, update an error flag and continue. Then say if the 2nd dimension goes through the happy path, should the `resultType`  be set to an error type regardless or should the below code be executed for that loop? 
```
arrayTypeSymbol.type = arrType;
resultType = arrayTypeSymbol.type;
markParameterizedType(arrType, arrType.eType);
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangParserListener.java
```diff
@@ -828,18 +834,18 @@ public void exitArrayTypeNameLabel(BallerinaParser.ArrayTypeNameLabelContext ctx
...
         int index = 1;
         int dimensions = 0;
-        List<Integer> sizes = new ArrayList<>();
+        List<BLangExpression> sizes = new ArrayList<>();
         List<ParseTree> children = ctx.children;
         while (index < children.size()) {
             if (children.get(index).getText().equals("[")) {
                 if (children.get(index + 1).getText().equals("]")) {
-                    sizes.add(UNSEALED_ARRAY_INDICATOR);
+                    sizes.add(new BLangLiteral(Integer.valueOf(OPEN_ARRAY_INDICATOR), symTable.intType));
```
MaryamZi, line 842: Creating these expressions and setting them is usually delegated off to the package builder. Also doesn't feel quite right for the parser listener to know of the symbol table. 

Since the antlr parser will be removed soon, shall we just pass an empty list in L858 and disable the tests on the old parser, instead of introducing these changes?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -118,9 +119,8 @@
 
 import static java.lang.String.format;
 import static org.wso2.ballerinalang.compiler.semantics.model.Scope.NOT_FOUND_ENTRY;
-import static org.wso2.ballerinalang.compiler.util.Constants.OPEN_SEALED_ARRAY_INDICATOR;
-import static org.wso2.ballerinalang.compiler.util.Constants.UNSEALED_ARRAY_INDICATOR;
-
+import static org.wso2.ballerinalang.compiler.util.Constants.INFERRED_ARRAY_INDICATOR;
+import static org.wso2.ballerinalang.compiler.util.Constants.OPEN_ARRAY_INDICATOR;
```
MaryamZi, line 123: Leave a new line after the imports. Please check and fix everywhere.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -899,24 +899,79 @@ public void visit(BLangArrayType arrayTypeNode) {
...
+                        arrType = new BArrayType(resultType,
+                                arrayTypeSymbol,
+                                sizeIndicator,
+                                BArrayState.OPEN_SEALED);
+                    } else {
+                        arrType =  new BArrayType(resultType,
+                                arrayTypeSymbol,
+                                sizeIndicator,
+                                BArrayState.CLOSED_SEALED);
+                    }
```
MaryamZi, line 929: Since it is only the array state that changes, IMO we can improve readability here by doing something like

```suggestion
                    BArrayState arrayState;
                    if (sizeIndicator == OPEN_ARRAY_INDICATOR) {
                        arrayState = BArrayState.UNSEALED;
                    } else if (sizeIndicator == INFERRED_ARRAY_INDICATOR) {
                        arrayState = BArrayState.OPEN_SEALED;
                    } else {
                        arrayState = BArrayState.CLOSED_SEALED;
                    }
                    arrType =  new BArrayType(resultType, arrayTypeSymbol,  sizeIndicator, arrayState);
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -899,24 +899,79 @@ public void visit(BLangArrayType arrayTypeNode) {
...
+
+                    BConstantSymbol sizeConstSymbol = (BConstantSymbol) sizeSymbol;
+                    BType lengthLiteralType = sizeConstSymbol.literalType;
+
+                    if (lengthLiteralType.tag != TypeTags.INT) {
+                        dlog.error(size.pos,
+                                DiagnosticCode.INCOMPATIBLE_TYPES,
+                                symTable.intType,
+                                sizeConstSymbol.literalType);
```
MaryamZi, line 956: Shall we reformat this to have whatever possible in a single line?

MaryamZi: Please check other places too.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArraySizeDefinitionTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private String invalidReferenceExpressionError = "invalid reference expression " +
+            "'intLength' as array size: expected a constant reference expression";
+    private String incompatibleTypeError = "incompatible types: expected 'int', found 'string'";
+    private String undefinedSymbolError = "undefined symbol 'length'";
+
+    @Test(groups = { "disableOnOldParser" })
+    public void testCompilationSizeReferenceErrors() {
+        int index = 0;
+        CompileResult result = BCompileUtil.compile(
+                "test-src/statements/arrays/array-size-test.bal");
```
MaryamZi, line 42: Can we not have this on one line?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArraySizeDefinitionTest.java
```diff
@@ -0,0 +1,95 @@
...
+        Assert.assertEquals(result.getDiagnostics().length, 20);
```
MaryamZi, line 43: Let's move this to the end and assert equality between `index` and `result.getDiagnostics().length`, that way we also make sure we've asserted all the errors.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArraySizeDefinitionTest.java
MaryamZi: Please check other places too.

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-fill-test.bal
```diff
@@ -13,7 +13,6 @@
 // KIND, either express or implied.  See the License for the
 // specific language governing permissions and limitations
 // under the License.
-
```
MaryamZi, line 16: Let's not remove this.

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-size-scope-test.bal
```diff
@@ -0,0 +1,39 @@
+const constLength = 2;
```
MaryamZi, line 1: Please add the license text to all the bal files too.

MaryamZi, line 1: Let's use underscores instead of hyphens in bal file names.

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-size-scope-test.bal
```diff
@@ -0,0 +1,39 @@
...
+int intLength = 2;
+const stringLength = "2";
+
+//in module scope
```
MaryamZi, line 5: We don't quite need these comments right?

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-size-scope-test.bal
MaryamZi: Let's also suffix the negative test bal files with `_negative`

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array-size-test.bal
```diff
@@ -0,0 +1,29 @@
...
+    int[stringLength][2] m = [[1,2],[1,2]];
+    int[length][2] n = [[1,2],[1,2]];
+    int[2][2] o = [[1,2],[1,2],[1,2]];
+
+    int[2][constLength][2] p = [[[1,2],[1,2],[1,2]],[[1,2],[1,2]]];
+    int[2][intLength][2] q = [[[1,2],[1,2]],[[1,2],[1,2]]];
+    int[2][stringLength][2] r = [[[1,2],[1,2]],[[1,2],[1,2]]];
+    int[2][length][2] s = [[[1,2],[1,2]],[[1,2],[1,2]]];
+    int[2][2][2] t = [[[1,2],[1,2],[1,2]],[[1,2],[1,2]]];
+}
```
MaryamZi, line 29: Missing a new line.

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array_fill_runtime_test.bal
```diff
@@ -120,6 +121,28 @@ function createMultipleConstLiteralAutoFilledSealedArray() {
     assertEqualPanic(1, sealedArray[3]);
 }
 
+function tesOneDimensionalArrayWithConstantSizeReferenceFill() {
+    int[constLength] a = [];
+    a[1]=2;
```
MaryamZi, line 126:
```suggestion
    a[1] = 2;
```

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/sealed_array.bal
```diff
@@ -15,6 +15,8 @@
 // under the License.
 import ballerina/lang.'int;
 
+const constLength = 2;
```
MaryamZi, line 18: Can we also add positive tests for module level, anon functions, etc. array definitions with constants?

MaryamZi: Need to also include positive and negative tests for scenarios where the constant is defined after the array in the source.

For example, the following should work.

```ballerina
int[TWO] x = [1, 2];

const TWO = 2;
```

and the following should result in an error.

```ballerina
int[ONE] x = [1, 2];

const ONE = 1;
```

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/sealed_array.bal
```diff
@@ -438,6 +440,13 @@ function createConstLiteralAutoFilledSealedArray() {
     assertArrayLengthPanic(5, sealedArray);
 }
 
+function testArrayWithConstantSizeReferenceFill() {
+    int[constLength] sealedArray = [1];
+    sealedArray[1]=2;
```
MaryamZi, line 445:
```suggestion
    sealedArray[1] = 2;
```

tests/jballerina-unit-test/src/test/resources/testng.xml
```diff
@@ -27,7 +27,7 @@
         <!--<listener class-name="org.ballerinalang.test.utils.TestNGListener"/>-->
     </listeners>
     <suite-files>
-        <suite-file path="./testng-jballerina.xml" />
+<!--        <suite-file path="./testng-jballerina.xml" />-->
```
dulajdilshan, line 30: Do we really need this to be commented out?

suleka96: There were some changes that I did in the BLangParserListener in order for the array tests in the old parser to not throw any errors. However, as suggested I reverted those changes. Since the old parser tests are bound to fail without the changes, I commented them out.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/arrays/ArraySizeDefinitionTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+import org.ballerinalang.test.util.BAssertUtil;
+import org.ballerinalang.test.util.BCompileUtil;
+import org.ballerinalang.test.util.CompileResult;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+/**
+ * Array size definition test.
+ *
```
MaryamZi, line 28: Shall we add `@since Swan Lake`?

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array_size_scope_test_negative.bal
```diff
@@ -0,0 +1,54 @@
...
+
+    if (x == 10) {
+        int[constLength] m = [1, 2, 3];
+        int[intLength] n = [1, 2];
+        int[stringLength] o = [1, 2];
+        int[length] p = [1, 2];
+    }
+}
+
+
```
MaryamZi, line 54: Extra new lines.

MaryamZi: Please check all files.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangArrayType.java
```diff
@@ -62,10 +65,15 @@ public String toString() {
         final StringBuilder[] sb = {new StringBuilder(getTypeName())};
         if (sizes.length == 0) {
             Arrays.stream(sizes).forEach(size -> {
-                if (size == -1) {
-                    sb[0].append("[]");
+                if (size.getKind() == NodeKind.LITERAL || size.getKind() == NodeKind.NUMERIC_LITERAL) {
```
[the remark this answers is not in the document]
suleka96: We actually discussed about this before in [this](https://github.com/ballerina-platform/ballerina-lang/pull/25551#discussion_r482686991) and it was suggested to use both. Since it is the same logic that is being used here I used both.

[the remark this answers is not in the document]
MaryamZi: IIRC digits are usually represented by `NUMERIC_LITERAL`s, while strings are represented by `LITERAL`. 

So once we've validated that the size is a numeric literal (probably done by the parser itself), we don't have to check for `LITERAL` thereafter. Because we know that if it is an array's size, and it is not inferred or a const ref, it has to be a `NUMERIC_LITERAL`.

hasithaa (approved):

PR #25604, opened 2020-09-03
Add `array:sort` function

### LINKED ISSUE #25421: Add `array:sort` function
Ref: https://github.com/ballerina-platform/ballerina-spec/issues/572

lasinicl:
Fixes #25421

## Samples
```
type Student record {|
   int id;
   string name;
|};

function testSort() {
    Student s1 = {id: 3, name: "Ben"};
    Student s2 = {id: 2, name: "Matthew"};
    Student s3 = {id: 1, name: "Amber"};

    Student[] arr = [s1, s2, s3];

    Student[] sortedArr = arr.sort("descending", function(Student s) returns string {
        return s.name;
    });
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.array/src/main/ballerina/src/lang.array/array.bal
```diff
@@ -152,14 +152,33 @@ public function lastIndexOf(PureType[] arr, PureType val, int startIndex = arr.l
...
+
+# Any ordered type is a subtype of this.
+type OrderedType BasicType|BasicType[];
+
+// TO DO: Add this when cyclic type reference in union type definitions is supported
+//# Any ordered type is a subtype of this.
+//public type OrderedType ()|boolean|int|float|decimal|string|OrderedType[];
+
+# Sorts an array.
+# If the member type of the array is not sorted, then the `key` function
```
pubudu91, line 172: I think this is meant to be "ordered", not "sorted".

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3474,6 +3474,15 @@ public BType resolveExprType(BType type) {
      * @return boolean whether the type is basic type or not.
      */
     public boolean isBasicType(BType type) {
+        if (type.tag == TypeTags.ARRAY) {
+            BType arrType = ((BArrayType) type).eType;
+            return checkBasicType(arrType);
+        } else {
+            return checkBasicType(type);
```
pubudu91, line 3481: No need to special case the array right? Can use the same switch case to do this.

pubudu91: Also, what's the exact purpose of this method? The name of the method conveys a completely different meaning from its implementation. e.g., this seems to consider union types with `()` as basic. But in the spec, unions aren't considered as a basic type.

Even without the union, this seems to be considering an arbitrary set of types from the basic types.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5390,6 +5395,28 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
...
+                }
+            } else {
+                BLangLambdaFunction func = (BLangLambdaFunction) iExpr.argExprs.get(2);
+                BType returnType = func.function.type.getReturnType();
+                if (!types.isBasicType(returnType)) {
+                    dlog.error(iExpr.expr.pos, DiagnosticCode.INVALID_SORT_FUNC_RETURN_TYPE, returnType);
+                }
+            }
+        }
+    }
```
pubudu91, line 5418: There shouldn't be any need to hard code the type checking for this. Why do we need to do this?

lasinicl: In the spec issue, James has mentioned that
"The compiler needs to check a few things that cannot be expressed in the signature:

- if a key function is specified as an argument, then its return value must be an ordered type

- if there is no key function, then the member type of the array must be ordered type"

https://github.com/ballerina-platform/ballerina-spec/issues/572#issuecomment-663838997

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
lasinicl: Will look into it and do the necessary changes

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,185 @@
...
+        Object[][] sortArrClone = new Object[arr.size()][2];
+        for (int i = 0; i < arr.size(); i++) {
+            if (function != null) {
+                Object x = function.call(new Object[]{strand, arr.get(i), true});
+                sortArr[i][0] = x;
+            } else {
+                sortArr[i][0] = arr.get(i);
+            }
+            sortArr[i][1] = arr.get(i);
+        }
```
MaryamZi, line 74: Since `function` doesn't change for a single invocation, wouldn't it be better to do 

```suggestion
        if (function != null) {
            for (int i = 0; i < arr.size(); i++) {
                sortArr[i][0] = function.call(new Object[]{strand, arr.get(i), true});
                sortArr[i][1] = arr.get(i);
            }
        } else {
            for (int i = 0; i < arr.size(); i++) {
                sortArr[i][0] = sortArr[i][1] = arr.get(i);
            }
        }
```

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,185 @@
...
+
+        if (function != null) {
+            elemType = ((BFunctionType) function.getType()).retType;
```
MaryamZi, line 77: If go with the suggestion to swap the null check above, we can move this there too.

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,185 @@
...
+        }
+
         return arr;
     }
 
-    // Adapted from https://algs4.cs.princeton.edu/22mergesort/Merge.java.html
-    private static void mergesort(ArrayValue input, ArrayValue aux, int lo, int hi, Strand strand,
-                                  FPValue<Object, Long> comparator) {
+    private static BType getType(BType type) {
+        Optional<BType> mainType = ((BUnionType) type).getMemberTypes().stream()
```
MaryamZi, line 99: We avoid using streams due to perf impact.

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,185 @@
...
+                .filter(m -> m.getTag() == TypeTags.INT_TAG || m.getTag() == TypeTags.BYTE_TAG ||
+                        m.getTag() == TypeTags.FLOAT_TAG || m.getTag() == TypeTags.DECIMAL_TAG ||
+                        m.getTag() == TypeTags.STRING_TAG || m.getTag() == TypeTags.BOOLEAN_TAG ||
+                        m.getTag() == TypeTags.ARRAY_TAG).findFirst();
```
MaryamZi, line 103: What if the union has multiple compatible types? Should it not be an error?

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,185 @@
...
+            arr.add(k, sortArr[k][1]);
+        }
+
         return arr;
     }
 
-    // Adapted from https://algs4.cs.princeton.edu/22mergesort/Merge.java.html
-    private static void mergesort(ArrayValue input, ArrayValue aux, int lo, int hi, Strand strand,
-                                  FPValue<Object, Long> comparator) {
+    private static BType getType(BType type) {
```
MaryamZi, line 98: Can we rename the function to indicate what we're actually looking for and change the param to a `BUnionType`?

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,185 @@
...
+        // () should come last irrespective of the sort direction.
+        if (value1 == null) {
+            if (value2 == null) {
+                return 0;
+            }
+            if (isAscending) {
+                return 1;
+            }
+            return -1;
+        } else if (value2 == null) {
```
MaryamZi, line 157: Since you're definitely returning in the previous if bock we can make this another if instead of an else-if right?

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,185 @@
...
+            if (isAscending) {
+                return -1;
+            }
+            return 1;
+        } else if (type.getTag() == TypeTags.INT_TAG) {
```
MaryamZi, line 162: What about subtypes like `SIGNED32_INT_TAG`, `SIGNED16_INT_TAG`, etc.? Same for string and char too.

The util methods in TypeTags.java may be useful. org.ballerinalang.jvm.types.TypeTags#isIntegerTypeTag

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,185 @@
...
+        } else if (type.getTag() == TypeTags.ARRAY_TAG) {
+            for (int i = 0; i < ((ArrayValue) value1).size(); i++) {
+                c = sortFunc(((ArrayValue) value1).get(i), ((ArrayValue) value2).get(i),
+                        ((ArrayValue) value1).getElementType(), isAscending);
+                if (c != 0) {
+                    break;
+                }
+            }
+        } else {
+            throw BallerinaErrors.createError("invalid sort key type", "Failed to sort array");
```
MaryamZi, line 198: This doesn't seem to follow the normal practice we follow for langlib errors. We currently have a simple string as the error message and the second arg goes into the detail. Please check other usage of this method.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
MaryamZi: This also seems to miss the int and string subtypes. I also think `isBasicType` (I think `isSimpleBasicType` is more accurate here) should just check for the simple basic types in line with the spec, and we can maybe introduce a separate method that handles arrays and unions, eventually calling `isBasicType`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi, line 5418: I think we can refactor this method a bit by returning early on error and merging the if blocks.
```suggestion
    private void checkArrayLibSortFuncArgs(BLangInvocation iExpr) {
        if (iExpr.argExprs.size() <= 2 && !types.isBasicType(iExpr.expr.type)) {
            dlog.error(iExpr.expr.pos, DiagnosticCode.INVALID_SORT_ARRAY_MEMBER_TYPE, iExpr.expr.type);
        } 
        
        if (iExpr.argExprs.size() != 3) {
            return;
        }

        BLangExpression argThree = iExpr.argExprs.get(2);
        BType argThreeType = argThree.type;
        if (argThreeType != null) {
            if (argThreeType.tag == TypeTags.SEMANTIC_ERROR) {
                return;
            }
            
            if (argThreeType.tag == TypeTags.NIL && !types.isBasicType(iExpr.expr.type)) {
                dlog.error(iExpr.expr.pos, DiagnosticCode.INVALID_SORT_ARRAY_MEMBER_TYPE, iExpr.expr.type);
                return;
            }
        }
        
        BLangLambdaFunction func = (BLangLambdaFunction) argThree;
        BType returnType = func.function.type.getReturnType();
        if (!types.isBasicType(returnType)) {
            dlog.error(iExpr.expr.pos, DiagnosticCode.INVALID_SORT_FUNC_RETURN_TYPE, returnType);
        }
    }
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5228,6 +5228,11 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
             requiredParams.add(nonRestParam);
         }
 
+        // check argument types in arr:sort function
+        if (iExpr.langLibInvocation && iExpr.name.value.equals("sort")) {
```
MaryamZi, line 5232: In org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker#checkInLangLib we do some langlib specific validations. Wondering if we should move this also there.

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
lasinicl: Handled this at compile time during type checking

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
lasinicl: Refactored this

langlib/lang.array/src/main/ballerina/src/lang.array/array.bal
```diff
@@ -152,14 +152,33 @@ public function lastIndexOf(PureType[] arr, PureType val, int startIndex = arr.l
...
+   DESCENDING = "descending"
+}
+
+# A type parameter that is a subtype of `()|boolean|int|float|decimal|string`.
+type BasicType ()|boolean|int|float|decimal|string;
+
+# Any ordered type is a subtype of this.
+type OrderedType BasicType|BasicType[];
+
+// TO DO: Add this when cyclic type reference in union type definitions is supported
```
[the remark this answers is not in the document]
lasinicl: Created issue https://github.com/ballerina-platform/ballerina-lang/issues/25627 to track this

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticCode.java
```diff
@@ -302,6 +302,8 @@
     INVALID_LIST_INDEX_EXPR("invalid.list.index.expr"),
     INVALID_ARRAY_INDEX_EXPR("invalid.array.index.expr"),
     SEALED_ARRAY_TYPE_CAN_NOT_INFER_SIZE("sealed.array.type.can.not.infer.size"),
+    INVALID_SORT_FUNC_RETURN_TYPE("invalid.sort.func.return.type"),
```
mohanvive, line 305: Isn't this invalid key function return type?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5384,6 +5389,37 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
         }
     }
 
+    private void checkArrayLibSortFuncArgs(BLangInvocation iExpr) {
+        if (iExpr.argExprs.size() <= 2 && !types.isOrderedType(iExpr.expr.type)) {
+            dlog.error(iExpr.expr.pos, DiagnosticCode.INVALID_SORT_ARRAY_MEMBER_TYPE, iExpr.expr.type);
+        }
+
+        if (iExpr.argExprs.size() != 3) {
+            return;
```
mohanvive, line 5398: Don't we need any error message for this case and it needs to be the first condition to verify in the method.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5384,6 +5389,37 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
...
+        }
+
+        BLangExpression argThree = iExpr.argExprs.get(2);
```
mohanvive, line 5401: Please use proper names for arguments. (argThreee does not provide any context)

mohanvive (changes requested):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3467,30 +3467,44 @@ public BType resolveExprType(BType type) {
...
+            case TypeTags.ARRAY:
+                BType elementType = ((BArrayType) type).eType;
+                if (elementType.tag == TypeTags.ARRAY) {
+                    elementType = ((BArrayType) elementType).eType;
                 }
+                if (elementType.tag == TypeTags.UNION) {
+                    Set<BType> memTypes = ((BUnionType) elementType).getMemberTypes();
+                    // can not sort (string?|int)[]/(string|int)[], can sort string?[]
+                    return memTypes.size() <= 2 && memTypes.contains(symTable.nilType);
+                }
```
pubudu91, line 3504: Shouldn't it simply be a recursive call to `isOrderedType()` on `elementType`? For example, what happens if the array has more than two dimensions?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3467,30 +3467,44 @@ public BType resolveExprType(BType type) {
...
+    public boolean isOrderedType(BType type) {
+        switch (type.tag) {
             case TypeTags.UNION:
-                if (((BUnionType) type).getMemberTypes().contains(symTable.nilType)) {
-                    return true;
-                } else {
-                    return false;
+                Set<BType> memberTypes = ((BUnionType) type).getMemberTypes();
+                // can not sort (string?|int)/(string|int), can sort string?
+                return memberTypes.size() <= 2 && memberTypes.contains(symTable.nilType);
```
pubudu91, line 3494: What happens here if the union has a non-ordered type as a member type?

pubudu91: Should be calling `isOrderedType()` recursively on the member types, in addition to the above checks right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2761,6 +2761,11 @@ private void checkInLangLib(BLangInvocation iExpr, BType varRefType) {
         }
 
         checkIllegalStorageSizeChangeMethodCall(iExpr, varRefType);
+
+        // check argument types in arr:sort function
+        if (iExpr.langLibInvocation && iExpr.name.value.equals("sort")) {
```
pubudu91, line 2766: At this point, don't need the `iExpr.langLibInvocation` check here since at this point, we know for sure (Line# 2753) that this is a lang lib call.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
lasinicl: `iExpr.argExprs` list can have 1 to a maximum of 3  members considering the scenarios given below
arr.sort(), arr.sort(direction), arr.sort(direction, key). Therefore this can not be taken as the first condition to be verified. 
When there are more than 3 arguments it'll be handled by the following https://github.com/ballerina-platform/ballerina-lang/blob/8481a2bd849a79154b46e06d56d3557762ce64c5/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java#L5299

mohanvive (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2761,6 +2761,11 @@ private void checkInLangLib(BLangInvocation iExpr, BType varRefType) {
         }
 
         checkIllegalStorageSizeChangeMethodCall(iExpr, varRefType);
+
+        // check argument types in arr:sort function
+        if (iExpr.name.value.equals("sort")) {
```
grainier, line 2766: Put the constant value on the left side of equals to avoid possible NPEs

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,230 @@
...
             } else {
-                input.add(k, aux.get(i++));
+                index = i++;
             }
+            input[k] = aux[index];
         }
     }
 
+    private static int sortFunc(Object value1, Object value2, BType type, boolean isAscending) {
+        int c = -999;
```
grainier, line 147: What is this -999? And Why it's -999?

grainier (dismissed):

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1050,6 +1050,12 @@ error.array.index.out.of.range=\
 error.sealed.array.type.can.not.infer.size=\
   invalid usage of sealed type: can not infer array size
 
+error.invalid.key.func.return.type=\
+  invalid array sort key function return type: type ''{0}'' is not an ordered type
+
+error.invalid.sort.array.member.type=\
+  invalid member type of the array to sort: type ''{0}'' is not an ordered type
```
pubudu91, line 1057: Just wondering whether it'd be better to avoid saying "array", considering that this applies to tuples as well? @MaryamZi 

Also, the first  "type" in the message after the colon seems redundant.

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -18,73 +18,229 @@
...
+                    break;
+                } else {
+                    if (i == len - 1 && lengthVal1 < lengthVal2) {
+                        return -1;
+                    }
+                }
+            }
+            return c;
+        }
+        throw BallerinaErrors.createError("Invalid type to sort", "Expected an ordered type, but found type: "
```
pubudu91, line 222: This error message needs to conform to the same format we follow for other lang lib errors. You can check the other places in the lang lib where panics are created for reference. Also, IIRC, we don't start the detail with a capital letter. @MaryamZi

stdlib/http/src/main/ballerina/src/http/http_request.bal
```diff
@@ -534,7 +535,14 @@ public type Request object {
...
         string cookieheader = "";
-        Cookie[] sortedCookies = cookiesToAdd.sort(comparator);
+        Cookie[] sortedCookies = cookiesToAdd.sort(array:ASCENDING, function(Cookie c) returns int {
+            var cookiePath = c.path;
+            int l = 0;
+            if (cookiePath is string) {
+                l = cookiePath.length();
+            }
+            return l;
+        });
```
pubudu91, line 545: This changes the original ordering behaviour implemented by the provided comparator. I think we'll have to sort it twice to achieve the same effect: first, sort using the length as above and second, sort using the created time. @chamil321 thoughts?

stdlib/http/src/main/ballerina/src/http/websocket/websocket_client.bal
```diff
@@ -237,7 +238,14 @@ public function addCookies(WebSocketClientConfiguration|WebSocketFailoverClientC
...
     if (cookiesToAdd is Cookie[]) {
-        Cookie[] sortedCookies = cookiesToAdd.sort(comparator);
+        Cookie[] sortedCookies = cookiesToAdd.sort(array:ASCENDING, function(Cookie c) returns int {
+            var cookiePath = c.path;
+            int l = 0;
+            if (cookiePath is string) {
+                l = cookiePath.length();
+            }
+            return l;
+        });
```
pubudu91, line 248: Same comment as above

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibArrayTest.java
```diff
@@ -419,66 +382,153 @@ public void callingLengthModificationFunctionsOnFixedLengthLists() {
...
+
+    @Test
+    public void testSort8() {
+        BRunUtil.invoke(compileResult, "testSort8");
+    }
+
+    @Test
+    public void testSort9() {
+        BRunUtil.invoke(compileResult, "testSort9");
+    }
```
pubudu91, line 533: Shall we add a couple of tests for the `array:sort()` style of usage as well? Both positive and negative.

pubudu91: And what about arrays having finite types?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5384,6 +5390,37 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
...
+            dlog.error(iExpr.expr.pos, DiagnosticCode.INVALID_SORT_ARRAY_MEMBER_TYPE, iExpr.expr.type);
+        }
+
+        if (iExpr.argExprs.size() != 3) {
+            return;
+        }
+
+        BLangExpression keyFunction = iExpr.argExprs.get(2);
+        BType keyFunctionType = keyFunction.type;
+        if (keyFunctionType != null) {
```
pubudu91, line 5404: For what scenarios would `keyFunctionType` be `null` here?

stdlib/http/src/main/ballerina/src/http/http_request.bal
lasinicl: Yes. aware of it. This was communicated with @daneshk and I was asked to create an issue in the HTTP module repo and fix it here just to get the PR checks passed.

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibArrayTest.java
lasinicl: > Shall we add a couple of tests for the `array:sort()` style of usage as well? Both positive and negative.

Have already added for both.
Some positive cases have been added in
[#L701](https://github.com/ballerina-platform/ballerina-lang/blob/62d470fa9d5069bfa89d83bc91d049141beb1712/langlib/langlib-test/src/test/resources/test-src/arraylib_test.bal#L701)
[#L793](https://github.com/ballerina-platform/ballerina-lang/blob/62d470fa9d5069bfa89d83bc91d049141beb1712/langlib/langlib-test/src/test/resources/test-src/arraylib_test.bal#L793)
[#L853](https://github.com/ballerina-platform/ballerina-lang/blob/62d470fa9d5069bfa89d83bc91d049141beb1712/langlib/langlib-test/src/test/resources/test-src/arraylib_test.bal#L853)
Some negative cases have been added in
[#[L136]](https://github.com/ballerina-platform/ballerina-lang/blob/62d470fa9d5069bfa89d83bc91d049141beb1712/langlib/langlib-test/src/test/resources/test-src/arraylib_test_negative.bal#L136)
[#[L148]](https://github.com/ballerina-platform/ballerina-lang/blob/62d470fa9d5069bfa89d83bc91d049141beb1712/langlib/langlib-test/src/test/resources/test-src/arraylib_test_negative.bal#L148)

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
lasinicl: seems it's an unnecessary check. will remove it :)

compiler/ballerina-lang/src/main/resources/compiler.properties
lasinicl: @pubudu91 @MaryamZi any suggestions on what it should be?

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibArrayTest.java
pubudu91: No, I meant using it as a normal function. For example,
```ballerina
int[] sorted = array:sort(arr);
```

stdlib/http/src/main/ballerina/src/http/http_request.bal
pubudu91: Ah ok.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1051,10 +1051,10 @@ error.sealed.array.type.can.not.infer.size=\
   invalid usage of sealed type: can not infer array size
 
 error.invalid.key.func.return.type=\
-  invalid array sort key function return type: type ''{0}'' is not an ordered type
+  invalid sort key function return type: ''{0}'' is not an ordered type
 
 error.invalid.sort.array.member.type=\
-  invalid member type of the array to sort: type ''{0}'' is not an ordered type
+  invalid member type of the array/tuple to sort: type ''{0}'' is not an ordered type
```
pubudu91, line 1057:
```suggestion
  invalid member type of the array/tuple to sort: ''{0}'' is not an ordered type
```

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -219,8 +222,8 @@ private static int sortFunc(Object value1, Object value2, BType type, boolean is
             }
             return c;
         }
-        throw BallerinaErrors.createError("Invalid type to sort", "Expected an ordered type, but found type: "
-                + type.toString());
+        throw BallerinaErrors.createError(getModulePrefixedReason(ARRAY_LANG_LIB, INVALID_TYPE_TO_SORT),
+                "expected an ordered type, but found type: '" + type.toString() + "'");
```
pubudu91, line 226: Shall we just simply say `but found '" + type.toString() + "'"`?

langlib/lang.array/src/main/ballerina/src/lang.array/array.bal
pubudu91: Can you check and confirm this with James? Should be corrected on the spec as well. I think this was meant to say "the array is not ordered". Because that's when you have a need for the key function.

pubudu91 (approved):

langlib/lang.array/src/main/ballerina/src/lang.array/array.bal
lasinicl: Will do :)

PR #25622, opened 2020-09-04
Split Testerina assertions test case into multiple tests

### LINKED ISSUE #25560: Testerina integration test 'testAssertions' needs to be fixed and enabled
**Description:**
The testAssertions test in the Testerina BasicTestCases has been disabled. Although it passes locally, it seems that the 87 tests do not finish executing before the LogLeecher checks the logs when a PR is sent

I have tried increasing the time the logLeecher has to wait for the test logs, but this has no impact and the test continues to fail. 

**Steps to reproduce:**

**Affected Versions:**

Ibaqu:
## Purpose
Split testerina assertions test case into smaller tests that can be run without the server timing out

Fixes #25560

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


Ibaqu:
I have enabled some of the test cases by removing the unnecessary io:println from the test cases. There are some that cannot be enabled since they use the `hhtp` and `config` modules.

tests/testerina-integration-test/src/test/resources/project-based-tests/outside-tests/src/pathVerification/tests/main_test.bal
```diff
@@ -14,20 +14,17 @@
...
 // under the License.
 
-import ballerina/io;
 import ballerina/test;
 
 # Before Suite Function
 
 @test:BeforeSuite
 function beforeSuiteFunc() {
-    io:println("I'm the before suite function!");
```
Dilhasha, line 24: @Ibaqu We need to add logic(appending test and asserting it) to test that the functions get executed in correct order. WDYT?

Ibaqu: There is a similar logic included in the `beforeAfter` and `beforeEachAfterEach` test cases. It is also there in the `dependsOn` test case as well. In those tests, a string is passed between the functions and the end result is asserted to see if the order is executed correctly. https://github.com/ballerina-platform/ballerina-lang/pull/25622/files#diff-43b6d1a6469d4251593b5734978c5e38

@Dilhasha Is this what you are referring to? If so, should I include such logic in the 'pathVerification' test case as well?

Dilhasha: Yes that's correct. 
Let's include similar logic in test cases wherever applicable rather than using println.

Ibaqu: Added changes to pathVerification test

tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/BasicCasesTest.java
```diff
@@ -31,13 +31,13 @@
     private BMainInstance balClient;
     private String projectPath;
 
-    @BeforeClass(enabled = false)
+    @BeforeClass(enabled = true)
```
azinneera, line 34: we can remove the `(enabled = true)`, All functions are enabled by default. We don't need to specify it explicitly.

Ibaqu: Removed the `enabled = true` from the test cases

Dilhasha, line 34: Shall we remove "enabled = true" from here too?

Dilhasha (approved):
LGTM

PR #25635, opened 2020-09-07
Verify and Fix XML related Formatter test case scenarios

### LINKED ISSUE #25347: [Formatter] Verify and fix the XML related formatting scenarios
**Description:**
The XML scenarios in the new formatter implementation have to be verified with the Ballerina specification and fixed there are any formatting issues. The XML related tests in the formatter test suite should then be enabled.

**Steps to reproduce:**

**Affected Versions:**

sanjana:
Fixes #25347

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
```diff
@@ -1186,12 +1186,16 @@ public XMLNamespaceDeclarationNode transform(XMLNamespaceDeclarationNode xMLName
         Token asKeyword = getToken(xMLNamespaceDeclarationNode.asKeyword().orElse(null));
         IdentifierToken namespacePrefix = this.modifyNode(xMLNamespaceDeclarationNode.namespacePrefix().orElse(null));
         Token semicolonToken = getToken(xMLNamespaceDeclarationNode.semicolonToken());
+        if (asKeyword != null || namespacePrefix != null) {
```
IrushiL, line 1189: Shouldn't this be an AND expression here? Since both the `namespacePrefix` and the `asKeyword` should not be null.

misc/formatter/modules/formatter-core/src/test/resources/testng.xml
```diff
@@ -29,6 +29,7 @@ under the License.
         <packages>
             <package name="org.ballerinalang.formatter.core.types.*" />
             <package name="org.ballerinalang.formatter.core.expressions.*" />
+            <package name="org.ballerinalang.formatter.core.statements.*" />
```
IrushiL, line 32: We might have to leave this out until fixes for the rest of the statement tests are merged, if there are failing statement tests.

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
sanjana: The purpose of using the `OR` is, if either one of `asKeyword` or `namespacePrefix` is available (not null), then the _if_ condition would be true and both of them should be present when applying the modifications. 
Otherwise if even a single one out of them is null, there would be an assertion error generated when checking in the `Objects.requireNonNull()` methods of _tree node modifier utility_. 

And, even if the `asKeyword` is `null`, still the parser would be returning a `MISSING[as]` token if the `namespacePrefix` is present. 

Is this approach correct, or, are there any other suggestions?

misc/formatter/modules/formatter-core/src/test/resources/testng.xml
sanjana: Fixed by limiting to `org.ballerinalang.formatter.core.statements.XMLNSDeclarationStatementsTest` only.

SupunS (commented):
Are we only adding formatting support for XML namespaces? Any plans for other XML related stuff like XML literal etc?

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
sanjana: @IrushiL, Thank you for pointing out the issue here. Fixed the logical error.

misc/formatter/modules/formatter-core/src/test/resources/testng.xml
sanjana: Removed the `statement` package and added the `XMLNSDeclarationStatementsTest` class separately.

IrushiL:
> Are we only adding formatting support for XML namespaces? Any plans for other XML related stuff like XML literal etc?

@SupunS XML literals are already covered under the following test cases.

https://github.com/ballerina-platform/ballerina-lang/tree/master/misc/formatter/modules/formatter-core/src/test/resources/types/sequence/assert

IrushiL (dismissed):

IrushiL (approved):

Issue #25037, opened 2020-07-30
Implement the class type


rdhananjaya:
**Description:**

Subtasks: 
[ ] - Implement the parser changes for class defn #25038
[ ] - Remove and migrate object\abstract objects to classes #25039
[ ] - Support object constructor expression #25043
[ ] - Improve parser error handling for class-definition #25071
[ ] - Support parsing class annotation declarations #25187
[ ] - Add dataflow and taint analysis test cases for class-defn #25306
[ ] - Align parser with new parser rules for object type #25312
[ ] - Support defining distinct classes #25361
[ ] - Support distinct object type in the runtime #25379
[ ] - Migrate rawtemplate desugaring to class defn #25517
[ ] - Add documentation generation support for Class definition #25536
[ ] - Add language server support for module-class-defn and object constructor expressions #25643
[ ] - Add Ballerina debugger support for module-class-defn and object constructor expressions #25644
[ ] - Add testerina support for module-class-defn and object constructor expressions #25645

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #25637, opened 2020-09-07
Introduce module level class definitions and object constructor expression

rdhananjaya:
## Purpose
This PR is to align the Ballerina compiler with class definitions and object constructor expressions introduced to the spec targetting Ballerina Swan Lake release.

With this PR object type definitions become just a type definition, which means we can't create objects with just the object type, we need to have a `class` or an `object constructor expression` to provide the implementation for an object type.

### Object type def
Prev:
```ballerina
type Obj object {
    int i;
    function init(int i) {
        self.i = i;
    }
};
```
Now:
```ballerina
class Obj {
    int i;
    function init(int i) {
        self.i = i;
    }
}
```

### Abstract objects
Prev:
```ballerina
type Obj abstract object {
    function foo(int i);
};
```
Now:
```ballerina
type Obj object {
    function foo(int i);
};
```

### Anon objects
Prev:
```ballerina
function bar() {
    object { function foo() { // do something } } o = new();
}
```
Now:
```ballerina
function bar() {
    var o = object { function foo() { // do something }};
}
```

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/25037

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/ballerinalang/model/tree/SortableTypeDefNode.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.ballerinalang.model.tree;
+
+/**
+ * Interface representing type definitions that are sortable using their dependencies to each other.
+ *
+ * @since 2.0
+ */
+public interface SortableTypeDefNode extends Node {
```
pubudu91, line 26: How about `OrderedNode` or `OrderedTypeDefNode`? Suggested the former since this precedence APIs seem generic enough that they can be used for other nodes as well, if needed

nadeeshaan (changes requested):
For the following source snippet, code analyzer fails with a NPE, when the `TOOLING_COMPILATION` option is on.
```
import ballerina/module1;

function testToStringMethodForTable() {
    var tesVar = table key<module1:>
}
```

rdhananjaya:
> For the following source snippet, code analyzer fails with a NPE, when the `TOOLING_COMPILATION` option is on.
> 
> ```
> import ballerina/module1;
> 
> function testToStringMethodForTable() {
>     var tesVar = table key<module1:>
> }
> ```

@nadeeshaan fixed the issue, please check

compiler/ballerina-lang/src/main/java/org/ballerinalang/model/tree/SortableTypeDefNode.java
rdhananjaya: +1

hasithaa (approved):

nadeeshaan (approved):
Reviewed the language server test cases and the latest signature help test cases

hasithaa (approved):

PR #25664, opened 2020-09-08
Add testcases for BIR variable optimization

riyafa:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/bir/BirVariableOptimizationTest.java
```diff
@@ -0,0 +1,89 @@
...
+        ((BLangPackage) result.getAST()).symbol.bir.typeDefs.forEach(
+                typeDefinition -> typeDefinition.attachedFuncs.forEach(this::assertFunctions));
+    }
+
+    private void assertFunctions(BIRNode.BIRFunction func) {
+        String expectedBir = null;
+        try {
+            expectedBir = readFile(func.getName().getValue());
+        } catch (IOException e) {
+            Assert.fail();
```
Kishanthan, line 66: will it print the reason as well?

riyafa: added reason

Kishanthan: don't we have to use information coming from 'e' ?

riyafa: didn't think it will add value. But passed the error the function as well now.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/bir/BirVariableOptimizationTest.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private void assertFunctions(BIRNode.BIRFunction func) {
+        String expectedBir = null;
+        try {
+            expectedBir = readFile(func.getName().getValue());
+        } catch (IOException e) {
+            Assert.fail("Failed when reading file", e);
+        }
+        if (!"".equals(expectedBir)) {
```
Kishanthan, line 68: So if the expectedBir is an empty string, then the test case would pass here?

riyafa: yes. The file is not in the folder. We test for files which are in the folder. For `...init` or `...start` or any such functions we chose to ignore there won't be a file for that. Then we ignore. If the file is empty also it will ignore I guess.

Kishanthan: ack

Kishanthan (approved):

Issue #12892, opened 2018-12-20
Assertion errors given to the user only contain the string representation of values


pubudu91:
When an assertion error occurs, it only uses the string representations of the values to communicate the error to the user. However, if the string representation is the same, but the types are different, the assertion still fails, but the user will be confused as to why the assertion failed when the two values are seemingly the same. Consider the following sample error which was a result of a type mismatch.
```
error: Assertion Failed!: expected '{name:"John Doe", age:25, address:{city:"Colombo", country:"Sri Lanka"}}' but found '{name:"John Doe", age:25, address:{city:"Colombo", country:"Sri Lanka"}}' {}
	    	at ballerina/test:createBallerinaError(assert.bal:42)
	    	ballerina/test:assertEquals(assert.bal:76)
	    	ballerina-examples/anonymous-records:0.0.1:testFunc(tests/anonymous_records_test.bal:19)
```

Dilhasha:
Following is how the assertion comparison is done in **JUnit5**.

If there is a type mismatch, the errors are shown as follows.

- String "1" vs int 1
expected: java.lang.String@28feb3fa<1> but was: java.lang.Integer@675d3402<1>

- Json vs String
expected: java.lang.String@5025a98f<{"name":"Dilhasha"}> but was: com.google.gson.JsonObject@49993335<{"name":"Dilhasha"}>

When the types are matched, but the values are different, the errors are shown as follows.

- expected: <8> but was: <7>

In the following 2 examples, only the type of the property 'age' is different.
 
- expected: <{"firstName":"Sergey","age":18}> but was: <{"firstName":"Sergey","age":"18"}>

- expected: <{"firstName":"Sergey","lastName":"Harvard","age":18,"address":"879, Colombo Road","school":"primary","telephone":"+96738373933","occupation":"SSE","favouriteColor":"Green"}> but was: <{"firstName":"Sergey","lastName":"Harvard","age":"18","address":"879, Colombo Road","school":"primary","telephone":"+96738373933","occupation":"SSE","favouriteColor":"Green"}>

Dilhasha:
Following is how the assertion comparison is done in **GoLang**.

If there is a type mismatch, the errors are shown as follows.

- string "1" vs int 1

        	Error:      	Not equal:
        	            	expected: int(1)
        	            	actual  : string("1")

- Json vs String

        	Error:      	Not equal: 
[...]

PR #25690, opened 2020-09-09
Add type information to assertion error message

Dilhasha:
Fixes part 1 of #12892

## Samples

```
> test:assertEquals(1,"1");

        [fail] testFunction:
            Assertion Failed!
                expected: <string> '1'
                actual  : <int> '1'
```

```
>  Person p1 = new;
    Employee p2 = new;
    test:assertExactEquals(p1,p2);

        [fail] testFunction:
            Assertion Failed!
                expected: <Employee> 'object diffmod:Employee'
                actual  : <Person> 'object diffmod:Person'
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/io/BallerinaTypeCheck.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.ballerinalang.testerina.natives.io;
```
Ibaqu, line 19: Does BallerinaTypeCheck fit into natives.io? Cant it be considered part of core?

hevayo (changes requested):
Can we add tests and check the behavior for all types. specially anonymous types and other built in types.

misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/io/BallerinaTypeCheck.java
Dilhasha: +1 It would be better to move this to core.

Dilhasha: +1 It would be better to move this to core.

> Can we add tests and check the behavior for all types. specially anonymous types and other built in types.

Dilhasha:
> Can we add tests and check the behavior for all types. specially anonymous types and other built in types.

Sure I will do that.

Ibaqu (dismissed):

hevayo (dismissed):

misc/testerina/modules/testerina-core/src/main/ballerina/src/test/assert.bal
```diff
@@ -19,6 +19,7 @@ import ballerina/java;
 const string assertFailureErrorCategory = "assert-failure";
 const string arraysNotEqualMessage = "Arrays are not equal";
 const string arrayLengthsMismatchMessage = " (Array lengths are not the same)";
+const int maxArgLength = 2048;
```
hevayo, line 22: This is too much to be effective shall we use 80 which is usually the default character width of a terminal. Or we can do 180 as per https://en.wikipedia.org/wiki/Characters_per_line

Dilhasha: Yes, true. Let's use 80 then.

hevayo (approved):

PR #25697, opened 2020-09-10
Add missing similarity criteria between array and tuple types 

### LINKED ISSUE #23676: Similarities between array and tuple types not fully honored
**Description:**
The arrays and tuple types are similar to each other with the difference that in the array all the members are of the same type while in the tuple they can be of different types.
The following types could be used interchangeably:
`int[2`] is the same as `[int, int]` - list of two int members

and hence the `is` type check should return true when used with either of them. But currently they return false;

**Steps to reproduce:**
```
import ballerina/io;

public function main() {
    [int...] x = [1, 2];
    any y = x;
    io:println(y is [int...]); // should return true
    io:println(y is int[]); // should return true (currently returns false)
    io:println(y is int[0]);  // should return false (currently returns false)
    io:println(y is [int]);  // should return false (currently returns true)

    int[] a=[1,2]; 
    y = a;
     io:println(y is [int...]); // should return true (currently returns false)
    io:println(y is int[]); // should return true (currently returns true)

     int[2] b=[1,2]; 
    y = b;
    io:println(y is [int, int]);  // should return true (currently returns false)


}
```
Consider other possible combinations as well
**Affected Versions:**
2.0.0-SNAPSHOT and previous versions

based on comment here: https://github.com/ballerina-platform/ballerina-lang/pull/23640#discussion_r433780341

suleka96:
Fixes #

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


KRVPerera:
Think these assertions should fail.

```Ballerina
public function testClosedArrayType2(){
  float[1] b=[1];
  any y = b;
  assertEquality(y is [float...], true);

  float[*] c =[1];
  any x = c;
  assertEquality(x is [float...], true); 
}
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1040,18 +1043,71 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+            for (int i = 0; i < sourceArrayType.getSize(); i++) {
+                fieldTypes.add(sourceArrayType.getElementType());
+            }
+            sourceTupleType = new BTupleType(fieldTypes);
+        }
+
+        List<BType> sourceTypes = new ArrayList<>(sourceTupleType.getTupleTypes());
+        BType sourceRestType = sourceTupleType.getRestType();
+
+        if ((targetTypes.size() != 0 && (sourceTypes.size() == 0 || sourceTypes.size() != 0))
```
KRVPerera, line 1102: Please check the logic here.

⛏️  Code may look simpler if we can use `sourceTypes.isEmpty()` method instead of comparisons like `0` and `!= 0`

♻️ But it looks like after adding rest types to both `sourceTypes` and `targetTypes` and checking the size in (L:1115) make the logic here redundant. 

🔧 Also I think `(sourceTypes.size() == 0 || sourceTypes.size() != 0)` is always true.

MaryamZi:
> Think these assertions should fail.
> 
> ```ballerina
> public function testClosedArrayType2(){
>   float[1] b=[1];
>   any y = b;
>   assertEquality(y is [float...], true);
> 
>   float[*] c =[1];
>   any x = c;
>   assertEquality(x is [float...], true); 
> }
> ```

These should pass, right? In both scenarios `x` and `y` are `float` lists of length 1. Since the `is` check is for a list with 0 or more `float` members, a `float` list of length 1 will always belong to it.

The "belongs-to" relationship is defined in the spec as

KRVPerera:
Yes, I was wrong. This is explained in detail for the context as in section [5.4.1.2 Tuple types](https://ballerina.io/ballerina-spec/spec.html#section_5.4.1.2)

> A tuple type descriptor T with m member type descriptors contains a list shape L of length n if and only if:
> - m is less than or equal to n
> - the i-th member type descriptor of T contains the i-th member of L for each i from 1 to m;
> - if n is greater than m, then T has a tuple-rest-descriptor R..., and R contains the j-th member of L for each j from m + 1 to n.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1005,10 +1005,20 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
             if (tupleTypes.isEmpty()) {
                 return targetType.getState() == ArrayState.UNSEALED || targetType.getSize() == 0;
             }
+
+            if (sourceTupleType.getRestType() != null && targetType.getState() == ArrayState.UNSEALED &&
+                    targetType.getElementType().getTag() != TypeTags.UNION_TAG) {
```
MaryamZi, line 1010: Why are we special casing union type?

Also instead of checking equality of type tag can't we call `checkIsType`?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1040,18 +1045,83 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+            BArrayType sourceArrayType = (BArrayType) sourceType;
+            BType sourceElementType = sourceArrayType.getElementType();
+
+            switch (sourceArrayType.getState()) {
+                case UNSEALED:
+                    if (targetRestType == null) {
+                        return false;
+                    } else {
+                        return checkIsType (sourceElementType, targetRestType, unresolvedTypes);
+                    }
```
MaryamZi, line 1078:
```suggestion
                    }
                    return checkIsType (sourceElementType, targetRestType, unresolvedTypes);
```

MaryamZi: Need to check non-rest member types of the target tuple too.

For example where `x` is `boolean[] x = [false, true];`, `<any> x is [string, boolean...]` evaluates to true now, but it shouldn't.

Basically an open array cannot belong to a tuple type unless it has a rest descriptor AND has no non-rest member types.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1040,18 +1045,83 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+                case CLOSED_SEALED:
```
MaryamZi, line 1079: Shall we also change these names like we did for compile time in the other PR?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1040,18 +1045,83 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+                    if (sourceArrayType.getSize() >= targetTypes.size()) {
```
MaryamZi, line 1080: Can't we also check `sourceArrayType.getSize() < targetTypes.size()` and return false here? We don't need to do the rest of the checks starting from L1107 if so, right?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1040,18 +1045,83 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+                        if (targetTypes.isEmpty()) {
+                            if (targetRestType != null) {
+                                return checkIsType (sourceElementType, targetRestType, unresolvedTypes);
+                            }
+                            return true;
```
MaryamZi, line 1085: Shouldn't we check the size here? Reaching here would mean the target is an empty list? This would mean the source also has to be empty?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1040,18 +1045,83 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+                            if (!(checkIsType(sourceElementType, targetElementType, unresolvedTypes))) {
+                                isSameType = false;
+                            }
+                        }
+                        if (sourceArrayType.getSize() > targetTypes.size()) {
+                            if (targetRestType != null) {
+                                return (checkIsType (sourceElementType, targetRestType, unresolvedTypes) && isSameType);
+                            } else {
+                                return false;
+                            }
```
MaryamZi, line 1100:
```suggestion
                                return checkIsType(sourceElementType, targetRestType, unresolvedTypes) && isSameType;
                            }
                            return false;
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1040,18 +1045,83 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+                case CLOSED_SEALED:
+                    if (sourceArrayType.getSize() >= targetTypes.size()) {
+                        if (targetTypes.isEmpty()) {
+                            if (targetRestType != null) {
+                                return checkIsType (sourceElementType, targetRestType, unresolvedTypes);
+                            }
+                            return true;
+                        }
+
+                        boolean isSameType = true;
```
MaryamZi, line 1088: Shall we use a name like `memberTypesMatch`? `isSameType` doesn't quite explain what the check is IMO.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1040,18 +1045,83 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+
+                        for (BType targetElementType : targetTypes) {
+                            if (!(checkIsType(sourceElementType, targetElementType, unresolvedTypes))) {
+                                isSameType = false;
+                            }
+                        }
+                        if (sourceArrayType.getSize() > targetTypes.size()) {
```
MaryamZi, line 1095: We can reduce the indentation a bit by doing
```suggestion
                        if (sourceArrayType.getSize() == targetTypes.size()) {
                            return isSameType;
                        }
                        if (targetRestType != null) {
                        ...
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1040,18 +1045,83 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+        List<BType> sourceTypes = new ArrayList<>(sourceTupleType.getTupleTypes());
+        BType sourceRestType = sourceTupleType.getRestType();
+
+        if ((!(targetTypes.isEmpty()) && (sourceRestType != null && targetRestType == null))) {
             return false;
         }
 
-        List<BType> sourceTypes = new ArrayList<>(((BTupleType) sourceType).getTupleTypes());
-        BType sourceRestType = ((BTupleType) sourceType).getRestType();
         if (sourceRestType != null) {
```
MaryamZi, line 1121: The logic from L1121 to L1131 doesn't seem right to me.

Where `x` is `[int, boolean, boolean...] x = [1, true]`,
`<any> x is [int, boolean...]` should evaluate to true.

But it doesn't because we seem to be doing a size check with the rest type also added as a separate type.

MaryamZi (changes requested):

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
suleka96: We did not change the array state names we only changes the indicator names.
```
 if (sizeIndicator == OPEN_ARRAY_INDICATOR) {
        arrayState = BArrayState.UNSEALED;
} else if (sizeIndicator == INFERRED_ARRAY_INDICATOR) {
        arrayState = BArrayState.OPEN_SEALED;
} else {
        arrayState = BArrayState.CLOSED_SEALED;
}
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
suleka96: Actually I was wrong here. Can a tuple  exist with its member type size being empty and also the rest type being null?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
suleka96: Actually I was wrong here. Can a tuple  exist with its member type size being empty and also the rest type being null?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
MaryamZi: We can change the state too, right? `OPEN`, `CLOSED`, and `INFERRED` maybe?

suleka96:
Please find the updated pull request #26173

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
MaryamZi: Although we don't seem to allow writing such a tuple atm, I believe it is valid according to the spec. IMO, we can add the validation now since it is a simple check. That way it will be future-proof too.

PR #25726, opened 2020-09-10
Refactor symbol position serialization logic

### LINKED ISSUE #25725: Refactored position writing logic breaks the BIR
The logic for persisting position info in the BIR was extended through #25524. In the process, the existing logic was also refactored a bit. That refactoring seems to have broken the BIR, resulting in build failures when std libraries are imported.

pubudu91:
## Purpose
Fixes #25725


pubudu91:
Closing this since the issue turned out to be a dependency issue, not an issue with the BIR change.

PR #25739, opened 2020-09-11
Fix identifying module level arrow expression with closure vars

### LINKED ISSUE #25288: Gets a "Bad Sad" when assigning a function value that returns another function to a module-level variable
**Steps to reproduce:**
```ballerina
type F function (int) returns function (int) returns int;
F f = a => b => a + b;

public function main() {
}
```

**Affected Versions:**
Swan Lake Preview 2

Kishanthan:
## Purpose
This PR fixes the issue where module level arrow expressions with closure vars are failing to compile.

Fixes #25288

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -4773,7 +4773,7 @@ public class TypeChecker extends BLangNodeVisitor {
 
     private void markAndRegisterClosureVariable(BSymbol symbol, DiagnosticPos pos) {
         BLangInvokableNode encInvokable = env.enclInvokable;
-        if (symbol.owner instanceof BPackageSymbol) {
+        if (symbol.owner instanceof BPackageSymbol && env.node.getKind() != NodeKind.ARROW_EXPR) {
             return;
         }
         if (encInvokable != null && encInvokable.flagSet.contains(Flag.LAMBDA)
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/lambda/ArrowExprTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/lambda/ArrowExprTest.java
@@ -265,6 +265,11 @@ public class ArrowExprTest {
         BRunUtil.invoke(basic, "testNestedArrowExpressionWithOneParameter");
     }
 
+    @Test(description = "Test global arrow expression with closures")
+    public void testGlobalArrowExpressionsWithClosure() {
+        BRunUtil.invoke(basic, "testGlobalArrowExpressionsWithClosure");
+    }
+
     @Test(description = "Test type narrowing in arrow expression")
     public void testTypeNarrowingInArrowExpression() {
         BRunUtil.invoke(basic, "testTypeNarrowingInArrowExpression");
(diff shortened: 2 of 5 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -4773,7 +4773,7 @@ private void checkFunctionInvocationExpr(BLangInvocation iExpr) {
 
     private void markAndRegisterClosureVariable(BSymbol symbol, DiagnosticPos pos) {
         BLangInvokableNode encInvokable = env.enclInvokable;
-        if (symbol.owner instanceof BPackageSymbol) {
+        if (symbol.owner instanceof BPackageSymbol && env.node.getKind() != NodeKind.ARROW_EXPR) {
```
pubudu91, line 4776: I'm guessing the issue here is that in `F1 f1 = a => b => a + b;`, `a`'s owner is the module?

pubudu91: If that's the case, this should ideally be fixed by properly setting the owner of `a` to the symbol for the arrow function.

pubudu91: If this is a blocker, I guess we can go with this for now and fix it properly later.

Kishanthan: I think the first check in this method is, we skip if the current symbol is a top level node. Is that correct for arrow functions which has closures?

MaryamZi: @pubudu91, isn't this related to the issue we discussed recently?

IIRC, even in the following example, the owner of `a` is not the arrow function - it is the function `foo`. Which again is not quite correct IMO.

```ballerina
function foo() {
    function (int) returns int x = a => a + 1;    
}
```

pubudu91: @Kishanthan Yes, we skip the symbol if it's a module level construct (e.g., a module level variable) because those are handled in a different manner and don't need to deal with closures when referring to those symbols. 

In a normal function the owner of the param symbols is the function symbol. But in the arrow functions, looks like the relationship of the params aren't accurately reflected in the symbols. 

@MaryamZi Yeah, it seems to be the same issue

Kishanthan: So having a lambda module level function does not have this issue. I guess the reason here is that arrowFunction is **not** treated like lambda function at typechecking phase, or altest at closure analysis phase, whereas lambdas are specially/properly handled.

Then there is a bigger issue here IMO.

rdhananjaya:
Merging as this fixes the bug, an issue was created to track the improvements suggested in the comments 
https://github.com/ballerina-platform/ballerina-lang/issues/25849

rdhananjaya (approved):

PR #25802, opened 2020-09-15
Improve the API for retrieving the symbol at cursor position

pubudu91:
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/NodeResolver.java
```diff
@@ -0,0 +1,1363 @@
...
+            // TODO: Check whether we can return from here.
+        }
+    }
+
+    private void lookupNode(BLangNode node) {
+        if (node == null) {
+            return;
+        }
+
+        if (PositionUtil.withinBlock(this.cursorPos, node.pos)) {
```
nadeeshaan, line 221: is it possible to merge the two if checks to one and do an early return?

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/NodeResolver.java
```diff
@@ -0,0 +1,1363 @@
...
+    }
+
+    @Override
+    public void visit(BLangPackage pkgNode) {
+        super.visit(pkgNode);
+    }
+
+    @Override
+    public void visit(BLangTestablePackage testablePkgNode) {
+        super.visit(testablePkgNode);
```
nadeeshaan, line 233: Do we need to explicitly call the super method here?

nadeeshaan (dismissed):

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/NodeResolver.java
pubudu91: Ah no. This is a piece of IDE generated code. Left it as-is so that we'll easily know if this is something that we need to support.

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/NodeResolver.java
pubudu91: Can merge. But it'd still be the same I guess

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/NodeResolver.java
pubudu91: I guess we can remove these type of overridden methods

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/impl/BallerinaSemanticModel.java
```diff
@@ -102,15 +105,16 @@ public BallerinaSemanticModel(BLangPackage bLangPackage, CompilerContext context
...
-            Scope.ScopeEntry value = entry.getValue();
-
-            if (isWithinSymbol(position, value.symbol.pos)) {
-                return Optional.of(SymbolFactory.getBCompiledSymbol(value.symbol, name.value));
-            }
+        BLangCompilationUnit compilationUnit = getCompilationUnit(srcFile);
+        NodeResolver nodeResolver = new NodeResolver();
+        BLangNode node = nodeResolver.lookup(compilationUnit, position);
+
+        if (node instanceof IdentifiableNode) {
```
SupunS, line 112: Can we check the node-kind and get the symbol?
Checking `instanceof` can be expensive, and we won't have to introduce another interface also.

pubudu91: If checking the node kind, we'll have to handle each node with a symbol separately. What I have done here is, introduce an API for getting/setting a symbol to a node (currently we don't have a unified API for accessing a symbol from a node). So any node which has a symbol associated to it implements this interface and at this point, once I find the node enclosing the specified position, I just simply call `getSymbol()` to get the symbol. No need to consider the nodes individually.

SupunS: +1 go with this for the initial version. But IMO we should remove `instanceof` checks in all places.

For e.g.: since the NodeResolver is already a visitor, if we can return the symbol directly rather than doing it in two-steps, then we can avoid these checks.

pubudu91: The performance of instanceof check isn't an issue in Java 8 right? 

Made it a generic node resolving visitor in case we need similar behaviours for other APIs. +1 for making it return the symbol itself according how this progresses

SupunS (approved):

PR #25815, opened 2020-09-15
Fix type checking issue in sort key function

lasinicl:
Fixes #25781

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #25834, opened 2020-09-16
Fix issue in arrow expression type checking in sort

### LINKED ISSUE #25848: NPE when trying to sort an any[]
**Description:**
$title.

**Steps to reproduce:**
```ballerina
    any[] x = [1, 3, 2];
    any[] sortResult = x.sort(array:ASCENDING, (i) => i);
```

**Affected Versions:**
n/a

n/a

https://github.com/ballerina-platform/ballerina-lang/issues/25781
https://github.com/ballerina-platform/ballerina-lang/issues/25682 blocked on this.

lasinicl:
Fixes #25848, https://github.com/ballerina-platform/ballerina-lang/pull/25815#discussion_r488719579

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/langlib-test/src/test/resources/test-src/arraylib_test.bal
```diff
@@ -1041,6 +1041,11 @@ function testSort10() {
 
     assertValueEquality(sortedArr3.toString(), "[0,1,10,2,3,6]");
     assertValueEquality(sortedArr3, arr);
+
+    int[] sortedArr4 = arr.sort(array:DESCENDING, (i) => i.toString());
+
+    assertValueEquality(sortedArr4.toString(), "[6,3,2,10,1,0]");
```
MaryamZi, line 1047: Instead of using toString shall we assert the value itself? We lose type information when we assert using toString.

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -68,15 +69,15 @@ public static ArrayValue sort(Strand strand, ArrayValue arr, Object direction, O
...
-                elemType = elementType;
-            }
+            elemType = ((BFunctionType) function.getType()).retType;
             for (int i = 0; i < arr.size(); i++) {
                 sortArr[i][0] = function.call(new Object[]{strand, arr.get(i), true});
                 sortArr[i][1] = arr.get(i);
             }
+            if (elemType.getTag() == TypeTags.UNION_TAG &&
+                    ((BUnionType) elemType).getMemberTypes().size() > 2) {
+                elemType = TypeChecker.getType(sortArr[0][0]);
```
MaryamZi, line 79: Can you please explain what we're trying to do with this block? Is there a sample invocation?

lasinicl:
```
int[] arr = [1, 4, 2];
int[] sortedArr = arr.sort(array:ASCENDING, (i) => i);
```
when there is an arrow expression `((BFunctionType) function.getType()).retType` is a union type containing all ordered types. This is to cater to getting the type of the array members we need to sort in such a scenario.

MaryamZi: 1. Shall we add a comment explaining why we're doing it? Also it's not just an arrow function right? There can even be a normal function with a compatible union type as the return type?
2. A union should have at least 2 members. If it has only 2, don't we have to change `elemType`?
3. Can't `sortArr` or `sortArr[0]` be empty here?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5382,12 +5382,15 @@ private void checkArrayLibSortFuncArgs(BLangInvocation iExpr) {
...
             BLangArrowFunction arrowFunction = ((BLangArrowFunction) keyFunction);
-            pos = arrowFunction.params.get(0).pos;
-            returnType = arrowFunction.params.get(0).type;
+            pos = arrowFunction.body.expr.pos;
+            returnType = arrowFunction.body.expr.type;
         } else {
             BLangLambdaFunction keyLambdaFunction = (BLangLambdaFunction) keyFunction;
             pos = keyLambdaFunction.function.pos;
             returnType = keyLambdaFunction.function.type.getReturnType();
+            if (returnType.tag == TypeTags.SEMANTIC_ERROR) {
```
MaryamZi, line 5391: Do we have tests covering this? Is this only an issue with lambda functions?

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
lasinicl: 1. Added a comment. That condition checking was only added for arrow functions. When it's an arrow function we definitely know that the function return type is a union with more than 2 member types.  In other cases if return type is a union it can contain maximum of two members. Thus this condition is checked `if (elemType.getTag() == TypeTags.UNION_TAG && ((BUnionType) elemType).getMemberTypes().size() > 2)` and the type of values in sortArr is taken.
2. This is already handled in https://github.com/ballerina-platform/ballerina-lang/blob/4fc699331c77169e5668cf013ce06ae69a572d84/langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java#L92
3. yes. therefore changed the position of the condition checking for arrow functions

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
lasinicl: Added a test. It's an issue with arrow functions

grainier:
@MaryamZi @mohanvive, shall we merge this and move further reviews/fixes to another issue/pr?  @lasinicl can you create an issue to keep track of that?

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -69,13 +70,17 @@ public static ArrayValue sort(Strand strand, ArrayValue arr, Object direction, O
...
             for (int i = 0; i < arr.size(); i++) {
                 sortArr[i][0] = function.call(new Object[]{strand, arr.get(i), true});
+                // Get the type of the sortArr elements when there is an arrow expression as the key function
+                if (elemType.getTag() == TypeTags.UNION_TAG &&
+                        ((BUnionType) elemType).getMemberTypes().size() > 2) {
+                    BType sortArrElemType = TypeChecker.getType(sortArr[i][0]);
+                    if (sortArrElemType.getTag() != TypeTags.NULL_TAG) {
+                        elemType = sortArrElemType;
+                    }
+                }
```
MaryamZi, line 83: Do we need to check this element by element? This gets overridden right?

MaryamZi (approved):
Approving this since this is blocking other tasks. @lasinicl can you send a PR to fix the issue asap?

**[Update]**: discussed offline with @grainier and @lasinicl and decided to fix this before merging.

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Sort.java
```diff
@@ -68,14 +69,20 @@ public static ArrayValue sort(Strand strand, ArrayValue arr, Object direction, O
 
         Object[][] sortArr = new Object[arr.size()][2];
         Object[][] sortArrClone = new Object[arr.size()][2];
+        boolean elementTypeIdentified = false;
```
MaryamZi, line 72: Shall we move this to within the if block?

MaryamZi (approved):

Issue #25254, opened 2020-08-13
Refactor non-blocking extern functions in stdlib


manuranga:
currently we use `NonBlockingCallback` class to manage extern functions in stdlib. It is a solution stdlib team has comeup with since there is no given solution from the Runtime API itself. It also can lead to some issues such as #18763. We need to give a specific API for this from the Runtime and replace usages in std lib.

manuranga:
All references to NonBlockingCallback is now removed, safe to delete.

PR #25854, opened 2020-09-16
Migrate NonBlockingCallback usage to BalEnv

manuranga:
sub task of #25254

### DIFF
--- a/stdlib/http/src/main/java/org/ballerinalang/net/http/DataContext.java
+++ b/stdlib/http/src/main/java/org/ballerinalang/net/http/DataContext.java
@@ -19,12 +19,12 @@
 package org.ballerinalang.net.http;
 
 import org.ballerinalang.jvm.api.BValueCreator;
+import org.ballerinalang.jvm.api.BalFuture;
 import org.ballerinalang.jvm.api.values.BError;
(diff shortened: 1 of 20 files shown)


stdlib/http/src/main/java/org/ballerinalang/net/http/DataContext.java
```diff
@@ -38,21 +38,21 @@
     private Strand strand;
     private HttpClientConnector clientConnector;
     private BObject requestObj;
-    private NonBlockingCallback callback;
+    private BalFuture future;
     private HttpCarbonMessage correlatedMessage;
 
-    public DataContext(Strand strand, HttpClientConnector clientConnector, NonBlockingCallback callback,
+    public DataContext(Strand strand, HttpClientConnector clientConnector, BalFuture future,
```
warunalakshitha, line 44: Do we still need Strand?

manuranga: It's still used for `dataContext.getStrand().getProperty(HttpConstants.REMOTE_ADDRESS)` so can't remove if fully yet.

warunalakshitha (approved):

Issue #25893, opened 2020-09-17
Can't refer functions, typedescs, etc. in isolated functions and as default values


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type Bar record {|
|};

function foo() {    
}

public isolated function main() {
    typedesc t = Bar;
    function () f = foo;
}
```

```cmd
$ ballerina build xyz.bal 
Compiling source
        xyz.bal
error: .::xyz.bal:9:18: invalid access of mutable storage in an 'isolated' function
error: .::xyz.bal:10:21: invalid access of mutable storage in an 'isolated' function
```

**Affected Versions:**
Current master.

Issue #25898, opened 2020-09-17
Introduce isolated module level variables


MaryamZi:
**Description:**
See https://github.com/ballerina-platform/ballerina-spec/blob/master/lang/proposals/isolated/isolated.md#isolated-module-level-variables

MaryamZi:
Also see https://github.com/ballerina-platform/ballerina-lang/pull/25899/files#r490368145

PR #25899, opened 2020-09-17
Allow using refs of module definitions within isolated functions

MaryamZi:
## Purpose
$title.

This PR also makes incorrect default value usage a warning only for ballerina and ballerinax modules. For other modules such usage will now result in an error.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/25893

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/record/RecordWithClosureTest.java
```diff
@@ -37,7 +37,7 @@ public void setup() {
         compileResult = BCompileUtil.compile("test-src/record/record_closure_default.bal");
     }
 
-    @Test(description = "Test record type with closure variable in default value")
+    @Test(description = "Test record type with closure variable in default value", enabled = false)
```
MaryamZi, line 40: The test disabled is https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/jballerina-unit-test/src/test/resources/test-src/record/record_closure_default.bal#L1.

With the requirement that record field defaults have to be isolated, based on the current implementation I don't think we can test what we originally tested here.

But disabled this without removing it, since we may be able to add back a different version of the test once we support isolated module level variables (https://github.com/ballerina-platform/ballerina-lang/issues/25898).

hasithaa (approved):

pubudu91 (approved):

PR #25902, opened 2020-09-17
Add an API for retrieving the workspace symbols

pubudu91:
## Purpose
This PR adds an API to the semantic model to get the list of symbols of the module level constructs in the model.

## Approach
- Get the scope entries from the module symbol.
- Iterate through the scope entries and create API symbol wrappers for symbols whose origin is `SOURCE`.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/api/SemanticModel.java
+++ b/compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/api/SemanticModel.java
@@ -51,6 +51,13 @@ public interface SemanticModel {
     Optional<Symbol> symbol(String srcFile, LinePosition position);
 
     /**
+     * Retrieves the symbols of module-scoped constructs in the semantic model.
+     *
+     * @return A list of module-scoped symbols
+     */
+    List<Symbol> moduleLevelSymbols();
+
+    /**
      * Get the diagnostics within the given text Span.
      *
      * @param textRange Text range to filter the diagnostics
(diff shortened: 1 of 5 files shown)


compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/api/SemanticModel.java
```diff
@@ -50,6 +50,13 @@
      */
     Optional<Symbol> symbol(String srcFile, LinePosition position);
 
+    /**
+     * Retrieves the symbols of module-scoped constructs in the semantic model.
+     *
+     * @return A list of module-scoped symbols
+     */
+    List<Symbol> moduleLevelSymbols();
```
pubudu91, line 58: Not entirely sure about this name. Please suggest any alternatives if you have any. This basically retrieves all the symbols in the module scope, both public and non-public.

SupunS (approved):

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/api/SemanticModel.java
nadeeshaan: We need to  consider this along with, how are we going to define the term `Workspace` in ballerina. Until then, may be we can go with a name similar to `symbolsInModule`

PR #25939, opened 2020-09-18
Change observe module functions used by ballerina/http to isolated

nadundesilva:
# Purpose
> Change observe module functions used by ballerina/http to isolated

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/observability/ObserverContext.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/observability/ObserverContext.java
@@ -20,11 +20,11 @@ package org.ballerinalang.jvm.observability;
 import org.ballerinalang.jvm.observability.metrics.Tag;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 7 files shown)


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/observability/ObserverContext.java
```diff
@@ -73,9 +73,9 @@
     private boolean isSystemSpan;
 
     public ObserverContext() {
-        this.properties = new HashMap<>();
-        this.mainTags = new HashMap<>();
-        this.additionalTags = new HashMap<>();
+        this.properties = new ConcurrentHashMap<>();
```
Kishanthan, line 76: do we need to change this to ConcurrentHashMap ?

nadundesilva: Since this had not been an issue so far, I have reverted this change with 62949558fcaffac152102ab246d94114db89cf97. I will verify this further and add back the ConcurrentHashMaps later if needed.

Kishanthan (approved):

Issue #25999, opened 2020-09-22
Formatting doesn't work for If-statements 


SupunS:
**Description:**

Sample code:
```ballerina
public function main() {
    if(true) {
        int a = 6;
        if(true) {
            int a = 6;
        } else {
            int a = 5;
        }
    } else {
        int a = 5;
    }
}
```
Formatted output:
```ballerina
public function main() {
    if (true) {
            int a = 6;
        if (true) {
                    int a = 6;
    } else {
                    int a = 5;
    }
    } else {
            int a = 5;
    }
}
```

IrushiL:
Fixed with PR: https://github.com/ballerina-platform/ballerina-lang/pull/26391

Issue #25988, opened 2020-09-21
[Query Expressions] Formatting not works in VSCode Plugin


mohanvive:
**Description:**
Code formatting not works for query expressions and streams in VSCode plugin.

**Steps to reproduce:**
Please checkout the Query and streams examples.
- https://ballerina.io/swan-lake/learn/by-example/query-expression.html
- https://ballerina.io/swan-lake/learn/by-example/query-expression-to-stream.html
- https://ballerina.io/swan-lake/learn/by-example/query-expression-to-string.html
- https://ballerina.io/swan-lake/learn/by-example/query-expression-to-table.html
- https://ballerina.io/swan-lake/learn/by-example/query-expression-to-xml.html
- https://ballerina.io/swan-lake/learn/by-example/query-action.html
- https://ballerina.io/swan-lake/learn/by-example/query-join.html
- https://ballerina.io/swan-lake/learn/by-example/streams.html

IrushiL:
Fixed with PR: https://github.com/ballerina-platform/ballerina-lang/pull/26391

PR #26008, opened 2020-09-22
[Formatter] Fix several formatting issues

### LINKED ISSUE #26001: Formatting doesn't work for unions with user-defined types
**Description:**
Sample code
```ballerina
public function main() {
    Person|error x;
}
```
Formatted output:
```ballerina
public function main() {
Person|error x;
}
```

### LINKED ISSUE #26002: A extra newline gets added every-time the code is formatted
**Description:**
Sample code
```ballerina
public function main() {
}
```
Output after formatting the code for five (5) times: 
```ballerina
public function main() {

}





```

Note the extra 5 newlines at the end of the file.

### LINKED ISSUE #26004: Formatting doesn't work properly for type-descriptors with parenthesis
**Description:**
Sample code:
```ballerina
public function main() {
    (int|string) x;
}
```
Formatted output:
```ballerina
public function main() {
(    int|string) x;
}
```

### LINKED ISSUE #26005: Formatting a while-block result in a syntax error
**Description:**
```ballerina
public function main() {
    while true {
    }
}
```
Formatted output:
```ballerina
public function main() {
    whiletrue {
    }
}
```
The whitespace between the 'while' keyword and the 'condition' has been removed, resulting a syntax error.

IrushiL:
## Purpose

Fixes #26002
Fixes #26001
Fixes #26005
Fixes #26004

Disables the formatting of nested if statements and query expressions in order to reduce the priority levels of the following issues.
#25999
#25988

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
+++ b/misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
@@ -3003,26 +3003,26 @@ public class FormattingTreeModifier extends TreeModifier {
 
     @Override
     public QueryExpressionNode transform(QueryExpressionNode queryExpressionNode) {
-        if (!isInLineRange(queryExpressionNode, lineRange)) {
+//        if (!isInLineRange(queryExpressionNode, lineRange)) {
             return queryExpressionNode;
-        }
-        QueryConstructTypeNode queryConstructType =
-                this.modifyNode(queryExpressionNode.queryConstructType().orElse(null));
-        QueryPipelineNode queryPipeline = this.modifyNode(queryExpressionNode.queryPipeline());
-        SelectClauseNode selectClause = this.modifyNode(queryExpressionNode.selectClause());
-        OnConflictClauseNode onConflictClause = this.modifyNode(queryExpressionNode.onConflictClause().orElse(null));
-        if (queryConstructType != null) {
-            queryExpressionNode = queryExpressionNode.modify()
-                    .withQueryConstructType(queryConstructType).apply();
-        }
-        if (onConflictClause != null) {
-            queryExpressionNode = queryExpressionNode.modify()
-                    .withOnConflictClause(onConflictClause).apply();
-        }
-        return queryExpressionNode.modify()
-                .withQueryPipeline(queryPipeline)
-                .withSelectClause(selectClause)
-                .apply();
+//        }
+//        QueryConstructTypeNode queryConstructType =
+//                this.modifyNode(queryExpressionNode.queryConstructType().orElse(null));
+//        QueryPipelineNode queryPipeline = this.modifyNode(queryExpressionNode.queryPipeline());
+//        SelectClauseNode selectClause = this.modifyNode(queryExpressionNode.selectClause());
+//        OnConflictClauseNode onConflictClause = this.modifyNode(queryExpressionNode.onConflictClause().orElse(null));
+//        if (queryConstructType != null) {
+//            queryExpressionNode = queryExpressionNode.modify()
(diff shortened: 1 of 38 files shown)


sanjana:
Fixes: #26004

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
```diff
@@ -2996,26 +3005,26 @@ public SelectClauseNode transform(SelectClauseNode selectClauseNode) {
...
+//                    .withQueryConstructType(queryConstructType).apply();
+//        }
+//        if (onConflictClause != null) {
+//            queryExpressionNode = queryExpressionNode.modify()
+//                    .withOnConflictClause(onConflictClause).apply();
+//        }
+//        return queryExpressionNode.modify()
+//                .withQueryPipeline(queryPipeline)
+//                .withSelectClause(selectClause)
+//                .apply();
```
sanjana, line 3027: Shall we get rid of these comments?

IrushiL: I think it's okay to leave it there since we have to enable it again anyway.

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
```diff
@@ -912,7 +914,7 @@ public SimpleNameReferenceNode transform(SimpleNameReferenceNode simpleNameRefer
 
     @Override
     public IfElseStatementNode transform(IfElseStatementNode ifElseStatementNode) {
-        if (!isInLineRange(ifElseStatementNode, lineRange)) {
+        if (!isInLineRange(ifElseStatementNode, lineRange) || !nestedIfBlock(ifElseStatementNode).isEmpty()) {
```
sanjana, line 917: This will create an impact when formatting the following type of nested-if code. 

```
public function foo() {
    if (true) {
        
        if (1 < 2) {
            int x = 2;
        }
    } else {
        int y = 0;
        return i;
    }
}
```

IrushiL: Yeah, that is added to disable the formatting of nested if blocks, since this fix requires some time. The intention is to leave the nested if block as it is now. The following issue is there to track the fix.
https://github.com/ballerina-platform/ballerina-lang/issues/25999

SupunS (changes requested):
Following seems to still give some malformed output:
```ballerina
public function foo() {
    if (true) {
        (string|error) txt = rp.getTextPayload();
        // hello world
        if (txt is string) {
            io:println("response.../n", txt);
        }
    }
}

function bar() {
    int a;
    // hello
}
```

IrushiL:
@SupunS Fixed the issue pointed and added a test case for it.

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
IrushiL: Since the formatting of nested if blocks are disabled for now if there are formatting errors in the above code snippet, it will not get formatted. 

But if the above code is formatted as it is, then the output will also be the same.

SupunS (approved):

PR #26028, opened 2020-09-23
Fix explicit paramTypes passing for BalEnv accepting functions

### LINKED ISSUE #26027: Fix explicit paramTypes passing for BalEnv accepting functions
**Description:**
When an extern methods accepts `BalEnv` as the first parameter, and the ballerina annotation specifics `paramTypes` it gives a compilation error.
```
error: ballerina/http:1.0.0::http_connection.bal:300:0: {ballerina/java}METHOD_NOT_FOUND 'No such public method 'cancelWebSocketUpgrade' that matches with parameter types '(org.ballerinalang.jvm.values.ObjectValue,long,org.ballerinalang.jvm.api.values.BString)' found in class 'class org.ballerinalang.net.http.nativeimpl.connection.CancelWebSocketUpgrade'
```

**Related PRs**
We have temporary removed such case, need to revert them after fixing this:

https://github.com/ballerina-platform/module-ballerinax-kafka/pull/20
https://github.com/ballerina-platform/module-ballerina-http/pull/58

manuranga:
## Purpose
Fixes #26027

## Approach
If normal exact match fails we re-match with `BalEnv` as the first param.

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/interop/JMethodResolver.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/interop/JMethodResolver.java
@@ -17,6 +17,7 @@
  */
 package org.wso2.ballerinalang.compiler.bir.codegen.interop;
 
+import org.ballerinalang.jvm.api.BalEnv;
 import org.ballerinalang.jvm.api.values.BArray;
 import org.ballerinalang.jvm.api.values.BDecimal;
 import org.ballerinalang.jvm.api.values.BError;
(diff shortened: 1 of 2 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/interop/JMethodResolver.java
```diff
@@ -625,13 +626,23 @@ private JMethod resolveExactMethod(Class<?> clazz, String name, JMethodKind kind
 
         Executable executable = (kind == JMethodKind.CONSTRUCTOR) ? resolveConstructor(clazz, paramTypes) :
                 resolveMethod(clazz, name, paramTypes);
+        if (executable == null) {
```
warunalakshitha, line 629: Another way we can do this is by changing validateArgumentTypes logic where we can skip first parameter if it is BalEnv class. But that has effect on normal interop function validation. This also looks good to me.

tests/jballerina-unit-test/src/test/resources/test-src/javainterop/basic/static_method_tests.bal
```diff
@@ -218,7 +218,8 @@ function decimalParamAndReturn(decimal a1) returns decimal = @java:Method {
 } external;
 
  public function addTwoNumbersSlowAsync(int a, int b) returns int = @java:Method {
-    'class:"org/ballerinalang/nativeimpl/jvm/tests/StaticMethods"
+    'class:"org/ballerinalang/nativeimpl/jvm/tests/StaticMethods",
```
warunalakshitha, line 221: Shall we write a test case for object attached interop function

manuranga: I think this scenario is still broken, added #26054 for this.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/interop/JMethodResolver.java
manuranga: I also consider that approach, I waned to support the case where people have put the `BalEnv` to the `paramTypes` manually as well, this works for both.

warunalakshitha (approved):

PR #26072, opened 2020-09-24
Add an API to retrieve position info from the public symbol API

pubudu91:
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/api/symbols/Symbol.java
+++ b/compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/api/symbols/Symbol.java
@@ -63,7 +63,7 @@ public interface Symbol {
      *
      * @return The position information of the symbol
      */
-    Location position();
+    Location location();
 
     // TODO: Add the annotation attachment API
 }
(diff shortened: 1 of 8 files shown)


compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/api/symbols/Symbol.java
```diff
@@ -44,17 +45,25 @@
...
      */
     Optional<Documentation> docAttachment();
 
+    /**
+     * This retrieves position information of the symbol in the source code. The position given here is the location of
+     * the definition of the symbol.
+     *
+     * @return The position information of the symbol
+     */
+    Location position();
```
pubudu91, line 66: Went with `position()` for this since this is the term we are used to in the compiler. Should this be changed to something else? `location()` for example?

nadeeshaan: What if we change the model name Position? Either way I would +1 for `Position position()` or `Location location()`

nadeeshaan (dismissed):
LGTM

compiler/ballerina-compiler-api/src/main/java/org/ballerina/compiler/api/symbols/Symbol.java
SupunS: +1 for `location()`.
We use the same in the syntax tree as well.

rasika (approved):

Issue #26007, opened 2020-09-22
Publish Ballerina SNAPSHOT on master branch push


ThisaruGuruge:
# Improving Ballerina standard library daily builds

## Ballerina SNAPSHOT Build 
Currently we have a Jenkins job to publish Ballerina SNAPSHOT version to the nexus. This is a manual process. The standard library modules need this SNAPSHOT version to build on. Current process is to trigger the Jenkins job manually and then wait for the build to complete, and then build the standard library modules.

## Proposal
- A Github workflow to be added, which will be triggered when a PR is merged to the `master` branch.
- This workflow will trigger a Ballerina build **without the tests**, and then publish it to the Github packages.
- Once the packages are published, the workflow will trigger top level standard library modules (modules which aren't dependent on the other standard library modules).
- Those modules then trigger the subsequent standard library module builds.

Related: #25789

Issue #25789, opened 2020-09-15
Improve Ballerina Standard Library Daily Build


shafreenAnfar:
# Background

At the moment each standard library module has it is own repository and the health of the each module is tested by running daily builds against the master. This daily build get triggered on a preconfigured cron expression.

Further, since there are standard libraries which depends on other standard libraries, each module must be built in certain order. Therefore, in order to decide When to build What, we have come up with a `level` system. Each module belongs to certain `level` starting from 0 to N. 

# Problem

Even though this allows us to check the status of the each module, this has some drawbacks such as,
- In case something goes wrong, we don't really get updated instantly. 
- Build order of standard library modules is ignored as all the modules get triggered at the same time. Of course, this can be fixed by changing the trigger time of each module but that is not the ideal solution.

# The short term solution

As the short term, it was decided to move from cron based daily builds to pub-sub style of builds. Idea is quite simple,

1. Each standard library module allows interested parties to get notified every time a build is `completed` on master.
2. Each module that depends on a certain module, need to subscribe to this build event.

What this means in terms of github actions is each module needs to have a final step which triggers a build on interested parties. Each interested party needs to have configured `repository_dispatch` event in their daily build workflow.

**Note**: Since all the modules depends the `ballerina-lang`, all the modules are required to subscribe for build events on `ballerina-lang`

## Issues with the short term solution

Even though this solves the aforementioned problems, it does have it is own problems such as,

- Each module is responsible for subscribing in every module it depends on.
- This could trigger redundant builds. For instance, some of the language changes require standard libraries to be built level by level. But in the propose solution ignores the level each module belongs to.

# The long term solution

Centralized build pipeline is needed to solve this problem properly. Something similar to the `release pipeline`. Following is the high level idea,

- Each module needs to updated an entry in the centralized build pipeline with all the dependencies. 
- Each time a module is built, that particular module notifies the centralized build pipeline. Then the centralized build pipeline triggers a list of builds based on dependencies. 

**Note:** This centralized build pipeline can be made smarter and smarter. For instance, it can be improved to automatically update the dependency entry of each modules by walking each repo. 

This solution is complex compared to the `the short term solution` but has no drawbacks.

shafreenAnfar:
@ThisaruGuruge @Maninda @BuddhiWathsala @niveathika do let me know if I have missed anything.

ThisaruGuruge:
[...]

PR #26083, opened 2020-09-24
Adding Standard Library Publishing Workflow

ThisaruGuruge:
## Purpose [Don't Merge Until All The STDLIB modules Update their Workflows] 

When there's a change in Ballerina-Lang repo, sometimes all the standard library modules should be re-built, incrementally, in-order to work. (For an example, when there's a change in BIR, all the modules have to build on top of that change).

Current process is to manually trigger all the stdlib modules, one-by-one sequentially. This PR is to automate this process.

For the standard library there are two main workflows:

1. stdlib_release_pipeline
2. stdlib_publish_snapshots

### Stdlib Release Pipeline
This workflow is used to release the standard library modules. The pipeline keeps a json file (`stdlib/release/resources/stdlib_modules.json`) to track the standard library modules and their information. This JSON files has the following information:

- **_`name`_**: The name of the module repository. (Eg.: `"module-ballerina-io"`)
- **_`version`_**: The **current development version** of the module. (Eg.: `"0.5.1-SNAPSHOT"`)
- **_`level`_**: The level of the module, corresponding to the standard library module dependency graph. (Eg.: `1`)
- **_`release`_**: Whether to release the module or not. This should be set to `true` if the module is expected to be released when the release pipeline triggers.

When standard library should be released, go to Github actions and run the `Stdlib Release` workflow. (This is a manual step).

### Stdlib Publsih Snapshots
This workflow too uses the aforementioned JSON file. Whenever a commit is pushed to the `master` branch of the `ballerina-lang` repo, this workflow will be triggered. It will first build and publish the `ballerina-lang` artifacts. Then it will build the standard library modules, sequentially. When a set of modules in a level are built, the workflow waits for a predefined amount of time, before trigger the next level module set. This time is determined by the maximum time it takes for a module to build, for a particular level.

Therefore, **if the build time of a module is increased, check for the `getWaitTimeForLevel()` function**, to check whether the there's enough time for that particular module to build, before triggering the next level of modules. If not, update the level wait time. 

Fixes: #26007


shafreenAnfar:
This is great. AFAIU, this is slightly different from the `The short term solution` and `The long term solution` we discussed in #25789. Just to make sure everyone is on the same page. Shall we have a quick meeting ?

Also, IMO, publishing `ballerina-lang` should not have to be part of the standard library but instead getting a notification when `ballerina-lang` published should be.

niveathika:
@ThisaruGuruge Shall we also check if we can indicate the failure for panics or any std release failures? IMO we can improve ballerina action easily for this, https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/setting-exit-codes-for-actions#setting-a-failure-exit-code-in-a-docker-container-action

ThisaruGuruge:
> This is great. AFAIU, this is slightly different from the `The short term solution` and `The long term solution` we discussed in #25789. Just to make sure everyone is on the same page. Shall we have a quick meeting ?

@shafreenAnfar This is similar to the long term solution IINM. Let's have a meeting for this. 

> Also, IMO, publishing `ballerina-lang` should not have to be part of the standard library but instead getting a notification when `ballerina-lang` published should be.

Agreed. Then we have to create a separate workflow to publish `ballerina-lang`

> @ThisaruGuruge Shall we also check if we can indicate the failure for panics or any std release failures? IMO we can improve ballerina action easily for this, https://docs.github.com/en/free-pro-team@latest/actions/creating-actions/setting-exit-codes-for-actions#setting-a-failure-exit-code-in-a-docker-container-action

@niveathika Yes. We should. Actually, there's an issue with Ballerina tool, when panic occurs, ballerina will return 0 as the exit code. This should be fixed from the Ballerina tool I guess. Already [created an issue](https://github.com/ballerina-platform/ballerina-lang/issues/26084) for this.

shafreenAnfar (approved):

PR #26173, opened 2020-10-01
Add missing similarity criteria between array and tuple types 

### LINKED ISSUE #23676: Similarities between array and tuple types not fully honored

suleka96:
Fixes #23676

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1242,32 +1247,45 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
                 tupleTypes.add(sourceTupleType.getRestType());
             }
             if (tupleTypes.isEmpty()) {
-                return targetType.getState() == ArrayState.UNSEALED || targetType.getSize() == 0;
+                return targetType.getState() == ArrayState.OPEN || targetType.getSize() == 0;
+            }
+
+            if (sourceTupleType.getRestType() != null && targetType.getState() == ArrayState.OPEN) {
+                boolean memberTypesMatch = true;
+                if (!(tupleTypes.isEmpty())) {
```
SupunS, line 1255: additional parenthesis

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1242,32 +1247,45 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+                    for (BType sourceElementType : tupleTypes) {
+                        if (!checkIsType(sourceElementType, targetElementType, unresolvedTypes)) {
+                            memberTypesMatch = false;
+                        }
+                    }
+                }
+                if (memberTypesMatch) {
+                    BType sourceRestType = sourceTupleType.getRestType();
+                    return checkIsType (sourceRestType, targetElementType, unresolvedTypes);
```
SupunS, line 1264: seems theres a formatting issue here

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1279,32 +1297,118 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+                case OPEN:
+                    if (targetRestType == null) {
+                        return false;
+                    }
+                    if (targetTypes.isEmpty()) {
+                        return checkIsType(sourceElementType, targetRestType, unresolvedTypes);
+                    }
+                    return false;
+                case CLOSED:
+                    if (sourceArrayType.getSize() >= targetTypes.size()) {
```
SupunS, line 1335: We can remove this if-condition, if you bring the condition in the else-if, before this one
i.e:
```java
if (sourceArrayType.getSize() < targetTypes.size()) {
    return false;
}

// rest of the code goes here...
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1279,32 +1297,118 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+                            if (!(checkIsType(sourceElementType, targetElementType, unresolvedTypes))) {
+                                memberTypesMatch = false;
+                            }
+                        }
+                        if (sourceArrayType.getSize() == targetTypes.size()) {
+                            return memberTypesMatch;
+                        }
+                        if (targetRestType != null) {
+                            return (checkIsType(sourceElementType, targetRestType, unresolvedTypes)
+                                    && memberTypesMatch);
```
SupunS, line 1358: better to do the `memberTypesMatch` check first. Then we can fail-fast

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1279,32 +1297,118 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+            boolean remainingMemberTypesMatch = true;
+
+            for (int i = targetTypeSize; i < sourceTypeSize; i++) {
+                if (!checkIsType(sourceTypes.get(i), targetRestType, unresolvedTypes)) {
+                    remainingMemberTypesMatch = false;
+                }
+            }
+            if (sourceRestType != null) {
+                return (checkIsType (sourceRestType, targetRestType, unresolvedTypes)
+                        && memberTypesMatch && remainingMemberTypesMatch);
```
SupunS, line 1406: same as above

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1279,32 +1297,118 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+        BType sourceRestType = sourceTupleType.getRestType();
+
+        if (sourceRestType != null && targetRestType == null) {
+            return false;
         }
+        int sourceTypeSize = sourceTypes.size();
+        int targetTypeSize = targetTypes.size();
 
-        if (sourceTypes.size() != targetTypes.size()) {
+        if ((sourceRestType == null && targetRestType == null) && sourceTypeSize != targetTypeSize) {
```
SupunS, line 1377: additional parenthesis

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1279,32 +1297,115 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
 
         List<BType> targetTypes = new ArrayList<>(targetType.getTupleTypes());
         BType targetRestType = targetType.getRestType();
-        if (targetRestType != null) {
-            targetTypes.add(targetRestType);
+
+        BTupleType sourceTupleType = null;
+        if (sourceTypeTag == TypeTags.TUPLE_TAG) {
+            sourceTupleType = (BTupleType) sourceType;
+        } else {
```
MaryamZi, line 1321: Since we seem to be returning within the else block shall we refactor the code to do something like

```java
if (sourceTypeTag == TypeTags.ARRAY_TAG) { 
    // move logic in else
}

BTupleType sourceTupleType = (BTupleType) sourceType;
```

Reduces the indentation and the variable will be defined closer to where it is used.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1279,32 +1297,115 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
-        for (int i = 0; i < sourceTypes.size(); i++) {
-            if (!checkIsType(sourceTypes.get(i), targetTypes.get(i), unresolvedTypes)) {
-                return false;
+        if (sourceRestType == null && targetRestType == null && sourceTypeSize != targetTypeSize) {
+            return false;
+        }
+
+        boolean memberTypesMatch = true;
+
+        if (sourceTypeSize >= targetTypeSize) {
```
MaryamZi, line 1380: We can switch the check here also, and remove the else block.

```java
        if (sourceTypeSize < targetTypeSize) {
            return false;
        }

        for (int i = 0; i < targetTypeSize; i++) {
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1279,32 +1297,115 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+            for (int i = 0; i < targetTypeSize; i++) {
+                if (!checkIsType(sourceTypes.get(i), targetTypes.get(i), unresolvedTypes)) {
+                    memberTypesMatch = false;
```
MaryamZi, line 1383: Can't we directly return false here instead of setting it to `memberTypesMatch`?

MaryamZi: Check `remainingMemberTypesMatch` too.

KavinduZoysa:
```
public function foo() returns string {
    any k = [1, 2, 3, 4];
    if (k is [int, any...]) {
        return "match";
    }
    return "No match";
}
```

Please consider the above example, I tested it with this PR, and it does not match. IMO it should be matched. WDYT?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1242,32 +1247,45 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+                    for (BType sourceElementType : tupleTypes) {
+                        if (!checkIsType(sourceElementType, targetElementType, unresolvedTypes)) {
+                            memberTypesMatch = false;
+                        }
+                    }
+                }
+                if (memberTypesMatch) {
+                    BType sourceRestType = sourceTupleType.getRestType();
+                    return checkIsType(sourceRestType, targetElementType, unresolvedTypes);
+                }
```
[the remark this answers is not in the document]
suleka96: Ya actually the logic here is a bit wrong. I re-did the whole tuple evaluation in the checkIsArrayType().

suleka96:
I think it should fail because the `k` holds a `any[]` and since the target type tuple is `[int, any...]` it should fail. It needs to be `[any...]` to be matched IMO.

> A tuple type descriptor T with m member type descriptors contains a list shape L of length n if and only if:
> 
> 1) m is less than or equal to n
> 2) the i-th member type descriptor of T contains the i-th member of L for each i from 1 to m;
> 3) if n is greater than m, then T has a tuple-rest-descriptor R..., and R contains the j-th member of L for each j from m + 1 to n.

This violates the 1st statement.

rdhananjaya:
> I think it should fail because the `k` holds a `any[]` and since the target type tuple is `[int, any...]` it should fail. It needs to be `[any...]` to be matched IMO.
> 
> > A tuple type descriptor T with m member type descriptors contains a list shape L of length n if and only if:
> > 
> > 1. m is less than or equal to n
> > 2. the i-th member type descriptor of T contains the i-th member of L for each i from 1 to m;
> > 3. if n is greater than m, then T has a tuple-rest-descriptor R..., and R contains the j-th member of L for each j from m + 1 to n.
> 
> This violates the 1st statement.

@suleka96 Agreed that this should not be true, but to me this doesn't seems to violate the first rule, rather it seems to violate the second rule.

m = 1 
n = 4
So the first rule (m < n) is not violated.

Sicne this is a `any` array, the first member's type descriptor (int) does not contain the first member's type of the list (which is any).

MaryamZi:
> > I think it should fail because the `k` holds a `any[]` and since the target type tuple is `[int, any...]` it should fail. It needs to be `[any...]` to be matched IMO.
> > > A tuple type descriptor T with m member type descriptors contains a list shape L of length n if and only if:
> > > 
> > > 1. m is less than or equal to n
> > > 2. the i-th member type descriptor of T contains the i-th member of L for each i from 1 to m;
> > > 3. if n is greater than m, then T has a tuple-rest-descriptor R..., and R contains the j-th member of L for each j from m + 1 to n.
> > 
> > 
> > This violates the 1st statement.
> 
> @suleka96 Agreed that this should not be true, but to me this doesn't seems to violate the first rule, rather it seems to violate the second rule.
> 
> m = 1
> n = 4
> So the first rule (m < n) is not violated.
> 
> Sicne this is a `any` array, the first member's type descriptor (int) does not contain the first member's type of the list (which is any).

IMO, target tuple type actually contains the list shape at this particular point in the evaluation.

Here `T` is `[int, any...]`, `L` is a list shape of 4 `int`s but with the inherent type `any[]`, `m` is 1 and `n` is 4.

So 
1. m is less than or equal to n - is satisfied
2. the i-th member type descriptor of T contains the i-th member of L for each i from 1 to m - satisfied since `T` expects an `int` as the first member and `L` has an `int` as the first member too
3. if n is greater than m, then T has a tuple-rest-descriptor R..., and R contains the j-th member of L for each j from m + 1 to n. - satisfied since `n` is greater than `m` and all the remaining members, from the second member to the last member of `L`, belongs to the tuple-rest-descriptor `any`

Why this should fail though is because the belongs-to relationship (upon which the `is` check is based for mutable values) is

Although at this point the value `[1, 2, 3, 4]` looks like `[int, any...]`, since the inherent type is `any[]`, the value can be mutated in a way such that it no longer belongs to the target type.

e.g.,
```
    any[] arr = <any[]> k;
    arr.removeAll();
```

So, the `is` check should fail.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1228,82 +1230,174 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+
         return checkIsType(sourceElementType, targetElementType, unresolvedTypes);
     }
 
     private static boolean checkIsTupleType(BType sourceType, BTupleType targetType, List<TypePair> unresolvedTypes) {
-        if (sourceType.getTag() != TypeTags.TUPLE_TAG) {
-            return false;
+        int sourceTypeTag = sourceType.getTag();
+
+        if (sourceTypeTag == TypeTags.UNION_TAG) {
```
MaryamZi, line 1308: Have we added tests for this? Where the source type is a union of lists?

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1228,82 +1230,174 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
+        if (sourceTypeTag != TypeTags.ARRAY_TAG && sourceTypeTag != TypeTags.TUPLE_TAG) {
+            return false;
         }
 
         List<BType> targetTypes = new ArrayList<>(targetType.getTupleTypes());
         BType targetRestType = targetType.getRestType();
-        if (targetRestType != null) {
-            targetTypes.add(targetRestType);
+
+        if (sourceTypeTag == TypeTags.ARRAY_TAG) {
```
MaryamZi, line 1324: Let's extract these two scenarios also to two methods.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/TypeChecker.java
```diff
@@ -1228,82 +1230,174 @@ private static boolean checkIsArrayType(BType sourceType, BArrayType targetType,
...
             sourceArrayType = (BArrayType) sourceType;
         } else {
             BTupleType sourceTupleType = (BTupleType) sourceType;
             Set<BType> tupleTypes = new HashSet<>(sourceTupleType.getTupleTypes());
-            if (sourceTupleType.getRestType() != null) {
-                tupleTypes.add(sourceTupleType.getRestType());
+            BType sourceRestType = sourceTupleType.getRestType();
+
+            if (sourceRestType != null) {
+                tupleTypes.add(sourceRestType);
```
MaryamZi, line 1249: Instead of adding the rest-descriptor to tupleTypes can we handle it separately?

For example, 
- if the target array is closed, the source cannot have a rest-descriptor (return false directly). If there is no rest-descriptor it may be valid, so check each member against the target array's element type - we seem to be doing this already
- if the target array is open, check member types against element type, and then check the rest-descriptor against the target element type separately - add another check to the else block below

tests/jballerina-unit-test/src/test/resources/test-src/jvm/types.bal
```diff
@@ -157,17 +157,69 @@ function tupleTest() returns int {
...
     assertEquality(y is string[], false);
+    assertEquality(y is [int...], true);
+    assertEquality(y is int[0], false);
+    assertEquality(y is int[], true);
+    assertEquality(y is [int], false);
+    assertEquality(y is int[2], false);
+
+    [int, int, int...] a = [1, 2];
+    any b = a;
+    assertEquality(b is int[2], false);
```
MaryamZi, line 192: Shall we also add tests like
```ballerina
    assertEquality(b is [int, int...], true);
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/TypeChecker.java
```diff
@@ -1246,8 +1246,62 @@ private static boolean hasIncompatibleReadOnlyFlags(Field targetField, Field sou
...
+                    return false;
+                }
+                break;
+        }
+        return checkIsType(sourceElementType, targetElementType, unresolvedTypes);
+    }
+
+    private static boolean checkIsArrayType(BTupleType sourceType, BArrayType targetType,
+                                            List<TypePair> unresolvedTypes) {
+        Set<Type> tupleTypes = new HashSet<>(sourceType.getTupleTypes());
```
MaryamZi, line 1272: Do we need a new set here? Can't we just work with the list returned by `sourceType.getTupleTypes()`?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/TypeChecker.java
```diff
@@ -1256,83 +1310,123 @@ private static boolean checkIsArrayType(Type sourceType, BArrayType targetType,
...
+                return false;
+        }
+    }
 
-        Type sourceElementType = sourceArrayType.getElementType();
+    private static boolean checkIsTupleType(BTupleType sourceType, BTupleType targetType,
+                                            List<TypePair> unresolvedTypes) {
+        List<Type> sourceTypes = new ArrayList<>(sourceType.getTupleTypes());
+        Type sourceRestType = sourceType.getRestType();
+        List<Type> targetTypes = new ArrayList<>(targetType.getTupleTypes());
```
MaryamZi, line 1370: Here again, for both `sourceTypes` and `targetTypes`, since we are no longer adding the rest types to these lists we can directly use the lists returned by `getTupleTypes()` instead of creating new lists.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/TypeChecker.java
```diff
@@ -1256,83 +1307,123 @@ private static boolean checkIsArrayType(Type sourceType, BArrayType targetType,
...
+        return checkIsArrayType((BTupleType) sourceType, targetType, unresolvedTypes);
+    }
 
-        switch (sourceArrayType.getState()) {
-            case UNSEALED:
-                if (targetType.getState() != ArrayType.ArrayState.UNSEALED) {
+    private static boolean checkIsTupleType(BArrayType sourceType, BTupleType targetType,
+                                            List<TypePair> unresolvedTypes) {
+        Type sourceElementType = sourceType.getElementType();
+        List<Type> targetTypes = new ArrayList<>(targetType.getTupleTypes());
```
MaryamZi, line 1323:
```suggestion
        List<Type> targetTypes = targetType.getTupleTypes();
```

MaryamZi (dismissed):

MaryamZi (approved):

PR #26203, opened 2020-10-05
Restrict zero match clauses in match statement

### LINKED ISSUE #25296: Zero match-clauses should be handled
**Description:**
If match-stmt does not include match-clause(s) it should be a syntax error, because according to the spec there should be at least one or more match-clauses

**Steps to reproduce:**

```
function func(any v){
    match v {
    }
}

public function main() {
    func(2);
}
```

rdulmina:
## Purpose
$Subject

Fixes #25296

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/test/resources/statements/match-stmt/match_stmt_source_16.bal
```diff
@@ -0,0 +1,4 @@
+function func(any v){
```
anupama-pathirage, line 1: Missing license header.

rdulmina: Added thanx.

SupunS: We haven't added licence headers to any of the parser test bal-sources. The reason for that was, these parser tests operates at a very granular level, and tests a specific grammar rule in each test-case.

Having the license header means, parser has to parse additional grammar rules as well - then we no longer have that granularity in the test..

Can we stick to the same?

rdulmina: reverted

anupama-pathirage: +1. Thanks for the info @SupunS

compiler/ballerina-parser/src/main/resources/syntax_diagnostic_message.properties
```diff
@@ -235,7 +235,8 @@ error.more.field.match.patterns.after.rest.field=more field match patterns after
 error.action.as.a.wait.expr=''wait'' cannot be used with actions
 error.invalid.usage.of.var='var' is not a valid type descriptor
 error.more.match.patterns.after.rest.match.pattern=more match patterns after rest match pattern
-error-match-pattern-not-allowed=match pattern not allowed
+error.match.pattern.not.allowed=match pattern not allowed
+error.at.least.one.match.clause.required=error at least one match clause required
```
SupunS, line 239: shall we reword the error message?
"match statement is required to have one or more match clauses"

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -11802,7 +11802,19 @@ private STNode parseMatchStatement() {
...
+        // Parse match clauses list
+        List<STNode> matchClausesList = new ArrayList<>();
+        while (!isEndOfMatchClauses(peek().kind)) {
+            STNode clause = parseMatchClause();
+            matchClausesList.add(clause);
+        }
+        STNode matchClauses =  STNodeFactory.createNodeList(matchClausesList);
+        // At least one match clause required
+        if (isNodeListEmpty(matchClauses)) {
+            return SyntaxErrors.addDiagnostic(openBrace,
```
SupunS, line 11814: shouldn't be returning here

SupunS (changes requested):

compiler/ballerina-parser/src/main/resources/syntax_diagnostic_message.properties
rdulmina: updated

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
rdulmina: updated thanx

SupunS (approved):

PR #26216, opened 2020-10-05
Map internal symbol flags to qualifiers in public symbol API

pubudu91:
## Purpose
This PR exposes some of the flags in the internal symbol as qualifiers in the public symbol API.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


nadeeshaan (commented):
Looks good to me.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
```diff
@@ -1095,7 +1095,7 @@ private BLangNode createService(Node serviceNode, IdentifierToken serviceNameNod
         // 1) Define type nodeDefinition for service type.
         BLangClassDefinition classDef = (BLangClassDefinition) TreeBuilder.createClassDefNode();
         BLangIdentifier serviceTypeID = createIdentifier(identifierPos, serviceTypeName);
-        serviceTypeID.pos = pos;
+        serviceTypeID.pos = symTable.builtinPos;
```
SupunS, line 1098: Won't this result in some incorrect line-numbers for server declarations?

compiler/ballerina-compiler-api/src/main/java/io/ballerina/compiler/api/symbols/FunctionSymbol.java
```diff
@@ -19,27 +19,26 @@
...
  * Represent Function Symbol.
  *
  * @since 2.0.0
  */
-public interface FunctionSymbol extends Symbol {
+public interface FunctionSymbol extends Symbol, Qualifiable, Deprecatable {
 
     /**
-     * Get the list of qualifiers.
+     * Get the Function Type Descriptor.
```
SupunS, line 32: Should this be sentence-case?

compiler/ballerina-compiler-api/src/main/java/io/ballerina/compiler/api/symbols/FunctionSymbol.java
```diff
@@ -19,27 +19,26 @@
...
      *
-     * @return {@link List} of qualifiers
+     * @return {@link BallerinaTypeDescriptor}
      */
-    List<Qualifier> qualifiers();
+    Optional<BallerinaTypeDescriptor> typeDescriptor();
```
SupunS, line 36: Can the type-desc be null?

compiler/ballerina-compiler-api/src/main/java/io/ballerina/compiler/api/symbols/Qualifiable.java
```diff
@@ -0,0 +1,35 @@
...
+ * @since 2.0.0
+ */
+public interface Qualifiable {
+
+    /**
+     * Gets the set of qualifiers the user has applied to a symbol.
+     *
+     * @return The set of qualifiers applied to the symbol
+     */
+    Set<Qualifier> qualifiers();
```
SupunS, line 34: IIUC only the worker-symbol (may be not?) and xmlns-symbol doesn't have any qualifiers?
So can we move this method to the `Symbol` interface, and return an empty list for whoever doesn't have any qualifiers?
That would simplify the class hierarchy.. and can move this qualifiers in to the abstract `BallerinaSymbol` class as well in the implementation

compiler/ballerina-compiler-api/src/main/java/io/ballerina/compiler/api/symbols/FunctionSymbol.java
pubudu91: Yeah, should be. Will change.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
pubudu91: Hmm.. Good point. Will check this and see. It looks like we are not skipping class defs generated for services. Will have to filter these in another way if that's the case.

compiler/ballerina-compiler-api/src/main/java/io/ballerina/compiler/api/symbols/FunctionSymbol.java
pubudu91: Hmm. Can't think of a reason why it'd be. Will check with Nadeeshaan and see. All the `typeDescriptor()` methods seem to return `Optional` values.

compiler/ballerina-compiler-api/src/main/java/io/ballerina/compiler/api/symbols/Qualifiable.java
pubudu91: And module and service symbols as well though.

compiler/ballerina-compiler-api/src/main/java/io/ballerina/compiler/api/symbols/FunctionSymbol.java
nadeeshaan: This was because the earlier BOperatorSymbols were converted into the InvokableSymbols and the type became null during the conversion, I think we can remove the optional here

compiler/ballerina-compiler-api/src/main/java/io/ballerina/compiler/api/symbols/Qualifiable.java
nadeeshaan: If the qualifiers are not allowed, shouldn't we avoid exposing the API as well?

pubudu91: Yeah, that was my thought as well. Hence this approach

SupunS (approved):

SupunS (approved):

PR #26315, opened 2020-10-12
Change the NodeResolver to return the symbol

pubudu91:
## Purpose
This PR changes the behaviour of the visitor used earlier to find the enclosing node for a given position to return the symbol itself.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -107,23 +103,14 @@ public BallerinaSemanticModel(BLangPackage bLangPackage, CompilerContext context
...
-        } else if (node != null && (node.getKind() == USER_DEFINED_TYPE
-                || node.getKind() == NodeKind.UNION_TYPE_NODE
-                || node.getKind() == NodeKind.INTERSECTION_TYPE_NODE
-                || node.getKind() == NodeKind.VALUE_TYPE
-                || node.getKind() == NodeKind.BUILT_IN_REF_TYPE
-                || node.getKind() == NodeKind.CONSTRAINED_TYPE)) {
-            return Optional.ofNullable(
-                    SymbolFactory.createTypeDefinition(node.type.tsymbol, node.type.tsymbol.name.value));
+        SymbolFinder symbolFinder = new SymbolFinder();
+        BSymbol symbolAtCursor = symbolFinder.lookup(compilationUnit, position);
```
rasika, line 107: Can we return an `Optional<BSymbol>` omitting the null check. Later in L113; we can use  a ternary for the return of `SymbolFactory.getBCompiledSymbol(...)` or `Optional.empty()`. As a further improvement we can return optional `Symbol` from the `SymbolFactory.getBCompiledSymbol(...)` as well.

pubudu91: Not different from doing a null check right? And `SymbolFactory.getBCompiledSymbol()` should always be returning a `Symbol`, provided a valid `BSymbol`. Will update it to validate the `BSymbol`. Any filtering should be done before calling the factory method.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/SymbolFactory.java
```diff
@@ -92,7 +92,11 @@ public static Symbol getBCompiledSymbol(BSymbol symbol, String name) {
             return createTypeDefinition((BTypeSymbol) symbol, name);
         }
 
-        return null;
+        if (symbol == null) {
+            throw new IllegalArgumentException("Symbol is 'null'");
+        }
+
+        throw new IllegalArgumentException("Unsupported symbol kind: " + symbol.getClass().getName());
```
nadeeshaan, line 99: Error message tells invalid kind, instead we show the class name. Shall we modify either the message or the input arg

nadeeshaan (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/SymbolFactory.java
pubudu91: Changed kind -> type.

SupunS (approved):

nadeeshaan (approved):

PR #26331, opened 2020-10-13
Add testcases to cover the edge scenarios of on-fail clause

mohanvive:


tests/jballerina-unit-test/src/test/resources/test-src/statements/onfail/on-fail-clause-negative-v2.bal
```diff
@@ -0,0 +1,15 @@
+import ballerina/io;
```
anupama-pathirage, line 1: Shouldn't we add the license header for the `.bal` files?

mohanvive: Hmm. That's a good question. If we check the test files in this repo, then we can find .bal files with and without license header. Do we have a standard on this?

anupama-pathirage: IINM, we didn't have this on old .bal files as we were not able to put comments before the imports on early days. Later that restriction was removed and we continued to have headers for .bal files. 

May be @hasithaa can give more accurate answer.

mohanvive: Oh, OK. Let me add it.

pcnfernando (approved):

PR #26333, opened 2020-10-13
Filter symbols of nodes added by the parser

pubudu91:
## Purpose
This PR marks symbols defined for nodes added by the parser when recovering as `VIRTUAL` so that they are filtered out when looking up symbols.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BSymbol.java
@@ -34,6 +34,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 
+import static org.ballerinalang.model.symbols.SymbolOrigin.VIRTUAL;
+
 /**
  * @since 0.94
  */
@@ -66,7 +68,12 @@ public class BSymbol implements Symbol {
         this.type = type;
         this.owner = owner;
         this.pos = pos;
-        this.origin = origin;
+
+        if (name.value.startsWith("$missingNode$")) {
+            this.origin = VIRTUAL;
+        } else {
+            this.origin = origin;
+        }
     }
 
     public MarkdownDocAttachment getMarkdownDocAttachment() {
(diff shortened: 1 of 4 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BSymbol.java
```diff
@@ -66,7 +68,12 @@ public BSymbol(int tag, int flags, Name name, PackageID pkgID, BType type, BSymb
         this.type = type;
         this.owner = owner;
         this.pos = pos;
-        this.origin = origin;
+
+        if (name.value.startsWith("$missingNode$")) {
```
SupunS, line 72: Can we do this check at the symbol enter phase? Then we can use the `BLangMissingNodesHelper`

pubudu91: Can do. The concern is whether symbol enter is the only place where a `SOURCE` symbol can be created.

nadeeshaan: I think it would be better to do it within this constructor if we can't ensure Pubudu's concern and this is always the common place

nadeeshaan (dismissed):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BSymbol.java
SupunS: IMO creator of the symbol should give the proper origin. 

Also, having to hardcode "$missingNode$" smells a problem

pubudu91: Will move it to symbol enter and see. If any issues arise, can address them case by case I guess

SupunS (approved):

PR #26422, opened 2020-10-20
Revamp ballerina parser qualifier parsing

### LINKED ISSUE #26014: Revisit isolated qualifier parsing in BallerinaParser
**Description:**
There's an ambiguity for isolated qualifier alone.
For example it could be a start of a,

- function-definition
- function-type-descriptor
- module-class-definition
- object-type-descriptor
- anonymous-function-expr

As with PR #25778 ambiguity is resolved considering the next immediate token.
For example, if isolated keyword is followed by transactional or function keyword it is considered to be a start of funcDefOrFuncTypeDesc.
This is not a very good approach as there could be invalid qualifiers following the isolated qualifier.
   
Solution: To overcome above issue, can parse all qualifiers at the very beginning and then validate at the end.
Can have something like parseTopLevelQualifiers and go to recovery with next rule defined as object/func rules instead of going through `ParserRuleContext.TOP_LEVEL_NODE_WITHOUT_MODIFIER`

When fixing this issue. Can enable following parser recovery test cases.

- https://github.com/ballerina-platform/ballerina-lang/pull/25778/files?file-filters%5B%5D=.java#diff-2f879b8b773f801dc8fd1fadded3fd8bR76
- https://github.com/ballerina-platform/ballerina-lang/pull/25778/files?file-filters%5B%5D=.java#diff-5dbb64fe81cd35489057bfce6a0f45caR39

**Suggested Labels:**
Team/CompilerFE, Area/Parser, Type/Improvement

### LINKED ISSUE #26176: Add parser recovery for transactional worker keyword
Parser recovery  is not added for transactional workers. As of the current implementation, The below code snippets can break the parser .

 if there are additional tokens like:
```ballerina
transactional someJunkToken worker w1 {

}
```
or if the worker keyword is missing like:

```ballerina
transactional  w1 {

}
```

lochana-chathura:
## Purpose
$subject.

Fixes #26014 
Fixes #26176 

## Approach
Methods performing qualifier based directions are given self qualifier parsing abilities thus allowing them to make decisions without taking qualifiers into account. Further, those methods are generalized so that It doesn't matter whether qualifiers are already passed in a previous stage and come. 

## Samples
N/A

## Remarks
Following methods perform qualifier based directions. 
- parseTopLevelNode
- parseTypeDescriptorInternal
- parseStatement
- parseObjectMemberWithoutMeta
- parseObjectMethodOrField
- parseTypedBindingPatternOrExpr
- parseTypeDescOrExpr
- parseStatementStartBracketedListMember

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples
  - [ ] API documentation


compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/declarations/FunctionDefinitionTest.java
```diff
@@ -157,11 +157,6 @@ public void testFunctionWithInvalidExpressionStatement() {
         testFile("func-definition/func_def_source_24.bal", "func-definition/func_def_assert_24.json");
     }
 
-    @Test
-    public void testMissingFunctionKeywordWithQualifiers() {
-        testFile("func-definition/func_def_source_25.bal", "func-definition/func_def_assert_25.json");
-    }
-
```
rdulmina, line 164: Any reason for removing these test cases?

lochana-chathura: Now that parsing is independent of qualifiers
```bal
public isolated transcational foo(int i) returns boolean { }
```
is equivalent to
```bal
public foo(int i) returns boolean { }
```
Removed the test cases(5 tests) which were previously recovered based on the preceding qualifiers with a875674

lochana-chathura: Anyway consider following code snippet.
```bal
public foo() returns boolean { }

public foo(int i) returns boolean { }
```

Recovery is
```bal
public  MISSING[function]foo() returns boolean { }

 INVALID[public] foo MISSING[error] MISSING[]( INVALID[int] i)  MISSING[;] MISSING[object] INVALID[returns]  INVALID[boolean] { } MISSING[] MISSING[;]
````
in the current master.

It seems something is wrong with the recovery of functions with parameters

File: compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java [hunk 2]
```diff
@@ -739,111 +739,117 @@ private STNode parseImportPrefix() {
...
 
-    private boolean isObjectTypeStart() {
-        return peek(2).kind == SyntaxKind.OBJECT_KEYWORD || peek(3).kind == SyntaxKind.OBJECT_KEYWORD;
+    private STNode parseModuleVarDecl(STNode metadata, STNode qualifier) {
+        List<STNode> topLevelQualifiers = new ArrayList<>();
+        return parseModuleVarDeclWithoutQuals(metadata, qualifier, topLevelQualifiers);
     }
 
-    private STNode parseModuleVarDecl(STNode metadata, STNode qualifier) {
+    private STNode parseModuleVarDeclWithoutQuals(STNode metadata, STNode qualifier, List<STNode> qualifiers) {
```
SupunS, line 849: Its fine to call this `parseModuleVarDecl`. The overloaded parameters will indicate this takes qualifiers from outside..

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -859,21 +865,122 @@ private void reportInvalidMetaData(STNode metadata) {
...
-            recover(token, ParserRuleContext.PUBLIC_KEYWORD);
-            return parseQualifier();
+    private void reportInvalidQualifierList(List<STNode> qualifiers) {
+        for (STNode qual : qualifiers) {
+            addInvalidNodeToNextToken(qual, DiagnosticErrorCode.ERROR_QUALIFIER_NOT_ALLOWED, ((STToken) qual).text());
+        }
+    }
+
+    private void reportInvalidStatementAnnots(STNode annots, List<STNode> qualifiers) {
+        if (!isNodeListEmpty(annots)) {
```
SupunS, line 875: Inverting the condition can avoid the nesting if's:
```java
if (isNodeListEmpty(annots)) {
    return;
}

File: compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java [hunk 4]
```diff
@@ -899,10 +1006,70 @@ private STNode parseFuncDefinition(STNode metadata, boolean isObjectMember, STNo
...
+        }
         return parseFuncDefOrFuncTypeDesc(metadata, qualifierList, isObjectMember, isObjectTypeDesc);
     }
 
+    private STNode createFuncDefNodeList(List<STNode> qualifierList) {
+        // This method is temporary. Will be removed when syncing to service-typing branch
+        // Validate qualifiers and create a STNodeList
+        for (int i = 0; i < qualifierList.size();) {
+            STNode qualifier = qualifierList.get(i);
+            if (qualifier.kind != SyntaxKind.ISOLATED_KEYWORD &&
```
SupunS, line 1024: Inverting the condition would be better

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -1690,93 +1863,85 @@ private STNode validateForUsageOfVar(STNode typeDesc) {
...
-                }
+                reportInvalidQualifierList(qualifiers);
+                STNode distinctKeyword = consume();
+                return parseDistinctTypeDesc(distinctKeyword, context);
             default:
+                if (isTypeDescQualifier(nextToken.kind)) {
+                    // Type desc qualifiers are present. Parse them first and then try to re-parse the method.
+                    parseTypeDescQualifiers(qualifiers);
+                    return parseTypeDescriptorInternal(qualifiers, context, isInConditionalExpr);
+                }
```
SupunS, line 1924: What if we move this, entirely before the switch-case, and only call `parseTypeDescQualifiers(qualifiers);`, (without the if-check and without the re-parse)?
inside `parseTypeDescQualifiers(qualifiers)` we can have a while loop, rather than the do-while.

SupunS: Can do the same for top-level nodes as well

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -2490,50 +2646,90 @@ private STNode parseClassDefinition(STNode metadata, STNode qualifier) {
...
-            STNode nextToken = peek();
-            if (isNodeWithSyntaxKindInList(qualifiers, nextToken.kind)) {
-                qualifier = consume();
-                updateLastNodeInListOrAddInvalidNodeToNextToken(qualifiers, nextToken,
-                        DiagnosticErrorCode.ERROR_DUPLICATE_QUALIFIER, ((STToken) qualifier).text());
+    private STNode createClassTypeQualNodeList(List<STNode> qualifierList) {
+        // Validate qualifiers and create a STNodeList
+        for (int i = 0; i < qualifierList.size();) {
+            STNode qualifier = qualifierList.get(i);
+            if (!isClassTypeQual(qualifier.kind)) {
```
SupunS, line 2684: Invert the condition. Let's do the early exit in all places

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -2490,50 +2646,90 @@ private STNode parseClassDefinition(STNode metadata, STNode qualifier) {
...
+     * <code> object-type-quals := (client | isolated)*</code>
+     *
+     * @param qualifierList Qualifier list to be validated
+     * @return Parsed node
+     */
+    private STNode createObjectTypeQualNodeList(List<STNode> qualifierList) {
+        // Validate qualifiers and create a STNodeList
+        for (int i = 0; i < qualifierList.size();) {
+            STNode qualifier = qualifierList.get(i);
+            if (!isObjectTypeQual(qualifier.kind)) {
```
SupunS, line 2716: same here

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -899,10 +1006,70 @@ private STNode parseFuncDefinition(STNode metadata, boolean isObjectMember, STNo
...
+    private STNode createMethodQualNodeList(List<STNode> qualifierList) {
+        // This method is temporary. Will be removed when syncing to service-typing branch
+        // Validate qualifiers and create a STNodeList
+        for (int i = 0; i < qualifierList.size();) {
+            STNode qualifier = qualifierList.get(i);
+            if (qualifier.kind != SyntaxKind.ISOLATED_KEYWORD &&
+                    qualifier.kind != SyntaxKind.TRANSACTIONAL_KEYWORD &&
+                    qualifier.kind != SyntaxKind.REMOTE_KEYWORD &&
+                    qualifier.kind != SyntaxKind.PUBLIC_KEYWORD &&
+                    qualifier.kind != SyntaxKind.PRIVATE_KEYWORD) {
```
SupunS, line 1054: Better to extract this logic inside the 'if-condition' to a separate method, something like `isValidMethodQualifier`

SupunS (changes requested):

compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/declarations/FunctionDefinitionTest.java
lochana-chathura: * https://github.com/ballerina-platform/ballerina-lang/blob/70c5342829156da8525211726b1548d2ad018271/compiler/ballerina-parser/src/test/resources/types/func-type/func_type_source_10.bal
* https://github.com/ballerina-platform/ballerina-lang/blob/70c5342829156da8525211726b1548d2ad018271/compiler/ballerina-parser/src/test/resources/types/func-type/func_type_source_12.bal

Consider above two test cases. They both same only different is qualifiers.  Also there are invalid tokens in between qualifiers in lot of places and now some of them match with the switch cases in the parsing methods and parser going haywire. Since those test cases don't  test any granular level recovery we don't need to keep them right?

* https://github.com/ballerina-platform/ballerina-lang/blob/70c5342829156da8525211726b1548d2ad018271/compiler/ballerina-parser/src/test/resources/declarations/isolated-object-methods/isolated_object_method_source_02.bal
* https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-parser/src/test/resources/declarations/func-definition/func_def_source_25.bal

Above two test cases the same and they are broken due to #26460. I think adding one test case when fixing that issue would be sufficient.

SupunS (approved):

PR #26465, opened 2020-10-21
Change the formatter API to return an exception

### LINKED ISSUE #26381: [Formatter] Change the APIs to return an exception
**Description:**
When an exception occurs during formatting, the current implementation logs the exception and returns back an unaltered syntax tree. This behavior should be changed to return an exception so that the module using the formatter core can handle it accordingly.

**Steps to reproduce:**

**Affected Versions:**
Swan Lake Preview 4

IrushiL:
## Purpose

Fixes #26381

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/Formatter.java
+++ b/misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/Formatter.java
@@ -109,7 +109,7 @@ public class Formatter {
         try {
             return syntaxTree.modifyWith(treeModifier.transform(modulePartNode));
         } catch (Exception e) {
-            throw new FormatterException("Error while formatting the source: " + e.getMessage(), e.getCause());
+            throw new FormatterException("Error while formatting: " + e.getMessage(), e.getCause());
         }
     }
 }
--- a/misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/Formatter.java
+++ b/misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/Formatter.java
@@ -20,23 +20,20 @@ import io.ballerina.compiler.syntax.tree.SyntaxTree;
 import io.ballerina.tools.text.LineRange;
 import io.ballerina.tools.text.TextDocument;
 import io.ballerina.tools.text.TextDocuments;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * Class that exposes the formatting APIs.
  */
 public class Formatter {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(Formatter.class);
-
     /**
      * Formats the provided source string and returns back the formatted source string.
      *
      * @param source A Ballerina source in string form
      * @return A modified source string after formatting changes
+     * @throws FormatterException Exception caught while formatting
      */
-    public static String format(String source) {
+    public static String format(String source) throws FormatterException {
         return format(source, new FormattingOptions());
     }
 
@@ -47,8 +44,9 @@ public class Formatter {
      * @param syntaxTree The complete SyntaxTree, of which a part is to be formatted
      * @param range LineRange which specifies the range to be formatted
      * @return The modified SyntaxTree after formatting changes
+     * @throws FormatterException Exception caught while formatting
      */
-    public static SyntaxTree format(SyntaxTree syntaxTree, LineRange range) {
+    public static SyntaxTree format(SyntaxTree syntaxTree, LineRange range) throws FormatterException {
         return format(syntaxTree, range, new FormattingOptions());
     }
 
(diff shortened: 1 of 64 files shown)


misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/Formatter.java
```diff
@@ -94,19 +96,20 @@ public static SyntaxTree format(SyntaxTree syntaxTree, LineRange range, Formatti
...
+    private static SyntaxTree modifyTree(SyntaxTree syntaxTree, FormattingOptions options, LineRange range)
+            throws FormatterException {
         FormattingTreeModifier treeModifier = new FormattingTreeModifier(options, range);
         ModulePartNode modulePartNode = syntaxTree.rootNode();
         try {
             return syntaxTree.modifyWith(treeModifier.transform(modulePartNode));
         } catch (Exception e) {
-            LOGGER.error(String.format("Error while formatting the source: %s", e.getMessage()));
-            return syntaxTree;
+            throw new FormatterException("Error while formatting the source: " + e.getMessage(), e.getCause());
```
nadeeshaan, line 112: Does `e.getMessage()` always return the source name? Otherwise let's change the exception message

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/FormatterTest.java
```diff
@@ -59,8 +59,13 @@ public void test(String source, String sourcePath) throws IOException {
...
         TextDocument textDocument = TextDocuments.from(content);
         SyntaxTree syntaxTree = SyntaxTree.from(textDocument);
-        SyntaxTree newSyntaxTree = Formatter.format(syntaxTree);
-        Assert.assertEquals(newSyntaxTree.toSourceCode(), getSourceText(assertFilePath));
+        try {
+            SyntaxTree newSyntaxTree = Formatter.format(syntaxTree);
+            Assert.assertEquals(newSyntaxTree.toSourceCode(), getSourceText(assertFilePath));
+        } catch (FormatterException e) {
+            // TODO: handle the test cases causing formatting exceptions
+//            Assert.fail(e.getMessage(), e);
```
nadeeshaan, line 67: Are we going to add test cases for these particular negative cases or just ignore them? If ignoring let's remove the commented code

nadeeshaan (changes requested):

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/FormatterTest.java
IrushiL: These test cases should ideally be enabled since these are scenarios with no diagnostic issues (positive test cases).

But at the moment these are failing due to some parser issues which are yet to be fixed. Once they are fixed this can be enabled, and ideally, these test cases shouldn't fail.

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/Formatter.java
IrushiL: Fixed

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/FormatterTest.java
```diff
@@ -74,8 +79,13 @@ public void testParserResources(String sourcePath) throws IOException {
         TextDocument textDocument = TextDocuments.from(content);
         SyntaxTree syntaxTree = SyntaxTree.from(textDocument);
         if (!syntaxTree.hasDiagnostics()) {
-            SyntaxTree newSyntaxTree = Formatter.format(syntaxTree);
-            Assert.assertEquals(newSyntaxTree.toSourceCode(), getSourceText(filePath));
+            try {
+                SyntaxTree newSyntaxTree = Formatter.format(syntaxTree);
+                Assert.assertEquals(newSyntaxTree.toSourceCode(), getSourceText(filePath));
```
SupunS, line 84: Can directly use `content` here

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/FormatterTest.java
SupunS: I think its better to fail the test here and disable any-failing tests, if there are issues from the parser. 
Otherwise, any newly failing test would go un-noticed .

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/FormatterTest.java
IrushiL: Done

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/FormatterTest.java
IrushiL: Done

SupunS (approved):

PR #26454, opened 2020-10-21
Set annotation values readonly, if only attach points contain dual attach point without source

chiranSachintha:
Fixes #26408 

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

Issue #23691, opened 2020-06-03
Update param and return types for clone, cloneReadOnly


irshadnilam:
**Description:**
Please refer https://github.com/ballerina-platform/ballerina-spec/issues/531#issuecomment-635739590

The clone, and cloneReadOnly should be changed to 
```ballerina
type Cloneable readonly|xml|Cloneable[]|map<Cloneable>|table<map<Cloneable>>;

@typeParam
type CloneableType Cloneable;

public function clone(CloneableType v) returns CloneableType = external;

public function cloneReadOnly(CloneableType v) returns CloneableType & readonly = external;
```
This depends on https://github.com/ballerina-platform/ballerina-lang/issues/23690

KRVPerera:
`Cloneable` type is added and `clone` is also updated. `cloneReadOnly` is not yet updated because this requires creating new intersection from `TypeParamAnalyzer`. 
Since intersection with `readonly` creates a new type with effective types we cannot rely on bound types. 
The discussion lead to the idea of we must use consistent types to create a new type using their bound types first. Then use this type and take the intersection with readonly.

MaryamZi:
Reopening this since https://github.com/ballerina-platform/ballerina-lang/pull/27909 only temporarily changed the signature of `cloneReadOnly` to 

```ballerina
public isolated function cloneReadOnly(CloneableType  v) returns CloneableType = external;
```

The return type needs to be changed to `CloneableType & readonly`.

PR #26476, opened 2020-10-21
Revert "Set annotation values readonly, if only attach points contain dual attach point without source"

chiranSachintha:
Reverts ballerina-platform/ballerina-lang#26454


MaryamZi (approved):
Merging this since the original change is causing runtime issues when the annotation has an object value - https://github.com/ballerina-platform/ballerina-lang/issues/26478. 

We can introduce this change when fixing https://github.com/ballerina-platform/ballerina-lang/issues/23691.

PR #26496, opened 2020-10-22
Model finite types as unions of singleton typedescs

pubudu91:
## Purpose
This PR models the public representation of finite types as unions of singleton type descs. 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/TypesFactory.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/TypesFactory.java
@@ -27,6 +27,7 @@ import io.ballerina.compiler.api.impl.types.BallerinaNilTypeDescriptor;
 import io.ballerina.compiler.api.impl.types.BallerinaObjectTypeDescriptor;
 import io.ballerina.compiler.api.impl.types.BallerinaRecordTypeDescriptor;
 import io.ballerina.compiler.api.impl.types.BallerinaSimpleTypeDescriptor;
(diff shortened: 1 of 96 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/TypesFactory.java
```diff
@@ -107,6 +112,16 @@ public static BallerinaTypeDescriptor getTypeDescriptor(BType bType, boolean raw
                 return new BallerinaTypeDescTypeDescriptor(moduleID, (BTypedescType) bType);
             case NIL:
                 return new BallerinaNilTypeDescriptor(moduleID, (BNilType) bType);
+            case FINITE:
```
SupunS, line 115: Don't we need to update the `getTypeDescKind` method?

pubudu91: For the singleton type? The kind() method is implemented in the abstract type desc class. And the type desc kind gets passed to it when creating an instance of the typedesc.

nadeeshaan (commented):
Tested the fix locally, and the original issue is resolved

SupunS (approved):

PR #26520, opened 2020-10-22
Upgrade runtime version

kalaiyarasiganeshalingam:
## Purpose
$Subject

Fixes [#20](https://github.com/ballerina-platform/module-ballerina-runtime/issues/20)

Need to fix [#21](https://github.com/ballerina-platform/module-ballerina-runtime/issues/21)

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/util/BLangConstants.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/util/BLangConstants.java
@@ -99,7 +99,7 @@ public class BLangConstants {
 
     public static final Module BALLERINA_BUILTIN_PKG_ID = new Module(BALLERINA_BUILTIN_PKG_PREFIX, "builtin");
     public static final Module BALLERINA_RUNTIME_PKG_ID = new Module(BALLERINA_BUILTIN_PKG_PREFIX,
-                                                                     "runtime", "0.5.0");
+                                                                     "runtime", "0.5.1");
     public static final Module BALLERINA_LANG_ERROR_PKG_ID = new Module(BALLERINA_BUILTIN_PKG_PREFIX,
                                                                         "lang.error", "1.0.0");
     public static final Module BALLERINA_AUTH_PKG_ID = new Module(BALLERINA_BUILTIN_PKG_PREFIX, "auth", "1.0.0");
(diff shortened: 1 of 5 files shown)


Maninda (approved):

stdlib/runtime-api/src/test/java/org/ballerinalang/stdlib/runtime/InvocationContextTest.java
```diff
@@ -40,13 +40,13 @@ public void setup() {
         compileResult = BCompileUtil.compile("test-src/invocation-context.bal");
     }
 
-    @Test(description = "Test case for accessing invocationId from invocation context")
+    @Test(enabled = false, description = "Test case for accessing invocationId from invocation context")
```
MadhukaHarith92, line 43: Any reason to disable the tests.

kalaiyarasiganeshalingam: ballerina-lang runtime version 0.5.0 and this also uses the same method. So, these tests are failing. But these tests are already moved to the runtime repo other than [this](https://github.com/ballerina-platform/ballerina-lang/pull/26520/files#diff-cb14e55430b1f8380deca06ad5c6c7701cf16e150d6118bfba3fa440985d7fc5R193). We need to move this missed test also. We have found this issue if this test had moved already. We are maintaining this repo for other dependent modules and not used to pack. I think it is no big issue. @daneshk  Can I proceed with this?

daneshk: We are not maintaining this code. Eventually, this will remove from ballerina-lang. So it's ok to disable

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/util/BLangConstants.java
```diff
@@ -99,7 +99,7 @@
 
     public static final Module BALLERINA_BUILTIN_PKG_ID = new Module(BALLERINA_BUILTIN_PKG_PREFIX, "builtin");
     public static final Module BALLERINA_RUNTIME_PKG_ID = new Module(BALLERINA_BUILTIN_PKG_PREFIX,
-                                                                     "runtime", "0.5.0");
+                                                                     "runtime", "0.5.1");
```
daneshk, line 102: What is the usage of this constant? . can't we remove this

kalaiyarasiganeshalingam: Now, we can't remove this as [getCallStack()](https://github.com/ballerina-platform/module-ballerina-runtime/blob/master/runtime-native/src/main/java/org/ballerinalang/stdlib/runtime/nativeimpl/GetCallStack.java#L35) is using this. So we need to implement this method properly.

daneshk (approved):

Issue #26521, opened 2020-10-22
`Type Conversion` BBE  fails with typecast error


chiranSachintha:
**Description:**
Type conversion BBE throws ` {ballerina}TypeCastError {"message":"incompatible types: 'map<anydata|readonly>' cannot be cast to 'lang.error:Detail'"} ` when running with currrent master. 

**Steps to reproduce:**
Run the example in https://ballerina.io/swan-lake/learn/by-example/type-conversion.html with current master.

Issue #24079, opened 2020-06-13
Map as record and record as map allowed at compile-time but not allowed at runtime


MaryamZi:
**Description:**
$title, seems like the "is-type" checks allowed between maps and records have only been introduced at compile-time.

**Steps to reproduce:**
**Map to Record**
```ballerina
import ballerina/io;

type Foo record {|
    int i?;
    int...;
|};

public function main() {
    map<int> mp = {
        i: 1,
        j: 2
    };
    Foo f = mp; // allowed

    any a = mp;

    // io:println(mp is Foo); // Compilation error - expression will always evaluate to 'true'

    io:println(a is Foo); // false!
}
```

**Record to Map**
```ballerina
import ballerina/io;

type Foo record {|
    int i?;
    int...;
|};

public function main() {
    Foo f = {
        i: 1,
        "j": 2
    };
    map<int> mp = f; // allowed

    any a = f;

    // io:println(f is map<int>); // Compilation error - expression will always evaluate to 'true'

    io:println(a is map<int>); // false!
}
```

**Affected Versions:**
Current master.

pubudu91:
Related issue for the compile time checking #17202. I've enabled the above at compile time in PR #22292, not sure why the runtime checks weren't added with that though. :sweat_smile:

MaryamZi:
Record to map assignability seems to be working on the current master. #26533 will fix map to record assignability.

dulajdilshan:
Getting `BLangRecordLiteral cannot be cast to class BLangVariableReference at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.visit(TypeChecker.java:2463)`

for 

```ballerina
type Alpha record {|
    int a;
|};

public function main() {
    Alpha rec = {a: 3;
    map<int> mm = {..rec}};
}
```

Will this also be solved with this ?

MaryamZi:
> ```ballerina
> type Alpha record {|
>     int a;
> |};
> 
> public function main() {
>     Alpha rec = {a: 3;
>     map<int> mm = {..rec}};
> }
> ```

No, this is a compile-time issue due to the syntax errors. Works when those are fixed.

```ballerina
type Alpha record {|
    int a;
|};

public function main() {
    Alpha rec = {a: 3};
    map<int> mm = {...rec};
}
```

Can you please create a separate issue for this?

dulajdilshan:
Created #26544

PR #26533, opened 2020-10-22
Allow using a compatible map as a record at runtime

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/24079
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/26521

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/TypeChecker.java
@@ -1048,20 +1048,26 @@ public class TypeChecker {
     }
 
     private static boolean checkIsRecordType(Type sourceType, BRecordType targetType, List<TypePair> unresolvedTypes) {
-        if (sourceType.getTag() != TypeTags.RECORD_TYPE_TAG) {
-            return false;
+        switch (sourceType.getTag()) {
+            case TypeTags.RECORD_TYPE_TAG:
+                return checkIsRecordType((BRecordType) sourceType, targetType, unresolvedTypes);
+            case TypeTags.MAP_TAG:
+                return checkIsRecordType((BMapType) sourceType, targetType, unresolvedTypes);
         }
+        return false;
+    }
 
+    private static boolean checkIsRecordType(BRecordType sourceRecordType, BRecordType targetType,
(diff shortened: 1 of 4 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/TypeChecker.java
```diff
@@ -1156,17 +1196,24 @@ private static boolean checkRecordBelongsToAnydataType(MapValue sourceVal, BReco
 
     private static boolean checkIsRecordType(Object sourceVal, Type sourceType, BRecordType targetType,
                                              List<TypePair> unresolvedTypes) {
-        if (sourceType.getTag() != TypeTags.RECORD_TYPE_TAG) {
-            return false;
+        switch (sourceType.getTag()) {
+            case TypeTags.RECORD_TYPE_TAG:
+                return checkIsRecordType((MapValue) sourceVal, (BRecordType) sourceType, targetType, unresolvedTypes);
+            case TypeTags.MAP_TAG:
+                return checkIsRecordType((BMapType) sourceType, targetType, unresolvedTypes);
```
pubudu91, line 1203: Is it enough to simply check if the two types are compatible? What about the behaviour if the record type contains a `never` field? :thinking:

MaryamZi: Since we check the source map's constraint against the field type in https://github.com/ballerina-platform/ballerina-lang/pull/26533/files#diff-19c04b50f4303ca57559920287e2ca9ef20fd7ddaad5b7c75d425c340ec5421dR1151, that shouldn't be an issue, right?

hasithaa (approved):

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/TypeChecker.java
pubudu91: Had a chat with Maryam and got this clarified.

Issue #25457, opened 2020-08-27
Refactor the usage globalProps in strands


manuranga:
We have `globalProps` property bag in stands, review and design a new implement if needed.

manuranga:
Had a meeting with @shafreenAnfar and @chamil321 , following is a summary ,
1) even though `globalProps` may not needed for correctness, it's needed for performance.
2) it's assumed `globalProps` is copied form parent stand to child strand, which is not implemented, so the use case is already broken for workers/start.

manuranga:
@nadundesilva it would be good if you can look at the performance of #26579

PR #26579, opened 2020-10-25
Move Strand properties access to Environment api

manuranga:
## Purpose

properties are access is moved to Environment api, now there is no need to use strand internal api to use properties. 

Properties are now inherited by child strands

Part of #25457

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
@@ -19,9 +19,6 @@ package io.ballerina.runtime.api;
 import io.ballerina.runtime.api.async.Callback;
 import io.ballerina.runtime.api.async.StrandMetadata;
 import io.ballerina.runtime.api.values.BObject;
-import io.ballerina.runtime.observability.ObservabilityConstants;
-import io.ballerina.runtime.observability.ObserveUtils;
-import io.ballerina.runtime.observability.ObserverContext;
 import io.ballerina.runtime.scheduling.Scheduler;
 import io.ballerina.runtime.scheduling.Strand;
 
@@ -86,15 +83,10 @@ public class Runtime {
      */
     public void invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
                                   Callback callback, Map<String, Object> properties, Object... args) {
-        Function<Object[], Object> func = objects -> {
-            Strand strand = (Strand) objects[0];
-            if (ObserveUtils.isObservabilityEnabled() && properties != null &&
-                    properties.containsKey(ObservabilityConstants.KEY_OBSERVER_CONTEXT)) {
-                strand.observerContext =
-                        (ObserverContext) properties.remove(ObservabilityConstants.KEY_OBSERVER_CONTEXT);
-            }
-            return object.call(strand, methodName, args);
-        };
+        if (object == null) {
+            throw new NullPointerException();
+        }
+        Function<?, ?> func = o -> object.call((Strand) (((Object[]) o)[0]), methodName, args);
         scheduler.schedule(new Object[1], func, null, callback, properties, PredefinedTypes.TYPE_NULL, strandName,
                            metadata);
     }
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Scheduler.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Scheduler.java
@@ -455,9 +455,6 @@ public class Scheduler {
     }
 
     private FutureValue createFuture(Strand parent, Callback callback, Type constraint, Strand newStrand) {
-        if (parent != null) {
-            newStrand.observerContext = parent.observerContext;
-        }
         FutureValue future = new FutureValue(newStrand, callback, constraint);
         future.strand.frames = new Object[100];
         return future;
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Strand.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Strand.java
@@ -22,7 +22,6 @@ import io.ballerina.runtime.api.PredefinedTypes;
 import io.ballerina.runtime.api.StringUtils;
 import io.ballerina.runtime.api.async.StrandMetadata;
 import io.ballerina.runtime.api.values.BError;
-import io.ballerina.runtime.observability.ObserverContext;
 import io.ballerina.runtime.transactions.TransactionLocalContext;
 import io.ballerina.runtime.values.ChannelDetails;
 import io.ballerina.runtime.values.ErrorValue;
@@ -71,7 +70,6 @@ public class Strand {
     public boolean blockedOnExtern;
     public Set<ChannelDetails> channelDetails;
     public Set<SchedulerItem> dependants;
-    public ObserverContext observerContext;
     public boolean cancel;
 
     SchedulerItem schedulerItem;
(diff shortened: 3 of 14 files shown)


manuranga:
Thanks @nadundesilva  for running below tests

Case | Throughput | Average Latency| Min Latency| Max Latency| Median Latency| 90th pct Latency| 95th pct Latency| 99th pct Latency
-- | -- | -- | -- | -- | -- | -- | -- | --
master with obs |  10806.13 |  9.21 | 0 | 101 | 5.00 | 13.00 | 48.00 | 53.00
master with obs |  10727.83 |  9.28 | 0 | 91 | 5.00 | 17.00 | 47.00 | 54.00
master with obs |  10739.38 |  9.27 | 0 | 114 | 5.00 | 14.00 | 48.00 | 53.00
master with obs |  10959.91 |  9.08 | 0 | 92 | 5.00 | 16.00 | 47.00 | 52.00
pr with obs |  10529.69 |  9.45 | 0 | 109 | 5.00 | 28.90 | 49.00 | 54.00
pr with obs |  10526.49 |  9.45 | 0 | 99 | 5.00 | 14.00 | 48.00 | 55.00
pr with obs |  10566.36 |  9.42 | 0 | 102 | 5.00 | 16.00 | 49.00 | 55.00
pr with obs |  10834.27 |  9.18 | 0 | 102 | 5.00 | 12.00 | 50.00 | 55.00
master without obs |  14959.65 |  6.64 | 0 | 86 | 4.00 | 9.00 | 37.00 | 41.00
master without obs |  14764.17 |  6.73 | 0 | 88 | 4.00 | 9.00 | 37.00 | 43.00
pr without obs |  14729.95 |  6.74 | 0 | 97 | 4.00 | 9.00 | 36.00 | 42.00
pr without obs |  14981.67 |  6.63 | 0 | 93 | 4.00 | 8.00 | 36.00 | 42.00

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Strand.java
```diff
@@ -71,15 +70,14 @@
...
     public boolean cancel;
 
     SchedulerItem schedulerItem;
     List<WaitContext> waitingContexts;
     WaitContext waitContext;
     ItemGroup strandGroup;
 
-    private Map<String, Object> globalProps;
+    public Map<String, Object> globalProps;
     public TransactionLocalContext currentTrxContext;
```
warunalakshitha, line 81: Can we move the transaction context to strand locals similar to observation context

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
```diff
@@ -86,15 +83,10 @@ public Object invokeMethodAsync(BObject object, String methodName, String strand
...
-            Strand strand = (Strand) objects[0];
-            if (ObserveUtils.isObservabilityEnabled() && properties != null &&
-                    properties.containsKey(ObservabilityConstants.KEY_OBSERVER_CONTEXT)) {
-                strand.observerContext =
-                        (ObserverContext) properties.remove(ObservabilityConstants.KEY_OBSERVER_CONTEXT);
-            }
-            return object.call(strand, methodName, args);
-        };
+        if (object == null) {
+            throw new NullPointerException();
```
riyafa, line 87: This is not necessary. Is it? Because in the next line it will automatically throw Nullpointer and throwing this exception explicitly is not recommended AFAIK

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Scheduler.java
```diff
@@ -456,7 +457,8 @@ public FutureValue createFuture(Strand parent, Callback callback, Map<String, Ob
 
     private FutureValue createFuture(Strand parent, Callback callback, Type constraint, Strand newStrand) {
         if (parent != null) {
-            newStrand.observerContext = parent.observerContext;
+            //TODO: improve by using a copy on write map
```
riyafa, line 460: Shall we also create an issue for the Todo and link here or fix the Todo before the PR is merged?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Strand.java
```diff
@@ -71,15 +70,14 @@
...
-    public ObserverContext observerContext;
     public boolean cancel;
 
     SchedulerItem schedulerItem;
     List<WaitContext> waitingContexts;
     WaitContext waitContext;
     ItemGroup strandGroup;
 
-    private Map<String, Object> globalProps;
+    public Map<String, Object> globalProps;
```
riyafa, line 80: Shall we try not to make public variables? We can use getters for future variables. I know this class already uses a lot of public variables, but we can try to reduce them in the future

Kishanthan, line 80: Why did we make this public? I thought we are getting rid of it?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
manuranga: below `object.call` is in a lambda, therefor error gets thrown in a different thread. so the original stack trace is lost. 
Explicit null pointer exception is ok for parameter checking. eg: http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/Objects.java#l201

riyafa: Reading on this, came across this: https://stackoverflow.com/questions/3322638/is-it-okay-to-throw-nullpointerexception-programmatically 
It seems to be a point of argument 😀

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Strand.java
manuranga: We will hide this from others using java 9 modules. But better to keep things public here, as long as they are only accessed by us (runtime module) for performance reasons. what do you think?

manuranga: We will hide this from others using java 9 modules. but needs to be public here because I need to access it from Environment.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Strand.java
manuranga: yes, we need to do that. But will be a separate pr because it's tightly coupled to transaction langlib.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Scheduler.java
manuranga: +1 will create

manuranga: #26710

manuranga, line 460:
```suggestion
            //TODO: improve by using a copy on write map #26710
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/scheduling/Strand.java
manuranga: will fix.

manuranga: fill fix

riyafa (approved):

PR #26581, opened 2020-10-26
Fixed parameter types of Syntax API factory methods

### LINKED ISSUE #26378: Factory method to create `SeparatedNodeList` is not exposed via the `NodeFactory` API
**Description:**
I'm currently working on the project to create a list of Ballerina Syntax API calls for given source code and I noticed that there currently isn't a way to generate a `SeparatedNodeList` from the `NodeFactory` API. Since the constructors are not public, directly creating the required object is also not possible. Is there any alternative way to create the object?

**Steps to reproduce:**
Following code results in a `FUNCTION_SIGNATURE` node with empty `SeperatedNodeList` as parameters.
So, there is no way to recreate the syntax tree.
```
public function main() { }
```

### LINKED ISSUE #26415: NodeFactory.createTemplateExpressionNode() content parameter type cannot hold all possible node types
**Description:**
The current implemented `NodeFactory.createTemplateExpressionNode()` accepts only a  `NodeList<InterpolationNode>` for the content parameter, but it is sometimes necessary to pass `Token`s into the node list as well. So the parameter type should be fixed as  `NodeList<Node>`.

**Steps to reproduce:**
- Attempt to recreate the string ``string `Hello ${name}!!!` `` via the Syntax API. With the current method calls, this doesn't seem to be possible.
- Nodes related to this string is as follows,

```
TemplateExpressionNode
├── Token(BACKTICK_TOKEN)
├── NodeList*
│   ├── LiteralValueToken("Hello ")
│   ├── InterpolationNode
│   │    ├── Token(INTERPOLATION_START_TOKEN)
│   │    ├── SimpleNameReferenceNode
│   │    │    └─ IdentifierToken("name")
│   │    └── Token(CLOSE_BRACE_TOKEN)
│   └── LiteralValueToken("!!!")
└── Token(BACKTICK_TOKEN)

```

- However, `createTemplateExpressionNode`'s `NodeList` parameter only accepts `NodeList<InterpolationNode>`.
But, since the literal tokens surrounding the interpolation node ("Hello " and "!!!" in this case) cannot be cast into interpolation nodes, that factory call cannot be used.

### LINKED ISSUE #26440: NodeFactory.createSeparateNodeList accepts the wrong type of parameter
**Description:**
`SeparatedNodeList` constructor takes a non-terminal node with an odd number of buckets.
And the constructor expects that every odd position of the buckets contain a separator.

```java
SeparatedNodeList(NonTerminalNode node) {
   // The list size is always an odd number
   // E.g., If the list size is 7, then there are 4 nodes and 3 separators
   super(node, (node.bucketCount() + 1) / 2);
   this.separatorSize = size != 0 ? size - 1 : 0;

   // TODO check whether the given node is a node list with separator
}
```

So, when creating the separator node list, we have to pass in the separators(eg: comma tokens) as well.

However, in the `NodeFactory.createSeparatedNodeList()` it directly passes the node list to the constructor after creating the node list node. 

```java
public static <T extends Node> NodeList<T> createSeparatedNodeList(T... nodes) {
   STNode[] internalNodes = new STNode[nodes.length];
   for (int index = 0; index < nodes.length; index++) {
      T node = nodes[index];
      Objects.requireNonNull(node, "node should not be null");
      internalNodes[index] = node.internalNode();
   }
   return new SeparatedNodeList<>(STNodeFactory.createNodeList(internalNodes).createUnlinkedFacade());
}
```

However, since the `SeparatedNodeList` constructor expects separator tokens between every 2 nodes, the callee of this function will have to pass in those as well. But in the for loop, each node has to be cast into the same type `T`. (Which token can't be cast into) So this method call cannot be used.

**Steps to reproduce:**

- Attempt to recreate `function main(){ f(a, b); }` via Syntax API. With the current method calls, this doesn't seem to be possible. Nodes related to this string is as follows,

```
FunctionDefinitionNode
   | Token[FUNCTION_KEYWORD]
   | IdentifierToken("main")
   | FunctionSignatureNode
   |   | Token[OPEN_PAREN_TOKEN]
   |   | Token[CLOSE_PAREN_TOKEN]
   | FunctionBodyBlockNode
   |   | Token[OPEN_BRACE_TOKEN]
   |   | NodeList*
   |   |   | ExpressionStatementNode[CALL_STATEMENT]
   |   |   |   | FunctionCallExpressionNode
   |   |   |   |   | SimpleNameReferenceNode
   |   |   |   |   |   | IdentifierToken("f")
   |   |   |   |   | Token[OPEN_PAREN_TOKEN]
   |   |   |   |   | SeparatedNodeList              <- THIS
   |   |   |   |   |   | PositionalArgumentNode
   |   |   |   |   |   |   | SimpleNameReferenceNode
   |   |   |   |   |   |   |   | IdentifierToken("a")
   |   |   |   |   |   | Token[COMMA_TOKEN]         <- TOKEN
   |   |   |   |   |   | PositionalArgumentNode
   |   |   |   |   |   |   |  SimpleNameReferenceNode
   |   |   |   |   |   |   |   |  IdentifierToken("b")
   |   |   |   |   | Token[CLOSE_PAREN_TOKEN]
   |   |   |   | Token[SEMICOLON_TOKEN]
   |   | Token[CLOSE_BRACE_TOKEN]
```

- However, `createFunctionCallExpressionNode`'s `SeparatedNodeList` parameter only accepts `SeparatedNodeList<FunctionArgumentNode>`.
    
- But, since the commas inside the parameter list has to be included in the node list but cannot be cast into function argument node, that factory call cannot be used.

kdsuneraavinash:
## Purpose
Some of the Syntax API factory calls accept the wrong type of parameter causing some syntax trees unable to create.

Fixes #26440 
Fixes #26415
Fixes #26378

## Approach
Modified the `treegen` configuration JSON file and regenerated the code. And the `SeparatedNodeList` creation method was also fixed to accept the correct type of node.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


sameerajayasoma:
LGTM @SupunS Please have a look as well.

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -2877,7 +2878,7 @@
                 },
                 {
                     "name": "intermediateClauses",
-                    "type": "IntermediateClauseNode",
+                    "type": "ClauseNode",
```
SupunS, line 2881: Any idea why do we need this change?

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -2173,7 +2174,7 @@
                 },
                 {
                     "name": "content",
-                    "type": "TemplateMemberNode",
+                    "type": "Node",
```
SupunS, line 2177: and this one too?

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.ballerinalang.formatter.core;
 
+
```
SupunS, line 20: Shall we remove this extra newline?

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
```diff
@@ -1391,7 +1391,7 @@ public TypeCastExpressionNode transform(TypeCastExpressionNode typeCastExpressio
     @Override
     public TypeCastParamNode transform(TypeCastParamNode typeCastParamNode) {
         NodeList<AnnotationNode> annotations = formatNodeList(typeCastParamNode.annotations(), 0, 1, 1, 0);
-        Node type = formatNode(typeCastParamNode.type(), env.trailingWS, env.trailingNL);
+        Node type = formatNode(typeCastParamNode.type().orElse(null), env.trailingWS, env.trailingNL);
```
SupunS, line 1394: Formatting properties of the annotations changes depending on the existence of the `type`.
Can you create an issue to track this?

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/AbstractNodeFactory.java
```diff
@@ -189,17 +189,17 @@ public static Minutiae createEndOfLineMinutiae(String text) {
                         .collect(Collectors.toList())).createUnlinkedFacade());
     }
 
-    public static <T extends Node> NodeList<T> createSeparatedNodeList(T... nodes) {
+    public static <T extends Node> SeparatedNodeList<T> createSeparatedNodeList(Node... nodes) {
```
SupunS, line 192: Do we need to change the parameter type from the generic type `T` to `Node`?

SupunS (changes requested):

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
SupunS: ok, just saw the issue reported. +1 for this change. 
Shall we also remove the `TemplateMemberNode` from the syntax tree, since there are no more usages of it?

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
kdsuneraavinash: <https://ballerina.io/swan-lake/learn/by-example/query-expression.html>

In this code, specifically in the `QueryPipelineNode`, `order by student.firstName descending` part is a `OrderByClauseNode` which is extended from `ClauseNode` directly. So, it cannot exists inside the `NodeList<IntermediateClauseNode`.

Here is a small code segment with this problem.

```ballerina
public function main() {
    Report[] reportList = from var student in studentList
       where student.gpa >= 2.0
       order by student.firstName descending
       limit 2
       select {};
}
```

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
kdsuneraavinash: Actually previously it was possible for the `typeCastParamNode.type()` to be null. However, it was handled inside the `formatNode` method accordingly.

```java
private <T extends Node> T formatNode(T node, int trailingWS, int trailingNL) {
        if (node == null || node.isMissing()) {
            return node;
        }
       // ...
}
```

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
SupunS: I see. 
Issue seems to be in the `OrderByClauseNode`. We should make the `base` type of `OrderByClauseNode` to `IntermediateClauseNode`.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/AbstractNodeFactory.java
kdsuneraavinash: Yes, I think so.
The `SeparatedNodeList` contains separator nodes as well.
These are not of the Type T.

See issue #26440

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
kdsuneraavinash: OK

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
kdsuneraavinash: Should I remove that in this PR as well?

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
SupunS: Trailing whitespaces/newlines etc. of a given node depends on the node to be followed. So here, if the `typeCastParamNode.type()` is not present, then the formatting the preceding node (metadata) must be adjusted accordingly.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/AbstractNodeFactory.java
SupunS: I have added a comment in 26440

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
SupunS: +1

kdsuneraavinash: `XMLItemNode` extends from `TemplateMemberNode`. So should I change `XMLItemNode` to extend from the superclass of `TemplateMemberNode` instead? (which is `Node`)

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/AbstractNodeFactory.java
kdsuneraavinash: What if `SeparatedNodeList` was changed to accept the `Node` type, but does a runtime check to see if every even index is of Type `T`?

SupunS: oh okay, I think I misunderstood the problem.

SupunS (approved):

PR #26607, opened 2020-10-26
Add an API to get the type of an expression

pubudu91:
## Purpose
This PR adds support for looking up the type of an AST node, given the location of the node.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/SemanticModel.java
```diff
@@ -57,6 +58,15 @@
      */
     List<Symbol> moduleLevelSymbols();
 
+    /**
+     * Retrieves the type of the expression in the specified text range. If it's not a valid expression, returns null.
```
nadeeshaan, line 62: Shall we mention if the range is not valid, we return an empty optional? Instead of returns null

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/NodeFinder.java
```diff
@@ -0,0 +1,1484 @@
...
+    private void lookupNode(BLangNode node) {
+        if (node == null) {
+            return;
+        }
+
+        if (PositionUtil.withinRange(this.range, node.pos)) {
+            node.accept(this);
+        }
+
+        if (this.enclosingNode == null && !node.internal) {
```
nadeeshaan, line 231: Do we need this null check since we do the null check within the `setEnclosingNode` method ?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/NodeFinder.java
```diff
@@ -0,0 +1,1484 @@
...
+    }
+
+    @Override
+    public void visit(BLangRollback rollbackNode) {
+        lookupNode(rollbackNode.expr);
+    }
+
+    @Override
+    public void visit(BLangLiteral literalExpr) {
+        // Do nothing
```
nadeeshaan, line 567: Shall we introduce a base visitor with empty implementation for all the visitor methods? We have a couple of visitors now and for some node we have to duplicate this empty visitor methods.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/NodeFinder.java
```diff
@@ -0,0 +1,1484 @@
...
+    public void visit(BLangQueryExpr queryExpr) {
+        if (setEnclosingNode(queryExpr, queryExpr.pos)) {
+            return;
+        }
+
+        lookupNodes(queryExpr.queryClauseList);
+    }
+
+    @Override
+    public void visit(BLangTableMultiKeyExpr tableMultiKeyExpr) {
```
nadeeshaan, line 947: Calling the super methods throws assertion error

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/NodeFinder.java
```diff
@@ -0,0 +1,1484 @@
...
+                || setEnclosingNode(syncSendExpr.expr, syncSendExpr.expr.pos)) {
+            return;
+        }
+
+        setEnclosingNode(syncSendExpr, syncSendExpr.workerIdentifier.pos);
+    }
+
+    @Override
+    public void visit(BLangWaitForAllExpr waitForAllExpr) {
+        super.visit(waitForAllExpr);
```
nadeeshaan, line 1403: Calls the super here

nadeeshaan (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/NodeFinder.java
pubudu91: Modified the logic related to this a bit. The `setEnclosingNode()` is usually used for nodes with names (e.g., variables, constants), to see if the provided range is the range of the name of it. With this latest cleanup, this is now required since now, for the typical case, this does what `setEnclosingNode()` method does.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/NodeFinder.java
pubudu91: Was intentional, since these are not yet supported. Throwing an unsupported expression exception would be more appropriate i guess?

nadeeshaan: This will be OK during the development until we finish the implementation. Otherwise these might cause broken usages even for valid cases. We had such issues with earlier LS visitor implementations

SupunS (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/NodeFinder.java
nadeeshaan: @SupunS yeah, we override the base expectation

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BaseVisitor.java
```diff
@@ -0,0 +1,988 @@
...
+import org.wso2.ballerinalang.compiler.tree.types.BLangUserDefinedType;
+import org.wso2.ballerinalang.compiler.tree.types.BLangValueType;
+
+/**
+ * A default visitor with empty implementations for the visit methods, to be used as the base visitor in visitors
+ * implemented.
+ *
+ * @since 2.0.0
+ */
+class BaseVisitor extends BLangNodeVisitor {
```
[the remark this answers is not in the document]
pubudu91: Yeah, can be.

Was planning on doing the refactoring to `SymbolFinder` with a separate PR to avoid unnecessary noise in this PR.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/PositionUtil.java
```diff
@@ -42,4 +43,22 @@ static boolean withinBlock(LinePosition cursorPos, Diagnostic.DiagnosticPosition
...
+        int specifiedStartLine = specifiedRange.startLine().line();
+        int specifiedEndLine = specifiedRange.endLine().line();
+        int specifiedStartOffset = specifiedRange.startLine().offset();
+        int specifiedEndOffset = specifiedRange.endLine().offset();
+
+        return (startLine < specifiedStartLine && endLine > specifiedEndLine)
+                || (startLine < specifiedStartLine && endLine == specifiedEndLine && endOffset > specifiedEndOffset)
+                || (startLine == specifiedStartLine && startOffset < specifiedStartOffset && endLine > specifiedEndLine)
+                || (startLine == endLine && startLine == specifiedStartLine && specifiedStartLine == specifiedEndLine
+                && startOffset <= specifiedStartOffset && endOffset >= specifiedEndOffset);
```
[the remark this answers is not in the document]
pubudu91: This doesn't seem to work in this case (tests failing). Here we're checking if a particular node encompasses the specified range. The above mentioned checks don't seem to cover all the cases for that. Will see if the checks can be simplified.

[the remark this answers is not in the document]
pubudu91: Can write it as follows, if it feels more readable
```java
if (specifiedStartLine < startLine || specifiedEndLine > endLine) {
    return false;
}

if (specifiedStartLine > startLine && specifiedEndLine < endLine) {
    return true;
}

if (specifiedStartLine == startLine && specifiedEndLine == endLine) {
    return specifiedStartOffset >= startOffset && specifiedEndOffset <= endOffset;
}

if (specifiedStartLine == startLine) {
    return specifiedStartOffset >= startOffset;
}

return specifiedEndOffset <= endOffset;
```

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -131,6 +133,22 @@ public BallerinaSemanticModel(BLangPackage bLangPackage, CompilerContext context
...
+    public Optional<BallerinaTypeDescriptor> getType(String fileName, LineRange range) {
+        BLangCompilationUnit compilationUnit = getCompilationUnit(fileName);
+        NodeFinder nodeFinder = new NodeFinder();
+        BLangNode node = nodeFinder.lookup(compilationUnit, range);
+
+        if (node == null) {
+            return Optional.empty();
+        }
+
+        return Optional.ofNullable(TypesFactory.getTypeDescriptor(node.type));
```
rasika, line 149: Why we are returning NULLs in`TypesFactory.getTypeDescriptor(node.type)` ? shouldn't it return an optional type-desc ?

pubudu91: We have kept the factories to return the relevant instance or `null`. Returning an `Optional` value straight from the factory may not be desirable for some of the uses of it. e.g., populating a list of types (in the union for instance). If we return `Optional` from the factory, then the list would be a `null` allowing list: `List<Optional<BallerinaTypeDescriptor>>`

rasika: When populating a list of types still we call `TypesFactory.getTypeDescriptor()` per each type right? I am not sure how usage of `null` avoid a `null` allowing list. Idea of using `Optional<T>` for a function-return is to explicitly notify that the value can be present or not. Otherwise, API user deliberately need to go and check the function implementation. Since this is a public API I still believe `Optional<BallerinaTypeDescriptor>` is the way to go. WDYT @nadeeshaan @SupunS  ?

nadeeshaan: I think, what @rasika says is valid if we are exposing this particular API to the public. IIRC we had to expose it to public due to the project API limitation. I would say, as long as we hide it as a public API, this is ok. WDYT rasika?

SupunS (approved):

nadeeshaan (approved):

rasika (approved):

PR #26643, opened 2020-10-28
Fix Completion Issues for snippet and pre-declared lang libs

### LINKED ISSUE #26243: Avoid adding imports for predeclared langlib modules
**Description:**
Swan Lake preview versions now predeclare prefixes for langlib modules that are keywords.

This happens for the following modules:

```
boolean
decimal
error
float
future
int
map
object
stream
string
table
typedesc
xml
```

So the user doesn't have to add the import for such a module.

For example, the following code is now valid.

```ballerina
public function main() {
    var x = 'int:fromString("1");
}
```

However, in VSCode, when we currently accept a suggestion for such a module, the import is added automatically. Shall we not do this? If the user wants, they can explicitly add the import.

**Steps to reproduce:**

For example, accepting

int_import

results in

```ballerina
import ballerina/lang.'int;
public function main() {
    'int<cursor>
}
```

**Affected Versions:**
slp4

### LINKED ISSUE #26314: Add context aware completion support for else and else-if clauses
**Description:**
$title for the completions for the contexts as follows,

function testFunction() {
      int testInt = 123;
      if (testInt == 12) {
      } <cursor>
}

nadeeshaan:
Remove the pre declared langlib suggestions with auto import

Fixes #26314
Fixes #26243

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
```diff
@@ -141,6 +141,10 @@
     public static final Path LS_STDLIB_CACHE_DIR = TEMP_DIR.resolve("ls_stdlib_cache").resolve(SDK_VERSION);
 
     public static final Path LS_CONNECTOR_CACHE_DIR = TEMP_DIR.resolve("ls_connector_cache").resolve(SDK_VERSION);
+    
+    public static final List<String> PRE_DECLARED_LANG_LIBS = Arrays.asList("lang.boolean", "lang.decimal",
+            "lang.error", "lang.float", "lang.future", "lang.int", "lang.map", "lang.object", "lang.stream",
+            "lang.string", "lang.table", "lang.typedesc", "lang.xml", "lang.annotations");
```
rasika, line 147: Any specific reason to maintain a set of pre-declared lang libs?

nadeeshaan: This is due to #26243 where the compiler now pre-declare langLibs with reserved keywords. So we do not need go add the auto import and also we do not need the suggestions at the import declaration.

rasika (approved):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
rasika: Thanks, eventually we need to incorporate this into auto-imports in code-actions.

PR #26674, opened 2020-10-29
Fix lang lib content suggestions

### LINKED ISSUE #25983: Suggestions support for predeclared modules.
With https://github.com/ballerina-platform/ballerina-lang/issues/25463, all lang lib module which ends with a keyboard now Predeclare as module prefixes. 

This has to support in code suggestions in Language Server implementation. 
vscode-1600689355121

nadeeshaan:
Fixes #25983 and add few more test cases.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/AssignmentStatementNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/AssignmentStatementNodeContext.java
@@ -17,6 +17,7 @@ package org.ballerinalang.langserver.completions.providers.context;
 
 import io.ballerina.compiler.api.symbols.Symbol;
 import io.ballerina.compiler.api.symbols.SymbolKind;
+import io.ballerina.compiler.api.symbols.VariableSymbol;
 import io.ballerina.compiler.api.types.ObjectTypeDescriptor;
 import io.ballerina.compiler.syntax.tree.AssignmentStatementNode;
 import io.ballerina.compiler.syntax.tree.Node;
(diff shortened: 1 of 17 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/AssignmentStatementNodeContext.java
```diff
@@ -67,7 +68,8 @@ public AssignmentStatementNodeContext() {
             (2) [module:]TypeName c = module:a<cursor>
              */
             QualifiedNameReferenceNode qNameRef = (QualifiedNameReferenceNode) node.expression();
-            Predicate<Symbol> filter = symbol -> symbol.kind() == SymbolKind.VARIABLE;
+            Predicate<Symbol> filter = symbol -> symbol instanceof VariableSymbol
```
rasika, line 71: Can't we use `SymbolKind.VARIABLE` instead?

rasika (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/AssignmentStatementNodeContext.java
nadeeshaan: No, if we use the VARIABLE symbol kind, then it will not capture the constants as per the current model. In order to avoid having two check with constant and variable kinds, added the instance of check

rasika: If the intention is to address both constants and variables I would still prefer two checks; since `==` comparison is far efficient  compared to `instanceOf` and for the better readability of allowing constants and variables.

rasika (approved):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/AssignmentStatementNodeContext.java
nadeeshaan: Yes, agreed. This will be addressed along with the newly proposed model

PR #26693, opened 2020-11-01
Add a multi-language extension support

nadeeshaan:
1. Completion Extension
2. Formatting Extension
3. Diagnostic Extension

Also, with this PR we migrated the `completion` feature to delegate to the particular completion extension(Currently Ballerina only)

## Samples
Writing a feature extension
- Implement the `CompletionExtension`, `FormattingExtension` or `DiagnosticsExtension`
- Implements the `execute` method
- `execute` method is the entry point
- LangExtensionDelegator is the publicly exposed delegate API.

Ex: Implementing a language extension for completion feature
```
@JavaSPIService("org.ballerinalang.langserver.commons.LanguageExtension")
public class BallerinaCompletionExtension implements CompletionExtension {
    @Override
    public boolean validate(CompletionParams inputParams) {
        // language validation based on the input parameters
    }

    @Override
    public Either<List<CompletionItem>, CompletionList> execute(CompletionParams inputParams, LSContext context) {
        // resolving logic
    }
}
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
@@ -49,7 +49,6 @@ import org.ballerinalang.langserver.compiler.config.LSClientConfig;
 import org.ballerinalang.langserver.compiler.config.LSClientConfigHolder;
 import org.ballerinalang.langserver.compiler.exception.CompilationFailedException;
 import org.ballerinalang.langserver.completions.exceptions.CompletionContextNotSupportedException;
-import org.ballerinalang.langserver.completions.util.CompletionUtil;
 import org.ballerinalang.langserver.diagnostic.DiagnosticsHelper;
 import org.ballerinalang.langserver.exception.UserErrorException;
 import org.ballerinalang.langserver.extensions.ballerina.semantichighlighter.HighlightingFailedException;
@@ -156,14 +155,13 @@ class BallerinaTextDocumentService implements TextDocumentService {
 
     @Override
     public CompletableFuture<Either<List<CompletionItem>, CompletionList>> completion(CompletionParams position) {
-        final List<CompletionItem> completions = new ArrayList<>();
         return CompletableFuture.supplyAsync(() -> {
             String fileUri = position.getTextDocument().getUri();
             Optional<Path> completionPath = CommonUtil.getPathFromURI(fileUri);
 
             // Note: If the source is a cached stdlib source or path does not exist, then return early and ignore
             if (completionPath.isEmpty() || CommonUtil.isCachedExternalSource(fileUri)) {
-                return Either.forLeft(completions);
+                return Either.forLeft(new ArrayList<>());
             }
 
             Path compilationPath = getUntitledFilePath(completionPath.toString()).orElse(completionPath.get());
@@ -174,13 +172,11 @@ class BallerinaTextDocumentService implements TextDocumentService {
                     .withCommonParams(position, fileUri, docManager)
                     .withCompletionParams(clientCapabilities.getTextDocCapabilities().getCompletion())
                     .build();
-
             try {
                 LSModuleCompiler.getBLangPackage(context, docManager, false, false);
                 // Fill the current file imports
                 context.put(DocumentServiceKeys.CURRENT_DOC_IMPORTS_KEY, CommonUtil.getCurrentFileImports(context));
-                CompletionUtil.resolveSymbols(context);
-                completions.addAll(CompletionUtil.getCompletionItems(context));
+                return LangExtensionDelegator.instance().completion(position, context);
             } catch (CompletionContextNotSupportedException e) {
                 // Ignore the exception
             } catch (Throwable e) {
@@ -190,7 +186,8 @@ class BallerinaTextDocumentService implements TextDocumentService {
             } finally {
                 lock.ifPresent(Lock::unlock);
             }
-            return Either.forLeft(completions);
+
+            return Either.forLeft(new ArrayList<>());
         });
     }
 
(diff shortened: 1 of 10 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/LangExtensionDelegator.java
```diff
@@ -0,0 +1,110 @@
...
+     * Get the completions.
+     *
+     * @param params completion parameters
+     * @return {@link Either} completion results
+     */
+    public Either<List<CompletionItem>, CompletionList> completion(CompletionParams params, LSContext context) {
+        return completionExtensions.stream().filter(ext -> ext.validate(params))
+                .map(ext -> ext.execute(params, context))
+                .findFirst()
+                .orElse(Either.forRight(new CompletionList()));
```
rasika, line 78: Specific reason to limit to one? We can merge the results from multiple filtered extensions. eg. there can be two completion extensions with the same uri pattern.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/LangExtensionDelegator.java
```diff
@@ -0,0 +1,110 @@
...
+     * Get the Diagnostics.
+     *
+     * @param uri document URI
+     * @return {@link PublishDiagnosticsParams} diagnostic params calculated
+     */
+    public PublishDiagnosticsParams diagnostics(String uri, LSContext context) {
+        return diagExtensions.stream().filter(ext -> ext.validate(uri))
+                .map(ext -> ext.execute(uri, context))
+                .findFirst()
+                .orElse(new PublishDiagnosticsParams());
```
rasika, line 104: Specific reason to limit to one? We can merge the results from multiple filtered extensions. eg. there can be two completion extensions with the same uri pattern.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/LangExtensionDelegator.java
```diff
@@ -0,0 +1,110 @@
...
+     * Get the formatting.
+     *
+     * @param params formatting parameters
+     * @return {@link List} of text edits
+     */
+    public List<? extends TextEdit> formatting(DocumentFormattingParams params, LSContext context) {
+        return formatExtensions.stream().filter(ext -> ext.validate(params))
+                .map(ext -> ext.execute(params, context))
+                .findFirst()
+                .orElse(new ArrayList<>());
```
rasika, line 91: Specific reason to limit to one? We can merge the results from multiple filtered extensions. eg. there can be two completion extensions with the same uri pattern.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
```diff
@@ -156,14 +155,13 @@ void setClientCapabilities(LSClientCapabilities clientCapabilities) {
...
     public CompletableFuture<Either<List<CompletionItem>, CompletionList>> completion(CompletionParams position) {
-        final List<CompletionItem> completions = new ArrayList<>();
         return CompletableFuture.supplyAsync(() -> {
             String fileUri = position.getTextDocument().getUri();
             Optional<Path> completionPath = CommonUtil.getPathFromURI(fileUri);
 
             // Note: If the source is a cached stdlib source or path does not exist, then return early and ignore
             if (completionPath.isEmpty() || CommonUtil.isCachedExternalSource(fileUri)) {
-                return Either.forLeft(completions);
+                return Either.forLeft(new ArrayList<>());
```
rasika, line 164: Instead of returning `new ArrayList<>()` for erroneous cases we can return an unmodifiable per-declared empty list with `Collections..emptyList()`. This will limit ad-hoc new list objects creations.

rasika (changes requested):
This is great @nadeeshaan. I feel we can move all lang-server extensions to these interfaces for consistency. So any extension-dev find can find all available extensions in a single place.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/LangExtensionDelegator.java
nadeeshaan: Yes, I added this one to force only one at this time because we need to decide whether we collect from all the extensions and if we do so, do we need a specific enable disable mechanism (Because as an example, for completions and signatures, ordering of the items matters). We need to decide that. Otherwise I also feel we should allow all and +1 for proceed.
@rasika Shall we allow all in the first phase and shall we decide later about the prioritisation and other similar capabilities?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
nadeeshaan: +1

nadeeshaan:
> This is great @nadeeshaan. I feel we can move all lang-server extensions to these interfaces for consistency. So any extension-dev find can find all available extensions in a single place.

Yes, we should migrate all the features to depend on the delegation. As the first phase I migrated the completion engine and it  worked seamlessly with the current test suite and manual testing.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/LangExtensionDelegator.java
nadeeshaan: Added the aggregated result support

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/LangExtensionDelegator.java
nadeeshaan: fixed

rasika (approved):

SupunS (approved):

PR #26718, opened 2020-11-03
Update signature help parameter detection by range

### LINKED ISSUE #25971: Signature help's selected parameter is invalid
**Description:**
Consider the following source capture and the selected parameter in the signature help is invalid.

Screenshot 2020-09-21 at 15 09 33

**Affected Versions:**
slp4 at least

nadeeshaan:
Fixes #25971

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFactory.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFactory.java
@@ -213,7 +213,7 @@ public class SymbolFactory {
         if (symbol == null) {
             return null;
         }
-        String name = symbol.getName().getValue();
+        String name = symbol.getName().getValue().isBlank() ? null : symbol.getName().getValue();
         TypeSymbol typeDescriptor = TypesFactory.getTypeDescriptor(symbol.getType());
         List<Qualifier> qualifiers = new ArrayList<>();
         if ((symbol.flags & Flags.PUBLIC) == Flags.PUBLIC) {
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
@@ -100,7 +100,8 @@ public class SignatureHelpUtil {
         labelBuilder.append("(");
         // Join the function parameters to generate the function's signature
         List<ParameterInfoModel> parameterInfoModels = signatureInfoModel.getParameterInfoModels();
-        for (ParameterInfoModel paramModel : parameterInfoModels) {
+        for (int i = 0; i < parameterInfoModels.size(); i++) {
+            ParameterInfoModel paramModel = parameterInfoModels.get(i);
             int labelOffset = labelBuilder.toString().length();
             labelBuilder.append(paramModel.parameter.getType());
             ParameterInformation paramInfo = new ParameterInformation();
@@ -112,7 +113,7 @@ public class SignatureHelpUtil {
                 paramEnd += (paramModel.parameter.getName().get() + " ").length();
                 labelBuilder.append(" ").append(paramModel.parameter.getName().get());
             }
-            if (parameterInfoModels.indexOf(paramModel) < parameterInfoModels.size() - 1) {
+            if (i < parameterInfoModels.size() - 1) {
                 labelBuilder.append(", ");
             }
             paramInfo.setLabel(Tuple.two(paramStart, paramEnd));
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
@@ -54,6 +54,7 @@ import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.SignatureInformation;
 import org.eclipse.lsp4j.SignatureInformationCapabilities;
 import org.eclipse.lsp4j.jsonrpc.messages.Either;
+import org.eclipse.lsp4j.jsonrpc.messages.Tuple;
 
 import java.nio.file.Path;
 import java.util.ArrayList;
@@ -61,7 +62,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.stream.Collectors;
 
 import static io.ballerina.compiler.api.symbols.SymbolKind.FUNCTION;
 import static io.ballerina.compiler.api.symbols.SymbolKind.METHOD;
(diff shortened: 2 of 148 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFactory.java
```diff
@@ -213,7 +213,7 @@ public static ParameterSymbol createBallerinaParameter(BVarSymbol symbol, Parame
         if (symbol == null) {
             return null;
         }
-        String name = symbol.getName().getValue();
+        String name = symbol.getName().getValue().isBlank() ? null : symbol.getName().getValue();
```
rasika, line 216: `.isBlank()` will capture cases such as "      "(empty spaces) as well. Do we really need this or just would `isEmpty()` is suffcient?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
```diff
@@ -92,21 +92,35 @@ private SignatureHelpUtil() {
...
-            parameterInformationList.add(getParameterInformation(parameterInfoModel));
-
-            return parameterInfoModel.toString();
-        }).collect(Collectors.joining(", "));
+        List<ParameterInfoModel> parameterInfoModels = signatureInfoModel.getParameterInfoModels();
+        for (ParameterInfoModel paramModel : parameterInfoModels) {
+            int labelOffset = labelBuilder.toString().length();
+            labelBuilder.append(paramModel.parameter.getType());
+            ParameterInformation paramInfo = new ParameterInformation();
+            paramInfo.setDocumentation(getParameterDocumentation(paramModel));
```
rasika, line 107: we can move adding documentation into `getParameterInfoModels()` method

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
```diff
@@ -92,21 +92,35 @@ private SignatureHelpUtil() {
...
+            int paramEnd = labelOffset + paramModel.parameter.getType().length();
+            if (paramModel.parameter.getName().isPresent()) {
+                paramStart = paramEnd + 1;
+                paramEnd += (paramModel.parameter.getName().get() + " ").length();
+                labelBuilder.append(" ").append(paramModel.parameter.getName().get());
+            }
+            if (parameterInfoModels.indexOf(paramModel) < parameterInfoModels.size() - 1) {
+                labelBuilder.append(", ");
+            }
+            paramInfo.setLabel(Tuple.two(paramStart, paramEnd));
```
rasika, line 118: can we move this logic related to setting `Tuple.two(paramStart, paramEnd)` into `getParameterInfoModels()`

rasika (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFactory.java
nadeeshaan: Yes, name should never be empty, and since the `isBlank` internally checks for both isEmpty and space chars, therefore for the safe hand added this.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
nadeeshaan: ParameterInformation model generation is separated from the meta model. SignatureInfoModel is a combination of meta info.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
nadeeshaan: Same as the above and it will be better to keep the final model generation in once place. WDYT?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFactory.java
rasika: My argument is will you ever get a symbol name with empty spaces? `isBlank` internally traverses through the whole char-array where `isEmpty` only checks char array `length == 0`.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
rasika: Still, paramStart, paramEnd calculation can be a part of `ParameterInfoModel` of the `SignatureInfoModel`. WDYT?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
```diff
@@ -92,21 +92,35 @@ private SignatureHelpUtil() {
...
+            paramInfo.setDocumentation(getParameterDocumentation(paramModel));
+            int paramStart = labelOffset;
+            int paramEnd = labelOffset + paramModel.parameter.getType().length();
+            if (paramModel.parameter.getName().isPresent()) {
+                paramStart = paramEnd + 1;
+                paramEnd += (paramModel.parameter.getName().get() + " ").length();
+                labelBuilder.append(" ").append(paramModel.parameter.getName().get());
+            }
+            if (parameterInfoModels.indexOf(paramModel) < parameterInfoModels.size() - 1) {
+                labelBuilder.append(", ");
```
rasika, line 116: we can get rid of `indexOf` call for each element if we move the parent loop into an indexed for loop. Or will be more readable if we can move parameters generation into a `new StringJoiner(", ")`.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
nadeeshaan: IMHO I think we don't need to, this is because the paramStart and paramEnd is calculated relative to the label. That's the reason for generating both the label and the params in a single run without calculating the parameters separately as per the earlier logic.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
nadeeshaan: converted to an indexed loop

rasika (approved):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
```diff
@@ -100,7 +100,8 @@ private SignatureHelpUtil() {
         labelBuilder.append("(");
         // Join the function parameters to generate the function's signature
         List<ParameterInfoModel> parameterInfoModels = signatureInfoModel.getParameterInfoModels();
-        for (ParameterInfoModel paramModel : parameterInfoModels) {
+        for (int i = 0; i < parameterInfoModels.size(); i++) {
+            ParameterInfoModel paramModel = parameterInfoModels.get(i);
             int labelOffset = labelBuilder.toString().length();
```
rasika, line 105: You can just call `labelBuilder.length()` otherwise whole purpose of StringBuilder is lost since we convert toString per each cycle.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
rasika: Yeah agree, seems paramStart requires other labelling positions as input.

Issue #26054, opened 2020-09-24
Fix and test BalEnv receiving extern instance and object functions


manuranga:
**Description:**
Fix and test BalEnv receiving extern instance and object functions

**Steps to reproduce:**
Add `BalEnv` as the first param to a object extern call, you will get an error like below
```
error: .::object_test.bal:53:0: {ballerina/java}METHOD_SIGNATURE_DOES_NOT_MATCH 'Incompatible param type for method 'getRandomInt' in class 'org.ballerinalang.nativeimpl.jvm.tests.InstanceMethods': Java type 'org.ballerinalang.jvm.api.BalEnv' will not be matched to ballerina type 'Person''
```

**Affected Versions:**
SLP4

Issue #24443, opened 2020-06-24
API to get current module info at runtime


warunalakshitha:
**Description:**
API should provide module org name, version and module name.

PR #26725, opened 2020-11-03
Support API to get current module from runtime

warunalakshitha:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/24443
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/26054

This will support an api to get current module for given interop function environment.
Also fixes the ballerina environment passing issue with instance methods(receivers) and ballerina object methods.
Added test cases for each scenario.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmPackageGen.java
```diff
@@ -256,20 +260,41 @@ private static void generateStaticInitializer(ClassWriter cw, String className,
...
+    }
+
+    private static void setCurrentModuleField(ClassWriter cw, MethodVisitor mv, BIRPackage module,
+                                              String moduleInitClass) {
+        FieldVisitor fv = cw.visitField(ACC_PUBLIC + ACC_STATIC, CURRENT_MODULE_VAR_NAME,
+                                        String.format("L%s;", MODULE), null, null);
+        fv.visitEnd();
+        mv.visitTypeInsn(Opcodes.NEW, MODULE);
+        mv.visitInsn(Opcodes.DUP);
+        mv.visitLdcInsn(module.org.value);
```
HindujaB, line 289:
```suggestion
        mv.visitLdcInsn(IdentifierUtils.decodeIdentifier(module.org.value));
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmPackageGen.java
```diff
@@ -256,20 +260,41 @@ private static void generateStaticInitializer(ClassWriter cw, String className,
...
+        mv.visitLdcInsn(module.name.value);
```
HindujaB, line 290:
```suggestion
        mv.visitLdcInsn(IdentifierUtils.decodeIdentifier(module.name.value));
```

HindujaB: We will be having the encoded names for module/org names at this point. So we might need to decode them for module initializing

warunalakshitha: Do we need to decode version?

HindujaB: No. version is not encoded.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmPackageGen.java
```diff
@@ -452,7 +477,7 @@ CompiledJarFile generate(BIRNode.BIRPackage module, InteropValidator interopVali
 
         // generate object/record value classes
         JvmValueGen valueGen = new JvmValueGen(module, this, methodGen, lambdaGen);
-        valueGen.generateValueClasses(jarEntries);
+        valueGen.generateValueClasses(jarEntries, moduleInitClass);
```
riyafa, line 480: I have been wondering if there could be a better way rather than having to pass this parameter to many methods

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/observability/ObserveUtils.java
```diff
@@ -257,7 +257,7 @@ public static void logMessageToActiveSpan(String logLevel, Supplier<String> logM
         if (!tracingEnabled) {
             return;
         }
-        Environment balEnv = new Environment(Scheduler.getStrand());
+        Environment balEnv = new Environment(Scheduler.getStrand(), null);
```
riyafa, line 260: I think it is a bad idea to pass a null to a constructor. We can have two constructors in the `Environment` class to handle this.

tests/jballerina-unit-test/src/test/resources/test-src/javainterop/basic/instance_method_tests.bal
```diff
@@ -191,3 +200,15 @@ function newByte(int val) returns handle = @java:Constructor {
    'class: "java.lang.Byte"
 } external;
 
+const ASSERTION_ERROR_REASON = "AssertionError";
+
+function assertEquality(any|error expected, any|error actual) {
```
riyafa, line 205: Since this is unit tests we can use the `test:assertEquals` function from the `ballerina/test` package. Then we need not define this function

tests/jballerina-unit-test/src/test/java/org/ballerinalang/nativeimpl/jvm/tests/InstanceMethods.java
```diff
@@ -172,4 +176,10 @@ public int getInt(ObjectValue objectValue, int x) {
     public int getRandomInt() {
         return 123;
     }
+
+    public BString getCurrentModule(Environment env, long b) {
+        Module callerModule = env.getCurrentModule();
+        return StringUtils.fromString(callerModule.getOrg() + "#" + callerModule.getName() + "#" +
+                                              callerModule.getVersion() + "#" + b);
+    }
```
riyafa, line 184: Shall we also have test where self is also passed as a parameter for an instance method? I mean self and Environment both.

riyafa: Shall we also see what happens if the methods with both without self and with self and Environment have the same name(overloaded)?

warunalakshitha: Self is not part of parameter list in instance method. As an example getCureentModule(receiver) will generated to receiver.getCurrentModule() . passing self as a parameter test already added. I also added another test case for overload methods.

riyafa (approved):

PR #26748, opened 2020-11-04
Fix RunExecutable task to reflect actual status of program execution

### LINKED ISSUE #26084: Ballerina program panics with exit code 0
**Description:**
When a Ballerina program panics, the exit code returns `0`. This should be `1` since the program did not exit normally. 

This is a serious issue since we use ballerina (1.2.8) for the standard library release pipeline processes, where the release is failed in the middle, the ballerina program panics, but the pipeline is marked as passed.

**Steps to reproduce:**
Try the following code:

```ballerina
public function main() {
    error err = error("Sample", message = "Sample Error");
    panic err;
}
```

```shell
$ ballerina run sample.bal
$ echo $?
```
This prints `0`.

**Affected Versions:**

jBallerina: 1.2.8
Ballerina Tool: 0.8.8

** Related Issues:**
#20967, #21624

Dilhasha:
## Purpose
Fix exit code to reflect actual status of task

Fixes #26084 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/RunExecutableTask.java
+++ b/cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/RunExecutableTask.java
@@ -22,6 +22,7 @@ import org.ballerinalang.packerina.buildcontext.BuildContext;
 import org.ballerinalang.packerina.buildcontext.BuildContextField;
 import org.ballerinalang.packerina.buildcontext.sourcecontext.SingleFileContext;
 import org.ballerinalang.packerina.buildcontext.sourcecontext.SingleModuleContext;
+import org.ballerinalang.tool.BLangPanicException;
 import org.ballerinalang.tool.util.BFileUtil;
 import org.wso2.ballerinalang.compiler.tree.BLangPackage;
 import org.wso2.ballerinalang.compiler.util.ProjectDirConstants;
@@ -132,7 +133,7 @@ public class RunExecutableTask implements Task {
             process.waitFor();
             int exitValue = process.exitValue();
             if (exitValue != 0) {
-                Runtime.getRuntime().exit(exitValue);
+                throw new BLangPanicException("Execution exited with the exit code " + exitValue);
             }
         } catch (IOException | InterruptedException e) {
             throw createLauncherException("Error occurred while running the executable ", e.getCause());
--- a/cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/RunExecutableTask.java
+++ b/cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/RunExecutableTask.java
@@ -130,6 +130,10 @@ public class RunExecutableTask implements Task {
             ProcessBuilder pb = new ProcessBuilder(commands).inheritIO();
             Process process = pb.start();
             process.waitFor();
+            int exitValue = process.exitValue();
+            if (exitValue != 0) {
+                Runtime.getRuntime().exit(exitValue);
+            }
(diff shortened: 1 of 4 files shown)


cli/ballerina-packerina/src/main/java/org/ballerinalang/packerina/task/RunExecutableTask.java
```diff
@@ -130,6 +130,10 @@ private void runGeneratedExecutable(BLangPackage executableModule, BuildContext
             ProcessBuilder pb = new ProcessBuilder(commands).inheritIO();
             Process process = pb.start();
             process.waitFor();
+            int exitValue = process.exitValue();
+            if (exitValue != 0) {
+                Runtime.getRuntime().exit(exitValue);
+            }
         } catch (IOException | InterruptedException e) {
```
warunalakshitha, line 137: Don't we need to exit if get exception while running the command?

Dilhasha: The thrown exceptions are handled and exits are covered at the main class at https://github.com/ballerina-platform/ballerina-lang/blob/ballerina-1.2.x/cli/ballerina-tool/src/main/java/org/ballerinalang/tool/Main.java

But when there is no exception thrown as in the case when the ballerina program panics with an error, the exit code does not reflect the actual status. That's why I am suggesting this fix.

Dilhasha: I realised, we can throw an exception when exit code is not zero and let the main class handle it to keep it consistent. I'll fix it.

cli/ballerina-tool/src/main/java/org/ballerinalang/tool/Main.java
```diff
@@ -68,6 +68,9 @@ public static void main(String... args) {
         } catch (BLauncherException e) {
             LauncherUtils.printLauncherException(e, errStream);
             Runtime.getRuntime().exit(1);
+        } catch (BLangPanicException e) {
+            errStream.println(e.getMessage());
```
warunalakshitha, line 72: I think the e.getMessage will not have any useful info as always print "Execution exited with the exit code <some non 0 value>". Runtime already print the actual cause when panic. Therefore we do not need to print the error.  Also this will miss the exit code when we print the echo $? which will be always 1. So we can save the exitValue in PanicException and set in here without printing any thing.

catch (BLangPanicException e) {
Runtime.getRuntime().exit(((BlangPanicException).getExitValue())
}

cli/ballerina-tool/src/main/java/org/ballerinalang/tool/BLangPanicException.java
```diff
@@ -0,0 +1,34 @@
...
+ *   under the License.
+ */
+package org.ballerinalang.tool;
+
+/**
+ * Thrown to indicate a non zero exit code from Ballerina executable.
+ *
+ * @since 1.2.10
+ */
+public class BLangPanicException extends RuntimeException {
```
warunalakshitha, line 25: Since actuall runtime gets panic shall we rename RuntimePanicException

warunalakshitha:
Also send a PR to master as well.

warunalakshitha (approved):

hevayo (approved):

PR #26783, opened 2020-11-05
Enable ballerina maven resolver

keizer619:


misc/maven-resolver/build.gradle
```diff
@@ -41,21 +43,7 @@ dependencies {
     implementation 'org.apache.maven.resolver:maven-resolver-transport-file:1.4.2'
     implementation 'org.apache.maven.resolver:maven-resolver-transport-http:1.4.2'
     implementation 'me.tongfei:progressbar'
-    testImplementation 'org.apache.maven:maven-model-builder:3.6.3' 
     testCompile 'org.testng:testng'
 }
 
-test {
```
keizer619, line 48: Still there is a packaging issue in maven resolver with jdk11 where we cannot run tests within module. For this i have created a separate issue to enable 
https://github.com/ballerina-platform/ballerina-lang/issues/26786

IrushiL (approved):

misc/maven-resolver/src/main/java/module-info.java
```diff
@@ -1,13 +0,0 @@
-module io.ballerina.maven.resolver {
```
IrushiL, line 1: Removing the `module-info.java` here is intentional right?

keizer619: Yes Apache Maven Resolver has issues with JPMS
https://stackoverflow.com/questions/60915727/java-11-unnamed-module-issue-when-using-maven-model

PR #26796, opened 2020-11-06
Fix BCompileUtil to fail only on compilation errors

NipunaRanasinghe:
## Purpose
$subject.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/ballerina-test-utils/src/main/java/org/ballerinalang/test/BCompileUtil.java
+++ b/tests/ballerina-test-utils/src/main/java/org/ballerinalang/test/BCompileUtil.java
@@ -59,9 +59,9 @@ public class BCompileUtil {
         Package currentPackage = project.currentPackage();
         PackageCompilation packageCompilation = currentPackage.getCompilation();
         JBallerinaBackend jBallerinaBackend = JBallerinaBackend.from(packageCompilation, JdkVersion.JAVA_11);
-        long errorCount = jBallerinaBackend.diagnostics().stream().filter(diagnostic ->
-                diagnostic.diagnosticInfo().severity() == DiagnosticSeverity.ERROR).count();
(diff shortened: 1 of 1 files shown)


tests/ballerina-test-utils/src/main/java/org/ballerinalang/test/BCompileUtil.java
```diff
@@ -58,7 +59,9 @@ public static CompileResult compile(String sourceFilePath) {
         Package currentPackage = project.currentPackage();
         PackageCompilation packageCompilation = currentPackage.getCompilation();
         JBallerinaBackend jBallerinaBackend = JBallerinaBackend.from(packageCompilation, JdkVersion.JAVA_11);
-        if (jBallerinaBackend.hasDiagnostics()) {
+        long errorCount = jBallerinaBackend.diagnostics().stream().filter(diagnostic ->
```
hemikak, line 62: We can use `findAny` function instead finding the count since there is no usage of the count - https://www.baeldung.com/java-stream-findfirst-vs-findany

NipunaRanasinghe: @hemikak WDYT of using `anyMatch(predicate)` instead? IMO it'll be more convenient for the above context, if we don't need the exact count.

hemikak: +1

hevayo (approved):

PR #26805, opened 2020-11-06
Implement API for getting the lang lib functions allowed by method call expr

### LINKED ISSUE #25929: Provide a way to lookup built-in methods attached to a type
**Description:**

A possible approach, quoted from one of the offline discussions related to this:

> Need a way to get the built-in methods coming from the lang-lib modules, associated with a type. Can facilitate this by adding a method called ‘builtinMethods()’ to the ‘TypeDescriptor’ model.
> - Any type descriptor class will inherit it.
> - Eg: ObjectTypeDescriptor will have a `methods()` function to get the user-defined methods, and a `builtinMethods()` function to get the methods coming from object-lang-lib and value-lang-lib.

### LINKED ISSUE #26392: Add qualifiers to method declarations and lang lib methods
$subject. Currently it's empty

pubudu91:
## Purpose
This PR adds the lang lib functions which can be called using a method call expression to the relevant types. An API is provided to then retrieve a list of its associated lang lib functions.

Fix #25929
Fix #26392

## Remarks
This is to be merged after PR #26758

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
@@ -61,22 +61,23 @@ public class LangLibrary {
     private LangLibrary(CompilerContext context) {
         context.put(LANG_LIB_KEY, this);
 
-        symbolFactory = SymbolFactory.getInstance(context);
-        SymbolTable symbolTable = SymbolTable.getInstance(context);
-        Map<String, BPackageSymbol> langLibs = new HashMap<>();
-        wrappedLangLibMethods = new HashMap<>();
+        this.symbolFactory = SymbolFactory.getInstance(context);
+        this.wrappedLangLibMethods = new HashMap<>();
+        this.langLibMethods = new HashMap<>();
 
+        SymbolTable symbolTable = SymbolTable.getInstance(context);
         for (Map.Entry<BPackageSymbol, SymbolEnv> entry : symbolTable.pkgEnvMap.entrySet()) {
             BPackageSymbol module = entry.getKey();
             PackageID moduleID = module.pkgID;
 
-            if (Names.BALLERINA_ORG.equals(moduleID.orgName) &&
-                    (moduleID.nameComps.size() >= 2 && Names.LANG.equals(moduleID.nameComps.get(0)))) {
-                langLibs.put(moduleID.nameComps.get(1).value, module);
+            if (isLangLibModule(moduleID)) {
+                if (!LANG_VALUE.equals(moduleID.nameComps.get(1).value)) {
+                    addLangLibMethods(moduleID.nameComps.get(1).value, module, this.langLibMethods);
+                } else {
+                    populateLangValueLibrary(module, this.langLibMethods);
+                }
             }
         }
-
-        langLibMethods = getLangLibMethods(langLibs);
     }
 
     public static LangLibrary getInstance(CompilerContext context) {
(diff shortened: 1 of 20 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
```diff
@@ -0,0 +1,225 @@
...
+                return TABLE.getName();
+            default:
+                return "value";
+        }
+    }
+
+    private static Map<String, Map<String, BInvokableSymbol>> getLangLibMethods(Map<String, BPackageSymbol> langLibs) {
+        Map<String, Map<String, BInvokableSymbol>> langLibMethods = new HashMap<>();
+
+        for (Map.Entry<String, BPackageSymbol> entry : langLibs.entrySet()) {
```
SupunS, line 180: Lang libs are looped twice, here and at L#80. Can we do it in a single loop?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
```diff
@@ -0,0 +1,225 @@
...
+
+    private static void populateLangValueLibrary(Map<String, BPackageSymbol> langLibs,
+                                                 Map<String, Map<String, BInvokableSymbol>> langLibMethods) {
+        BPackageSymbol langValue = langLibs.get(LANG_VALUE);
+        Map<String, BInvokableSymbol> methods = new HashMap<>();
+
+        for (Map.Entry<Name, Scope.ScopeEntry> nameScopeEntry : langValue.scope.entries.entrySet()) {
+            BSymbol symbol = nameScopeEntry.getValue().symbol;
+
+            if (symbol.kind != SymbolKind.FUNCTION || !Symbols.isFlagOn(symbol.flags, Flags.LANG_LIB)) {
```
SupunS, line 216: Can there be symbols in the `lang.value` module without the `LANG_LIB` flag?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
```diff
@@ -0,0 +1,225 @@
...
+                return "value";
+        }
+    }
+
+    private static Map<String, Map<String, BInvokableSymbol>> getLangLibMethods(Map<String, BPackageSymbol> langLibs) {
+        Map<String, Map<String, BInvokableSymbol>> langLibMethods = new HashMap<>();
+
+        for (Map.Entry<String, BPackageSymbol> entry : langLibs.entrySet()) {
+            String key = entry.getKey();
+            BPackageSymbol value = entry.getValue();
```
SupunS, line 182: Can rename:
`key` -> `pkgName`
 `value` to `pkgSymbol`

SupunS (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
```diff
@@ -0,0 +1,225 @@
...
+        SymbolTable symbolTable = SymbolTable.getInstance(context);
+        Map<String, BPackageSymbol> langLibs = new HashMap<>();
+        wrappedLangLibMethods = new HashMap<>();
+
+        for (Map.Entry<BPackageSymbol, SymbolEnv> entry : symbolTable.pkgEnvMap.entrySet()) {
+            BPackageSymbol module = entry.getKey();
+            PackageID moduleID = module.pkgID;
+
+            if (Names.BALLERINA_ORG.equals(moduleID.orgName) &&
+                    (moduleID.nameComps.size() == 2 && Names.LANG.equals(moduleID.nameComps.get(0)))) {
```
nadeeshaan, line 85: will the number of name components be 2 in future as well? What if we check for startsWith("lang.")?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
```diff
@@ -0,0 +1,225 @@
...
+        for (Map.Entry<String, BInvokableSymbol> entry : langLib.entrySet()) {
+            FunctionSymbol method = symbolFactory.createFunctionSymbol(entry.getValue(), entry.getKey());
+            list.add(method);
+        }
+    }
+
+    private String getAssociatedLangLibName(TypeDescKind typeDescKind) {
+        switch (typeDescKind) {
+            case INT:
+            case BYTE:
```
nadeeshaan, line 142: We can simplify the cases and the returns here. What if get the common cases to the bottom where we can directly call getName() over the typeDescKind

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
pubudu91: Ack. I think we can, by populating the lang lib map on-demand, instead of populating the function list at the lang lib initialization.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
pubudu91: There can be I guess. Currently we have added the `LANG_LIB` flag to lang lib functions. I guess we can get rid of the flag check

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
pubudu91: Good point. Don't think there's a restriction on that. Will change

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
pubudu91: Checked this. Forgot that I explicitly checked the lang lib flag to filter out the module init, start, stop functions.

SupunS (approved):

Issue #646, opened 2017-01-06
Fix Hardcoded port in ballerina server mode


hasithaa:
This was added with PR #645

PR #26806, opened 2020-11-06
Fix langlib functions to support readonly

### LINKED ISSUE #26114: Can't use certain langlib methods with immutable values
**Description:**
$title.

**Steps to reproduce:**
e.g., `ballerina/lang.array:filter` - [the type of the original array is set as the type of the newly created array](https://github.com/ballerina-platform/ballerina-lang/blob/v2.0.0-Preview4/langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Filter.java#L57) (i.e., with the immutable bit also) resulting in errors when trying to add to the new array members that match the filter criteria. Ideally we should create a new type with the relevant member type.

```ballerina
import ballerina/lang.'array;

final int[] & readonly arr = [1, 2, 1];

public function main() {
    int[] x = array:filter(arr, val => val == 1);
}
```

ballerina-internal.log
```log
[2020-09-25 23:14:44,054] SEVERE {b7a.log.crash} - {ballerina/lang.array}InvalidUpdate 
error: {ballerina/lang.array}InvalidUpdate {"message":"modification not allowed on readonly value"} 
```

Same happens for maps too.

**Affected Versions:**
slp4

HindujaB:
## Purpose

Fixes #26114 

## Approach

- Modify the `filter` function to clone the element type instead of original `readonly` type

## Remarks

The API can be used to return the expected type regarding to this spec-issue [#646](https://github.com/ballerina-platform/ballerina-spec/issues/646)

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BArray.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BArray.java
@@ -17,6 +17,7 @@
 */
 package io.ballerina.runtime.api.values;
 
+import io.ballerina.runtime.api.types.ArrayType;
 import io.ballerina.runtime.api.types.Type;
 
 /**
@@ -210,4 +211,10 @@ public interface BArray extends BRefValue, BCollection {
     void setLength(long i);
 
     long getLength();
+
+    /**
+     * Check and and set the array type as readonly.
+     * @param type original Array type
+     */
+    BArray checkAndSetReadonly(ArrayType type);
 }
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BArray.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BArray.java
@@ -17,7 +17,6 @@
 */
 package io.ballerina.runtime.api.values;
 
-import io.ballerina.runtime.api.types.ArrayType;
 import io.ballerina.runtime.api.types.Type;
 
 /**
@@ -211,10 +210,4 @@ public interface BArray extends BRefValue, BCollection {
     void setLength(long i);
 
     long getLength();
-
-    /**
-     * Check and and set the array type as readonly.
-     * @param type original Array type
-     */
-    BArray checkAndSetReadonly(ArrayType type);
 }
--- a/langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Filter.java
+++ b/langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Filter.java
@@ -20,7 +20,6 @@ package org.ballerinalang.langlib.array;
 
 import io.ballerina.runtime.api.async.StrandMetadata;
 import io.ballerina.runtime.api.creators.ValueCreator;
-import io.ballerina.runtime.api.types.ArrayType;
 import io.ballerina.runtime.api.values.BArray;
 import io.ballerina.runtime.api.values.BFunctionPointer;
 import io.ballerina.runtime.internal.scheduling.AsyncUtils;
@@ -44,7 +43,7 @@ public class Filter {
                                                                       ARRAY_VERSION, "filter");
 
     public static BArray filter(BArray arr, BFunctionPointer<Object, Boolean> func) {
-        BArray newArr = ValueCreator.createArrayValue((ArrayType) arr.getType());
+        BArray newArr = ValueCreator.createArrayValue(TypeCreator.createArrayType(arr.getElementType()));
         int size = arr.size();
         AtomicInteger newArraySize = new AtomicInteger(-1);
         AtomicInteger index = new AtomicInteger(-1);
--- a/langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Filter.java
+++ b/langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Filter.java
@@ -19,6 +19,7 @@
 package org.ballerinalang.langlib.array;
 
 import io.ballerina.runtime.api.async.StrandMetadata;
+import io.ballerina.runtime.api.creators.TypeCreator;
 import io.ballerina.runtime.api.creators.ValueCreator;
 import io.ballerina.runtime.api.values.BArray;
 import io.ballerina.runtime.api.values.BFunctionPointer;
--- a/langlib/lang.map/src/main/java/org/ballerinalang/langlib/map/Filter.java
+++ b/langlib/lang.map/src/main/java/org/ballerinalang/langlib/map/Filter.java
@@ -22,6 +22,7 @@ import io.ballerina.runtime.api.TypeTags;
 import io.ballerina.runtime.api.async.StrandMetadata;
 import io.ballerina.runtime.api.creators.TypeCreator;
 import io.ballerina.runtime.api.creators.ValueCreator;
+import io.ballerina.runtime.api.types.MapType;
 import io.ballerina.runtime.api.types.RecordType;
 import io.ballerina.runtime.api.types.Type;
 import io.ballerina.runtime.api.values.BFunctionPointer;
@@ -55,7 +56,7 @@ public class Filter {
         Type newConstraint;
         switch (mapType.getTag()) {
             case TypeTags.MAP_TAG:
-                BMapType type = (BMapType) mapType;
+                MapType type = (MapType) mapType;
                 newConstraint = type.getConstrainedType();
                 break;
             case TypeTags.RECORD_TYPE_TAG:
--- a/langlib/lang.map/src/main/java/org/ballerinalang/langlib/map/Filter.java
+++ b/langlib/lang.map/src/main/java/org/ballerinalang/langlib/map/Filter.java
@@ -52,21 +52,19 @@ public class Filter {
 
     public static BMap filter(BMap<?, ?> m, BFunctionPointer<Object, Boolean> func) {
         Type mapType = m.getType();
-        Type newMapType;
-        Type newConstraint;
+        Type constraint;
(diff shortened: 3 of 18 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BArray.java
```diff
@@ -208,4 +209,10 @@
     BArray slice(long startIndex, long endIndex);
 
     void setLength(long i);
+
+    /**
+     * Check and and set the array type as readonly.
+     * @param type original Array type
+     */
+    BArray checkAndSetReadonly(ArrayType type);
```
MaryamZi, line 217: Is this change needed? The inherent type of a value cannot change once the value is created, so IMO, we shouldn't have such a method (which allows changing the type) in the public API.

If your requirement is to create an immutable value, IMO it should be done by allowing the user to pass both the type and the values and internally calling the relevant constructors that accept the same;
- [io.ballerina.runtime.values.ArrayValueImpl#ArrayValueImpl(io.ballerina.runtime.api.types.ArrayType, long, io.ballerina.runtime.values.ListInitialValueEntry[])](https://github.com/ballerina-platform/ballerina-lang/blob/v2.0.0-Preview5/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/values/ArrayValueImpl.java#L187) 
- [io.ballerina.runtime.values.MapValueImpl#MapValueImpl(io.ballerina.runtime.api.types.Type, io.ballerina.runtime.api.values.BMapInitialValueEntry[])](https://github.com/ballerina-platform/ballerina-lang/blob/v2.0.0-Preview5/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/values/MapValueImpl.java#L106)

HindujaB: With the clarification from spec issue #646 we wouldn't need this API.

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Filter.java
```diff
@@ -44,7 +44,7 @@
                                                                       ARRAY_VERSION, "filter");
 
     public static BArray filter(BArray arr, BFunctionPointer<Object, Boolean> func) {
```
warunalakshitha, line 46: Do we have set of methods we need to support read only?. I can see array reverse also we create array value with given array type without using element type. @MaryamZi @HindujaB

HindujaB: The doc does not specifically say that we return a new array (unlike some other functions as filter, map etc.). 

So I found that currently we are not allowing the usage of `reverse` functionality for `readonly` types from front-end. 
https://ballerina.io/swan-lake/learn/api-docs/ballerina/lang.array/functions.html#reverse

MaryamZi: With reverse I think there's actually a spec deviation, we shouldn't be returning a new array. See https://github.com/ballerina-platform/ballerina-lang/issues/26843.

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibArrayTest.java
```diff
@@ -551,4 +551,9 @@ public void testSort9() {
     public void testSort10() {
         BRunUtil.invoke(compileResult, "testSort10");
     }
+
+    @Test
+    public void testReadOnlyArrayFilter() {
+        BRunUtil.invoke(compileResult, "testReadOnlyArrayFilter");
+    }
```
riyafa, line 558: Since we are ding the same thing in all the methods, maybe we can use a `DataProvider` to simplify the code

langlib/langlib-test/src/test/resources/test-src/arraylib_test.bal
```diff
@@ -1114,6 +1114,16 @@ function testAsyncFpArgsWithArrays() returns [int, int[]] {
     return [reduce, filter];
 }
 
+function testReadOnlyArrayFilter() {
+    int[] & readonly numbers = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+    int[] evenNumbers = numbers.filter(val => val % 2 == 0);
+    int count = 0;
+    foreach int number in evenNumbers {
+        assertValueEquality(count * 2, number);
```
riyafa, line 1122: can we use `test:assertEquals`?

langlib/langlib-test/src/test/resources/test-src/maplib_test.bal
```diff
@@ -298,3 +302,16 @@ function testAsyncFpArgsWithMaps() returns [int, map<int>] {
...
+function testReadOnlyMapFilter() {
+    map<int> & readonly marks = {
+        science: 75,
+        maths: 45,
+        english: 85
+    };
+    map<int> passed = marks.filter(val => val > 50);
+    assert(passed.length(), 2);
+    passed.forEach(function(int value) {
+        assertTrue(value > 50);
```
riyafa, line 315: can we use `test:assertTrue`?

langlib/langlib-test/src/test/resources/test-src/tablelib_test.bal
```diff
@@ -683,3 +683,37 @@ function testPutValidDataToKeylessTbl() returns boolean {
...
+    });
+}
+
+const ASSERTION_ERROR_REASON = "AssertionError";
+
+function assertTrue(boolean actual) {
+    assertEquals(true, actual);
+}
+
+function assertEquals(anydata expected, anydata actual) {
```
riyafa, line 709: I think we can use function in `ballerina/test` package instead of defining our own functions in all places

MaryamZi: We usually try to avoid importing non-langlib modules in language tests. Once the compiler test framework (most of the  changes done in https://github.com/ballerina-platform/ballerina-lang/pull/24557) is fully integrated, the plan is to migrate all of these lang tests to that.

IMO, we can go ahead with this approach for the time-being.

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibArrayTest.java
```diff
@@ -517,38 +493,23 @@ public void testSort3() {
...
-    @Test
-    public void testSort9() {
-        BRunUtil.invoke(compileResult, "testSort9");
-    }
-
-    @Test
-    public void testSort10() {
-        BRunUtil.invoke(compileResult, "testSort10");
+    @DataProvider(name = "FunctionList")
+    public Object[][] testFunctions() {
```
HindujaB, line 497: use Object[]

manuranga (dismissed):

langlib/lang.map/src/main/java/org/ballerinalang/langlib/map/Filter.java
```diff
@@ -52,17 +53,19 @@
     public static BMap filter(BMap<?, ?> m, BFunctionPointer<Object, Boolean> func) {
         Type mapType = m.getType();
         Type newMapType;
+        Type newConstraint;
```
MaryamZi, line 56: Can't we directly create `newMapType` for the map and record cases instead of introducing a separate variable for the constraint?

langlib/lang.table/src/main/java/org/ballerinalang/langlib/table/Filter.java
```diff
@@ -51,8 +51,10 @@
                                                                       TABLE_VERSION, "filter");
 
     public static BTable filter(BTable tbl, BFunctionPointer<Object, Boolean> func) {
-        Type newTableType = tbl.getType();
-        BTable newTable = ValueCreator.createTableValue((TableType) newTableType);
+        TableType newTableType = (TableType) tbl.getType();
```
MaryamZi, line 54: The variable name isn't quite correct here, right? It's not the new type.

langlib/langlib-test/src/test/resources/test-src/arraylib_test.bal
```diff
@@ -1114,6 +1114,16 @@ function testAsyncFpArgsWithArrays() returns [int, int[]] {
...
+        count += 1;
+    }
```
MaryamZi, line 1124: Shall we also assert that the new value is not immutable? Please add for the map, table, and record tests too.
```suggestion
    }
    assertFalse(evenNumbers.isReadOnly());
```

langlib/langlib-test/src/test/resources/test-src/tablelib_test.bal
```diff
@@ -683,3 +683,37 @@ function testPutValidDataToKeylessTbl() returns boolean {
...
+      { name: "Hermione", age: 28 },
+      { name: "Ron", age: 11 },
+      { name: "Draco", age: 23 }
+    ];
+    table<Person> children = personTable.filter(function (Person person) returns boolean {
+                                                      return person.age < 18;
+                                                  });
+    assertEquals(children.length(), 2);
+    children.forEach(function(Person person) {
+        assertTrue(person.age < 18);
```
MaryamZi, line 699: Shall we also assert that each `person` value is immutable?

HindujaB: Updated in last commit

manuranga (approved):

Issue #26340, opened 2020-10-14
Error in how an `xml-template-expr` is handled with a sequence


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    xml x1 = xml `<foo>foo</foo><!--bar-->baz`;
    xml x2 = xml `<foo>foo</foo>`;
    io:println(x1);
    io:println(x1);
    io:println(x1 == x2);
}
```

This currently prints
```cmd
<foo>foo</foo>
<foo>foo</foo>
true
```

**Affected Versions:**
slp4

rdhananjaya:
Seems to be an issue with the xml sequence literal. Does not occure when xml sequence is constructed using `xml:concat` function.

rdhananjaya:
Same as https://github.com/ballerina-platform/ballerina-lang/issues/26589

PR #26825, opened 2020-11-09
Update XML iterator  return type

### LINKED ISSUE #24562: XML iterator still returns `xml|string` this should be just `xml`
**Description:**
$subject.

**Steps to reproduce:**

**Affected Versions:**

### LINKED ISSUE #26731: Inconsistencies in xml:concat behavior
**Description:**
$subject.

**Steps to reproduce:**
The below example is failing due to `r.length()` returning 2 instead of 3. 
```
'xml:Element catalog = xml `<CATALOG>
                       <CD>
                           <TITLE>Empire Burlesque</TITLE>
                           <ARTIST>Bob Dylan</ARTIST>
                       </CD>
                       <CD>
                           <TITLE>Hide your heart</TITLE>
                           <ARTIST>Bonnie Tyler</ARTIST>
                       </CD>
                       <CD>
                           <TITLE>Greatest Hits</TITLE>
                           <ARTIST>Dolly Parton</ARTIST>
                       </CD>
                   </CATALOG>`;

xml r = 'xml:concat();
foreach var x in catalog/* {
    if (x is 'xml:Element) {
         r = 'xml:concat(r, x);
    }
}

assert(r.length(), 3);
```

### LINKED ISSUE #27532: xml iterator does not support xml:Text
**Description:**
In the spec it says: ```xml<xml<T>> is the same as xml<T> and that xml<xml:Text> is the same as xml:Text.``` Even though 'xml<xml:Text> ' is supported by the iterator currently it does not support `xml:Text`.

**Steps to reproduce:**
```
'xml:Text txt = xml `bit of text\u2702\u2705`;

int i = 0;
foreach 'xml:Text elem in txt {
     i += 1;
}

```

> ERROR [test5.bal:(22:21,22:33)] incompatible types: expected 'other', found 'xml:Text'
> ERROR [test5.bal:(22:42,22:45)] incompatible types: 'xml:Text' is not an iterable collection
> error: compilation contains errors

### LINKED ISSUE #27533: xml iterator does not return T of xml<T>
**Description:**
$title

**Steps to reproduce:**
```
public function main() {
    xml<'xml:Text> txt = xml `bit of text`;
    int i = 0;
    foreach 'xml:Text item in txt {
        i += 1;
    }

record {| 'xml:Text value; |}? nextResult = txt.iterator().next();
}

```

### LINKED ISSUE #27715: xml iterator does not iterate over  xml<xml:Comment> and xml<xml:ProcessingInstruction>
**Description:**
$title

**Steps to reproduce:**

```
function concatIntString (int i, string v) {
     output = output + i.toString() + ":" + v + " ";
}

function testXmlCommentSequenceIteration() {
     xml<'xml:Comment> comment1 = xml `<!--I am a comment-->`;

     int i = 0;
     string output = "";
     foreach (xml|string) elem in comment1 {
         concatIntString(i, elem.toString());
         i += 1;
     }

     record {| (xml|string) value; |}? nextComment1 = comment1.iterator().next();

     panic error("out for loop: "+ output + "\n out record: "+ nextComment1.toString());
}

function testXmlPISequenceIteration() {
     xml<'xml:ProcessingInstruction> pi1 = xml `<?target data?>`;

     int i = 0;
     string output = "";
     foreach (xml|string) elem in pi1 {
         concatIntString(i, elem.toString());
         i += 1;
     }

     record {| (xml|string) value; |}? nextPI1 = pi1.iterator().next();

     panic error("out for loop: "+ output + "\n out record: "+ nextPI1.toString());
}

```
output for both scenarios: 

> error: out for loop: 
>  out record:

suleka96:
Fixes #24562 
Fixes #26731
Fixes #27532
Fixes #27533
Fixes #27715

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.xml/src/main/java/org/ballerinalang/langlib/xml/Concat.java
```diff
@@ -51,6 +53,14 @@ public static BXml concat(Object... arrayValue) {
...
                 lastItem = xmlText;
             } else {
+                if (refValue instanceof BXmlSequence) {
+                    List<BXml> sequenceChildren = ((BXmlSequence) refValue).getChildrenList();
+                    for (int j = 0; j < sequenceChildren.size(); j++) {
+                        backingArray.add(sequenceChildren.get(j));
+                    }
+                    lastItem = (BXml) refValue;
+                    continue;
+                }
```
MaryamZi, line 63: Can't we keep the else logic as is and make this an `else if`?

langlib/lang.xml/src/main/java/org/ballerinalang/langlib/xml/Concat.java
```diff
@@ -51,6 +53,14 @@ public static BXml concat(Object... arrayValue) {
                 backingArray.add(xmlText);
                 lastItem = xmlText;
             } else {
+                if (refValue instanceof BXmlSequence) {
```
MaryamZi, line 56: Have we added tests for this fix?

langlib/lang.xml/src/main/java/org/ballerinalang/langlib/xml/Concat.java
```diff
@@ -51,6 +53,14 @@ public static BXml concat(Object... arrayValue) {
...
+                    List<BXml> sequenceChildren = ((BXmlSequence) refValue).getChildrenList();
+                    for (int j = 0; j < sequenceChildren.size(); j++) {
+                        backingArray.add(sequenceChildren.get(j));
+                    }
```
MaryamZi, line 60: Can't we just do
```suggestion
                    backingArray.addAll(((BXmlSequence) refValue).getChildrenList());
```

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-literals-negative.bal
```diff
@@ -72,3 +72,16 @@ function testXmlNsInterpolation() returns xml {
...
+function testXMLLiteralWithEscapeSequence(){
+    xml x1 = xml `hello &lt; &gt; &amp;`;
+    int i = 0;
+    string[] strs = [];
+    foreach string|xml e in x1 {
+        if e is string {
+            strs[i] = e;
+            i += 1;
+        }
+    }
```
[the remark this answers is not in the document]
suleka96: since we changed the iterator return type to xml fom xml|string this tests if we get the below error.

>  incompatible types: 'xml' will not be matched to 'string'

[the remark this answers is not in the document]
MaryamZi: Sorry, missed that this was a negative test. In the long term though, this will become outdated once we allow the type test with intersections. Let's leave this as is until then.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlText.java
```diff
@@ -113,7 +114,7 @@ public boolean hasNext() {
             public Object next() {
                 if (!read) {
                     this.read = true;
-                    return data;
+                    return that;
```
MaryamZi, line 117: Can't we just do 
```suggestion
                    return this;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1433,7 +1433,7 @@ public void setForeachTypedBindingPatternType(BLangForeach foreachNode) {
                 varType = inferRecordFieldType(recordType);
                 break;
             case TypeTags.XML:
-                varType = BUnionType.create(null, symTable.xmlType, symTable.stringType);
+                varType = symTable.xmlType;
```
MaryamZi, line 1436: I think there's a deviation here too.

The [spec](https://ballerina.io/spec/lang/master/#section_5.1.6) says for `xml<T>` the iteration value type is `T`.

So the following should work, right?

```ballerina
public function main() {
    xml<'xml:Element> el = xml `<foo>foo</foo><bar/>`;

    foreach 'xml:Element item in el {
        
    }
}
```

MaryamZi: Similarly, the `lang.xml:iterator()` also needs to change.

The following doesn't work either.

```ballerina
public function main() {
    xml<'xml:Element> el = xml `<foo>foo</foo><bar/>`;

    record {| 'xml:Element value; |}? nextResult = el.iterator().next();
}
```

MaryamZi: Maybe we can merge this PR first and then fix the `xml<T>` scenario.

langlib/langlib-test/src/test/resources/test-src/xmllib_test.bal
```diff
@@ -62,6 +62,25 @@ function testConcat() returns xml {
     return 'xml:concat(x, <xml> testFromString(), "hello from String");
 }
 
+function testConcatWithXMLSequence() {
+    string a = "string one";
+    string b = "string two";
+
+    xml c = 'xml:concat(catalog, a, b);
```
MaryamZi, line 69: Shall we also assert here that the original `catalog` does not change?

Before calling `concat` we can create a clone, and then check `==` for that clone and `catalog` after calling `clone`.

tests/ballerina-spec-conformance-tests/src/S05_values_types_variables/tests/iterable_types.bal
```diff
@@ -110,9 +110,7 @@ function testIterableTypeXML() {
                         </bookstore>`;
     string result = "";
     foreach var x in bookstore["book"] {
-        if x is xml {
             result += x["title"].getTextValue();
```
MaryamZi, line 113:
```suggestion
        result += x["title"].getTextValue();
```

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-literals-negative.bal
```diff
@@ -72,3 +71,13 @@ function testXmlNsInterpolation() returns xml {
...
+
+function testXMLLiteralWithEscapeSequence() {
+    xml x1 = xml `hello &lt; &gt; &amp;`;
+    string[] strs = [];
+    foreach xml e in x1 {
+        if e is string {
+            strs.push(e);
+        }
+    }
+}
```
MaryamZi, line 83: Missing a new line.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlText.java
suleka96: Not really sure why but when using `this` the program crashes with a bad sad error for examples like: 

```
xml bitOfText = xml `bit of text\u2702\u2705`;

function xmlCharItemIter() returns string {
    string result = "";
    int i = 0;
    foreach xml elem in bitOfText {
        string str = io:sprintf("%s\n", elem);
        result += str;
        i += 1;
    }
    return result;
}
```

> class io.ballerina.runtime.internal.values.XmlText$1 cannot be cast to class io.ballerina.runtime.internal.values.XmlValue (io.ballerina.runtime.internal.values.XmlText$1 and io.ballerina.runtime.internal.values.XmlValue are in unnamed module of loader 'app')

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml_iteration.bal
```diff
@@ -142,10 +140,103 @@ function xmlSequenceIter() returns string {
...
+    foreach xml elem in bitOfText {
         string str = io:sprintf("%s\n", elem);
         result += str;
+        i += 1;
     }
     return result;
 }
+
+function xmlTypeParamElementIter() {
+    xml<'xml:Element> el1 = xml `<foo>foo</foo><bar/>`;
```
MaryamZi, line 153: These tests are bound to break once we fix https://github.com/ballerina-platform/ballerina-lang/issues/26340, right?

Instead shall we rewrite these tests similar to 

```ballerina
public function main() {
    xml el = xml `<foo>foo</foo>`;
    xml<'xml:Element> seq = <xml<'xml:Element>> el.concat(xml `<bar/>`);


    foreach 'xml:Element x in seq {
        
    }
}
```

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml_iteration.bal
```diff
@@ -142,10 +140,103 @@ function xmlSequenceIter() returns string {
...
+    int i = 0;
+    foreach xml elem in bitOfText {
         string str = io:sprintf("%s\n", elem);
         result += str;
+        i += 1;
     }
     return result;
 }
+
+function xmlTypeParamElementIter() {
```
MaryamZi, line 152: I think we can call these tests `testXmlElementSequenceIteration`, `testXmlElementSequenceIteration`, etc.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -294,6 +295,29 @@ private void findTypeParam(Location loc, BType expType, BType actualType, Symbol
         }
         // Bound type is a structure. Visit recursively to find bound type.
         switch (expType.tag) {
+            case TypeTags.XML:
+                if (TypeTags.isXMLTypeTag(actualType.tag)) {
```
MaryamZi, line 299: We can check `(!TypeTags.isXMLTypeTag(actualType.tag))` and return early here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -294,6 +295,29 @@ private void findTypeParam(Location loc, BType expType, BType actualType, Symbol
...
+                    switch (actualType.tag) {
+                        case TypeTags.XML:
+                            BType constraint = ((BXMLType) actualType).constraint;
+                            while (constraint.tag == TypeTags.XML) {
+                                actualType = constraint;
+                                constraint = ((BXMLType) actualType).constraint;
+                            }
+                            findTypeParam(loc, ((BXMLType) expType).constraint, constraint, env,
+                                    resolvedTypes, result);
+                            break;
```
MaryamZi, line 309: Do we need to break in these cases? We can directly return, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -294,6 +295,29 @@ private void findTypeParam(Location loc, BType expType, BType actualType, Symbol
...
         // Bound type is a structure. Visit recursively to find bound type.
         switch (expType.tag) {
+            case TypeTags.XML:
+                if (TypeTags.isXMLTypeTag(actualType.tag)) {
+                    switch (actualType.tag) {
+                        case TypeTags.XML:
+                            BType constraint = ((BXMLType) actualType).constraint;
+                            while (constraint.tag == TypeTags.XML) {
+                                actualType = constraint;
+                                constraint = ((BXMLType) actualType).constraint;
```
MaryamZi, line 305: Can't we directly do 
```suggestion
                                constraint = ((BXMLType) constraint).constraint;
```
instead?

langlib/lang.xml/src/main/ballerina/xml.bal
```diff
@@ -77,7 +77,7 @@ type XmlType xml;
 # + return - iterator object
 # Each item is represented by an xml singleton.
 public isolated function iterator(xml<ItemType> x) returns object {
```
MaryamZi, line 79: Shall we add some tests in LangLibXMLTest for the direct invocation of `lang.xml:iterator()` too?

```ballerina
    xml x = xml `<!--first-->`;
    xml<'xml:Comment> seq = <xml<'xml:Comment>> x.concat(xml `<!--second-->`);

    object {
        public isolated function next() returns record {| 'xml:Comment value; |}?;
    } iter = seq.iterator();

    // tests based on `iter.next()`.
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/PredefinedTypes.java
```diff
@@ -170,9 +170,15 @@
...
     public static final RecordType STRING_ITR_NEXT_RETURN_TYPE =
             IteratorUtils.createIteratorNextReturnType(PredefinedTypes.TYPE_STRING);
-    public static final RecordType XML_ITR_NEXT_RETURN_TYPE = IteratorUtils
-            .createIteratorNextReturnType(
-                    new BUnionType(Arrays.asList(PredefinedTypes.TYPE_STRING, PredefinedTypes.TYPE_XML)));
+
+    public static final RecordType XML_ITR_NEXT_RETURN_ELEMENT_TYPE =
+            IteratorUtils.createIteratorNextReturnType(TYPE_ELEMENT);
+    public static final RecordType XML_ITR_NEXT_RETURN_TEXT_TYPE =
+            IteratorUtils.createIteratorNextReturnType(TYPE_TEXT);
```
MaryamZi, line 177: How come we don't need to do this for processing instruction and comment sequences?

suleka96: They are not iterable so I didn't not create iteration return types for them.

MaryamZi: > They are not iterable so I didn't not create iteration return types or write type param logic for them.

`xml:Element` also isn't iterable, right? But there's a `XML_ITR_NEXT_RETURN_ELEMENT_TYPE`?

These are the record value types for iterator next right?

We can iterate `xml`, `xml<T>`, and `xml:Text`. 

So I assume we will need record values with all of the following field types
- `xml:Element|xml:Text|xml:ProcessingInstruction|xml:Comment`
- `xml:Element`
- `xml:Text`
- `xml:ProcessingInstruction`
- `xml:Comment`
- and other combinations of unions of these (e.g., `xml:Comment|xml:Element` for `xml<xml:Comment|xml:Element>`).

Thus my question re: why we don't need the others. How do we handle the union case btw?

suleka96: xml<xml:Text> and xml<xml:ProcessingInstruction> are not iterable in the sense it returns null from `next`

for examples like:
```
xml d = xml `<!--first-->`;
        xml<'xml:Element|'xml:Comment> seq4 = <xml<'xml:Element|'xml:Comment>> d.concat(xml `<foo>fu</foo>`);

         int i =0;
         foreach 'xml:Element|'xml:Comment item in seq4 {
             concatIntString(i, item.toString());
             i += 1;
         }

         record {| 'xml:Element|'xml:Comment value; |}? nextEle = seq4.iterator().next();

         panic error("out for loop: "+ output + "\n out record: "+ nextEle.toString());
```
the current implementation returns: 
```
error: out for loop: 0:<!--first--> 1:<foo>fu</foo> 
 out record: {"value":`<!--first-->`}
```
This is the expected output right?

The BXml type for this is : 

> xml<lang.xml:Element|lang.xml:Comment|lang.xml:ProcessingInstruction|lang.xml:Text>

I'm switching from the  BXml type.


`XML_ITR_NEXT_RETURN_ELEMENT_TYPE` is to handle examples like:

```
xml xdata = xml `<p:person xmlns:p="foo" xmlns:q="bar">
                <p:name>bob</p:name>
                <p:address>
                    <p:city>NY</p:city>
                    <q:country>US</q:country>
                </p:address>
                <q:ID>1131313</q:ID>
            </p:person>`;
```

who's BXml type is `lang.xml:Element`

MaryamZi: > xmlxml:Text and xmlxml:ProcessingInstruction are not iterable in the sense it returns null from `next`
> 

IINM
- `xml<'xml:Text>` is the same as `'xml:Text` and iteration will always return one value which is the `'xml:Text` itself
- Iterating `xml<'xml:ProcessingInstruction>` will return 0 or more `'xml:ProcessingInstruction`s

I'm not sure I get what you mean by `next` returning null. 

> for examples like:
> 
> ```
> xml d = xml `<!--first-->`;
>         xml<'xml:Element|'xml:Comment> seq4 = <xml<'xml:Element|'xml:Comment>> d.concat(xml `<foo>fu</foo>`);
> 
>          int i =0;
>          foreach 'xml:Element|'xml:Comment item in seq4 {
>              concatIntString(i, item.toString());
>              i += 1;
>          }
> 
>          record {| 'xml:Element|'xml:Comment value; |}? nextEle = seq4.iterator().next();
> 
>          panic error("out for loop: "+ output + "\n out record: "+ nextEle.toString());
> ```
> 
> the current implementation returns:
> 
> ```
> error: out for loop: 0: 1:<foo>fu</foo> 
>  out record: {"value":`<!--first-->`}
> ```
> 
> This is the expected output right?
> 
> The BXml type for this is :
> 
> > xmllang.xml:Element|lang.xml:Comment|lang.xml:ProcessingInstruction|lang.xml:Text

The output is correct, but shouldn't the iteration value type here be `'xml:Element|'xml:Comment` since we're iterating over `xml<'xml:Element|'xml:Comment>`?

> 
> I'm switching from the BXml type.
> 
> `XML_ITR_NEXT_RETURN_ELEMENT_TYPE` is to handle examples like:
> 
> ```
> xml xdata = xml `<p:person xmlns:p="foo" xmlns:q="bar">
>                 <p:name>bob</p:name>
>                 <p:address>
>                     <p:city>NY</p:city>
>                     <q:country>US</q:country>
>                 </p:address>
>                 <q:ID>1131313</q:ID>
>             </p:person>`;
> ```
> 
> who's BXml type is `lang.xml:Element`

Not sure why we need to special case this. Again, IMO the iteration value type should be based on the variable's type, and not the type of the actual value it holds.

When we iterate over `xdata`, we're iterating over an `xml` which is `xml<'xml:Element|'xml:Comment|'xml:Text|'xml:ProcessingInstruction>`, so IMO the iteration type here should be `'xml:Element|'xml:Comment|'xml:Text|'xml:ProcessingInstruction`.

suleka96: As we discussed `xml<xml:Comment> and xml<xml:ProcessingInstruction>` is now iterable. Also regarding unions, we are unable to write a xml union sequence with lang capabilities at the moment. However, after discussing with @rdhananjaya,  instead of switching from 'BXml' types it is now switched from the 'xml value type' so a union case won't exist now.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlComment.java
```diff
@@ -45,6 +46,28 @@ public XmlComment(String data, boolean readonly) {
...
+            @Override
+            public boolean hasNext() {
+                return !read;
+            }
+
+            @Override
+            public Object next() {
+                if (!read) {
+                    this.read = true;
+                    return that;
```
MaryamZi, line 63: Using `XmlComment.this` may work instead of introducing the new variable `that`.

langlib/langlib-test/src/test/resources/test-src/xmllib_test.bal
```diff
@@ -382,6 +397,53 @@ function testElementChildrenNS() {
     assert(toNoNs.toString(), "<to>Irshad</to><to>Irshad</to>");
 }
 
+function testXMLIteratorInvocation() {
```
MaryamZi, line 400: Shouldn't the following result in a compilation error?

```ballerina
public function main() {
    'xml:Comment x = xml `<!-- foo -->`;
    var iterator = x.iterator();
}
```

Please check for the other non-text types too. 

This seems to fail on slp8 too, let's create an issue.

MaryamZi:
@suleka96, let's sync the PR too.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlComment.java
suleka96: Seems like we'll have to use `"that"` here as well. It gives the same error as when testing with XmlText.
```
org.ballerinalang.core.util.exceptions.BLangRuntimeException: error: java.lang.ClassCastException {"message":"class io.ballerina.runtime.internal.values.XmlComment$1 cannot be cast to class io.ballerina.runtime.api.values.BXml (io.ballerina.runtime.internal.values.XmlComment$1 and io.ballerina.runtime.api.values.BXml are in unnamed module of loader 'app')"}
        at ballerina.lang.xml.0_8_0:externNext(internal.bal:35)
           ballerina.lang.xml.0_8_0.XMLIterator:next(internal.bal:31)
           xml_iteration:testXmlCommentSequenceIteration(xml_iteration.bal:195)
        at org.ballerinalang.test.BRunUtil.invoke(BRunUtil.java:440)
        at org.ballerinalang.test.BRunUtil.invokeFunction(BRunUtil.java:1198)
        at org.ballerinalang.test.BRunUtil.invoke(BRunUtil.java:188)
        at org.ballerinalang.test.BRunUtil.invoke(BRunUtil.java:1241)
        at org.ballerinalang.test.types.xml.XMLIterationTest.testXMLTypesForeachOp(XMLIterationTest.java:148)
```

langlib/langlib-test/src/test/resources/test-src/xmllib_test.bal
suleka96: #27848

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -492,6 +517,12 @@ private void findTypeParamInUnion(Location loc, BType expType, BUnionType actual
             if (type.tag == TypeTags.MAP) {
                 members.add(((BMapType) type).constraint);
             }
+            if (TypeTags.isXMLTypeTag(type.tag)) {
+                if (type.tag == TypeTags.XML) {
+                    members.add(((BXMLType) type).constraint);
+                }
+                members.add(type);
```
MaryamZi, line 524: Do we need to add the type also when `type.tag == TypeTags.XML` is true?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1494,7 +1494,55 @@ public void setForeachTypedBindingPatternType(BLangForeach foreachNode) {
...
+                        varType = symTable.xmlPIType;
+                        break;
+                    default:
+                        Set<BType> collectionTypes = getEffectiveMemberTypes((BUnionType) constraint);
+                        Set<BType> builtinXMLConstraintTypes = getEffectiveMemberTypes
+                                ((BUnionType) ((BXMLType) symTable.xmlType).constraint);
+                        if (collectionTypes.size() == 4 && builtinXMLConstraintTypes.equals(collectionTypes)) {
+                            varType = symTable.xmlType;
+                        } else {
+                            LinkedHashSet<BType> collectionTypesInSymTable = new LinkedHashSet<>();
```
MaryamZi, line 1522: Is this variable name accurate?

langlib/lang.xml/src/main/java/org/ballerinalang/langlib/xml/Next.java
```diff
@@ -49,10 +50,21 @@ public static Object next(BObject m) {
...
+                case TypeTags.XML_TEXT_TAG:
+                    return ValueCreator.createRecordValue(ValueCreator.createMapValue
+                            (PredefinedTypes.XML_ITR_NEXT_RETURN_TEXT_TYPE), xmlValue);
+                case TypeTags.XML_COMMENT_TAG:
+                    return ValueCreator.createRecordValue(ValueCreator.createMapValue
+                            (PredefinedTypes.XML_ITR_NEXT_RETURN_COMMENT_TYPE), xmlValue);
+                case TypeTags.XML_PI_TAG:
+                    return ValueCreator.createRecordValue(ValueCreator.createMapValue
+                            (PredefinedTypes.XML_ITR_NEXT_RETURN_PI_TYPE), xmlValue);
+            }
```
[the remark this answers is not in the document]
suleka96: #27849

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1494,7 +1494,55 @@ public void setForeachTypedBindingPatternType(BLangForeach foreachNode) {
...
+                                    case TypeTags.XML_TEXT:
+                                        collectionTypesInSymTable.add(symTable.xmlTextType);
+                                        break;
+                                    case TypeTags.XML_PI:
+                                        collectionTypesInSymTable.add(symTable.xmlPIType);
+                                        break;
+                                }
+
+                            }
+                            varType = BUnionType.create(null, collectionTypesInSymTable);
```
MaryamZi, line 1540: Why can't we directly use `constraint` as the var type here?

langlib/langlib-test/src/test/resources/test-src/xmllib_test.bal
```diff
@@ -382,6 +397,53 @@ function testElementChildrenNS() {
...
+    xml a = xml `<!--first-->`;
+    xml<'xml:Comment> seq1 = <xml<'xml:Comment>> a.concat(xml `<!--second-->`);
+
+    object {
+        public isolated function next() returns record {| 'xml:Comment value; |}?;
+    } iter1 = seq1.iterator();
+
+    assert((iter1.next()).toString(), "{\"value\":`<!--first-->`}");
```
MaryamZi, line 408: We should ideally test iterating the entire sequence. Including `()` being returned after everything has been iterated.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/xml/XMLIterationTest.java
```diff
@@ -54,14 +54,64 @@ public void setup() {
...
-                                  "invalid tuple binding pattern: expected a tuple type, but found '(xml|string)'",
-                                  11, 17);
+                                  "invalid tuple binding pattern: attempted to infer a tuple type, but found 'xml'",
+                                  13, 17);
         BAssertUtil.validateError(negative, index++, "incompatible types: expected " +
                 "'function ((xml:Element|xml:Comment|xml:ProcessingInstruction|xml:Text)) returns ()'," +
-                " found 'function ([int,xml,string]) returns ()'", 16, 19);
+                " found 'function ([int,xml,string]) returns ()'", 18, 19);
+        BAssertUtil.validateError(negative, index++,
+                "incompatible types: expected 'other', found 'xml:Element'",
```
MaryamZi, line 66: Ideally, this error shouldn't be logged. The actual error is the next one, `'xml:Element' is not an iterable collection`.

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml_iteration.bal
```diff
@@ -103,6 +100,7 @@ function chainedIterableOps() returns xml {
 xml theXml = xml `<book>the book</book>`;
 xml bitOfText = xml `bit of text\u2702\u2705`;
 xml compositeXml = theXml + bitOfText;
+string output = "";
```
MaryamZi, line 103: Why do we need a module level variable? Why can't we use local variables and pass them as arguments to `concatString`?

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml_iteration_negative.bal
```diff
@@ -7,6 +7,8 @@ xml xdata = xml `<p:person xmlns:p="foo" xmlns:q="bar">
                     <q:ID>1131313</q:ID>
                   </p:person>`;
 
+string result = "";
```
MaryamZi, line 10: Same comment as above re: using local variables instead.

MaryamZi (approved):
@suleka96, please send a separate PR addressing the suggestions here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
suleka96: Tracked in issue: #27851

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
suleka96: Tracked in issue: #27851

PR #26829, opened 2020-11-10
Fix the addition of multiple whitespaces through missing tokens

### LINKED ISSUE #26767: [Formatter] Handle addition of multiple whitespaces through missing tokens
**Description:**
When there are missing tokens, the formatter adds the trailing minutiaes to ensure the formatting in the remaining code is not distorted. But if there are multiple whitespaces added due to this, they should be reduced to 1 whitespace. 

For instance, a sample code would be as follows:

**Source:**
```ballerina
public function foo() returns {

}
```

**Current Output:**
```ballerina
public function foo() returns  {

}
```
Notice the 2 whitespaces after the `returns` keyword.

**Expected Output:**
```ballerina
public function foo() returns {

}
```

**Steps to reproduce:**
SLP5

**Affected Versions:**

IrushiL:
## Purpose

Fixes #26767

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
```diff
@@ -264,6 +264,9 @@
     // Range of the file to be formatted.
     private final LineRange lineRange;
 
+    // Previous token's trailing whitespaces.
+    private int prevTokensTrailingWS;
```
SupunS, line 268: Shall we move this to `env`?

IrushiL: Done

SupunS (approved):

PR #26882, opened 2020-11-12
Add support for included record parameters

### LINKED ISSUE #26047: Add support for included record parameters
**Description:**
When considering this function
```ballerina
// Definition
type Options record {
   boolean a = false;
   boolean b = false;
   boolean c = false;
};
// At the moment, there are two patterns that can be used when a function needs named parameters.

// A
function foo(string s, Options options = {}) { }

// B
public function foo(string x, int size = 10, boolean a = true) { }
```
With the included record parameter feature, the function definition and call become:
```ballerina
function foo(string s, *Options options) {}
foo("s", a = true);
```
Please refer https://github.com/ballerina-platform/ballerina-spec/issues/578

### LINKED ISSUE #26123: Parser changes for included record parameter feature
**Description:**
Please refer https://github.com/ballerina-platform/ballerina-spec/issues/578#issue-671556255

chiranSachintha:
```ballerina
type Options record {
   boolean a = false;
   boolean b;
   boolean c? ;
};
function foo(string s, *Options options) {}
```
The last parameter in the definition of `foo` is an included record parameter. With the use of include record parameter, we can pass field of an included record parameter as named argument to the function as below example, in addition to using it for named parameters.

```ballerina
type Person record {
    string firstName;
    string secondName;
};

//function definition
function bar(*Person person) returns string {
    return person.firstName + " " + person.secondName;
}

//function call
public function main() {
    string fullName = bar(firstName = "chiran", secondName = "sachintha");
}
```
If these [two rules](https://github.com/ballerina-platform/ballerina-spec/issues/578#issuecomment-709664731) are allowed, It can pass additional arguments to a function call as below example. (Mentioned additional arguments can be added as fields in  the inclusive included record parameter)
```ballerina
type Foo record {
    never a?;
    never b?;
    never c?;
    never d?;
};

type Bar record {|
    int a;
    int b;
|};

function value(int c, int d, *Bar bar, *Foo foo) {
}

public function main() {
    value(10, 10, a = 20, b = 20, id = 5, value = “foo”, path = “/var/log”);
}
```

Fixes #26047 
Fixes #26123 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -6390,6 +6392,39 @@ private void reorderNamedArgs(BLangInvocation iExpr, BInvokableSymbol invokableS
...
     }
 
+    private void createIncludedRecordLiteral(BLangInvocation iExpr, BVarSymbol param, Map<String,
+            BLangExpression> namedArgs, List<BLangExpression> args) {
+        BLangRecordLiteral recordLiteral = TreeBuilder.createBLangRecordLiteralNode();
+        recordLiteral.type = param.type;
+        if (param == ((BInvokableSymbol) iExpr.symbol).incRecordParamAllowAdditionalFields) {
+            for (String name : namedArgs.keySet()) {
+                BLangNamedArgsExpression expr = (BLangNamedArgsExpression) namedArgs.get(name);
+                if (!((BRecordType) param.type).fields.containsKey(name)) {
```
MaryamZi, line 6402: Wouldn't it be better to 
1. iterate through the param list and identify
  - all included record params of a closed record type
  - the single included record param of an open record type, if present
2. then iterate through the named args and add to the relevant records based on the expected names? 
  - if the name is present in a particular included record param, and the type is not a subtype of `never`, the field belongs to that record
  - if the name is present in a particular included record param, and the type is a subtype of `never`, the field does not belongs to that record
  - if the name is not present in a particular included record param, and it is an open record, the field belongs to that record

MaryamZi:
Quoting from https://github.com/ballerina-platform/ballerina-spec/issues/578#issuecomment-709664731

> when a record type R is included with *R, the names of fields allowed by individual field descriptors must be distinct from the parameter names and from the names of fields allowed by individual field descriptors of other included record parameters

shouldn't the following thus result in a compilation error?

```ballerina
type Options record {
    string options;
};

function foo(int id, *Options options) { // `Options` has a field with the same name as a parameter

}

public function main() {
    foo(1, options = {options: "bar"});
}
```

compiler/ballerina-lang/src/main/java/org/ballerinalang/model/TreeBuilder.java
```diff
@@ -450,6 +450,10 @@ public static RecordLiteralNode createRecordLiteralNode() {
         return new BLangRecordLiteral();
     }
 
+    public static BLangRecordLiteral createBLangRecordLiteralNode() {
+        return new BLangRecordLiteral();
```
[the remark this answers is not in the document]
chiranSachintha: Fixed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -6390,6 +6392,39 @@ private void reorderNamedArgs(BLangInvocation iExpr, BInvokableSymbol invokableS
         iExpr.requiredArgs = args;
     }
 
+    private void createIncludedRecordLiteral(BLangInvocation iExpr, BVarSymbol param, Map<String,
+            BLangExpression> namedArgs, List<BLangExpression> args) {
```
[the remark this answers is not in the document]
chiranSachintha: Fixed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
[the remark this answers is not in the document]
chiranSachintha: Changed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2378,13 +2385,34 @@ private void defineInvokableSymbolParams(BLangInvokableNode invokableNode, BInvo
...
                 symbol.defaultableParam = true;
             }
+            if (varNode.flagSet.contains(Flag.INCLUDED) && varNode.type.getKind() == TypeKind.RECORD) {
+                symbol.flags |= Flags.INCLUDED;
+                if (((BRecordType) varNode.type).restFieldType != symTable.noType) {
+                    openIncludedRecordParams.add(symbol);
+                }
+                LinkedHashMap<String, BField> fields = ((BRecordType) varNode.type).fields;
+                for (String field : fields.keySet()) {
+                    if (!Symbols.isFlagOn(Flags.asMask(fields.get(field).symbol.getFlags()), Flags.OPTIONAL)) {
```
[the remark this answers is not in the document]
chiranSachintha: this should be work. I misunderstood with optional fields and the disallow fields. I fixed it and now it works for optional fields also.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5302,7 +5303,8 @@ private BType checkInvocationParam(BLangInvocation iExpr) {
             switch (expr.getKind()) {
                 case NAMED_ARGS_EXPR:
                     foundNamedArg = true;
-                    if (i < parameterCount) {
+                    if (i < parameterCount + invokableSymbol.includedRecordParams.size() ||
```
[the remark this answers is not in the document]
chiranSachintha: I have fixed this. Now it fails with `too many arguments in call to 'foo()'` error.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BInvokableSymbol.java
```diff
@@ -40,6 +40,8 @@
 public class BInvokableSymbol extends BVarSymbol implements InvokableSymbol {
 
     public List<BVarSymbol> params;
+    public List<BVarSymbol> includedRecordParams;
+    public BVarSymbol incRecordParamAllowAdditionalFields;
```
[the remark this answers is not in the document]
chiranSachintha: If we are not keeping these two fields we need to check for an open included record which allows extra fields in typechecker and desugar phase also. We can avoid the effect of repeating the same task through this.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2412,6 +2440,26 @@ private void defineInvokableSymbolParams(BLangInvokableNode invokableNode, BInvo
         invokableSymbol.type.tsymbol = functionTypeSymbol;
     }
 
+    private boolean isIncRecordParamAllowAdditionalFields(List<BVarSymbol> inclusiveIncludedRecordParams,
+                                                          Set<String> requiredParamNames) {
+        if (inclusiveIncludedRecordParams.size() != 1) {
+            return false;
+        }
+
+        LinkedHashMap<String, BField> fields = ((BRecordType) inclusiveIncludedRecordParams.get(0).type).fields;
```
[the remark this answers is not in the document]
chiranSachintha: This happens due to early mention issue. I have fixed it. This program now fails with the following errors.
```ballerina
Compiling source
        test3.bal
ERROR [test3.bal:(149:23,149:39)] redeclared symbol 'id'
ERROR [test3.bal:(149:42,149:57)] redeclared symbol 'name'
ERROR [test3.bal:(154:19,154:24)] incompatible types: expected 'boolean', found 'string'

```

chiranSachintha:
> Quoting from [ballerina-platform/ballerina-spec#578 (comment)](https://github.com/ballerina-platform/ballerina-spec/issues/578#issuecomment-709664731)
> 
> > when a record type R is included with *R, the names of fields allowed by individual field descriptors must be distinct from the parameter names and from the names of fields allowed by individual field descriptors of other included record parameters
> 
> shouldn't the following thus result in a compilation error?
> 
> ```ballerina
> type Options record {
>     string options;
> };
> 
> function foo(int id, *Options options) { // `Options` has a field with the same name as a parameter
> 
> }
> 
> public function main() {
>     foo(1, options = {options: "bar"});
> }
> ```

In the [second point](https://github.com/ballerina-platform/ballerina-spec/issues/578#issuecomment-709664731), it has mentioned that `the record type has individual field descriptors that disallow all parameter names' and It needs to have a field with own name of it for allowing extra fields due to that fact.
example:
``` ballerina
type KVPairs record {
	never message?;
	never level?; 
        never pairs?; 
};
function log(string message, LogLevel level, *KVPairs pairs);
```
But the given [example](https://github.com/ballerina-platform/ballerina-spec/issues/578#issuecomment-690851647) is not in an above-mentioned way(It not contain a field with its own name). to allow that example I have excluded its name when checking. due to that reason above mention program has compiled properly.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5303,10 +5303,25 @@ private BType checkInvocationParam(BLangInvocation iExpr) {
...
+                if (!fields.isEmpty()) {
+                    for (String field : fields.keySet()) {
+                        if (fields.get(field).type.tag != TypeTags.NEVER) {
+                            parameterCountForNamedArgs = parameterCountForNamedArgs - 1;
+                            break;
+                        }
+                    }
+                }
+            }
+        }
```
KavinduZoysa, line 5325: Shall we rearrange the login inside for loop like this, it will be more readable,

```
for (BVarSymbol symbol : invokableSymbol.params) {
   if (!Symbols.isFlagOn(Flags.asMask(symbol.getFlags()), Flags.INCLUDED)) {
       continue;
   }
   .......
   if (fields.isEmpty()) {
       continue;
   }
   for (String field : fields.keySet()) {
   }
}
```

compiler/ballerina-parser/src/main/resources/syntax_diagnostic_message.properties
```diff
@@ -242,6 +242,8 @@ error.required.parameter.after.the.defaultable.parameter=required parameter afte
 error.named.arg.followed.by.positional.arg=named arg followed by positional arg
 error.arg.followed.by.rest.arg=arg followed by rest arg
 error.binding.pattern.not.allowed=binding pattern not allowed
+error.defaultable.parameter.cannot.be.included.record.parameter = defaultable parameter cannot be included \
```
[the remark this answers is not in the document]
chiranSachintha: In [here](https://github.com/ballerina-platform/ballerina-lang/blob/e4d3b21058bb46e5d3a73e3ef3d909897ebf4e6f/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java#L1761) I took the invalid node as `*`. So due to that, I set the error as `defaultable parameter cannot be included`. So if set the error as the mentioned way I want to change the invalid node as `=`.  Is there any special reason to make such a change?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -918,6 +918,9 @@ error.rest.arg.defined.after.named.arg=\
 error.missing.required.parameter=\
   missing required parameter ''{0}'' in call to ''{1}''()
 
+error.missing.required.included.record.parameter.field=\
+  missing non-defaultable required record field ''{0}'' in call to ''{1}''()
```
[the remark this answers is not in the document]
chiranSachintha: Changed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -1543,6 +1544,9 @@ public void visit(BLangSimpleVariable varNode) {
 
         BVarSymbol varSymbol = defineVarSymbol(varNode.name.pos, varNode.flagSet, varNode.type, varName, env,
                                                varNode.internal);
+        if (varNode.expr != null) {
+            varSymbol.defaultableParam = true;
+        }
```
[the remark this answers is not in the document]
chiranSachintha: Changed. Also, work with optional fields with the new changes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5571,6 +5593,28 @@ private void checkArrayLibSortFuncArgs(BLangInvocation iExpr) {
         }
     }
 
+    private BVarSymbol checkParameterNameForDefaultArgument(BLangIdentifier argName,
+                       BLangExpression expr, List<BVarSymbol> nonRestParams, List<BVarSymbol> includedRecordParams,
+                                                            BVarSymbol incRecordParamAllowAdditionalFields) {
```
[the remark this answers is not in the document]
chiranSachintha: Fixed.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/functions/FunctionWithIncludedRecordParam.java
```diff
@@ -0,0 +1,161 @@
...
+import org.ballerinalang.test.BAssertUtil;
+import org.ballerinalang.test.BCompileUtil;
+import org.ballerinalang.test.BRunUtil;
+import org.ballerinalang.test.CompileResult;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * Test function signatures and calling with included record params.
```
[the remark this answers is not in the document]
chiranSachintha: Fixed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
chiranSachintha: Changed.

compiler/ballerina-parser/src/main/resources/syntax_diagnostic_message.properties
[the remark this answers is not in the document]
MaryamZi: This was based on the assumption that when the user sees this error it's more likely that they'll continue to have it as an included param rather than a defaultable param. 

Given that it's just an assumption, let's go ahead with this.

But shall we say something like "a defaultable parameter cannot be an include parameter"?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BInvokableSymbol.java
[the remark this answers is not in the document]
MaryamZi: In desugar it's just the additional checks for two flags, right? One to see if the parameter has the `INCLUDED` flag and the other to see if the record is open? IMO, that is OK rather than adding these fields just for the desugar phase.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Refactor the logic according to the mentioned way.

compiler/ballerina-parser/src/main/resources/syntax_diagnostic_message.properties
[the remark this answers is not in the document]
chiranSachintha: Changed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BInvokableSymbol.java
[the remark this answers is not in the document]
chiranSachintha: Removed `includedRecordParams` and `incRecordParamAllowAdditionalFields` fields.

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -945,13 +945,17 @@
         {
             "name": "RequiredParameterNode",
             "base": "ParameterNode",
-            "kind": "REQUIRED_PARAM",
```
SupunS, line 948: Do we need to remove the node kind?

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -1669,9 +1669,8 @@ private STNode getUpdatedParamList(STNode parameters, int index) {
                 case REQUIRED_PARAM:
                     STRequiredParameterNode requiredParam = (STRequiredParameterNode) param;
                     if (isEmpty(requiredParam.paramName)) {
-                        param = STNodeFactory
-                                .createRequiredParameterNode(requiredParam.annotations, requiredParam.typeName,
-                                        paramName);
+                        param = STNodeFactory.createRequiredParameterNode(SyntaxKind.REQUIRED_PARAM,
+                                requiredParam.annotations, null, requiredParam.typeName, paramName);
```
SupunS, line 1673: Can we use `STNodeFactory.createEmptyNode()` instead of `null`?

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -1949,15 +1948,23 @@ private STNode parseParameterRhs(SyntaxKind tokenKind) {
      */
     private STNode parseParameter(SyntaxKind prevParamKind, boolean isParamNameOptional) {
         STNode annots;
+        STNode inclusionSymbol;
         STToken nextToken = peek();
         switch (nextToken.kind) {
             case AT_TOKEN:
+                inclusionSymbol = STNodeFactory.createEmptyNode();
```
SupunS, line 1955: Will this support `function foo(@someAnnot *Bar b)`?

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
SupunS: Looks like we are using two different kinds for the same node.
But that allows someone to have the asterisk-token in the node, but use the `REQUIRED_PARAM` as the node kind, which is not correct. We can avoid that if we introduce a separate node for `INCLUDED_RECORD_PARAM`.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -2047,12 +2053,18 @@ private STNode parseEllipsis() {
...
      * @return Parsed parameter node
      */
-    private STNode parseParameterRhs(SyntaxKind prevParamKind, STNode annots, STNode type,
+    private STNode parseParameterRhs(SyntaxKind prevParamKind, STNode annots, STNode inclusionSymbol, STNode type,
                                      STNode paramName) {
         STToken nextToken = peek();
         // Required parameters
         if (isEndOfParameter(nextToken.kind)) {
-            return STNodeFactory.createRequiredParameterNode(annots, type, paramName);
+            if (inclusionSymbol != null && inclusionSymbol.kind == SyntaxKind.ASTERISK_TOKEN) {
```
SupunS, line 2061: Do we need this second check `inclusionSymbol.kind == SyntaxKind.ASTERISK_TOKEN`?

compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/declarations/FunctionParamsTest.java
```diff
@@ -66,4 +66,14 @@ public void testIncompleteBinarExprInDefaultableParams() {
...
+
+    @Test
+    public void testIncludedRecordParam() {
+        test("func-definition/func_params_source_09.bal", "func-definition/func_params_assert_09.json");
+    }
+
+    @Test
+    public void testIncludedRecordParamWithDefaultValues() {
+        test("func-definition/func_params_source_10.bal", "func-definition/func_params_assert_10.json");
+    }
```
SupunS, line 78: Shall we add some negative syntax tests as well?
e.g: additional tokens before and after asterisk, missing tokens before and after asterisk, etc

SupunS (changes requested):
Need to include this change in the error handler as well.

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
chiranSachintha: I did this because the asterisk is the only additional token in `included record param` when compared with `required param`. yes. It can be done. Isn't it an intentional work done in the parser side?

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
chiranSachintha: Fixed.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
chiranSachintha: Fixed. Now it works with `annotations` also.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
chiranSachintha: Removed.

compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/declarations/FunctionParamsTest.java
chiranSachintha: Added.

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
SupunS: Catch is, with the existence of asterisk, the node becomes a completely different node, with a different node kind. So the issue I mentioned above is there.

The intention of optional fields is to support truly optional fields for a given node. For eg. metadata (documentation and annotations) is an optional node. But the owner doesn't become a different node, even if the metadata is there or not.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -1978,11 +1989,10 @@ private STNode parseParameter(SyntaxKind prevParamKind, boolean isParamNameOptio
         }
 
         STNode type = parseTypeDescriptor(ParserRuleContext.TYPE_DESC_BEFORE_IDENTIFIER);
```
lochana-chathura, line 1991: Parsing the inclusionSymbol in a single place like this is cleaner to me
```suggestion
        STNode inclusionSymbol = parseInclusionSymbol();
        STNode type = parseTypeDescriptor(ParserRuleContext.TYPE_DESC_BEFORE_IDENTIFIER);
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -7606,6 +7626,21 @@ private STNode parseAtToken() {
...
+     */
+    private STNode parseAsteriskToken() {
+        STToken nextToken = peek();
+        if (nextToken.kind == SyntaxKind.ASTERISK_TOKEN) {
+            return consume();
+        } else {
+            recover(nextToken, ParserRuleContext.ASTERISK);
+            return parseAsteriskToken();
+        }
+    }
```
lochana-chathura, line 7642: Related to above comment.
```suggestion
    private STNode parseInclusionSymbol() {  
        STToken nextToken = peek();
        if (nextToken.kind == SyntaxKind.ASTERISK_TOKEN) {
            return consume();
        } else {
            return STNodeFactory.createEmptyNode();
        }
    }
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -1949,15 +1948,27 @@ private STNode parseParameterRhs(SyntaxKind tokenKind) {
      */
     private STNode parseParameter(SyntaxKind prevParamKind, boolean isParamNameOptional) {
         STNode annots;
+        STNode inclusionSymbol;
         STToken nextToken = peek();
         switch (nextToken.kind) {
             case AT_TOKEN:
                 annots = parseOptionalAnnotations();
+                if (peek().kind == SyntaxKind.ASTERISK_TOKEN) {
+                    inclusionSymbol = parseAsteriskToken();
```
rdulmina, line 1957: We can directly use `consume()` here

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -1949,15 +1948,27 @@ private STNode parseParameterRhs(SyntaxKind tokenKind) {
...
+                } else {
+                    inclusionSymbol = STNodeFactory.createEmptyNode();
+                }
+                break;
+            case ASTERISK_TOKEN:
+                inclusionSymbol = parseAsteriskToken();
```
rdulmina, line 1963: can use `consume()` and remove `parseAsteriskToken()` method

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/ParserTestFormatter.java
```diff
@@ -173,7 +173,8 @@ public void test(String fileName, String path) throws IOException {
                 "annot_decl_source_02.bal", "class_def_source_11.bal", "object_type_def_source_26.bal",
                 "ambiguity_source_32.bal", "service_object_constructor_source_01.bal",
                 "service_object_constructor_source_02.bal", "object_constructor_source_08.bal",
-                "annotations_source_01.bal", "do_stmt_source_08.bal");
+                "annotations_source_01.bal", "do_stmt_source_08.bal", "func_params_source_10.bal",
+                "func_params_source_11.bal", "func_params_source_13.bal");
```
SupunS, line 177: Why do we have to skip these tests?

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -1949,15 +1948,27 @@ private STNode parseParameterRhs(SyntaxKind tokenKind) {
...
+        STNode inclusionSymbol;
         STToken nextToken = peek();
         switch (nextToken.kind) {
             case AT_TOKEN:
                 annots = parseOptionalAnnotations();
+                if (peek().kind == SyntaxKind.ASTERISK_TOKEN) {
+                    inclusionSymbol = parseAsteriskToken();
+                } else {
+                    inclusionSymbol = STNodeFactory.createEmptyNode();
+                }
```
SupunS, line 1960: This will not handle syntax errors that could come after annotations. We need to introduce a another level to parse the content after the annots.

rdulmina:
```
public function foo(*int a) {
}
```
No error given at compile time. Is this the expected behavior?

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
rdulmina: +1 I think we need to handle recovery for the included record param

lochana-chathura: isn't it enough going to recovery inside the parseTypeDescriptor when there's an error followed by annotations? 
Reason is error handler will never pick `INSERT *` as the optimal path. It will always go for the required-param, defaultable-param or rest-param

lochana-chathura: sorry. we can something like `@untainted invalid token *Options options`
+1

SupunS (approved):

SupunS:
As discussed offline, we can do the error handling improvements and the node changes in a separate PR.

chiranSachintha:
> As discussed offline, we can do the parser error handling and the node changes in a separate PR.
> 
> @chiranSachintha Shall we create issues to track these?

Add a comment on this [issue](https://github.com/ballerina-platform/ballerina-lang/issues/26123) to track those suggestions.

hasithaa (approved):

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
chiranSachintha: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/26882

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
chiranSachintha: Fixed.

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/ParserTestFormatter.java
chiranSachintha: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/27662

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
chiranSachintha: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/27662

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
chiranSachintha: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/27662

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
chiranSachintha: Changed as we discussed offline. https://github.com/ballerina-platform/ballerina-lang/pull/27662

PR #26888, opened 2020-11-12
Add parser support for isolated services and object constructor expressions

### LINKED ISSUE #26696: Parser support for isolated services and object constructor expressions
**Description:**
Need to add $title.

lochana-chathura:
## Purpose
$subject.

Fixes #26696 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -924,7 +924,8 @@ protected Result seekMatch(ParserRuleContext currentCtx, int lookahead, int curr
                     hasMatch = nextToken.kind == SyntaxKind.CLOSE_PAREN_TOKEN;
                     break;
                 case OBJECT_CONSTRUCTOR_QUALIFIER:
-                    hasMatch = nextToken.kind == SyntaxKind.CLIENT_KEYWORD;
+                    hasMatch = nextToken.kind == SyntaxKind.CLIENT_KEYWORD ||
+                            nextToken.kind == SyntaxKind.ISOLATED_KEYWORD;
```
rdulmina, line 928: Do we need to address `isolated client object {..` case in error handler?

rdulmina: Currently, it is not handled right?

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -3847,24 +3855,28 @@ private STNode parseTerminalExpression(boolean isRhsExpr, boolean allowActions,
...
      * @param allowActions Allow actions
      * @return Parsed node
      */
-    private STNode parseTerminalExpression(STNode annots, boolean isRhsExpr, boolean allowActions,
-                                           boolean isInConditionalExpr) {
+    private STNode parseTerminalExpression(STNode annots, List<STNode> qualifiers, boolean isRhsExpr,
+                                           boolean allowActions, boolean isInConditionalExpr) {
         // Whenever a new expression start is added, make sure to
         // add it to all the other places as well.
+        parseTypeDescQualifiers(qualifiers);
```
SupunS, line 3878: parsing the type-def qualifiers may not be the best approach. May be we could introduce a separate method?
For. e.g: in an expression context, `transactional` is a valid expression, not a qualifier. so we should not parse it as a qualifier.

SupunS (changes requested):

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -3874,131 +3886,141 @@ private STNode parseTerminalExpression(STNode annots, boolean isRhsExpr, boolean
             case FALSE_KEYWORD:
             case DECIMAL_FLOATING_POINT_LITERAL_TOKEN:
             case HEX_FLOATING_POINT_LITERAL_TOKEN:
+                reportInvalidExpressionAnnots(annots, qualifiers);
+                reportInvalidQualifierList(qualifiers);
```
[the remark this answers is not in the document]
lochana-chathura: If we employ the previous approach we have following issues when re-parsing the parseTerminalExpression().
1. Validation will take place again and invalid nodes are reported twice. We can avoid this using flag though.
2.  We can't go to recovery with the qualifiers. Bcz we have already validated them before reaching the recovery method call.  

I'll try to think of another way.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
lochana-chathura: In error handler we don't have a proper way to solve this issue. But, in a very few cases we have handled it like below.
```java
    private static final ParserRuleContext[] OBJECT_CONSTRUCTOR_START =
            { ParserRuleContext.OBJECT_KEYWORD, ParserRuleContext.ISOLATED_KEYWORD, 
              ParserRuleContext.CLIENT_KEYWORD };

    private static final ParserRuleContext[] OBJECT_CONSTRUCTOR_WITHOUT_ISOLATED =
            { ParserRuleContext.OBJECT_KEYWORD, ParserRuleContext.CLIENT_KEYWORD };

    private static final ParserRuleContext[] OBJECT_CONSTRUCTOR_WITHOUT_CLIENT =
            { ParserRuleContext.OBJECT_KEYWORD, ParserRuleContext.ISOLATED_KEYWORD };
```

But, my concern with above logic is; it adds **cyclic** alternative paths.
For example, something like `isolated client isolated client isolated` is an alternative path.

So, If we assume `LOOKAHEAD_LIMIT = 5` and entry point is `OBJECT_CONSTRUCTOR_START`, 
In 2 qualifier case: we check for `2 + 2x1 + 2x1x1 + 2x1x1x1 + 2x1x1x1x1 = 10` alternative qualifier paths.
In 4 qualifier case: we check for `4 + 4x3 + 4x3x3 + 4x3x3x3 + 4x3x3x3x3 = 484` alternative qualifier paths.
which are all unnecessary. (IINM)

We could rather define rules like below. (aliasing `OBJECT_CONSTRUCTOR_QUALIFIER`)
```java
case FIRST_OBJECT_CONSTRUCTOR_QUALIFIER:
case SECOND_OBJECT_CONSTRUCTOR_QUALIFIER:
       hasMatch = nextToken.kind == SyntaxKind.CLIENT_KEYWORD || 
                  nextToken.kind == SyntaxKind.ISOLATED_KEYWORD;
       break;
```
In 2 qualifier case: we check for 2 alternative qualifier paths.
In 4 qualifier case: we check for 4 alternative qualifier paths.

or do we need to worry about this? @SupunS @rdulmina

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -3966,38 +3996,65 @@ private STNode parseTerminalExpression(STNode annots, boolean isRhsExpr, boolean
...
+            case XML_KEYWORD:
+            case STRING_KEYWORD:
+            case NEW_KEYWORD:
+            case FLUSH_KEYWORD:
+            case LEFT_ARROW_TOKEN:
+            case WAIT_KEYWORD:
+            case COMMIT_KEYWORD:
+            case TRANSACTIONAL_KEYWORD:
+            case BASE16_KEYWORD:
+            case BASE64_KEYWORD:
```
SupunS, line 4042: Can we use `isValidExprStart` method here?

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
SupunS: I think second option is a viable and a simple solution.

rdulmina: +1

SupunS (approved):

PR #26976, opened 2020-11-17
Implement find all references API 

### LINKED ISSUE #25364: Add an API to find all references of a given symbol
**Description:**

Input is a cursor position/ text position

pubudu91:
## Purpose
This PR adds an API to the semantic API for looking up all the references of a given symbol.

Fixes #25364 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


nadeeshaan (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -149,6 +151,25 @@ public BallerinaSemanticModel(BLangPackage bLangPackage, CompilerContext context
         return compiledSymbols;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<Location> references(String fileName, LinePosition position) {
```
rasika, line 158: Use of references across the modules is a common case. Have we thought of how user would query such references across modules? How one could query a single symbol across multiple semantic models of the modules. Is it possible to allow query `references()` through a Symbol as input?

nadeeshaan: I think @rasika 's concern is a valid requirement. There is a catch though, which is the performance. With the symbol being an input then we query the symbol API and then again query the references API.

pubudu91: This is what we talked about having an integration layer between the project API and the semantic API for right? Can definitely add another API which accepts a symbol as the input. In fact, I tried that approach first, and then reverted to position. From a performance POV, I don't think it'll be an issue since we don't need to query the symbol API again.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -235,7 +256,8 @@ private BLangCompilationUnit getCompilationUnit(String srcFile) {
     }
 
     private boolean isTypeSymbol(BSymbol symbol) {
-        return symbol instanceof BTypeSymbol && !(symbol instanceof BPackageSymbol);
+        return symbol instanceof BTypeSymbol && !(symbol instanceof BPackageSymbol)
+                && !(symbol instanceof BAnnotationSymbol);
```
SupunS, line 260: We could use symTag to check for `BPackageSymbol` and `BAnnotationSymbol`

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/SemanticModel.java
```diff
@@ -58,6 +59,16 @@
...
 
+    /**
+     * If there's an identifier associated with a symbol at the specified cursor position, finds all the references of
+     * the specified symbol within the relevant scope.
+     *
+     * @param fileName name of the file in which to look up the specified position
+     * @param position a cursor position in the source
+     * @return A {@link List} of line ranges of all the references
+     */
+    List<Location> references(String fileName, LinePosition position);
```
SupunS, line 70: I feel accepting a symbol as a parameter would make more sense in-terms of the API's perspective..
Then it reads as "give me the references of this symbol"

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
```diff
@@ -108,6 +110,10 @@ public boolean isLangLib() {
         return isLangLib;
     }
 
+    public BSymbol getInternalSymbol() {
```
SupunS, line 113: unused method?

SupunS (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/SemanticModel.java
pubudu91: Added an overloaded version of this which takes a `Symbol`.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
pubudu91: In use, after the addition of the above mentioned API.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/SemanticModel.java
```diff
@@ -58,6 +59,25 @@
...
     List<Symbol> moduleLevelSymbols();
 
+    /**
+     * Finds all the references of the specified symbol within the relevant scope.
+     *
+     * @param fileName name of the file in which to look up the specified position
+     * @param symbol   a {@link Symbol} insance
+     * @return A {@link List} of line ranges of all the references
+     */
+    List<Location> references(String fileName, Symbol symbol);
```
SupunS, line 69: Do we need the file name here?

pubudu91: Don't need it. Removed with 345a9e0

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
rasika: Since now we allow `Symbol` as input, i think rest of the query can be possible with the introduction of the integration layer.

SupunS (approved):

nadeeshaan (approved):

PR #26992, opened 2020-11-18
Set detail map as immutable

### LINKED ISSUE #26545: Return value of `lang.error.detail` should be immutable
**Description:**
According to the spec, the return value of `lang.error.detail` should be immutable.
https://github.com/ballerina-platform/ballerina-spec/blob/1b6cc59c5cd52375ae6ba4747c8a7716f30ccf5a/lang/lib/error.bal#L50

**Affected Versions:**
Swan Lake preview-5

chiranSachintha:
Fixes #26545 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/PredefinedTypes.java
```diff
@@ -164,9 +164,10 @@
     public static final ServiceType TYPE_ANY_SERVICE = new BServiceType(TypeConstants.SERVICE, EMPTY_MODULE, 0);
     public static final HandleType TYPE_HANDLE = new BHandleType(TypeConstants.HANDLE_TNAME, EMPTY_MODULE);
     public static final UnionType ANYDATA_OR_READONLY = new BUnionType(Arrays.asList(TYPE_ANYDATA, TYPE_READONLY));
-    public static final MapType TYPE_ERROR_DETAIL = new BMapType(TypeConstants.MAP_TNAME, ANYDATA_OR_READONLY,
-                                                                 EMPTY_MODULE);
-    public static final ErrorType TYPE_ERROR = new BErrorType(TypeConstants.ERROR, EMPTY_MODULE, TYPE_ERROR_DETAIL);
+    public static final Type TYPE_ERROR_DETAIL = ReadOnlyUtils.setImmutableTypeAndGetEffectiveType(
+                                            new BMapType(TypeConstants.MAP_TNAME, ANYDATA_OR_READONLY, EMPTY_MODULE));
+    public static final ErrorType TYPE_ERROR = new BErrorType(TypeConstants.ERROR, EMPTY_MODULE,
+                                            new BMapType(TypeConstants.MAP_TNAME, ANYDATA_OR_READONLY, EMPTY_MODULE));
```
MaryamZi, line 170: Can't we use `TYPE_ERROR_DETAIL` here? This also has to be immutable, right?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/MapValueImpl.java
```diff
@@ -271,6 +271,9 @@ public V put(K key, V value) {
 
     protected void populateInitialValues(BMapInitialValueEntry[] initialValues) {
         for (BMapInitialValueEntry initialValue : initialValues) {
+            if (initialValue == null) {
```
MaryamZi, line 274: IMO we shouldn't do this. We should create and pass the initialValues array only if the values are present.

chiranSachintha: I have made a such change because there are some  cases(reale to [this](https://github.com/ballerina-platform/ballerina-lang/blob/d042e7043f64f09138932557b2b7be435000542f/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/ErrorUtils.java#L54)) which only the detail become null. Is there any other option to handle this ?

MaryamZi: I think we should refactor that code to create an array of the exact length depending on how many entries there are, instead of always creating an array of length 2.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/PredefinedTypes.java
chiranSachintha: In here we set error values as immutable. So if we change `error` type as immutable we can't do below kind of casting.
```ballerina
type CustomErrorData record {
    string message?;
    error cause?;
    int accountID?;
};
type CustomError1 distinct error<CustomErrorData>;
type CustomError distinct error<CustomErrorData>;

type Error CustomError | CustomError1;

class Foo {
    public isolated function next() returns record {| int value; |}|Error? {

        CustomError e = CustomError("CustomError", message = "custom error occured", accountID = 2);
        return e;
        
    }
}

public function main() {
    Foo val = new;
    object {
        public isolated function next() returns record {|int value;|}|error?;} err = <object {
         public isolated function next() returns record {|int value;|}|error?;}> val;
}
```

MaryamZi: All error values/types and their detail values/types have to be immutable. If using the immutable detail here causes something to fail, it's possible that we're creating an error with a non-immutable detail somewhere, which is wrong. Can you check why exactly it fails?

chiranSachintha: When creating `CustomError1` we are not set detail(`CustomErrorData`) as immutable. Is it effect when creating values from that type( `CustomErrorData`) If we set it to `immutable`.

MaryamZi: That's wrong, right? When the error value is being created, cloneReadOnly should get called for the detail, so at runtime the detail type should be a subtype of `readonly`?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/MapValueImpl.java
chiranSachintha: Fixed.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/PredefinedTypes.java
MaryamZi: As discussed offline, let's have the mutable detail type here for the time-being because that's how we've modeled the errors defined in the source too.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/PredefinedTypes.java
```diff
@@ -164,9 +164,10 @@
     public static final ServiceType TYPE_ANY_SERVICE = new BServiceType(TypeConstants.SERVICE, EMPTY_MODULE, 0);
     public static final HandleType TYPE_HANDLE = new BHandleType(TypeConstants.HANDLE_TNAME, EMPTY_MODULE);
     public static final UnionType ANYDATA_OR_READONLY = new BUnionType(Arrays.asList(TYPE_ANYDATA, TYPE_READONLY));
-    public static final MapType TYPE_ERROR_DETAIL = new BMapType(TypeConstants.MAP_TNAME, ANYDATA_OR_READONLY,
-                                                                 EMPTY_MODULE);
-    public static final ErrorType TYPE_ERROR = new BErrorType(TypeConstants.ERROR, EMPTY_MODULE, TYPE_ERROR_DETAIL);
+    public static final Type TYPE_ERROR_DETAIL = ReadOnlyUtils.setImmutableTypeAndGetEffectiveType(
+                                            new BMapType(TypeConstants.MAP_TNAME, ANYDATA_OR_READONLY, EMPTY_MODULE));
```
MaryamZi, line 168: Shall we define a private final variable for new `BMapType(TypeConstants.MAP_TNAME, ANYDATA_OR_READONLY, EMPTY_MODULE)` and use that in these two types?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/ErrorUtils.java
```diff
@@ -50,14 +51,28 @@
...
     public static ErrorValue createInteropError(Throwable e) {
-        BMap<BString, Object> detailMap = new MapValueImpl<>(PredefinedTypes.TYPE_ERROR_DETAIL);
-        if (e.getMessage() != null) {
-            detailMap.put(ERROR_MESSAGE_FIELD, StringUtils.fromString(e.getMessage()));
-        }
-        if (e.getCause() != null) {
-            detailMap.put(ERROR_CAUSE_FIELD, createError(StringUtils.fromString(e.getCause().getClass().getName()),
-                                                         StringUtils.fromString(e.getCause().getMessage())));
+        MappingInitialValueEntry[] initialValues;
+        if (e.getMessage() != null && e.getCause() != null) {
```
MaryamZi, line 55: Shall we extract `e.getMessage()` and `e.getCause()` out to variables?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/creators/ErrorCreator.java
```diff
@@ -105,11 +101,11 @@ public static BError createError(Type type, BString message, BError cause, BMap<
      * @return new error
      */
     public static BError createError(Type type, BString message, BString details) {
-        MapValueImpl<BString, Object> detailMap =
-                new MapValueImpl<>(setImmutableTypeAndGetEffectiveType(PredefinedTypes.TYPE_ERROR_DETAIL));
+        MappingInitialValueEntry[] initialValues = new MappingInitialValueEntry[1];
```
MaryamZi, line 104: Wouldn't this lead to the same issue where the entry becomes null?

chiranSachintha: In [here](https://github.com/ballerina-platform/ballerina-lang/blob/d042e7043f64f09138932557b2b7be435000542f/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/ErrorUtils.java#L54) for java exceptions like UnsupportedOperationException, the detail message become null.  But when considering the ballerina error values I think detail cannot become null. I kept the null check for detail value as it was in that manner previously

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/ErrorUtils.java
chiranSachintha: Fixed

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/PredefinedTypes.java
chiranSachintha: Fixed.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/creators/ErrorCreator.java
MaryamZi: Wouldn't the iteration fail in io.ballerina.runtime.internal.values.MapValueImpl#populateInitialValues?

chiranSachintha: It can happen. I have changed this in above mentioned way.

MaryamZi (approved):

PR #27045, opened 2020-11-20
Fix zero-based error line and column numbers printing when compiling

### LINKED ISSUE #27043: Wrong line  and column numbers in compile errors in cli
**Description:**

**Steps to reproduce:**
Please see the below code
```ballerina
public function main() {
   string p = 1234;
}
````

The line numbers and column numbers are zero-based
```
Compiling source
        nn2.bal
ERROR [nn2.bal:(1:14,1:18)] incompatible types: expected 'string', found 'int'
```

**Affected Versions:**
SLP6

@dulajdilshan

### PR DESCRIPTION
**Pull request by dulajdilshan**
## Purpose
$title

Fixes #27043 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
@@ -24,6 +24,9 @@ import io.ballerina.projects.JdkVersion;
 import io.ballerina.projects.PackageCompilation;
 import io.ballerina.projects.Project;
 import io.ballerina.projects.directory.SingleFileProject;
+import io.ballerina.tools.diagnostics.Diagnostic;
+import io.ballerina.tools.text.LinePosition;
+import io.ballerina.tools.text.LineRange;
 
 import java.io.PrintStream;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by hasithaa — APPROVED** [review]

File: cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java [hunk 1]
```diff
@@ -63,9 +66,21 @@ public void execute(Project project) {
...
+    private String convertDiagnosticToString(Diagnostic diagnostic) {
+        LineRange lineRange = diagnostic.location().lineRange();
+
+        LineRange oneBasedLineRange = LineRange.from(
+                lineRange.filePath(),
+                LinePosition.from(lineRange.startLine().line() + 1, lineRange.startLine().offset() + 1),
+                LinePosition.from(lineRange.endLine().line() + 1, lineRange.endLine().offset() + 1));
+
+        return diagnostic.diagnosticInfo().severity().toString() + " [" +
+                oneBasedLineRange.filePath() + ":" + oneBasedLineRange + "] " + diagnostic.message();
```
pubudu91, line 84: Is this the best way to handle this? :thinking: My concern here is that the `toString()` of the Diagnostic impl is duplicated here. How about creating a copy of the whole Diagnostic instance? Maybe add a builder API to `Diagnostic` to make it easier to create copies of Diagnostic instances?

pubudu91: We can go with this for now though, and maybe revisit later.

dulajdilshan: IMO, creating a new diagnostic kinda going out of its definition, isn't it? Here, We only need to change the position and print right ?

pubudu91: Yeah, converting the diagnostic to a 1-based position. My concern is this feels like a rather ad hoc solution and duplicates the `toString()` of the diagnostic implementation. My point was whether we should consider having a proper util/API for converting a given diagnostic to 1-based positions. 

But I guess we can go with this for now and revisit this later.

pubudu91 (approved):

SupunS (approved):

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
dulajdilshan: Cool

Issue #26632, opened 2020-10-27
ClassCastException occurs when converting a JSON to a record type.


ThisaruGuruge:
**Description:**
There is a `ClassCastException` throwing when trying to build a record from a JSON using the `cloneWithType()` function.
Following is the error message:

```
java.lang.ClassCastException {"message":"class io.ballerina.runtime.values.DecimalValue cannot be cast to class java.math.BigDecimal (io.ballerina.runtime.values.DecimalValue is in unnamed module of loader 'app'; java.math.BigDecimal is in module java.base of loader 'bootstrap')"}
                at ballerina.lang_value.1_0_0:fromJsonWithType(value.bal:205)
```

**Steps to reproduce:**

Define the following record types:

```ballerina
public enum ArgumentType {
    INLINE,
    VARIABLE
}

public type Scalar int|string|float|boolean;

public type Location record {|
    int line;
    int column;
|};

public type Argument record {|
    string name;
    Scalar value;
    ArgumentType 'type = INLINE;
    Location nameLocation;
    Location valueLocation;
|};

public type Field record {|
    string name;
    Argument[] arguments?;
    Field[] selections?;
    Location location;
|};

public type Operation record {|
    string name;
    OperationType 'type;
    Field[] fields?;
|};

public type Document record {|
    map<Operation> operations;
|};
```

Use the following JSON to create the `Document` record.

```JSON
{
    "operations": {
        "operation": {
            "name": "getData",
            "type": "query",
            "fields": [
                {
                    "name": "name",
                    "arguments": [
                        {
                            "name": "id",
                            "value": 132,
                            "type": "INLINE",
                            "nameLocation": {
                                "line": 8,
                                "column": 10
                            },
                            "valueLocation": {
                                "line": 8,
                                "column": 14
                            }
                        },
                        {
                            "name": "name",
                            "value": "Prof. Moriarty",
                            "type": "INLINE",
                            "nameLocation": {
                                "line": 8,
                                "column": 19
                            },
                            "valueLocation": {
                                "line": 8,
                                "column": 25
                            }
                        },
                        {
                            "name": "negative",
                            "value": -123,
                            "type": "INLINE",
                            "nameLocation": {
                                "line": 8,
                                "column": 43
                            },
                            "valueLocation": {
                                "line": 8,
                                "column": 53
                            }
                        },
                        {
                            "name": "weight",
                            "value": 75.4,
                            "type": "INLINE",
                            "nameLocation": {
[...]

PR #27100, opened 2020-11-23
Fix TypeConverter error

### LINKED ISSUE #27549: Error in using `cloneWithType` to union targets
**Description:**

When using cloneWithType to convert to union targets, when the actual type of the source value is a member of the union, and there is another valid type as well, the current implementation cannot proceed.

**Steps to reproduce:**

Please consider the following scenario

```public type Scalar int|string|float|boolean;

public type Argument record {|
    Scalar value;
|};

public function testConvertToUnionWithActualType() {
    json expectedJson = {"value": 132};

    Argument expected = {"value": 132};
    var actual = expectedJson.cloneWithType(Argument);
}
```
**Affected Versions:**

SLP7

MacOS BigSur

dulvinw:
When a converting to a record, if a union type is present and the value was convertible to multiple options of the union type converter assumed this is an error.

Also added test cases to check json with decimal using `fromJsonWithType` langlib function

Fixes #27549

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
@@ -355,7 +355,7 @@ public class TypeConverter {
             } else {
                 if (targetFieldTypes.containsKey(fieldName)) {
                     if (getConvertibleTypes(valueEntry.getValue(), targetFieldTypes.get(fieldName),
-                            unresolvedValues).size() != 1) {
+                            unresolvedValues).size() == 0) {
                         return false;
                     }
                 } else if (!targetType.sealed) {
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
@@ -245,6 +245,9 @@ public class TypeConverter {
         switch (targetTypeTag) {
             case TypeTags.UNION_TAG:
                 for (Type memType : ((BUnionType) targetType).getMemberTypes()) {
+                    if (checkIsType(inputValue, memType)) {
+                        return List.of(memType);
+                    }
                     convertibleTypes.addAll(getConvertibleTypes(inputValue, memType, unresolvedValues));
                 }
                 break;
@@ -265,6 +268,15 @@ public class TypeConverter {
                 }
         }
         return convertibleTypes;
+    }
+
+    private static boolean checkIsType(Object value, Type targetType) {
+        Type valueType = TypeChecker.getType(value);
+        if (valueType.equals(targetType)) {
+            return true;
+        }
+
+        return false;
     }
 
     public static List<Type> getConvertibleTypesFromJson(Object value, Type targetType,
@@ -355,7 +367,7 @@ public class TypeConverter {
             } else {
                 if (targetFieldTypes.containsKey(fieldName)) {
                     if (getConvertibleTypes(valueEntry.getValue(), targetFieldTypes.get(fieldName),
-                            unresolvedValues).size() == 0) {
+                            unresolvedValues).size() != 1) {
                         return false;
                     }
                 } else if (!targetType.sealed) {
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
@@ -245,7 +245,7 @@ public class TypeConverter {
         switch (targetTypeTag) {
             case TypeTags.UNION_TAG:
                 for (Type memType : ((BUnionType) targetType).getMemberTypes()) {
-                    if (checkIsType(inputValue, memType)) {
+                    if (TypeChecker.getType(inputValue) == memType) {
                         return List.of(memType);
                     }
                     convertibleTypes.addAll(getConvertibleTypes(inputValue, memType, unresolvedValues));
@@ -268,15 +268,6 @@ public class TypeConverter {
                 }
         }
         return convertibleTypes;
-    }
-
-    private static boolean checkIsType(Object value, Type targetType) {
-        Type valueType = TypeChecker.getType(value);
-        if (valueType.equals(targetType)) {
-            return true;
-        }
-
-        return false;
     }
 
     public static List<Type> getConvertibleTypesFromJson(Object value, Type targetType,
--- a/langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
+++ b/langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
@@ -1281,8 +1281,8 @@ type OpenRecordWithUnionTarget record {|
 
 function tesFromJsonWithTypeMapWithDecimal() {
     map<json> mp = {
-            name: "foo",
-            factor: 1.23d
+        name: "foo",
+        factor: 1.23d
     };
     var or = mp.fromJsonWithType(OpenRecordWithUnionTarget);
 
@@ -1293,4 +1293,22 @@ function tesFromJsonWithTypeMapWithDecimal() {
     OpenRecordWithUnionTarget castedValue = <OpenRecordWithUnionTarget>or;
     assertEquality(castedValue["factor"], mp["factor"]);
     assertEquality(castedValue["name"], mp["name"]);
+}
+
+public type Maps record {|int i; int...;|}|record {|int i?;|};
+
+public type Value record {|
+    Maps value;
+|};
+
+public function testConvertJsonToAmbiguousType() {
+    json j = {"value": <map<int>> {i: 1}};
+    Value|error res = j.cloneWithType(Value);
+
+    if res is error {
+        assertEquality("'map<json>' value cannot be converted to 'Value'", res.detail()["message"]);
+        return;
+    }
+
+    panic error("Invalid respone.", message = "Expected error");
 }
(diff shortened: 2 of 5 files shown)


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/conversion/NativeConversionWithStampTypesTest.java
```diff
@@ -133,4 +133,9 @@ public void testConvertMapJsonWithDecimalToOpenRecord() {
     public void testConvertMapJsonWithDecimalUnionTarget() {
         BRunUtil.invoke(compileResult, "testConvertMapJsonWithDecimalUnionTarget");
     }
+
+    @Test
```
MaryamZi, line 137: Shall we move these tests to org.ballerinalang.langlib.test.LangLibValueTest#testFromJsonWithType?

dulvinw: Fixed.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/conversion/native-conversion-stampable-values.bal
```diff
@@ -104,6 +104,25 @@ function testConvertMapJsonWithDecimalUnionTarget() {
...
 }
 
+public type Scalar int|string|float|boolean;
+
+public type Argument record {|
+    Scalar value;
+|};
+
+public function testConvertUnion() {
+    json expectedJson = { "value": 132 };
```
MaryamZi, line 114:
```suggestion
    json expectedJson = {"value": 132};
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/conversion/native-conversion-stampable-values.bal
```diff
@@ -104,6 +104,25 @@ function testConvertMapJsonWithDecimalUnionTarget() {
...
+
+    Argument expected = { "value": 132 };
```
MaryamZi, line 116:
```suggestion
    Argument expected = {"value": 132};
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/conversion/native-conversion-stampable-values.bal
```diff
@@ -104,6 +104,25 @@ function testConvertMapJsonWithDecimalUnionTarget() {
...
+    var actual = expectedJson.cloneWithType(Argument);
+
+    if (actual is error) {
+        panic error("`cloneWithType` returned an error.");
+    } else {
+        assert(actual, expected);
+    }
```
MaryamZi, line 123:
```suggestion
    }
    assert(actual, expected);
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/conversion/native-conversion-stampable-values.bal
```diff
@@ -78,7 +78,7 @@ type OpenRecordWithUnionTarget record {|
 map<json> mp = {
         name: "foo",
         factor: 1.23d
```
MaryamZi, line 80:
```suggestion
    name: "foo",
    factor: 1.23d
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/conversion/NativeConversionWithStampTypesTest.java
```diff
@@ -133,4 +133,9 @@ public void testConvertMapJsonWithDecimalToOpenRecord() {
...
+    void testConvertUnion() {
```
MaryamZi, line 138: Shouldn't the name reflect that it's conversion with numeric types?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/conversion/native-conversion-stampable-values.bal
```diff
@@ -104,6 +104,25 @@ function testConvertMapJsonWithDecimalUnionTarget() {
...
+
+public type Argument record {|
+    Scalar value;
+|};
+
+public function testConvertUnion() {
+    json expectedJson = { "value": 132 };
+
+    Argument expected = { "value": 132 };
+    var actual = expectedJson.cloneWithType(Argument);
```
MaryamZi, line 117: Were we able to reproduce the original CCE mentioned in https://github.com/ballerina-platform/ballerina-lang/issues/26632? Both this example and the one in the issue seem to result in "ballerina/lang.typedesc}ConversionError" errors on slp7.

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -1245,3 +1245,23 @@ function testEnsureTypeNegative() {
...
 }
+
+type OpenRecordWithUnionTarget record {|
+    string|decimal...;
+|};
+
+function tesFromJsonWithTypeMapWithDecimal() {
+    map<json> mp = {
+            name: "foo",
+            factor: 1.23d
```
MaryamZi, line 1256:
```suggestion
        name: "foo",
        factor: 1.23d
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -355,7 +355,7 @@ private static boolean isConvertibleToRecordType(Object sourceValue, BRecordType
             } else {
                 if (targetFieldTypes.containsKey(fieldName)) {
                     if (getConvertibleTypes(valueEntry.getValue(), targetFieldTypes.get(fieldName),
-                            unresolvedValues).size() != 1) {
+                            unresolvedValues).size() == 0) {
```
MaryamZi, line 358: Don't we still have to check for > 1 convertible types? That should also be an error, right? Let's add a test for the same.

The following should fail.

```ballerina
public type Maps record {|int i; int...;|}|record {|int i?;|};

public type Value record {|
    Maps value;
|};

public function main() {
    json j = {"value": <map<int>> {i: 1}};
    Value res = checkpanic j.cloneWithType(Value);
}
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/conversion/native-conversion-stampable-values.bal
dulvinw: CCE is not reproducible. Found a different bug and fixed it.

MaryamZi: I tried the same on slp3 also but still couldn't reproduce the CCE. Fails with the following which I believe has/is been fixed.

```cmd
error: {ballerina/lang.typedesc}ConversionError message='map<json>' value cannot be converted to 'Argument'
        at ballerina.lang_value.1_0_0:cloneWithType(value.bal:67)
           xyz:main(xyz.bal:174)
```

But since the CCE has been observed shall we keep that issue open to try and reproduce the CCE and fix it, and create a new issue for what we're fixing here? Let's also add a comment on the issue to say the CCE is not reproducible with that code.

dulvinw: +1. Will do that

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -267,6 +270,15 @@ static boolean isConvertibleToFloatingPointTypes(Object value) {
         return convertibleTypes;
     }
 
+    private static boolean checkIsType(Object value, Type targetType) {
+        Type valueType = TypeChecker.getType(value);
+        if (valueType.equals(targetType)) {
```
MaryamZi, line 275: In L248 shouldn't we directly check `io.ballerina.runtime.internal.TypeChecker#checkIsType(java.lang.Object, io.ballerina.runtime.api.types.Type)` instead? 

Checking `.equals()` may not work for subtypes, right?

dulvinw: For anything which is not the exact type shouldn't we resort to getConvertibleTypes method?

MaryamZi: Yeah, I guess we should look for an exact match, or we won't be able to widen the type using `cloneWithType`. This is OK.

MaryamZi: We should ideally be able to rely on `==` here though, right?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -267,6 +270,15 @@ static boolean isConvertibleToFloatingPointTypes(Object value) {
         return convertibleTypes;
     }
 
+    private static boolean checkIsType(Object value, Type targetType) {
```
MaryamZi, line 273: Shall we call this `checkIsSameType` instead? Because `checkIsType` means something else in `TypeChecker`?

dulvinw: +1. Will do

MaryamZi: Should we just move this in-line?

```
TypeChecker.getType(value) == targetType
```

MaryamZi (approved):

PR #27165, opened 2020-11-25
Refactor the model and behaviour for constants

pubudu91:
## Purpose
This PR addresses some of the issues that were there with constants.
- `ConstantSymbol` is now a type symbol as well. 
- When used in type defs, `symbol()` will return the constant symbol
- The type of a constant is the singleton type of its value space.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/ConstantSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/ConstantSymbol.java
@@ -22,7 +22,7 @@ package io.ballerina.compiler.api.symbols;
  *
  * @since 2.0.0
  */
-public interface ConstantSymbol extends VariableSymbol {
+public interface ConstantSymbol extends VariableSymbol, SingletonTypeSymbol {
 
     /**
      * Get the constant value.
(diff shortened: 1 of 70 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/ConstantSymbol.java
```diff
@@ -22,12 +22,19 @@
...
      * @return {@link Object} value of the constant
      */
     Object constValue();
+
+    /**
+     * Gets the broader type of constant expression associated with this symbol.
+     *
+     * @return {@link TypeSymbol} of the broader type
+     */
+    TypeSymbol broaderTypeDescriptor();
```
nadeeshaan, line 39: What if we return an Optional here?

nadeeshaan (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/ConstantSymbol.java
pubudu91: From a language POV, no reason for broader type of a constant expr to be optional right? If there's a constant, it also has a broader type. It may become `null` for incomplete constant declarations?

rasika: Shouldn't incomplete constant declarations is a expected case?

pubudu91: Yes, but the APIs for the most part designed to be in-line with the spec. Optionals were used in places where the spec has specified it's optional. e.g., rest param in function type

SupunS (approved):

PR #27189, opened 2020-11-26
Add missing type descriptors

### LINKED ISSUE #26459: Missing representation in semantic API for the internal  "semantic error" type
The compiler, we have a separate type created for semantic errors. If an expression is semantically invalid, we set its type to semantic error. In the current semantic API, there isn't a corresponding construct. When semantic errors are encountered when building types, it's simply ignored and `null` returned as the type desc. This causes issues for the API user since the user simply expects the type to be there. Therefore, we need to decide how we want to handle semantic errors in the public API.

### LINKED ISSUE #27135: [Semantic API] Add support for intersection, distinct and never types
$subject. The above mentioned types are currently missing in the API.

### LINKED ISSUE #27269: [Semantic API] Model builtin sub types using individual type symbols
Curently we have modeled builtin subtypes using the same type symbols as their broader types. Instead, need to model them using individual type symbols for the sub types.

pubudu91:
## Purpose
This PR,
- adds support for `never` and intersection types. 
- improves the support for distinct types. 
- adds a type symbol for the internal semantic error type: `CompilationErrorTypeSymbol`. The text representation of this is `$CompilationError$`
- adds separate type symbols for each of the simple types and builtin subtypes:
    - `IntUnsigned8TypeSymbol`
    - `IntSigned8TypeSymbol`
    - `IntUnsigned16TypeSymbol`
    - `IntSigned16TypeSymbol`
    - `IntUnsigned32TypeSymbol`
    - `IntSigned32TypeSymbol`
 
    - `XMLElementTypeSymbol`
    - `XMLProcessingInstructionTypeSymbol`
    - `XMLCommentTypeSymbol`
    - `XMLTextTypeSymbol`
 
    - `CharTypeSymbol`

Fix #27135 
Fix #26459
Fix #27269 

## TODO
- Add typedesc for `byte`

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


nadeeshaan (approved):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaCharTypeSymbol.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Represents the string:Char type descriptor.
+ *
+ * @since 2.0.0
+ */
+public class BallerinaCharTypeSymbol extends AbstractTypeSymbol implements CharTypeSymbol {
+
+    public BallerinaCharTypeSymbol(CompilerContext context, ModuleID moduleID, BStringSubType charType) {
+        super(context, TypeDescKind.STRING, moduleID, charType);
```
nadeeshaan, line 34: will it be possible to assign CHAR kind?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaIntSigned16TypeSymbol.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Represents the int:Signed16 type descriptor.
+ *
+ * @since 2.0.0
+ */
+public class BallerinaIntSigned16TypeSymbol extends AbstractTypeSymbol implements IntSigned16TypeSymbol {
+
+    public BallerinaIntSigned16TypeSymbol(CompilerContext context, ModuleID moduleID, BIntSubType signed16Type) {
+        super(context, TypeDescKind.INT, moduleID, signed16Type);
```
nadeeshaan, line 35: Same as the Char type, will it be possible to have a subtype kind? Otherwise in order to distinguish we will have to have an instance of check

nadeeshaan (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaIntSigned16TypeSymbol.java
pubudu91: We can include. The downside would be if we want to check if the type is an integer kind, we'll have to consider all the types and/or provide a util to check if the type is an integer kind

nadeeshaan: My suggestion is, that we keep a separate API for the subtype. We have to consider this when we order the suggestions with the priority

pubudu91: Added the discussed APIs to the respective type symbols.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaTypeSymbol.java
```diff
@@ -17,29 +17,23 @@
 package io.ballerina.compiler.api.impl.symbols;
 
 import io.ballerina.compiler.api.ModuleID;
-import io.ballerina.compiler.api.symbols.SimpleTypeSymbol;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BType;
 import org.wso2.ballerinalang.compiler.util.CompilerContext;
 
 /**
- * Represents the built-in simple type descriptor.
+ * Represents type descriptors which do not fall in to any other type descriptor.
```
SupunS, line 24: What are the possible types that could fall on to this group?

SupunS (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/StringTypeSymbol.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public interface StringTypeSymbol extends TypeSymbol {
+
+    /**
+     * Retrieves the string kind represented by the type symbol. Intended to be used when one needs to differentiate
+     * between the different kinds of string subtypes.
+     *
+     * @return The kind of string represented
+     */
+    Kind stringKind();
```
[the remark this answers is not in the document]
pubudu91: What would the return type of `subtype()` be?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaTypeSymbol.java
pubudu91: With the changes introduced in this PR, this becomes a sort of a fall back type symbol for any typedescs we may have missed. Currently the only one I can think of is the "dependent type". Plan is to get rid of this altogether for alpha.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/StringTypeSymbol.java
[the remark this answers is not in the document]
SupunS: can we return a similar enum (a merged enum with all subtypes)?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaByteTypeSymbol.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public class BallerinaByteTypeSymbol extends AbstractTypeSymbol implements ByteTypeSymbol {
+
+    public BallerinaByteTypeSymbol(CompilerContext context, ModuleID moduleID, BType byteType) {
+        super(context, TypeDescKind.BYTE, moduleID, byteType);
+    }
+
+    @Override
+    public String name() {
+        return "byte";
```
SupunS, line 39: can we use `this.typeDescKind.getName()`? would be better than hardcoding

SupunS (approved):

PR #27307, opened 2020-12-02
Improve utils in Project API

### LINKED ISSUE #26945: Improve Project API to return valid source root paths for single file projects
**Description:**
Please note that the current project API returns a temporary directory path when trying to access the source root of a `SingleFileProject` instance. Therefore it can be improved so that the project API will return the file path of one of the following. 
- path of the single file
- path of the parent directory of the single file

**Steps to reproduce:**

Try accessing the source root of a single file via project API, as follows.
```ballerina
 String srcRoot = ProjectLoader.loadProject(Paths.get(<FILE_PATH>)).sourceRoot().toAbsolutePath().toString();

```

**Affected Versions:**

### LINKED ISSUE #27273: Need a project-api util function to get sourceRoot from the documentPath
**Description:**
Need a project-api util function to get **sourceRoot** from the **documentPath**. With the current APIs it always required to create a project instance to get the projectRoot. But we need to calculate sourceRoot **without creating a project instance** for example when maintaining a mapping of project instances against it's sourceRoot path.

```
{sourceRoot -> projectInstance}
```

**Steps to reproduce:**

**Affected Versions:**
2.0.0-SLP7

### LINKED ISSUE #27274: Need a project-api util function to get Document providing documentPath from Project
**Description:**
Need a project-api util function to get Document providing documentPath from Project.

**Affected Versions:**
2.0.0-SLP7

azinneera:
- Set Standalone Ballerina file path as `sourceRoot` in SingleFileProject. Fixes #26945
- Improve util methods in ProjectLoader. Fixes #27274, Fixes #27273
- Bind target to project

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


NipunaRanasinghe (commented):
Reviewed debugger related changes and LGTM. Found few places where we can further simply the debug server implementation with these fixes, so I'll send a separate PR once this is merged. +1 to proceed.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
```diff
@@ -95,31 +136,41 @@ public static Project loadProject(Path path, ProjectEnvironmentBuilder projectEn
      * @return documentId of the document
      */
 
-    public static Optional<DocumentId> getDocumentId(Path documentFilePath, BuildProject project) {
+    public static Document getDocument(Path documentFilePath, Project project) {
```
NipunaRanasinghe, line 139: IMO adding an overloaded API so that the users can retrieve the document instance just by passing its path might makes more sense from the users point of view. WDYT?

```
public static Document getDocument(Path documentFilePath) {
    Project project = loadProject(documentFilePath);
    return getDocument(documentFilePath, project);
}
```

NipunaRanasinghe (dismissed):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/Project.java
```diff
@@ -36,14 +38,18 @@
     private final BuildOptions buildOptions;
     private final ProjectEnvironment projectEnvironment;
     private final ProjectKind projectKind;
+    private final Target target;
 
     protected Project(ProjectKind projectKind,
                       Path projectPath,
-                      ProjectEnvironmentBuilder projectEnvironmentBuilder, BuildOptions buildOptions) {
+                      ProjectEnvironmentBuilder projectEnvironmentBuilder,
+                      Target target,
```
sameerajayasoma, line 46: It is not correct to expose a class in the internal package via the public API. Here `Target` is in the internal package.

sameerajayasoma, line 46: Passing the target path via this `Target` object is not the correct way IMO. We should allow the user to configure such things via the ProjectEnvironmentBuilder.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
```diff
@@ -87,6 +91,43 @@ public static Project loadProject(Path path, ProjectEnvironmentBuilder projectEn
         return SingleFileProject.load(projectEnvironmentBuilder, absProjectPath);
     }
 
+    public static Optional<Path> findProjectRoot(Path filepath) throws ProjectException, IllegalArgumentException {
```
sameerajayasoma, line 94: This API should not be part of the ProjectLoader utility

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
sameerajayasoma, line 139: This method should not be part of the ProjectLoader utility

compiler/ballerina-lang/src/main/java/org/ballerinalang/compiler/plugins/CompilerPlugin.java
```diff
@@ -137,9 +136,9 @@ default void process(ClassDefinition classDefinition, List<AnnotationAttachmentN
      * Notifies when the code generated phase is completed.
      *
      * @param project Project instance
-     * @param target Target directory
+     * @param executablePath path of the executable jar
      */
-    default void codeGenerated(Project project, Target target) {
+    default void codeGenerated(Project project, Path executablePath) {
```
sameerajayasoma, line 141: This is an API change. We need to communicate this change especially with the C2C team.

sameerajayasoma (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
rasika: I think the use case tried to solve here is to get a document out of an **existing** project instance. Hence, I suggest this should be moved to `ProjectUtils`. 

Or I suppose this should be a basic-functionality of an **Project**. 

```java
public interface Project {
    Optional<Document> getDocument(Path documentFilePath);
}
```

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
rasika: Can we move this into ProjectUtils ?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
```diff
@@ -95,31 +136,41 @@ public static Project loadProject(Path path, ProjectEnvironmentBuilder projectEn
      * @return documentId of the document
      */
 
-    public static Optional<DocumentId> getDocumentId(Path documentFilePath, BuildProject project) {
+    public static Document getDocument(Path documentFilePath, Project project) {
+        if (project.kind().equals(ProjectKind.SINGLE_FILE_PROJECT)) {
+            DocumentId documentId = project.currentPackage().getDefaultModule().documentIds().iterator().next();
+            return project.currentPackage().getDefaultModule().document(documentId);
+        }
+        if (!(project instanceof BuildProject)) {
```
rasika, line 144: Since this method is only supported for SINGLE_FILE_PROJECT and BUILD_FILE_PROJECT, we can have a top level check to throw 'unsupported project type' to avoid `instanceof` check

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
```diff
@@ -95,31 +136,41 @@ public static Project loadProject(Path path, ProjectEnvironmentBuilder projectEn
...
                     }
+                }
 
-                    for (DocumentId documentId : module.testDocumentIds()) {
-                        if (module.document(documentId).name().equals(ProjectConstants.TEST_DIR_NAME + "/"
-                                + Optional.of(documentFilePath.getFileName()).get().toString())) {
-                            return Optional.of(documentId);
-                        }
+                for (DocumentId documentId : module.testDocumentIds()) {
+                    if (module.document(documentId).name().equals(ProjectConstants.TEST_DIR_NAME + "/"
```
rasika, line 166: Can you verify whether use of '/' here is platform independent. If yes, shall we move that into a more convincing constant such as `TEST_FILE_SEPARATOR`

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
sameerajayasoma: How about creating a separate class called `ProjectPaths` and add all project path related utility APIs there? @azinneera @rasika

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
sameerajayasoma: We don't need this `instanceof` check. One should be using the `ProjectKind`

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
sameerajayasoma: AFAIU, the functionality of this method should be this: 
- First, check whether this path lies inside of a Ballerina project
- If true, return the root path of the project
- If false, return `Optional.empty()`;

if the above statement is correct, then the implementation should be simple IMO. Are there any other requirements here? @rasika @nadeeshaan @azinneera

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
sameerajayasoma: My gut feeling is that this method shouldn't be in the project API. Project API offers methods to retrieve modules by their names, and ids. Working with paths should not be part of the project API code.  

Can we keep a map of paths to document IDs?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/ProjectLoader.java
nadeeshaan: @sameerajayasoma yes, this should be the behaviour for both the single file and multi-module project. Also, I think it would be better if we can get this particular method out of the ProjectLoader.

rasika: @sameerajayasoma +1 for the util. Can return root path of the project for both single-file and build projects, `Optional.empty()` otherwise.

azinneera:
Closing this, since the relevant issues will be addressed separately by individual PRs

PR #27326, opened 2020-12-03
Align observability test-utils with new service type changes.

nadundesilva:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/observability-test-utils/src/main/ballerina/Ballerina.toml
```diff
@@ -4,7 +4,7 @@ name = "testobserve"
 version = "0.0.0"
 
 [[platform.java11.dependency]]
-path = "../libs/observability-test-utils-2.0.0-Preview7.jar"
+path = "../libs/observability-test-utils-2.0.0-Preview8-SNAPSHOT.jar"
```
rdhananjaya, line 7: do we need to commit these toml changes

tests/observability-test-utils/src/main/java/org/ballerina/testobserve/listenerendpoint/Resource.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public Resource(BObject serviceObject, ResourceFunctionType resourceFunctionType, String basePath) {
+        this.serviceObject = serviceObject;
+        this.resourceFunctionType = resourceFunctionType;
+        this.basePath = basePath;
+
+        // Validate resource
+        int paramCount = this.getParamTypes().length;
+        String resourcePath = this.getResourcePath();
+        if (paramCount > 2) {
```
rdhananjaya, line 46: What does this validation do? Is this specific to the test case?

tests/observability-test-utils/src/main/ballerina/Ballerina.toml
nadundesilva: Most of the lang modules and other modules seem to have been committed with the new version. I can revert if this shouldn't be done.

tests/observability-test-utils/src/main/java/org/ballerina/testobserve/listenerendpoint/Resource.java
nadundesilva: This is a validation for the mock listener. The listener at the moment, only supports resource functions and only accepts two arguments.

The validation is there as a way to throw an error if a developer doesn't adhere to the expected format.

rdhananjaya (approved):

Issue #26885, opened 2020-11-12
Re-enable  MappingConstructorExprTest.testVarNameField[8](testVarNameFieldInAnnotation) 


rdhananjaya:
**Description:**

Re-enable  MappingConstructorExprTest. testVarNameField[8](testVarNameFieldInAnnotation) in service-change branch after fixing #26838 

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Issue #26838, opened 2020-11-10
Object constructor does not capture local values in annotations


rdhananjaya:
**Description:**
```ballerina
import ballerina/io;

type Foo record {|
    string s;
|};


annotation Foo foo on class;

string s = "g";

public function main() {
    s = "local";
    object {} o = @foo{ s } object { };

    var t = typeof o;
    io:println(t.@foo);
}
```

Output: `{"s":"g"}`

Expected `{"s":"local"}`

If we refer a local variable in the annotation, it gives the error `undefined symbol`.

**Steps to reproduce:**

**Affected Versions:**
SLP5

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

rdhananjaya:
We should be able to re-enable disabled test case on this issue. https://github.com/ballerina-platform/ballerina-lang/issues/26885

rdhananjaya:
Fixed in  #27347

PR #27347, opened 2020-12-05
Add missing annotation support for service objects and improve listener

rdhananjaya:
## Purpose
Add back missing annotation support for services. 

And allow listener's `attach` method to control what to accept as service path.
Second param of listener.attach method can be any combination of `string[], string, ()`.

Fixes: https://github.com/ballerina-platform/ballerina-lang/issues/26885, https://github.com/ballerina-platform/ballerina-lang/issues/27348, and https://github.com/ballerina-platform/ballerina-lang/issues/26838

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/AnnotationUtils.java
```diff
@@ -43,12 +43,17 @@
...
      */
-    public static void processAnnotations(MapValue globalAnnotMap, Type bType) {
+    public static void processAnnotations(MapValue globalAnnotMap, Type bType, Strand strand) {
         if (!(bType instanceof BAnnotatableType)) {
             return;
         }
 
         BAnnotatableType type = (BAnnotatableType) bType;
+        if (type.getTag() == TypeTags.SERVICE_TAG) {
+            processServiceAnnotations(globalAnnotMap, (BServiceType) bType, strand);
```
MaryamZi, line 53: `processServiceAnnotations` used to be codegen-ed separately for services to specifically recompute the annotations in services that are created within a function (via https://github.com/ballerina-platform/ballerina-lang/blob/v2.0.0-Preview7/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java#L1625).

To continue to support the same, IMO we need to do something similar.

The second annotation access in the following sample used to return `{i: 202}`, but now returns `{i: 101}`.

```ballerina
annotation record {int i;} x on service;

int globalInt = 101;

public function getService() returns service object {} {
    return @x {i: globalInt} service object {
        
    };
}

public function main() {
    typedesc<service object {}> t = typeof getService();
    var v = t.@x; // {"i":101}

    globalInt = 202;
    t = typeof getService();
    v = t.@x; // used to be {"i":202}, now {"i":101}
}
```

Will we be fixing this anytime soon given that it's a breaking change?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3798,18 +3796,25 @@ private boolean checkAttachMethod(BAttachedFunction func) {
...
+
+        private boolean isServiceObject(BType type) {
+            if (type.tag != TypeTags.OBJECT) {
+                return false;
+            }
+
+            if (!types.isServiceObject((BObjectTypeSymbol) type.tsymbol)) {
+                return false;
+            }
+            return true;
```
MaryamZi, line 3817:
```suggestion
            return types.isServiceObject((BObjectTypeSymbol) type.tsymbol);
```

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -775,6 +775,22 @@ error.invalid.listener.var=\
...
   invalid listener attachment
 
+error.service.absolute.path.or.literal.required.by.listener=\
+  service absolute path or literal is required by listener
+
+error.service.path.literal.is.not.supported.by.listener=\
+  service path literal is not supported by listener
+
+error.service.absolute.path.is.not.supported.by.listener=\
+  service absolute path is not supported by listener
```
MaryamZi, line 785: Will these be supported in the future?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -775,6 +775,22 @@ error.invalid.listener.var=\
...
+
+error.service.path.literal.required.by.listener=\
+  service path literal is required by the listener
+
+error.service.absolute.path.required.by.listener=\
+  service absolute path is required by the listener
+
+
```
MaryamZi, line 793: Extra new line.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/AnnotationUtils.java
rdhananjaya: Could this be this issue, I mean this behavior https://github.com/ballerina-platform/ballerina-lang/issues/26838

compiler/ballerina-lang/src/main/resources/compiler.properties
rdhananjaya: No this is the expected behavior,

if listener attach method is `attach(service object s, string? path)`, then it's an error to have following:
```
service \abs\path on new http:Listener(4000) {

}
```

It should be  either `service "name-literal" on lsn { }` or `service on lsn { }`

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BServiceType.java
```diff
@@ -99,4 +93,8 @@ public void setResourceFunctions(ResourceFunctionType[] resourceFunctions) {
     public int getTag() {
         return TypeTags.SERVICE_TAG;
     }
+
+    public boolean isAnnotationsAvailable() {
```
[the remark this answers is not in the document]
rdhananjaya: will update

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/AnnotationUtils.java
rdhananjaya: Issue to re-visit this annot impl: https://github.com/ballerina-platform/ballerina-lang/issues/27348

I think we will have to consider object constructors for the above issue.

hasithaa (approved):

PR #27358, opened 2020-12-07
Improve Numeric values, Literal Strings  few bug fixes for Toml Parser

### LINKED ISSUE #27377: Need to support numeric values with exponents in TOML parser
**Description:**
$subject

As the TOML parser will be used for `configurable` variables, there is a requirement of parsing the exponent values such as 
- `5e+22`
- `1e6`
- `-2E-2`
- `6.626e-34`

referred : [https://toml.io/en/v0.4.0#float](https://toml.io/en/v0.4.0#float)

**Steps to reproduce:**

**Affected Versions:**
SLP7

#27129
#27131

Team/DevTools, Area/TOMLParser

### LINKED ISSUE #27378: Need to support literal string values in TOML parser
**Description:**
$subject

As the TOML parser will be used for `configurable` variables, It's good to support TOML literal strings that can include special characters witth no escaping. 

Eg. - 
```toml
path = '\\ServerX\admin$\system32\'
regexKey = '''I [dw]on't need \d{2} apples'''
```

referred :[https://toml.io/en/v0.4.0#string]( https://toml.io/en/v0.4.0#string)

**Steps to reproduce:**

**Affected Versions:**
slp7

#27129 

Team/DevTools, Area/TOMLParser

xlight05:
## Purpose
$Subject

## Features Introduced
- Literal String support Resolves #27378 
- Exponents Support for floats. Resolves #27377 
- Hex support for integers
- Octal support for integers
- Binary support for integers
- Underscore support for numerical values

## Bug Fixes

- Fix Empty String bug
- Fix Leading Zero bug in Integer bug in Integers and Floats
- Fix the issue with single character string. Fixes - Issue 1 of #27399 
- Fix the bug where compiler fails when theres newline in arrays  - Fixes Issue 2 of #27399 
- Resolves https://github.com/ballerina-platform/ballerina-lang/issues/27399
- Fix the Jar issue in the distribution


misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlLexer.java
```diff
@@ -570,7 +570,7 @@ private STToken processNumericLiteral(int startChar) {
             type = SyntaxKind.IDENTIFIER_LITERAL;
         }
         // Integer cannot have a leading zero
-        if (startChar == '0' && len > 1) {
+        if (startChar == '0' && len > 1 && type == SyntaxKind.DECIMAL_INT_TOKEN) {
```
SupunS, line 573: what would be the outcome for something like `004.12`?

SupunS (changes requested):

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -389,27 +399,42 @@ private STNode parseValue() {
...
+        if (isNumerticalLiteral(token)) {
             return consume();
         } else {
             recover(token, ParserRuleContext.NUMERICAL_LITERAL);
             return parseNumericalToken();
         }
     }
 
+    private boolean isNumerticalLiteral(STToken token) {
+        return token.kind == SyntaxKind.DECIMAL_INT_TOKEN ||
```
lochana-chathura, line 431: Since there are lot of cases let's use a switch case

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -389,27 +399,42 @@ private STNode parseValue() {
...
         STNode sign = parseSign();
         STNode token = parseNumericalToken();
-        SyntaxKind kind;
-        if (token.kind == SyntaxKind.DECIMAL_INT_TOKEN) {
-            kind = DEC_INT;
+        SyntaxKind kind = getNodeKindNumericNode(token.kind);
+        return STNodeFactory.createNumericLiteralNode(kind, sign, token);
+    }
+
+    private SyntaxKind getNodeKindNumericNode(SyntaxKind tokenKind) {
```
lochana-chathura, line 406: Shall we use something like getNumericLiteralKind as the name?
Le'ts use switch case here also

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -446,6 +461,11 @@ private STNode parseBoolean() {
     private STNode parseStringValue() {
         STNode startingDoubleQuote = parseDoubleQuoteToken(ParserRuleContext.STRING_START);
         STNode content = parseStringContent();
+        //Empty String
+        if (content.kind == SyntaxKind.DOUBLE_QUOTE_TOKEN) {
+            return STNodeFactory.createStringLiteralNode(startingDoubleQuote, STNodeFactory.createEmptyNode(), content);
+        }
```
lochana-chathura, line 467: we need to do this check inside the parseStringContent().

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -468,7 +488,7 @@ private STNode parseDoubleQuoteToken(ParserRuleContext ctx) {
 
     private STNode parseStringContent() {
         STToken nextToken = peek();
-        if (nextToken.kind == SyntaxKind.IDENTIFIER_LITERAL) {
+        if (nextToken.kind == SyntaxKind.IDENTIFIER_LITERAL || nextToken.kind == SyntaxKind.DOUBLE_QUOTE_TOKEN) {
```
lochana-chathura, line 491: if the syntaxkind is DOUBLE_QUOTE_TOKEN then return STNodeFactory.createEmptyNode();
Also need to fix recovery for this method with above change

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -549,6 +569,9 @@ private STNode parseValueEnd() {
             case CLOSE_BRACKET_TOKEN:
                 // null marks the end of values
                 return null;
+            case NEWLINE:
+                consume();
+                return parseValueEnd();
```
lochana-chathura, line 574: why we need this? what happens to the consumed token in L573?

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -389,27 +399,42 @@ private STNode parseValue() {
...
 
     private STNode parseNumericalToken() {
         STToken token = peek();
-        if (token.kind == SyntaxKind.DECIMAL_INT_TOKEN) {
-            return consume();
-        } else if (token.kind == SyntaxKind.DECIMAL_FLOAT_TOKEN) {
+        if (isNumerticalLiteral(token)) {
             return consume();
         } else {
             recover(token, ParserRuleContext.NUMERICAL_LITERAL);
```
lochana-chathura, line 425: Need to update the error handler for the above change. (when recovering with NUMERICAL_LITERAL ctx)

lochana-chathura (changes requested):

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
lochana-chathura: 1. if there's IDENTIFIER_LITERAL we consume it and return
2. if there's DOUBLE_QUOTE_TOKEN we simply create emptyNode and return (need to make the token optional in the Syntax API)
3. something is wrong. Hence go to recovery and re-parse

Similar case from BallerinaParser.java https://github.com/ballerina-platform/ballerina-lang/blob/7289b5b33d06787e4ebfa7159d6eab19e5545f80/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java#L6055-L6066

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlLexer.java
```diff
@@ -544,37 +546,161 @@ private STNode processComment() {
...
+                    // Integer part of the float cannot have a leading zero
+                    if (startChar == '0' && len > 1) {
+                        reportLexerError(DiagnosticErrorCode.ERROR_LEADING_ZEROS_IN_NUMERIC_LITERALS);
+                    }
+
+                    // Code would not reach here if the floating point starts with a dot
+                    return processDecimalFloatLiteral();
+                default:
+                    if (isAlphabeticChar(nextChar)) {
+                        return processKey();
```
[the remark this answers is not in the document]
xlight05: Yes.

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
xlight05: We only use newlines for validation. Ex - There should be a new line after each key value pair. 
But in few cases, in thats different. Here, we are ignoring the new line token passed to parser as that validation is not needed. Ex -  Arrays 
```toml
arr = [
"hello",
"hello1"
,
"hello2"
]
```

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlLexer.java
```diff
@@ -382,11 +384,8 @@ private STToken getLiteral(SyntaxKind kind) {
         return STNodeFactory.createLiteralValueToken(kind, lexeme, leadingTrivia, trailingTrivia);
     }
 
-    private STToken getDoubleQuoteToken(SyntaxKind kind) {
+    private STToken getQuoteToken(SyntaxKind kind) {
         STNode leadingTrivia = getLeadingTrivia();
-        // Trivia after the back-tick including whitespace belongs to the content of the back-tick.
-        // Therefore do not process trailing trivia for starting back-tick. We reach here only for
-        // starting back-tick. Ending back-tick is processed by the template mode.
```
lochana-chathura, line 389: Let's keep this comment update according this scenario. 
```
// Trivia after the single or double quote including whitespace belongs to the content inside the quotes.
// Therefore do not process trailing trivia for starting quote. We reach here only for
// starting single or double quote. Ending quote is processed by the LITERAL_STRING mode.
```

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -448,7 +466,7 @@ private STNode parseStringValue() {
         STNode content = parseStringContent();
         STNode endingDoubleQuote = parseDoubleQuoteToken(ParserRuleContext.STRING_END);
         return STNodeFactory.createStringLiteralNode(startingDoubleQuote, content, endingDoubleQuote);
-    }
+    }//TODO change optional
```
lochana-chathura, line 469: what does this TODO suppose to mean? I guess we don't need this.

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -470,10 +488,12 @@ private STNode parseStringContent() {
...
         if (nextToken.kind == SyntaxKind.IDENTIFIER_LITERAL) {
             return consume();
-        } else {
-            recover(nextToken, ParserRuleContext.STRING_CONTENT);
-            return parseStringContent();
         }
+        if (nextToken.kind == SyntaxKind.DOUBLE_QUOTE_TOKEN) {
+            return STNodeFactory.createEmptyNode();
+        }
+        recover(nextToken, ParserRuleContext.STRING_BODY);
```
lochana-chathura, line 495: shouldn't we go to recovery with STRING_CONTENT parser rule context?

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParserErrorHandler.java
```diff
@@ -82,6 +83,9 @@
     private static final ParserRuleContext[] KEY_LIST_END = {
             ParserRuleContext.ASSIGN_OP, ParserRuleContext.TABLE_END, ParserRuleContext.ARRAY_TABLE_FIRST_END};
 
+    private static final ParserRuleContext[] STRING_CONTENT = { ParserRuleContext.STRING_END,
+            ParserRuleContext.STRING_BODY};
```
lochana-chathura, line 87:
```suggestion
            ParserRuleContext.STRING_BODY };
```

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParserErrorHandler.java
```diff
@@ -65,7 +65,8 @@
             { ParserRuleContext.ARRAY_VALUE_LIST_END, ParserRuleContext.ARRAY_VALUE_START };
 
     private static final ParserRuleContext[] NUMERICAL_LITERAL = {ParserRuleContext.DECIMAL_INTEGER_LITERAL,
-            ParserRuleContext.DECIMAL_FLOATING_POINT_LITERAL};
+            ParserRuleContext.DECIMAL_FLOATING_POINT_LITERAL, ParserRuleContext.HEX_INTEGER_LITERAL,
+            ParserRuleContext.OCTAL_INTEGER_LITERAL, ParserRuleContext.BINARY_INTEGER_LITERAL};
```
lochana-chathura, line 69: correct formatting should be
```suggestion
            ParserRuleContext.OCTAL_INTEGER_LITERAL, ParserRuleContext.BINARY_INTEGER_LITERAL };
```
add a space after { here also.

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
lochana-chathura: I see. I guess it should be fine then. Usually syntax tree represents everything including minutiae that user inputs. (useful when formatting, etc)

Anyway since we capture the newline explicitly from the lexer, we need to add NEWLINE as an alternative path for ARRAY_VALUE_END ctx in error handler I think. There could be invalid tokens before the NEWLINE token.
e.g. 
```
[ "abc"  + invalid +
, "ABC" ]
```
@SupunS please have a look.

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -565,6 +588,9 @@ private STNode parseArrayValue() {
             return parseArray();
         } else if (nextToken.kind == CLOSE_BRACKET_TOKEN) {
             return null;
+        } else if (nextToken.kind == NEWLINE) {
+            consume();
+            return parseArrayValue();
         } else {
             recover(peek(), ParserRuleContext.ARRAY_VALUE_START);
```
lochana-chathura, line 595: same thing valid here also.

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
SupunS: Interesting point.

We designed the error handler to be same as the ballerina error handler, to not to consider newlines. Because newlines are just for validations and parsing could done even without newlines - also simplifies things.

@xlight05 I just noticed that we are not ignoring the newlines in the error-handler though. We should skip the newline tokens or do as @lochana-chathura suggested.

lochana-chathura: +1 for skipping. @xlight05 May be we can create a new issue and address this in a separate PR.

xlight05: Done. #27641

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
xlight05: Created an issue for this. #27641

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -291,9 +296,9 @@ private STNode parseSingleKey() {
             case IDENTIFIER_LITERAL:
                 return parseIdentifierLiteral();
             case DOUBLE_QUOTE_TOKEN:
-            case TRIPLE_DOUBLE_QUOTE_TOKEN:
```
lochana-chathura, line 294: Can we have a single key like this?
```toml
 [["""fruit"""]]
```
In this method we go to recovery with KEY_START ctx which allows both TRIPLE_DOUBLE_QUOTE_TOKEN and TRIPLE_SINGLE_QUOTE_TOKEN, but the parsing method does not allow them.

Therefore, for above example parseSingleKey will keep re-parsing again and again.

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -450,15 +472,41 @@ private STNode parseStringValue() {
...
+        STNode startingDoubleQuote = parseSingleQuoteToken(ParserRuleContext.LITERAL_STRING_START);
+        STNode content = parseStringContent();
+        STNode endingDoubleQuote = parseSingleQuoteToken(ParserRuleContext.LITERAL_STRING_END);
+        return STNodeFactory.createLiteralStringLiteralNode(startingDoubleQuote, content, endingDoubleQuote);
+    }
+
+    /**
+     * Parse Double quote token.
+     *
+     * @return Double quote token
```
lochana-chathura, line 490: single quote token

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -470,10 +518,12 @@ private STNode parseStringContent() {
...
         if (nextToken.kind == SyntaxKind.IDENTIFIER_LITERAL) {
             return consume();
-        } else {
-            recover(nextToken, ParserRuleContext.STRING_CONTENT);
-            return parseStringContent();
         }
+        if (nextToken.kind == SyntaxKind.DOUBLE_QUOTE_TOKEN) {
+            return STNodeFactory.createEmptyNode();
+        }
+        recover(nextToken, ParserRuleContext.STRING_CONTENT);
```
lochana-chathura, line 525: seems this method is called within both string and literal string. if it is literal string shouldn't we need to go to recovery with LITERAL_STRING_CONTENT?

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
xlight05: It seems like we cant reuse same context for multi strings. Addressed this issue by introducing separate tokens for parsing multiline (literal) strings.

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlLexer.java
```diff
@@ -356,6 +365,32 @@ private STToken readMultilineStringToken() {
...
+        reader.mark();
+        if (reader.isEOF()) {
+            return getSyntaxToken(SyntaxKind.EOF_TOKEN);
+        }
+
+        char nextChar = this.reader.peek();
+        char secondNextChar = this.reader.peek(1);
+        char thirdNextChar = this.reader.peek(2);
+        if (nextChar == LexerTerminals.SINGLE_QUOTE && secondNextChar == LexerTerminals.SINGLE_QUOTE && thirdNextChar
+                == LexerTerminals.SINGLE_QUOTE) {
```
lochana-chathura, line 378:
```suggestion
        if (nextChar == LexerTerminals.SINGLE_QUOTE && secondNextChar == LexerTerminals.SINGLE_QUOTE && 
        thirdNextChar == LexerTerminals.SINGLE_QUOTE) {
```

misc/toml-parser/src/main/java/io/ballerina/toml/internal/parser/TomlParser.java
```diff
@@ -445,35 +470,118 @@ private STNode parseBoolean() {
...
+    private STNode parseStringContent(ParserRuleContext ctx) {
         STToken nextToken = peek();
         if (nextToken.kind == SyntaxKind.IDENTIFIER_LITERAL) {
             return consume();
-        } else {
-            recover(nextToken, ParserRuleContext.STRING_CONTENT);
-            return parseStringContent();
         }
+        if (nextToken.kind == SyntaxKind.DOUBLE_QUOTE_TOKEN || nextToken.kind == SINGLE_QUOTE_TOKEN ||
+                nextToken.kind == TRIPLE_SINGLE_QUOTE_TOKEN || nextToken.kind == TRIPLE_DOUBLE_QUOTE_TOKEN) {
```
lochana-chathura, line 580: We have a small problem here. We call parseStringContent in 4 places and depending on the context we need only one kind to be checked here. Here we are heavily depending on the lexer logic. i.e we assume SINGLE_QUOTE_TOKEN is always followed by SINGLE_QUOTE_TOKEN when the string content is empty. We can make parsing independent of lexer logic by having 4 separate methods to parse each.

misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlTransformer.java
```diff
@@ -338,24 +340,93 @@ private TomlNodeLocation getPosition(Node node) {
...
+            }
+        }
+        TomlNodeLocation position = getPosition(literalStringLiteralNode);
+
+        return new TomlStringValueNode(valueString, position);
+    }
+
+    private boolean isMultilineString(Token token) {
+        return token.kind().equals(SyntaxKind.TRIPLE_SINGLE_QUOTE_TOKEN) ||
+                token.kind().equals(SyntaxKind.TRIPLE_DOUBLE_QUOTE_TOKEN);
```
lochana-chathura, line 390: can use == to check. (it can prevent possible NPE, complie time type safe and faster compared to a method call)

misc/toml-parser/src/main/java/io/ballerina/toml/syntax/tree/SyntaxKind.java
```diff
@@ -57,6 +62,7 @@
     // Invalid nodes
     INVALID_TOKEN(1600),
     MISSING_VALUE(1601),
+    INVALID_TOKEN_MINUTIAE_NODE(1603),
```
lochana-chathura, line 65:
```suggestion
    INVALID_TOKEN_MINUTIAE_NODE(1602),
```

lochana-chathura (approved):

anuruddhal (approved):

PR #27393, opened 2020-12-09
Remove packerina

keizer619:


cli/ballerina-cli/src/main/java/io/ballerina/cli/utils/JarFileWriter.java
```diff
@@ -15,7 +15,7 @@
  * specific language governing permissions and limitations
  * under the License.
```
hevayo, line 16: We do not need the JarFileWriter lets merge this PR and create a new issue to remove this class. I guess we use the writer in some test utils which also should be removed.

keizer619: I have created an issue
https://github.com/ballerina-platform/ballerina-lang/issues/27395

hevayo (approved):

PR #27479, opened 2020-12-14
Fix casting with special character fields

### LINKED ISSUE #27365: Casting fails when the field names contain special characters
**Description:**
$title. We don't seem to be using the decoded key when retrieving the field at https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java#L1242 resulting in `sourceField` becoming null, causing `false` to be returned from `checkIsRecordType`.

Need to check objects too.

**Steps to reproduce:**
```ballerina
type Foo record {
    string 'abc\.def;
};

type Bar record {
    string 'abc\.def;
};

public function main() {
    Foo foo = {'abc\.def: "abc"};
    Bar bar = <Bar> foo;
}
```

**Affected Versions:**
slp7

https://github.com/ballerina-platform/ballerina-lang/issues/27364

HindujaB:
## Purpose
$subject
Fixes #27365 

## Approach
- Change usage of encoded field names in `TypeChecker`
- Add support for Quoted Identifier in table keys

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
```diff
@@ -1433,7 +1434,7 @@ private static void loadTableType(MethodVisitor mv, BTableType bType) {
                 mv.visitInsn(DUP);
                 mv.visitLdcInsn((long) i);
                 mv.visitInsn(L2I);
-                mv.visitLdcInsn(fieldName);
+                mv.visitLdcInsn(StringEscapeUtils.unescapeJava(fieldName));
```
manuranga, line 1437: Why the java escape? Not identifier encode?

HindujaB: This field name is used to refer the `record` field name of the table key.  The record fields are mapped using the decoded field names. [escaped names == decode(encoded identifiers) ]
Eg: 
```ballerina
table key('emp\-id)

//field name passed as "emp\-id" at code gen (instruction-level)
//map key needed for record field access is "emp-id"
```

manuranga (approved):

PR #27571, opened 2020-12-18
Introduce extension points for websub message persistence

Bhashinee:
## Purpose
Introduce some extension points to support message persistence. A couple of newly introduced fields to the HubConfiguration record to configure functions to be invoked at the events like receiving updates, update delivery, and update delivery failure. These functions will include information like the message, callback URL, and topic details. 

## Approach
Newly introduced functions,
```
public type HubConfiguration record {|
    int leaseSeconds = 86400;
    SignatureMethod signatureMethod = SHA256;
    RemotePublishConfig remotePublish?;
    boolean topicRegistrationRequired = true;
    http:ClientConfiguration clientConfig?;
    HubPersistenceStore hubPersistenceStore?;
    function (WebSubContent content) tapOnMessage?;
    function (string callback, string topic, WebSubContent content) tapOnDelivery?;
    function (string callback, string topic, WebSubContent content, http:Response|error response) tapOnDeliveryFailure?;
|};
```
## Samples
```
import ballerina/http;
import ballerina/io;
import ballerina/runtime;
import ballerina/websub;
import ballerina/lang.'string;

public function main() {
    io:println("Starting up the Ballerina Hub Service");

    websub:Hub webSubHub;
    var result = websub:startHub(new http:Listener(9191), "/websub", "/hub",
        hubConfiguration = {
            remotePublish: {
                enabled: true
            },
            tapOnMessage: onMessageImpl,
            tapOnDelivery: onDeliveryImpl,
            tapOnDeliveryFailure: onDeliveryFailureImpl
        }
    );
    if (result is websub:Hub) {
        webSubHub = result;
    } else if (result is websub:HubStartedUpError) {
        webSubHub = result.startedUpHub;
    } else {
        io:println("Hub start error:" + <string>result.detail()?.message);
        return;
    }
    runtime:sleep(60000);
}

function onMessageImpl (websub:WebSubContent content) {
    var payload = content.payload;
    if (payload is byte[]) {
        string|error cont = 'string:fromBytes(payload);
        if (cont is string) {
            io:println(cont);
        }
    }
   io:println("on message");
}

function onDeliveryImpl (string callback, string topic, websub:WebSubContent content) {
   io:println("on Delivery");
}

function onDeliveryFailureImpl (string callback, string topic, websub:WebSubContent content, http:Response|error response, websub:FailureReason reason) {
   io:println("on Delivery failure");
}

```

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/connector/Executor.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/connector/Executor.java
@@ -160,14 +160,14 @@ public class Executor {
     /**
      * This method will invoke Ballerina function in a non-blocking manner.
      *
-     * @param scheduler   current scheduler
-     * @param classLoader normal classLoader
-     * @param orgName     org which the package belongs to
-     * @param packageName package which the class belongs to
-     * @param className   which the function resides/ or file name
-     * @param methodName  to be invokable unit
-     * @param callback    to be executed when execution completes.
-     * @param paramValues to be passed to invokable unit
+     * @param scheduler   The Current scheduler
+     * @param classLoader The classLoader
+     * @param orgName     The org which the package belongs to
+     * @param packageName The package which the class belongs to
+     * @param className   The class name or the file name in which the function resides
+     * @param methodName  The method name which is to be the invokable unit
+     * @param callback    The callback to be executed when the execution completes.
+     * @param paramValues The Parameters to be passed to the invokable unit
      */
     public static void executeFunctionAsync(Scheduler scheduler, ClassLoader classLoader, final String orgName,
             String packageName, String className, String methodName, CallableUnitCallback callback,
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/connector/Executor.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/connector/Executor.java
@@ -118,41 +118,33 @@ public class Executor {
                                          String packageName, String className, String methodName,
                                          Object... paramValues) {
         try {
-            Class<?> clazz = classLoader.loadClass(orgName + "." + packageName + "." + className);
-            int paramCount = paramValues.length * 2 + 1;
-            Class<?>[] jvmParamTypes = new Class[paramCount];
-            Object[] jvmArgs = new Object[paramCount];
-            jvmParamTypes[0] = Strand.class;
-            jvmArgs[0] = scheduler;
-            for (int i = 0, j = 1; i < paramValues.length; i++) {
-                jvmArgs[j] = paramValues[i];
-                jvmParamTypes[j++] = getJvmType(paramValues[i]);
-                jvmArgs[j] = true;
-                jvmParamTypes[j++] = boolean.class;
-            }
-            Method method = clazz.getDeclaredMethod(methodName, jvmParamTypes);
-            Function<Object[], Object> func = args -> {
-                try {
-                    return method.invoke(null, args);
-                } catch (IllegalAccessException | InvocationTargetException e) {
-                    throw new BallerinaException(methodName + " function invocation failed: " + e.getMessage());
-                }
-            };
+
             CountDownLatch completeFunction = new CountDownLatch(1);
-            FutureValue futureValue = scheduler.schedule(jvmArgs, func, null, new CallableUnitCallback() {
+
+            class Callback implements CallableUnitCallback {
+
+                private CountDownLatch countDownLatch;
+
+                private Callback(CountDownLatch countDownLatch) {
+                    this.countDownLatch = countDownLatch;
+                }
+
                 @Override
                 public void notifySuccess() {
-                    completeFunction.countDown();
+                    countDownLatch.countDown();
                 }
 
                 @Override
                 public void notifyFailure(ErrorValue error) {
-                    completeFunction.countDown();
+                    countDownLatch.countDown();
                 }
-            }, new HashMap<>(), BTypes.typeNull);
+            }
+
+            FutureValue futureValue = (FutureValue) executeFunctionAsync(scheduler, classLoader, orgName, packageName,
+                    className, methodName, new Callback(completeFunction), paramValues);
             completeFunction.await();
             return futureValue.result;
-        } catch (NoSuchMethodException | ClassNotFoundException | InterruptedException e) {
+        } catch (InterruptedException e) {
             throw new BallerinaException("invocation failed: " + e.getMessage());
         }
     }
@@ -169,7 +161,7 @@ public class Executor {
      * @param callback    The callback to be executed when the execution completes.
      * @param paramValues The Parameters to be passed to the invokable unit
      */
-    public static void executeFunctionAsync(Scheduler scheduler, ClassLoader classLoader, final String orgName,
+    public static Object executeFunctionAsync(Scheduler scheduler, ClassLoader classLoader, final String orgName,
             String packageName, String className, String methodName, CallableUnitCallback callback,
             Object... paramValues) {
         try {
@@ -193,8 +185,7 @@ public class Executor {
                     throw new BallerinaException(methodName + " function invocation failed: " + e.getMessage());
                 }
             };
-            FutureValue futureValue = scheduler
-                    .schedule(jvmArgs, func, null, callback, new HashMap<>(), BTypes.typeNull);
+            return scheduler.schedule(jvmArgs, func, null, callback, new HashMap<>(), BTypes.typeNull);
         } catch (NoSuchMethodException | ClassNotFoundException e) {
             throw new BallerinaException("invocation failed: " + e.getMessage());
         }
--- a/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/connector/Executor.java
+++ b/bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/connector/Executor.java
@@ -118,53 +118,6 @@ public class Executor {
                                          String packageName, String className, String methodName,
                                          Object... paramValues) {
         try {
-
-            CountDownLatch completeFunction = new CountDownLatch(1);
-
-            class Callback implements CallableUnitCallback {
-
-                private CountDownLatch countDownLatch;
-
-                private Callback(CountDownLatch countDownLatch) {
-                    this.countDownLatch = countDownLatch;
-                }
-
-                @Override
-                public void notifySuccess() {
-                    countDownLatch.countDown();
-                }
-
-                @Override
-                public void notifyFailure(ErrorValue error) {
-                    countDownLatch.countDown();
-                }
-            }
-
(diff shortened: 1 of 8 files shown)


stdlib/websub/src/main/ballerina/src/websub/hub_configuration.bal
```diff
@@ -36,6 +36,9 @@ RemotePublishConfig remotePublishConfig = {};
 boolean hubTopicRegistrationRequired = false;
 string hubPublicUrl = "";
 http:ClientConfiguration? hubClientConfig = ();
+function (WebSubContent content)? onMessageFunction = ();
+function (string callback, string topic, WebSubContent content)? onDeliveryFunction = ();
+function (string callback, string topic, WebSubContent content, http:Response|error response)? onDeliveryFailureFunction = ();
```
MaryamZi, line 41: Do we need to execute the current logic (logging) if the user has specified a custom function? 

Instead of making this nilable shall we just have the functions and assign as the default implementation the current logic that logs?

Given that delivery failure can happen due to three reasons, we can add maybe add another `FailureReason` parameter to quickly figure out what went wrong, and in the default case log the appropriate log.


```ballerina
const FAILURE_REASON_SUBSCRIPTION_GONE = "SUBSCRIPTION_GONE";
const FAILURE_REASON_FAILURE_STATUS_CODE = "FAILURE_STATUS_CODE";
const FAILURE_REASON_DELIVERY_FAILURE = "DELIVERY_FAILURE";

type FailureReason FAILURE_REASON_SUBSCRIPTION_GONE|FAILURE_REASON_FAILURE_STATUS_CODE|FAILURE_REASON_DELIVERY_FAILURE;

function (string callback, string topic, WebSubContent content, http:Response|error response, FailureReason reason) onDeliveryFailureFunction = defaultOnDeliveryFailureFunction;
```

This way we can avoid having to use the type test per delivery.

MaryamZi: The `onMessageFunction` default implementation can be an empty function or a debug log.

stdlib/websub/src/main/ballerina/src/websub/hub_service.bal
```diff
@@ -550,6 +559,21 @@ function distributeContent(string callback, SubscriptionDetails subscriptionDeta
     return;
 }
 
+function callOnDeliveryFailure(string callback, string topic, WebSubContent notification, http:Response|error response) {
+    function (string callback, string topic, WebSubContent content, http:Response|error response)?
+                               onDeliveryFailureFunc = onDeliveryFailureFunction;
+    if (onDeliveryFailureFunc is function (string callback, string topic, WebSubContent content, http:Response|error response)) {
```
MaryamZi, line 565: If we go with just functions as suggested above, we wouldn't need the type test. But generally, it is less expensive to check if a variable is `()` rather than a specific type, so if we are going with this approach let's invert the check.

stdlib/websub/src/main/ballerina/src/websub/hub_configuration.bal
Bhashinee: +1. Changed the code as requested.

stdlib/websub/src/main/ballerina/src/websub/hub_service.bal
Bhashinee: Moved to the approach suggested in https://github.com/ballerina-platform/ballerina-lang/pull/27571#discussion_r545610193

stdlib/websub/src/main/ballerina/src/websub/commons.bal
```diff
@@ -480,6 +480,18 @@ public type SubscriptionChangeResponse record {|
 /////////////////////////////////////////////////////////////
 //////////////////// WebSub Hub Commons /////////////////////
 /////////////////////////////////////////////////////////////
+# Represents subscription delivery failures related to the HTTP 410 Gone status code.
```
MaryamZi, line 483:
```suggestion
# Represents subscription delivery failure due to the subscriber's callback URL
# replying with a HTTP 410 Gone status code.
```

stdlib/websub/src/main/ballerina/src/websub/commons.bal
```diff
@@ -557,6 +579,19 @@ public function startHub(http:Listener hubServiceListener,
     hubSignatureMethod = getSignatureMethod(hubConfiguration.signatureMethod);
     remotePublishConfig = getRemotePublishConfig(hubConfiguration["remotePublish"]);
     hubTopicRegistrationRequired = hubConfiguration.topicRegistrationRequired;
+    var onMessageFunc = hubConfiguration["onMessage"];
```
MaryamZi, line 582: Shall we use optional field access instead?

```suggestion
    var onMessageFunc = hubConfiguration?.onMessage;
```

stdlib/websub/src/main/ballerina/src/websub/commons.bal
```diff
@@ -557,6 +579,19 @@ public function startHub(http:Listener hubServiceListener,
...
+    }
+    var onDeliveryFunc = hubConfiguration["onDelivery"];
+    if (!(onDeliveryFunc is ())) {
+       onDeliveryFunction = onDeliveryFunc;
+    }
+    var onDeliveryFailureFunc = hubConfiguration["onDeliveryFailure"];
+    if (!(onDeliveryFailureFunc is ())) {
+       onDeliveryFailureFunction = onDeliveryFailureFunc;
+    }
+    asyncContentDelivery = hubConfiguration["asyncContentDelivery"];
```
MaryamZi, line 594:
```suggestion
    asyncContentDelivery = hubConfiguration,asyncContentDelivery;
```

stdlib/websub/src/main/ballerina/src/websub/hub_configuration.bal
```diff
@@ -78,3 +85,27 @@ function getSignatureMethod(SignatureMethod? signatureMethod) returns string {
 function getRemotePublishConfig(RemotePublishConfig? remotePublish) returns RemotePublishConfig {
     return remotePublish is RemotePublishConfig ? remotePublish : {};
 }
+
+function defaultOnMessageFunction(WebSubContent content) {
+    log:printDebug("Message received at the hub");
+}
+
+function defaultOnDeliveryFunction(string callback, string topic, WebSubContent content) {
+    log:printDebug("Content delivery to callback[" + callback + "] successful for topic[" + topic + "]");
```
MaryamZi, line 94: Let's use the function pointer approach instead, because there's concatenations. https://ballerina.io/learn/by-example/log-api.html

```suggestion
    log:printDebug(() => "Content delivery to callback[" + callback + "] successful for topic[" + topic + "]");
```

stdlib/websub/src/main/ballerina/src/websub/hub_configuration.bal
```diff
@@ -78,3 +85,27 @@ function getSignatureMethod(SignatureMethod? signatureMethod) returns string {
...
+                     + "], topic[" + topic + "]");
+    } else if (reason is FAILURE_REASON_FAILURE_STATUS_CODE) {
+       http:Response resp = <http:Response> response;
+       log:printError("Error delivering content to callback[" + callback + "] for topic["
+                                + topic + "]: received response code " + resp.statusCode.toString());
+    } else {
+       error err = <error> response;
+       log:printError("Error delivering content to callback[" + callback + "] for topic["
+                                   + topic + "]: " + <string> err.detail()?.message);
+    }
```
MaryamZi, line 110:
```suggestion
                     + "], topic[" + topic + "]");
       return;                     
    } 
    
    if (reason is FAILURE_REASON_FAILURE_STATUS_CODE) {
       http:Response resp = <http:Response> response;
       log:printError("Error delivering content to callback[" + callback + "] for topic["
                                + topic + "]: received response code " + resp.statusCode.toString());
       return;
    }
    
    error err = <error> response;
    log:printError("Error delivering content to callback[" + callback + "] for topic["
                          + topic + "]: " + <string> err.detail()?.message);
```

stdlib/websub/src/main/ballerina/src/websub/commons.bal
```diff
@@ -489,13 +501,23 @@ public type SubscriptionChangeResponse record {|
 #                               to the topic
 # + clientConfig - The configuration for the hub to communicate with remote HTTP endpoints
 # + hubPersistenceStore - The `HubPersistenceStore` to use to persist hub data
+# + onMessage - The configuration for passing a function to be invoked when an update is received at the hub
+# + onDelivery - The configuration for passing a function to be invoked when an update is successfully delivered to the subscribers
+# + onDeliveryFailure - The configuration for passing a function to be invoked when an update is failed to be delivered to the subscribers
+# + asyncContentDelivery - The configuration for specifying the way(sync/async) of the update delivery to subscribers. Setting this to true
+#                          may change the order
```
MaryamZi, line 508:
```suggestion
# + asyncContentDelivery - The configuration to specify is content delivery to subscribers should be done asynchronously.
#                          Setting this to `true` may result in the order not being preserved in message delivery.
```

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/connector/Executor.java
```diff
@@ -157,6 +157,49 @@ public void notifyFailure(ErrorValue error) {
...
+     * @param scheduler   The Current scheduler
+     * @param classLoader The classLoader
+     * @param orgName     The org which the package belongs to
+     * @param packageName The package which the class belongs to
+     * @param className   The class name or the file name in which the function resides
+     * @param methodName  The method name which is to be the invokable unit
+     * @param callback    The callback to be executed when the execution completes.
+     * @param paramValues The Parameters to be passed to the invokable unit
+     */
+    public static void executeFunctionAsync(Scheduler scheduler, ClassLoader classLoader, final String orgName,
```
warunalakshitha, line 172: Since this new executeFunctionAsync takes callback as a param, existing 'executeFunctionAsync'  can call this function passing the callback with CountDownLatch. So we do not need to duplicate common logic

stdlib/websub/src/main/ballerina/src/websub/hub_configuration.bal
shafreenAnfar: I think in all the other cases, we simply tap the flow. In other words, the core functionality stays the same regardless of what we are doing. Doing what is suggested make this function in consistent with the other two if I am not mistaken.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/values/connector/Executor.java
Bhashinee: Removed this code segment as we decided not to make this async.

stdlib/websub/src/main/ballerina/src/websub/hub_service.bal
Bhashinee: Inverted the check.

shafreenAnfar (approved):

Issue #27586, opened 2020-12-19
Add the platform annotation @display


sameerajayasoma:
Refer to [this issue](https://github.com/ballerina-platform/ballerina-spec/issues/681) for more information

rdhananjaya:
Related: https://github.com/ballerina-platform/ballerina-lang/issues/27587

MaryamZi:
Fixed in #27626, merged via #27652.

Issue #18342, opened 2019-08-30
Fix ATTEMPT_EXPOSE_NON_PUBLIC_SYMBOL validation for Objects, Unions and other types.


hasithaa:
**Description:**
See Referred pull.
https://github.com/ballerina-platform/ballerina-lang/pull/18341/files#diff-4025beda7fd769ab390b6d789e6cec16R1241 

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

hasithaa:
Spec allows exposing non-public types. Hence we should reconsider this adding this as a warning.

MaryamZi:
Related issues
https://github.com/ballerina-platform/ballerina-lang/issues/30324
https://github.com/ballerina-platform/ballerina-lang/issues/23049

MaryamZi:
Should we prioritize this issue given that users would have to make constructs public to get rid of the error right now? 

In addition to the compiler changes, we will also have to consider aspects such as doc generation.

Issue #27587, opened 2020-12-19
Remove @icon annotation 


sameerajayasoma:
With the `@display` annotation, we can remove the `@icon` annotation. Refer to [this issue](https://github.com/ballerina-platform/ballerina-spec/issues/681) for more information.

MaryamZi:
Removed in #27626, merged via #27652.

Issue #22307, opened 2020-03-30
Docerina support for record/object field-level documentation


pramodya1994:
**Description:**
#21529 allows field level documentation for records/objects.

```ballerina
# Docs for this record.
public type Foo record {

    # Docs for this field.
    string bar;
};

# Docs for this object.
public type Baz record {

    # Docs for this field.
    string qux;
};
```

Doc generation needs to be updated to consider such documentation.
In this issue, we are only considering **BuildTools & Docerina** capabilities.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**
Tooling aspect
https://github.com/ballerina-platform/ballerina-lang/issues/21599

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #27626, opened 2020-12-23
Introduce display annotation

rdhananjaya:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/27586 and https://github.com/ballerina-platform/ballerina-lang/issues/27587

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


rdhananjaya:
@nadeeshaan @rasika there are some lang-server test fixes due to introduction of this annot and it's datatype. Please check.

langlib/lang.annotations/src/main/ballerina/annotations.bal
```diff
@@ -73,5 +73,16 @@ public type StrandData record {|
 # Denotes new Strand execution semantics.
 public const annotation StrandData strand on source worker;
 
-# Denotes icon metadata related to types and functions.
-public const annotation record {| string path; |} icon on source type, source function, source class;
+# Describes display metadata.
+#
+# + label - label to be dispaled for a Ballerina construct
```
MaryamZi, line 78:
```suggestion
# + label - label to be displayed for a Ballerina construct
```

MaryamZi: Should this just be "label for the Ballerina construct" instead?

tests/jballerina-unit-test/src/test/resources/test-src/annotations/display_annot.bal
```diff
@@ -14,15 +14,33 @@
 // specific language governing permissions and limitations
 // under the License.
 
-@icon { path: "/fooIconPath.icon" }
+@display {iconPath: "fooIconPath.icon", label: "Foo function"}
```
MaryamZi, line 17: Shall we add a test for the openness of `DisplayData` too? With additional fields in the annotation?

nadeeshaan (dismissed):
Reviewed the language server test changes and looks good. Hence approving

nadeeshaan:
@rdhananjaya does this `DisplayData` record is supposed to be used externally as well? This is because we suggest the particular record for the users since it is globally visible.

rdhananjaya:
> @rdhananjaya does this `DisplayData` record is supposed to be used externally as well? This is because we suggest the particular record for the users since it is globally visible.

This is what being approved in the spec issue https://github.com/ballerina-platform/ballerina-spec/issues/681#issuecomment-750454471

But we can't expose the module visible `DisplayData` via a public visible annotation, hence had to make it public.

MaryamZi:
> > @rdhananjaya does this `DisplayData` record is supposed to be used externally as well? This is because we suggest the particular record for the users since it is globally visible.
> 
> This is what being approved in the spec issue [ballerina-platform/ballerina-spec#681 (comment)](https://github.com/ballerina-platform/ballerina-spec/issues/681#issuecomment-750454471)
> 
> But we can't expose the module visible `DisplayData` via a public visible annotation, hence had to make it public.

What if we use an anonymous record in-line instead of introducing the `DisplayData` record separately, for the time-being? Wouldn't impact usage since if it were of module-level visibility we won't be able to access it outside the module anyway?

rdhananjaya:
> > > @rdhananjaya does this `DisplayData` record is supposed to be used externally as well? This is because we suggest the particular record for the users since it is globally visible.
> > 
> > 
> > This is what being approved in the spec issue [ballerina-platform/ballerina-spec#681 (comment)](https://github.com/ballerina-platform/ballerina-spec/issues/681#issuecomment-750454471)
> > But we can't expose the module visible `DisplayData` via a public visible annotation, hence had to make it public.
> 
> What if we use an anonymous record in-line instead of introducing the `DisplayData` record separately, for the time-being? Wouldn't impact usage since if it were of module-level visibility we won't be able to access it outside the module anyway?

Yes, we can do that but then there are no doc and I don't see an issue exposing this publicly as there are other things like `StrandData`

MaryamZi:
> > > > @rdhananjaya does this `DisplayData` record is supposed to be used externally as well? This is because we suggest the particular record for the users since it is globally visible.
> > > 
> > > 
> > > This is what being approved in the spec issue [ballerina-platform/ballerina-spec#681 (comment)](https://github.com/ballerina-platform/ballerina-spec/issues/681#issuecomment-750454471)
> > > But we can't expose the module visible `DisplayData` via a public visible annotation, hence had to make it public.
> > 
> > 
> > What if we use an anonymous record in-line instead of introducing the `DisplayData` record separately, for the time-being? Wouldn't impact usage since if it were of module-level visibility we won't be able to access it outside the module anyway?
> 
> Yes, we can do that but then there are no doc and I don't see an issue exposing this publicly as there are other things like `StrandData`

In this scenario what we need is the field level documentation, right? We can add them in-line? (Related issues - https://github.com/ballerina-platform/ballerina-lang/issues/22307, https://github.com/ballerina-platform/ballerina-lang/issues/21599)

```ballerina
# Denotes general-purpose metadata to customize how Ballerina symbols are displayed in a UI environment.
public const annotation record {
    # label for the Ballerina construct
    string label;
    # icon path relative to module's resouce directory
    string iconPath;
} display on source type, source class,
      source function, source return, source parameter, source field, source listener,
      source var, source const, source annotation;
```

Another issue with a public type in `lang.annotations` (other than the tooling aspect) is that it becomes part of the root scope, and that would mean one less identifier available to users.

rdhananjaya:
> > > > > @rdhananjaya does this `DisplayData` record is supposed to be used externally as well? This is because we suggest the particular record for the users since it is globally visible.
> > > > 
> > > > 
> > > > This is what being approved in the spec issue [ballerina-platform/ballerina-spec#681 (comment)](https://github.com/ballerina-platform/ballerina-spec/issues/681#issuecomment-750454471)
> > > > But we can't expose the module visible `DisplayData` via a public visible annotation, hence had to make it public.
> > > 
> > > 
> > > What if we use an anonymous record in-line instead of introducing the `DisplayData` record separately, for the time-being? Wouldn't impact usage since if it were of module-level visibility we won't be able to access it outside the module anyway?
> > 
> > 
> > Yes, we can do that but then there are no doc and I don't see an issue exposing this publicly as there are other things like `StrandData`
> 
> In this scenario what we need is the field level documentation, right? We can add them in-line? (Related issues - #22307, #21599)
> 
> ```ballerina
> # Denotes general-purpose metadata to customize how Ballerina symbols are displayed in a UI environment.
> public const annotation record {
>     # label for the Ballerina construct
>     string label;
>     # icon path relative to module's resouce directory
>     string iconPath;
> } display on source type, source class,
>       source function, source return, source parameter, source field, source listener,
>       source var, source const, source annotation;
> ```
> 
> Another issue with a public type in `lang.annotations` (other than the tooling aspect) is that it becomes part of the root scope, and that would mean one less identifier available to users.

Ack.

MaryamZi:
> > > > > > @rdhananjaya does this `DisplayData` record is supposed to be used externally as well? This is because we suggest the particular record for the users since it is globally visible.
> > > > > 
> > > > > 
> > > > > This is what being approved in the spec issue [ballerina-platform/ballerina-spec#681 (comment)](https://github.com/ballerina-platform/ballerina-spec/issues/681#issuecomment-750454471)
> > > > > But we can't expose the module visible `DisplayData` via a public visible annotation, hence had to make it public.
> > > > 
> > > > 
> > > > What if we use an anonymous record in-line instead of introducing the `DisplayData` record separately, for the time-being? Wouldn't impact usage since if it were of module-level visibility we won't be able to access it outside the module anyway?
> > > 
> > > 
> > > Yes, we can do that but then there are no doc and I don't see an issue exposing this publicly as there are other things like `StrandData`
> > 
> > 
> > In this scenario what we need is the field level documentation, right? We can add them in-line? (Related issues - #22307, #21599)
> > ```ballerina
> > # Denotes general-purpose metadata to customize how Ballerina symbols are displayed in a UI environment.
> > public const annotation record {
> >     # label for the Ballerina construct
> >     string label;
> >     # icon path relative to module's resouce directory
> >     string iconPath;
> > } display on source type, source class,
> >       source function, source return, source parameter, source field, source listener,
> >       source var, source const, source annotation;
> > ```
> > 
> > 
> > Another issue with a public type in `lang.annotations` (other than the tooling aspect) is that it becomes part of the root scope, and that would mean one less identifier available to users.
> 
> Ack.

The separate type wouldn't be an issue anyway when we fix https://github.com/ballerina-platform/ballerina-lang/issues/18342 though, since it can then be of module-level visibility.

hasithaa:
> > > > > > > @rdhananjaya does this `DisplayData` record is supposed to be used externally as well? This is because we suggest the particular record for the users since it is globally visible.
> > > > > > 
> > > > > > 
> > > > > > This is what being approved in the spec issue [ballerina-platform/ballerina-spec#681 (comment)](https://github.com/ballerina-platform/ballerina-spec/issues/681#issuecomment-750454471)
> > > > > > But we can't expose the module visible `DisplayData` via a public visible annotation, hence had to make it public.
> > > > > 
> > > > > 
> > > > > What if we use an anonymous record in-line instead of introducing the `DisplayData` record separately, for the time-being? Wouldn't impact usage since if it were of module-level visibility we won't be able to access it outside the module anyway?
> > > > 
> > > > 
> > > > Yes, we can do that but then there are no doc and I don't see an issue exposing this publicly as there are other things like `StrandData`
> > > 
> > > 
> > > In this scenario what we need is the field level documentation, right? We can add them in-line? (Related issues - #22307, #21599)
> > > ```ballerina
> > > # Denotes general-purpose metadata to customize how Ballerina symbols are displayed in a UI environment.
> > > public const annotation record {
> > >     # label for the Ballerina construct
> > >     string label;
> > >     # icon path relative to module's resouce directory
> > >     string iconPath;
> > > } display on source type, source class,
> > >       source function, source return, source parameter, source field, source listener,
> > >       source var, source const, source annotation;
> > > ```
> > > 
> > > 
> > > Another issue with a public type in `lang.annotations` (other than the tooling aspect) is that it becomes part of the root scope, and that would mean one less identifier available to users.
> > 
> > 
> > Ack.
> 
> The separate type wouldn't be an issue anyway when we fix #18342 though, since it can then be of module-level visibility.

+1, Looks good. Once we fix the root module issue, we can define `DisplayData` as a separate type.

tests/jballerina-unit-test/src/test/resources/test-src/annotations/display_annot.bal
rdhananjaya: Done

langlib/lang.annotations/src/main/ballerina/annotations.bal
```diff
@@ -73,5 +73,12 @@ public type StrandData record {|
 # Denotes new Strand execution semantics.
 public const annotation StrandData strand on source worker;
 
-# Denotes icon metadata related to types and functions.
-public const annotation record {| string path; |} icon on source type, source function, source class;
+# Denotes general-purpose metadata to customize how Ballerina symbols are displayed in a UI environment.
+public const annotation record {
+    # label for the Ballerina construct
+    string label;
+    # icon path relative to module's resouce directory
```
MaryamZi, line 80:
```suggestion
    # icon path relative to module's resource directory
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/annotations/DisplayAnnotationTest.java
```diff
@@ -29,57 +29,65 @@
...
         List<BLangAnnotationAttachment> attachedFuncAttachments =
                 ((BLangClassDefinition) clz).functions.get(0).annAttachments;
         String annotAsString =
                 getActualExpressionFromAnnotationAttachmentExpr(attachedFuncAttachments.get(0).getExpression())
                         .toString();
-        Assert.assertEquals(annotAsString, " {path: /kMemberFuncIconPath.icon}");
+        Assert.assertEquals(annotAsString, " {iconPath: kMemberFuncIconPath.icon,label: k method}");
     }
 
     @Test void testIconAnnotationNegative() {
```
MaryamZi, line 87:
```suggestion
    @Test void testDisplayAnnotationNegative() {
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/annotations/DisplayAnnotationTest.java
```diff
@@ -29,57 +29,65 @@
...
+    @Test
+    public void testDisplayAnnotOnServiceDecl() {
+        BLangService service = (BLangService) result.getAST().getServices().get(0);
+        BLangAnnotationAttachment attachment = service.getAnnotationAttachments().get(0);
+        Assert.assertEquals(getActualExpressionFromAnnotationAttachmentExpr(attachment.expr).toString(),
+                " {iconPath: service.icon,label: service,misc: Other info}");
     }
 
     @Test (enabled = false)
     public void testIconOnObjectAndMemberFunction() {
```
MaryamZi, line 72:
```suggestion
    public void testDisplayOnObjectAndMemberFunction() {
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/annotations/DisplayAnnotationTest.java
```diff
@@ -29,57 +29,65 @@
...
+
+    @Test
+    public void testDisplayAnnotOnServiceDecl() {
+        BLangService service = (BLangService) result.getAST().getServices().get(0);
+        BLangAnnotationAttachment attachment = service.getAnnotationAttachments().get(0);
+        Assert.assertEquals(getActualExpressionFromAnnotationAttachmentExpr(attachment.expr).toString(),
+                " {iconPath: service.icon,label: service,misc: Other info}");
     }
 
     @Test (enabled = false)
```
MaryamZi, line 71: Can this be enabled?

rdhananjaya: We can by removing ``.cloneReadOnly()` bit, but then when we fix the annotation cloning issue this will fail.

MaryamZi: That's OK, right? The rest will probably fail anyway so we can fix the set of tests together? WDYT?

rdhananjaya: done

tests/jballerina-unit-test/src/test/resources/test-src/annotations/display_annot_negative.bal
```diff
@@ -14,11 +14,11 @@
...
-@icon { path: "/fooIconPath.icon" }
+@display {iconPath: "fooIconPath.icon", label: "l"}
+@display {iconPath: "fooIconPath.icon", label: "j"}
 function foo(int i, string k) returns int {
     return i;
 }
 
-@icon { path: "/fooIconPath.icon" }
+@display {iconPath: "fooIconPath.icon", label: "k"}
 int GLOBAL = 3;
```
MaryamZi, line 24: Do we need this now?

rdhananjaya: Removed

MaryamZi (approved):

PR #27642, opened 2021-01-04
Improve performance by caching typedesc

### LINKED ISSUE #27142: Optimize creating new type descriptor for new record values
**Description:**
We are creating new type descriptor when creating new record values. This was done part of supporting default value initializing with closures.  But when we do not have any closures, type descriptors can be reused. This can be done using cache mechanism or may be compile time (code gen to use same type descriptors were types which not have have closures). This will improve the performance of simple record value creation.

Related issues https://github.com/ballerina-platform/ballerina-lang/issues/20851

manuranga:
Fixes #27142


warunalakshitha:
LGTM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
```diff
@@ -1895,18 +1896,28 @@ void generateNegateIns(BIRNonTerminator.UnaryOP unaryOp) {
     }
 
     void generateNewTypedescIns(BIRNonTerminator.NewTypeDesc newTypeDesc) {
-
+        List<BIROperand> closureVars = newTypeDesc.closureVars;
         BType type = newTypeDesc.type;
+        if (type.tag == TypeTags.RECORD && closureVars.size() == 0 && type.tsymbol != null) {
```
warunalakshitha, line 1901: Can type.tsymbol be null?

manuranga: It seems to coming null for `map` this is good for me because I anyway needed to avoid that.
Maybe I am relaying on a bug

warunalakshitha (approved):

PR #27650, opened 2021-01-04
Cache the public type symbols for internal types

pubudu91:
## Purpose
This PR adds a caching mechanism for the public type symbols so that we won't be creating the same type symbol over and over again for a given internal type.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/TypesFactory.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/TypesFactory.java
@@ -57,8 +57,11 @@ import org.wso2.ballerinalang.compiler.semantics.model.types.BXMLSubType;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BXMLType;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangExpression;
 import org.wso2.ballerinalang.compiler.util.CompilerContext;
+import org.wso2.ballerinalang.compiler.util.TypeTags;
 import org.wso2.ballerinalang.util.Flags;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 
 import static org.ballerinalang.model.types.TypeKind.OBJECT;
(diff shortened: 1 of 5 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/TypesFactory.java
```diff
@@ -129,6 +134,27 @@ public TypeSymbol getTypeDescriptor(BType bType, boolean rawTypeOnly) {
...
+
+            if (bType == ((AbstractTypeSymbol) typeSymbol).getBType()) {
+                return typeSymbol;
+            }
+        }
+
+        TypeSymbol typeSymbol = createTypeDescriptor(bType, moduleID);
+
+        // Because of the above explained reason, equivalent invokable types won't get cached either here.
+        typeCache.putIfAbsent(bType, typeSymbol);
```
nadeeshaan, line 153: Does this means for the invokable types we just create the typeSymbol on query and does not replace the symbol? Will there be any harm for replacing the typeSymbol then?

nadeeshaan (approved):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/TypesFactory.java
pubudu91: Since others don't override equals, this means we get type symbols cached for each unique `BType` instance. For invokable types, the `equals()` is basically a type equivalence test. For example, consider the following functions:
```ballerina
function add(int x, int y) returns int => x + y;
For the above  
function sum(int a, int b) returns int => a + b;
```
The above are two distinct functions. But the types of these are equivalent. Hence `equals()` will be true. So can't really cache the function type symbols against their internal counterparts. We can replace the cache entry. But still, for a set of equivalent invokable types, there will always be only one cache entry.

nadeeshaan: Cool, then +1 for this approach

nadeeshaan (approved):

Issue #27625, opened 2020-12-23
Enable bir spec tests


warunalakshitha:
**Description:**
Bir spec tests have been disabled after jdk 11 migration due to issue with kaitai. This will be fixed with https://github.com/ballerina-platform/ballerina-lang/pull/26931.

But due to recent language changes bir tests are broken. We need to,

1) Migrate bir test bal files to new lanugage syntax
2) Fix bir format due to new bir changes

PR #27735, opened 2021-01-08
Update bir-spec tests

rdhananjaya:
## Purpose
$title.

Note: JBallerina back-end is modifying the `Bir` object while generating JVM code. So the BIR module that is used in serialization of BIR differs from the one used in tests.

fixes: https://github.com/ballerina-platform/ballerina-lang/issues/27625

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/projects/NullBackend.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * Ballerina compiler backend that does not emit backend code, used for testing BIR.
+ *
+ * @since 2.0
+ */
+public class NullBackend extends CompilerBackend {
```
Kishanthan, line 33: Do we need a backend here? The idea behind bancked is it take cares of compilation backend related tasks, such as codegen, interop validation, etc. But BIRGen is actually part of compiler frontend right? So having a Backend here does not make sense to me.

rdhananjaya: This class was created because of this [comment](https://github.com/ballerina-platform/ballerina-lang/blob/d87f41f6f39e83424ddab7efe2d48b453a9a7c71/compiler/ballerina-lang/src/main/java/io/ballerina/projects/JBallerinaBackend.java#L76) wanting to make `PackageContext` an internal package. 
Would it be ok to access it from `BCompileUtil` class?

Kishanthan: You don't need the backend class here. Just calling pckCompilation.getCompilation would give you the symbol to access BIR

Kishanthan: I was wrong here. I was thinking codegen phase does not invoke birgen phase and was seperated out but looks like they are both part of generateCodeInternal method in ModuleContext class. Ideally, we should separate them out to different methods.

We have to go with separate backend class here as per the original PR.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/NullBackend.java
```diff
@@ -0,0 +1,101 @@
...
+    private DefaultDiagnosticResult diagnosticResult;
+
+    public NullBackend(PackageCompilation compilation) {
+        this.compilation = compilation;
+        this.packageContext = compilation.packageContext();
+        ProjectEnvironment projectEnvContext = this.packageContext.project().projectEnvironmentContext();
+        this.compilerContext = projectEnvContext.getService(CompilerContext.class);
+    }
+
+    public void compile() {
```
Kishanthan, line 46: This is actually not compile, but generate BIR right? Shall we rename it accordingly? Also, should we have this method explicitly called? Why don't we move the call to this method to the constructor?

rdhananjaya: Updated to use a static factory method similar to JBallerinaBackend. Please check.

warunalakshitha:
LGTM

Kishanthan (approved):
LGTM

compiler/ballerina-lang/src/main/java/io/ballerina/projects/NullBackend.java
```diff
@@ -0,0 +1,105 @@
+/* Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
Kishanthan, line 1: 2021 ?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/NullBackend.java
```diff
@@ -0,0 +1,105 @@
...
+import org.wso2.ballerinalang.compiler.util.CompilerContext;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * Ballerina compiler backend that does not emit backend code, used for testing BIR.
+ *
+ * @since 2.0
```
Kishanthan, line 31: 2.0.0

tests/ballerina-test-utils/src/main/java/org/ballerinalang/test/BCompileUtil.java
```diff
@@ -70,6 +72,17 @@ public static CompileResult compile(String sourceFilePath) {
         return compileResult;
     }
 
+    public static BPackageSymbol generateBIR(String sourceFilePath) {
```
Kishanthan, line 75: Is it the symbol we need to return or the BIRPackageFile?

rdhananjaya: I think best would be to move it [BIRCompileResult](https://github.com/ballerina-platform/ballerina-lang/blob/62191c2566e9e35121d1169e6fc57021ef0b962a/docs/bir-spec/src/test/java/org/ballerinalang/birspec/BIRTestUtils.java#L646) into BCompileUtil and return that. WDYT?

Kishanthan: +1

Kishanthan (approved):

warunalakshitha (approved):

PR #27797, opened 2021-01-12
[Ballerina Shell] Add REPL implementation for Ballerina

### LINKED ISSUE #27796: [Ballerina Shell] Implement REPL support for Ballerina
**Description:**

The Ballerina-shell tool is an interactive tool for learning the Ballerina programming language and prototyping Ballerina code. Ballerina-shell is a Read-Evaluate-Print Loop (REPL), which evaluates declarations, statements, and expressions as they are entered and immediately shows the results. Currently, the tool is run via the command line using `ballerina shell` command. Using Ballerina-shell, one can enter program statements one at a time and immediately see the result.

kdsuneraavinash:
## Purpose

The Ballerina-shell tool is an interactive tool for learning the Ballerina programming language and prototyping Ballerina code. Ballerina-shell is a Read-Evaluate-Print Loop (REPL), which evaluates declarations, statements, and expressions as they are entered and immediately shows the results. Currently, the tool is run via the command line by using `ballerina shell` command. Using Ballerina-shell, one can enter program statements one at a time and immediately see the result.

Fixes #27796

## Approach

Uses a wrapper-based technique used in similar projects such as RCRL. The current implementation will generate a ballerina source with the necessary code to evaluate a given expression by using a static memory class that is accessible via both the host application and the generated ballerina code. More implementation details are included in [implementation README](https://github.com/kdsuneraavinash/ballerina-shell/blob/master/shell-core/README.md).

## Samples

Following is a sample REPL session.

```ballerina
=$ int i = 200 

=$ i + 1000 
1200

=$ string|error s = "Hello" 

=$ s  
Hello

=$ s = error("Erro") 

=$ s  
Exception occurred: error("Erro")
| Unhandled Runtime Error.

=$ int t = "Error"   
| incompatible types: expected 'int', found 'string'
|     int t = "Error";
|             ^
| Compilation aborted because of errors.

=$ s is error 
true
```

## Remarks

- **Assignments to global variables in closures or class methods will not work** - Assignments done to global variables in closures will not be reflected after the execution. The changes will be visible only for the scope belonging to the snippet where the closure was defined.
- **Enum definition with explicit expression is not fully supported** - Enum declarations with explicit expressions are not supported fully. Some functionalities may not work properly.
- **Support with external libraries not tested completely** 
- **Arrays with a fixed length will lose that information once defined** - `int[3] a = [1, 2, 3]` and `int a = [1, 2, 3]` will behave in the same manner. As a result, some array bindings may not be possible.
- **Parser timeout issues** - Currently, the parser phase is implemented with a timeout (default 100ms when running via CLI, but can be changed by an argument) to ensure a better response time for erroneous snippets. However, this might cause issues in some old hardware where the execution might be slower than expected (where even valid executions might exceed the timeout).
- **Using var with types that are not visible to the public** - This is not currently possible. One has to explicitly define a type for these types of variable declarations (eg: `any|error` will work)
- **Cyclic dependencies of module-level types not possible** - Obviously, because a REPL is run line by line, One may not use a type that would be defined later. As a result, cyclic dependencies are not possible in module level declarations.
- **Type guards for global variables do not work** - Since all the variables are defined as global variables, type guards will not work for them. This will be fixed once this support is added to the ballerina compiler. However, type guards will work as intended for local variables.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples


Kishanthan (commented):
All license texts in files/classes should be for year 2021
All new files/classes addition should have @ since version text 

@kdsuneraavinash , can you update the PR accordingly ?

kdsuneraavinash:
I updated the PR.

Kishanthan (commented):
@kdsuneraavinash , could you arrange a code review for this PR?

Kishanthan:
@hasithaa @MaryamZi @manuranga @warunalakshitha @lochana-chathura 

Folks, could you please review this PR and approve? We are going to merge this for alpha.

ballerina-shell/modules/shell-cli/src/main/resources/command.header.txt
```diff
@@ -0,0 +1,9 @@
+  ____        _ _           _                _____ _          _ _
+ |  _ \      | | |         (_)              / ____| |        | | |
+ | |_) | __ _| | | ___ _ __ _ _ __   __ _  | (___ | |__   ___| | |
+ |  _ < / _` | | |/ _ | '__| | '_ \ / _` |  \___ \| '_ \ / _ | | |
+ | |_) | (_| | | |  __| |  | | | | | (_| |  ____) | | | |  __| | |
+ |____/ \__,_|_|_|\___|_|  |_|_| |_|\__,_| |_____/|_| |_|\___|_|_|
+=============================================== REPL for Ballerina
+Welcome to Ballerina Shell REPL.
+Type /exit to exit and /help to get about information.
```
Kishanthan, line 9: How about `Type /exit to exit and /help for help` ?

Kishanthan: Or list of available commands ?

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
```diff
@@ -0,0 +1,19 @@
+Ballerina Shell Help
+
+Type a ballerina statement, expression or a declaration.
+
+Imports should be done with prefixes and the modules
+should be available in the classpath.
+To see available imports use /imports.
+
+Following command are the in-built commands.
```
Kishanthan, line 9: `commands`  plural

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
```diff
@@ -0,0 +1,19 @@
...
+
+/exit       - Exit the terminal shell.
+/debug      - Toggle debug mode. You will see debug messages.
+/reset      - Reset REPL state.
+              All previous code snippets will be removed from REPL memory.
+/imports    - List the available imports.
+/vars       - List the declared variables and their values.
+/dclns      - List the user defined module level declarations.
+/help       - Get this message.
+/help TOPIC - Get help on TOPIC.
```
Kishanthan, line 19: `<TOPIC>` in the placeholder mentioning we can type in

ballerina-shell/modules/shell-cli/src/main/resources/command.header.txt
sameerajayasoma: Btw, @kdsuneraavinash should remove the BIG Ballerina logo at the start of the shell?

kdsuneraavinash: Ok I'll remove everything except for 
```
Welcome to Ballerina Shell REPL.
Type /exit to exit and /help to get about information.
```

or should I remove the welcome as well?

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
kdsuneraavinash: Is this ok?

```
/exit           - Exit the terminal shell.
/debug          - Toggle debug mode. You will see debug messages.
/reset          - Reset REPL state.
                  All previous code snippets will be removed from REPL memory.
/imports        - List the available imports.
/vars           - List the declared variables and their values.
/dclns          - List the user defined module level declarations.
/help           - Get this message.
/help <TOPIC>   - Get help on <TOPIC>. You may type in a topic to search for
                  in place of <TOPIC>. This will fetch information from
                  ballerina-by-example site.
                    eg: /help var
```

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
```diff
@@ -0,0 +1,22 @@
+Ballerina Shell Help
+
+Type a ballerina statement, expression or a declaration.
+
+Imports should be done with prefixes and the modules
+should be available in the classpath.
+To see available imports use /imports.
+
+Following commands are the in-built commands.
```
Kishanthan, line 9: Is it in-built or `builtin commands`? We are mostly using builtin keyword.

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
```diff
@@ -0,0 +1,22 @@
+Ballerina Shell Help
+
+Type a ballerina statement, expression or a declaration.
+
+Imports should be done with prefixes and the modules
+should be available in the classpath.
```
Kishanthan, line 6: This text is not so user friendly. What are we trying to ask the user?

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
```diff
@@ -0,0 +1,22 @@
...
+
+/exit           - Exit the terminal shell.
+/debug          - Toggle debug mode. You will see debug messages.
+/reset          - Reset REPL state.
+                  All previous code snippets will be removed from REPL memory.
+/imports        - List the available imports.
+/vars           - List the declared variables and their values.
+/dclns          - List the user defined module level declarations.
+/help           - Get this message.
+/help <TOPIC>   - Get help on <TOPIC>. You may type in a topic to search for
```
Kishanthan, line 19: Do we allow arbitrary text also here with <TOPIC>? Or is it only the available commands? Then we need to fix the wording here.

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
kdsuneraavinash: This message is outdated. I'll remove this section.

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
kdsuneraavinash: There are a set of available commands but there are a lot of available prefixes. (basically, this command will search for the topic in BBE with the same name and give the output)

I added `help topics` to see available topics. So how about this?

```
Ballerina Shell Help

Type a Java language expression, statement, or declaration.
Or type one of the following commands:

/exit           - Exit the terminal shell.
/debug          - Toggle debug mode. You will see debug messages.
/reset          - Reset REPL state.
                  All previous code snippets will be removed from REPL memory.
/imports        - List the available imports.
/vars           - List the declared variables and their values.
/dclns          - List the user defined module level declarations.
/help           - Get this message.
/help <TOPIC>   - Get help on <TOPIC>. eg: /help var
/help topics    - See all available topic names.
```

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
kdsuneraavinash: I rename to built-in commands

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
Kishanthan: Is it actually correct to call it TOPIC? That is my question. Can you post some examples and their output here?

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -0,0 +1,33 @@
+NAME
+       ballerina-shell - Run ballerina interactive REPL
+
+SYNOPSIS
+       bal shell [-d|--debug] [-f|--force-dumb]  [-t|--time-out]
+
+
+DESCRIPTION
+       Shell runs a REPL instance of ballerina to enable users to run small
```
MaryamZi, line 9:
```suggestion
       Shell runs a REPL instance of Ballerina to enable users to run small
```

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -0,0 +1,33 @@
...
+       snippets to code to check their outputs. Please not that this is an
```
MaryamZi, line 10:
```suggestion
       snippets of code and check the output. Please note that this is an
```

Do we need "check the output" part at all?

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -0,0 +1,33 @@
...
+       experimental feature.
+
+       Debug messages can be enabled via -d parameter.
```
MaryamZi, line 13:
```suggestion
       Debug messages can be enabled using the '-d' option.
```

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -0,0 +1,33 @@
...
+
+OPTIONS
+       -d
+           Enable debug mode from the beginning. Not enabled by default.
```
MaryamZi, line 17: Do we need the " from the beginning" part? This cannot be enabled after, right?
```suggestion
           Enable debug mode. Disabled by default.
```

MaryamZi: Also I don't think we need to mention the default for debug and force-dumb. The options indicate that we're toggling it on.

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -0,0 +1,33 @@
...
+
+DESCRIPTION
+       Shell runs a REPL instance of ballerina to enable users to run small
+       snippets to code to check their outputs. Please not that this is an
+       experimental feature.
+
+       Debug messages can be enabled via -d parameter.
+
+OPTIONS
+       -d
```
MaryamZi, line 16:
```suggestion
       -d, --debug
```

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -0,0 +1,33 @@
...
+           Enable debug mode from the beginning. Not enabled by default.
+       -f
```
MaryamZi, line 18:
```suggestion
       -f, --force-dumb
```

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -0,0 +1,33 @@
...
+           Force dumb terminal mode. Not dumb by default.
+       -t <time-out-duration>
```
MaryamZi, line 20:
```suggestion
       -t, --time-out
```

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -0,0 +1,33 @@
...
+           Change tree parsing timeout. Default is 100ms. Value should be
```
MaryamZi, line 21:
```suggestion
           Set tree parsing timeout. Default is 100ms. Value should be
```

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -0,0 +1,33 @@
...
+           given in milliseconds. Increase in slow devices.
```
MaryamZi, line 22: Do we need "Increase in slow devices."? Not sure if it belongs in the help text.

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
MaryamZi, line 16: We don't usually use short options except for maybe help. Given that this is for REPL, if we think they'll be used often, I guess it's OK to include them. Or for consistency, we can stick to the long options.

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
kdsuneraavinash: The user can enable by debug mode later using `/debug` command while the shell is running. I'll remove the default value mention.

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
kdsuneraavinash: I'll remove it. I included this because the timeout *can* cause some problems in old hardware.

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
kdsuneraavinash:
```
=$ /help var                                                                                                                                    
| Ballerina allows you to declare variables using the `var` keyword instead of specifying a type.
| The type of the declared variable is inferred from the expression on the right-hand side.

=$ /help enums                                                                                                                                   
| An enum member is defined as a constant. An enum member can explicitly specify a constant expression
| as an associated value. If specified, the result of evaluating the relevant expression must be a string. If the
| constant expression is omitted, it defaults to be the same as the identifier.
```

The current help provider is a simple implementation that basically fetches the entry from BBE for the word given.

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
MaryamZi: Oh okay. But then do we need this option at all? The user can do `/debug` at the very beginning? Is there anything different between the two?

ballerina-shell/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+The Ballerina-shell tool is an interactive tool for learning the Ballerina programming language and prototyping Ballerina code. Ballerina-shell is a Read-Evaluate-Print Loop (REPL), which evaluates declarations, statements, and expressions as they are entered and immediately shows the results. Currently, the tool is run via the command line. Using Ballerina-shell, you can enter program statements one at a time and immediately see the result.
+
+## Demo
+
+![Recording](./docs/demo.gif)
+
+## Modules
+
+The project is implemented in two base modules.
```
MaryamZi, line 13:
```suggestion
The project is implemented in three base modules.
```
?

ballerina-shell/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+- **shell-rt** - Module including runtime dependencies for ballerina programs generated. You may find the source code for this
+  module [here](shell-rt).
+- **shell-core** - Module including all the base evaluation classes. This has all the base components to evaluate and run a
+  string. All other components are built on top of this module. You may find the source code for this
+  module [here](shell-core).
+- **shell-cli** - A command-line interface built on top of shell. Includes multi-line inputs, color-coded outputs,
+  keyword-based auto-completion, etc... You may find the source code for this module [here](shell-cli).
+
+## Known Issues
```
MaryamZi, line 23: Have we created issues for these?

ballerina-shell/README.md
```diff
@@ -0,0 +1,130 @@
...
+- **Enum definition with explicit expression is not fully supported** - Enum declarations with explicit expressions are
+  not supported fully. Some functionalities may not work properly. Following is an example.
+
+  ```ballerina
+  // This will not work due to type EN converting to english
+  enum Language {EN="english", TA="tamil", SI="sinhala"}
+  EN variable = "english"  // <- Error
+  ```
+
+- **Support with external libraries not tested completely** - Even though it is possible to run this against external libraries, this is not fully tested. Some issues, especially relating to invincible types, may occur. For an example following is a possible problematic scenario.
```
MaryamZi, line 45: Invisible types?
```suggestion
- **Support with external libraries not tested completely** - Even though it is possible to run this against external libraries, this is not fully tested. Some issues, especially relating to types that are not visible outside the library, may occur. For an example following is a possible problematic scenario.
```

ballerina-shell/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+  ```ballerina
+  import abc/pqr
+  // If pqr:call() returns a value of which the type is private, following will fail.
```
MaryamZi, line 49: Since we don't have private types shall we say
```suggestion
  // If pqr:call() returns a value of which the type is not public, following will fail.
```

ballerina-shell/README.md
```diff
@@ -0,0 +1,130 @@
...
+- **Enum type signature should be the type symbol name** - If an enum is defined with expression values, enum elements'
+  type signature would be the expression. However, this should instead be the original enum name.
+
+  ```ballerina
+  enum Language {EN="english", TA="tamil", SI="sinhala"}
+  // Type signature of variable would be `english` but should instead be `EN`.
+  EN variable = "english" 
+  ```
+
+- **Type signature of arrays with a fixed length should reflect the length** - Any array definition with a fixed length
```
MaryamZi, line 64: Do we have issues for these?

ballerina-shell/README.md
```diff
@@ -0,0 +1,130 @@
...
+  [int, int, int] [a, b, c] = variable
+  ```
+
+### Notes
+
+- **The parser is imperfect** - Current parser is imperfect and is sometimes unable to detect the type of statement.
+  Please file an issue if you come across any wrong categorization of a snippet. The parser is also relatively slow
+  compared to the compilation phase, acting as a bottle-neck. So a timeout is employed to stop invalid statement parsing
+  from taking too much time. However, this might cause issues in some old hardware where the execution might be slower
+  than expected (where even valid executions might exceed the timeout).
```
MaryamZi, line 81: We can maybe add something here about using the `--time-out` option?

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
kdsuneraavinash: The user would miss the initialization debug messages (if there are any that is) if they toggled it later.

ballerina-shell/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+- **Type guards for global variables do not work** - Since all the variables defined in REPL top level act as global
+  variables, type guards won't work. To remedy this you can explicitly cast the variable to the required type. However,
+  note that this is the expected behavior and not a bug. This will be fixed once the language supports global type
+  guards.
```
MaryamZi, line 86: From the language POV, AFAIK this will not change. Since global variables can be updated in different functions, we cannot guarantee that a type guard will hold.

I think we should move this to "Known Issues" and just mention that type guards won't work because everything is defined as a global variable.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BShellConfiguration.java
```diff
@@ -0,0 +1,207 @@
...
+     * @param mode Mode to create the evaluator on.
+     * @return Created evaluator.
+     */
+    private Evaluator createEvaluator(EvaluatorMode mode) {
+        if (mode == EvaluatorMode.DEFAULT) {
+            return new EvaluatorBuilder()
+                    .treeParser(TrialTreeParser.defaultParser(treeParsingTimeout))
+                    .build();
+        }
+        throw new RuntimeException("Unknown mode given.");
```
MaryamZi, line 63: Will probably be useful to include the invalid mode too.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BShellConfiguration.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /**
+     * Creates and returns an evaluator based on the config.
+     *
+     * @param mode Mode to create the evaluator on.
+     * @return Created evaluator.
+     */
+    private Evaluator createEvaluator(EvaluatorMode mode) {
+        if (mode == EvaluatorMode.DEFAULT) {
```
MaryamZi, line 58: Can this be anything else right now?

kdsuneraavinash: No, at the moment only that mode is available.

ballerina-shell/README.md
kdsuneraavinash: OK

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BShellConfiguration.java
kdsuneraavinash: Previously there was another mode implementation (when I was trying out different implementations) but I decided to keep the mode in because it would be easier, later on, to implement a different implementation/configurations.

ballerina-shell/README.md
kdsuneraavinash: Typo. Previously it was only 2 modules and I think I forgot to update it. I'll fix this.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/NoExitVmSecManager.java
```diff
@@ -0,0 +1,78 @@
...
+package io.ballerina.shell.invoker.classload;
+
+import java.security.Permission;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A helper class for class load invoker so that
+ * the system.exit calls will not exit the VM.
+ * TODO: Remove this and fix VM exit bug.
```
Kishanthan, line 28: Rather than a custom sec manager, which is an overkill, I would prefer to fix this at the actual place. Say if we are starting the shell, then we don't have to exit the program. We could include this logic to the backend generation phase.

ballerina-shell/README.md
kdsuneraavinash: I just created issues for them.
#28036 #28037 #28038 #28039 #28042

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/NoExitVmSecManager.java
kdsuneraavinash: I created an issue for this #28048

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
MaryamZi: I was wrong here in removing `<time-out-duration>`.

Shall we make this

```
       -t <time-out-duration>, --time-out <time-out-duration>
```

kdsuneraavinash: OK

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
kdsuneraavinash: Should I leave short options as is then?

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
kdsuneraavinash: I'll remove the "check the output" part

Kishanthan:
@kdsuneraavinash , could also do a verification by updating to the latest master? The ballerina command is renamed to bal and just want to make sure that has no impact here.

kdsuneraavinash:
@Kishanthan updated

Kishanthan:
I was trying out locally. Is the following an issue? 

```

=$ int a = 4;                                                                                                                                                                                        took 357 ms

=$ a                                                                                                                                                                                                   took 0 ms
4

=$ int b = 0;                                                                                                                                                                                          took 0 ms

=$ a + b;                                                                                                                                                                                            took 221 ms
4

=$ int c = a + b;                                                                                                                                                                                    took 324 ms

=$                                                                                                                                                                                                   took 196 ms

=$ c                                                                                                                                                                                                   took 0 ms
4

=$                                                                                                                                                                                                   took 167 ms

=$ b = 3;                                                                                                                                                                                              took 0 ms

=$ c                                                                                                                                                                                                 took 180 ms
4
```

Kishanthan:
And this

```
=$ int a = 1;                                                                                                                                                                                          took 0 ms

=$ int b = 3;                                                                                                                                                                                        took 218 ms

=$ int c = a + b;                                                                                                                                                                                    took 158 ms

=$                                                                                                                                                                                                   took 154 ms

=$                                                                                                                                                                                                     took 0 ms

=$ c                                                                                                                                                                                                   took 0 ms
4

=$                                                                                                                                                                                                   took 145 ms

=$                                                                                                                                                                                                     took 0 ms

=$                                                                                                                                                                                                     took 0 ms

=$ c =+ 1;                                                                                                                                                                                             took 0 ms

=$ c                                                                                                                                                                                                 took 137 ms
1
```

Kishanthan:
> And this
> 
> ```
> =$ int a = 1;                                                                                                                                                                                          took 0 ms
> 
> =$ int b = 3;                                                                                                                                                                                        took 218 ms
> 
> =$ int c = a + b;                                                                                                                                                                                    took 158 ms
> 
> =$                                                                                                                                                                                                   took 154 ms
> 
> =$                                                                                                                                                                                                     took 0 ms
> 
> =$ c                                                                                                                                                                                                   took 0 ms
> 4
> 
> =$                                                                                                                                                                                                   took 145 ms
> 
> =$                                                                                                                                                                                                     took 0 ms
> 
> =$                                                                                                                                                                                                     took 0 ms
> 
> =$ c =+ 1;                                                                                                                                                                                             took 0 ms
> 
> =$ c                                                                                                                                                                                                 took 137 ms
> 1
> ```

This is correct.

Kishanthan:
> I was trying out locally. Is the following an issue?
> 
> ```
> 
> =$ int a = 4;                                                                                                                                                                                        took 357 ms
> 
> =$ a                                                                                                                                                                                                   took 0 ms
> 4
> 
> =$ int b = 0;                                                                                                                                                                                          took 0 ms
> 
> =$ a + b;                                                                                                                                                                                            took 221 ms
> 4
> 
> =$ int c = a + b;                                                                                                                                                                                    took 324 ms
> 
> =$                                                                                                                                                                                                   took 196 ms
> 
> =$ c                                                                                                                                                                                                   took 0 ms
> 4
> 
> =$                                                                                                                                                                                                   took 167 ms
> 
> =$ b = 3;                                                                                                                                                                                              took 0 ms
> 
> =$ c                                                                                                                                                                                                 took 180 ms
> 4
> ```

This is correct. c should still refer to the old value of b

Kishanthan:
```
=$ /q                                                                                                                                                                                                took 284 ms
| Tree parsing failed: invalid token '/'
| /q;
|  ^
| Parsing aborted because of errors.

```

We should hide internal errors. Shall we create an issue to improve these?

kdsuneraavinash:
Sure, that is actually not an internal error, but a syntax error. I think the error messages should be improved to show it as such. I created an issue for this #28066

Kishanthan (dismissed):
OK to merge for alpha pack.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BShellConfiguration.java
MaryamZi: Since we have only one mode atm, IMO, we should refactor this. We can add back the checks and the enum when we are introducing more modes, WDYT? Right now, we're doing an unnecessary check that will always evaluate to true.

We can decide on this and introduce any changes separately.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BallerinaShell.java
```diff
@@ -0,0 +1,201 @@
...
+            return;
+        }
+
+        if (diagnostic.getKind() == DiagnosticKind.ERROR) {
+            terminal.error(diagnostic.toString());
+        } else if (diagnostic.getKind() == DiagnosticKind.WARN) {
+            terminal.warn(diagnostic.toString());
+        } else {
+            terminal.debug(diagnostic.toString());
+        }
```
MaryamZi, line 140: We can extract `diagnostic.getKind()` out to a variable.

Also, we usually try to return early. e.g,
```suggestion
        DiagnosticKind kind = diagnostic.getKind();
        if (kind == DiagnosticKind.DEBUG) {
            if (!configuration.isDebug()) {
                return;
            }

            terminal.debug(diagnostic.toString());
            return;
        }

        if (kind == DiagnosticKind.ERROR) {
            terminal.error(diagnostic.toString());
            return;
        }
        
        terminal.warn(diagnostic.toString());
```

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BallerinaShell.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /**
+     * Output a diagnostic to the terminal.
+     *
+     * @param diagnostic Diagnostic to output.
+     */
+    protected void outputDiagnostic(Diagnostic diagnostic) {
+        if (diagnostic.getKind() == DiagnosticKind.DEBUG &&
```
MaryamZi, line 129: We can handle the debug case here itself, right? Or we unnecessarily check if `diagnostic.getKind() == DiagnosticKind.ERROR` and `diagnostic.getKind() == DiagnosticKind.WARN` again.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BallerinaShell.java
MaryamZi: May apply elsewhere too.

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
MaryamZi: Yeah this is OK for the time-being.

cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
MaryamZi: Yeah, then we'll have to have both.

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
```diff
@@ -0,0 +1,15 @@
+Ballerina Shell Help
+
+Type a Java language expression, statement, or declaration.
```
MaryamZi, line 3: Ballerina?

kdsuneraavinash: Typo. Fixed.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BallerinaShell.java
kdsuneraavinash: I refactored as following,
```java
    protected void outputDiagnostic(Diagnostic diagnostic) {
        DiagnosticKind diagnosticKind = diagnostic.getKind();
        if (diagnosticKind == DiagnosticKind.DEBUG) {
            if (configuration.isDebug()) {
                terminal.debug(diagnostic.toString());
            }
            return;
        }

        if (diagnosticKind == DiagnosticKind.ERROR) {
            terminal.error(diagnostic.toString());
        } else {
            terminal.warn(diagnostic.toString());
        }
    }
```

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BallerinaShell.java
kdsuneraavinash: I refactored the mode out.

MaryamZi (approved):
Reviewed command and help text related changes and tried out several scenarios. LGTM.

Kishanthan (approved):

PR #27861, opened 2021-01-14
Extract resource name for http filters

ldclakmal:
## Purpose
The `resourceNode.name.value` is evaluated as the resource name concatenated with the `accessor` and `$` symbols. In order to engage HTTP filters, the exact resource name should be desugard in to the HTTP filter context.

Example 1:
`resource function get sayHello` is evaluated as `"$get$sayHello"` and converted by the PR as `"sayHello"`

Example 2:
`resource function myAccessor sayHello` is evaluated as `"$myAccessor$sayHello"` and converted by the PR as `"sayHello"`

Example 3:
`resource function my\$Accessor say\$Hello` is evaluated as `"$my\$Accessor$say\$Hello"` and converted by the PR as `"sayHello"`

Related to https://github.com/ballerina-platform/ballerina-lang/pull/27820
Related to https://github.com/ballerina-platform/ballerina-standard-library/issues/584

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/HttpFiltersDesugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/HttpFiltersDesugar.java
@@ -53,6 +53,7 @@ import org.wso2.ballerinalang.compiler.tree.BLangFunctionBody;
 import org.wso2.ballerinalang.compiler.tree.BLangIdentifier;
 import org.wso2.ballerinalang.compiler.tree.BLangImportPackage;
 import org.wso2.ballerinalang.compiler.tree.BLangNode;
+import org.wso2.ballerinalang.compiler.tree.BLangResourceFunction;
 import org.wso2.ballerinalang.compiler.tree.BLangSimpleVariable;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangBinaryExpr;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangExpression;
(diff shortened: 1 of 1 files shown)


chamil321 (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/HttpFiltersDesugar.java
```diff
@@ -233,6 +234,11 @@ private BLangSimpleVariable addFilterContextCreation(BLangFunction resourceNode,
         return filterContextVar;
     }
 
+    private String extractResourceName(BLangFunction resourceNode) {
```
rdhananjaya, line 237: Is this method supposed to extract the resource paths?
say we have a resource function such as `resource function get foo/bar/[int i]/[int j]() { }` what is the expected behavior?

ldclakmal: The objective of this function is to extract the resource name by removing the accessor part attached to it. The expected value for the given resource signature is `"foo$bar$*$*"` and this function returns that.

rdhananjaya: Ack, I was confused with the term `resourceName` as resource functions doesn't have a name anymore. Shall we concatenate the text parts of [this](https://github.com/ballerina-platform/ballerina-lang/blob/70ed2d553613073ed0bfd4ace1fcc20bd5f56755/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/BLangResourceFunction.java#L33) instead of removing from the generated string.

ldclakmal:
The `resourceNode.name.value` is evaluating is correct according to the new service type changes. It should be fixed at `ballerina/reflect` module.
Related issue: https://github.com/ballerina-platform/ballerina-standard-library/issues/837

PR #27890, opened 2021-01-16
Refactor symbols to include name() only for symbols with names

pubudu91:
## Purpose
Previously, `name()` was defined in `Symbol`. With this PR, we are clearly marking the distinction between named symbols and ones without a name (e.g., type symbols in general). This forces the user to be mindful of when `name()` is used and ensures that you rely on `name()` really for symbols associated with a name.

## Approach
- `name()` moved to a separate interface called `Identifiable`
- `moduleId()` and `location()` moved to a new interface called `NamedSymbol` which extends from both `Symbol` and `Identifiable`.

The following shows how the new design looks like, using a few of the symbols.

named-symbol

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


pubudu91:
Closing this for now since we decided to defer this.

PR #27911, opened 2021-01-18
Fix signature() of union type symbol

pubudu91:
## Purpose
Fixes a bug in the `signature()` method in union type symbol

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaUnionTypeSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaUnionTypeSymbol.java
@@ -49,6 +49,7 @@ public class BallerinaUnionTypeSymbol extends AbstractTypeSymbol implements Unio
     private static final Pattern pCloneableType = Pattern.compile(CLONEABLE_TYPE);
 
     private List<TypeSymbol> memberTypes;
+    private String signature;
 
     public BallerinaUnionTypeSymbol(CompilerContext context, ModuleID moduleID, BUnionType unionType) {
         super(context, TypeDescKind.UNION, moduleID, unionType);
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaUnionTypeSymbol.java
```diff
@@ -97,4 +115,18 @@ public String signature() {
             return joiner.toString();
         }
     }
+
+    private String getSignatureForFiniteType() {
+        List<TypeSymbol> memberTypes = this.memberTypeDescriptors();
+        if (memberTypes.size() == 2 &&
+                memberTypes.get(1).typeKind() == TypeDescKind.NIL) {
```
hasithaa, line 122: is it possible to have NIL in the first element?

hasithaa (approved):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaUnionTypeSymbol.java
pubudu91: Ah no.. Can't really assume that. Was wondering whether to get rid of this altogether for finite types :thinking:

pubudu91: Even for unions, should we always print unions which contain `()` as `T1|T2|T3?`?

hasithaa: That is a readability issue. For optional types saying `T?` is more natural than saying `T|()` . But When we have more I find `T1|T2|T3|()` is more natural than `T1|T2|T3?`

pubudu91: Yeah, true. Removed this for finite types and fixed the optional type logic for unions and updated.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaUnionTypeSymbol.java
```diff
@@ -84,9 +102,10 @@ public String signature() {
             return "...";
         }
         List<TypeSymbol> memberTypes = this.memberTypeDescriptors();
-        if (memberTypes.size() == 2 &&
-                memberTypes.get(1).typeKind() == TypeDescKind.NIL) {
-            return memberTypes.get(0).signature() + "?";
+        if (memberTypes.size() == 2) {
+            TypeSymbol member1 = memberTypes.get(0);
+            return member1.typeKind() == TypeDescKind.NIL ?
+                    memberTypes.get(1).signature() + "?" : member1.signature() + "?";
```
hasithaa, line 108: Is this correct, what if T1|T2 where T2 is not nil,

pubudu91: Fixed this with e53c44b

hasithaa (approved):

Issue #27913, opened 2021-01-18
Call stack shows multiple thread frames in VS Code debug view


praveennadarajah:
**Description:**
$subject

**Steps to reproduce:**
- Consider the below code segment
```
public function main() {
    int x = 5;
}
```
- Add a debug breakpoint inside the main method.
- Observe the call stack in debug view. 
- Following call stack will be observed:
![image](https://user-images.githubusercontent.com/11292766/104891439-f5477700-5996-11eb-9e20-9363b6f0bbcf.png)

- Expected call stack view:
![image](https://user-images.githubusercontent.com/11292766/104891647-3d669980-5997-11eb-9ff9-eb7d4468cd29.png)

**Affected Versions:**
```
Ballerina Swan Lake Preview 9 SNAPSHOT
Language specification v2020-12-17
Update Tool 0.8.11-SNAPSHOT
```

praveennadarajah:
Similar issue https://github.com/ballerina-platform/ballerina-lang/issues/28089

PR #27914, opened 2021-01-18
Fix call stack shows multiple thread frames issue

praveennadarajah:
## Purpose
This PR will fix call stack shows multiple thread frames issues

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/27913


praveennadarajah:
> This `isAtBreakpoint()` check had been there and AFAIR you have removed it with [8b73a2e](https://github.com/ballerina-platform/ballerina-lang/commit/8b73a2e18d613bc7a4f5fc03d520cbc52cc63900) sometimes back when fixing debugger integration tests. I'm just wondering why we had to remove it, just to avoid any regressions due to this change.

Earlier we had both `threadReference.isSuspended()` and `threadReference.isAtBreakpoint()` checks. When both checks are there together, we had issues with STEP_OVER operation as it is not working as expected. Then we removed `threadReference.isAtBreakpoint()` check. Now, I'm replacing the `threadReference.isSuspended()` check with `threadReference.isAtBreakpoint()` to overcome the $subject. As we have enabled all the tests now, I believe this won't lead to any regressions.

praveennadarajah:
With the above change, `STEP_IN` and `STEP_OUT` operations are not working as expected. We will have to find a proper fix for this.

Issue #27845, opened 2021-01-13
Compiler crashes for inclusion in `readonly` class


dulvinw:
**Description:**
$title

FTM produce a bad sad.

MaryamZi:
The currently commented out `testComplexIntersectionsInUnions` test in `test_intersection_type.bal`, fails due to this after changes from `readonly` to `final` for object fields.

```ballerina
type IntermediateInboundHandler object {
    int allow;
};

readonly class InboundHandlerImpl {
    *IntermediateInboundHandler;

    public function init(int allow) {
        self.allow = allow;
    }
}
```

```log
[2021-01-18 19:47:19,840] SEVERE {b7a.log.crash} - null 
java.util.NoSuchElementException
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:1000)
	at org.wso2.ballerinalang.compiler.util.ImmutableTypeCloner.markFieldsAsImmutable(ImmutableTypeCloner.java:126)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.setReadOnlynessOfClassDef(SymbolEnter.java:2365)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.validateFieldsAndSetReadOnlyType(SymbolEnter.java:2267)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.defineUndefinedReadOnlyTypes(SymbolEnter.java:2226)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.defineConstructs(SymbolEnter.java:395)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.visit(SymbolEnter.java:297)
	at org.wso2.ballerinalang.compiler.tree.BLangPackage.accept(BLangPackage.java:163)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.defineNode(SymbolEnter.java:259)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.definePackage(SymbolEnter.java:252)
	at io.ballerina.projects.ModuleContext.compileInternal(ModuleContext.java:342)
	at io.ballerina.projects.ModuleCompilationState$1.compile(ModuleCompilationState.java:45)
	at io.ballerina.projects.ModuleCompilationState$1.generatePlatformSpecificCode(ModuleCompilationState.java:53)
[...]

Issue #27501, opened 2020-12-15
Support intersection types as a type reference in an object type inclusion


pcnfernando:
**Description:**
Following sample throws an incompatible type exception: `incompatible types: 'Timestamp' is not an object`.

**Steps to reproduce:**
Sample:
```ballerina
public type Timestamp readonly & object {
    public function toMillisecondsInt() returns int;
    public function toString() returns string;
};

class TimestampImpl  {
    *Timestamp;
    public function toMillisecondsInt() returns int {
        return 0;
    }
    public function toString() returns string {
        return "test";
    }
}

```
**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

MaryamZi:
Spec discussion - https://github.com/ballerina-platform/ballerina-spec/issues/649#issuecomment-745788381 onward.

MaryamZi:
Even though inclusion will be allowed with a `readonly` and object intersection, according to the spec, this will
- not be allowed in an object type descriptor
- will be allowed in a class only if the class is also `readonly`

PR #27921, opened 2021-01-18
Allow object type inclusion with `readonly` and object intersections

MaryamZi:
## Purpose
$title, where the inclusion is done in a `readonly` class.

```ballerina
public type Foo readonly & object {
    int x;

    public function foo() returns int;
};

readonly class Bar  {
    *Foo;

    function init(int x) {
        self.x = x;
    }

    public function foo() returns int => self.x;
}
```

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/27501 - Support intersection types as a type reference in an object type inclusion

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/27845 - Compiler crashes for inclusion in `readonly` class 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1644,6 +1644,12 @@ error.invalid.inclusion.with.mismatched.qualifiers=\
 error.invalid.reference.with.mismatched.qualifiers=\
   invalid type reference: missing ''{0}'' qualifier(s) in the referencing object constructor expression
 
+error.invalid.read.only.typedesc.inclusion.in.object.typedesc=\
+  object type inclusion cannot be used with a ''readonly'' type descriptor in an ''object'' type descriptor
+
+error.invalid.read.only.typedesc.inclusion.in.non.read.only.class=\
+  object type inclusion cannot be used with a ''readonly'' type descriptor in a ''class'' that is not ''readonly''
```
KRVPerera, line 1651: I think this message is unclear. When we say `that is not` here that means the class including the object. But we don't say we cannot include because we are trying to include a readonly object.
Isn't it clear just saying something similar to "Immutable type descriptor can only include immutable types"

KRVPerera: Or can we put the readonly type descriptor name here.

`object type inclusion cannot be used with a 'readonly' type descriptor in a 'class' that is not 'readonly'`

`object type inclusion cannot be used with a 'readonly' type descriptor `Timestamp` in a 'class' that is not 'readonly'`

```Ballerina
public type Timestamp readonly & object {
    int value;

    public function toMillisecondsInt() returns int;
    public function toString() returns string;
};

class TimestampImpl  {
    *Timestamp;

    function init(decimal value) {
        self.value = <int> value;
    }

    public function toMillisecondsInt() returns int => self.value;

    public function toString() returns string => "test";
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2414,15 +2494,31 @@ private void validateIntersectionTypeDefinitions(List<BLangTypeDefinition> typeD
...
         // Any newly added typedefs are due to `T & readonly` typed fields. Once the fields are set for all
         // type-definitions we can revisit the newly added type-definitions and define the fields and members for them.
         populateImmutableTypeFieldsAndMembers(typeDefNodes, pkgEnv);
 
         // If all the fields of a structure are readonly or final, mark the structure type itself as readonly.
         // If the type is a `readonly object` validate if all fields are compatible.
-        validateFieldsAndSetReadOnlyType(typDefs, pkgEnv);
+        validateFieldsAndSetReadOnlyType(typeDefs, pkgEnv);
+
+        for (BLangNode typeDef : typeDefs) {
```
rdhananjaya, line 2507: This could be a method right, like `populateImmutableTypeFieldsAndMembers` and `validateFieldsAndSetReadOnlyType`.

compiler/ballerina-lang/src/main/resources/compiler.properties
MaryamZi: > I think this message is unclear. When we say `that is not` here that means the class including the object. But we don't say we cannot include because we are trying to include a readonly object.

It's an error specifically because it is trying to include `readonly` that is part of an intersection, which I'm trying to convey via the "object type inclusion cannot be used with a ''readonly'' type descriptor". 

> Isn't it clear just saying something similar to "Immutable type descriptor can only include immutable types"

Doesn't this imply that we cannot use inclusion with a mutable type in a class that is immutable? But that is not the case. 

> Or can we put the readonly type descriptor name here.

Since the error is specifically for `readonly` via `Timestamp` and not directly for `Timestamp`, IMO it's not quite accurate to include the name.

rdhananjaya (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -556,73 +586,123 @@ private void defineReferencedClassFields(BLangClassDefinition classDefinition, S
             fieldNames.put(fieldVariable.name.value, fieldVariable);
         }
 
-        classDefinition.referencedFields = classDefinition.typeRefs.stream().flatMap(typeRef -> {
+        List<BLangSimpleVariable> referencedFields = new ArrayList<>();
```
rdhananjaya, line 589: This is not about this PR but I think we should consider refactoring this method, maybe into a some sort of method object.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
MaryamZi: Fixed in 6b923b6.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
MaryamZi: We also need to revisit the field/method population logic for intersection types (https://github.com/ballerina-platform/ballerina-lang/issues/27948), which will probably help clean this up a bit too.

compiler/ballerina-lang/src/main/resources/compiler.properties
MaryamZi: Let's change the error in a separate PR, if we can come up with a better error message.

PR #27928, opened 2021-01-18
Update project and packages APIs with project kind

prabushi:
## Purpose
Update `ballerinaDocument/project` and `ballerinaProject/packages` requests' LS APIs to consider project kind.

Fixes https://github.com/ballerina-platform/plugin-vscode/issues/139

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/project/BallerinaProjectServiceImpl.java
```diff
@@ -53,7 +55,12 @@
      * @return {@link JsonArray} Array of packages
      */
     private JsonArray createJSONResponse(String sourceRoot) {
-        Package currentPackage = ProjectLoader.loadProject(Paths.get(URI.create(sourceRoot))).currentPackage();
+
+        Project project = ProjectLoader.loadProject(Paths.get(URI.create(sourceRoot)));
+        if (project.kind() == ProjectKind.SINGLE_FILE_PROJECT) {
```
nadeeshaan, line 60: Instead of returning an empty `jsonArray`, what if we add the kind as a property and at the consumer side of the API, change the logic based on that? I think that approach will be reliable, which is to depend on an available property rather than an empty object. WDYT?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/project/BallerinaProjectServiceImpl.java
prabushi: True that it'll have more info in the response. But, this API is '/packages' and returns an array of packages. Project kind doesn't relate to a package. And, we already have /project API to get the kind. Wouldn't this approach be redundant?

prabushi: Removed BallerinaProjectService with https://github.com/ballerina-platform/ballerina-lang/pull/27928/commits/ef323d436273c9330aa5e31a3bd5bddbcb3bc72e

nadeeshaan: No longer valid now, after removing the particular APIs

nadeeshaan (approved):

PR #27933, opened 2021-01-18
Refactor error handler rules for qualifiers

### LINKED ISSUE #25300: Revisit error recovery for function qualifiers
**Description:**
Need to revisit error recovery for function qualifiers since

1. We have new qualifiers like `isolated` and the current model may not be scalable
2. There are several possibly related issues when there are errors
   - Bad, sad error on incomplete service source with `transactional function` - #25290 
   - Bad, sad error for duplicate function qualifiers - #25287 
   - Bad, sad error with missing `function` keyword and invalid qualifiers - #25261

### LINKED ISSUE #26460: Invalid recovery for functions having parameters
**Description:**
Consider following code snippet.
```bal
public foo() returns boolean { }

public foo(int i) returns boolean { }
```

Recovery is
```bal
public  MISSING[function]foo() returns boolean { }

 INVALID[public] foo MISSING[error] MISSING[]( INVALID[int] i)  MISSING[;] MISSING[object] INVALID[returns]  INVALID[boolean] { } MISSING[] MISSING[;]
````
It seems something is wrong with the recovery of functions having parameters. We can expect function keyword to be recovered in both cases.

**Affected Versions:**
Current Master

Team/CompilerFE, Area/Parser, Type/Bug, Priority/High

### PR DESCRIPTION
**Pull request by lochana-chathura**
## Purpose
$subject.

Fixes #25300
Also resolves #26460 as test case is added.

## Approach
Define qualifier rules in the error handler based on the qualifier position rather than the qualifier kind. 

## Samples
N/A

## Remarks
With new approach we expect to reduce the amount of alternative paths traverse by the error handler for qualifiers greatly.

For example:
Assume LOOKAHEAD_LIMIT = 5 and entry point to the error handler is a qualifier starting context.

In Old approach:
2 qualifier case: we traverse 2 + 2x1 + 2x1x1 + 2x1x1x1 + 2x1x1x1x1 = 10 alternative qualifier paths.
4 qualifier case: we traverse 4 + 4x3 + 4x3x3 + 4x3x3x3 + 4x3x3x3x3 = 484 alternative qualifier paths.

In New approach:
2 qualifier case: we traverse 2 alternative qualifier paths.
4 qualifier case: we traverse 4 alternative qualifier paths.

Ref https://github.com/ballerina-platform/ballerina-lang/pull/26888#discussion_r527382327

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### CODE REVIEW

**Comment by lochana-chathura** [comment]
Closing temporarily.

File: compiler/ballerina-parser/src/test/resources/declarations/class-def/class_def_source_35.bal [hunk 1]
```diff
@@ -1,5 +1,6 @@
-class client client client Foo {
-    string name;
+class 
+client client client Foo {
+string name;
```
rdulmina, line 3: what is the need for this change?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/globalvar/GlobalVarNegativeTest.java
```diff
@@ -84,7 +84,8 @@ public void testConfigurableModuleVarDeclNegative() {
         BAssertUtil.validateError(result, i++, "configurable variable cannot be declared with var", 20, 1);
         BAssertUtil.validateError(result, i++, "type of configurable variable must be anydata&readonly", 22, 22);
         BAssertUtil.validateError(result, i++, "type of configurable variable must be anydata&readonly", 24, 14);
-        BAssertUtil.validateError(result, i++, "invalid token 'configurable'", 32, 18);
+        BAssertUtil.validateError(result, i++, "missing close brace token", 31, 1);
+        BAssertUtil.validateError(result, i++, "invalid token '}'", 35, 1);
```
rdulmina, line 88: since local configurable var declaration is the issue here, we should at least give `invalid token 'configurable'` or try to validate after parsing and give an error message like `configurable variable cannot be declared locally`

compiler/ballerina-parser/src/test/resources/declarations/class-def/class_def_source_35.bal
lochana-chathura: Recovered as,
```
class MISSING[] MISSING[{] 
INVALID[client]  INVALID[client]  INVALID[client] Foo  INVALID[{]
 INVALID[string] name;
}
```
and there for is formatted as above.

rdulmina: Okay

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/globalvar/GlobalVarNegativeTest.java
lochana-chathura: +1.  Create an issue #28429

rdulmina (approved):

PR #27934, opened 2021-01-18
Refactor field symbols

### LINKED ISSUE #27310: [Semantic API] Create field symbols for fields
Currently when you use `symbol()` to get the symbol of a field of a record, object etc. what you get is a variable symbol. Ideally, what you should be getting is a `FieldSymbol` since there may be more to be described in a field than in a regular variable.

### LINKED ISSUE #27311: [Semantic API] Separate field symbols for object and record fields
Currently a single `FieldSymbol` API is used for both object and record fields. But it's better if we can have separate field symbols for these since there are properties which aren't common to both types of fields. e.g., optional fields

pubudu91:
## Purpose
This PR addresses several issues we had with field symbols.

Fix #27310 
Fix #27311

## Approach
- Removed `FieldSymbol`
- Introduced 3 separate interfaces instead: `RecordFieldSymbol`, `ObjectFieldSymbol` and `ClassFieldSymbol` (which is an extension of `ObjectFieldSymbol`). These 3 also has their own, corresponding symbol kinds as well: `RECORD_FIELD`, `OBJECT_FIELD` and `CLASS_FIELD`.
- Removed the `qualifier()` method which we had in `FieldSymbol`. Instead, the above mentioned interfaces extends `Qualifiable`.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordFieldSymbol.java
```diff
@@ -124,26 +129,45 @@ public boolean deprecated() {
...
     @Override
     public String signature() {
-        StringBuilder signature = new StringBuilder(this.typeDescriptor().signature() + " " + this.name());
+        if (this.signature != null) {
+            return this.signature;
+        }
+
+        StringJoiner joiner = new StringJoiner(" ");
+
+        if (!this.qualifiers().isEmpty()) {
```
rasika, line 159: this check is redundant

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordFieldSymbol.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
rasika, line 2: We don't need to update year, usually we put initially-committed year here.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaObjectFieldSymbol.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @Override
+    public String signature() {
+        if (this.signature != null) {
+            return this.signature;
+        }
+
+        StringJoiner joiner = new StringJoiner(" ");
+
+        if (!this.qualifiers().isEmpty()) {
```
rasika, line 134: this check is redundant

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaClassFieldSymbol.java
```diff
@@ -0,0 +1,74 @@
...
+        final long symFlags = this.bField.symbol.flags;
+        List<Qualifier> quals = new ArrayList<>();
+
+        if (Symbols.isFlagOn(symFlags, Flags.PUBLIC)) {
+            quals.add(PUBLIC);
+        } else if (Symbols.isFlagOn(symFlags, Flags.PRIVATE)) {
+            quals.add(PRIVATE);
+        }
+
+        if (Symbols.isFlagOn(symFlags, Flags.FINAL)) {
```
rasika, line 67: Can't we combine this `if` check into same if-else ladder?

rasika (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaClassFieldSymbol.java
hasithaa: That is not right, Public/Private and Final are two different things.

pubudu91: In the previous one there's an if-else ladder because there won't be cases where both `public` and `private` appear on the same symbol. But there can be `public final` or `private final` combinations.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaObjectFieldSymbol.java
```diff
@@ -0,0 +1,141 @@
...
+    private List<AnnotationSymbol> annots;
+    private String signature;
+    private boolean deprecated;
+
+    public BallerinaObjectFieldSymbol(CompilerContext context, BField bField, SymbolKind kind) {
+        super(bField.name.value, bField.symbol.pkgID, kind, bField.symbol);
+        this.context = context;
+        this.bField = bField;
+        this.docAttachment = new BallerinaDocumentation(bField.symbol.markdownDocumentation);
+        this.deprecated = Symbols.isFlagOn(bField.symbol.flags, Flags.DEPRECATED);
```
nadeeshaan, line 63: Suggestion: What if we populate these two fields also on demand as we have done for the other symbols such as annotations, typeDescriptor, etc.

nadeeshaan (approved):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaObjectFieldSymbol.java
pubudu91: +1. Will go ahead and merge this for now. Will send a separate PR for improvements.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaClassFieldSymbol.java
rasika: Ops, didn't spot the combinations in the first sight. Thanks for the clarification!

PR #27961, opened 2021-01-19
Fix keyed table to json conversion

mohanvive:

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TableJsonDataSource.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TableJsonDataSource.java
@@ -88,6 +88,7 @@ public class TableJsonDataSource implements JsonDataSource {
         BIterator itr = this.tableValue.getIterator();
         while (itr.hasNext()) {
             TupleValueImpl tupleValue = (TupleValueImpl) itr.next();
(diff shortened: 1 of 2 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TableJsonDataSource.java
```diff
@@ -88,7 +88,7 @@ public Object build() {
         BIterator itr = this.tableValue.getIterator();
         while (itr.hasNext()) {
             TupleValueImpl tupleValue = (TupleValueImpl) itr.next();
-            MapValueImpl record = ((MapValueImpl) tupleValue.get(0));
```
gimantha, line 91: Shall we make a comment saying what this index means?

gimantha (dismissed):

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TableJsonDataSource.java
gimantha: @mohanvive

mohanvive: Added a comment for the opeartion.

grainier (approved):

PR #27993, opened 2021-01-20
Fix path undefined error at file open

prabushi:
## Purpose
Fix the issue of undefined project info when `project' API is called before the project is loaded to the workspace manager.

Fixes https://github.com/ballerina-platform/plugin-vscode/issues/146

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/document/BallerinaDocumentServiceImpl.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/document/BallerinaDocumentServiceImpl.java
@@ -20,6 +20,7 @@ import io.ballerina.compiler.api.SemanticModel;
 import io.ballerina.projects.Document;
 import io.ballerina.projects.Project;
 import io.ballerina.projects.ProjectKind;
+import io.ballerina.projects.directory.ProjectLoader;
 import org.ballerinalang.diagramutil.DiagramUtil;
 import org.ballerinalang.langserver.LSClientLogger;
 import org.ballerinalang.langserver.LSContextOperation;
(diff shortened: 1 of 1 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/document/BallerinaDocumentServiceImpl.java
```diff
@@ -149,14 +150,11 @@ public BallerinaDocumentServiceImpl(WorkspaceManager workspaceManager, LanguageS
...
                 }
-                Optional<Project> project = this.workspaceManager.project(filePath.get());
-                if (project.isEmpty()) {
-                    return ballerinaProject;
-                }
-                ballerinaProject.setPath(project.get().sourceRoot().toString());
-                ProjectKind projectKind = project.get().kind();
+                Project project = ProjectLoader.loadProject(filePath.get());
+                ballerinaProject.setPath(project.sourceRoot().toString());
+                ProjectKind projectKind = project.kind();
```
nadeeshaan, line 155: Why do we use ProjectLoader here? Can't we use the workspace manager to access the project, given the filePath?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/document/BallerinaDocumentServiceImpl.java
prabushi: VSCode has a popup message upon file open, which calls this API before the file is actually open in the editor. The workspace manager doesn't contain the required info at that time. We have to newly load the project to access the info.

nadeeshaan: Cool, +1 to proceed as such

nadeeshaan (approved):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/document/BallerinaDocumentServiceImpl.java
rasika: Can't we get these details through `ProjectPaths.isStandaloneBalFile(path)` and `ProjectPaths.packageRoot(path)` util methods? `ProjectLoader.loadProject(filePath.get())` is a costly operation.

nadeeshaan: In order to get the package name then we have to do some path operations right? If there are any package structure change wouldn't it be an issue?

rasika: Hmm yeah, that can cause problems in future. This is a valid use-case`workspace.project()` api,  for how about extending functionality of `workspace.project()` functionality to load a project if not loaded. Reason is here we load a new project instance and discard it.

PR #28008, opened 2021-01-20
Fix record implicit readonly-ness with inclusion and fix cyclic union comparison

### LINKED ISSUE #27981: Invalid `readonly` flag for record that has non-`readonly` fields via inclusions
**Description:**
$title.

**Steps to reproduce:**
Results in a compilation error for the following.
```ballerina
type CommonResponse record {|
    string mediaType?;
    map<string|string[]> headers?;
    anydata body?;
|};

public type Forbidden record {|
    *CommonResponse;
|};

public type Unauthorized record {|
    *CommonResponse;
    readonly byte[] body;
    readonly boolean valid;
|};

public function main() {
    Unauthorized|Forbidden? x = tryAuthenticate(); // incompatible types: expected '(Unauthorized|Forbidden)?', found '(Unauthorized|Forbidden)?'
}

isolated function tryAuthenticate() returns Unauthorized|Forbidden? {
}
```

**Affected Versions:**
slp9-SNAPSHOT

### LINKED ISSUE #28003: Cyclic union types are incomplete at times
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    record {} x = {};

    io:println(x is record {| int|boolean|decimal|float|string|xml?...; |}); // prints true, should print false
}
```

**Affected Versions:**
slp9-SNAPSHOT

MaryamZi:
## Purpose
This PR fixes
- considering included fields when setting the `readonly` flag for record types with all `readonly` fields
- `readonly` flag not being set for some unions with implicitly `readonly` members
- certain cyclic unions not having all members at runtime
- `equals` check for cyclic unions with the fix for the previous issue

Fixes #27981 - Invalid `readonly` flag for record that has non-`readonly` fields via inclusions
Fixes #28003 - Cyclic union types are incomplete at times

## Remarks
Not convinced that the approach to add the read-only flag to unions at compile-time is the best approach (even though this only happens at times). Will create an issue to revisit this.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -728,7 +728,7 @@ public class Types {
 
         if (!Symbols.isFlagOn(source.flags, Flags.PARAMETERIZED) &&
                 !isInherentlyImmutableType(target) && Symbols.isFlagOn(target.flags, Flags.READONLY) &&
-                !isInherentlyImmutableType(source) && !Symbols.isFlagOn(source.flags, Flags.READONLY)) {
+                !isInherentlyImmutableType(source) && isMutable(source)) {
             return false;
         }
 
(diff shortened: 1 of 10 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -889,6 +889,27 @@ private boolean isAssignable(BType source, BType target, Set<TypePair> unresolve
...
+
+        BUnionType unionType = (BUnionType) type;
+        for (BType memberType : unionType.getMemberTypes()) {
+            if (!Symbols.isFlagOn(memberType.flags, Flags.READONLY)) {
+                return true;
+            }
+        }
+
+        unionType.flags |= Flags.READONLY;
+        unionType.tsymbol.flags |= Flags.READONLY;
```
KRVPerera, line 909: Do we need to fix this later? I think union creation should properly add this flag.

KRVPerera (approved):
LGTM

hasithaa (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
MaryamZi: The union creation actually already handles setting the read-only flag if all the members have the read-only flag set. The issue here was due to implicitly read-only types - for such types the read-only flag may be set after the creation of the union type and thus the read-only flag may not be set even though all the members have it set.

I've created https://github.com/ballerina-platform/ballerina-lang/issues/28012 to track revisiting this.

PR #28010, opened 2021-01-20
Modify return type of fieldDescriptors() and methods() to Map<String, *FieldSymbol> and Map<String, MethodSymbol>

pubudu91:
## Purpose
This PR modifies the return types of the `fieldDescriptors()` method in records, objects and classes to return `Map<String, *FieldSymbol>` instead of a `List<*FieldSymbol>`. This is because a common use case is to lookup fields in the aforementioned types by name and previously the user had to loop through the fields list and filter to get the relevant field symbol. The map used for this is a `LinkedHashMap` so that the original order is preserved when iterating.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


Kishanthan (commented):
LGTM

nadeeshaan (approved):
Looks good. Hence Approving

PR #27596, opened 2020-12-20
Add support for mapping type rest arguments

chiranSachintha:
- Update record equivalency rules for optional fields.
According to the spec(`a value belongs to a type if it looks like the type, and it will necessarily continue to look like the type no matter how the value is mutated`) this type of assignment can be done. Its has been done with this. 
```ballerina
type Person record {|
    string name;
    int id;
    float avg?;
|};

type Student record {|
    string name;
    int id;
|};

public function main() {
    Student st = {name: "Amy", id: 1234};
    Person pr = st;
}
```

- Allow closed record as rest argument.
```ballerina
type Foo record {|
    int c = 10;
    int d?;
|};

function getSum(int a, int b, int c, int d = 1) returns int {
    return a + b + c + d;
}

function testFunctionWithMappingTypeRestArg() {
    Foo f = {};
    Foo f1 = {c:20, d:15};
    int sum = getSum(10, 15, ...f); // case 1
    int sum1 = getSum(10, 15, ...f1); // case 2
}
```
For the `case 1` instance need to check whether value for `d` is given or not(if value for `d` is given use that value if not use default value). Change the argument representation in the BIR to do that. Futher, keep one of the below state for argument in BIR.
- User provided
- User not provided
- User conditionally provided(relate to the case 1)

**Further information relate to Implementation design has been attached with** [this](https://docs.google.com/document/d/1hnvWui5uLIuzuFmR4ONqrMQjq2tGn0Fpw0953uAYHGg/edit).

Fixes #21097 
Fixes #23755

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #28011, opened 2021-01-20
Update type checker to use context information when deriving type of a let expression

### LINKED ISSUE #26906: Let expressions don't seem to be using the contextually-expected type for type-checking
**Description:**
$title. Results in the actual expression being type-checked with `noType`, resulting in errors later on.

**Steps to reproduce:**
```ballerina
public function main() {
    record {
        int i;
        int j?;
    } rec = let int v = 1 in {i: v};
}
```

Compilation fails for this with

```cmd
$ ballerina run qwe.bal 
Compiling source
        qwe.bal
error: .::qwe.bal:5:13: incompatible types: expected 'record {| int i; int j?; anydata...; |}', found 'record {| int i; |}'
```

**Affected Versions:**
slp5

IMS94:
## Purpose
This fixes #26906. 

## Approach
- `TypeCheker` was using `noType` as the expected type of the expression when an anonymous record is defined inline using a let expression. Updated the `TypeChecker.visit(BLangExpression)` method to use the expected type from let expression when checking the expression instead of `noType`.
- Added a test case to cover this scenario as described in #26906

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -2821,7 +2821,7 @@ public class TypeChecker extends BLangNodeVisitor {
         for (BLangLetVariable letVariable : letExpression.letVarDeclarations) {
             semanticAnalyzer.analyzeDef((BLangNode) letVariable.definitionNode, letExpression.env);
         }
-        BType exprType = checkExpr(letExpression.expr, letExpression.env, letExpression.expectedType);
+        BType exprType = checkExpr(letExpression.expr, letExpression.env, this.expType);
         types.checkType(letExpression, exprType, this.expType);
     }
 
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -2821,7 +2821,7 @@ public class TypeChecker extends BLangNodeVisitor {
         for (BLangLetVariable letVariable : letExpression.letVarDeclarations) {
             semanticAnalyzer.analyzeDef((BLangNode) letVariable.definitionNode, letExpression.env);
         }
-        BType exprType = checkExpr(letExpression.expr, letExpression.env);
+        BType exprType = checkExpr(letExpression.expr, letExpression.env, letExpression.expectedType);
         types.checkType(letExpression, exprType, this.expType);
     }
 
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
@@ -54,9 +54,9 @@ public class LetExpressionTest {
         BAssertUtil.validateError(negativeResult, i++, "redeclared symbol 'x'", 19, 21);
         BAssertUtil.validateError(negativeResult, i++, "undefined symbol 'y'", 23, 27);
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found 'string'", 27, 25);
-        BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found 'string'", 28, 13);
+        BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found 'string'", 28, 39);
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'string', found 'int'", 29, 28);
-        BAssertUtil.validateError(negativeResult, i, "incompatible types: expected 'string', found 'int'", 30, 16);
+        BAssertUtil.validateError(negativeResult, i, "incompatible types: expected 'string', found 'int'", 30, 42);
     }
 
     @Test(description = "Test cases for scenarios where let expression is not yet supported")
@@ -102,6 +102,7 @@ public class LetExpressionTest {
 //                {"testLetExpressionErrorBindingSimple"},
 //                {"testLetExpressionErrorBindingVar"},
 //                {"testLetExpressionRecordConstrainedErrorBinding"},
+                {"testAnonymousRecordWithLetExpression"}
         };
     }
 
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
@@ -58,7 +58,8 @@ public class LetExpressionTest {
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'string', found 'int'", 29, 28);
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'string', found 'int'", 30, 42);
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'byte', found 'int'", 41, 37);
-        BAssertUtil.validateError(negativeResult, i, "incompatible types: expected 'int', found 'string'", 48, 14);
+        BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found 'string'", 48, 14);
+        BAssertUtil.validateError(negativeResult, i, "too many arguments in call to 'new()'", 57, 37);
     }
 
     @Test(description = "Test cases for scenarios where let expression is not yet supported")
@@ -105,7 +106,8 @@ public class LetExpressionTest {
 //                {"testLetExpressionErrorBindingVar"},
 //                {"testLetExpressionRecordConstrainedErrorBinding"},
                 {"testAnonymousRecordWithLetExpression"},
-                {"testRecordWithLetExpression"}
+                {"testRecordWithLetExpression"},
+                {"testLetWithClass"}
         };
     }
 
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
@@ -56,7 +56,9 @@ public class LetExpressionTest {
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found 'string'", 27, 25);
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found 'string'", 28, 39);
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'string', found 'int'", 29, 28);
-        BAssertUtil.validateError(negativeResult, i, "incompatible types: expected 'string', found 'int'", 30, 42);
+        BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'string', found 'int'", 30, 42);
+        BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'byte', found 'int'", 41, 37);
+        BAssertUtil.validateError(negativeResult, i, "incompatible types: expected 'int', found 'string'", 48, 14);
     }
 
     @Test(description = "Test cases for scenarios where let expression is not yet supported")
@@ -102,7 +104,8 @@ public class LetExpressionTest {
 //                {"testLetExpressionErrorBindingSimple"},
 //                {"testLetExpressionErrorBindingVar"},
 //                {"testLetExpressionRecordConstrainedErrorBinding"},
-                {"testAnonymousRecordWithLetExpression"}
+                {"testAnonymousRecordWithLetExpression"},
+                {"testRecordWithLetExpression"}
         };
     }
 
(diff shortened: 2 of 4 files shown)


dulajdilshan:
The PR looks good. Shall we rename the commit messages according to the guidelines?

IMS94:
@dulajdilshan Updated the commit message accordingly. Can you check now?

anupama-pathirage:
@IMS94  The commit messages and PR heading is not according to the best practices. Please correct them.

A properly formed Git commit subject line should always be able to complete the following sentence:

If applied, this commit will **your subject line here**

Refer [1] for more details.
 
[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md#submit-your-contribution

IMS94:
@anupama-pathirage I updated the commit message and PR title. Can you check now?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
```diff
@@ -54,9 +54,9 @@ public void testLetExpressionNegative() {
         BAssertUtil.validateError(negativeResult, i++, "redeclared symbol 'x'", 19, 21);
         BAssertUtil.validateError(negativeResult, i++, "undefined symbol 'y'", 23, 27);
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found 'string'", 27, 25);
-        BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found 'string'", 28, 13);
+        BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found 'string'", 28, 39);
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'string', found 'int'", 29, 28);
-        BAssertUtil.validateError(negativeResult, i, "incompatible types: expected 'string', found 'int'", 30, 16);
+        BAssertUtil.validateError(negativeResult, i, "incompatible types: expected 'string', found 'int'", 30, 42);
```
dulajdilshan, line 59: Shall we add some negative tests for this?

IMS94: @dulajdilshan Can you give some examples? For the specific case of #26906, we had a compilation error. Therefore, adding the 
```
public function testAnonymousRecordWithLetExpression() {
    record {
        int i;
        int j?;
    } rec = let int v = 1 in {i: v};

    assertTrue(rec.i == 1, "rec.i == 1");
}
``` 
to the `let-expression-test.bal` file ensures that the compilation no longer fails (otherwise, test initialization fails). Apart from that, what specific cases can we add?

dulajdilshan:
The mentioned issue is fixed with https://github.com/ballerina-platform/ballerina-lang/pull/27596

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
```diff
@@ -316,6 +316,15 @@ function getRecordConstrainedError() returns FooError {
     return e;
 }
 
+public function testAnonymousRecordWithLetExpression() {
+    record {
+        int i;
+        int j?;
+    } rec = let int v = 1 in {i: v};
+    
+    assertTrue(rec.i == 1, "rec.i == 1");
```
MaryamZi, line 325: Shall we use `assert` instead?

```suggestion
    assert(1, rec.i);
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2821,7 +2821,7 @@ public void visit(BLangLetExpression letExpression) {
         for (BLangLetVariable letVariable : letExpression.letVarDeclarations) {
             semanticAnalyzer.analyzeDef((BLangNode) letVariable.definitionNode, letExpression.env);
         }
-        BType exprType = checkExpr(letExpression.expr, letExpression.env);
+        BType exprType = checkExpr(letExpression.expr, letExpression.env, letExpression.expectedType);
```
MaryamZi, line 2824: Is `letExpression.expectedType` here different from `this.expType`? If not, I think we should use `this.expType` instead because that may be what is set consistently. 

`expectedType` is set to the expression more for post type-checking analysis.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
```diff
@@ -316,6 +316,15 @@ function getRecordConstrainedError() returns FooError {
     return e;
 }
 
+public function testAnonymousRecordWithLetExpression() {
```
MaryamZi, line 319: As discussed in https://github.com/ballerina-platform/ballerina-lang/issues/26906, since this test may no longer cover the issue we are fixing here, shall we add another test that does?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
IMS94: Since the assignability is addressed by #27596, negative tests seems to be not required.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
IMS94: Added 
```
public function testAnonymousRecordWithLetExpression() {
    record {
        int i;
        int j?;
    } rec = let int v = 1 in {i: v};
    
    assert(rec.i, 1);
    
    rec.j = 5;
    assert(rec?.j, 5);
}
```
to cover the scenario you had described in #26906 which relies on context information rather than assignability.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
IMS94: Done. Thanks!

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
IMS94: Yes. Both are same here. Changed to use `this.expType` instead.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
MaryamZi: We can add a negative test like

```ballerina
    record {
        byte i;
        int j?;
    } rec = let int v = 160 in {i: v};
```

Without this fix, the error was something like
```
incompatible types: expected 'record {| byte i; int j?; anydata...; |}', found 'record {| int i; |}'
```

With this fix I believe it'll be something like
```
incompatible types: expected 'byte', found 'int'
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
```diff
@@ -316,6 +316,18 @@ function getRecordConstrainedError() returns FooError {
...
```
MaryamZi, line 319: Shall we add another test like

where `Rec` is
```ballerina
type Rec record {|
    int i;
    int j = 100;
|};
```

```ballerina
Rec rec1 = let int v = 160 in {i: v};
assert(160, rec1.i);
assert(100, rec1.j);

Rec rec2 = let int v = 161 in {i: v, j: v};
assert(161, rec2.i);
assert(161, rec2.j);
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
```diff
@@ -316,6 +316,18 @@ function getRecordConstrainedError() returns FooError {
...
+    record {
+        int i;
+        int j?;
+    } rec = let int v = 1 in {i: v};
+    
+    assert(rec.i, 1);
```
MaryamZi, line 325: The expected value seems to be the first param.
```suggestion
    assert(1, rec.i);
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
```diff
@@ -316,6 +316,18 @@ function getRecordConstrainedError() returns FooError {
...
+    
+    rec.j = 5;
+    assert(rec?.j, 5);
```
MaryamZi, line 328:
```suggestion
    assert(5, rec?.j);
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
IMS94: Added!

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
IMS94: Done!

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-negative.bal
```diff
@@ -33,3 +33,17 @@ function wrongTypes() {
...
+        int j?;
+    } rec1 = let int v = 160 in {i: v};
+
+    record {
+        string i;
+        int j?;
+    } rec2 = let string v = "160" in {i: v};
+    
+    rec2.j = "invalid";
+}
```
MaryamZi, line 49: Missing new line.
```suggestion
}

```

MaryamZi (commented):
Shall we also add a test for a non-record type/expression? Maybe with objects?

```ballerina
class Foo {
    function init(int x) {

    }
}

Foo f1 = let int m = 100 in new (m); // valid
Foo f2 = let int m = 100 in new (m, m); // invalid, mismatched args
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/let/LetExpressionTest.java
IMS94: Yes. That's the error we get. Added.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
```diff
@@ -316,6 +316,51 @@ function getRecordConstrainedError() returns FooError {
...
+class FooClass {
+    int m;
+    
+    public function init(int m) {
+        self.m = m;
+    }
+}
+
+function testLetWithClass() {
+    FooClass foo = let int m = 5 in new(m);
```
MaryamZi, line 355:
```suggestion
    FooClass foo = let int m = 5 in new (m);
```

We usually leave a space between `new` and `(`. 

Since this is effectively `new FooClass(m)` without `FooClass`.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
```diff
@@ -316,6 +316,51 @@ function getRecordConstrainedError() returns FooError {
...
+    assert(5, foo.m);
+    
+    FooClass foo2 = new(let var arr = [1, 2, 3] in arr.reduce(function(int sum, int x) returns int {
```
MaryamZi, line 358:
```suggestion
    FooClass foo2 = new (let var arr = [1, 2, 3] in arr.reduce(function(int sum, int x) returns int {
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-negative.bal
```diff
@@ -33,3 +33,26 @@ function wrongTypes() {
...
+
+class FooClass {
+    public function init(int m) {
+    }
+}
+
+function testLetWithClass() {
+    FooClass foo = let int m = 5 in new(m, m);
```
MaryamZi, line 57:
```suggestion
    FooClass foo = let int m = 5 in new (m, m);
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-negative.bal
```diff
@@ -33,3 +33,26 @@ function wrongTypes() {
...
+        byte i;
+        int j?;
+    } rec1 = let int v = 160 in {i: v};
+
+    record {
+        string i;
+        int j?;
+    } rec2 = let string v = "160" in {i: v};
+    
+    rec2.j = "invalid";
```
MaryamZi, line 48: Since `rec2`'s type will be the anonymous record type irrespective of the RHS expression, IMO as a let expression test it may be better to test an invalid expression for `j` in the mapping constructor expression itself.

```suggestion
    } rec2 = let string v = "160" in {i: v, j: "invalid"};
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/let/let-expression-test.bal
IMS94: Got it. Thanks!

MaryamZi (approved):

Issue #28040, opened 2021-01-21
Can assign structure types with error to anydata


KRVPerera:
**Description:**

```Ballerina
public function main() {
    (anydata|error)[] x = [error("")];
    anydata y = x;
}
```

This is a side effect of `Types:containsErrorType` not going recursively inside the array element type. But being used in many other places there could be breaks due to this.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**
https://github.com/ballerina-platform/ballerina-lang/issues/22429 Remove error from anydata

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #28062, opened 2021-01-21
Remove error type from pure data internal compiler type

rdhananjaya:
## Purpose
With this PR error is not part of `anydata` datatype.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/28040

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/testerina/modules/testerina-core/src/main/ballerina/assert.bal
```diff
@@ -69,9 +71,8 @@ public isolated function assertFalse(boolean condition, string msg = "Assertion
 # + actual - Actual value
 # + expected - Expected value
 # + msg - Assertion error message
-public isolated function assertEquals(anydata|error actual, anydata|error expected, string msg = "Assertion Failed!") {
-    boolean isEqual = (actual == expected);
-    if (!isEqual) {
+public isolated function assertEquals(Comparable actual, Comparable expected, string msg = "Assertion Failed!") {
```
MaryamZi, line 74: IIRC, `assertEquals` was meant to do the same as `==` and `assertExactEquals` the same as `===`. I think it'll be consistent to make this
```suggestion
public isolated function assertEquals(anydata actual, anydata expected, string msg = "Assertion Failed!") {
```
instead.

misc/testerina/modules/testerina-core/src/main/ballerina/mock.bal
```diff
@@ -107,7 +109,7 @@ public class MockObject {
 public class MemberFunctionStub {
     object {} mockObject;
     string functionName = "";
-    anydata|error args = [];
+    AnydataOrError[] args = [];
```
MaryamZi, line 112: Why do we need this change?

tests/jballerina-integration-test/src/test/resources/configurables/configurableProject/main/main.bal
```diff
@@ -38,10 +38,10 @@ public function main() {
...
     test:assertEquals(result, decimalVar);
 
-    test:assertEquals([1,2,3], intArr);
-    test:assertEquals([9.0, 5.6], floatArr);
-    test:assertEquals(["red", "yellow", "green" ], stringArr);
-    test:assertEquals([true, false, false, true], booleanArr);
+    test:assertEquals(<int[]>[1,2,3], intArr);
+    test:assertEquals(<float[]>[9.0, 5.6], floatArr);
+    test:assertEquals(<string[]>["red", "yellow", "green" ], stringArr);
+    test:assertEquals(<boolean[]>[true, false, false, true], booleanArr);
```
MaryamZi, line 44: Shouldn't these work without the casts?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/access/field_access.bal
```diff
@@ -279,3 +279,11 @@ function testFieldAccessOnMapConstruct() returns boolean {
...
     return "Sanjiva" == name;
 }
+
+isolated function isEqual(anydata|error actual, anydata|error expected) returns boolean {
+    if (actual is anydata && expected is anydata) {
+        return (actual == expected);
+    } else {
+        return (actual === expected);
+    }
+}
```
MaryamZi, line 289:
```suggestion
}

```

tests/jballerina-unit-test/src/test/resources/test-src/statements/ifelse/type-guard.bal
```diff
@@ -923,7 +923,7 @@ function testTypeGuardForCustomErrorPositive() returns [boolean, boolean] {
 
         Details m1 = e5.detail();
         Details m2 = e6.detail();
-        isSpecificError = e5.message() == ERR_REASON && e6.message() == ERR_REASON_TWO && m1 == d && m2 == d;
+        isSpecificError = e5.message() == ERR_REASON && e6.message() == ERR_REASON_TWO;
```
MaryamZi, line 926: Should we check the "message" field in the detail instead of what we are removing?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/clone-operation.bal
```diff
@@ -384,27 +385,24 @@ public function testCloneArrayWithError() returns boolean {
...
+                                <string> check ma["two"] == <string> check clonedMap["two"];
 
-    map<error> clonedErrorMap = <map<error>> clonedMap["errMap"];
+    map<error> clonedErrorMap = <map<error>> check clonedMap["errMap"];
     foreach [string, error] [x, y] in errMapFromValue.entries() {
         cloneSuccessful = cloneSuccessful && y === clonedErrorMap[x];
     }
     return cloneSuccessful;
-}
+}
```
MaryamZi, line 408:
```suggestion
}

```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/clone-operation.bal
```diff
@@ -384,27 +385,24 @@ public function testCloneArrayWithError() returns boolean {
...
     };
 
-    map<anydata> ma = {
+    map<value:Cloneable> ma = {
         one: 1,
         two: "two",
         errMap: errMap
     };
-    map<error> errMapFromValue = <map<error>> ma["errMap"];
+    map<error> errMapFromValue = <map<error>> check ma["errMap"];
```
MaryamZi, line 396: This should be `checkpanic` right? There seems to be a bug here with `value:Cloneable`.

MaryamZi: Created https://github.com/ballerina-platform/ballerina-lang/issues/28081.

misc/testerina/modules/testerina-core/src/main/ballerina/assert.bal
rdhananjaya: @MaryamZi I changed this like this to have the same behavior testerina users used to have with this API.

@hevayo please confirm if the suggested is the expected behavior.

hevayo: The reason we didn't include object are because we didn't do a value compare. And the assert equals support printing diff as well. For object equals we use assertExactEquals. is that preserved with this change.

Dilhasha: @rdhananjaya 

What's the possibility of the below scenario with this change? If the function intAdd returns an error?

`test:assertEquals(4,intAdd(1,3))`

rdhananjaya: Yes, this does not change the diff printing mechanism, only the value comparison.
```ballerina
import ballerina/test;

public function main() {
    error a = error("hello", name = "oh no!!");
    error b = error("hello", name = "oh no!");
    test:assertEquals(a, b);
}
```
Out:
```
error: Assertion Failed!
 
expected: 'error("hello",name="oh no!")'
actual  : 'error("hello",name="oh no!!")'
        at ballerina.test.0_0_0:createBallerinaError(assert.bal:45)
           ballerina.test.0_0_0:assertEquals(assert.bal:77)
           dhananjaya.newProj.0_1_0:main(main.bal:7)
```

If we are to constrain parameters to `anydata`, a programmer who wants to compare two errors will have to write the comparison logic them self, or maybe testerina could supply a separate function to compare errors for similarity.

rdhananjaya: @Dilhasha 
```ballerina
import ballerina/test;

public function main() {
    test:assertEquals(intAdd(1,3), 4);
}

function intAdd(int i, int j) returns int|error {
    return error("opsi");
}
```
Outupt:
```
error: Assertion Failed!
 
expected: <int> '4'
actual  : <error> 'error("opsi")'
        at ballerina.test.0_0_0:createBallerinaError(assert.bal:45)
           ballerina.test.0_0_0:assertEquals(assert.bal:77)
           main:main(main.bal:4)
```

MaryamZi: @rdhananjaya, agreed about the API change. 

My concern though is that by doing this we are doing a special comparison just for errors, a value equality check that is not defined in the language spec. 

We would then have to explicitly define how we are doing this comparison for `ballerina/test`. 

Additionally, I think users may raise queries as to why this special comparison is only done for errors, and not other behavioural values like objects. One can argue that there can be a similar comparison for objects by comparing field-wise?

IMO, even though it may break code, it would be cleaner to base
- `assertEquals` and `assertNotEquals` on `==` defined in the language
- `assertExactEquals` and `assertNotExactEquals`  on `===` defined in the language

This way we'll have a clear separation in the methods and a rationale based on the language itself.

Yes, the user will have to write their own method to compare errors and structures with errors then, but I think it'll still be consistent with other behavioural values.

rdhananjaya: For me at least, testerina is not a part of the language, it's a test framework just like any other, which helps me execute tests and compare values. 

About objects vs error:
Comparing fields of the object is not sufficient, as there is a behavior associated with each object. I think we don't have a way to compare those.
But for errors, even though they are categorized as behavioral values, I don't see any behavior that we can't compare. And I also think we don't need to compare the stack trace as it renders this comparison pretty useless. 

Note: If we are to go with the current change we need to compare distinct type ids as well.

rdhananjaya: I asked why the error is a behavioral value in the spec: https://github.com/ballerina-platform/ballerina-spec/issues/699

docs/bir-spec/src/test/java/org/ballerinalang/birspec/BIRTestUtils.java
```diff
@@ -88,11 +88,12 @@
 
     static void validateBIRSpec(String testSource) {
         BCompileUtil.BIRCompileResult compileResult = BCompileUtil.generateBIR(testSource);
+        Assert.assertNotNull(compileResult, "Compilation may contain errors!");
+
         BIRNode.BIRPackage expectedBIRModule = compileResult.getExpectedBIR();
         byte[] actualBIRbinary = compileResult.getActualBIR();
         Bir kaitaiBir = new Bir(new ByteBufferKaitaiStream(actualBIRbinary));
 
-        Assert.assertNotNull(kaitaiBir, "Compilation may contain errors!");
```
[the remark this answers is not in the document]
rdhananjaya: I don't think so. If anything goes wrong, `generateBIR` function should return `null`.

langlib/langlib-test/src/test/resources/test-src/type-param/type_param_bound_type_as_cet.bal
```diff
@@ -193,3 +193,11 @@ function assertValueEquality(anydata|error expected, anydata|error actual) {
     panic error(ASSERTION_ERROR_REASON,
                 message = "expected '" + expectedValAsString + "', found '" + actualValAsString + "'");
 }
+
+isolated function isEqual(anydata|error actual, anydata|error expected) returns boolean {
+    if (actual is anydata && expected is anydata) {
+        return (actual == expected);
+    } else {
+        return (actual === expected);
+    }
```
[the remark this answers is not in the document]
rdhananjaya: Why do we need this change?

langlib/langlib-test/src/test/resources/test-src/valuelib_fromBalString_test.bal
```diff
@@ -188,7 +186,7 @@ function testArrayFromBalString() {
     string[] arr5 = ["ABC", "XYZ", "LMN"];
     decimal[] arr6 = [12.65, 1, 2, 90.0];
     (anydata|error)[] arr7 = ["str", 23, 23.4, true, {"x":"AA","y":(1.0/0.0),"z":1.23}, x1, ["X", (0.0/0.0),
-    x1], underGradTable, err, xmlVal];
```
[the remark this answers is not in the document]
rdhananjaya: will do

langlib/langlib-test/src/test/resources/test-src/booleanlib_test.bal
```diff
@@ -23,24 +23,30 @@ function testFromString(string s, boolean|error expected) {
...
+
+    string expectedValAsString = expected is error ? expected.toString() : expected.toString();
+    string actualValAsString = actual is error ? actual.toString() : actual.toString();
+    string reason = "expected [" + expectedValAsString + "] of type [" + expT.toString()
+                        + "], but found [" + actualValAsString + "] of type [" + actT.toString() + "]";
+    error e = error(reason);
+    panic e;
+}
+
+isolated function isEqual(anydata|error actual, anydata|error expected) returns boolean {
```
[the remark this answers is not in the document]
rdhananjaya: I think this is same as this https://github.com/ballerina-platform/ballerina-lang/pull/28062#discussion_r562162805

tests/jballerina-integration-test/src/test/resources/configurables/configurableProject/main/main.bal
rdhananjaya: Without this we are getting `ambiguous type 'Comparable'`, `Comparable` is `anydata|error|Comparable[]|map<Comparable>`

rdhananjaya: This also relates to https://github.com/ballerina-platform/ballerina-lang/pull/28062#discussion_r562162805

misc/testerina/modules/testerina-core/src/main/ballerina/assert.bal
rdhananjaya: Created a discussion to conclude this https://github.com/ballerina-platform/ballerina-lang/discussions/28314
Please provide your opinions.

misc/testerina/modules/testerina-core/src/main/ballerina/mock.bal
rdhananjaya: Removed

langlib/langlib-test/src/test/resources/test-src/type-param/type_param_bound_type_as_cet.bal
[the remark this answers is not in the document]
MaryamZi: The else block is unnecessary, right?

Dilhasha (approved):
LGTM

langlib/langlib-test/src/test/resources/test-src/type-param/type_param_bound_type_as_cet.bal
```diff
@@ -184,7 +184,7 @@ function assertTrue(any|error actual) {
 }
 
 function assertValueEquality(anydata|error expected, anydata|error actual) {
-    if expected == actual {
+    if isEqual(expected, actual) {
```
MaryamZi, line 187: Shouldn't the arguments be switched? `isEqual` seems to have `actual` as the first param?

MaryamZi: Applicable for some other functions too.

rdhananjaya: It doesn't matter here right, it says if first argument equals to second argument do this.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/IsPureTypeUniqueVisitor.java
```diff
@@ -83,7 +83,7 @@ private boolean isAnydata(BType type) {
     }
 
     private boolean isPureType(BType type) {
-        return type.tag == TypeTags.ERROR || type.tag == TypeTags.READONLY || isAnydata(type);
+        return type.tag == TypeTags.ANYDATA || type.tag == TypeTags.READONLY || isAnydata(type);
```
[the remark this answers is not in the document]
rdhananjaya: https://github.com/ballerina-platform/ballerina-lang/issues/28640

langlib/langlib-test/src/test/resources/test-src/type-param/type_param_bound_type_as_cet.bal
MaryamZi: True that the order doesn't matter in terms of the logic, but isn't it misleading when you consider the parameter names? IMO we should either fix the order here to be in-line with `isEqual` or update the signature of `isEqual` to have names that imply a generic value comparison, maybe `v1` and `v2`.

rdhananjaya: `isEqual(expected, actual)` misleading in to what?

MaryamZi: The fact that we're passing `expected` as the first argument and `actual` as the second argument to a function that has the inverse as the signature.

```ballerina
isolated function isEqual(anydata|error actual, anydata|error expected) returns boolean {
    return actual is anydata && expected is anydata && actual == expected || actual === expected;
}
```

rdhananjaya: Misleading is when it says one thing and does some other thing, this says that it checks if two args are equal and it does exactly that. 
In my view what variable it passes to what parameter is just a preference (talking in this context)

MaryamZi: What I found potentially misleading was that we first pass `expected` and then `actual` to a function that seems to accept `actual` as the first parameter and `expected` as the second - not the implementation itself.

`isEqual` currently says it takes an `actual` value and an `expected` value and compares the two to see if they are equal, right? While highly unlikely in this scenario, what if we eventually add like an actual vs expected log there, given that it is an arbitrary comparison function?

Why can't we change the parameter names of the function to reflect that it's a generic comparison of two values, if the order cannot be changed?

rdhananjaya: We can absolutely change the param names.

If we decide to log something from a isolated function called `isEqual` that returns and boolean I think that function signature is misleading.

MaryamZi: +1, seems to be the case in several other files too.

tests/jballerina-unit-test/src/test/resources/test-src/types/anydata/anydata_test.bal
```diff
@@ -885,70 +871,7 @@ type MyErrorDetail record {|
...
-
-    anydata a1 = m1;
-    anydata a2 = <anydata> m2;
-
-    MyRecord m3 = <MyRecord> a1;
-    MyRecordTwo m4 = <MyRecordTwo> a2;
-
-    return m3["err"] === e1 && m4["err"] === e3 && m3 === a1 && m2 === m4;
-}
+MyError e3 = error MyError("err reason 3", e1 = e1, e2 = e2);
```
MaryamZi, line 874: Is this used now? Do we need this and e1 and e2?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/clone-operation.bal
```diff
@@ -384,25 +385,22 @@ public function testCloneArrayWithError() returns boolean {
 
 public function testCloneMapWithError() returns boolean {
     map<error> errMap = {
-        e1: err1,
-        e2: err2,
-        e3: err3,
-        e4: err4
```
MaryamZi, line 390: Any particular reason to change this?

misc/testerina/modules/testerina-core/src/main/ballerina/assert.bal
```diff
@@ -69,9 +69,8 @@ public isolated function assertFalse(boolean condition, string msg = "Assertion
 # + actual - Actual value
 # + expected - Expected value
 # + msg - Assertion error message
-public isolated function assertEquals(anydata|error actual, anydata|error expected, string msg = "Assertion Failed!") {
-    boolean isEqual = (actual == expected);
-    if (!isEqual) {
+public isolated function assertEquals(anydata|error actual, anydata expected, string msg = "Assertion Failed!") {
+    if (actual is error || actual != expected) {
```
MaryamZi, line 73: If we don't have one already, shall we add a test for when this function is called with an `error` value as the arg for `actual`?

rdhananjaya: Issue created https://github.com/ballerina-platform/ballerina-lang/issues/28818

tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/clone-operation.bal
rdhananjaya: I really don't remember why I did this, removed the change

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/variabledef/RecordVariableDefinitionTest.java
```diff
@@ -274,8 +274,8 @@ public void testIgnoreVariable() {
     public void testRecordVariableWithOnlyRestParam() {
         BValue[] returns = BRunUtil.invoke(result, "testRecordVariableWithOnlyRestParam");
         Assert.assertEquals(returns.length, 1);
-        Assert.assertEquals(returns[0].stringValue(), "{\"name\":\"John\", \"age\":{\"age\":30, \"format\":\"YY\", " +
-                "\"year\":1990}, \"married\":true, \"work\":\"SE\"}");
+        Assert.assertEquals(returns[0].stringValue(),
+                "{\"name\":\"John\", \"age\":{age:30, format:\"YY\", year:1990}, \"married\":true, \"work\":\"SE\"}");
```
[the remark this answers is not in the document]
rdhananjaya: It's due to this line https://github.com/ballerina-platform/ballerina-lang/pull/28062/commits/051379d6c54c4db2298b01e025ae36a32a350feb#diff-a5961a28254babde0aa44e4c1cb05126b5ee68494baccca1c76879dec267b8aeL1858

We can't use cloneWithType with rest binding/match pattern processing as it only works with `anydata`.
I read the spec again about the rest binding/match pattern and it seems we don't need to create a clone for the rest pattern.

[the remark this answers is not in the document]
MaryamZi: Ah okay, so I'm assuming previously `age` was a map at https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-core/src/main/java/org/ballerinalang/core/model/values/BMap.java#L283 but is now the original record?

Agreed that we shouldn't have used cloneWithType.

MaryamZi (approved):

PR #28133, opened 2021-01-22
[Testerina] Remove use of log leecher from standard test cases

Dilhasha:
## Purpose
Remove use of log leecher from standard test cases

Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/DisableTestsTestCase.java
+++ b/tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/DisableTestsTestCase.java
@@ -19,6 +19,7 @@ package org.ballerinalang.testerina.test;
 
 import org.ballerinalang.test.context.BMainInstance;
 import org.ballerinalang.test.context.BallerinaTestException;
+import org.ballerinalang.testerina.test.utils.AssertionUtils;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
@@ -42,8 +43,6 @@ public class DisableTestsTestCase extends BaseTestCase {
     public void testDisablingTests() throws BallerinaTestException {
         String errorOutput = balClient.runMainAndReadStdOut("test", new String[]{"disable-test.bal"},
                 new HashMap<>(), projectPath, true);
-        if (errorOutput.contains("error: there are test failures")) {
-            throw new BallerinaTestException("Test failed due to disable test failure in test framework");
-        }
+        AssertionUtils.assertForTestFailures(errorOutput, "disable test failure");
     }
 }
--- a/tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/DisableTestsTestCase.java
+++ b/tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/DisableTestsTestCase.java
@@ -19,9 +19,10 @@ package org.ballerinalang.testerina.test;
 
 import org.ballerinalang.test.context.BMainInstance;
 import org.ballerinalang.test.context.BallerinaTestException;
-import org.ballerinalang.test.context.LogLeecher;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
+
+import java.util.HashMap;
 
 /**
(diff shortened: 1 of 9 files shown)


tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/DisableTestsTestCase.java
```diff
@@ -39,10 +40,10 @@ public void setup() throws BallerinaTestException {
...
-        String msg = "2 passing";
-        LogLeecher clientLeecher = new LogLeecher(msg);
-        balClient.runMain("test", new String[]{"disable-test.bal"}, null, new String[]{},
-                new LogLeecher[]{clientLeecher}, projectPath);
-        clientLeecher.waitForText(20000);
+        String errorOutput = balClient.runMainAndReadStdOut("test", new String[]{"disable-test.bal"},
+                new HashMap<>(), projectPath, true);
+        if (errorOutput.contains("error: there are test failures")) {
+            throw new BallerinaTestException("Test failed due to disable test failure in test framework");
+        }
```
NipunaRanasinghe, line 47: Since I've noticed this same logic is being duplicated in many places, how about defining an assertion util as follows? WDYT? 

```java
private void assertForTestFailures(String programOutput, String errorMessage) throws BallerinaTestException {
    if (errorOutput.contains("error: there are test failures")) {
         throw new BallerinaTestException(errMessage);
    }
}

Comment (Dilhasha) [reply tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/DisableTestsTestCase.java]: Fixed with 8aae1b9

File: tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/utils/AssertionUtils.java [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.ballerinalang.testerina.test.utils;
+
+import org.ballerinalang.test.context.BallerinaTestException;
+
+/**
+ * Util class for test assertions.
+ */
```
azinneera, line 24: lets add `@since 2.0.0`

azinneera (approved):

tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/utils/AssertionUtils.java
```diff
@@ -0,0 +1,32 @@
...
+public class AssertionUtils {
+
+    public static void assertForTestFailures(String programOutput, String errMessage) throws BallerinaTestException {
+        if (programOutput.contains("error: there are test failures")) {
```
Ibaqu, line 28: Say hypothetically, a test failed due to a bad-sad or npe exception. 

Would the test automatically fail? Or will the `balClient` still return the buffered output?

Dilhasha: If a ballerina test fails it will anyway have the mentioned text in the program output. But, if it fails due to a compilation error, this fails to capture that. So, I'll add an else statement to capture unforeseen errors.

Ibaqu: I think the else statement seems to be causing the normal execution of tests to fail. Since if it can't find `error: there are test failures` it will automatically throw an exception.

Dilhasha: yeah will revert the else, and add a 'else if' to capture compilation issues in bal file. 
Every other test failure will be handled with the existing output.

azinneera (approved):

PR #28160, opened 2021-01-25
Fix spec deviations lang libs

### LINKED ISSUE #23360: Use distinct object type for Iterable
Refer Commit https://github.com/ballerina-platform/ballerina-spec/commit/54725c33fa7afdb203c4175e06070d900c042c2b

### LINKED ISSUE #25446: Set return types of `fromJsonFloatString` and `fromJsonDecimalString` as `JsonFloat` and `JsonDecimal` respectively
**Description:**
```ballerina
public type JsonDecimal ()|boolean|string|decimal|JsonDecimal[]|map<JsonDecimal>;
public function fromJsonDecimalString(string str) returns JsonDecimal|error = external;

public type JsonFloat ()|boolean|string|float|JsonFloat[]|map<JsonFloat>;
public function fromJsonFloatString(string str) returns JsonFloat|error = external;
```
To fix this issue need to allow cyclic reference in unions typedefs.

### LINKED ISSUE #27899: Align lang libs in the compiler with the ones in the spec.
**Description:**
We need to go through and update langlibs with the spec.

Things to consider:
- Addition or removal of functions.
- Signature or data type changes.

**Steps to reproduce:**

**Affected Versions:**

KRVPerera:
Fixes #27899
Fixes #25446
Fixes #23360

RawTemplate type is now a distinct type
`indexOf` and `lastIndexOf` lang.array now work only on pure types (anydata|error).
Object iterators must be a subtype of `object:Iterable`

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [x] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.object/src/main/ballerina/object.bal
```diff
@@ -13,3 +13,69 @@
...
+};
+
+// moved here from __internal
+
+# Integer range expression is represented using `IntRange` object.
+#
+# + iStart - start expression of range expression
+# + iEnd - second expression on range expression
+# + iCurrent - current cursor
+public class __IntRange {
```
MaryamZi, line 36: These aren't in the spec, right? Probably why we had it in the __internal module in the first place.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -3404,6 +3404,14 @@ public void visit(BLangMatchStructuredBindingPatternClause patternClause) {
     public void visit(BLangForeach foreach) {
         // Check the collection's type.
         typeChecker.checkExpr(foreach.collection, env);
+        // object type collection should be a subtype of 'object:Iterable
+        if (foreach.collection.type.tag == TypeTags.OBJECT
+                && !types.isAssignable(foreach.collection.type, symTable.iterableType)) {
```
MaryamZi, line 3409: Wouldn't this allow in foreach statements iterable objects where the iterator's next may return error too? https://github.com/ballerina-platform/ballerina-lang/issues/28264

langlib/lang.object/src/main/ballerina/object.bal
KRVPerera: Yes, then internal will depend on `lang.object` vice versa.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -766,7 +767,9 @@ public BSymbol lookupConstructorSpaceSymbolInPackage(Location pos,
     }
 
     public BSymbol lookupMethodInModule(BPackageSymbol moduleSymbol, Name name, SymbolEnv env) {
-
+        if (moduleSymbol == null) {
```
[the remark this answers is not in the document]
KRVPerera: This is called by the programmer. It can happen if the programmer called for a module which is not loaded yet to memory.
Check the usage pattern around here.
https://github.com/ballerina-platform/ballerina-lang/blob/0418e82129063c92a2209aaf81d43ffde13bf62d/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java#L646

Do you think it is better to let it crash for NPE for an invalid call? Then yes this is not needed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
KRVPerera: Is the only place we use iterable is in `foreach`? I changed the error message here please check.
Check https://github.com/ballerina-platform/ballerina-lang/pull/28160/commits/032b8e63532af4fc28843be212b8696d37f0c340

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1664,3 +1664,9 @@ error.invalid.readonly.field.type=\
 error.continue.not.allowed=continue not allowed here
 
 error.break.not.allowed=break not allowed here
+
+error.invalid.iterable.type=\
+  invalid iterable type ''{1}'': an iterable object must be subtypes of ''{0}''
+
+error.invalid.iterable.completion.type.in.next.function=\
+  invalid iterable completion type ''{0}'': next method completion type cannot contain type ''{1}''
```
MaryamZi, line 1672: This is specifically for `foreach` right? Shall we be more specific in the error? Having error is not an issue for iterable, it is only an issue when used in a foreach statement.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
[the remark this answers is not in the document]
MaryamZi: Given that this wouldn't be null for user code, IMO it will be better if the developer can identify early on that the expected module is not available when we get here. WDYT?

langlib/lang.object/src/main/ballerina/object.bal
MaryamZi: The problem with this is that the users will start seeing this in suggestions, API docs, etc. Is the dependency required for `Iterable`?

KRVPerera: We use `createIntRange` in desugar. `__IntRange` should be a subtype of `Iterable`. So Iterable should be available when we create `__IntRange`. Does creating interanal.bal inside lang.object module work?

MaryamZi: Is `createIntRange` used in langlibs too? As long as is lang.object is also available at the time this check is done we can work around this, right?

Issue #28245, opened 2021-01-27
Add test to ensure only single strand's stack frames are visible in debug view.


praveennadarajah:
**Description:**
We need to add tests to ensure only single strand's stack frames are visible in call stack section in debug view.

Related to https://github.com/ballerina-platform/ballerina-lang/issues/27913 and https://github.com/ballerina-platform/ballerina-lang/pull/28238

PR #28250, opened 2021-01-27
Add call stack representation test for single strand visibility

praveennadarajah:
## Purpose
This PR will add call stack representation test for single strand visibility.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/28245

### DIFF
--- a/tests/jballerina-debugger-integration-test/src/main/java/org/ballerinalang/debugger/test/utils/DebugTestRunner.java
+++ b/tests/jballerina-debugger-integration-test/src/main/java/org/ballerinalang/debugger/test/utils/DebugTestRunner.java
@@ -40,6 +40,7 @@ import org.eclipse.lsp4j.debug.StackTraceResponse;
 import org.eclipse.lsp4j.debug.StepInArguments;
 import org.eclipse.lsp4j.debug.StepOutArguments;
 import org.eclipse.lsp4j.debug.StoppedEventArguments;
+import org.eclipse.lsp4j.debug.ThreadsResponse;
 import org.eclipse.lsp4j.debug.Variable;
 import org.eclipse.lsp4j.debug.VariablesArguments;
 import org.eclipse.lsp4j.debug.VariablesResponse;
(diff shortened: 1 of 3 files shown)


tests/jballerina-debugger-integration-test/src/main/java/org/ballerinalang/debugger/test/utils/DebugTestRunner.java
```diff
@@ -415,6 +416,29 @@ public void resumeProgram(StoppedEventArguments context, DebugResumeKind kind)
         return stackFrames;
     }
 
+    /**
+     * Can be used to fetch threads when a debug hit is occurred.
+     *
+     * @return Thread array with threads information.
+     * @throws BallerinaTestException if an error occurs when fetching threads.
+     */
+    public org.eclipse.lsp4j.debug.Thread[] fetchThreads() throws BallerinaTestException {
```
NipunaRanasinghe, line 425: Shall we avoid using full qualified names?

tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/CallStackDebugTest.java
```diff
@@ -55,6 +57,10 @@ public void setup() throws BallerinaTestException {
     public void callStackTest() throws BallerinaTestException {
         Pair<BallerinaTestDebugPoint, StoppedEventArguments> debugHitInfo = debugTestRunner.waitForDebugHit(25000);
         StackFrame[] frames = debugTestRunner.fetchStackFrames(debugHitInfo.getRight());
+        Thread[] threads = debugTestRunner.fetchThreads();
+
+        // Call stack representation test for single strand visibility.
+        Assert.assertEquals(threads.length, 1);
```
NipunaRanasinghe, line 63: IMO it would be better if we can separate out strand related tests from stack trace tests, by moving these strand related testing into a separate test case. In there, we can assert strand representation(count) under different contexts, like
- inside main function
- inside a function which runs asynchronously (invoked using `start` action)
- inside a worker
- inside the init function of a class

tests/jballerina-debugger-integration-test/src/main/java/org/ballerinalang/debugger/test/utils/DebugTestRunner.java
praveennadarajah: I'm using the full qualified name to avoid conflicts between `java.lang.Thread[]` and `org.eclipse.lsp4j.debug.Thread[]`.  Here Thread[] already referring to `java.lang.Thread[]`. Therefore I'm using full qualified name `org.eclipse.lsp4j.debug.Thread[]` to refer debug Thread[].

tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/CallStackDebugTest.java
praveennadarajah: IMO, strand and stack frame related tests should be together. Tests should be written in such a way that given different scenarios how many strands are running and what are the stack frames belongs to each strand.

With the current thread filtration mechanism, there will be only one main strand throughout the execution flow(main, init and other functions). For workers and `start` action, strands will be created and attached to the currently running threads. We can further improve these test once we support multiple strands debug in future.

praveennadarajah: +1 for moving strand related tests into a separate test case.

tests/jballerina-debugger-integration-test/src/main/java/org/ballerinalang/debugger/test/utils/DebugTestRunner.java
NipunaRanasinghe: Yeah I also noticed that this is unavoidable. +1

NipunaRanasinghe (approved):

PR #28265, opened 2021-01-28
Generate jacoco coverage for integration tests

azinneera:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/BasicCasesTest.java
```diff
@@ -41,98 +41,112 @@ public void setup() throws BallerinaTestException {
 
     @Test
     public void testAssertions() throws BallerinaTestException {
-        String errorOutput = balClient.runMainAndReadStdOut("test", new String[]{"assertions"},
+        String errorOutput = balClient.runMainAndReadStdOut("test",
```
manuranga, line 44: can we use `@BeforeMethod` annotation to avoid putting this in each method ?

azinneera: Couldn't use a BeforeMethod since the rest of the args to the command are only available within the test and hence we cannot merge the args. Wrote a protected method and used it to prefix the coverage args.

manuranga (approved):

PR #28286, opened 2021-01-29
Fix not exposing symbols within on fail clause through semantic API 

### LINKED ISSUE #28189: On Fail variable definition not suggested within the clause
**Description:**
onFailVar

On fail variable definition is visible within its clause. Atm, the reference is not among the suggestions.

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**
Swan Lake Alpha

**Steps to reproduce:**

**Related Issues:**

pcnfernando:
Fixes #28189

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
@@ -411,6 +411,7 @@ public class EnvironmentResolver extends BLangNodeVisitor {
             this.scope = this.symbolEnv;
             this.acceptNode(whileNode.body, this.symbolEnv);
         }
+        this.acceptNode(whileNode.onFailClause, symbolEnv);
     }
 
     @Override
@@ -421,6 +422,8 @@ public class EnvironmentResolver extends BLangNodeVisitor {
         if (transactionNode.transactionBody != null) {
             this.acceptNode(transactionNode.transactionBody, transactionEnv);
         }
+
+        this.acceptNode(transactionNode.onFailClause, symbolEnv);
     }
 
     @Override
@@ -461,11 +464,13 @@ public class EnvironmentResolver extends BLangNodeVisitor {
     @Override
     public void visit(BLangLock lockNode) {
         this.acceptNode(lockNode.body, symbolEnv);
+        this.acceptNode(lockNode.onFailClause, symbolEnv);
     }
 
     @Override
     public void visit(BLangForeach foreach) {
         this.acceptNode(foreach.body, symbolEnv);
+        this.acceptNode(foreach.onFailClause, symbolEnv);
     }
 
     @Override
@@ -474,6 +479,7 @@ public class EnvironmentResolver extends BLangNodeVisitor {
             this.scope = this.symbolEnv;
             matchNode.patternClauses.forEach(patternClause -> acceptNode(patternClause, this.symbolEnv));
         }
+        this.acceptNode(matchNode.onFailClause, symbolEnv);
     }
 
     @Override
@@ -1123,6 +1129,7 @@ public class EnvironmentResolver extends BLangNodeVisitor {
     @Override
     public void visit(BLangDo doNode) {
         this.acceptNode(doNode.body, symbolEnv);
+        this.acceptNode(doNode.onFailClause, symbolEnv);
     }
 
     @Override
@@ -1140,10 +1147,13 @@ public class EnvironmentResolver extends BLangNodeVisitor {
 
     @Override
     public void visit(BLangRetry retryNode) {
+        this.acceptNode(retryNode.retryBody, symbolEnv);
+        this.acceptNode(retryNode.onFailClause, symbolEnv);
     }
 
     @Override
     public void visit(BLangRetryTransaction retryTransaction) {
+        this.acceptNode(retryTransaction.transaction, symbolEnv);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
```diff
@@ -411,6 +411,14 @@ public void visit(BLangWhile whileNode) {
...
             this.acceptNode(whileNode.body, this.symbolEnv);
         }
+        visitOnFailClause(whileNode.onFailClause);
+    }
+
+    private void visitOnFailClause(BLangOnFailClause onFailClause) {
+        if (onFailClause != null) {
+            SymbolEnv onFailEnv = SymbolEnv.createOnFailEnv(onFailClause, this.symbolEnv);
+            this.acceptNode(onFailClause, onFailEnv);
+        }
```
pubudu91, line 421: Shouldn't this logic be in the visit() method for the on fail clause?

pcnfernando: this was to avoid null check duplication within multiple nodes. we can move the environment creating to onfail visit if needed. wdyt?

pcnfernando: this was to avoid null check duplication in multiple places

pcnfernando: we can move the env creation to onfail clause if needed? wdyt?

pubudu91: There won't be any duplication right? What I meant was to move this logic to the `visit(BLangOnFailClause)` method. And since the on fail clause seem to have a block stmt, I think we can just simply call `acceptNode(onFail.body, symbolEnv)` in the visit.

pcnfernando: Updated in 1b130a542273ac48fa116aeee3fb56c14953dba2

anupama-pathirage:
@pcnfernando The commit messages and PR heading is not according to the best practices. Please correct them.

A properly formed Git commit subject line should always be able to complete the following sentence:

If applied, this commit will your subject line here

Refer [1] for more details.

[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md#submit-your-contribution

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
```diff
@@ -411,6 +411,7 @@ public void visit(BLangWhile whileNode) {
             this.scope = this.symbolEnv;
             this.acceptNode(whileNode.body, this.symbolEnv);
         }
+        this.acceptNode(whileNode.onFailClause, symbolEnv);
```
pubudu91, line 414: The while stmt's pos includes the on fail clause right? If so, this call should be within the above if block.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
```diff
@@ -474,6 +479,7 @@ public void visit(BLangMatch matchNode) {
             this.scope = this.symbolEnv;
             matchNode.patternClauses.forEach(patternClause -> acceptNode(patternClause, this.symbolEnv));
         }
+        this.acceptNode(matchNode.onFailClause, symbolEnv);
```
pubudu91, line 482: Same comment as above regarding this.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolEnv.java
```diff
@@ -222,13 +222,15 @@ public static SymbolEnv createRetryEnv(BLangRetry node, SymbolEnv env) {
     }
 
     public static SymbolEnv createOnFailEnv(BLangOnFailClause node, SymbolEnv env) {
```
pubudu91, line 224: Not exactly related to this PR. But if the on fail clause has a block stmt in it, do we need to be creating a separate env like this? Wouldn't it automatically be handled when analyzing the block stmt?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
pcnfernando: Updated in b12ab10395f186458914384cb49212a460a18833

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolEnv.java
pcnfernando: Seems when the node changed, the environment creation was not reflected. Updated in b12ab10395f186458914384cb49212a460a18833

pcnfernando: It seems unless the OnFailClause is used to create the symbol environment, the variable definition symbol within on fail clause needs to be added manually while desugaring it to a lambda function. Hence reverted this change

mohanvive (approved):

PR #28288, opened 2021-01-29
Add Custom error message support for Toml Schema validator

### LINKED ISSUE #27903: Need to add custom error messages in toml parser
**Description:**
Currently, if some property in a toml file does not align with its validations mentioned in the validation schema, it gives errors like this.
```
Key "name" value does not match the Regex provided in Schema ^lang.[a-z0-9_]*$|^[a-zA-Z0-9_]*$
```
But we need to give custom error messages like below.
```
invalid Ballerina.toml file: Invalid 'name' under [package]: 'winery-project' : 'name' can only contain alphanumerics, underscores and the maximum length is 256 characters.
```
For that, we need a mechanism to add custom messages to the validation scheme as given below.
```json
{
"id": "DOBChecker",
"type": "object",
"properties": {
    "DOB": {
        "type": "string",
        "required": true,
        "pattern": "/^(([1-9]|0[1-9]|1[0-9]|2[1-9]|3[0-1])[-](JAN|FEB|MAR|APR|MAY|JUN|JULY|AUG|SEP|OCT|NOV|DEC)[-](\d{4}))$/i",
        "message": {
            "required": "Date of Birth is Required Property",
            "pattern": "Correct format of Date Of Birth is dd-mmm-yyyy"
        }
    }
}
```

**Steps to reproduce:**

**Affected Versions:**

xlight05:
## Purpose
$Subject.

Fixes #27903 

## Approach
Introduced a seperate field for each schema, called `message`. Message can contain error messages for all the validations. Keys of the message field is equals to the validation key. See the sample below.  and check `CustomErrorTest.java` for usage. 

## Samples
```json 
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "title": "Ballerina Manifest Spec",
    "description": "Schema for Ballerina Manifest",
    "type": "object",
    "additionalProperties": false,
    "properties": {
        "package": {
            "type": "object",
            "additionalProperties": false,
            "properties": {
                "name": {
                    "type": "string",
                    "message" : {
                        "type" : "Invalid Type"
                    }
                },
                "org": {
                    "type": "string",
                    "pattern": "^(?!\\s*$).+",
                    "message" : {
                        "pattern" : "Org cant be empty"
                    }
                },
                "version": {
                    "type": "string",
                    "pattern": "^(?!\\s*$).+"
                },
                "port": {
                    "type": "integer",
                    "minimum": 0,
                    "message" : {
                        "minimum" : "Port should be positive int"
                    }
                },
                "capacity": {
                    "type": "integer",
                    "maximum": 5,
                    "message" : {
                        "maximum" : "Max Capacity is 5"
                    }
                },
                "repository": {
                    "type": "string"
                },
                "keywords": {
                    "type": "array",
                    "items": {
                        "type": "string"
                    }
                },
                "exported": {
                    "type": "array",
                    "items": {
                        "type": "string"
                    }
                }
            },
            "required": ["name","org"],
            "message" : {
                "required" : "Field ${property} is required",
                "additionalProperties" : "Field ${property} is not supported"
            }
        }
    }
}

```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java
```diff
@@ -55,6 +55,8 @@
  */
 public class SchemaValidator extends TomlNodeVisitor {
 
+    public static final String PROPERTY_HOLDER = "${property}";
```
pramodya1994, line 58: Access can be private?

misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java
```diff
@@ -66,12 +68,11 @@ public SchemaValidator(Schema schema) {
     public void visit(TomlTableNode tomlTableNode) {
         if (schema.type() != Type.OBJECT) {
             TomlDiagnostic diagnostic = getTomlDiagnostic(tomlTableNode.location(), "TVE0002", "error.invalid.type",
```
pramodya1994, line 70: Shall we define constants for `"TVE0002"` and `"error.invalid.type"`?

misc/toml-parser/src/main/java/io/ballerina/toml/validator/schema/Schema.java
```diff
@@ -58,9 +65,7 @@ public static Schema from(Path jsonPath) throws IOException {
         Gson gson = new GsonBuilder()
                 .registerTypeAdapter(AbstractSchema.class, new SchemaDeserializer()).create();
         BufferedReader reader = Files.newBufferedReader(jsonPath);
-        Schema rootSchema = gson.fromJson(reader, Schema.class);
-        rootSchema.setType(Type.OBJECT);
-        return rootSchema;
+        return (Schema) gson.fromJson(reader, AbstractSchema.class);
```
pramodya1994, line 68: Seems like this casting is unnecessary

misc/toml-parser/src/main/java/io/ballerina/toml/validator/schema/Schema.java
```diff
@@ -72,16 +77,30 @@ public static Schema from(Path jsonPath) throws IOException {
     public static Schema from(String jsonContent) {
         Gson gson = new GsonBuilder()
                 .registerTypeAdapter(AbstractSchema.class, new SchemaDeserializer()).create();
-        Schema rootSchema = gson.fromJson(jsonContent, Schema.class);
-        rootSchema.setType(Type.OBJECT);
-        return rootSchema;
+        return (Schema) gson.fromJson(jsonContent, AbstractSchema.class);
```
pramodya1994, line 80: Seems like this casting is unnecessary

misc/toml-parser/src/main/java/io/ballerina/toml/validator/schema/SchemaDeserializer.java
```diff
@@ -162,6 +175,51 @@ private String parseOptionalStringFromJson(JsonObject jsonObject, String key) {
...
+        if (!(minimum == null || minimum.isJsonNull())) {
+            String customMinMessage = minimum.getAsString();
+            customMessages.put(MINIMUM, customMinMessage);
+        }
+
+        JsonElement maximum = customMessageJson.get(MAXIMUM);
+        if (!(maximum == null || maximum.isJsonNull())) {
+            String customMaxMessage = maximum.getAsString();
+            customMessages.put(MAXIMUM, customMaxMessage);
+        }
```
pramodya1994, line 219: Can we replace this code with follwoing method to avoid duplicates?
```java
private void addFieldToCustomMessagesMap(Map<String, String> customMessages, JsonObject customMessageJson,
            String field) {
        JsonElement type = customMessageJson.get(field);
        if (!(type == null || type.isJsonNull())) {
            String customTypeMessage = type.getAsString();
            customMessages.put(field, customTypeMessage);
        }
    }
```
Then you can invoke
```java
        addFieldToCustomMessagesMap(customMessages, customMessageJson, TYPE);
        addFieldToCustomMessagesMap(customMessages, customMessageJson, PATTERN);
        addFieldToCustomMessagesMap(customMessages, customMessageJson, REQUIRED);
        addFieldToCustomMessagesMap(customMessages, customMessageJson, ADDITIONAL_PROPERTIES);
        addFieldToCustomMessagesMap(customMessages, customMessageJson, MINIMUM);
        addFieldToCustomMessagesMap(customMessages, customMessageJson, MAXIMUM);

**Review by pramodya1994 — CHANGES_REQUESTED** [review]
Pls check comments added

File: misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java (hunk 2 above)
Comment (xlight05) [reply misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java:70]: This should be addressed with templates with all the error messages. We've created an issue to keep track of this. #27041
Shall we do this change with that PR?

File: misc/toml-parser/src/main/java/io/ballerina/toml/validator/schema/Schema.java (hunk 3 above)
Comment (xlight05) [reply misc/toml-parser/src/main/java/io/ballerina/toml/validator/schema/Schema.java:68]: Casting is needed as the static method should return a ObjectSchema(Schema) instead of AbstractSchema. 
Note in the old code we called `gson.fromJson(reader, Schema.class);` and now we call `gson.fromJson(reader, AbstractSchema.class);`

File: misc/toml-parser/src/main/java/io/ballerina/toml/validator/schema/Schema.java (hunk 4 above)
Comment (xlight05) [reply misc/toml-parser/src/main/java/io/ballerina/toml/validator/schema/Schema.java:80]: Casting is needed as the static method should return a ObjectSchema(Schema) instead of AbstractSchema. 
Note in the old code we called `gson.fromJson(reader, Schema.class);` and now we call `gson.fromJson(reader, AbstractSchema.class);`

**Comment by pubudu91** [comment]
What would the full form of these errors look like when printed on the console?

**Comment by xlight05** [comment]
> Pls check comments added
Thanks. Addressed the issues, please check

> What would the full form of these errors look like when printed on the console?
When we add the following changes to C2C(Kubernetes.toml)
```toml
"repository": {
    "type": "string",
    "pattern": "^(?!\\s*$).+",
    "message" : {
        "pattern" : "repository cant be empty"
    }
}
```
If the repository is empty in the Kubernetes.toml, `repository=""` 

It'll give the following error on the console while compiling. 
ERROR [Kubernetes.toml:(4:12,4:14)] repository cant be empty

misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java
```diff
@@ -88,26 +89,43 @@ public void visit(TomlTableNode tomlTableNode) {
...
+    }
+
+    private String getUnexpectedPropertyErrorMessage(String key) {
+        Map<String, String> message = this.schema.message();
+        String typeCustomMessage = message.get(SchemaDeserializer.ADDITIONAL_PROPERTIES);
+        if (typeCustomMessage == null) {
+            return "Unexpected Property \"" + key + "\"";
+        }
+        return typeCustomMessage.replace(PROPERTY_HOLDER, key);
+    }
```
pubudu91, line 121: These errors get printed in the format you mentioned right? I think we need to revisit the format and phrasing of these sort of error messages to align with the other compiler/parser error messages. e.g., start with lower case letters, using single quotes to quote keys/fields etc.

pubudu91: @MaryamZi

pramodya1994: @pubudu91 @MaryamZi  I'm currently integrating this toml parser with schema validation to replace the validations we do in the `ManifestBuilder` class. I will mention you in the PR, once I'm done with it.

xlight05: @pubudu91 Yes. removed caps and double quotes from all the default error messages and test cases in the recent commit.

misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java
```diff
@@ -88,26 +89,43 @@ public void visit(TomlTableNode tomlTableNode) {
...
+            return "missing required field '" + field + "'";
+        }
+        return typeCustomMessage.replace(PROPERTY_HOLDER, field);
+    }
+
+    private String getUnexpectedPropertyErrorMessage(String key) {
+        Map<String, String> message = this.schema.message();
+        String typeCustomMessage = message.get(SchemaDeserializer.ADDITIONAL_PROPERTIES);
+        if (typeCustomMessage == null) {
+            return "unexpected property '" + key + "'";
```
pubudu91, line 118: Shall we rephrase this error message so that from the error message itself it's evident that the key/val pair is invalid because it's not in the schema? Maybe something like,
```
key 'foo' not supported in schema 'MySchema'
```

misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java
```diff
@@ -130,13 +148,31 @@ public void visit(TomlValueNode tomlValue) {
         visitNode(tomlValue);
     }
 
+    private String getTypeErrorMessage(Type found) {
+        Map<String, String> message = this.schema.message();
+        String typeCustomMessage = message.get(SchemaDeserializer.TYPE);
+        if (typeCustomMessage == null) {
+            return String.format("key '%s' expects %s . found %s", this.key, schema.type(), found);
```
pubudu91, line 155: How about something like,
```
incompatible type for key 'foo': expected 'T1', found 'T2'
```

misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java
```diff
@@ -130,13 +148,31 @@ public void visit(TomlValueNode tomlValue) {
...
+        }
+        return typeCustomMessage;
+    }
+
+    private String getPatternErrorMessage(String pattern) {
+        Map<String, String> message = this.schema.message();
+        String typeCustomMessage = message.get(SchemaDeserializer.PATTERN);
+        if (typeCustomMessage == null) {
+            return String.format("key '%s' value does not match the regex provided in schema %s", this.key,
```
pubudu91, line 164: How about,
```
value for key 'foo' expected to match the regex: <<pattern>>
```

misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java
```diff
@@ -190,38 +221,53 @@ public void visit(TomlArrayValueNode tomlArrayValueNode) {
...
+        }
+        return maxCustomMessage;
+    }
+
+    private String getMinValueDeceedErrorMessage(Double min) {
+        Map<String, String> message = this.schema.message();
+        String minCustomMessage = message.get(SchemaDeserializer.MINIMUM);
+        if (minCustomMessage == null) {
+            return String.format("key '%s' value can't be lower than %f", this.key,
+                    min);
```
pubudu91, line 250: `value for key 'foo'` instead of `key 'foo' value` for these?

misc/toml-parser/src/main/java/io/ballerina/toml/validator/SchemaValidator.java
pramodya1994: @pubudu538 ATM T1 values are all capital. eg: `ARRAY`. Shall we make it like `array`?

pubudu91 (approved):

pramodya1994 (approved):

PR #28306, opened 2021-01-30
Enable Ctrl+D to exit Ballerina shell

### LINKED ISSUE #28270: Ctrl-D should exit Ballerina shell
It's conventional that Ctrl-D exit a shell. I can (if I read the intro text) using /exit, but people may have forgotten that by the time they want to exit (I did).

Ctrl-D works but gives a `Something went wrong while executing REPL: org.jline.reader.EndOfFileException`. You should just catch this an exit cleanly.

kdsuneraavinash:
## Purpose
Enables user exit with Ctrl+D

Fixes #28270

## Approach
Both /exit and Ctrl+D (which causes an `EndOfFileException`) will both throw an `ShellExitException` which will be handled at the main shell loop. It will now print `Bye!!!` there and exit.

## Samples

Can exit by pressing Ctrl+D
```
Welcome to Ballerina Shell REPL.
Type /exit to exit and /help to list available commands.

=$
| Bye!!!
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/ShellExitException.java
```diff
@@ -0,0 +1,26 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.ballerina.shell.cli;
+
+/**
+ * Exception to signify that shell wants to exit.
+ * After this is called no new input/output can be done on shell.
```
MaryamZi, line 23: Let's add the `@since` annotation.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BallerinaShell.java
```diff
@@ -98,24 +98,31 @@ public void run() {
...
-                if (!evaluator.hasErrors()) {
-                    terminal.fatalError("Something went wrong: " + e.getMessage());
+                String source = terminal.readLine(leftPrompt, rightPrompt).trim();
+                start = Instant.now();
+                try {
+                    if (!commandHandler.handle(source)) {
+                        String result = evaluator.evaluate(source);
+                        terminal.result(result);
+                    }
+                } catch (ShellExitException e) {
```
MaryamZi, line 109: Is it an issue if the finally block is executed here? If not, we can have a single try block, right?

MaryamZi:
@kdsuneraavinash, as discussed, let's check if we can add a test for this.

MaryamZi (approved):

PR #28329, opened 2021-02-01
Introduce dotted key support for Toml parser

### LINKED ISSUE #27769: Revisit Toml Parser API Design
**Description:**
Currently toml parser API follows chaining approach. 
Ex - toml.getTable("foo").getTable("bar").get("bear")
This causes high change of getting NPE when chaining.

As a solution, dotted key based retrival should be supported with java optionals.

### LINKED ISSUE #27789: TOML Parser returns wrong nodes for dotted TOML keys
**Description:**
$subject

**Steps to reproduce:**
The following TOML table and the key-value pairs must be returning similar table nodes when parsing. 
```toml
[abc.main]
intKey = 42
floatKey = 3.5
stringKey = "x" 

abc.main.intKey = 42
abc.main.floatKey = 3.5
abc.main.stringKey = "x" 
```
The resulting node of the table is 
```
TOML table - abc 
           | --- TOML table - main
                            | --- TOML Key-Value - "intKey" = 42                              
                            | --- TOML Key-Value - "floatKey" = 3.5  
                            | --- TOML Key-Value - "stringKey" = "x" 

```
And the resulting node of the dotted keys are 
``` 
TOML table - abc 
           | --- TOML table - main
                            | --- TOML Key-Value - "intKey" = 42    
TOML table - main 
           | --- TOML Key-Value - "floatKey" = 3.5    
TOML Key-Value - "stringKey" = "x" 
``` 

**Affected Versions:**
slp9-SNAPSHOT

#27431 

xlight05:
## Purpose
$Subject. Removes Deprecated status from `Toml` class. Contains breaking changes to `Toml` class. AFAIK only code to cloud uses this class for now, i'll be updating that code once this is merged.

Earlier Toml Parser API followed query chaining approach instead. `read.getTable("key").get("hi")` The main issue with approach is if you get NPE in the middle of the chain it becomes really hard to capture it. The reason for going for this chaining method was quoted string keys in toml spec.

However, with this PR, quoted string keys will be treated differently from regular keys. If you want to query a key value pair with quoted string key you have to include quotes with the key. 
Retrieving Key that has quoted string key.
`read.get("\"key\"")`

Toml API now also supports dotted keys for easy retrieval.
`read.get("key.hi");`

In order to avoid the NPE issue stated above, the API now supports Optionals as chaining is not needed.
`TomlStringValueNode key1 = (TomlStringValueNode) read.get("key").get();`

This PR also fixes some bugs related to dotted Keys and existing node diagnostics
Fixes #27789
Fixes #27769

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlTransformer.java
+++ b/misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlTransformer.java
@@ -226,7 +226,9 @@ public class TomlTransformer extends NodeTransformer<TomlNode> {
                 if ((targetTable).generated()) {
                     parentTable.replaceGeneratedTable(newTableNode);
                 } else {
-                    appendEntriesToTable(targetTable, tableChild);
+                    TomlDiagnostic nodeExists = dlog.error(tableChild.location(),
+                            DiagnosticErrorCode.ERROR_EXISTING_NODE, newTableNode.key().name());
+                    parentTable.addDiagnostic(nodeExists);
                 }
             } else if (topLevelNode instanceof TomlKeyValueNode) {
                 TomlDiagnostic nodeExist = dlog.error(newTableNode.location(),
@@ -234,20 +236,6 @@ public class TomlTransformer extends NodeTransformer<TomlNode> {
                 parentTable.addDiagnostic(nodeExist);
             } else {
                 throw new UnsupportedOperationException();
-            }
-        }
-    }
-
-    private void appendEntriesToTable(TomlTableNode targetTable, TomlTableNode newTable) {
-        for (Map.Entry<String, TopLevelNode> entry : newTable.entries().entrySet()) {
-            String key = entry.getKey();
-            TopLevelNode value = entry.getValue();
-            boolean isExist = targetTable.entries().containsKey(key);
-            if (!isExist) {
-                targetTable.entries().put(key, value);
-            } else {
-                TomlDiagnostic nodeExists = dlog.error(value.location(), DiagnosticErrorCode.ERROR_EXISTING_NODE, key);
-                targetTable.addDiagnostic(nodeExists);
             }
         }
     }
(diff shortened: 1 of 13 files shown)


misc/toml-parser/src/test/resources/semantic/existing-node.toml
```diff
@@ -0,0 +1,22 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+key = "first"
+key = "second"
+foo.mow = "test"
+
+[foo]
```
KRVPerera, line 21: Shouldn't we also say existing node/table `foo` here.

misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlTransformer.java
```diff
@@ -203,33 +219,45 @@ private void addChildTableToParent(TomlTableNode rootTable, TomlTableNode tableC
...
             }
         }
     }
 
-    private TomlTableNode generateTable(Map<String, TopLevelNode> childs, TomlKeyEntryNode parentString,
-                                        boolean isGenerated) {
+    private void appendEntriesToTable(TomlTableNode targetTable, TomlTableNode newTable) {
+        for (Map.Entry<String, TopLevelNode> entry : newTable.entries().entrySet()) {
+            String key = entry.getKey();
+            TopLevelNode value = entry.getValue();
```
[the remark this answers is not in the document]
xlight05: We need value for else block too, to get the location for the diagnostics.
Remove isExist as normal expression makes more sense :D

misc/toml-parser/src/test/resources/semantic/existing-node.toml
xlight05: I'm not clear on this. 
```toml
foo.mow = "test"

[foo]
mow = "wew"
```
in the following example, line `mow = "wew"` is the duplicate entry,

if you change the file to
```toml
foo.mow = "test"

[foo]
bar = "wew"
```
this is a valid toml syntax.

In Toml AST this gets represented as.
[foo]
mow = "test"
bar = "wew"

xlight05: Discussed offline and Fixed.

KRVPerera (approved):
LGTM

lochana-chathura (approved):

Issue #28315, opened 2021-01-31
Ballerina Shell should partially parse to identify if snippets are incomplete


kdsuneraavinash:
**Description:**
Currently, shell will simply use parenthesis to decide to wait for user input even on a new line. However, there are many cases where other factors affect whether or not to let user continue the snippet.
For an example,
```ballerina
int x = 1 +
```
will give a parser error rather than waiting for the rest of the expression. This is not ideal. To support these cases, Ballerina shell should support identifying whether the snippet is partially entered,

**Steps to reproduce:**
Eg:
```ballerina
int x = 1 +
```
should wait for new input even on new line. But currently it terminates.

**Affected Versions:**
Ballerina Swan Lake Alpha 1

**OS, DB, other environment details and versions:**

**Related Issues (optional):**
#28282 #28274 #28313 

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

jclark:
How do you decide whether to parse user input as an expression, a statement or a module-level declaration?

kdsuneraavinash:
Basically by running some trials, there are about 4 trials, that tries to parse as an import, module-dcln, statement, expression. 
The first trial that passes is used to determine the type. A trial simply wraps the user code and tries to parse. If it fails, then the next trial is checked, so on and so forth. (because this can take some time there is also a timeout in effect)

jclark:
Here's how I suggest you approach this.  Stage 1 and 2 are implementable in a few hours, and will make things quite a bit better than now. Stage 3 is what is needed to get it absolutely right but will need help from the compiler team.

### Stage 1

In this stage, extend the "counting parentheses" approach so that it is consistent with Ballerina syntax, and correct in the sense that if it says the input is incomplete, then it will in Ballerina definitely be incomplete. There are several cases where it isn't at the moment e.g.

```
"\"("
```

or

```
// (
```

Have a stack whose members are either {, (, [ or `. Then use a state-machine with the following states:

1. normal
2. following \
3. in double quotes
4. in double quotes, following \
5. in template (i.e. following \`)
6. in template, following $
7. following /
8. in comment (either # or // style)

Note the following points about transitions:
* when you get \` in state 1, push \` onto stack, and goto state 5
* if you get { in state 6, push { and goto state 1
* when you get a }, pop { from the stack; if the top of stack is \`, then goto state 5
* at the end of line, if you're in state 3 or 4, it's an error, regardless of the stack
[...]

Issue #28274, opened 2021-01-29
Ballerina shell cannot handle newline before opening curly brace of function definition


jclark:
If I type:

```
function parse(string s) returns int|error {
```

the shell prompts for more input. But if I do e.g.

```
function parse(string s)
```

it gives a parse error rather than waiting.

pubudu91:
Fixed with #34355

PR #28344, opened 2021-02-01
Improve Ballerina Shell CLI parser to support some partial snippets

### LINKED ISSUE #28271: Ballerina shell does not accept multiline template string
I type this:

```
byte[] data = base64`
```

I get an error. Of course, it should allow accept further lines until the closing backtick.

### LINKED ISSUE #28282: Ballerina shell cannot handle newlines in expressions
For example:

```
int x = 1 +
```

will give a parse error rather than waiting for the rest of the expression.

Related to #28274.

kdsuneraavinash:
## Purpose
Implements a custom parser to the Ballerina Shell CLI module that is customized to Ballerina Language. This will partially implement waiting on user inputs that are incomplete.

Fixes #28271 
Fixes #28282
This also partially addresses #28315

## Approach
- Implement a state machine to parse user inputs
- Implement a jline parser using the state machine
- State machine implementation is as [this](https://github.com/ballerina-platform/ballerina-lang/issues/28315#issuecomment-770507673) comment (upto stage 2)
- Remove remaining stack check in preprocessor, the error message will be shown by ballerina parser instead

## Samples
After the implementation following session is possible. (previously line breaks would trigger submit inconsistently)
```
=$ "abc   
| Tree parsing failed: missing double quote
| "abc;
| ^
| Parsing aborted because of errors.

=$ io:println("abc 
| Tree parsing failed: missing double quote
| io:println("abc;
|            ^
| Parsing aborted because of errors.

=$ string s = string `
 > Hello
 > `

=$ s 

Hello


=$ int x = 1 +
 > 2

=$ x
3
```

## Remarks
- The implementation is only a partial fix to #28315. To implement the fix completely a new parser API is required to check if a statement/expression is complete as addressed by [stage 3](https://github.com/ballerina-platform/ballerina-lang/issues/28315#issuecomment-770507673). As a result, issues like #28274 will be unaffected

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/jline/JlineBallerinaParser.java
```diff
@@ -0,0 +1,91 @@
...
+ * @since 2.0.0
+ */
+public class JlineBallerinaParser implements Parser {
+    private static final char SPACE = ' ';
+    private static final char NEW_LINE = '\n';
+
+
+    @Override
+    public ParsedLine parse(String line, int cursor, ParseContext context) throws SyntaxError {
+        assert line != null;
```
MaryamZi, line 42: Assertions aren't enabled by default, right? Is it possible that `line` will actually be null? 

https://stackoverflow.com/questions/2758224/what-does-the-java-assert-keyword-do-and-when-should-it-be-used

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/jline/JlineBallerinaParser.java
```diff
@@ -0,0 +1,91 @@
...
+        List<String> words = new LinkedList<>();
+        StringBuilder currentWord = new StringBuilder();
+
+        int wordCursor = -1;
+        int wordIndex = -1;
+        ParserStateMachine stateMachine = new ParserStateMachine();
+        for (int i = 0; i < line.length(); i++) {
+            char character = line.charAt(i);
+
+            // once we reach the cursor, set selected index.
```
MaryamZi, line 53: Shall we fix all the new comments to start with an upper case letter and end with a period?
```suggestion
            // Once we reach the cursor, set selected index.
```

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/jline/parser/ParserStateMachine.java
```diff
@@ -0,0 +1,317 @@
...
+
+/**
+ * State machine implementation for ballerina parser.
+ * Specifically used by {@link io.ballerina.shell.cli.jline.JlineBallerinaParser}.
+ * <p>
+ * In this stage, the "counting parentheses" approach is extended
+ * so that it is consistent with Ballerina syntax, and correct in the sense that if it
+ * says the input is incomplete, then in Ballerina it will definitely be incomplete.
+ * <p>
+ * There are several cases where it isn't at the moment e.g. {@code "\"(", // (}
```
MaryamZi, line 33: Is this still to be done? Do we need to create an issue?

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/jline/parser/ParserStateMachine.java
```diff
@@ -0,0 +1,317 @@
...
+            case AFTER_FORWARD_SLASH:
+                afterForwardSlashState(character);
+                break;
+            case IN_COMMENT:
+                inCommentState(character);
+                break;
+            case AFTER_OPERATOR:
+                afterOperatorState(character);
+                break;
+            default:
```
MaryamZi, line 110: We don't need this, right? Or do need to do something here for other states (if any).

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/jline/parser/ParserStateMachine.java
```diff
@@ -0,0 +1,317 @@
...
+            return;
+        }
+
+        switch (state) {
+            case NORMAL:
+                normalState(character);
+                break;
+            case AFTER_BACKWARD_SLASH:
+                afterBackwardSlashState();
+                break;
```
MaryamZi, line 88: We can just return in each case instead of breaking, right?

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/jline/parser/ParserStateMachine.java
```diff
@@ -0,0 +1,317 @@
...
+            case IN_COMMENT:
+                inCommentState(character);
+                break;
+            case AFTER_OPERATOR:
+                afterOperatorState(character);
+                break;
+            default:
+        }
+
```
MaryamZi, line 112: Extra new line.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/jline/JlineBallerinaParser.java
kdsuneraavinash: No they are not,
but this should not get null, but I added so that test cases would fail if they did. I can remove if wanted

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/jline/parser/ParserStateMachine.java
kdsuneraavinash: I think #28315 can track this. This PR is only a partial fix to that issue, solving it completely would require a parser API.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/jline/parser/ParserStateMachine.java
kdsuneraavinash: The default case? spotbugs complained when I didn't include a default case, so thats why I added it.

MaryamZi (approved):

PR #28350, opened 2021-02-01
Fix several issues with tuples

### LINKED ISSUE #21586: Rest descriptor not given in the string representation of a tuple typedesc with rest member
Consider a tuple such as the following:
```ballerina
[string, int, Person[]...] tup = ["records", 100];
```
When you print the typedesc of `tup`, you get the following:
```
typedesc [string,int]
```
This string representation should also include the rest member of the tuple

### LINKED ISSUE #23747: BIRPackageSymbolEnter does not read the tuple-rest-descriptor
**Description:**
$title - see https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java#L995.

It is [written in the BIRTypeWriter](https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/writer/BIRTypeWriter.java#L241) though. 

A balo test with the rest-descriptor fails atm.

Module foo
```ballerina
public function foo() returns [int, string...] {
    return [1, "hello"];
}
```

Test
```ballerina
function test() {
    string s = se:foo()[3]; // list index out of range: index: '3'
}
```

### LINKED ISSUE #24764: Issue in type checking for tuples when the source tuple has a rest descriptor
**Description:**
$title.

**Steps to reproduce:**
Compilation should fail for the following, for both `y` and `z`, but it doesn't atm.

```ballerina
public function main() {
    [int, (string|int)...] x = [1, 10, "foo"];

    [int, string...] y = x; // no error!
    [int] z = x; // no error!
}
```

**Affected Versions:**
slp1

### LINKED ISSUE #28320: Type test for tuples produces incorrect results when the source tuple has a rest descriptor
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    [int, (string|boolean)...] x = [2, "hello", true, false];
    io:println(<any> x is [int, string...]);
}
```

Prints true!

**Affected Versions:**
slalpha1

**Related issues:**
https://github.com/ballerina-platform/ballerina-lang/issues/24764

MaryamZi:
## Purpose
$title.

Fixes #23747 - BIRPackageSymbolEnter does not read the tuple-rest-descriptor
Fixes #24764 - Issue in type checking for tuples when the source tuple has a rest descriptor
Fixes #28320 - Type test for tuples produces incorrect results when the source tuple has a rest descriptor
Fixes #21586 - Rest descriptor not given in the string representation of a tuple typedesc with rest member

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BTupleType.java
```diff
@@ -144,7 +144,17 @@ public boolean equals(Object o) {
             return false;
         }
         BTupleType that = (BTupleType) o;
-        return this.readonly == that.readonly && Objects.equals(tupleTypes, that.tupleTypes);
+
+        if ((this.restType == null || that.restType == null) && this.restType != that.restType) {
+            return false;
+        }
+
+        if (this.restType == null) {
```
KRVPerera, line 152: Shall we refactor here differently to take common branches out. I think it is easy to understand when should two tuple types should be equal that way.

eg:
```Java
if (this.readonly != that.readonly) {
  return false;
}

if (!Objects.equals(tupleTypes, that.tupleTypes)) {
  return false;
 }

if (this.restType != null) {
this.restType.equals(that.restType);
}

```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2577,7 +2584,12 @@ public Boolean visit(BTupleType t, BType s) {
                     return false;
                 }
             }
-            return true;
+
+            if (sourceRestType == null || targetRestType == symTable.noType) {
+                return true;
```
KRVPerera, line 2589: Do we need null check here since the logic in line 2575. If we need to allow **noType** for `targetRestType ` we need to change previous logic right? Now when we reach here both `sourceRestType` and `targetRestType` are null.

```Java
if ((sourceRestType == null || targetRestType == null) && sourceRestType != targetRestType) {
   return false;
}
```

MaryamZi: No, right? 

We can reach here if 
- `sourceRestType` and `targetRestType` are both null or
- when neither is null 

So `sourceRestType == null || targetRestType == symTable.noType` will evaluate to true if 
- both `sourceRestType` and `targetRestType` are null or
- `sourceRestType` is not null and `targetRestType` is `symTable.noType`?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BTupleType.java
MaryamZi: Moved the read-only check to the top. But for the rest vs individual member comparisons, I think it'll be better to handle and be done with the rest scenario first, because it's either null or a comparison of a single pair of types. WDYT?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: Is the output for this situation as False correct?
`sourceRestType == null && targetRestType == symTable.noType`

MaryamZi: Yes, this noType case is for scenarios where var is used. For example,

```
var [a, ...b] = getTuple();  

function getTuple() returns [int] => [1];
```

The same type check is done with a tuple created for the binding pattern used with var - `[noType, noType...]`. Here `targetRestType` is `symTable.noType` and `sourceRestType` is null, so this returning false and causing a compilation error is correct.

KRVPerera: Okay I agree but my concern is that logic is complex here.
1. When `sourceRestType == null` we are not actually testing whether it is null but whether both are null. 
2. ```sourceRestType == null || targetRestType == symTable.noType```
  here looking at this line I would expect `sourceRestType == null && targetRestType == symTable.noType` to be true too but it false.

KRVPerera (approved):
LGTM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
MaryamZi: Agreed that you need to have the preceding logic in mind. But was trying to avoid duplicate checks. 

Also, the noType scenario is kind of a special case.

PR #28392, opened 2021-02-02
Remove the use of a Security Manager in Ballerina Shell

### LINKED ISSUE #28048: [Ballerina Shell] Remove the use of custom Security Manager
**Description:**
Ballerina Shell uses a custom security manager to stop the VM from exiting when executing snippets.
However, this approach is overkill and should be replaced with a better approach.

Alternative approaches:
1. Say if we are starting the shell, then we don't have to exit the program. This logic can be included in the backend generation phase.
2. Use a similar approach as testerina to directly invoke the method, rather than invoking the generated main method.
3. Start the invocation as a separate process

kdsuneraavinash:
## Purpose
Removes the use of Security Manager in Ballerina Shell and instead uses a similar approach as testerina by directly invoking the required functions.

Fixes #28048 

## Approach
- Directly invokes `$_init.$moduleInit()` method and `main.main()` methods
- Followed approach is similar to that of testerina
- Remove security manager and silencing STDERR

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
```diff
@@ -53,16 +69,36 @@
  * @since 2.0.0
  */
 public abstract class Invoker extends DiagnosticReporter {
+    public static final String MODULE_NOT_FOUND_CODE = "BCE2003";
+    // TODO: After configurables can be supported, change this to that file location
```
riyafa, line 73: Shall we create an issue also and point the issue here along with the TODO? or can we fix this because the configurables are now supported

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
```diff
@@ -163,6 +241,168 @@ protected PackageCompilation compile(Project project) throws InvokerException {
...
+        try {
+            addDiagnostic(Diagnostic.debug(String.format("Running %s.%s on schedule", className, methodName)));
+
+            // Get class and method references
+            Class<?> clazz = classLoader.loadClass(className);
+            Method method = clazz.getDeclaredMethod(methodName, Strand.class);
+            Function<Object[], Object> methodInvocation = createInvokerCallback(method);
+
+            // Schedule and run the function and return if result is valid
+            BFuture out = scheduler.schedule(new Object[1], methodInvocation, null, null, null,
```
riyafa, line 320: Use the apis in `io.ballerina.runtime.api.Runtime` to invoke methods. The methods in the schedulers are not supposed to be used

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
```diff
@@ -53,16 +69,36 @@
...
+    private final Scheduler scheduler;
     /**
      * File object that is used to create projects and write.
      * Depending on USE_TEMP_FILE flag, this may be either a file in cwd
      * or a temp file.
      */
     private File bufferFile;
 
+    protected Invoker() {
+        this.scheduler = new Scheduler(false);
```
riyafa, line 99: It is incorrect to new a scheduler. This class is exported to only few packages and not supposed to be used by all.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
```diff
@@ -163,6 +241,168 @@ protected PackageCompilation compile(Project project) throws InvokerException {
...
+            // Get class and method references
+            addDiagnostic(Diagnostic.debug(String.format("Running %s.%s directly", className, methodName)));
+            Class<?> clazz = classLoader.loadClass(className);
+            Method method = clazz.getDeclaredMethod(methodName, argTypes);
+            Object result = method.invoke(null, args);
+            addDiagnostic(Diagnostic.debug("Result: " + result));
+            return result;
+        } catch (ClassNotFoundException e) {
+            addDiagnostic(Diagnostic.error(className + " class not found: " + e.getMessage()));
+            throw new InvokerException(e);
```
riyafa, line 370: We can move this to a method and pass the required string value as a parameter. This can reduce code duplication in multiple places.

bvm/ballerina-runtime/src/main/java/module-info.java
```diff
@@ -48,7 +48,7 @@
     exports io.ballerina.runtime.internal.scheduling to io.ballerina.cli.utils, io.ballerina.java,
             io.ballerina.lang.array, io.ballerina.lang.error, io.ballerina.lang.internal, io.ballerina.lang.map,
             io.ballerina.lang.table, io.ballerina.lang.transaction, io.ballerina.lang.value, io.ballerina.lang.xml,
-            io.ballerina.log.api, io.ballerina.testerina.core, io.ballerina.testerina.runtime;
+            io.ballerina.log.api, io.ballerina.testerina.core, io.ballerina.testerina.runtime, io.ballerina.shell;
```
riyafa, line 51: I don't think we need to export this to shell module. Shall we remove it? As mentioned please use the `Runtime` util functions to invoke methods.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
kdsuneraavinash: These are not object methods per-se. They are `main`, `$moduleInit` and `$configureInit` functions.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
kdsuneraavinash: The issue is that shell executes each snippet by compiling and running a function on the generated jar. To invoke them with reflection, we need a `Schedular`. And shell invocations are not one program actually, each snippet runs as its own program, so during a shell session it will run multiple programs.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
```diff
@@ -53,16 +69,36 @@
  * @since 2.0.0
  */
 public abstract class Invoker extends DiagnosticReporter {
```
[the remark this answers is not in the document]
kdsuneraavinash: How about `ShellInvoker`? because this class invokes shell snippets.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
kdsuneraavinash: The problem was what should be the config.toml to use. Because for shell, there is no specific config file.
Maybe we should have a different ShellConfig.toml file.? I added this todo to change the location if is happens.
I'll create an issue for this.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
riyafa: since there's no access to strand in your case it is okay to create a `new Scheduler()`

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
riyafa: same as above

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/Invoker.java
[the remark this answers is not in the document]
riyafa: what does it invoke in the shell? Probably `ShellSnippetsInvoker` if that makes sense 🙂

riyafa (dismissed):

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
```diff
@@ -0,0 +1,455 @@
...
+ * <p>
+ * Context of an invoker is the context that will be used to
+ * fill the template. This should be a logic-less as much as possible.
+ * Invoker and its context may be tightly coupled.
+ *
+ * @since 2.0.0
+ */
+public abstract class ShellSnippetsInvoker extends DiagnosticReporter {
+    public static final String MODULE_NOT_FOUND_CODE = "BCE2003";
+    // TODO: After configurables can be supported, change this to that file location
```
riyafa, line 73: Shall we link the issue created here? Along with the todo?

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
```diff
@@ -0,0 +1,455 @@
...
+                }
+            }
+
+            return packageCompilation;
+        } catch (InvokerException e) {
+            throw e;
+        } catch (Exception e) {
+            addErrorDiagnostic("Something went wrong: " + e);
+            throw new InvokerException(e);
+        } catch (Error e) {
```
riyafa, line 238: Let us not try to catch Errors. It says in the java doc:

> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch. Most such errors are abnormal conditions. The ThreadDeath error, though a "normal" condition, is also a subclass of Error because most applications should not try to catch it. 
https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
```diff
@@ -0,0 +1,455 @@
...
+            }
+            return result;
+        } catch (ClassNotFoundException e) {
+            addErrorDiagnostic(className + " class not found: " + e.getMessage());
+            throw new InvokerException(e);
+        } catch (NoSuchMethodException e) {
+            addErrorDiagnostic(methodName + " method not found: " + e.getMessage());
+            throw new InvokerException(e);
+        } catch (RuntimeException e) {
+            addErrorDiagnostic("Unexpected error: " + e.getMessage());
```
riyafa, line 340: We are not supposed to catch `RuntimeException` but they should be fixed as much as possible. Where can we expect `RuntimeException` in this code? Can we fix them without catching them?

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
```diff
@@ -0,0 +1,455 @@
...
+                    addDiagnostic(Diagnostic.debug(diagnostic.message()));
+                }
+            }
+
+            return packageCompilation;
+        } catch (InvokerException e) {
+            throw e;
+        } catch (Exception e) {
+            addErrorDiagnostic("Something went wrong: " + e);
+            throw new InvokerException(e);
```
riyafa, line 237: Shall we try to fix potential `RuntimeException` if any?

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
riyafa: Shall we fix in all places if it is applicable?

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
```diff
@@ -0,0 +1,455 @@
...
+     *
+     * @param classLoader Class loader to find the class.
+     * @param className   Class name with the method.
+     * @param methodName  Method name to invoke.
+     * @param argTypes    Types of arguments.
+     * @param args        Arguments to provide.
+     * @return The result of the invocation.
+     * @throws InvokerException If invocation failed.
+     */
+    protected Object invokeDirectMethod(ClassLoader classLoader, String className, String methodName,
```
riyafa, line 358: Did we mean `invokeMethodDirectly`?

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
kdsuneraavinash: #28662 Created an issue

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
kdsuneraavinash: The runtime exception can occur and from the user snippet as well. Here, and in the next position, we try to invoke a method using reflection and it can throw a RuntimeException. IMO if that happens, a REPL should continue to work without crashing and just notify the user that something went wrong.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
kdsuneraavinash: The reason to name this as `invokeDirectMethod` was because its companion method name is `invokeScheduledMethod`.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
riyafa: I mean shall we add it to the comment in the code along with the todo?

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/ShellSnippetsInvoker.java
riyafa: I think if  `RuntimeException` is thrown inside a method call it is propagated as `InvocationTargetException`. I think we need to catch that instead of the `RuntimeException`.

kdsuneraavinash: I'll remove the handlers for `RuntimeException` then :)

riyafa (approved):

PR #28402, opened 2021-02-03
Fix chained invocations with rest args

### LINKED ISSUE #18968: Invalid runtime error on chained method calls for type int
**Description:**

$title

**Steps to reproduce:**
```ballerina
public function main() {
    var e = 0;
    int a = e.max(1).min(2);
}
```
Runtime Error
```
error: {ballerina/lang.array}InherentTypeViolation message=incompatible types: expected 'int', found 'int[]'
        at irshadnilam.linkedmap:main(linked_map.bal:3)
```
**Affected Versions:**
Ballerina 1.0.0
Language specification 2019R3

### LINKED ISSUE #25783: Method chaining not working for string.concat
**Description:**

$title

From Spec:- 

> if both operands are a string, then the result is a string that is the concatenation of the operands;

But returning a `string[]`.

**Steps to reproduce:**

```ballerina

// Not working. Error:- `expected string and provided string[]`
string foo = "foo";
foo = foo.concat("bar").concat("foo");

// Working
string foo = "foo";
foo = foo.concat("bar");
foo = foo.concat("foo");


```

**Affected Versions:**
jBallerina 1.2.0
Language specification 2020R1
Ballerina tool 0.8.5

grainier:
## Purpose
$subject

Fixes #25783
Fixes #18968

## Approach
n/a

## Samples
ref: tests.

## Remarks
n/a

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -7365,18 +7365,25 @@ private void reorderArguments(BLangInvocation iExpr) {
...
             List<BLangExpression> exprs = new ArrayList<>();
-
-            BArrayType arrayType = (BArrayType) invokableSymbol.restParam.type;
-            BType elemType = arrayType.eType;
-
             for (BLangExpression restArg : restArgs) {
-                exprs.add(addConversionExprIfRequired(restArg, elemType));
+                // If the restArg is an array (where expected elemType is not an array),
+                // we have to flatten that and create a single array.
+                if (elemType.tag != TypeTags.ARRAY && restArg.getKind() == NodeKind.ARRAY_LITERAL_EXPR) {
```
MaryamZi, line 7377: According to the comment in L7362, this should not happen, right? 

If it is a list that is passed for the rest param, the node kind should have been REST_ARGS_EXPR here. 

I think we should look into if there is an issue in how we're desugaring this, the order, or if we are desugaring something that has already been desugared.

grainier: Correct, but we are creating the array literal for the rest args here, and when the invocation is chained, that array literal will get passed to the next chained invocation, and L7375 cast tries to cast it to the rest param type. i.e `<string> [foo, foo1]`. So it's not `REST_ARGS_EXPR`, and it'll be the array literal we create in the previous iteration and that's why it comes here.

KRVPerera, line 7377: Shall we check `elemType.tag != TypeTags.ARRAY` outside the for loop?

grainier: You mean outside of `for (BLangExpression restArg : restArgs) {}`?

rdhananjaya (dismissed):

KRVPerera (dismissed):

mohanvive (dismissed):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: This happens because we're desugaring the expression on which the method is called twice, right? 

Shouldn't both the following be desugared in the same way.

```ballerina
int e = 0;
int a = e.max(1).min(2);
```

```ballerina
int e = 0;
int a = int:min(int:max(e, 1), 2);
```

Since the latter works and similar object method calls work, this could specifically be an issue due to how we add the expression on which the method is called as the first arg for langlib methods.

I guess we may have to revisit that logic anyway when fixing https://github.com/ballerina-platform/ballerina-lang/issues/28580.

grainier: @MaryamZi, updated the PR 666f225.

mohanvive (approved):

Issue #27744, opened 2021-01-08
[Debugger] Order of the child elements are not preserved for ballerina arrays in the debug view variable presentation 


NipunaRanasinghe:
**Description:**
$subject. Please refer the below example.
![Screenshot from 2021-01-07 16-38-33](https://user-images.githubusercontent.com/29032600/104013709-d519fa00-51d7-11eb-8bf0-42d0ce937688.png)

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Issue #24605, opened 2020-07-06
[Debugger] Add support for variable querying 


NipunaRanasinghe:
**Description:**
Current debugger implementation does not support variable querying, which comes useful in contexts such as 
-  when the number of visible variables is too large to be processed at once.
- when processing compound variables (arrays,json, xml) which may contain a large number of elements/children.

Therefore the debugger implementation should be refactored to honor variable request arguments, to limit the number of variables that should be contained in the response.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #28435, opened 2021-02-03
[Debugger] Add variable paging support

NipunaRanasinghe:
## Purpose
This PR introduces variable paging support for the following variable types in vscode debug view.
- arrays
- JSON
- maps
- XML 
- tables
- tuples

With this feature, ballerina variables which contains a large number of child variables will be shown in a paged view, as shown below.

ezgif com-gif-maker (1)

Along with the paging support, this PR also introduces following improvements to debugger variable representation.

- preserve child variable order in the debug view for ballerina ordered variable types (i.e. arrays, tables)
- Optimize map and JSON child variable loading mechanism

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/27744 
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/24605

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -655,12 +677,37 @@ private String getStackFrameName(StackFrameProxyImpl stackFrame) {
...
+                // Handles list-type indexed variables.
+                return createVariableArrayFrom(args, childVars.getRight());
+            }
+            return new Variable[0];
+        } else if (parentVar instanceof NamedCompoundVariable) {
+            // Handles named variables.
+            Map<String, Value> childVars = ((NamedCompoundVariable) parentVar).getNamedChildVariables();
+            return createVariableArrayFrom(args, childVars);
+        }
+
```
praveennadarajah, line 705:
```suggestion
```

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/ExecutionContext.java
```diff
@@ -16,27 +16,36 @@
...
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+
 /**
  * Debugging session context.
  */
-public class DebugContext {
+public class ExecutionContext {
```
praveennadarajah, line 32: Any specific reasons for changing the class name here?
If there is a valid reason shall we update the doc comment accordingly?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -686,13 +754,89 @@ private void loadProjectInfo(Map<String, Object> clientArgs) {
...
+     * and thread objects that have completed their execution are not included in the returned list.
+     */
+    Map<Long, ThreadReferenceProxyImpl> getThreadsMap() {
+        if (context.getDebuggee() == null) {
+            return null;
+        }
+        Collection<ThreadReference> threadReferences = context.getDebuggee().getVirtualMachine().allThreads();
+        Map<Long, ThreadReferenceProxyImpl> breakPointThreads = new HashMap<>();
+
+        // Filter thread references which are at breakpoint, suspended and whose thread status is running.
```
praveennadarajah, line 769:
```suggestion
        // Filter thread references which are suspended and whose thread status is running.
```

praveennadarajah (commented):
IMO it is better to add test case(s) covering all the possible scenarios mentioned un the PR description. WDYT?

Maybe we can create an issue for now and send as a separate PR.

hevayo:
What are the thresholds for this feature to get triggered ? and what's the number of pages that we would create ?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/ExecutionContext.java
NipunaRanasinghe: yeah thought of using a more specific name, since we are already having a separate context holder for debugging suspended state.

NipunaRanasinghe:
> What are the thresholds for this feature to get triggered ? and what's the number of pages that we would create ?

@hevayo Pagination view is automatically triggered from the vscode side, if the child variable limit is more than 100 (which seems to be the default threshold defined in vscode client).

The number of pages are also decided by the vscode client, depending on the total number of child variables. And when the count gets higher (i.e. array with 100000 elements) vscode uses multi-level paging as shown in the sample below. 

ezgif com-gif-maker (1)

Therefore from the debug server side, all we have to do is 
1. computing and attaching the total number of child variables when sending variable responses and 
2. add support to compute and send a range of child variables, which will be requested from the vscode client.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
NipunaRanasinghe: I think we are allowed to keep black lines (maximum 1) within a method to improve readability :slightly_smiling_face: WDYT?

hevayo:
BTW can we check with some complex types which can have large arrays as attributes ie. json and record ? Also how the pagination works on tables.

NipunaRanasinghe:
> BTW can we check with some complex types which can have large arrays as attributes ie. json and record ? Also how the pagination works on tables.

@hevayo I have already tested the mentioned scenarios and please refer to below samples on how the feature works for different variable types and scenarios, which you have mentioned

1. complex types which can have large arrays as attributes
Peek 2021-02-05 12-46

1. tables with large number of entries
Peek 2021-02-05 12-47

NipunaRanasinghe:
> IMO it is better to add test case(s) covering all the possible scenarios mentioned in the PR description. WDYT?
> 
> Maybe we can create an issue for now and send as a separate PR.

I've created a separate issue for the moment to update test framework + add integration tests to cover all pagination scenarios.
https://github.com/ballerina-platform/ballerina-lang/issues/28451

praveennadarajah (approved):

Issue #28262, opened 2021-01-27
Dependencies can become incompatible due to anonymous types


MaryamZi:
**Description:**
Names of anonymous types in a module seem to change across compilations. 

Seems like this can cause incompatibilities when using dependencies built on a different lang build

This is probably why we've often been seeing errors like "'lang.array:ArrayIterator' cannot be cast to 'lang.array:$anonType$_1'" and no such field errors for downstream builds.

Assume v1 is the older version which will add the cast from `lang.array:ArrayIterator` to `lang.array:$anonType$_1'` which is an object type. Assume the HTTP module is built on this version.
 
Assume v2 is the newer version where `lang.array:$anonType$_1` represents a different anonymous type (may even be a record). Now if we use the HTTP module built on v1 with lang v2, the cast will fail.

**Steps to reproduce:**
Run the following sample on slp8.1

```ballerina
import ballerina/http;

public function main() {
    http:Cookie c1 = new("foo", "foo");
    http:Cookie c2 = new("bar", "bar");

    http:Request req = new;

    req.addCookies([c1, c2]);
}
```

```cmd
ballerina run qwe.bal 

Compiling source
        qwe.bal

Running executable

error: {ballerina}TypeCastError {"message":"incompatible types: 'lang.array:ArrayIterator' cannot be cast to 'lang.array:$anonType$_1'"}
        at ballerina.http.1_0_4.Request:addCookies(http_request.bal:546)
           qwe:main(qwe.bal:8)
```

**Affected Versions:**
slalpha1

PR #28462, opened 2021-02-06
Remove casts to imported anonymous types and add a validation to disallow the same

MaryamZi:
## Purpose
$title, as a temporary fix for https://github.com/ballerina-platform/ballerina-lang/issues/28262.

Not sure if we can write a test for either change.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
@@ -5606,8 +5606,7 @@ public class Desugar extends BLangNodeVisitor {
 
         List<BVarSymbol> params = ((BInvokableSymbol) iExpr.symbol).params;
 
-        // Start from index `1`, since for langlib methods index `0` will be the value itself.
-        for (int i = 1; i < requiredArgs.size(); i++) {
+        for (int i = 0; i < requiredArgs.size(); i++) {
             requiredArgs.set(i, addConversionExprIfRequired(requiredArgs.get(i), params.get(i).type));
         }
     }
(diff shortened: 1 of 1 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -4435,8 +4438,11 @@ private BLangBlockStmt desugarForeachToWhile(BLangForeach foreach, BLangSimpleVa
         // However, we are within the while loop. hence the $result$ can never be nil nor error.
         // Therefore cast $result$ to non-nilable type. i.e `int item = <>$result$.value;`
         BLangFieldBasedAccess valueAccessExpr = getValueAccessExpression(foreach.pos, foreach.varType, resultSymbol);
-        valueAccessExpr.expr = addConversionExprIfRequired(valueAccessExpr.expr,
-                types.getSafeType(valueAccessExpr.expr.type, true, false));
+
+        BLangExpression expr = valueAccessExpr.expr;
+        // Since `$result$` is always a record (with a single `value` field), add a cast to `map<any|error>`
+        // to avoid casting to an anonymous type - https://github.com/ballerina-platform/ballerina-lang/issues/28262.
+        valueAccessExpr.expr = addConversionExprIfRequired(expr, symTable.mapAllType);
```
rdhananjaya, line 4445: Out of curiosity, what happens if we don't have this conversion here? Issues in jvm codegen or does it get re-added somehow via desugaring the desugared constructs?

rdhananjaya (approved):

KRVPerera (approved):
Better if we can add a positive test case which was failing previously with anonemous type cast.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: So the type of `expr` (`$result$`) here is the union of a record and nil - `record {|T value;|}?`. Although semantically it is invalid to do `$result$.value` in such a scenario, desugaring proceeds as a field access on a record union due to the check at https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java#L5171. And then at org.wso2.ballerinalang.compiler.bir.codegen.JvmInstructionGen#generateMapLoadIns we don't seem to be adding a cast so probably works as is there?

MaryamZi:
> Better if we can add a positive test case which was failing previously with anonemous type cast.

Any suggestions on how we can add such a test? 

There are several conditions that needs to be met. For example if we consider using `foreach` with an array
1. need to have two compatible lang.array versions where both have different names for the anonymous types
2. need to build the JAR using the pack with one version and then run it using a pack with the other

MaryamZi:
> > Better if we can add a positive test case which was failing previously with anonemous type cast.
> 
> Any suggestions on how we can add such a test?
> 
> There are several conditions that needs to be met. For example if we consider using `foreach` with an array
> 
> 1. need to have two compatible lang.array versions where both have different names for the anonymous types
> 2. need to build the JAR using the pack with one version and then run it using a pack with the other

Will go ahead and merge this for now, will add a test in a separate PR if we can figure out a way to test this.

PR #28523, opened 2021-02-09
Fix tuple reverse and filter functions

riyafa:
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Reverse.java
+++ b/langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Reverse.java
@@ -18,23 +18,41 @@
 
 package org.ballerinalang.langlib.array;
 
+import io.ballerina.runtime.api.TypeTags;
+import io.ballerina.runtime.api.creators.TypeCreator;
+import io.ballerina.runtime.api.creators.ValueCreator;
+import io.ballerina.runtime.api.types.Type;
 import io.ballerina.runtime.api.values.BArray;
+import org.ballerinalang.langlib.array.utils.ArrayUtils;
+
+import static org.ballerinalang.langlib.array.utils.ArrayUtils.createOpNotSupportedError;
 
 /**
  * Native implementation of lang.array:reverse((any|error)[]).
  *
  * @since 1.0
  */
-//@BallerinaFunction(
-//        orgName = "ballerina", packageName = "lang.array", functionName = "reverse",
-//        args = {@Argument(name = "arr", type = TypeKind.ARRAY)},
-//        returnType = {@ReturnType(type = TypeKind.ARRAY)},
-//        isPublic = true
-//)
 public class Reverse {
 
     public static BArray reverse(BArray arr) {
-        arr.reverse();
-        return arr;
+        Type arrType = arr.getType();
+        BArray reversedArr;
+        switch (arrType.getTag()) {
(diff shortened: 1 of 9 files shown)


langlib/langlib-test/src/test/resources/test-src/arraylib_test.bal
```diff
@@ -231,13 +232,6 @@ function testReverseRecord() {
         {name: "John Doe", age: 25, designation: "Software Engineer"}]);
 }
 
-function testArrayReverseEquality() {
-    int[] x = [1, 2, 3, 4, 5];
-    int[] y = x.reverse();
-    assertValueEquality(x == y, true);
```
KRVPerera, line 237: I think we are still modifying the argument. Here `assertValueEquality(x == y, true)`
Result should be false but it is still true. Instead of removing these we can keep them and revert the assertions.

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/Reverse.java
```diff
@@ -18,23 +18,82 @@
...
+            memTypes.add(sameType);
+        }
+        for (int i = 1; i < tupleTypes.size(); i++) {
+            isSameType &= sameType == tupleTypes.get(i);
+            memTypes.add(tupleTypes.get(i));
+        }
+        Type restType = tupleType.getRestType();
+        if (restType != null) {
+            isSameType = false;
+            memTypes.add(restType);
```
warunalakshitha, line 78: create empty array from tuple has been duplicated. Can we have a common method. We can put into ArrayUtils class

langlib/lang.array/src/main/java/org/ballerinalang/langlib/array/utils/ArrayUtils.java
```diff
@@ -81,6 +89,38 @@ public static BError createOpNotSupportedError(Type type, String op) {
...
+            isSameType &= sameType == tupleTypes.get(i);
+            memTypes.add(tupleTypes.get(i));
+        }
+        Type restType = tupleType.getRestType();
+        if (restType != null) {
+            isSameType = false;
+            memTypes.add(restType);
+        }
+        if (isSameType) {
+            ArrayType type = TypeCreator.createArrayType(sameType);
```
warunalakshitha, line 116: Should the new array type always be a union type? AFAIU we check duplicate types? Can we simply have a set for memTypes?

warunalakshitha: Shall we add a comment here?

riyafa: According to James it has to be union type: https://github.com/ballerina-platform/ballerina-spec/issues/731#issuecomment-774936212

riyafa: Added a comment in the code

warunalakshitha (approved):

Issue #28300, opened 2021-01-30
toString() on xml values sometimes fails to use character references


jclark:
toString doesn't escape properly on text items that are not nested inside elements

```
import ballerina/io;

public function main() {
   xml xy = xml`x&amp;y`;
   xml:Element e = xml`<p/>`;
   e.setChildren(xy);
   xml exy = xy + e + xy;
   io:println("Expect: x&amp;y Got: ", xy.toString());
   io:println("Expect: x&amp;y<p>x&amp;y</p>x&amp;y Got: ", exy.toString());
}
```

Gives me:

```
Expect: x&amp;y Got: x&y
Expect: x&amp;y<p>x&amp;y</p>x&amp;y Got: x&y<p>x&amp;y</p>x&y
```

Maybe the confusion comes from implicit conversion of xml:Text to string.  Note that are two distinct operations:

- toString() works on any xml value (including a value that is xml:Text) and will give you a string in XML syntax that you can parse with xml:fromString()
- implicit conversion to string works only for xml values that are xml:Text and gives you a string containing the same characters as the parsed XML; this is the getContent() lang.xml function

For example:

```
import ballerina/io;

public function main() {
   xml:Text xy = xml`x&amp;y`;
   string s = xy;
   io:println("Expect: x&amp;y Got: ", xy.toString());
   io:println("Expect: x&y Got: ", xy.getContent());
   io:println("Expect: x&y Got: ", s);
}
```

gives me the incorrect output:

```
Expect: x&amp;y Got: x&y
Expect: x&y Got: x&y
Expect: x&y Got: x&y
```

PR #28548, opened 2021-02-10
Make XMLText.stringValue go through XMLSerializer

rdhananjaya:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/28300

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -810,8 +810,8 @@ private boolean isAssignable(BType source, BType target, Set<TypePair> unresolve
             return true;
         }
 
-        if (TypeTags.isXMLTypeTag(sourceTag) && (TypeTags.isXMLTypeTag(targetTag) || targetTag == TypeTags.STRING)) {
-            return isXMLTypeAssignable(source, target, unresolvedTypes);
+        if (TypeTags.isXMLTypeTag(sourceTag) && isXMLTypeAssignable(source, target, unresolvedTypes)) {
```
MaryamZi, line 813: If the target type is an XML type or string when the source type is XML, if `isXMLTypeAssignable` returns false, we don't have to do any of the subsequent checks, right? By moving the `isXMLTypeAssignable` check to the condition aren't we introducing some additional unnecessary checks?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -822,10 +822,6 @@ private boolean isAssignable(BType source, BType target, Set<TypePair> unresolve
             return true;
         }
 
-        if (sourceTag == TypeTags.XML_TEXT && targetTag == TypeTags.CHAR_STRING) {
```
MaryamZi, line 825: Maybe we can move this also to the check above, but then again since this is getting removed anyway, maybe we can leave it here for now?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1173,8 +1167,19 @@ private boolean isXMLTypeAssignable(BType sourceType, BType targetType, Set<Type
...
-            return true;
+        } else if (sourceTag == TypeTags.XML_TEXT) {
+            if (targetTag == TypeTags.STRING || targetTag == TypeTags.CHAR_STRING) {
+                return true;
+            }
+
+            if (targetTag == TypeTags.FINITE) {
+                BFiniteType finiteType = (BFiniteType) targetType;
+                for (BLangExpression finiteValue : finiteType.getValueSpace()) {
+                    if (isXMLTypeAssignable(sourceType, finiteValue.type, unresolvedTypes)) {
```
MaryamZi, line 1178: Checking by type will not work, right? The following compiles with these changes, but should result in a compilation error?

```ballerina
type XmlValue 1|"foo";

public function main() {
    xml:Text notFoo = xml `not foo`;
    XmlValue xmlVal = notFoo;
}
```

MaryamZi: We also need to introduce these changes to the runtime type checker, right?

MaryamZi: Actually, we probably don't need this check at all at compile time, right? We only need it at runtime?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
rdhananjaya: Will re-introduce the short circuit

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
rdhananjaya: I think this gets invalidated when we remove the implicity conversion from xml text to string

rdhananjaya: Allowing this was a mistake. Will remove this, thank you.

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml_text_to_string_conversion.bal
```diff
@@ -42,16 +41,6 @@ function assignToString() {
     panic error(ASSERTION_ERR_REASON, message = "expected 'true', found 'false'");
 }
 
-function castToString() {
```
rdhananjaya, line 45: Note https://github.com/ballerina-platform/ballerina-lang/pull/28548/files#diff-7124eceb07ff413bd49ec71b405bfe3eca12f9ad154df1e0134a5decf7727c0aL31

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1173,8 +1168,10 @@ private boolean isXMLTypeAssignable(BType sourceType, BType targetType, Set<Type
                 }
                 return isAssignable(source.constraint, targetType, unresolvedTypes);
             }
-        } else if (sourceTag == TypeTags.XML_TEXT && targetTag == TypeTags.STRING) {
-            return true;
+        } else if (sourceTag == TypeTags.XML_TEXT) {
+            if (targetTag == TypeTags.STRING || targetTag == TypeTags.CHAR_STRING) {
```
MaryamZi, line 1172: Is the second condition correct? This is only true if the XML text value is of length 1 which we can't guarantee here, right?

tests/jballerina-unit-test/src/test/resources/testng-jballerina.xml
```diff
@@ -1,170 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
MaryamZi, line 1: There seem to be some deviations between these files and testng.xml. Most importantly there seem to be some additional tests here. WDYT think about leaving these files as is for the moment and removing them when fixing https://github.com/ballerina-platform/ballerina-lang/issues/28770 just to make sure we're tracking all of the missed tests? 

I don't think copying the additional tests directly to testng.xml will work since some of them actually have issues - futures for example - https://github.com/ballerina-platform/ballerina-lang/pull/28786/commits/221cdee629f6824809b8711fbaa6fa9395f31ee6, https://github.com/ballerina-platform/ballerina-lang/issues/28758

rdhananjaya: Sure will reset these

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
rdhananjaya: This is just this https://github.com/ballerina-platform/ballerina-lang/pull/28548/files#diff-51851f1451506fedd7c81b67a92282fada4eb00fe4728c70445d7262aed401f3L825

MaryamZi: Yeah, but that's wrong, right? Shall we remove the check?

The following fails at the backend atm.

```ballerina
public function main() {
    xml:Text x = xml `foo`;
    string:Char y = x;
}
```

Created an issue too https://github.com/ballerina-platform/ballerina-lang/issues/28814

rdhananjaya: Removed

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml_text_to_string_conversion.bal
```diff
@@ -81,9 +70,22 @@ function getXMLText() returns 'xml:Text {
     return xml `my world`;
 }
 
+function testXMLWithAngleBrackets() {
```
MaryamZi, line 73: Sorry, missed this earlier, but is this test file and the class the correct place to add this test? This is to test xml:Text to string conversion, right?

rdhananjaya: Agree, will move to langlib tests

MaryamZi:
@rdhananjaya, the test seems to have failed? 

```
expected: x&amp;y<p>x&amp;y</p>x&amp;y found: x&y<p>x&amp;y</p>x&y
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/BallerinaXmlSerializer.java
```diff
@@ -137,10 +137,42 @@ private void writeXMLComment(XmlComment xmlValue) throws XMLStreamException {
...
-        if (!textValue.isEmpty()) {
-            xmlStreamWriter.writeCharacters(textValue);
+        // No need to escape xml text when they are within xml element, it's handle from xml stream writer.
+        if (this.withinElement) {
+            String textValue = xmlValue.getTextValue();
+            if (!textValue.isEmpty()) {
+                xmlStreamWriter.writeCharacters(textValue);
+            }
+        } else {
+            char[] textValue = escapeCharacters(xmlValue.getTextValue());
```
MaryamZi, line 147: Wouldn't it be cleaner to do this in io.ballerina.runtime.internal.values.XmlText#stringValue that is being removed (and maybe even cache the value) instead of changing anything here?

rdhananjaya: That's a workaround to fix the current case, IMO XmlSerializer should be able to serialize anything

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/BallerinaXmlSerializer.java
```diff
@@ -163,7 +195,12 @@ private void writeElement(XmlItem xmlValue) throws XMLStreamException {
         // Write attributes
         writeAttributes(currentNSLevel, attributeMap);
 
+        // Track and override xml text escape when xml text is within an element.
+        boolean prevWithinElementFlag = withinElement;
+        withinElement = true;
         xmlValue.getChildrenSeq().serialize(this);
+        withinElement = prevWithinElementFlag;
```
MaryamZi, line 202: Wouldn't it be clearer to use `this.withinElement` here?
```suggestion
        this.withinElement = true;
        xmlValue.getChildrenSeq().serialize(this);
        this.withinElement = prevWithinElementFlag;
```

MaryamZi (approved):

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/BallerinaXmlSerializer.java
rdhananjaya: +1 updated.

MaryamZi (approved):

PR #28554, opened 2021-02-10
Improve completion support for the mapping expression context

### LINKED ISSUE #28526: No suggestion given for record fields
**Description:**
Check the below attachment. 
Screen Shot 2021-02-09 at 2 22 09 PM

I don't get the suggestion for the record type fields when trying to create one with the map.

**Steps to reproduce:**

**Affected Versions:**
ballerina-swan-lake-alpha1 (both Ballerina and VSCode plugin versions)

### LINKED ISSUE #28571: true and false Keywords are not suggested in the expression context
**Description:**
$subject

**Affected Versions:**
SL-Alpha1 at least

nadeeshaan:
## Purpose
1. Improve the mapping expression context's completion support to capture the record type from the assignment statement's left side
2. Add completion support within the conditional expression context

Fixes #28526
Fixes #28571

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MappingConstructorExpressionNodeContext.java
```diff
@@ -210,6 +207,45 @@ private boolean withinComputedNameContext(BallerinaCompletionContext context, No
...
+        if (varRef.kind() == SyntaxKind.INDEXED_EXPRESSION
+                && ((IndexedExpressionNode) varRef).containerExpression().kind() == SyntaxKind.SIMPLE_NAME_REFERENCE) {
+            varName = ((SimpleNameReferenceNode) ((IndexedExpressionNode) varRef).containerExpression()).name().text();
+        } else if (varRef.kind() == SyntaxKind.SIMPLE_NAME_REFERENCE) {
+            varName = ((SimpleNameReferenceNode) varRef).name().text();
+        } else {
+            return Optional.empty();
+        }
+
+        List<Symbol> visibleSymbols = ctx.visibleSymbols(ctx.getCursorPosition());
```
pubudu91, line 223: Wouldn't `symbol()` do for this?

pubudu91:
What about contexts such as the following for example:
```ballerina
class Person {
    string name = "";
    Address adr;

    function init(string name, Address adr) {
        self.name = name;
        self.adr = adr;
    }
}

type Address record {|
    string city;
    string country;
|};

public function main() {
    Person p = new("Pubudu", {<cursor>});
    p.adr = {<cursor>};
}
```
The above cases aren't handled right? Also, on a separate note, wouldn't it be cleaner to implement a feature like this using a syntax tree visitor?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MappingConstructorExpressionNodeContext.java
nadeeshaan: The cursor position is the position where the completion has been triggered and we need to capture the LHS symbol, which should be handled with syntax tree visitor. Syntax tree visitor based implementation should be bottom up and need a revamping in the current implementation. Which is planned for the next iteration.

nadeeshaan:
> What about contexts such as the following for example:
> 
> ```ballerina
> class Person {
>     string name = "";
>     Address adr;
> 
>     function init(string name, Address adr) {
>         self.name = name;
>         self.adr = adr;
>     }
> }
> 
> type Address record {|
>     string city;
>     string country;
> |};
> 
> public function main() {
>     Person p = new("Pubudu", {<cursor>});
>     p.adr = {<cursor>};
> }
> ```
> 
> The above cases aren't handled right? Also, on a separate note, wouldn't it be cleaner to implement a feature like this using a syntax tree visitor?

These are not handled yet and need to be addressed with a proper bottom up traversing logic. This is planned for the next iteration

pubudu91 (approved):

PR #28568, opened 2021-02-11
Synchronize compile() in PackageCompilation

pubudu91:
## Purpose
Fixes the issue we had with the compiler going to an inconsistent state due to the `lang.annotation` module's scope being corrupted due to the `env` field in `SymbolEnter` being concurrently updated at times.

## Approach
Synchronized the body of `compile()` method in `PackageCompilation` on the `CompilerContext` field of the `PackageCompilation` class to ensure that the state variables in the compiler phases don't get modified concurrently.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageCompilation.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageCompilation.java
@@ -140,16 +140,19 @@ public class PackageCompilation {
     }
 
     private void compile() {
-        List<Diagnostic> diagnostics = new ArrayList<>();
-        for (ModuleContext moduleContext : packageResolution.topologicallySortedModuleList()) {
-            moduleContext.compile(compilerContext);
-            moduleContext.diagnostics().forEach(diagnostic ->
-                    diagnostics.add(new PackageDiagnostic(diagnostic, moduleContext.moduleName())));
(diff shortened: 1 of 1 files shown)


hasithaa (approved):
Looks good to me.

rasika:
One suggestion to fix; while this synchronize the compilations there can be unnecessary compilations since all waiting threads will continue to compile since the lock happens after the `compiled` flag check. 

We have two places such in https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageCompilation.java#L110-L113 and https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageCompilation.java#L102-L105.

I mean something like below with double-checking;
```java
if (!compiled) {
            synchronized (compilerContext) {
                if (!compiled) {
                    compile();
                }
            }
        }
```

Kishanthan:
Can I suggest to move the `!compiled` boolean check completely within the compile method? That way, it is only one place we need to double check for locking.

PR #28591, opened 2021-02-11
Throw error when complete is called twice

riyafa:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Future.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Future.java
@@ -15,9 +15,15 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 package io.ballerina.runtime.api;
 
+import io.ballerina.runtime.api.creators.ErrorCreator;
+import io.ballerina.runtime.api.utils.StringUtils;
 import io.ballerina.runtime.internal.scheduling.Strand;
+import io.ballerina.runtime.internal.values.MapValueImpl;
+
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * A future that will resume the underling strand when completed.
@@ -26,12 +32,16 @@ import io.ballerina.runtime.internal.scheduling.Strand;
  */
 public class Future {
     private final Strand strand;
-
+    private final AtomicBoolean visited = new AtomicBoolean();
      Future(Strand strand) {
         this.strand = strand;
     }
 
     public void complete(Object returnValue) {
+        if (visited.getAndSet(true)) {
+            throw ErrorCreator.createError(StringUtils.fromString("cannot complete the same future twice."),
+                                           new MapValueImpl<>(PredefinedTypes.TYPE_ERROR_DETAIL));
+        }
         strand.returnValue = returnValue;
         strand.scheduler.unblockStrand(strand);
     }
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Environment.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Environment.java
@@ -24,16 +24,15 @@ import io.ballerina.runtime.internal.scheduling.Strand;
 import java.util.Optional;
 
 /**
- * When this class is used as the first argument of an interop method, Ballerina
- * will inject an instance of the class when calling. That instance can be used to
- * communicate with currently executing Ballerina runtime.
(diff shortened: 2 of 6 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Future.java
```diff
@@ -26,13 +31,19 @@
...
 public class Future {
     private final Strand strand;
+    private final AtomicBoolean visited = new AtomicBoolean(false);
 
      Future(Strand strand) {
         this.strand = strand;
     }
 
     public void complete(Object returnValue) {
+        if (visited.get()) {
```
warunalakshitha, line 41: Can we return, if the strand is already unblocked?

manuranga, line 41: let's use `compareAndSet` or `getAndSet`

warunalakshitha: +1 to the current design as its good to have an error so API user knows there are multiple complete calls.

warunalakshitha: What happens if I use markAsync twice which returns two futures and try to complete two futures. Still, strand will get unblocked twice.

warunalakshitha: strand.lock();
        if (strand.getState() != State.BLOCK_AND_YIELD) {
             throw ErrorCreator.createError(StringUtils.fromString("cannot complete the same future twice."),
                                           new MapValueImpl<>(PredefinedTypes.TYPE_ERROR_DETAIL));
         }
         strand.unlock();

manuranga: Good point. Maybe we can do the same fix for `markAsync` as well.

warunalakshitha (changes requested):
Can we write a test with interop method with Environment and then call markAsync. We can spawn a multiple java threads inside the interop function and call complete function.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Future.java
riyafa: I made the change suggested by @warunalakshitha . @manuranga I don't think we need to change `markAsync` WDYT?

manuranga: I see you have removed the atomic boolean and used a lock, isn't this slower?

riyafa: This is to prevent the strand state changing before the  if check completes.

riyafa:
@warunalakshitha I did try to write a test, but there was no way to validate the exception because it is written to the ballerina-internal log.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Future.java
manuranga: this is an edge case, should we slow down the whole thing to catch this error?

warunalakshitha: If we depend on atomic boolean inside the future object, if we have multiple futures still strand can be unblocked multiple times. We already have a lock inside unblock strand method. If this is slow, we can have another method called checkAndUnblockStarnd inside the scheduler and do the check there.

https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/scheduling/Scheduler.java#L422

warunalakshitha: IMO its good to have another check-in markAsync as manu said. Since these are user apis, its good to have validations which prevent unnecessary states in the scheduler.

riyafa: Made the future in `markAsync` as an instance variable and used atomic boolean in the future

warunalakshitha:
LGTM

warunalakshitha (approved):

PR #28623, opened 2021-02-12
Introduce a fail-safe mechanism for the parser error recovery

### LINKED ISSUE #21882: Introduce a fail-safe mechanism for the parser and error recorvery
**Description:**

Currently both the parser and error recovery have paths that are not supposed to be reached. Such invalid states cause runtime errors at the moment. i.e.: parser get crashed.

However, for some unforeseen reason, if the parser/error-recovery comes to a such an unhandled-state, the parser would crashed, and this is not an acceptable behaviour from a user's perspective. Therefore, instead of throwing a runtime error, implement a mechanism to recover from a unhandled path as well.

This is more of a fail-safe. Hence the requirement is only to be able to produce a valid tree, without the parser getting crashed. The recovered tree doesn't need to be the optimal one.

lochana-chathura:
## Purpose
$subject.

Fixes #21882 

## Approach
Let the exception be thrown, then catch and handle it. Two places to handle:

- When visiting alternative paths, one path could be throwing an exception. Catch the exception and simply ignore that path. The best alternative path would get picked from the remaining contenders.

- At the entry point exception can be thrown without going for alternative paths. Catch the exception and provide the solution as a `REMOVE` which guarantees the parser won't get stuck at the same place.

## Samples
N/A

## Remarks
In both cases mentioned in the approach, we will give `assert false` as we need to report and fix the issue.
Config your IDE to run  `io.ballerinalang.compiler.parser.test.TestParser` with  the VM option `-ea` to enable assertions.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/AbstractParserErrorHandler.java
+++ b/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/AbstractParserErrorHandler.java
@@ -178,7 +178,21 @@ public abstract class AbstractParserErrorHandler {
      * @return Recovery result
      */
     private Result seekMatch(ParserRuleContext currentCtx) {
-        return seekMatchInSubTree(currentCtx, 1, 0, true);
+        Result bestMatch;
+        try {
+            bestMatch = seekMatchInSubTree(currentCtx, 1, 0, true);
+        } catch (IllegalStateException exception) {
+            // This is a fail-safe mechanism to avoid parser being crashed in the production.
+            // We catch the exception and since we don't have any other path, return the solution as a REMOVE.
+            // We should never reach here. If we do, please open an issue.
+            assert false : "Oh no, something went bad with parser error handler: \n" +
+                    "seekMatch caught " +
+                    exception.toString();
+            bestMatch = new Result(new ArrayDeque<>(), LOOKAHEAD_LIMIT - 1);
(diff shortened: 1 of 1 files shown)


rdulmina (commented):
can we use a single try-catch block inside `seekMatchInSubTree` to get this done?

lochana-chathura:
> can we use a single try-catch block inside `seekMatchInSubTree` to get this done?

The thing is we handle two scenarios differently. 

1.If there is a problem in an alternative path we can simply ignore it and pick the best path out of the remaining alternative paths.
2. We could get an exception even without going for alternative paths. In that case, there are no alternative paths to pick as the best path, as the one and the only path we traverse is throwing an exception. In that scenario, we simply give the solution as a REMOVE action.

rdulmina:
> > can we use a single try-catch block inside `seekMatchInSubTree` to get this done?
> 
> The thing is we handle two scenarios differently.
> 
> 1.If there is a problem in an alternative path we can simply ignore it and pick the best path out of the remaining alternative paths.
> 2. We could get an exception even without going for alternative paths. In that case, there are no alternative paths to pick as the best path, as the one and the only path we traverse is throwing an exception. In that scenario, we simply give the solution as a REMOVE action.

ACK

rdulmina (approved):

Issue #28531, opened 2021-02-09
Duplicate code fixes suggestions


Kishanthan:
**Description:**
Check the below attached image where you get duplicate suggestions.

<img width="663" alt="Screen Shot 2021-02-09 at 5 19 24 PM" src="https://user-images.githubusercontent.com/690559/107359988-5cba9780-6afb-11eb-8e5c-a4616345e510.png">

**Steps to reproduce:**

**Affected Versions:**
ballerina-swan-lake-alpha1

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

nadeeshaan:
This can be caused due to caching issue. Hence we need to check this again after the #28568 merged.

PR #28652, opened 2021-02-15
Fix code actions getting duplicated issue

rasika:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/28531

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionRouter.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionRouter.java
@@ -20,6 +20,8 @@ import io.ballerina.compiler.api.symbols.TypeSymbol;
 import io.ballerina.compiler.syntax.tree.Node;
 import io.ballerina.compiler.syntax.tree.NonTerminalNode;
 import io.ballerina.compiler.syntax.tree.SyntaxTree;
+import io.ballerina.projects.Project;
+import io.ballerina.projects.ProjectKind;
 import org.ballerinalang.langserver.LSClientLogger;
 import org.ballerinalang.langserver.common.utils.CommonUtil;
 import org.ballerinalang.langserver.commons.CodeActionContext;
@@ -29,6 +31,7 @@ import org.eclipse.lsp4j.CodeAction;
 import org.eclipse.lsp4j.Position;
 import org.eclipse.lsp4j.Range;
 
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionRouter.java
```diff
@@ -78,11 +81,15 @@
...
+        Project project = ctx.workspace().project(ctx.filePath()).orElseThrow();
+        Path projectRoot = (project.kind() == ProjectKind.SINGLE_FILE_PROJECT)
+                ? project.sourceRoot().getParent() :
+                project.sourceRoot();
         // Get available diagnostics based code-actions
         ctx.allDiagnostics().stream().
-                filter(diag -> CommonUtil.isWithinRange(ctx.cursorPosition(),
-                                                        CommonUtil.toRange(diag.location().lineRange())))
+                filter(diag -> projectRoot.resolve(diag.location().lineRange().filePath()).equals(ctx.filePath()) &&
+                        CommonUtil.isWithinRange(ctx.cursorPosition(), CommonUtil.toRange(diag.location().lineRange()))
```
IMS94, line 91: While I was trying to implement a fix for this, I noticed that the `ctx.filePath()` refers to the absolute path while the `diagnostic.location().lineRange().filePath()` refers to just file name. This will cause the above equals check to fail. 

We can use something like below instead:
```
        context.filePath().endsWith(diagnostic.location().lineRange().filePath())
```
Screenshot from 2021-02-16 08-37-52

IMS94 (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionRouter.java
rasika: @IMS94 I didn't get any issues when testing this in a project, anyway will have a look. `endsWith()` is a string matching and has some more weird behaviors such as matching with other module's file with the same name(diag location.filePath() contains the relative path only).

IMS94: @rasika You are correct. I missed the `projectRoot.resolve()` part you have used. No issue with this then.

IMS94 (approved):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionRouter.java
```diff
@@ -78,11 +81,15 @@
...
-
+        Project project = ctx.workspace().project(ctx.filePath()).orElseThrow();
+        Path projectRoot = (project.kind() == ProjectKind.SINGLE_FILE_PROJECT)
+                ? project.sourceRoot().getParent() :
+                project.sourceRoot();
         // Get available diagnostics based code-actions
         ctx.allDiagnostics().stream().
-                filter(diag -> CommonUtil.isWithinRange(ctx.cursorPosition(),
-                                                        CommonUtil.toRange(diag.location().lineRange())))
+                filter(diag -> projectRoot.resolve(diag.location().lineRange().filePath()).equals(ctx.filePath()) &&
```
nadeeshaan, line 90: What if, without doing the filtering at this point, we do the filtering within the context API (`allDiagnostics `)? This is because the code action is associated with a single file. Is there a specific reason where we will have to access all the diagnostics? Otherwise we are going through all the module diagnostics again and again

rasika: Hmm yeah true, there are other api usage for `ctx.allDiagnostics()`. let's have a separate api for a particular file diagnostics. Shall i send it in another PR?

nadeeshaan: +1

nadeeshaan (approved):

Issue #28289, opened 2021-01-29
Fix generating two different BIR files for cache and --bir-dump-file flag


warunalakshitha:
**Description:**
ATM there are changes in two birs generated from --bir-dump-file and one we cached. We should have one bir file.

PR #28725, opened 2021-02-17
Move BIR serialization to project api

warunalakshitha:
## Purpose
Previously we have serialized the bir into binary form during compilation (birgen) so there is no control from project api side. Then we store the serialized byte content in the BPackageSymbol. We already store the birBytes in ModuleContext as well. In this PR I have moved serialization to project api and make the getBirBytes function to have on-demand serialization. 

I also moved the bir writer to project api as well. 

Resolves https://github.com/ballerina-platform/ballerina-lang/issues/28289

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/projects/ModuleContext.java
```diff
@@ -133,6 +137,13 @@ ModuleName moduleName() {
         return moduleDescriptor.name();
     }
 
+    byte[] getBirBytes() {
+        if (birBytes == null) {
```
warunalakshitha, line 141: function to get bit bytes.

tests/ballerina-test-utils/src/main/java/org/ballerinalang/test/bala/BalaCreator.java
```diff
@@ -179,7 +179,7 @@ public static CompileResult compileWithTestsAndWrite(Path sourceRootPath, String
 
         BLangPackage bLangPackage = (BLangPackage) compileResult.getAST();
         Compiler compiler = Compiler.getInstance(context);
-        compiler.write(bLangPackage, targetPath);
+        //compiler.write(bLangPackage, targetPath);
```
warunalakshitha, line 182: This class already deprecated but still, some functions used. We need to use use {@link org.ballerinalang.test.BCompileUtil} instead.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/ModuleContext.java
```diff
@@ -133,6 +137,13 @@ ModuleName moduleName() {
...
+            birBytes = new BIRBinaryWriter(this.bLangPackage.symbol.bir).serialize();
```
manuranga, line 142: this.bLangPackage can be `null`, let's use `getBLangPackageOrThrow` so we get a better error message.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/Module.java
```diff
@@ -87,6 +87,10 @@ public ModuleDescriptor descriptor() {
         return new DocumentIterable(this.testSrcDocs.values());
     }
 
+    public byte[] birBytes() {
+        return this.moduleContext.getBirBytes();
```
manuranga, line 91: I don't see any other compiler outputs (such as getBLangPackageOrThrow) exposed from this class, Isn't it enough to expose this form the `ModuleContext`

warunalakshitha: I have removed exposing bir from Module. This was only used for test bir. Added api in nullbackend (which is use for test bir purpose) to get default module bir.

warunalakshitha:
There are conflicts. Will fix and send a new PR.

Issue #28292, opened 2021-01-29
[Ballerina Shell] Cyclic types cannot be used to define variables


kdsuneraavinash:
**Description:**
Because of current implementation of type traverse, cyclic types cause a `StackOverflowException` when they are attempted to converted to a string type signature. This causes the inability to use types that were defined in a cyclic manner.

If the type is `Y` and `Y` is defined as `Y[]|()`, when the type is found, it attempts to expand the union, but since that union cannot be expanded, or rather can be expanded infinitely, it fails. The type should not be expanded but rather kept as it is.

**Steps to reproduce:**
In REPL,
```bash
=$ type Y Y[]|()

=$ Y c = ()

ballerina: Oh no, something really went wrong. Bad. Sad.

There should be a file named "ballerina-internal.log" in the current directory.
If you are able to share with us the code that broke Ballerina then
we would REALLY appreciate if you would report this to us:
go to https://github.com/ballerina-platform/ballerina-lang/issues and
create a bug report with both this log file and the sample code.

We thank you for helping make us better dancers.
```

And the error log
```
[2021-01-29 12:16:02,116] SEVERE {b7a.log.crash} - null
java.lang.StackOverflowError
        at java.base/java.util.regex.Pattern$Branch.match(Pattern.java:4749)
        at java.base/java.util.regex.Pattern$BranchConn.match(Pattern.java:4713)
        at java.base/java.util.regex.Pattern$GroupTail.match(Pattern.java:4863)
        at java.base/java.util.regex.Pattern$BmpCharPropertyGreedy.match(Pattern.java:4344)
        at java.base/java.util.regex.Pattern$GroupHead.match(Pattern.java:4804)
        at java.base/java.util.regex.Pattern$Branch.match(Pattern.java:4749)
        at java.base/java.util.regex.Pattern$Branch.match(Pattern.java:4747)
        at java.base/java.util.regex.Pattern$BmpCharProperty.match(Pattern.java:3964)
        at java.base/java.util.regex.Pattern$Start.match(Pattern.java:3619)
        at java.base/java.util.regex.Matcher.search(Matcher.java:1729)
        at java.base/java.util.regex.Matcher.find(Matcher.java:773)
        at java.base/java.util.Formatter.parse(Formatter.java:2702)
        at java.base/java.util.Formatter.format(Formatter.java:2655)
        at java.base/java.util.Formatter.format(Formatter.java:2609)
        at java.base/java.lang.String.format(String.java:2897)
[...]

PR #28728, opened 2021-02-17
Fix xml, array and record type signatures to be syntactically valid

### LINKED ISSUE #28041: Type signature of arrays with a fixed length should reflect the length
**Description:**
Any array definition with a fixed-length should have the type signature that reflects the length. As a result there will be no difference of type signature between types of `int[]` or `int[3]`.

**Steps to reproduce:**
```ballerina
int[] a = [1, 2, 3];
int[3] b = [1, 2, 3];
```
Type symbols of `a` and `b` do not reflect that `b` has a fixed length.

### LINKED ISSUE #28434: ArrayTypeSymbol `size()` should return the array size if the associated type has a fixed size
**Description:**
`size()` method of `ArrayTypeSymbol` always returns `Optional.zero()` but should return the size of the array if the associated type has a fixed size.
https://github.com/ballerina-platform/ballerina-lang/blob/7b64f439850bdb674b7bb701b037fbcc25bb8d87/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaArrayTypeSymbol.java#L55-L58

**Steps to reproduce:**

**Affected Versions:**

#28042 #28041
 
### LINKED ISSUE #28693: XML Type Symbol should use xml<never> instead of xml<>
**Description:**
In,
```ballerina
xml<never> xmlValue = <xml<never>>'xml:concat();
```
the Type Symbol signature of `xmlValue` is `xml<>`. But this should instead be `xml<never>` since `xml<>` is not valid.

**Steps to reproduce:**

**Affected Versions:**
Ballerina SLP Alpha 2

`Area/SemanticAPI`

### LINKED ISSUE #28695: Record Type signature should not contain the rest type descriptor if inclusive
**Description:**
```ballerina
record {
        int maths;
        int physics;
        int chemistry;
    } g = {maths: 10, chemistry: 0, physics: 10};
```
Type Symbol signature of above is `record {int maths; int physics; int chemistry; anydata...;}` which is invalid. The rest type descriptors should not be included in the signature if the record is inclusive.

**Steps to reproduce:**

**Affected Versions:**
Ballerina SLP Alpha 2

`Area/SemanticAPI`

kdsuneraavinash:
## Purpose

Fix xml, array and record type symbol signatures.

Fixes #28041
Fixes #28695
Fixes #28693
Fixes #28434 

## Approach
- Array Type Symbol signature would have the size only if the size >= 0
- Record type will contain rest member only if exclusive
- `xml` will default to `xml<never>` if member doesn't have a name

## Remarks
This is a blocker for the fix of #28292 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaArrayTypeSymbol.java
```diff
@@ -49,11 +53,13 @@ public TypeSymbol memberTypeDescriptor() {
 
     @Override
     public String signature() {
-        return memberTypeDescriptor().signature() + "[]";
+        // If the array is of fixed size, the signature should reflect that.
+        String sizeStr = size().map(Objects::toString).orElse("");
+        return memberTypeDescriptor().signature() + "[" + sizeStr + "]";
```
nadeeshaan, line 58: @rasika does this particular change affects the code actions?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaXMLTypeSymbol.java
```diff
@@ -60,7 +60,9 @@ public BallerinaXMLTypeSymbol(CompilerContext context, ModuleID moduleID, BXMLTy
             if (xmlType == symbolTable.xmlType || this.typeParameter().isEmpty()) {
                 this.typeName = "xml";
             } else {
-                this.typeName = "xml<" + this.typeParameter().get().getName().orElse("") + ">";
+                // if type param name is missing, it should default to xml<never>
+                String typeParamString = this.typeParameter().get().getName().orElse("never");
```
hasithaa, line 64: @rdhananjaya @suleka96 Can you check this? I think this is not correct.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaArrayTypeSymbol.java
rasika: @nadeeshaan seems like these are valid by the spec https://ballerina.io/spec/lang/2020R1/#section_5.3.1.1. 

```ballerina
int[1] fooResult = foo();
function foo() returns int[1]{
    
}
```

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaXMLTypeSymbol.java
hasithaa: @pubudu91 Can you check this.

pubudu91: While the spec doesn't seem to mention this case, I also don't think this behaviour is implied. In the implementation, we have considered `xml` type as the union of `Element`, `Text`, `ProcessingInstruction` and `Comment`. And I think that's the implied behaviour of the spec as well.

kdsuneraavinash: The issue is with `xml<never>` being shown as `xml<>` from the signature, and `xml<>` being syntactically incorrect. #28693 
And I don't think assuming `xml<never>` as same as `xml` is correct either, since following would be possible then, no?

```ballerina
public function main(){
    var g = <xml<never>> xml``;
    g = xml `<a></a>`;
}
```

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordTypeSymbol.java
```diff
@@ -114,9 +114,12 @@ public String signature() {
             joiner.add(ballerinaFieldSignature);
         }
 
-        restTypeDescriptor().ifPresent(typeDescriptor -> {
-            joiner.add(typeDescriptor.signature() + "...;");
-        });
+        // No rest descriptor if the record in inclusive
+        if (!this.isInclusive) {
```
pubudu91, line 118: Do you need this inclusivity info? The issue with using this is this check is not accurate. We don't keep this info in the internal symbol, so it's not available here either.

kdsuneraavinash: The issue is because of #28695 
And I'm not sure why it cannot be used, the same check is used in L107 few lines above this.
https://github.com/ballerina-platform/ballerina-lang/blob/b235daf9bc6576b5a015ca256751efbec98023bb/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordTypeSymbol.java#L107-L109

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaXMLTypeSymbol.java
pubudu91: Shall we change this like the following? 
- Move the logic in `getName()` to `signature()`. In `getName()`, let's just return `xml` for now
- Then, call `String typeParamString = this.typeParameter().get().signature();`

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordTypeSymbol.java
pubudu91: The issue is this check doesn't always return the correct result. For example, this would incorrectly evaluate to true for the following:
```ballerina
type Person record {|
    string name;
    int age;
    string...;
|};
```

pubudu91: This is something we need to fix in the compiler to get this check to correctly work. At the moment, this syntactic difference between inclusive and exclusive record typedescs is not captured by internal type info since it has no bearing on the compiler analysis phases.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaXMLTypeSymbol.java
kdsuneraavinash: Returning `xml` in `getName()` caused some other issues, so I refactored that code to the `signature` and did this change there

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordTypeSymbol.java
kdsuneraavinash: I changed the signature to treat the symbol as inclusive iif `this.isInclusive` is true and the rest type (if there is any)  is `anydata`.

pubudu91: That still wouldn't be accurate right? Suppose the user defines a record as follows: 
```ballerina
type Person record {|
    string name;
    int age;
    anydata...;
|};
```
Now `Person` would be incorrectly identified as an inclusive record typedesc.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordTypeSymbol.java
```diff
@@ -103,8 +103,21 @@ public BallerinaRecordTypeSymbol(CompilerContext context, ModuleID moduleID, BRe
...
+        // otherwise it is actually exclusive.
+        // This is a counter for following being identified as inclusive.
+        // type Person record {|
+        //    string name;
+        //    int age;
+        //    string...;
+        //|};
+        boolean isRestAnyData = restTypeDescriptor()
+                .map(t -> TypeDescKind.ANYDATA.equals(t.typeKind()))
+                .orElse(true);
```
pubudu91, line 116: Missing rest type desc does not imply that the rest type is `anydata`. A closed record doesn't have a rest type desc.

kdsuneraavinash: Yes, but this is actually representing `is rest anydata if there is a rest type descriptor`. So if there isn't a rest type descriptor, the value should be true. Maybe I have to rename it?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordTypeSymbol.java
kdsuneraavinash: Isn't,
```ballerina
type Person record {|
    string name;
    int age;
    anydata...;
|};
```

same as 
```ballerina
type Person record {
    string name;
    int age;
};
```

so cant we treat closed records with anydata field as the same as open records.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordTypeSymbol.java
```diff
@@ -103,12 +101,8 @@ public BallerinaRecordTypeSymbol(CompilerContext context, ModuleID moduleID, BRe
 
     @Override
     public String signature() {
-        StringJoiner joiner;
-        if (this.isInclusive) {
-            joiner = new StringJoiner(" ", "{ ", " }");
-        } else {
-            joiner = new StringJoiner(" ", "{| ", " |}");
-        }
+        // Treating every record as if they are closed records.
```
pubudu91, line 104:
```suggestion
        // Treating every record typedesc as exclusive record typedescs.
```

pubudu91 (approved):

Issue #28390, opened 2021-02-02
Module-level query expressions don't work


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
Example 1
```ballerina
type Employee record {|
    readonly int id;
    string name;
|};

table<Employee> key(id) employees = table [
    {name: "Amy", id: 1234},
    {name: "John", id: 4567}
];

int[] salaries = from record { int id; } rec in employees
                    select rec.id;
```

```cmd
$ bal run xyz.bal 

Compiling source
        xyz.bal
ERROR [xyz.bal:(12:28,12:31)] undefined symbol 'rec'
```

Example 2
```ballerina
type Employee record {|
    readonly int id;
    string name;
|};

type Person record {|
    string name;
    string dob;
|};

table<Employee> key(id) tb = table [
    {name: "Amy", id: 1234},
    {name: "John", id: 4567}
];

Person[] arr = [
    {dob: "01-01-1990", name: "John"},
    {dob: "07-07-1985", name: "Kate"}
];

string[] dob = from var person in arr
                    join var employee in tb 
                    on person.name equals employee.name
                    select person.name + ":" + person.dob;
```

```cmd
$ bal run xyz.bal 

Compiling source
        xyz.bal
ERROR [xyz.bal:(23:24,23:30)] undefined symbol 'person'
ERROR [xyz.bal:(23:43,23:51)] undefined symbol 'employee'
ERROR [xyz.bal:(24:28,24:34)] undefined symbol 'person'
ERROR [xyz.bal:(24:48,24:54)] undefined symbol 'person'
```

**Affected Versions:**
slalpha1

**Related Issues (optional):**
#28389

Issue #28389, opened 2021-02-02
Compiler crashes for module-level query expression which uses var


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type Employee record {|
    readonly int id;
    string name;
|};

table<Employee> key(id) employees = table [
    {name: "Amy", id: 1234},
    {name: "John", id: 4567}
];

int[] salaries = from var {id} in employees
                    select id;
```

```log
[2021-02-02 18:07:47,655] SEVERE {b7a.log.crash} - null 
java.lang.NullPointerException
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.visit(SemanticAnalyzer.java:771)
	at org.wso2.ballerinalang.compiler.tree.BLangSimpleVariable.accept(BLangSimpleVariable.java:53)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.analyzeNode(SemanticAnalyzer.java:3585)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.analyzeNode(SemanticAnalyzer.java:3553)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.validateRecordVariable(SemanticAnalyzer.java:1530)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer.handleDeclaredVarInForeach(SemanticAnalyzer.java:1192)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.handleInputClauseVariables(TypeChecker.java:4702)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.visit(TypeChecker.java:4591)
	at org.wso2.ballerinalang.compiler.tree.clauses.BLangFromClause.accept(BLangFromClause.java:37)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.lambda$visit$26(TypeChecker.java:4415)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.visit(TypeChecker.java:4415)
	at org.wso2.ballerinalang.compiler.tree.expressions.BLangQueryExpr.accept(BLangQueryExpr.java:99)
[...]

Issue #28037, opened 2021-01-21
[Ballerina Shell] Enum definitions with explicit expressions incorrect behavior


kdsuneraavinash:
**Description:**
Enum declarations with explicit expressions are do not work correctly.

**Steps to reproduce:**
In REPL,
```ballerina
// This will not work due to type EN converting to english
=$ enum Language {EN="english", TA="tamil", SI="sinhala"}
=$ EN variable = "english"  // <- Error
```

DamithaSenevirathne:
Can not reproduce this issue with latest ballerina master.
![issue-28037](https://user-images.githubusercontent.com/31839712/124077575-38893400-da65-11eb-81c7-8c9ea9191cef.png)

DamithaSenevirathne:
@kdsuneraavinash can you please describe more about the error state. 
Got above results in the ballerina shell with the latest ballerina master.

kdsuneraavinash:
It previously went to an error state. 
![Screenshot_20210705_134202](https://user-images.githubusercontent.com/34407567/124439456-1100d700-dd97-11eb-9a0a-0991637d7c12.png)

If it is not going to an error state like this, I think this issue might have been fixed by some change done in master.

DamithaSenevirathne:
Managed to reproduce the error state. will work on a fix
![Screenshot from 2021-07-06 00-34-00](https://user-images.githubusercontent.com/31839712/124510755-27884c00-ddf2-11eb-9cf6-0e4f62e4cf6d.png)

nipunayf:
Cannot seem to reproduce this error with 2201.7.2.

<img width="412" alt="image" src="https://github.com/ballerina-platform/ballerina-lang/assets/59343084/a10c2519-44f0-4c07-a674-2ead743110e2">

Issue #27855, opened 2021-01-13
Allow public qualifier for module-level variables


pubudu91:
With https://github.com/ballerina-platform/ballerina-spec/issues/324 `public` qualifier is now allowed on module-level vars. Need to update the compiler to reflect this.

hasithaa:
Related to https://github.com/ballerina-platform/ballerina-lang/issues/27982

hasithaa:
@rdulmina can you add https://github.com/ballerina-platform/ballerina-lang/issues/27855 as well.

rdulmina:
> @rdulmina can you add [ballerina-platform/ballerina-spec#741](https://github.com/ballerina-platform/ballerina-spec/issues/741) as well.

Sure

PR #28746, opened 2021-02-18
Refactor Ballerina Shell core to support redefining module-level declarations and variable declarations

### LINKED ISSUE #28042: [Ballerina Shell] Unexpected behavior with array binding patterns
**Description:**
Some array binding patterns defined using a fixed length array are not possible in REPL.

**Steps to reproduce:**
In REPL,
```ballerina
=$ int[3] variable = [1, 2, 3]
// This will cause an error
=$ [int, int, int] [a, b, c] = variable
```

#28041

### LINKED ISSUE #28272: Ballerina shell should allow redefining module level functions and variables
In the shell, this gives an error about redeclaring `add`:

```
function add(int x, int y) returns int => x * y;
add(1, 1)
function add(int x, int y) returns int => x = y;
```

But in a REPL, you often want to develop code interactively, which involves redefining functions that are wrong.

A good start would be to allow redefining the last thing you defined.

### LINKED ISSUE #28276: Improve Ballerina shell reporting of runtime errors
If I type:

```
error("whoops");
```

the shell says:

```
Exception occurred: error("whoops")
| Unhandled Runtime Error.
```

This is not an accurate description of what is happening from a Ballerina perspective. An expression has been evaluated and the result is an error. Ballerina  does not have exceptions. You should display the error as a value, e.g.

```
error("whoops")
```

If I do:

```
fail error("whoops");
```

I get the same message. In this case you should use the error value to report an error, e.g.

```
Error: whoops
```

or

```
Fail: whoops
```

If I do 

```
panic error("whoops");
```

it should say

```
Panic: whoops
```

### LINKED ISSUE #28283: Ballerina shell should allow `public`
Ballerina shell gives me an error for e.g.

```
public int x = 0;
```

I am guessing the thinking was that `public` is pointless, because there is no way for another module to reference symbols. But often people will want to use the shell to try out snippets of code that will be incorporated into programs, and these snippets may well have `public `in them.

### LINKED ISSUE #28292: [Ballerina Shell] Cyclic types cannot be used to define variables

### LINKED ISSUE #28298: `final` variables can be re-assinged in Ballerina shell
The following should give a compilation error 
```
=$ final int k = 10; 
=$ function foo() {k = 11;}
```

### LINKED ISSUE #28316: Ballerina Shell should preserve qualifiers in a variable declaration
**Description:**
Currently, when declaring a variable, all the qualifiers are removed. But instead, it should preserve the qualifiers to be used. Otherwise, it will create some unexpected behavior. (eg: final can be reassigned)
Also, some qualifiers may be required to ignore. (eg: public) But these should not cause any errors in shell. 

**Steps to reproduce:**
Following should not be possible,
```ballerina
final int k = 10
k = 100
```

**Affected Versions:**
Ballerina Swan Lake Alpha 1

#28298 #28283 

### LINKED ISSUE #28413: Ballerina shell does not allow isolated function to access final variable with readonly type
If I paste this into the shell:

```
type Entry map<json>;
type RoMap readonly & map<Entry>;

function loadMap() returns RoMap {
   //...
  return {};
}

final RoMap m = loadMap();

isolated function lookup(string s) returns readonly & Entry? {
   // OK because m is final and type of m is subtype of readonly
   return m[s];
}
```

I get the error:

```
| invalid access of mutable storage in an 'isolated' function
|    return m[s];
|           ^
| Compilation aborted because of errors.
```

The error is incorrect, and I do not get it outside the shell.

kdsuneraavinash:
## Purpose

Currently, REPL does not support redefining module-level declarations and variable declarations. However, since REPL is meant as an interactive feature, it should support this. This PR will fix issues related to redefining a module-level declaration/variable declaration.
Fixes #28272

The PR also restructures Ballerina Shell code base to support multiple core changes related to declaration definition. These changes will fix several issues in Ballerina Shell related to its template and declaration evaluation.

**Uses signature instead of `TypeSignatureTransformer` to find the type**
Change the type signature parsing logic to use the already implemented `signature()` call of `TypeSymbol`s. Previously the logic was re-implemented to traverse and parse the type.
#28728 should be merged before this PR to fix signature issues found.
Fixes #28292
Fixes #28042

**Updates templates to provide a better wrapper**
Previous ballerina shell template implementation ignores qualifiers in variable declarations and rejects public keywords. Also, previously runtime errors were reported as if they exceptions. This PR will also update templates to support better runtime reporting as well as qualifiers.
Fixes #28283
Addresses #28413 (Would not work if all submitted at once unless declarations are separated with semicolons)
Fixes #28316
Fixes #28298
Fixes #28276

## Approach
- Remove type signature converting logic
- Use `signature()` call directly
- Remove logic to reject public keyword
- Keep qualifiers in the `VariableDeclarationSnippet`
- Error-values will simply be evaluated
- Fails will output, `fail: ERROR` instead of the previous message
- Panics and other runtime errors will output `panic: ERROR` instead of the previous message
- The red `Unhandled Runtime Error`. will only be displayed for panics
- Variable values will be preserved even after runtime errors
- Allow rewriting old module declarations
- Allow rewriting old variable declarations
- Change import prefix finding to support this change
- Refactor shell invoker code to assist in adding this change
- Add a new method to directly run the file in REPL (command `/open <FILE>`)
- Add shell CLI parameter to run the file (`-o <FILE_NAME>`)
- Remove previous invoker memory class util functions and use file running the command in tests
- Supports `/remove` to delete previous declarations
- New var declarations are placed in module level to make sure re-declaration errors are correct

## Samples

**Public is allowed and qualifiers are preserved**
```ballerina
=$ public function f() returns int {  
 > return 1;
 > }

=$ f()  
1

=$ /dclns 
| ('f) public function f() returns int {return 1;}

=$ final int y = 10  

=$ y = 100   
| error: cannot assign a value to final 'y'
| 	y = 100;
| 	^
| Compilation aborted due to errors.

=$     
```

**Runtime errors are shown appropriately**
```ballerina
=$ 1/0   
panic: {ballerina}DivisionByZero {"message":" / by zero"}
| Execution aborted due to unhandled runtime error.

=$ error e = <error> int:fromString("Hello") 

=$ e  
error("{ballerina/lang.int}NumberParsingError",message="'string' value 'Hello' cannot be converted to 'int'")

=$ fail e  
fail: {ballerina/lang.int}NumberParsingError {"message":"'string' value 'Hello' cannot be converted to 'int'"}

=$  
```

**Allows redefining variables**
```ballerina
=$ string y = "hello" 

=$ y     
"hello"

=$ int y = 100   

=$ y   
100

=$ /remove y    

=$ y 
| error: undefined symbol 'y'
| 	y
| 	^
| Compilation aborted due to errors.

=$   
```

**Can load declarations from files (and tab-completes files)**
file

## Remarks
- should be merged after #28728 
- public variable declarations will still cause an error because of #27855
- enums with explicit expressions can pollute the state (#28037)
- global variables are now placed at the global level in the wrapper, so there might be additional issues relating to constraints in module level initializers (eg: #28389 #28390)
- Refactored elevated type transformer to use `assignableTo()` directly
- Last expression value was changed to a global variable named `__last__`. It will hold the last expression evaluation result.
- All execution methods now have a preceding `__`.
- Imports will now compile in online mode if initial offline import resolution failed

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


cli/ballerina-cli/src/main/resources/cli-help/ballerina-shell.help
```diff
@@ -30,3 +32,6 @@ EXAMPLES
 
        Run ballerina shell with 1s time out.
           $ bal shell -t 1000
+
+       Run ballerina shell initial declarations from "main.bal".
```
MaryamZi, line 36:
```suggestion
       Run ballerina shell loading initial declarations from "main.bal".
```

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/ShellCommand.java
```diff
@@ -47,6 +48,9 @@
     @CommandLine.Option(names = {"-t", "--time-out"}, description = "Timeout to use for tree parsing.")
     private long timeOut = 100;
 
+    @CommandLine.Option(names = {"-o", "--open-file"}, description = "Open file and load initial declarations.")
```
MaryamZi, line 51: Shouldn't  the option simply be something like `--file`? IMO, --open-file kinda implies a boolean option.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/PropertiesLoader.java
```diff
@@ -46,6 +46,8 @@
     public static final String COMMAND_VARS = "commands.vars";
     public static final String COMMAND_IMPORTS = "commands.imports";
     public static final String COMMAND_DCLNS = "commands.dclns";
```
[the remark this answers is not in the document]
kdsuneraavinash: this refers to the inbuilt command /dcln 
https://github.com/ballerina-platform/ballerina-lang/blob/36329a92866904fe4536ec477bf81dd65e99ef9b/ballerina-shell/modules/shell-cli/src/main/resources/shell.properties#L32

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/ShellCommand.java
kdsuneraavinash: Sure, but the command argument will have to be `-o` still because `-f` was taken for `--force-dumb`
So maybe this should be named to a different name?

MaryamZi: IMO we don't need a short option for every single option. Can we use `-f` for --file and remove it from --force-dumb? If not maybe we can remove the short option here? WDYT?

kdsuneraavinash: I think removing the short option from `--force-dumb` might be better, there not much use case for that option

MaryamZi: +1, let's also mention this in the release note.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/parser/SerialTreeParser.java
```diff
@@ -73,8 +76,50 @@ public Node parse(String source) throws TreeParserException {
...
+        if (declarationNode instanceof FunctionDefinitionNode) {
+            String functionName = ((FunctionDefinitionNode) declarationNode).functionName().text();
+            if (functionName.equals("main")) {
+                addWarnDiagnostic("Found 'main' function in the declarations.\n" +
+                        "Discarded 'main' function without loading.");
+                return false;
+            }
+            if (functionName.equals("init")) {
+                addWarnDiagnostic("Found 'init' function in the declarations.\n" +
+                        "Discarded 'init' function without loading.");
```
MaryamZi, line 114: Can we get the file name here? If so shall we say

```
Ignoring the 'init' function in the declarations loaded from 'file_name.bal'.
```

MaryamZi, line 114: Can we combine these two checks and build the message using the functionName variable?

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BShellConfiguration.java
```diff
@@ -159,14 +168,26 @@ public Builder setDumb(boolean dumb) {
             return this;
         }
 
+        /**
+         * Start file is the file containing declarations that should
+         * be loaded into the shell initially.
+         * If the file loading fails, the shell will not start and fail.
+         * The file should not contain any definitions for built-in names.
+         * (eg: main, run, init)
```
MaryamZi, line 176: Why do we need this restriction? Is it due to how we generate code?

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/EvaluatorImpl.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Override
+    public void evaluateDeclarationFile(String filePath) throws BallerinaShellException {
+        try {
+            String statements = Files.readString(Paths.get(filePath), Charset.defaultCharset());
+            Collection<Node> nodes = treeParser.parseDeclarations(statements);
+            Collection<Snippet> snippets = snippetFactory.createSnippets(nodes);
+            invoker.execute(snippets);
+        } catch (IOException e) {
+            addErrorDiagnostic("File open failed: " + filePath);
```
MaryamZi, line 88: How about
```suggestion
            addErrorDiagnostic("Error loading from file '" + filePath + "'");
```

Can we add some information from the exception as to what exactly went wrong?

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/handlers/FileCommand.java
```diff
@@ -16,39 +16,30 @@
...
+        super(ballerinaShell);
     }
 
     @Override
-    public String toString() {
-        return repr;
+    public void run(String... args) {
+        if (args.length == 0) {
+            ballerinaShell.outputError("No arguments provided. " +
+                    "You have to provide the file name to read.");
```
MaryamZi, line 37:
```suggestion
                    "You have to provide the name of the file to read.");
```

or
```suggestion
                    "You have to provide the name of the file to load.");
```

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/handlers/FileCommand.java
```diff
@@ -16,39 +16,30 @@
...
+        } else if (args.length == 1) {
+            ballerinaShell.runFile(args[0]);
+        } else {
+            ballerinaShell.outputError("Multiple arguments provided. " +
+                    "You have to provide the file name to read as the only argument.");
```
MaryamZi, line 42:
```suggestion
                    "You have to provide the name of the file as the only argument.");
```

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
```diff
@@ -7,6 +7,8 @@ Or type one of the following commands:
 /debug          - Toggle debug mode. You will see debug messages.
 /reset          - Reset REPL state.
                   All previous code snippets will be removed from REPL memory.
+/open <FILE>    - Load declarations defined in <FILE>. eg: /file my_file.bal
+/remove <NAMES> - Delete declarations specified by <NAMES>. eg: /file function1 var1 var2
```
MaryamZi, line 11:
```suggestion
/open <FILE>    - Load declarations defined in <FILE>. eg: /open my_file.bal
/remove <NAMES> - Delete declarations specified by <NAMES>. eg: /remove function1 var1 var2
```

ballerina-shell/modules/shell-cli/src/main/resources/shell.properties
```diff
@@ -30,6 +30,8 @@ commands.debug=/debug
 commands.vars=/vars
 commands.imports=/imports
 commands.dclns=/dclns
+commands.file=/open
```
MaryamZi, line 33: Can we use 
```suggestion
commands.file=/file
```

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
```diff
@@ -7,6 +7,8 @@ Or type one of the following commands:
 /debug          - Toggle debug mode. You will see debug messages.
 /reset          - Reset REPL state.
                   All previous code snippets will be removed from REPL memory.
+/open <FILE>    - Load declarations defined in <FILE>. eg: /file my_file.bal
```
MaryamZi, line 10: Also wondering if this should be file or load instead of open.

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/parser/SerialTreeParser.java
kdsuneraavinash: Actually this function does not know about a file. It just receives the delcarations inside the file, already parsed, not the file itself.

ballerina-shell/modules/shell-cli/src/main/java/io/ballerina/shell/cli/BShellConfiguration.java
kdsuneraavinash: yeah, they are already kinda defined in the shell, so user should not be able to redefine them

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
pubudu91, line 11: eg: => e.g.,

pubudu91: How's the exact usage of these? `/open /file my_file.bal`?

kdsuneraavinash: It is a typo, the example was supposed to say `/open my_file.bal`, so use the command as `/open my_file.bal` 
But the command is `/file my_file.bal` if using the last commit because I just changed that command to /file instead of /open as per @MaryamZi 's suggestion :)

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/ImportsManager.java
```diff
@@ -0,0 +1,314 @@
...
+     * We need to traverse all the sub-typed because any sub-type
+     * may need to be imported.
+     * Also, this will compute all the replacements that should be done.
+     * In above case, abc/z:TypeA will be replaced with z:TypeA.
+     *
+     * @param typeSymbol Type to process.
+     * @param imports    Imports set to store imports.
+     * @return Type signature after parsing.
+     */
+    protected String extractImportsFromType(TypeSymbol typeSymbol, Set<QuotedIdentifier> imports) {
```
pubudu91, line 234: IIUC, the purpose of this method is to figure out which module to import from the given type symbol right? Can't the module symbol returned from the `getModule()` method be used for that?

ballerina-shell/modules/shell-cli/src/main/resources/command.help.txt
```diff
@@ -7,6 +7,8 @@ Or type one of the following commands:
 /debug          - Toggle debug mode. You will see debug messages.
 /reset          - Reset REPL state.
                   All previous code snippets will be removed from REPL memory.
+/file <FILE>    - Load declarations defined in <FILE>. e.g.: /file my_file.bal
+/remove <NAMES> - Delete declarations specified by <NAMES>. e.g.: /remove function1 var1 var2
```
pubudu91, line 11: Missing comma. Should be: `e.g.,`

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/ImportsManager.java
kdsuneraavinash: The purpose is to figure out the required imports to use a type. so the type is something like (xyz/pqr:C | xyz/abc:D) and from this we have to extract imports (eg: import xyz/pqr, import xyz/abc) and replace signature as (pqr:C | abc:D)

This is the same approach taken in lang server iirc

ballerina-shell/modules/shell-cli/src/main/resources/shell.properties
```diff
@@ -30,6 +30,8 @@ commands.debug=/debug
...
+commands.delete=/remove
```
MaryamZi, line 34: Shouldn't we ideally use `/file` and `/delete` here?

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/ImportsManager.java
pubudu91: Still for a union type symbol, it'd be easier to derive the import statements using the module symbols of the member types of the union though.

MaryamZi (approved):
A panic statement still seems to print "Execution aborted due to unhandled runtime error. " too.

```cmd
=$ panic error("error!")                                      
panic: error! {}
| Execution aborted due to unhandled runtime error. 
```

I don't think we need this. Shall we check and remove this in a separate PR?

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/ImportsManager.java
kdsuneraavinash: I'll create an issue for this to track this. #29109

PR #28794, opened 2021-02-21
Fix inconsistency in typed binding patterns with var for lists

### LINKED ISSUE #28326: Inconsistency in typed binding patterns with var for lists
**Description:**
$title.

**Steps to reproduce:**
Seems like it is only allowed for an exact match 

```ballerina
public function main() {
    var [a, ...b] = tuple();

    var [c, ...d] = array();
}

public function tuple() returns [int, int, string...] {
    return [1, 2, "hello", "world"];
}

public function array() returns int[4] {
    return [1, 2, 3, 4];
}
```

These fail with
```cmd
$ bal run xyz.bal 

Compiling source
        xyz.bal
ERROR [xyz.bal:(2:21,2:28)] incompatible types: expected '[other]', found '[int,int,string...]'
ERROR [xyz.bal:(2:21,2:28)] invalid tuple binding pattern: attempted to infer a tuple type, but found 'other'
ERROR [xyz.bal:(4:21,4:28)] incompatible types: expected '[other]', found 'int[4]'
ERROR [xyz.bal:(4:21,4:28)] invalid tuple binding pattern: attempted to infer a tuple type, but found 'other'
```

After fixing https://github.com/ballerina-platform/ballerina-lang/issues/24764 the errors are
```
$ bal run xyz.bal 

Compiling source
        xyz.bal
ERROR [xyz.bal:(2:21,2:28)] incompatible types: expected '[other,other...]', found '[int,int,string...]'
ERROR [xyz.bal:(2:21,2:28)] invalid tuple binding pattern: attempted to infer a tuple type, but found 'other'
ERROR [xyz.bal:(4:21,4:28)] incompatible types: expected '[other,other...]', found 'int[4]'
ERROR [xyz.bal:(4:21,4:28)] invalid tuple binding pattern: attempted to infer a tuple type, but found 'other'
```

But this is not the case for records. The following work

```ballerina
public function main() {
    var {i, j, ...x} = getRecOne();

    var {k, ...m} = getRecTwo();
}

public function getRecOne() returns record {int i;} {
    return {i: 1, "j": "str"};
}


public function getRecTwo() returns record {|int k; string l; int...;|} {
    return {k: 1, l: "str", "m": 123};
}
```

The spec says
> When the binding pattern is more complicated, the binding pattern must be consistent with the type-descriptor, so that the type-descriptor unambiguously determines a type for each variable occurring in the binding pattern.

IMO, this is possible with the list examples above.

**Affected Versions:**
slalpha1

lasinicl:
Fixes #28326

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1283,6 +1301,15 @@ private boolean isArrayTypeAssignableToTupleType(BArrayType source, BTupleType t
                 return false;
             }
 
+            if (target.tupleTypes.stream().allMatch(this::isNoType)) {
```
gimantha, line 1304: Seems we are iterating two times here.. first we iterate at line number 1264, then here again, what if we refactor isNoType() to deal with BArrayType also and return isNoType and done with a single iteration?

lasinicl: Fixed in 10dc20b

gimantha (approved):

tests/jballerina-unit-test/src/test/resources/test-src/types/tuples/tuple_negative_test.bal
```diff
@@ -167,3 +167,17 @@ function testInvalidConstIndex() {
 
 type NoFillerObject object {
 };
+
+function testInvalidTupleDeclaredWithVar() {
+    var [a1, a2, a3, ...a4] = getData();
+    var [b1,b2,b3] = getData2();
```
MaryamZi, line 173:
```suggestion
    var [b1, b2, b3] = getData2();
```

tests/jballerina-unit-test/src/test/resources/test-src/types/tuples/tuple_basic_test.bal
```diff
@@ -249,6 +249,174 @@ function testAnonRecordsInTupleTypeDescriptor() {
...
+    public boolean b;
+    public int i;
+    public function init(boolean b, int i) {
+        self.b = b;
+        self.i = i;
+    }
+}
+
+Foo foo = {name:"Test", age:23};
+Bar bar = {id:34, flag:true};
```
MaryamZi, line 411:
```suggestion
Foo foo = {name: "Test", age: 23};
Bar bar = {id: 34, flag: true};
```

Let's fix wherever applicable.

tests/jballerina-unit-test/src/test/resources/test-src/statements/vardeclr/module_tuple_var_decl.bal
```diff
@@ -79,6 +79,25 @@ function testDeclaredWithVar2() {
...
 
+var [[intVal], {a: intVal2}, ...otherValues] = getComplexTuple();
+
+function testDeclaredWithVar3() {
+    assertEquality(5, intVal);
+    assertEquality(6, intVal2);
+    int|error err = otherValues[0];
+    if (err is error) {
+        assertEquality("error msg", err.message());
+    }
```
MaryamZi, line 90: We need to handle the else scenario and fail here, right? Please check wherever applicable.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/tuples/BasicTupleTest.java
```diff
@@ -243,6 +243,27 @@ public void testNegativesOfTupleType() {
         BAssertUtil.validateError(resultNegative, i, "list index out of range: index: '-1'", 165, 19);
     }
 
+    @Test(description = "Test tuple declared with var")
+    public void testTupleDeclaredWithVar() {
+        BRunUtil.invoke(result, "testTupleDeclaredWithVar1");
+        BRunUtil.invoke(result, "testTupleDeclaredWithVar2");
+        BRunUtil.invoke(result, "testTupleDeclaredWithVar3");
+        BRunUtil.invoke(result, "testTupleDeclaredWithVar4");
```
MaryamZi, line 251: Better to use a data provider so that the rest of the tests run even if one fails.

grainier (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2008,28 +2008,63 @@ boolean checkTypeAndVarCountConsistency(BLangTupleVariable varNode, BTupleType t
...
             memberVariables.add(varNode.restVariable);
+            if (varNodeMemCount < tupleNodeMemCount) {
+                LinkedHashSet<BType> varTypes = new LinkedHashSet<>();
+                for (int i = varNodeMemCount; i < tupleNodeMemCount; i++) {
+                    varTypes.add(tupleTypeNode.tupleTypes.get(i));
+                }
+                if (restType != null) {
+                    varTypes.add(restType);
+                }
+                restType = BUnionType.create(null, varTypes);
```
MaryamZi, line 2051: Shall we check the size of `varTypes` here before creating the union and if the size is only 1 return that member? We shouldn't create unions if there isn't at least 2 types.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2008,28 +2008,63 @@ boolean checkTypeAndVarCountConsistency(BLangTupleVariable varNode, BTupleType t
...
+            }
         }
-        for (int i = 0; i < memberVariables.size(); i++) {
+        int memVarCount = memberVariables.size();
+        for (int i = 0; i < memVarCount; i++) {
             BLangVariable var = memberVariables.get(i);
-            BType type = (i <= tupleTypeNode.tupleTypes.size() - 1) ? tupleTypeNode.tupleTypes.get(i) :
-                    new BArrayType(tupleTypeNode.restType);
+            BType type;
+            if (varNode.restVariable != null && i == memVarCount - 1) {
```
MaryamZi, line 2058: We don't have to repeat the `varNode.restVariable != null` check for each iteration, right? Shall we extract this out to a variable outside the loop?

MaryamZi: Wouldn't it be better to not add the restVariable to member variables and handle it separately? We are unnecessarily repeating this check for each member variable, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -1474,11 +1474,14 @@ private void createRestFieldVarDefStmts(BLangTupleVariable parentTupleVariable,
             //     t[t.length()] = <T> tupleLiteral[$foreach$i];
             // }
             BLangExpression tupleExpr = parentTupleVariable.expr;
+            // Make sure the tupleExpr rest type can hold parent tuple's rest types
+            if (tupleExpr.type.tag == TypeTags.TUPLE) {
+                ((BTupleType) tupleExpr.type).restType = ((BArrayType) parentTupleVariable.restVariable.type).eType;
```
MaryamZi, line 1479: We shouldn't be changing anything in the original tuple type, right? The following now panics at runtime.

```ballerina
public function main() {
    var [a, ...b] = tuple();
    [int, int, string...] m = <[int, int, string...]> tuple();
}

public function tuple() returns [int, int, string...] {
    return [1, 2, "hello", "world"];
}
```

```cmd
$ bal run test.bal 

Compiling source
        test.bal

Running executable

error: {ballerina}TypeCastError {"message":"incompatible types: '[int,int,int|string...]' cannot be cast to '[int,int,string...]'"}
        at test:main(test.bal:3)
```

lasinicl:
Suggestions have been addressed via 7d86f39

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2008,28 +2008,63 @@ boolean checkTypeAndVarCountConsistency(BLangTupleVariable varNode, BTupleType t
                 case TypeTags.TUPLE:
                     tupleTypeNode = (BTupleType) varNode.type;
                     break;
+                case TypeTags.ARRAY:
+                    List<BType> tupleTypes = new ArrayList<>();
+                    BArrayType arrayType = (BArrayType) varNode.type;
+                    for (int i = 0; i < arrayType.size; i++) {
+                        tupleTypes.add(arrayType.eType);
+                    }
+                    tupleTypeNode = new BTupleType(tupleTypes);
```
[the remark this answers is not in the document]
lasinicl: Checked the feasibility of removing this. According to my understanding, I think we'll have to add many checks to places where we have used the created tuple if we are not creating a new tuple here since all other cases also create a tuple and the rest of the logic is written catering to that. Shall we create an issue for this and revisit it if possible?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2053,6 +2083,19 @@ boolean checkTypeAndVarCountConsistency(BLangTupleVariable varNode, BTupleType t
...
+        if (varNode.isDeclaredWithVar) {
+            return (varNode.restVariable == null ||
+                    varNode.memberVariables.size() <= tupleTypeNode.tupleTypes.size()) &&
+                    (varNode.restVariable != null || (tupleTypeNode.tupleTypes.size() == varNode.memberVariables.size()
+                            && tupleTypeNode.restType == null));
+        } else {
+            return tupleTypeNode.tupleTypes.size() == varNode.memberVariables.size() &&
+                    (tupleTypeNode.restType != null || varNode.restVariable == null) &&
+                    (tupleTypeNode.restType == null || varNode.restVariable != null);
+        }
```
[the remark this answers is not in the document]
MaryamZi: This is an existing issue though, so let's maybe create an issue and fix it separately.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1096,6 +1096,18 @@ private boolean isTupleTypeAssignable(BType source, BType target, Set<TypePair>
         return true;
     }
 
+    private boolean checkAllTupleTypeMembersBelongNoType(BTupleType tupleType) {
+        for (BType memberType : tupleType.tupleTypes) {
+            switch (memberType.tag) {
+                case TypeTags.NONE:
+                    return true;
```
MaryamZi, line 1103: The method name says check**All** but we seem to return on the first match of noType, is this correct?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1096,6 +1096,18 @@ private boolean isTupleTypeAssignable(BType source, BType target, Set<TypePair>
...
+                case TypeTags.TUPLE:
+                    return checkAllTupleTypeMembersBelongNoType((BTupleType) memberType);
+            }
```
MaryamZi, line 1106: Don't we have to have a default case that returns false for anything else?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2460,6 +2471,12 @@ private boolean hasSameReadonlyFlag(BType source, BType target) {
         }
 
         public Boolean visit(BTupleType t, BType s) {
+            if (((!t.tupleTypes.isEmpty() && checkAllTupleTypeMembersBelongNoType(t)) ||
```
MaryamZi, line 2474: We can probably directly pass `t.tupleTypes` to `checkAllTupleTypeMembersBelongNoType`, right?

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
```diff
@@ -318,7 +318,7 @@
     INVALID_TUPLE_BINDING_PATTERN("BCE2588", "invalid.tuple.binding.pattern"),
     INVALID_TYPE_FOR_TUPLE_VAR_EXPRESSION("BCE2589", "invalid.type.for.tuple.var.expr"),
     INVALID_TUPLE_BINDING_PATTERN_DECL("BCE2590", "invalid.tuple.binding.pattern.decl"),
```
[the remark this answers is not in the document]
lasinicl: Created an issue. https://github.com/ballerina-platform/ballerina-lang/issues/28836

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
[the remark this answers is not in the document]
lasinicl: Created an issue. https://github.com/ballerina-platform/ballerina-lang/issues/28838

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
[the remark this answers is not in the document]
lasinicl: Created an issue. https://github.com/ballerina-platform/ballerina-lang/issues/28839

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
lasinicl: Fixed in https://github.com/ballerina-platform/ballerina-lang/issues/28839

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1096,6 +1096,18 @@ private boolean isTupleTypeAssignable(BType source, BType target, Set<TypePair>
         return true;
     }
 
+    private boolean checkTupleMemberIsNoType(BType memberType) {
+        switch (memberType.tag) {
+            case TypeTags.NONE:
+                return true;
+            case TypeTags.TUPLE:
+                BTupleType tupleType = (BTupleType) memberType;
+                return tupleType.tupleTypes.stream().allMatch(this::checkTupleMemberIsNoType);
```
MaryamZi, line 1105: We try to avoid using streams due to perf impact identified previously. Shall we replace this with a loop?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2460,6 +2471,13 @@ private boolean hasSameReadonlyFlag(BType source, BType target) {
         }
 
         public Boolean visit(BTupleType t, BType s) {
+            if (((!t.tupleTypes.isEmpty() && t.tupleTypes.stream()
+                    .allMatch(mem -> checkTupleMemberIsNoType(mem))) ||
```
MaryamZi, line 2475: Shall we replace the stream here too?  I think we can change `checkTupleMemberIsNoType` to accept `t.tupleTypes` instead?

tests/jballerina-unit-test/src/test/resources/test-src/types/tuples/tuple_basic_test.bal
```diff
@@ -249,6 +249,178 @@ function testAnonRecordsInTupleTypeDescriptor() {
...
+    assertEquality(23, d1.b);
+    assertEquality(34, e1[0][0].id);
+    assertEquality(true, e1[0][0].flag);
+    assertEquality(56, e1[0][1]);
+
+    var [a2, ...b2] = getComplexTuple2();
+    assertEquality("Test", a2);
+    assertEquality(foo, b2[0][0]);
+    assertEquality(true, b2[0][1] is [BarObj,FooObj]);
+    [BarObj,FooObj] objects = <[BarObj,FooObj]> b2[0][1];
```
MaryamZi, line 373:
```suggestion
    [BarObj, FooObj] objects = <[BarObj, FooObj]> b2[0][1];
```

tests/jballerina-unit-test/src/test/resources/test-src/statements/vardeclr/module_tuple_var_decl.bal
MaryamZi, line 90: Still need to fail the else condition, right?

I suggest we do
```ballerina
    assertEquality(true, err is error);
    error err0 = <error> err;
    assertEquality("error msg", err0.message());
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1096,6 +1096,23 @@ private boolean isTupleTypeAssignable(BType source, BType target, Set<TypePair>
...
+    private boolean checkAllTupleMembersBelongNoType(List<BType> tupleTypes) {
+        boolean isNoType = false;
+        for (BType memberType : tupleTypes) {
+            switch (memberType.tag) {
+                case TypeTags.NONE:
+                    isNoType = true;
+                    break;
+                case TypeTags.TUPLE:
+                    isNoType = checkAllTupleMembersBelongNoType(((BTupleType) memberType).tupleTypes);
+                    break;
```
MaryamZi, line 1108: If `isNoType` here is `false` but there are subsequent member types that are noType, we will still be returning after this tuple member since we break here. Is this the expected behaviour?

lasinicl: Since this breaks the switch statement it will go to the next member in the foreach?

MaryamZi: Ugh, sorry, my bad! But then again, the name of the check implies we're checking if all is `noType`, right? Don't we need to return false in the default case or if `isNoType` is false here? Or a subsequent iteration being `noType` will hide the non-noTypes?

lasinicl: ahh.. yes! :) fixed in 946ca73

MaryamZi (approved):

PR #28798, opened 2021-02-22
Validate for invalid identifier when renaming

### LINKED ISSUE #28664: [LS] Validate new symbol name before executing rename operation
If the user provides an invalid symbol name (eg: `hello]` ), it will get replaced in all the places and breaks the code. 

So better to validate the given symbol before executing the rename operation.

### LINKED ISSUE #28817: [LS] Renaming the variable in match statement fails due to references being duplicated
**Description:**
Consider the following snippet where the variable `counter` in for loop to be renamed.
```
   foreach var counter in nums {
        match counter {
            0 => {
                found = true;
            }
            1 => {
                found = false;
            }
            2 => {
                found = true;
            }
        }
    }
```
The rename operation fails because the location of `counter` variable within the match statement being duplicated multiple times causing LS to return multiple text edits for the same location back to VSCode.

Language Server, Priority/High

IMS94:
## Purpose
$subject

Fixes #28664
Fixes #28817

## Approach
1. LS now validates the `newName` sent by the VSCode side for a valid identifier according to the [spec](https://ballerina.io/ballerina-spec/spec.html#lexical_structure) and shows an error in VSCode UI if the identifier is invalid. See the new behavior in following screen capture:

https://user-images.githubusercontent.com/11285165/108664088-0c2e3b80-74f8-11eb-86e6-0e4c2c1c79bc.mp4

2. Update `ReferenceFinder` to not visit `BLangMatchClause.matchExpr` since it is a cached node pointing to the expression of the match statement

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/constants/PatternConstants.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/constants/PatternConstants.java
@@ -16,7 +16,7 @@
 package org.ballerinalang.langserver.common.constants;
 
 /**
- * Holds constants related to regex patterns
+ * Holds constants related to regex patterns.
  *
  * @since 2.0.0
  */
@@ -54,7 +54,7 @@ public class PatternConstants {
             IDENTIFIER_ESCAPE_PATTERN + ")+";
     
     /**
-     * Regex to match a valid identifier name as per the ballerina specification
+     * Regex to match a valid identifier name as per the ballerina specification.
      */
     public static final String IDENTIFIER_PATTERN = UNQUOTED_IDENTIFIER_PATTERN + "|" + QUOTED_IDENTIFIER_PATTERN;
 }
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/constants/PatternConstants.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/constants/PatternConstants.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2021, WSO2 Inc. (http://wso2.com) All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ballerinalang.langserver.common.constants;
+
+/**
+ * Holds constants related to regex patterns
+ *
+ * @since 2.0.0
+ */
+public class PatternConstants {
+
+    private PatternConstants() {
+    }
+
+    public static final String UNICODE_PATTERN_WHITESPACE_CHAR_PATTERN = "[\u200E\u200F\u2028\u2029]";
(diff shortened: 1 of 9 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
```diff
@@ -692,6 +693,20 @@ private static String generateVariableName(int suffix, String name, Set<String>
         return newName;
     }
 
+    /**
+     * Checks if the provided identifier is valid as per the ballerina specification.
+     *
+     * @param identifier Identifier to be checked for validity
+     * @return True, if the identifier is valid as per the ballerina specification
+     */
+    public static boolean isIdentifierValid(String identifier) {
```
nadeeshaan, line 702: Maybe we can change the name to `isValidIdentifier`

nadeeshaan (dismissed):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
IMS94: Done!

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/constants/PatternConstants.java
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+package org.ballerinalang.langserver.common.constants;
+
+/**
+ * Holds constants related to regex patterns.
+ *
+ * @since 2.0.0
+ */
+public class PatternConstants {
```
rasika, line 23: Any reason to duplicate the identifier matching logic here. Can't we reuse something that Parser/Runtime uses for this? My point is any changes to identifier matching patterns will not reflected here and we might need to maintain this.

IMS94: I had the same doubt. Do you know where the identifier matching logic is in the parser?

rasika: there's a similar class called [IdentifierUtils](https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/IdentifierUtils.java), but need to verify it from the compiler/runtime team. @lochana-chathura @rdulmina anyway to derive `isIdentifier()` capability through Compiler-APIs?

lochana-chathura: We do the identifier matching here in the lexer. https://github.com/ballerina-platform/ballerina-lang/blob/317188576353a17d5abbde7fb1e3bb153811d0ac/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaLexer.java#L783

It is not merely an `isIdentifier(String txt)` method. We read the input sequence char by char using the attached token reader to capture the identifier. So cannot expose as it is. But, maybe we can write a method to boot up a parser and try to parse an identifier from a given string and then return whether it is valid. But, currently, we don't have a place to expose such API.

nadeeshaan: In order to have a common approach we need to expose it via the Syntax API. Hence I'm +1 for the current approach which @IMS94 has followed.

nadeeshaan (approved):

Issue #28806, opened 2021-02-22
Ballerina version should be sent to central for all the resources


pramodya1994:
**Description:**
Ballerina version should be set as User-Agent for all the resources of the central.
The following header should be added to all the remote methods of the central API.
```java
conn.setRequestProperty(USER_AGENT, ballerinaVersion);
```

User-Agent has added to pushing and pulling for this issue https://github.com/ballerina-platform/ballerina-lang/issues/28627, need to add this for all other resources.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #28810, opened 2021-02-22
Add ballerina version header to all central client methods

pramodya1994:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/28806

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
+++ b/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
@@ -85,13 +85,15 @@ public class CentralAPIClient {
     /**
      * Get package with version.
      *
-     * @param orgNamePath     The organization name of the package. (required)
-     * @param packageNamePath The name of the package. (required)
-     * @param version         The version or version range of the module. (required)
+     * @param orgNamePath       The organization name of the package. (required)
+     * @param packageNamePath   The name of the package. (required)
+     * @param version           The version or version range of the module. (required)
+     * @param supportedPlatform The ballerina platform. (required)
+     * @param ballerinaVersion  The ballerina version. (required)
      * @return PackageJsonSchema
      */
-    public Package getPackage(String orgNamePath, String packageNamePath, String version, String supportedPlatform)
-            throws CentralClientException {
+    public Package getPackage(String orgNamePath, String packageNamePath, String version, String supportedPlatform,
+            String ballerinaVersion) throws CentralClientException {
         initializeSsl();
         String url = PACKAGES + "/" + orgNamePath + "/" + packageNamePath;
         // append version to url if available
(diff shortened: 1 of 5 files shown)


cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
```diff
@@ -175,6 +180,7 @@ public Package getPackage(String orgNamePath, String packageNamePath, String ver
 
         // Set headers
         conn.setRequestProperty(BALLERINA_PLATFORM, supportedPlatform);
+        conn.setRequestProperty(USER_AGENT, ballerinaVersion);
```
hemikak, line 183: Can we move these to the `createHttpUrlConnection` function ?

pramodya1994: Is that the best approach? Else we can create a new method and reuse it inside all the methods. WDYT?

hemikak: I think the best way to avoid resurfacing the issues we are facing is to move them to that function. Any new function added to this class that uses the client will have User-Agent header then.

pramodya1994: Resolved in bf00c0696a3ebedef52553ece1642d5f705129d1

cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
```diff
@@ -402,22 +398,34 @@ public PackageSearchResult searchPackage(String query, String ballerinaVersion)
...
             }
         } catch (IOException e) {
             throw new ConnectionErrorException("Creating connection to '" + url + "' failed:" + e.getMessage());
         }
     }
+
+    private HttpURLConnection setRequiredHeaders(HttpURLConnection conn, String supportedPlatform,
+            String ballerinaVersion) {
+        conn.setRequestProperty(BALLERINA_PLATFORM, supportedPlatform);
+        conn.setRequestProperty(USER_AGENT, ballerinaVersion);
```
hemikak, line 428: We can directly use `RepoUtils.getBallerinaVersion()` for the `User-Agent` header instead of passing the value. Please check if its correct.

pramodya1994: For `pullPackage` in `CentralApiClient` we use the ballerina version for identifying whether it is a snapshot version or not, So we need to pass this for that. Thats why I kept all resources to pass the ballerina version.
If we need we can refactor other resources as you have mentioned in the above comment.

hemikak: Okay, I am okay with that.

hemikak (approved):

Issue #28139, opened 2021-01-22
Create variable code action generates invalid variable for functions returns multi type arrays


pramodya1994:
**Description:**
Create variable code completion action creates invalid code for the following function.

method signature:
```ballerina
remote function getColumn(string column) returns @tainted (string | int | float)[] | error
```

Before action:
<img width="721" alt="Screenshot 2021-01-22 at 22 26 00" src="https://user-images.githubusercontent.com/19742678/105520523-d16e9300-5d00-11eb-9844-6bb70e9521ac.png">

After action:
<img width="725" alt="Screenshot 2021-01-22 at 22 26 46" src="https://user-images.githubusercontent.com/19742678/105520618-eba87100-5d00-11eb-8064-2b5f8a235727.png">

Generated code should be like this:
```ballerina
(string|int|float)[]|error column = sheet->getColumn("A");
```

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #28816, opened 2021-02-22
Fix signature fails for unions of array type members

rasika:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/28139

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaArrayTypeSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaArrayTypeSymbol.java
@@ -49,6 +49,9 @@ public class BallerinaArrayTypeSymbol extends AbstractTypeSymbol implements Arra
 
     @Override
     public String signature() {
+        if (memberTypeDescriptor().typeKind() == TypeDescKind.UNION) {
+            return "(" + memberTypeDescriptor().signature() + ")[]";
+        }
         return memberTypeDescriptor().signature() + "[]";
     }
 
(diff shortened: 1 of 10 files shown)


IMS94 (approved):
LGTM!

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaArrayTypeSymbol.java
```diff
@@ -49,6 +49,9 @@ public TypeSymbol memberTypeDescriptor() {
 
     @Override
     public String signature() {
+        if (memberTypeDescriptor().typeKind() == TypeDescKind.UNION) {
```
nadeeshaan, line 52: Shall we introduce a `ParanthesizedTypeDescriptor`, so we don't need to special case the union here?

nadeeshaan (changes requested):

nadeeshaan (approved):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaArrayTypeSymbol.java
rasika: As discussed, this could also appear without parenthesis as well. hence we need this.
```ballerina
type Value string|int|float;
Value[]|error myValue = [];
```

PR #28845, opened 2021-02-23
Refactor native unit test methods to use api types and values 

warunalakshitha:
## Purpose
ATM unit tests are using internal types and values. But actual interop users will api types and values. Hence we need to use api types and values inside unit tests as well.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ArrayValueImpl.java
```diff
@@ -228,7 +229,7 @@ public ArrayValueImpl(ArrayType type, long size) {
         }
     }
 
-    public ArrayValueImpl(ArrayType type, long size, ListInitialValueEntry[] initialValues) {
+    public ArrayValueImpl(ArrayType type, long size, BInitialValueEntry[] initialValues) {
```
HindujaB, line 232: Why do we change the initial values into the abstract type here? 
Because `BInitialValueEntry` is common for both list and mapping values. 
Wouldn't it be more specific to keep them as `ListInitialValueEntry[]` for arrays & tuples?

PR #28877, opened 2021-02-24
Improve parameter symbol implementation

pubudu91:
## Purpose
This PR makes `ParameterSymbol` a proper symbol by extending from `Symbol`. This also resolves a bug we had where if the parameters were looked up directly, the returned symbols were var symbols.

## Remarks
I introduced separate flags for the different types of params since the other solution of introducing a separate internal symbol for the params turned out to be a bigger task than expected.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/ParameterCompletionItemBuilder.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/ParameterCompletionItemBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
  *
  * WSO2 Inc. licenses this file to you under the Apache License,
  * Version 2.0 (the "License"); you may not use this file except
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/ParameterCompletionItemBuilder.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/ParameterCompletionItemBuilder.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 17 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -164,6 +167,13 @@ public void sort(BallerinaCompletionContext context, T node, List<LSCompletionIt
                 CompletionItem variableCItem = VariableCompletionItemBuilder.build(varSymbol, varSymbol.getName().get(),
                         typeName);
                 completionItems.add(new SymbolCompletionItem(ctx, symbol, variableCItem));
+            } else if (symbol.kind() == PARAMETER) {
+                ParameterSymbol varSymbol = (ParameterSymbol) symbol;
+                TypeSymbol typeDesc = (varSymbol).typeDescriptor();
```
nadeeshaan, line 172: unnecessary parenthesis

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -164,6 +167,13 @@ public void sort(BallerinaCompletionContext context, T node, List<LSCompletionIt
...
+                String typeName = typeDesc == null ? "" : CommonUtil.getModifiedTypeName(ctx, typeDesc);
```
nadeeshaan, line 173: here the type desc could be null, instead what if we make the typeDescriptor API to return an Optional, which is to be consistent with the other APIs

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
pubudu91: Removed this null check. There shouldn't be a case where the typedesc becomes null.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/ParameterCompletionItemBuilder.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
nadeeshaan, line 2: Year 2021?

nadeeshaan (changes requested):

nadeeshaan (approved):
LGTM

rdhananjaya (approved):

MaryamZi:
Can't we get rid of the the `defaultableParam` field in `BVarSymbol` with this change? https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BVarSymbol.java#L40

MaryamZi (approved):

pubudu91:
> Can't we get rid of the the `defaultableParam` field in `BVarSymbol` with this change? https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BVarSymbol.java#L40

Should be able to. Will check and address separately.

PR #28938, opened 2021-03-01
Allow `mapping-constructor-expr` to be optional in annotations

### LINKED ISSUE #27498: Mapping constructor in annotation can be optional
**Description:**
In the spec under annotation section it says 
`If the type of the annotation value is true, then a mapping-constructor-expr is not allowed, and the annotation value is true. Otherwise the type of the annotation value must be a mapping type M; if a mapping-constructor-expr is not specified, then it defaults to { }; the mapping-constructor-expr must have static type M and the annotation value is the result of evaluating the mapping-constructor-expr with M as the contextually expected type.
`

```ballerina
annotation T N on function;

type T record {|
    int i = 0;
|};

@N  // annotation value expected for annotation 'N'
public function main() {
    
}
```

It seems current implementation requires the empty mapping constructor even when mapping value can be constructed without it.

**Steps to reproduce:**

**Affected Versions:**
SLP7.

### LINKED ISSUE #28885: Allow `mapping-constructor-expr` to be optional in annotations
The tracking issue for https://github.com/ballerina-platform/ballerina-spec/issues/447

lasinicl:
## Purpose
$title.

Fixes #28885
Fixes #27498

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -1294,6 +1294,12 @@ public void visit(BLangAnnotation annotationNode) {
 
     public void visit(BLangAnnotationAttachment annAttachmentNode) {
         annAttachmentNode.expr = rewrite(annAttachmentNode.expr, env);
```
grainier, line 1296: Can take `annAttachmentNode.expr != null` before re-writing. 
```suggestion
        if(annAttachmentNode.expr != null) {
            annAttachmentNode.expr = rewrite(annAttachmentNode.expr, env);
```

grainier (dismissed):

rdhananjaya:
@lasinicl can you please create an issue to update BBEs with this change

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
```diff
@@ -300,6 +300,10 @@ private void attachSchedulerPolicy(BLangFunction function) {
             if (!annotation.annotationName.value.equals("strand")) {
                 continue;
             }
+            if (annotation.expr == null) {
```
MaryamZi, line 303: Shouldn't this get added at desugar?

I'm assuming we're adding this here because of the field-wise check after this? IMO, we should fix that check instead of adding an empty mapping constructor expression here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -3500,25 +3500,29 @@ private void validateAnnotationAttachmentExpr(BLangAnnotationAttachment annAttac
...
-        // map<anydata>[]|record{ anydata...; }[], thus an expression is required.
-        if (annAttachmentNode.expr == null) {
+        // map<anydata>[]|record{ anydata...; }[]. For map<anydata>[]|record{ anydata...; }[] an expression is
+        // required while map<anydata>|record{ anydata...; } can be empty.
+        if (annAttachmentNode.expr == null && annotationSymbol.attachedType != null &&
+                !(annotationSymbol.attachedType.type.tag == TypeTags.MAP ||
+                        annotationSymbol.attachedType.type.tag == TypeTags.RECORD)) {
             this.dlog.error(annAttachmentNode.pos, DiagnosticErrorCode.ANNOTATION_ATTACHMENT_REQUIRES_A_VALUE,
-                            annotationSymbol);
+                    annotationSymbol);
```
MaryamZi, line 3510: Is the assumption regarding arrays here correct? The spec says 

> For an annotation tag declared with type T, **if T is M[] for some type mapping type M, then the type of the annotation value is M**, multiple annotations are allowed, and the effective value is an array of the values for each annotation; otherwise the type of the annotation value is T, multiple annotations are not allowed, the effective value is the value of the single annotation.

and

> If the type of the annotation value is true, then a mapping-constructor-expr is not allowed, and the annotation value is true. **Otherwise the type of the annotation value must be a mapping type M; if a mapping-constructor-expr is not specified, then it defaults to { }**; the mapping-constructor-expr must have static type M and the annotation value is the result of evaluating the mapping-constructor-expr with M as the contextually expected type.

So even in the array scenario, the type of the annotation value can be a mapping for which there is no mapping constructor expression specified, right?

MaryamZi: Basically, shouldn't the following work?

```ballerina
annotation map<int>[] x on function;

@x
@x
function func() {
} 
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -3500,25 +3500,29 @@ private void validateAnnotationAttachmentExpr(BLangAnnotationAttachment annAttac
...
             return;
         }
 
-        BType annotType = annotationSymbol.attachedType.type;
-        this.typeChecker.checkExpr(annAttachmentNode.expr, env,
-                                   annotType.tag == TypeTags.ARRAY ? ((BArrayType) annotType).eType : annotType);
+        if (annAttachmentNode.expr != null) {
```
MaryamZi, line 3514: If `annAttachmentNode.expr` is null don't we have to do some semantic validation to check if the empty mapping constructor expression is valid for that particular type?

For example, IMO, the following should result in a compilation error, since the record type has a required field. Right now, this seems to be allowed at compile-time resulting in invalid values at runtime.

This crashes at runtime.
```ballerina
type Record record {| int[] i; |};

annotation Record x on function;

@x
function func() {
} 

public function main() { 
   Record rec = <Record> (typeof func).@x; 
   int i = rec.i[0];
}
```

```log
[2021-03-03 10:15:03,754] SEVERE {b7a.log.crash} - null 
    at test.main(test.bal:22) 
    at $_init.$lambda$main$(.) 
    at io.ballerina.runtime.internal.scheduling.SchedulerItem.execute(Scheduler.java:571) 
    at io.ballerina.runtime.internal.scheduling.Scheduler.run(Scheduler.java:301) 
    at io.ballerina.runtime.internal.scheduling.Scheduler.runSafely(Scheduler.java:269) 
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -3500,25 +3500,29 @@ private void validateAnnotationAttachmentExpr(BLangAnnotationAttachment annAttac
...
         }
 
-
         // At this point the type is a subtype of  map<anydata>|record{ anydata...; } or
-        // map<anydata>[]|record{ anydata...; }[], thus an expression is required.
-        if (annAttachmentNode.expr == null) {
+        // map<anydata>[]|record{ anydata...; }[]. For map<anydata>[]|record{ anydata...; }[] an expression is
+        // required while map<anydata>|record{ anydata...; } can be empty.
+        if (annAttachmentNode.expr == null && annotationSymbol.attachedType != null &&
+                !(annotationSymbol.attachedType.type.tag == TypeTags.MAP ||
```
rdhananjaya, line 3507: When annotation declaration support multiple annotations, `annotationSymbol.attachedType` is a list, IMO we don't need to provide empty mapping constructor for these scenarios.

```ballerina
type T map<int>;
annotation T[] ANNOT on function;

@ANNOT {}
@ANNOT {}
public function main(string... argv) {
}
```

rdhananjaya: Also current changes allow below code, which is wrong as we can't provide empty mapping constructor here.
```ballerina
type T record { int i; };

annotation T ANNOT on function;

@ANNOT
public function main(string... argv) {
}
```

rdhananjaya: My comments are as same as @MaryamZi please ignore these.

lasinicl:
Created https://github.com/ballerina-platform/ballerina-lang/issues/28986 to update the BBEs and rest of the comments have been addressed via 4467ec7

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -3500,18 +3500,14 @@ private void validateAnnotationAttachmentExpr(BLangAnnotationAttachment annAttac
...
         }
 
-
-        // At this point the type is a subtype of  map<anydata>|record{ anydata...; } or
-        // map<anydata>[]|record{ anydata...; }[], thus an expression is required.
         if (annAttachmentNode.expr == null) {
-            this.dlog.error(annAttachmentNode.pos, DiagnosticErrorCode.ANNOTATION_ATTACHMENT_REQUIRES_A_VALUE,
-                            annotationSymbol);
-            return;
+            annAttachmentNode.expr = new BLangRecordLiteral();
```
MaryamZi, line 3504: I'm not sure if we should be adding this here (we usually defer modifying the AST until we get to desugar).

Also, wouldn't it be confusing to have this go in the normal type checking path for a mapping constructor, since there's no visible mapping constructor, here? How about checking if the type is a record with required fields (that do not have defaults) here and logging a specific error like

```
annotation value expected for annotation of record type ''{0}'' with required fields
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
```diff
@@ -300,35 +300,37 @@ private void attachSchedulerPolicy(BLangFunction function) {
...
-                BLangRecordLiteral.BLangRecordKeyValueField keyValue =
-                        (BLangRecordLiteral.BLangRecordKeyValueField) field;
-                BLangExpression expr = keyValue.key.expr;
-                if (expr.getKind() != NodeKind.SIMPLE_VARIABLE_REF) {
-                    continue;
-                }
-                BLangIdentifier variableName = ((BLangSimpleVarRef) expr).variableName;
-                if (variableName.value.equals("name")) {
-                    if (keyValue.valueExpr.getKind() != NodeKind.LITERAL) {
+            if (annotation.expr != null) {
```
MaryamZi, line 303: We can probably check the inverse and return/continue early to avoid the indentation in the block, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -1292,8 +1292,15 @@ public void visit(BLangAnnotation annotationNode) {
     }
 
     public void visit(BLangAnnotationAttachment annAttachmentNode) {
-        annAttachmentNode.expr = rewrite(annAttachmentNode.expr, env);
+        if (annAttachmentNode.expr == null && annAttachmentNode.annotationSymbol.attachedType != null) {
+            BType attachedType = annAttachmentNode.annotationSymbol.attachedType.type;
+            if (!(attachedType.tag == TypeTags.FINITE)) {
```
MaryamZi, line 1297:
```suggestion
            if (attachedType.tag != TypeTags.FINITE) {
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7722,4 +7722,19 @@ private FieldInfo(List<BType> types, boolean required, boolean readonly) {
...
         }
     }
+
+    /**
+     * Check whether a record type has required fields.
+     *
+     * @param recordType Record type.
+     * @return true if the record type has required fields; false otherwise.
+     */
+    public boolean checkRecordTypeForRequiredFields(BRecordType recordType) {
```
MaryamZi, line 7732: I think we can have this in the SemanticAnalyzer itself. Shall we also rename this to something like
```suggestion
    public boolean hasRequiredFields(BRecordType recordType) {
```

MaryamZi (commented):
@lasinicl, shall we also add content about this change to the release note file?

lasinicl:
> @lasinicl, shall we also add content about this change to the release note file?

Added in https://github.com/ballerina-platform/ballerina-release/pull/101

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/annotations/AnnotationAttachmentTest.java
```diff
@@ -367,4 +369,89 @@ public void testAnnotsWithConstLists() {
         Assert.assertEquals(element.getKind(), NodeKind.LITERAL);
         Assert.assertEquals(((BLangLiteral) element).value, "test");
     }
+
```
MaryamZi, line 372: Sorry, missed this previously. But shall we add some runtime tests too? See org.ballerinalang.test.annotations.AnnotationRuntimeTest. The data provider at https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/annotations/AnnotationRuntimeTest.java#L87 doesn't seem to be used though  (the test that used that seems to have gotten removed :/), so we'll have to add that back too.

MaryamZi: Basically let's test the following,
1. map type
2. record type with required fields that have defaults (at runtime we can check if the default was set)
3. array of maps or records

lasinicl: Added runtime tests in b75e5a2. 
`testServiceAnnotAccess1` test is disabled since it was failing. Created https://github.com/ballerina-platform/ballerina-lang/issues/29033 to track it.

MaryamZi (approved):

PR #28968, opened 2021-03-03
[Debugger] Add conditional breakpoint support

### LINKED ISSUE #28565: Ballerina debugger support for conditional breakpoints
**Description:**
Currently Ballerina debugger does not seems to support conditional breakpoints. Since now we are supporting evaluate expressions, how about working on conditional breakpoints support. This would be really useful as we used to do the same for other languages and IDEs.

We can implement it even without runtime support; for instance runtime sees a conventional breakpoint. When the breakpoint got hit; debugger implementation evaluate the current condition against runtime values. if it is evaluated as true, we break; or else auto resume the breakpoint. BTW performance will be further improved if the runtime natively support conditional breakpoints.

conditiona-bp

**Steps to reproduce:**

**Affected Versions:**

praveennadarajah:
This PR will add conditional breakpoint support for Ballerina.

Fixes #28565

https://user-images.githubusercontent.com/11292766/109760854-70967c80-7c15-11eb-9233-2a80d75b3ced.mov


misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -179,7 +179,7 @@ ClientConfigHolder getClientConfigHolder() {
         // Todo - Implement
         capabilities.setSupportsCompletionsRequest(false);
         capabilities.setSupportsRestartRequest(false);
-        capabilities.setSupportsConditionalBreakpoints(false);
+        capabilities.setSupportsConditionalBreakpoints(true);
```
NipunaRanasinghe, line 182: Lets move up this line to the `supported capabilities` list.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -464,6 +472,19 @@ public void sendOutput(String output, String category) {
...
+            SuspendedContext ctx = new SuspendedContext(project, context.getDebuggeeVM(),
+                    getAllThreads().get(threadReference.uniqueID()), frame);
+            ExpressionEvaluator evaluator = new ExpressionEvaluator(ctx);
+            return evaluator.evaluate(expression);
+        } catch (JdiProxyException e) {
+            LOGGER.error(e.getMessage(), e);
+            return null;
+        }
+    }
+
```
NipunaRanasinghe, line 487: Shall we make this package-private + static?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -478,6 +499,15 @@ private Breakpoint toBreakpoint(SourceBreakpoint sourceBreakpoint, Source source
...
     }
 
+    private BalBreakpoint toBalBreakpoint(SourceBreakpoint sourceBreakpoint, Source source) {
+        BalBreakpoint balBreakpoint = new BalBreakpoint();
+        balBreakpoint.setLine(sourceBreakpoint.getLine());
+        balBreakpoint.setSource(source);
+        balBreakpoint.setVerified(true);
+        balBreakpoint.setCondition(sourceBreakpoint.getCondition());
+        return balBreakpoint;
+    }
```
NipunaRanasinghe, line 509: Why cant we use the `SourceBreakpoint` instances itself instead of creating a custom breakpoint type, as it seems like `BalBreakpoint` is used only to hold a subset of information of the `SourceBreakpoint`?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
```diff
@@ -103,11 +104,31 @@ private boolean processEvent(EventSet eventSet, Event event) {
...
-            stoppedEventArguments.setThreadId(((BreakpointEvent) event).thread().uniqueID());
-            stoppedEventArguments.setAllThreadsStopped(true);
-            context.getClient().stopped(stoppedEventArguments);
+            Value value = null;
+            String qualifiedClassName =
+                getQualifiedClassName(context, ((BreakpointEvent) event).location().declaringType());
+            Map<Integer, BalBreakpoint> balBreakpoints = this.balBreakpoints.get(qualifiedClassName);
+            int lineNumber = ((BreakpointEvent) event).location().lineNumber();
+
+            if (balBreakpoints != null && balBreakpoints.containsKey(lineNumber)) {
```
NipunaRanasinghe, line 113: Is this null check for `balBreakpoints` necessary in here?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
```diff
@@ -103,11 +104,31 @@ private boolean processEvent(EventSet eventSet, Event event) {
...
+                value = context.getAdapter().evaluateExpression(balBreakpoints.get(lineNumber).getCondition(),
+                    ((BreakpointEvent) event).thread());
+            }
+            if (balBreakpoints != null && balBreakpoints.containsKey(lineNumber)
```
NipunaRanasinghe, line 117: Is this null check for `balBreakpoints` necessary in here?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
```diff
@@ -103,11 +104,31 @@ private boolean processEvent(EventSet eventSet, Event event) {
...
+                && balBreakpoints.get(lineNumber).getCondition() != null
+                && !Boolean.parseBoolean(Objects.requireNonNull(value).toString())) {
```
NipunaRanasinghe, line 119: Directly trying to parse into boolean is unsafe, as we cannot guarantee about the type of conditional expression result (can be non-boolean)

NipunaRanasinghe (changes requested):

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -464,6 +472,19 @@ public void sendOutput(String output, String category) {
         }
     }
 
+    public Value evaluateExpression(String expression, ThreadReference threadReference) {
```
NipunaRanasinghe, line 475: How about using a name like `evaluateBreakpointCondition`, as this util function is specifically written of that purpose? Also, Shall we modify the method to return a `boolean` instead of returning the direct result (which is a JDI value instance)?

NipunaRanasinghe:
@praveennadarajah Are we planning add some integration tests to cover conditional break point scenarios?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Removed custom `balBreakpoint` class and used `SourceBreakpoint` information

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/BalBreakpoint.java
```diff
@@ -0,0 +1,64 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.ballerinalang.debugadapter;
+
+import org.eclipse.lsp4j.debug.Source;
+
+/**
+ * Information about ballerina breakpoint (Properties).
```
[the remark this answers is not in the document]
praveennadarajah: Refer [this](https://github.com/ballerina-platform/ballerina-lang/pull/28968#discussion_r586286907) comment.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/BalBreakpoint.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class BalBreakpoint {
+
+    private Long line;
+    private Source source;
+    private Boolean verified;
+    private String condition;
```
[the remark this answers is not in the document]
praveennadarajah: Refer [this](https://github.com/ballerina-platform/ballerina-lang/pull/28968#discussion_r586286907) comment.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Done.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Cannot make to `static` as this method is using non-static members. This method is written in `JBallerinaDebugServer` class to get the maximum use of its class variable and used in `JDIEventProcessor` class. Therefore it cannot be made `private`. Therefore keeping it as `public`

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
praveennadarajah: There are instances where `balBreakpoints` can be null. In that case `balBreakpoints.containsKey(lineNumber)` throwing a NPE. Therefore this check is required.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
praveennadarajah: Refer to [this](https://github.com/ballerina-platform/ballerina-lang/pull/28968#discussion_r586292855) comment.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
praveennadarajah: Fixed

praveennadarajah:
> @praveennadarajah Are we planning add some integration tests to cover conditional breakpoints related scenarios?

@NipunaRanasinghe Current debug framework does not support conditional breakpoints. We need to implement that first and then need to add possible integration tests.

I have created https://github.com/ballerina-platform/ballerina-lang/issues/28981 to track this issue.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Done

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
NipunaRanasinghe: @praveennadarajah I meant, `package-private` :slightly_smiling_face: Also, whats the reason for keeping this util inside `JballerinaDebugServer`, since this is used only within `JDIEventProcessor`? (If your intention is to use the state variables of `JBallerinaDebugServer`, those are already accessible from the `JDIEventProcessor`)

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
NipunaRanasinghe: I'm still not sure how it can be null, given that its declared as final and already initialized as an empty map :slightly_smiling_face:

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
```diff
@@ -103,11 +105,30 @@ private boolean processEvent(EventSet eventSet, Event event) {
...
+                continuedEventArguments.setThreadId(((BreakpointEvent) event).thread().uniqueID());
+                continuedEventArguments.setAllThreadsContinued(true);
+                context.getClient().continued(continuedEventArguments);
+            } else {
+                StoppedEventArguments stoppedEventArguments = new StoppedEventArguments();
+                stoppedEventArguments.setReason(StoppedEventArgumentsReason.BREAKPOINT);
+                stoppedEventArguments.setThreadId(((BreakpointEvent) event).thread().uniqueID());
+                stoppedEventArguments.setAllThreadsStopped(true);
+                context.getClient().stopped(stoppedEventArguments);
+            }
```
NipunaRanasinghe, line 131: Can't we simplify the logic in here as follows?
```java
if (condition) {
   // suspend
} else {
  // continue
} 
```

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
praveennadarajah: According to the recent,
`Map<Integer, SourceBreakpoint> sourceBreakpoints = this.sourceBreakpoints.get(qualifiedClassName);` 

Here `sourceBreakpoints` can be null when `this.sourceBreakpoints` doesn't contain any values for key `qualifiedClassName`

NipunaRanasinghe: Sorry I've missed it since you've used the same variable name locally. +1

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -464,6 +469,20 @@ public void sendOutput(String output, String category) {
...
     }
 
+    public boolean evaluateBreakpointCondition(String expression, ThreadReference threadReference) {
+        try {
+            StackFrameProxyImpl frame = getAllThreads().get(threadReference.uniqueID()).frame(0);
+            SuspendedContext ctx = new SuspendedContext(project, context.getDebuggeeVM(),
+                    getAllThreads().get(threadReference.uniqueID()), frame);
+            ExpressionEvaluator evaluator = new ExpressionEvaluator(ctx);
+            String condition = evaluator.evaluate(expression).toString();
+            return condition.equalsIgnoreCase(CONDITION_TRUE);
```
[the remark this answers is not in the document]
praveennadarajah: Here if the expression returns a boolean `true`, then String `condition` will get the value `"true"`, and
if the expression returns String `"true"`, the String `condition` will get the value `""true""`. 
Therefore this logic might not cause false positive.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Moved the util method to `JDIEventProcess` class and made it private.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
praveennadarajah: Done

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
```diff
@@ -144,6 +166,20 @@ private boolean processEvent(EventSet eventSet, Event event) {
...
+            SuspendedContext ctx = new SuspendedContext(context.getSourceProject(), context.getDebuggeeVM(),
+                    context.getAdapter().getAllThreads().get(threadReference.uniqueID()), frame);
+            ExpressionEvaluator evaluator = new ExpressionEvaluator(ctx);
+            String condition = evaluator.evaluate(expression).toString();
+            return condition.equalsIgnoreCase(CONDITION_TRUE);
+        } catch (JdiProxyException e) {
+            LOGGER.error(e.getMessage(), e);
+            return false;
+        }
+    }
```
NipunaRanasinghe, line 181: In here, Your logic returns `false` for the expressions which produces a non-boolean results as well. Shall we improve this behavior by notifying the user that it is an invalid expression instead?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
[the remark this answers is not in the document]
NipunaRanasinghe: Refer https://github.com/ballerina-platform/ballerina-lang/pull/28968#discussion_r587085780

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JDIEventProcessor.java
praveennadarajah: Yes, we can notify the user when they add an invalid expression. 
I have created https://github.com/ballerina-platform/ballerina-lang/issues/28992 to track this.

NipunaRanasinghe (dismissed):

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/BalBreakpoint.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public void setCondition(String condition) {
+        this.condition = condition;
+    }
+
+    public Breakpoint getBreakpoint() {
+        Breakpoint breakpoint = new Breakpoint();
+        breakpoint.setLine(line);
+        breakpoint.setSource(source);
+        breakpoint.setVerified(true);
```
NipunaRanasinghe, line 72:
```suggestion
        breakpoint.setVerified(verified);
```

NipunaRanasinghe (approved):

Issue #28285, opened 2021-01-29
Incorrect error with 'is'


jclark:
This correct program:

```
import ballerina/io;

public function main() {
   boolean b = "str" is string;
   io:println(b);
}
```

gets rejected with the error:

```
ERROR [demo8.bal:(4:16,4:31)] unnecessary condition: expression will always evaluate to 'true'
```

hasithaa:
Related to https://github.com/ballerina-platform/ballerina-spec/issues/724

jclark:
@hasithaa I’m not using this in a condition, so it’s clearly not an error regardless of https://github.com/ballerina-platform/ballerina-spec/issues/724

hasithaa:
@jclark This is one of the root cause for ballerina-platform/ballerina-spec#724, That is why I liked it.

PR #28985, opened 2021-03-03
Restrict deep value equality for readonly type 

### LINKED ISSUE #28324: Operator `==` can be used with `readonly` type
**Description:**

> For both == and !=, both operands must have a static type that is a subtype of anydata.

This means == is not defined for `readonly` type.

This fix will propagate to `value:Cloneable` type as `readonly` type is a subtype in the `Cloneable` union.

**Steps to reproduce:**

**Affected Versions:**

#28324

pcnfernando:
Fixes #28324

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/freeze-and-isfrozen.bal
```diff
@@ -456,15 +455,6 @@ function testValidSelfReferencingValueFreeze() returns [string, boolean] {
     return [FREEZE_SUCCESSFUL, !m.isReadOnly() && res.isReadOnly()];
 }
 
-function testStructureWithErrorValueFreeze() returns boolean {
```
rdhananjaya, line 459: Why is this removed?

pcnfernando: This test was checking whether an error is readonly. As per spec, isReadOnly() accepts only anydata. Hence, removed because the test is not applicable

rdhananjaya: I think we can you `res is readonly` here. 
Created a spec issue to clarify why that function is restricted to `anydata`.  https://github.com/ballerina-platform/ballerina-spec/issues/769

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/IsPureTypeUniqueVisitor.java
```diff
@@ -83,7 +83,7 @@ private boolean isAnydata(BType type) {
     }
 
     private boolean isPureType(BType type) {
-        return type.tag == TypeTags.ANYDATA || type.tag == TypeTags.READONLY || isAnydata(type);
+        return type.tag == TypeTags.ANYDATA || isAnydata(type);
```
rdhananjaya, line 86: I think we should replace all calls to `isPureType` with `isAnydata` as it doens't do anything other than calling `isAnydata`

tests/jballerina-unit-test/src/test/resources/test-src/expressions/builtinoperations/freeze-and-isfrozen.bal
pcnfernando: Once https://github.com/ballerina-platform/ballerina-lang/issues/23691 get fixed, cloneReadOnly() would anyway return   `CloneableType & readonly` which would make `res is readonly` always true

rdhananjaya: We are removing that `always true` error with https://github.com/ballerina-platform/ballerina-lang/issues/28285

pcnfernando: What I meant was even though the compiler validation for always true will be removed; I believe the testing scenario is not correct as we assert something that's always true. WDYT?

rdhananjaya: `is` test is a runtime check, so we verify the readonly-ness of the value (whether the readonly bit is set or not in the cloned value). 
Not sure how it's always true, as we are testing the behavior of `cloneReadOnly`.

KRVPerera: I believe that due to the current restriction on error casting we cannot do a is check with `Cloneable` of `map<anydata|error>`. May we have to disable this test case only temporarily

pcnfernando: My confusion was with the below sample as the `if check` will always be true. Clarified it from @rdhananjaya and added the test back in e3df68af6b6da7ae7b9a4e539c815fd41e26cf56.  For the moment, commented out the `is readonly` check. The test should be updated once #23691 is fixed.
```ballerina
import ballerina/io;

public function main() {
    var er = getReadOnlyError();
    if(er is readonly) {
        io:println("er is readonly");
    }
}

function getReadOnlyError() returns error & readonly {
    return error("Custom error");
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/IsPureTypeUniqueVisitor.java
pcnfernando: Updated in 6cfc931120f36be6ef71cfda98de8882e18c14e4

rdhananjaya (approved):

rdhananjaya:
Changes look good to me, we need to have the release note PR to merge this.

Issue #29051, opened 2021-03-08
Refactor service-based transaction to sync with http changes


pcnfernando:
Ref: https://github.com/ballerina-platform/ballerina-standard-library/issues/1047

PR #29054, opened 2021-03-08
Create a transaction branch for new strands if parent is in transaction

pcnfernando:
Addresses #29051

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Environment.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Environment.java
@@ -21,6 +21,7 @@ import io.ballerina.runtime.api.async.StrandMetadata;
 import io.ballerina.runtime.internal.scheduling.State;
 import io.ballerina.runtime.internal.scheduling.Strand;
 
+import java.util.Map;
 import java.util.Optional;
 
 /**
@@ -118,5 +119,14 @@ public class Environment {
      */
     public Object getStrandLocal(String key) {
         return strand.getProperty(key);
+    }
+
+    /**
+     * Gets the property map stored in the strand.
+     *
+     * @return property map stored in the strand.
+     */
+    public Map<String, Object> getStrandLocalProperties() {
+        return strand.globalProps;
     }
 }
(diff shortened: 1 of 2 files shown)


riyafa:
Shall we add an explanation on why we need to expose strand related properties via the Environment?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Environment.java
```diff
@@ -119,4 +120,13 @@ public void setStrandLocal(String key, Object value) {
     public Object getStrandLocal(String key) {
         return strand.getProperty(key);
     }
+
+    /**
+     * Gets the property map stored in the strand.
+     *
+     * @return property map stored in the strand.
+     */
+    public Map<String, Object> getStrandLocalProperties() {
```
pcnfernando, line 129: FYI; When a new strand is created, there are instances where we need to propagate the property map. For instance, during async invocation(https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java#L96), we do not pass the parent strand to the scheduler. Hence there's no means of propagating the properties. Unless we expose the property map so that it can be passed during async invocation, we need to extract the properties calling getStrandLocal() passing each key.

manuranga: @pcnfernando I am afraid StrandLocal will become a kitchen sink people will through everything at. Can we please take a look the reason you needed this in the async invocation. Shall we schedule a meeting.

pcnfernando: Had a chat with @manuranga and decided not to expose the property map, but to create a submap with needed properties to propagate. For the moment, we'll be passing transaction context alone. @chamil321, please  note

gimantha (approved):

grainier (approved):

PR #29056, opened 2021-03-08
Add support for reading configs from multiple providers 

warunalakshitha:
## Purpose
$subject

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BArray.java
```diff
@@ -202,6 +202,25 @@
      */
     long[] getIntArray();
 
+    /**
+     * Get a copy of boolean array.
+     * @return boolean array
+     */
+    boolean[] getBooleanArray();
```
warunalakshitha, line 209: Some apis were missing from BArray.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/launch/LaunchUtils.java
```diff
@@ -129,17 +126,13 @@ private static void loadConfigurations(Map<String, String> configArgs, String co
...
     }
 
-    public static void initConfigurableVariables(Path filePath, Map<Module, VariableKey[]> configurationData) {
-        try {
-            ConfigTomlParser.populateConfigMap(filePath, configurationData);
-        } catch (TomlException exception) {
-            throw ErrorCreator.createError(StringUtils.fromString(exception.getMessage()));
-        }
+    public static void initConfigurableVariables(Path configFilePath, Map<Module, VariableKey[]> configurationData) {
+        ConfigMap.initialize(configFilePath, configurationData);
```
HindujaB, line 130: As discussed offline, is it possible to move the path parameter to `ConfigTomlProvider`?
Because, the file path is specific for only toml files. 
This may need some changes in Testerina.

warunalakshitha: I try to move this to ConfigTomlProvider, then we need to rely only on env variables. So we cannot run tests in parallel. Because of that, I have kept the old behavior.

manuranga (approved):

PR #29058, opened 2021-03-08
Fix completion issues

### LINKED ISSUE #28665: Suggest only record types as the type of an included record parameter
**Description:**
The type reference in an included record parameter can only refer to a record type defined by a record typedesc. Can we suggest only record types after the `*` in a parameter?

**Steps to reproduce:**
```ballerina
function foo(*F<cursor>) {
    
}

type Foo record {|
    int id;
    string name;
|};
```

Currently doesn't seem to have any filtering.

included_record_param_1

**Affected Versions:**
slalpha2

### LINKED ISSUE #29070: Invalid context identified for suggestions
**Description:**
Consider the following source snippets

```
while true {
} <cursor>

foreach ... {
} <cursor>

do {
} <cursor>

lock {
} <cursor>

match ... {
} <cursor>
```

For all the cursor positions, the contexts are identified incorrectly, hence the `on fail ` clause is not suggested

**Affected Versions:**
sl-alpha2 at least

nadeeshaan:
1. Add a new common service snippet
2. Use the semantic API to capture the type of a field access expression

Fixes #29070 
Fixes #28665 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/RightArrowActionNodeContext.java
```diff
@@ -55,34 +50,14 @@ public RightArrowActionNodeContext(Class<T> attachmentPoint) {
...
-                return completionItems;
-        }
-
-        Optional<Symbol> filteredEntry = visibleSymbols.stream().filter(predicate).findAny();
-
-        if (filteredEntry.isEmpty()) {
-            return completionItems;
-        }
-        if (SymbolUtil.isClient(filteredEntry.get())) {
+        Optional<TypeSymbol> expressionType = context.currentSemanticModel().orElseThrow().type(node);
```
IMS94, line 53: Instead of throwing, we can do `flatMap()` as below:
```suggestion
        Optional<TypeSymbol> expressionType = context.currentSemanticModel().flatMap(semanticModel -> semanticMode.type(node));
```
WDYT? IMO, this approach is much cleaner.

IMS94 (approved):
LGTM! Except for a minor suggestion.

rasika (approved):

PR #29082, opened 2021-03-09
[LS] Fix 'fix return type' code action inserting invalid type names

### LINKED ISSUE #28855: Generated suggestion is not formatted properly
**Description:**
Check the below example. I used the suggestion to add a return type to the function. But the resulting code (the function signature) is not formatted.

Before applying the suggestion.
Screen Shot 2021-02-24 at 12 06 01 AM

After applying the suggestion.
Screen Shot 2021-02-24 at 12 06 09 AM

**Steps to reproduce:**

**Affected Versions:**
Ballerina Swan Lake Alpha 2

### LINKED ISSUE #29100: [LS] Create variable code action (with typeguard/check) adds the same import twice
**Description:**
When the created variable is from a type coming from another module, and if that import is not already present, it adds the same import twice for `create variable and check error` option and `create variable and type guard` option.

Screenshot from 2021-03-10 12-05-38

**Steps to reproduce:**
See description

**Affected Versions:**
SLA 2

IMS94:
## Purpose
- $subject and improve code action's support for imports
- Fix duplicate imports being added by `create variable and typeguard/check` code actions which requires importing an external module.

Fixes #28855
Fixes #29100

## Approach
Updated to use `CodeActionUtil.getPossibleTypes()` method to determine return type and handle imports. Updated the `FixReturnTypeCodeAction` to use diagnostic properties to determine expected type.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/ErrorHandleInsideCodeAction.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/ErrorHandleInsideCodeAction.java
@@ -89,7 +89,8 @@ public class ErrorHandleInsideCodeAction extends CreateVariableCodeAction {
         }
 
         edits.add(createVarTextEdits.edits.get(0));
-        edits.addAll(createVarTextEdits.imports);
+        // Add all the import text edits excluding duplicates
+        createVarTextEdits.imports.stream().filter(edits::contains).forEach(edits::add);
         return Collections.singletonList(AbstractCodeActionProvider.createQuickFixCodeAction(commandTitle, edits, uri));
     }
 }
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
@@ -190,11 +190,11 @@ public class CodeActionUtil {
      * Returns a list of possible types for this type descriptor.
      *
      * @param typeDescriptor {@link TypeSymbol}
-     * @param edits          a list of {@link TextEdit}
+     * @param importEdits    a list of import {@link TextEdit}
      * @param context        {@link CodeActionContext}
      * @return a list of possible type list
      */
-    public static List<String> getPossibleTypes(TypeSymbol typeDescriptor, List<TextEdit> edits,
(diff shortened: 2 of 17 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/ErrorHandleInsideCodeAction.java
```diff
@@ -89,7 +89,8 @@ public int priority() {
         }
 
         edits.add(createVarTextEdits.edits.get(0));
-        edits.addAll(createVarTextEdits.imports);
+        // Add all the import text edits excluding duplicates
+        createVarTextEdits.imports.stream().filter(edits::contains).forEach(edits::add);
```
nadeeshaan, line 93: What if we filter the duplicate imports before reaching this point, within `getPossibleTypes`? Otherwise we will have to do this filtering multiple places

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/ErrorHandleInsideCodeAction.java
IMS94: @nadeeshaan I had a look at the possibility and seems like we cannot do that. Here, we rely on 2 methods to get 
1. create variable text edit -> `CreateVariableOut createVarTextEdits = getCreateVariableTextEdits(range, positionDetails, typeDescriptor.get(), context);` line
2. type guard text edit -> `List<TextEdit> edits = CodeActionUtil.getTypeGuardCodeActionEdits(createVarTextEdits.name, range, unionType, context);` line

And we combine both to come up with the text edits for `create variable and typeguard` action. The above 2 methods return the text edit required for the import statement separately (and may be calling `getPossibleTypes()` twice internally.

Therefore, seems like we have to process this manually as we rely on utils of another 2 atomic code action outputs to provide this code action. WDYT?

nadeeshaan: Okay, then we might need to revisit the implementation later. Since this is acceptable now, let's proceed. +1

nadeeshaan (approved):

PR #29116, opened 2021-03-10
Implement a mechanism for cancelling stale compilation invocations

pubudu91:
## Purpose
This PR implements a mechanism for canceling stale compilation invocations when multiple compilation requests are made.

## Approach
- Introduced a `CompilationRequest` entity which represents and uniquely identifies a compilation invocation.
- Add a per-compiler-context compilation request manager for accepting compilation requests from multiple sources.
- When invoking a compilation, a compilation request is submitted to the request manager. 
- By the time that particular invocation reaches the point where it's about to begin the compilation, it will check with the request manager whether the compilation request is still valid. 
- If it has become stale, the request manager will throw an exception. 
- The one attempting to invoke the compilation will then handle this exception and set the compilation state of that module context to `CANCELLED_COMPILATION`. This particular module context will not recover from this state.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompilationRequestManager.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompilationRequestManager.java
@@ -66,7 +66,7 @@ public class CompilationRequestManager {
      *
      * @param request The request to be checked if it was cancelled
      */
-    public synchronized void throwIfCancelled(CompilationRequest request) {
+    public synchronized void throwIfCancelled(CompilationRequest request) throws StaleCompilationRequestException {
         if (!this.currentCompileRequest.equals(request)) {
             throw new StaleCompilationRequestException(currentCompileRequest, request);
         }
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompilationRequestManager.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompilationRequestManager.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.ballerinalang.compiler;
+
+import org.wso2.ballerinalang.compiler.util.CompilerContext;
+import org.wso2.ballerinalang.compiler.util.StaleCompilationRequestException;
+
+/**
+ * This class is responsible for collecting compilation requests that are coming in from various sources and ensuring
+ * that the latest request is maintained.
+ *
+ * @since 2.0.0
+ */
+public class CompilationRequestManager {
+
+    private static final CompilerContext.Key<CompilationRequestManager> COMPILATION_REQUEST_MANAGER_KEY =
+            new CompilerContext.Key<>();
+
+    private CompilationRequest currentCompileRequest;
+
+    private CompilationRequestManager(CompilerContext context) {
+        context.put(COMPILATION_REQUEST_MANAGER_KEY, this);
+    }
+
+    public static synchronized CompilationRequestManager getInstance(CompilerContext context) {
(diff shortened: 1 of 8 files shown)


nadeeshaan:
It seems in this particular scenario we return an empty optional from the semantic API whenever we occur a canceled compilation. What if we return a flagged semantic model to identify that the associated compilation has been cancelled? This is because there are situations that the semantic model can be empty due to other reasons as well, such as incorrect project update, incorrect source URI and etc. We should be able to distinguish such scenarios from the compilation cancelled scenarios. WDYT?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompilationRequestManager.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * This method accepts a {@link CompilationRequest} instance and compares it against the currently held compilation
+     * request. If this new request is more recent than the currently held request, it will discard the current request
+     * and set this new request as the current request. The provided request must not be null.
+     *
+     * @param request A request to perform a compilation
+     */
+    public synchronized void submit(CompilationRequest request) {
+        if (currentCompileRequest == null || currentCompileRequest.getTimestamp().isBefore(request.getTimestamp())) {
```
rasika, line 57: Do we really need to track the time here? Creating and comparing timestamp will slower down the comparison IMO. An `int` based **incrementing** **counter** would be ideal in this scenario. So you can compare freshness with primitive comparisons such as `>`.

pubudu91:
> It seems in this particular scenario we return an empty optional from the semantic API whenever we occur a canceled compilation. What if we return a flagged semantic model to identify that the associated compilation has been cancelled? This is because there are situations that the semantic model can be empty due to other reasons as well, such as incorrect project update, incorrect source URI and etc. We should be able to distinguish such scenarios from the compilation cancelled scenarios. WDYT?

What would be the behaviour of such a semantic model though? I mean, if a person attempts to use the methods in it? In the current version in this PR, I haven't made that change where an empty optional is returned for semantic models of cancelled compilations. At the moment, it has the default behaviour, which throws an `IllegalStateException`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompilationRequestManager.java
pubudu91: Should be fine I think. It's basically just a `long` comparison

rasika: No, when you calling `Instant.now()` it causes a **sys call** to get current instant from the system UTC clock. Why we are unnecessarily keeping a timestamp?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompilationRequestManager.java
```diff
@@ -0,0 +1,74 @@
...
+ * that the latest request is maintained.
+ *
+ * @since 2.0.0
+ */
+public class CompilationRequestManager {
+
+    private static final CompilerContext.Key<CompilationRequestManager> COMPILATION_REQUEST_MANAGER_KEY =
+            new CompilerContext.Key<>();
+
+    private CompilationRequest currentCompileRequest;
```
rasika, line 35: This should be marked as volatile to avoid L1, L2, L3 cache line memory visibility issue.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/CompilationRequestManager.java
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Given a compilation request, this method will check if it is still valid. If there is a more recent request, this
+     * will throw a {@link StaleCompilationRequestException} to signal to the one initiating this compilation that it
+     * should abort this compilation.
+     *
+     * @param request The request to be checked if it was cancelled
+     */
+    public synchronized void throwIfCancelled(CompilationRequest request) throws StaleCompilationRequestException {
+        if (!this.currentCompileRequest.equals(request)) {
```
rasika, line 70: This `equals()` check is way more costly IMO. You don't need to track the `currentCompileRequest` IMO. You can check for `this.currentCompilationVersion.equals(version)` where `version` is just a primitive `int`(should handle overflowing though).

rasika:
> It seems in this particular scenario we return an empty optional from the semantic API whenever we occur a canceled compilation. What if we return a flagged semantic model to identify that the associated compilation has been cancelled? This is because there are situations that the semantic model can be empty due to other reasons as well, such as incorrect project update, incorrect source URI and etc. We should be able to distinguish such scenarios from the compilation cancelled scenarios. WDYT?

Are we returning empty optional from the semantic API when there's cancelled compilation. @pubudu91 please confirm. IMO that should be handled internally since callee doesn't need to check whether the compilation is cancelled or not.

pubudu91:
Closing this for now.

PR #29167, opened 2021-03-12
Remove restrictions of using `null`

### LINKED ISSUE #28342: Restriction of null to json type
In alpha1, this gets a compile-error:

```
int? x = null;
```

This is a correct program according to the spec, since the spec says:

> the use of null should be restricted to JSON-related contexts

Note this is **should**, and it does not say `json` type.

This is not a helpful restriction. For example, I wanted to copy and paste some data from a JSON file and use it as the initializer of a user-defined record type (to see whether my data-binding would work), but I couldn't because of this.

lasinicl:
Fixes #28342

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -225,17 +225,31 @@ public BType checkType(Location pos,
     }
 
     public boolean isJSONContext(BType type) {
-        if (type.tag == TypeTags.UNION) {
-            return ((BUnionType) type).getMemberTypes().stream().anyMatch(memType -> memType.tag == TypeTags.JSON);
+        switch (type.tag) {
+            case TypeTags.MAP:
+                return isJSONContext(((BMapType) type).constraint);
+            case TypeTags.ARRAY:
+                return isJSONContext(((BArrayType) type).eType);
```
pubudu91, line 232: Shouldn't this only be applicable for `map<json>` and JSON arrays? For example, with this check, types such as `map<json|Foo>` (where `Foo` is an object) would also be considered as a JSON context right?

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1449,7 +1449,7 @@ error.invalid.use.of.experimental.feature=\
   using experimental feature ''{0}''. use ''--experimental'' flag to enable the experimental features
 
 error.invalid.use.of.null.literal=\
-''null'' literal is only supported for ''json''
+''null'' literal is only allowed in a json-related context
```
pubudu91, line 1452: `json-related` -> `JSON-related`. The lower case is only used when referring to the `json` type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
MaryamZi: I think we should clarify what JSON context means. I guess one could argue that even `map<json|Foo>` can be considered a JSON context since it still allows JSON, and similarly something like `map<int|Foo>`.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/nullable/NullableTypeTest.java
```diff
@@ -78,6 +78,51 @@ public void testNilableTypeArrayIteration() {
         validateError(result, 0, "incompatible types: expected '(()|any)', found '(()|any)?'", 33, 19);
     }
 
+    @Test(description = "Test null with basic types")
```
MaryamZi, line 81: Shall we do these assertions in b7a itself?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
lasinicl: Refer https://github.com/ballerina-platform/ballerina-lang/issues/28342#issuecomment-806403455 for clarification regarding the JSON-related context

pubudu91: My concern is not cases where we add elements to maps such as `map<json|Foo>`. I think those count as JSON context. But the above makes the following case to be considered as a JSON context too right? 
```ballerina
map<json|Foo>? m = null;
```

MaryamZi: If we consider `map<json|Foo>` as a JSON context isn't this similar to the original example `int?`?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/nullable/NullableTypeTest.java
```diff
@@ -78,6 +79,22 @@ public void testNilableTypeArrayIteration() {
         validateError(result, 0, "incompatible types: expected '(()|any)', found '(()|any)?'", 33, 19);
     }
 
+    @Test(dataProvider = "dataToTestNullInJSONRelatedContext", description = "Test null in JSON related context")
+    public void testNullInJSONRelatedContext(String functionName) {
+        BRunUtil.invoke(result, functionName);
+    }
+
+    @DataProvider
+    public Object[] dataToTestNullInJSONRelatedContext() {
```
MaryamZi, line 88: We don't need to say "JSON context" anymore, right?

MaryamZi (commented):
Shall we also update the PR title?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/nullable/NullableTypeTest.java
lasinicl: Ah my bad! missed this. Will fix :)

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
pubudu91: Ah now only I saw that @lasinicl has already raised this in the issue. So according to that, use of `null` anywhere is allowed by the compiler I guess? 🤔

pubudu91: ^ Ah nvm my bad, ignore the above. Really should start checking before commenting 😅

pubudu91 (approved):

MaryamZi (approved):

PR #29209, opened 2021-03-15
Fix type inclusions for class symbols and add BIR tests 

### LINKED ISSUE #27035: Add BIR tests for type inclusions in Semantic API
Need to $subject after the test utils are updated to support BIRs of modules.

### LINKED ISSUE #29204: [Semantic API] Type inclusions not working for class symbols
Currently `typeInclusions()` return an empty list for classes which have type inclusions. It seems that info is not getting added to the object type of the class.

pubudu91:
## Purpose
This PR fixes an issue with type inclusions for class symbols and adds BIR tests for type inclusions.

Fix #27035
Fix #29204 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
@@ -616,6 +616,9 @@ class SymbolFinder extends BaseVisitor {
 
     @Override
     public void visit(BLangConstRef constRef) {
+        if (constRef.symbol != null && setEnclosingNode(constRef.symbol.owner, constRef.pkgAlias.pos)) {
+            return;
+        }
         this.symbolAtCursor = constRef.symbol;
     }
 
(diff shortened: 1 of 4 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
```diff
@@ -616,6 +616,9 @@ public void visit(BLangLiteral literalExpr) {
 
     @Override
     public void visit(BLangConstRef constRef) {
+        if (constRef.symbol != null && setEnclosingNode(constRef.symbol.owner, constRef.pkgAlias.pos)) {
+            return;
+        }
         this.symbolAtCursor = constRef.symbol;
```
Kishanthan, line 622: If 619 fails, then at 622 we will assign a null value?

pubudu91: Yes, that would be fine. If `symbolAtCursor` is `null` at the end it'll result in `Optional.empty()` being returned to the user.

nadeeshaan (approved):
LGTM

PR #29226, opened 2021-03-15
[LS] Add completion support for query expressions and query actions

### LINKED ISSUE #24138: Language server support for new query expression clauses
**Description:**
There are a few new syntax added to query expression clauses. We have to provide language server support for those new clauses.

**Examples**

_**Join Clause**_

```ballerina
    DeptPerson[] deptPersonList =
       from var person in personList
       join Department dept in deptList
       on person.id equals dept.id
       select {
           fname : person.fname,
           lname : person.lname,
           dept : dept.name
       };
```

**_Limit Clause_**

```ballerina
    Person[] outputPersonList =
            from var person in personList
            let int newAge = 34
            where person.age == 33
            select {
                   firstName: person.firstName,
                   lastName: person.lastName,
                   age: newAge
            }
            limit 2;
```

**_On-conflict clause_**

```ballerina
    CustomerTable|error customerTable = table key(id, name) from var customer in customerList
        select {
            id: customer.id,
            name: customer.name,
            noOfItems: customer.noOfItems
        }
        on conflict onConflictError;
```

**_Stream generation from query-expr_**

```ballerina
    stream<Person> outputPersonStream = stream from var person in personList
        where person.firstName == "John"
        let int newAge = 34
        select {
            firstName: person.firstName,
            lastName: person.lastName,
            age: newAge
        };
```

### LINKED ISSUE #26860: [Tooling] Code suggestion for query expression not works
**Description:**
Please refer the below code. In that, code completion does not work for query expression.

```ballerina
type Student record {
    readonly id;
    string firstName;
    string lastName;
    int intakeYear;
    float gpa;
};

public function main() {

    Student s1 = { id: 1, firstName: "Michelle", lastName: "Sadler", 
                   intakeYear: 1990, gpa: 3.5 };
    Student s2 = { id: 2, firstName: "Ranjan", lastName: "Fonseka", 
                   intakeYear: 2001, gpa: 1.9 };
    Student s3 = { id: 3, firstName: "Martin", lastName: "Guthrie", 
                   intakeYear: 2002, gpa: 3.7 };
    Student s4 = { id: 4, firstName: "George", lastName: "Fernando", 
                   intakeYear: 2005, gpa: 4.0 };
    Student[] studentList = [s1, s2, s3, s4];

    Student[] duplicateStdList = [s1, s2, s1];

    error onConflictError = error("Key Conflict",
                                  message = "cannot insert report");                              

}

```

### LINKED ISSUE #28202: Query action is not suggested
**Description:**
doAction

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

IMS94:
## Purpose
$subject

Fixes #28202 
Fixes #24138 
Fixes #26860

## Approach
- Update each `ClauseNode` context to provide completions under their contexts for query expressions.
- Moved query expression completion tests to a separate test

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/JoinClauseNodeContext.java
```diff
@@ -115,10 +129,65 @@ private boolean onSuggestBindingPattern(BallerinaCompletionContext context, Join
...
+            Node nodeAtCursor = context.getNodeAtCursor();
+
+            /*
+             * Captures:
+             * (1) join var varName i<cursor>
+             * (2) join var varName i<cursor> expression
+             */
+            if (nodeAtCursor.kind() == SyntaxKind.SIMPLE_NAME_REFERENCE) {
+                SimpleNameReferenceNode nameReferenceNode = (SimpleNameReferenceNode) nodeAtCursor;
+                return node.expression().textRange().startOffset() + 1 == cursor &&
```
nadeeshaan, line 152: Say, when at the cursor the particular letter typed is `n`. Then this logic is invalidated right? If so, I think we should drop the starts with check here.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/JoinClauseNodeContext.java
```diff
@@ -115,10 +129,65 @@ private boolean onSuggestBindingPattern(BallerinaCompletionContext context, Join
...
+    private boolean onSuggestOnKeyword(BallerinaCompletionContext context, JoinClauseNode node) {
+        int cursor = context.getCursorPositionInTree();
+
+        if (node.typedBindingPattern().isMissing() ||
+                node.typedBindingPattern().isMissing() ||
+                node.inKeyword().isMissing() ||
+                node.expression().isMissing()) {
+            return false;
+        } else if (!node.joinOnCondition().onKeyword().isMissing()) {
+            return false;
```
nadeeshaan, line 170: Is is possible to combine the above two conditions?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/JoinClauseNodeContext.java
```diff
@@ -115,10 +129,65 @@ private boolean onSuggestBindingPattern(BallerinaCompletionContext context, Join
...
+            /*
+             * Captures:
+             * (1) join var varName in expr o<cursor>
+             * (2) join var varName in expr o<cursor> expr
+             */
+            if (nodeAtCursor.kind() == SyntaxKind.SIMPLE_NAME_REFERENCE) {
+                SimpleNameReferenceNode nameReferenceNode = (SimpleNameReferenceNode) nodeAtCursor;
+                return node.expression().textRange().endOffset() == cursor &&
+                        nameReferenceNode.textRange().endOffset() == cursor &&
+                        ItemResolverConstants.ON.startsWith(nameReferenceNode.name().text());
```
nadeeshaan, line 187: Same as in the previous comment

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/OnClauseNodeContext.java
```diff
@@ -89,9 +97,34 @@ private boolean onSuggestEqualsKeyword(BallerinaCompletionContext context, OnCla
...
         Token equalsKeyword = node.equalsKeyword();
 
-        return !lhs.isMissing() && cursor > lhs.textRange().endOffset()
-                && (rhs.isMissing() || cursor < rhs.textRange().startOffset())
-                && (equalsKeyword.isMissing() || (!equalsKeyword.isMissing()
-                && cursor < equalsKeyword.textRange().startOffset()));
+        if (lhs.isMissing()) {
+            return false;
+        } else if (!equalsKeyword.isMissing()) {
+            return false;
```
nadeeshaan, line 103: Can we merge the two conditions?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/WhereClauseNodeContext.java
```diff
@@ -57,12 +58,21 @@ public WhereClauseNodeContext() {
...
+
         return completionItems;
     }
 
     @Override
     public boolean onPreValidation(BallerinaCompletionContext context, WhereClauseNode node) {
         return !node.whereKeyword().isMissing();
     }
+
+    private boolean cursorAtTheEndOfExpression(BallerinaCompletionContext context, WhereClauseNode node) {
```
nadeeshaan, line 70: Shall we move this method to `IntermediateClauseNodeContext`? If an extending context needs to have a specific logic, then the particular context can override this. WDYT?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/FromClauseNodeContext.java
```diff
@@ -134,4 +135,13 @@ private boolean onBindingPatternContext(BallerinaCompletionContext context, From
                 (cursor > typeDescriptor.textRange().endOffset() && cursor < inKeyword.textRange().startOffset()))
                 && (bindingPattern.isMissing() || bindingPattern.textRange().endOffset() >= cursor);
     }
+
+    protected boolean cursorAtTheEndOfClause(BallerinaCompletionContext context, FromClauseNode node) {
+        if (node.expression() == null || node.expression().isMissing()) {
```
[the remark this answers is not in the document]
pubudu91: This is a syntax node right? If so, if it's becoming null, that should be fixed from the parser side I think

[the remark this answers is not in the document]
IMS94: I think we don't get a null here. I just reused the logic that was there. Will remove the null check.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/WhereClauseNodeContext.java
IMS94: I thought the same. But the issue is, `IntermediateClauseNode` doesn't have the `expression()` method. If we move this to parent class, we will have to do `instanceof` checks for each sub type.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/JoinClauseNodeContext.java
IMS94: I have a question regarding your suggestion. If the user types `n` (anything other than `i`), IMO `in` is not a suggestion. Even if we return the completion item `in`, VSCode will filter that out because the letter already typed is starting from `n`. WDYT?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/JoinClauseNodeContext.java
IMS94: Done!

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/OnClauseNodeContext.java
IMS94: Sure!

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/WhereClauseNodeContext.java
nadeeshaan: If we manipulate only the expression what if we pass the expression as an argument?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/JoinClauseNodeContext.java
nadeeshaan: I think, in this case, still the `in` keyword will be shown. Since the editor do the distance based filtering it will suggest the nearest matching entry.

IMS94: I just tested that. Seems like returning `in` while the user has types `o` doesn't show `in` as a suggestion in VSCode.

IMS94: Anyway the `startsWith` check is redundant as we check if the cursor is at the end of the expression. Therefore, removed that.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/JoinClauseNodeContext.java
IMS94: Removed the `startsWith()` check. See above thread for more details.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/WhereClauseNodeContext.java
IMS94: Refactored to get the expression from child classes and do the above check within `IntermediateClauseNodeContext`.

nadeeshaan (approved):

Issue #29158, opened 2021-03-12
Use the service name if available in Observability Data


nadundesilva:
**Description:**
Currently when the `display` annotation is not present, we use `<package_name>_svc_<counter>` as the service name in Observability Data.

However as the Compiler Team is going to inject the service name into the BIR, we need to use the service name itself whenever it is available. This will improve the experience in Observability as it is much easier to map data to the correct service by default.

Related to: https://github.com/ballerina-platform/ballerina-lang/issues/27402

PR #29228, opened 2021-03-16
Get basepath from service declaration symbol for observability

madushajg:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/29158

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmObservabilityGen.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmObservabilityGen.java
@@ -121,7 +121,7 @@ class JvmObservabilityGen {
     private int defaultServiceIndex;
 
     private final Map<Object, BIROperand> compileTimeConstants;
-    private final Map<Name, List<String>> svcAttachPoints;
+    private final Map<Name, String> svcAttachPoints;
 
     JvmObservabilityGen(PackageCache packageCache, SymbolTable symbolTable) {
         this.compileTimeConstants = new HashMap<>();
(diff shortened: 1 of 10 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmObservabilityGen.java
```diff
@@ -170,8 +175,13 @@ void instrumentPackage(BIRPackage pkg) {
                     }
                 }
                 if (serviceName == null) {
-                    serviceName = pkg.packageID.orgName.value + "_" + pkg.packageID.name.value + "_svc_" +
-                            defaultServiceIndex++;
+                    List<String> attachPoint = this.svcAttachPoints.get(typeDef.name);
+                    if (attachPoint != null) {
+                        serviceName = String.join("/", attachPoint);
```
nadundesilva, line 180: Don't we need to add a "/" here to the beginning ?
```suggestion
                        serviceName = "/" + String.join("/", attachPoint);
```

Also what happens when the user specify the base path as `/` or `.` ?

nadundesilva: Also what happens when the user specify the base path as `/` or `.` in the code ?

madushajg: The `/` prefix is not appended with `@display` annotation as well.
Do we need to add it here?

nadundesilva: I think we should.
Difference is that
* In `@display` annotation, we give the name as `testSvcFour`
* In basepath we give it as `/test/serviceThree`
I think we need to show the basepath that they give if we can. (check my second comment also)

pubudu91: `.` is not allowed as the base path right? At the BIR level, the behaviour of `attachPoint` in `BIRServiceDeclaration` is as follows:
- If it's just `/`, it will be an empty list
- For `/foo/bar` style base path, it will return a list containing the strings `"foo"` and `"bar"`
- If a base path is not specified, `attachPoint` will be `null`.

If a literal is specified instead of a base path, `attachPoint` will be `null` and the `attachPointLiteral` will contain the literal specified.

nadundesilva: Then adding "/" as a prefix should fix it I guess. But then we need to handle `attachPointLiteral` literal case also in this logic.

Maybe we can generate these correctly and store in `svcAttachPoints` @madushajg  ?

madushajg: Fixed with eef9a7fa417e0d3b9f796dcdb0b31e4f7cd704f9

nadundesilva (approved):
LGTM

pubudu91 (approved):

PR #29291, opened 2021-03-17
Fix NPE in FunctionTypeSymbol

pubudu91:
## Purpose
Fixes the unhandled case for the any function typedesc in `FunctionTypeSymbol`.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionTypeSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionTypeSymbol.java
@@ -25,8 +25,10 @@ import io.ballerina.compiler.api.symbols.TypeDescKind;
 import io.ballerina.compiler.api.symbols.TypeSymbol;
 import org.ballerinalang.model.symbols.SymbolKind;
 import org.wso2.ballerinalang.compiler.semantics.model.symbols.BInvokableTypeSymbol;
+import org.wso2.ballerinalang.compiler.semantics.model.symbols.Symbols;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BType;
 import org.wso2.ballerinalang.compiler.util.CompilerContext;
+import org.wso2.ballerinalang.util.Flags;
 
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionTypeSymbol.java
```diff
@@ -60,8 +60,13 @@ public BallerinaFunctionTypeSymbol(CompilerContext context, ModuleID moduleID,
     @Override
     public List<ParameterSymbol> parameters() {
         if (this.requiredParams == null) {
-            SymbolFactory symbolFactory = SymbolFactory.getInstance(this.context);
+            // Becomes null for the function typedesc.
+            if (this.typeSymbol.params == null) {
+                this.requiredParams = Collections.emptyList();
+                return this.requiredParams;
```
MaryamZi, line 66: For the compiler and runtime we decided on not setting an empty list for the `function` type descriptor to differentiate between `function` and `function ()`. Shouldn't we do the same for the semantic API? Should this method return `Optional<List<ParameterSymbol>>`?

Also, for the compiler and the runtime we also have a flag `ANY_FUNCTION` that can be used to check if it is `function`. Would the users of the semantic API be able to figure this out directly given that we are returning an empty list here?

MaryamZi (commented):
`io.ballerina.compiler.api.impl.symbols.BallerinaFunctionTypeSymbol#signature` also needs to be updated?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionTypeSymbol.java
pubudu91: Need to check that with the others and see. I think having an empty list should be ok. I don't think there'd be any requirement to distinguish between `function` and full form of the function typedesc. Usually the users are interested in details such as the params, return type, meta data etc. The only place this would be relevant AFAIU is the signature.

MaryamZi: Shall we create an issue to revisit this since `function` is a bit different from other function types?

MaryamZi (approved):

PR #29298, opened 2021-03-17
Disallow modeling vars of nilable function types as invokable symbols

pubudu91:
## Purpose
Due to code remaining from nil-lifting days, a variable of type `T?` where `T` is a function type is modeled as an invokable symbol. This PR removes this behaviour and models such vars as regular var symbols. 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
@@ -389,6 +389,7 @@ public enum DiagnosticErrorCode implements DiagnosticCode {
     EXPRESSION_DOES_NOT_SUPPORT_FIELD_ACCESS("BCE2648", "expression.does.not.support.field.access"),
     CANNOT_USE_TYPE_INCLUSION_WITH_MORE_THAN_ONE_OPEN_RECORD_WITH_DIFFERENT_REST_DESCRIPTOR_TYPES("BCE2649",
             "cannot.use.type.inclusion.with.more.than.one.open.record.with.different.rest.descriptor.types"),
+    METHOD_CALL_NOT_ALLOWED_FOR_FIELD("BCE2650", "method.call.expr.not.allowed"),
 
     // Error codes related to iteration.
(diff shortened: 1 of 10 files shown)


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1690,3 +1690,6 @@ error.invalid.readonly.field.type=\
 error.continue.not.allowed=continue not allowed here
 
 error.break.not.allowed=break not allowed here
+
+error.method.call.expr.not.allowed=\
+  method call expression not allowed on field ''{0}'': expected a function type, but found ''{1}''
```
MaryamZi, line 1695: Shouldn't the error be something like

```
invalid method call expression: expected a function type, found ''{0}''
```

When we say "on field", it kinda feels like we are trying to access the field and then call a method on it, but we get this error when we are actually trying to call the field itself?

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
```diff
@@ -389,6 +389,7 @@
     EXPRESSION_DOES_NOT_SUPPORT_FIELD_ACCESS("BCE2648", "expression.does.not.support.field.access"),
     CANNOT_USE_TYPE_INCLUSION_WITH_MORE_THAN_ONE_OPEN_RECORD_WITH_DIFFERENT_REST_DESCRIPTOR_TYPES("BCE2649",
             "cannot.use.type.inclusion.with.more.than.one.open.record.with.different.rest.descriptor.types"),
+    METHOD_CALL_NOT_ALLOWED_FOR_FIELD("BCE2650", "method.call.expr.not.allowed"),
```
MaryamZi, line 392: Shall we update this so that the code and the key match?

MaryamZi (approved):

PR #29302, opened 2021-03-17
Add xml:getDescendants() method

### LINKED ISSUE #28901: Add xml:getDescendants langlib method
The tracking issue for https://github.com/ballerina-platform/ballerina-spec/issues/701

pcnfernando:
Fixes #28901

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.xml/src/main/java/org/ballerinalang/langlib/xml/GetDescendants.java
```diff
@@ -0,0 +1,44 @@
...
+public class GetDescendants {
+
+    @SuppressWarnings("unchecked")
+    public static BXml getDescendants(BXml element) {
+        if (!IsElement.isElement(element)) {
+            throw BLangExceptionHelper.getRuntimeException(RuntimeErrors.XML_FUNC_TYPE_ERROR,
+                    "getDescendants", "element");
+        }
+        try {
+            return (BXml) element.descendants();
```
rdhananjaya, line 38: I don't think we need this try catch as `descendants` method doesn't throw

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibXMLTest.java
```diff
@@ -281,6 +284,36 @@ public void testSelectingTextFromXml() {
         BRunUtil.invoke(compileResult, "testSelectingTextFromXml");
     }
 
+    @Test
+    public void testGetDescendants() {
+        String[] expectedDescendants = {"<CD>\n" +
```
rdhananjaya, line 289: Shall we add more tests, which include other xml types, (comments and processing instructions).
Also shall we try to assert them in the Ballerina itself.

pcnfernando: Added and refactored in 2b6c68b820bb0c13ce59393ce873213c381a4224

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlValue.java
```diff
@@ -186,6 +186,15 @@ protected void addDescendants(List<BXml> descendants, XmlItem currentElement, Li
         }
     }
 
+    protected void addDescendants(List<BXml> descendants, XmlItem currentElement) {
+        for (BXml child : currentElement.getChildrenSeq().getChildrenList()) {
+            if (child.getNodeType() == XmlNodeType.ELEMENT) {
```
rdhananjaya, line 191: I think this should include non-elements too. Did you confirmed in the spec that this only include elements?

pcnfernando: Yes, the spec was rephrased to remove ambiguity. https://github.com/ballerina-platform/ballerina-spec/issues/701
It could include non-elements

pcnfernando: eg: 
```ballerina
<A>
   <B>
      <C>Hi</C>
   </B>
</A>
```

calling descendants on this results in the following sequence.
```ballerina
<B>
  <C>Hi</C>
</B>

<C>Hi</C>
```

rdhananjaya: This test case feels wrong to me. https://github.com/ballerina-platform/ballerina-lang/pull/29302/files#diff-17b929237f4f118a94e336c2e1c226975c3ec2402973ec77a078689921db9cc7R530
The processing instruction is missing from the result

pcnfernando: This is because the processing instruction is an immediate child of `bs` which gets filtered as it's not an element.
If we had something like below where it's not an immediate child. It will be included in the sequence along with its parent element and then filtered out moving down the elements.
```ballerina
<bs><bk><?xml-stylesheet type=""text/xsl""?><t><en><!-- english --><txt>Everyday Italian</txt></en></t></bk></bs>
```

getDescendants() result:
```ballerina
<bk>
  <?xml-stylesheet type="text/xsl"?>
  <t>
    <en>
      <!-- english -->
      <txt>Everyday Italian</txt>
    </en>
  </t>
</bk>

<t>
  <en>
    <!-- english -->
    <txt>Everyday Italian</txt>
  </en>
</t>
<en>
  <!-- english -->
  <txt>Everyday Italian</txt>
</en>
<txt>Everyday Italian</txt>
```

rdhananjaya: Function documentation says `The descendants of an element are the children of the element together with, for each of those children that is an element, ...`
Here the (in the test) we need to take all the children of `bk` + apply the same function to each element child.
So children of `bk` includes non elements.

pcnfernando: Updated in f1161288d594fxc24acddd7fdf86134030619d0fb

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlSequence.java
```diff
@@ -393,6 +393,18 @@ public XmlValue descendants(List<String> qnames) {
...
+        List<BXml> descendants = new ArrayList<>();
+        for (BXml child : children) {
+            if (child.getNodeType() == XmlNodeType.ELEMENT) {
+                XmlItem element = (XmlItem) child;
+                descendants.add(element);
+                addDescendants(descendants, element);
+            }
+        }
+        return new XmlSequence(descendants);
+    }
```
rdhananjaya, line 406: Consider following:
```ballerina
    xml x = xml `<a><b>helo</b></a>`;
    xml b = x/*;
    if (b is xml:Element) {
        xml c = b.getDescendants();
    }
 ```

`c` should only contain `hello`, at the moment it's `<b>helo</b>helo`

rdhananjaya: `b` is a sequence with a single element, hence it's a xml:Element.
calling `getDescendants` on a xml:Element should only return children of that element.

pcnfernando: Handled with a4dbd75508baab046999c18b524530019c699be9

rdhananjaya (approved):

rdhananjaya (approved):

Issue #29149, opened 2021-03-12
NPE from Semantic API when getting completion suggestions


rasika:
**Description:**
There NPE thrown from Semantic API getting completion suggestion when there's an semantically invalid object initialization before.
 
**Source Code:**
```ballerina
public class Foo {
    bar:Unknown c = new();
    f<cursor>
}
```
**Error:**
```
[Error - 8:58:22 PM] Operation 'text/completion' failed! { [3:6], error: 'java.lang.NullPointerException'}
java.lang.NullPointerException
	at io.ballerina.compiler.api.impl.symbols.BallerinaObjectFieldSymbol.<init>(BallerinaObjectFieldSymbol.java:58)
	at io.ballerina.compiler.api.impl.symbols.BallerinaClassFieldSymbol.<init>(BallerinaClassFieldSymbol.java:44)
	at io.ballerina.compiler.api.impl.SymbolFactory.createClassFieldSymbol(SymbolFactory.java:326)
	at io.ballerina.compiler.api.impl.SymbolFactory.getBCompiledSymbol(SymbolFactory.java:144)
	at io.ballerina.compiler.api.impl.BallerinaSemanticModel.addToCompiledSymbols(BallerinaSemanticModel.java:369)
	at io.ballerina.compiler.api.impl.BallerinaSemanticModel.visibleSymbols(BallerinaSemanticModel.java:114)
	at org.ballerinalang.langserver.contexts.AbstractDocumentServiceContext.visibleSymbols(AbstractDocumentServiceContext.java:113)
	at org.ballerinalang.langserver.completions.providers.AbstractCompletionProvider.getTypeItems(AbstractCompletionProvider.java:209)
	at org.ballerinalang.langserver.completions.providers.context.ObjectFieldNodeContext.getClassBodyCompletions(ObjectFieldNodeContext.java:107)
	at org.ballerinalang.langserver.completions.providers.context.ObjectFieldNodeContext.getCompletions(ObjectFieldNodeContext.java:66)
	at org.ballerinalang.langserver.completions.providers.context.ObjectFieldNodeContext.getCompletions(ObjectFieldNodeContext.java:47)
	at org.ballerinalang.langserver.completions.util.CompletionUtil.route(CompletionUtil.java:99)
	at org.ballerinalang.langserver.completions.util.CompletionUtil.getCompletionItems(CompletionUtil.java:56)
[...]

PR #29383, opened 2021-03-22
Fix issue in creating field symbols

pubudu91:
## Purpose
This PR fixes an issue there was in creating field symbols when the field's type is an undefined type. The issue was if there was an issue when defining the field, the compiler sets the type of the field as `semanticError` and later on, when creating the type, we filtered out those fields. This caused the NPE when creating the relevant field symbols in semantic API because when it tries to look up the field by name, it can't find such a field, hence it returns null.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/incompletesource/InvalidFieldTypeTest.java
+++ b/tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/incompletesource/InvalidFieldTypeTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
  *
  * WSO2 Inc. licenses this file to you under the Apache License,
  * Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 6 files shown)


nadeeshaan:
Shall we create an issue and tag in the PR?

tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/incompletesource/InvalidFieldTypeTest.java
```diff
@@ -0,0 +1,110 @@
+/*
+ * Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
nadeeshaan, line 2: Fix year

nadeeshaan (changes requested):

pubudu91:
> Shall we create an issue and tag in the PR? Also, since we do not filter the fields now, does this means we are getting fields where their types can be `SemanticError`?

Issue is already linked (#29149). And yes, that can happen now.

nadeeshaan:
> > Shall we create an issue and tag in the PR? Also, since we do not filter the fields now, does this means we are getting fields where their types can be `SemanticError`?
> 
> Issue is already linked (#29149). And yes, that can happen now.

Showing the SemanticError as the type to the user might be misleading though. If we don't have a solution for that better if we can track it as a known issue

pubudu91:
> > > Shall we create an issue and tag in the PR? Also, since we do not filter the fields now, does this means we are getting fields where their types can be `SemanticError`?
> > 
> > 
> > Issue is already linked (#29149). And yes, that can happen now.
> 
> Showing the SemanticError as the type to the user might be misleading though. If we don't have a solution for that better if we can track it as a known issue

We can skip the fields with semantic error for its type at the semantic API if needed. Either have to skip, or still make it visible with the error. So basically have to decide between having missing field symbols vs. having field symbols whose type is semantic error

grainier (approved):

PR #29386, opened 2021-03-22
Improve symbol lookup for query expressions

### LINKED ISSUE #29262: Symbol lookup for incomplete query expression misses symbols
**Description:**
$subject.

symbolsForOrderBy assertion in testSymbolLookupInQuery test commented out with c4c93cd in PR #29181 due to some symbols are not visible in that scope. 

when enabling the following top-level construct resides inside the function def can also be taken out.

https://github.com/ballerina-platform/ballerina-lang/blob/02fbdd0e6027e437241e5c1953789fe01faff5eb/tests/ballerina-compiler-api-test/src/test/resources/test-src/symbol_lookup_in_query.bal#L40-L44

pcnfernando:
Fixes #29262

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
- [] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Ballerina By Examples


mohanvive (dismissed):
Approving the PR, but it is a workaround/temp fix.

mohanvive:
It seems query expressions related latest test cases are failing with this PR, hence I will be moving this to next milestone to avoid unexpected issues. Let's verify this fix for the next milestone.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/PositionUtil.java
```diff
@@ -81,4 +81,19 @@ static boolean withinRange(LineRange specifiedRange, Location nodePosition) {
...
+    static boolean withinRightInclusive(LinePosition cursorPos, Location symbolPosition) {
+        int startLine = symbolPosition.lineRange().startLine().line();
+        int startColumn = symbolPosition.lineRange().startLine().offset();
+        int endLine = symbolPosition.lineRange().endLine().line();
+
+        int cursorLine = cursorPos.line();
+        int cursorColumn = cursorPos.offset();
+        return (startLine < cursorLine && endLine > cursorLine)
+                || (startLine < cursorLine && endLine == cursorLine)
+                || (startLine == cursorLine && endLine > cursorLine)
```
IMS94, line 95: Is this check correct? What if the cursor is on the start line of the node, but before the start column? (Or this is correct based on the writing pattern?)

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/PositionUtil.java
```diff
@@ -81,4 +81,19 @@ static boolean withinRange(LineRange specifiedRange, Location nodePosition) {
 
         return specifiedEndColumn <= nodeEndColumn;
     }
+
+    //  todo to be removed once https://github.com/ballerina-platform/ballerina-lang/discussions/28983 is sorted
+    static boolean withinRightInclusive(LinePosition cursorPos, Location symbolPosition) {
```
pubudu91, line 86: Shall we rename this to `nodePosition`? `symbolPosition` would mean something slightly different from node position.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
```diff
@@ -1220,17 +1223,16 @@ public void visit(BLangMatch.BLangMatchTypedBindingPatternClause patternClauseNo
 
     @Override
     public void visit(BLangJoinClause joinClause) {
-        if (!PositionUtil.withinBlock(this.linePosition, joinClause.getPosition())) {
-            return;
+        if (PositionUtil.withinRightInclusive(this.linePosition, joinClause.getPosition())) {
```
nadeeshaan, line 1226: Is it possible to use early return, to be consistent with the other logics?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
```diff
@@ -1220,17 +1223,16 @@ public void visit(BLangMatch.BLangMatchTypedBindingPatternClause patternClauseNo
...
         }
-        this.scope = joinClause.env;
-        this.acceptNode(joinClause.collection, joinClause.env);
-        this.acceptNode((BLangNode) joinClause.onClause, joinClause.env);
     }
 
     @Override
     public void visit(BLangOnClause onClause) {
-        if (PositionUtil.withinBlock(this.linePosition, onClause.getPosition())) {
+        if (PositionUtil.withinRightInclusive(this.linePosition, onClause.getPosition())) {
```
nadeeshaan, line 1235: Use early return here. We can avoid the nested if check

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/PositionUtil.java
```diff
@@ -81,4 +81,19 @@ static boolean withinRange(LineRange specifiedRange, Location nodePosition) {
...
+        int startLine = symbolPosition.lineRange().startLine().line();
+        int startColumn = symbolPosition.lineRange().startLine().offset();
+        int endLine = symbolPosition.lineRange().endLine().line();
+
+        int cursorLine = cursorPos.line();
+        int cursorColumn = cursorPos.offset();
+        return (startLine < cursorLine && endLine > cursorLine)
+                || (startLine < cursorLine && endLine == cursorLine)
```
nadeeshaan, line 94: Shouldn't we consider the start column as well? There can be scenarios that the cursor can be in the same line but the offset is less than the start column

pcnfernando: Updated in 54acb97631c9572295a6e6dcc3a00b66e5add39f

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/PositionUtil.java
pcnfernando: Updated in 54acb97631c9572295a6e6dcc3a00b66e5add39f

IMS94 (approved):
Approved with a minor comment for completions within a `join clause`. We can ignore that comment for now if needed.

language-server/modules/langserver-core/src/test/resources/completion/query_expression/config/query_expr_ctx_join_clause_config5.json
```diff
@@ -423,6 +423,30 @@
...
+    {
+      "label": "c3",
+      "kind": "Variable",
+      "detail": "Customer",
+      "sortText": "B",
+      "insertText": "c3",
+      "insertTextFormat": "Snippet"
+    },
+    {
+      "label": "customer",
```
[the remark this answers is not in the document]
nadeeshaan: IIRC, we have this same issue in some other places as well. We need to find a common solution to this. Hence I think this is okay for now.

nadeeshaan (approved):

PR #29565, opened 2021-03-24
[LS] Update create function to return new expression parameters and fix several documentation issues

### LINKED ISSUE #29225: Create function code action generates a new expr as the return value of the generated function
Consider the following:
```ballerina
object {
        string name;
        int age;
    } person = foo();
```
The above generates the following:
```ballerina
function foo() returns object {string name;int age} {
    
    return new();
}
```
Since the return typedesc is an object typedesc, I think we should refrain from adding a new expr as the return value. How about generating an object constructor expr instead?

### LINKED ISSUE #29397: [LS] Update document code action doesn't work
See attached screenshot. 
addressbook-1616480271384

Observed in Alpha3

### LINKED ISSUE #29402: Honor user defined order of the field when using "Document this" code action
See attached screenshot. 

addressbook-1616480611195

It is convenient to honer the user-defined order when generating documentation. In the language record fields are unordered. So we will get random fields in random order from the semantic API. But This information is available with SyntaxTree.

### LINKED ISSUE #29475: "Document this" code action for function, creates a comment with out of order parameters
**Description:**
I believe the `toCurrency` and `fromCurrency` order should be swapped in the comments syncing with the parameter order.

Screenshot 2021-03-23 at 16 18 25

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

### LINKED ISSUE #29541: Updating undocumented parameter document breaks the existing document
**Description:**
Check the images before and after applying the update documentation suggestion.

Screenshot 2021-03-24 at 01 18 28

Screenshot 2021-03-24 at 01 18 46

**Steps to reproduce:**

**Affected Versions:**

### LINKED ISSUE #29570: Wrong terminology used in sample docs generated
Refer to the following generated docs from the code action. In the sample descriptions they say "Parameter description". This should ideally be "Field description".
```ballerina
# Description
#
# + author - Parameter Description  
# + name - Parameter Description  
# + id - Parameter Description  
public type Book record {
    int id;
    string name;
    string author;
};
```

### LINKED ISSUE #29571: Change parameter type not working for resource method params
The change parameter type code action doesn't get suggested for resource method params. For similar scenarios in regular functions, it gets suggested.

IMS94:
## Purpose
$subject including fix for parameter ordering in add documentation and update documentation code actions.

Fixes #29225
Fixes #29541
Fixes #29475
Fixes #29402
Fixes #29571
Fixes #29397
Fixes #29570

See the discussion in that issue for more details.

## Approach
Updated `FunctionGenerator`/`CommonUtil` to return a `new ()` with default values for parameters of the initializer for classes when generating default values.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/docs/DocAttachmentInfo.java
```diff
@@ -87,20 +88,24 @@ public DocAttachmentInfo mergeDocAttachment(Documentation other) {
     }
 
     public String getDocumentationString() {
-        String result = String.format("# %s%n", this.description.trim());
+        StringBuilder result = new StringBuilder();
+        String[] descriptionLines = this.description.trim().split("\n");
```
nadeeshaan, line 92: Shouldn't we use the platform specific line separator here?

nadeeshaan (dismissed):

language-server/modules/langserver-core/src/test/resources/codeaction/update-documentation/source/updateDocumentation2.bal
```diff
@@ -0,0 +1,9 @@
+# Description
+#
+# + name - Parameter Description  
+# + id - Parameter Description  
+public type Student record {
+    string name;
+    int id;
+    float vals;
+};
```
mohanvive, line 9: Add a line space here.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/docs/DocAttachmentInfo.java
IMS94: @nadeeshaan  Actually in this commit: https://github.com/ballerina-platform/ballerina-lang/pull/29565/commits/85e91526d20b8fdb8e3972641ad7bdb059cdb806, I had the `System.lineSeparator()` for splitting. However, this caused the windows unit test for `UpdateDocumentationCommandExecutorTest` to fail. Assuming that this was due to the `\r\n` in windows, I changed it to `\n` as now, and the windows test passed afterwards. It's a bit confusing :thinking:

language-server/modules/langserver-core/src/test/resources/codeaction/update-documentation/source/updateDocumentation2.bal
IMS94: Done!

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/docs/DocAttachmentInfo.java
IMS94: Updated again to use `System.lineSeparator()`. Let's see what happens now.

nadeeshaan: If these are still failing, I think our tests are not aware of the platform risgt?

IMS94: Seems like the `description` in `DocAttachmentInfo` doesn't contain platform specific line separators. This cause the tests to fail on windows when `System.lineSeparator()` is used because the `description` is separated only with `\n`. For now, we are going to use `\n` for separating lines as discussed via zoom.

nadeeshaan (approved):
LGTM

PR #29614, opened 2021-03-25
Add an API to service decl symbol to get the attached listener types

pubudu91:
## Purpose
This PR adds a new method(`listenerTypes()`) to `ServiceDeclarationSymbol` for retrieving the types of the listeners to which the service is attached. 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
@@ -97,7 +97,7 @@ public class BallerinaServiceDeclarationSymbol extends BallerinaSymbol implement
         BServiceSymbol serviceSymbol = (BServiceSymbol) this.getInternalSymbol();
         Set<TypeSymbol> listenerTypes = new LinkedHashSet<>();
 
-        for (BType listenerType : serviceSymbol.getAttachExprTypes()) {
+        for (BType listenerType : serviceSymbol.getListenerTypes()) {
             listenerTypes.add(typesFactory.getTypeDescriptor(listenerType));
         }
 
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
@@ -41,11 +41,9 @@ import org.wso2.ballerinalang.compiler.util.Name;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.StringJoiner;
 
 import static io.ballerina.compiler.api.symbols.SymbolKind.SERVICE_DECLARATION;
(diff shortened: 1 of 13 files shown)


rdhananjaya (approved):
Looks good to me

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
```diff
@@ -59,6 +62,7 @@
     private final TypeSymbol typeDescriptor;
     private final ServiceAttachPoint attachPoint;
 
+    private Set<TypeSymbol> listenerTypes;
```
hasithaa, line 65: Shall we change this to List? if have multiple expressions with the same type reference we will get only one record. Which is not correct IMO.

docs/bir-spec/src/main/resources/kaitai/bir.ksy
```diff
@@ -542,6 +542,12 @@ types:
       - id: attach_point_literal
         type: s4
         if: has_attach_point_literal != 0
+      - id: listener_types_count
+        type: s4
+      - id: listener_types
```
Kishanthan, line 547: For better readability, it is better to create a new type called `listener_type` and refer it here, than defining inline. We must have followed such pattern in other places in this file. Please check and change it accordingly.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
pubudu91: Wouldn't that be redundant? For example,
```ballerina
listener http:Listener foo  = new(9090);
listener http:Listener bar  = new(8080);

service / on foo, bar {
}
```
In the above, if we return a list, it'll be `[http:Listener, http:Listener]`. Isn't this repetitive info that doesn't really add any additional insight?

docs/bir-spec/src/main/resources/kaitai/bir.ksy
pubudu91: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/29614/commits/34fb75349204470eef574baab2827c07c5caabf8

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
hasithaa: This can mislead those who use the API. Sometimes we need to related expression with the type. this will break this. Example. 

```
listener http:Listener foo  = new(9090);
listener http:Listener2 bar  = new(8080);
listener http:Listener baz  = new(8088);

service / on foo, bar, baz{
}
```

Above giving [http:Listener, http:Listener2] is wrong.

pubudu91: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/29614/commits/a5708ea50a63172b7420097e93abcdea7e289aa0

hasithaa (approved):

hasithaa:
@pubudu91 We need to add more test cases to cover cases reported by codecov. Could you please create an issue?

pubudu91:
> @pubudu91 We need to add more test cases to cover cases reported by codecov. Could you please create an issue?

Created #29628 to track this.

PR #29659, opened 2021-03-27
Refactor Transaction Info record

gimantha:
$Subject


langlib/lang.transaction/src/main/ballerina/transaction.bal
```diff
@@ -21,9 +21,7 @@ configurable boolean managerEnabled = false;
 # Config to specify transaction log directory
 configurable string logBase = "transaction_log_dir";
 
-# Information about a transaction that does not change
-# after the transaction is started.
-public type Info readonly & record {|
+public type InfoInternal record {|
```
gimantha, line 24: may be we have to make this not public

grainier (approved):

langlib/lang.transaction/src/main/ballerina/transaction.bal
gimantha: We cannot make this non-public as it gives me `"Attemp to refer to non-public.. "` error

langlib/lang.transaction/src/main/ballerina/transaction.bal
```diff
@@ -35,6 +33,10 @@ public type Info readonly & record {|
    Timestamp startTime;
 |};
 
+# Information about a transaction that does not change
+# after the transaction is started.
+public type Info readonly & InfoInternal;
```
gimantha, line 38: @manuranga it seems, we have a different issue here. Info type is referred in function parameters in function pointers (Commit and Rollback handlers). If user assigned an incompatible function pointer to these handlers, compiler gives the incompatible error, but in that error, the Internal record is exposed. Please refer to https://github.com/ballerina-platform/ballerina-lang/issues/29672

PR #29675, opened 2021-03-29
Fix several issues with `ensureType` function

### LINKED ISSUE #29219: Weird error with value:ensureType
**Description:**
$title. Observed when changing the return type of `ensureType` to be dependently-typed.

**Steps to reproduce:**
Consider the following example.
```ballerina
json p = {};

    int age = check p.name;
}

public function main() {
    any a = <string[]> ["hello", "world"];
    string[]|error strArray = a.ensureType(StringArray);
}

type StringArray string[];
```

Since the return type on master/alpha2 for `ensureType` is not dependently-typed we can expect an`incompatible types: expected '(string[]|error)', found '(any|error)'` error here. But the actual error logged is 

```
incompatible types: expected '(string[]|error)', found '(int|error)'
```

But when the `foo` function is commented out the expected error is logged!

**Affected Versions:**
slalpha2

### LINKED ISSUE #29368: ensureType doesn't works well with built-in subtypes
```ballerina
function test1() {
    int number1= 10;
    int:Unsigned32|error x = number1.ensureType(int:Signed32));
}
```

Gives 

`incompatible types: expected '(int:Unsigned32|error)', found '(any|error)'`

chiranSachintha:
Fixes #29219 
Fixes #29368 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -4821,21 +4819,21 @@ private void visitCheckAndCheckPanicExpr(BLangCheckedExpr checkedExpr) {
...
-            invocation.pkgAlias = (BLangIdentifier) TreeBuilder.createIdentifierNode();
-            checkedExpr.expr = invocation;
+            BType varRefType = getTypeOfExprInFieldAccess(((BLangFieldBasedAccess) exprWithCheckingKeyword).expr);
+            if (types.isLax(varRefType)) {
+                ArrayList<BLangExpression> argExprs = new ArrayList<>();
+                BType typedescType = new BTypedescType(expType, null);
+                BLangTypedescExpr typedescExpr = new BLangTypedescExpr();
+                typedescExpr.resolvedType = expType;
+                typedescExpr.type = typedescType;
+                argExprs.add(typedescExpr);
```
MaryamZi, line 4829: Can't we move this to desugar like we discussed?

MaryamZi: Created https://github.com/ballerina-platform/ballerina-lang/issues/29932.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2178,6 +2175,9 @@ public void visit(BLangSimpleVarRef varRefExpr) {
                     actualType = symTable.semanticError;
                     dlog.error(varRefExpr.pos, DiagnosticErrorCode.CANNOT_UPDATE_CONSTANT_VALUE);
                 }
+            } else if ((symbol.tag & SymTag.TYPE_DEF) == SymTag.TYPE_DEF || (symbol.tag & SymTag.TYPE) == SymTag.TYPE) {
```
MaryamZi, line 2178: I'm not sure if adding a check for SymTag.TYPE here is correct. IINM, SymTag.TYPE_DEF should have been set for the problem cases here.

@pubudu91 thoughts please?

pubudu91: What prompted to make this change? I also don't feel like checking for the `TYPE` sym tag here is correct.

chiranSachintha:
```ballerina
function testEnsureTypeFunction() {
    int number = 10;
    int:Unsigned32|error number1 = number.ensureType(int:Unsigned32);
}
```
The expected type for the parameter of `ensureType` is `typedesc`. Here `int:Unsigned32` is a simple variable reference. I used above change to set `actualType` as `typedesc<int:Unsigned32>`

MaryamZi: The original issue was that we couldn't refer to a builtin-subtype in a context where a typedesc is expected.

e.g.,
```ballerina
typedesc td = int:Unsigned32; // undefined symbol 'Unsigned32'
```

I just checked this with a user-defined type in a different module and it works without an issue. I think this can be an issue with the symbols defined specifically for builtin-subtypes (we create the symbol manually even though the type is defined in the langlib). Ideally, the symbol should have had the TYPE_DEF tag. 

@chiranSachintha, can you please check?

MaryamZi: This isn't an issue only via the BIR btw. Trying to refer to the type within the relevant langlib also fails with the same error.

chiranSachintha: `builtin-subtypes` only contains the TYPE tag. Should it have a `TYPE_DEF` tag though it is not a type definition?

MaryamZi: Right now, I believe for types associated with a type definition, we also use the TYPE_DEF tag. You can check this with something like

```ballerina
type Foo record {|
    
|};

Foo f = {}; // check the tag for Foo's type symbol here
```

I think we'll have to do the same here at least for now.

langlib/langlib-test/src/test/resources/test-src/valuelib_test_negative.bal
```diff
@@ -30,3 +30,12 @@ function testCloneWithTypeWithInvalidInferredType() {
...
     MyClass|error z = value:cloneWithType(x);
 }
+
+type Person record {
+    int name;
+};
+
+function testEnsureTypeFunction() returns error? {
+    Person p = {name: 12};
+    int age = check p.name;
```
MaryamZi, line 40: These tests should ideally go in a separate file, rather than in the langlib tests, since we're aren't directly testing the langlib. ensureType should have its own tests.

MaryamZi (approved):

langlib/langlib-test/src/test/resources/test-src/valuelib_test_negative.bal
MaryamZi: We can do this separately.

chiranSachintha: Created an issue to track this(https://github.com/ballerina-platform/ballerina-lang/issues/30009)

PR #29681, opened 2021-03-29
Fix issue with different field visibility qualifiers in object inclusions

### LINKED ISSUE #29385: No compilation error with object type inclusion when the field visibility qualifiers are different in the including type
**Description:**
$title. With object type inclusion, including types are expected to be subtypes of included types, and allowing different visibility qualifiers violates this.

**Steps to reproduce:**
```ballerina
public type Type object {
    public int[] vals;
};

public class Class {
    *Type;
    int[] vals = []; // no `public` qualifier, no error atm
}
```

**Affected Versions:**
slalpha3

lasinicl:
## Purpose
$title

Fixes #29385
<pre>
public type Type object {
    public int[] vals;
};

public class Class {
    *Type;
    int[] vals = []; // no `<b>public</b>` qualifier, <b>compilation error</b>
}
</pre>

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -698,6 +698,13 @@ private void defineReferencedClassFields(BLangClassDefinition classDefinition, S
             for (BField field : ((BStructureType) effectiveIncludedType).fields.values()) {
                 if (fieldNames.containsKey(field.name.value)) {
                     BLangSimpleVariable existingVariable = fieldNames.get(field.name.value);
+                    if ((Symbols.isFlagOn(Flags.asMask(existingVariable.flagSet), Flags.PUBLIC) !=
```
MaryamZi, line 701: Can't we do
```suggestion
                    if ((existingVariable.flagSet.contains(Flag.PUBLIC) !=
```

MaryamZi: And similarly for private in L703.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1676,6 +1676,9 @@ error.invalid.read.only.class.inclusion.in.non.read.only.class=\
 error.invalid.field.in.object.constructor.expr.with.readonly.reference=\
   invalid field in an object constructor expression with a ''readonly'' reference: ''{0}'' can never be ''readonly''
 
+error.mismatching.visibility.qualifiers.in.object.field=\
+  mismatching visibility qualifiers in referenced object field and referencing object field ''{0}''
```
MaryamZi, line 1680: Shall we change this to something like
```suggestion
  mismatched visibility qualifiers for field ''{0}'' in the including object and the included object
```

KRVPerera (commented):
[ ] I think we have more scenarios to test. 
> object-type-inclusion must be an object-type-descriptor, distinct-type-descriptor, intersection-type-descriptor or a class

  So we have object including a class, class including a object, object including object, class including class etc.
[ ] We also include method-decl from referenced types (object-field-descriptor or method-decl)

[ ]  Also it is worth investigating why we don't provide this error now 
INCOMPATIBLE_TYPE_REFERENCE_NON_PUBLIC_MEMBERS

[ ] We also can test with bala tests

lasinicl:
> [] I think we have more scenarios to test.
> 
> > object-type-inclusion must be an object-type-descriptor, distinct-type-descriptor, intersection-type-descriptor or a class
> 
> So we have object including a class, class including a object, object including object, class including class etc.
> [ ] We also include method-decl from referenced types (object-field-descriptor or method-decl)
> 
Added more tests to cover the above-mentioned scenarios. 

> [ ] Also it is worth investigating why we don't provide this error now
> INCOMPATIBLE_TYPE_REFERENCE_NON_PUBLIC_MEMBERS
> 
Tried out some scenarios and found that we do provide this error when the included type is in a different package and it contains non-public members.
Consider the following example
`Employee` is in `testOrg/foo`
```
public type Employee object {
    float salary;
};
```
And we try to include `Employee` 
```
import testorg/foo as foo;
public class PartTimeEmployee {
    *foo:Employee;
}
```
This would result in the following compilation error
```
incompatible type reference 'foo:Employee': a referenced object cannot have non-public fields or methods
```
Couldn't find any test that covered this. Added some tests to this PR just in case. 

> [ ] We also can test with bala tests

Added bala tests as well.

KRVPerera (commented):
LGTM

KRVPerera (approved):

IrushiL (dismissed):
Bindgen changes LGTM

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1676,6 +1676,9 @@ error.invalid.read.only.class.inclusion.in.non.read.only.class=\
 error.invalid.field.in.object.constructor.expr.with.readonly.reference=\
   invalid field in an object constructor expression with a ''readonly'' reference: ''{0}'' can never be ''readonly''
 
+error.mismatched.visibility.qualifiers.in.object.field=\
+  mismatched visibility qualifiers for field ''{0}'' in object type inclusion
```
MaryamZi, line 1680: Should we say
```suggestion
  mismatched visibility qualifiers for field ''{0}'' with object type inclusion
```

tests/jballerina-unit-test/src/test/resources/test-src/isolated-objects/isolated_objects.bal
```diff
@@ -41,7 +41,6 @@ final readonly & string[] immutableStringArray = ["hello", "world"];
 
 type IsolatedObjectType isolated object {
     int a;
-    string[] b;
 };
 
 isolated class IsolatedClassOverridingMutableFieldsInIncludedIsolatedObject {
```
MaryamZi, line 46: With the above change, this isn't quite overriding, right? Shall we change the field of `a` in this class to byte maybe?

MaryamZi (approved):

PR #29699, opened 2021-03-30
Make "always evaluate to true" error a hint

### LINKED ISSUE #28285: Incorrect error with 'is'

gimantha:
Fixes #28285

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticHintCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticHintCode.java
@@ -1,0 +1,58 @@
+/*
+ *  Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  WSO2 Inc. licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.ballerinalang.util.diagnostic;
+
+import io.ballerina.tools.diagnostics.DiagnosticCode;
+import io.ballerina.tools.diagnostics.DiagnosticSeverity;
+
+/**
+ * This class contains a list of diagnostic hint codes.
+ *
+ * @since Swan Lake
+ */
+public enum DiagnosticHintCode implements DiagnosticCode {
+
+    UNNECESSARY_CONDITION("BCE40000", "unnecessary.condition");
+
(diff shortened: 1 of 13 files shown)


compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticHintCode.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package org.ballerinalang.util.diagnostic;
+
+import io.ballerina.tools.diagnostics.DiagnosticCode;
+import io.ballerina.tools.diagnostics.DiagnosticSeverity;
+
+/**
+ * This class contains a list of diagnostic hint codes.
+ *
+ * @since Swan Lake
```
pubudu91, line 26: IINM, we use `2.0.0` for this.

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticHintCode.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public enum DiagnosticHintCode implements DiagnosticCode {
+
+    UNNECESSARY_CONDITION("BCE40000", "unnecessary.condition");
```
pubudu91, line 30: @dulajdilshan What does BCE stand for? If it's Ballerina Compiler Error, should we consider a different prefix for other categories of diagnostics?

dulajdilshan:
Shall we add a uniqueness check for `DiagnosticHintCodes`? Just need to add a function. Have a look at this https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/test/java/org/ballerinalang/util/DiagnosticCodeTest.java

https://github.com/ballerina-platform/ballerina-lang/blob/a9e90ceabc096524cd7d962a257931b751dc177d/compiler/ballerina-lang/src/test/java/org/ballerinalang/util/DiagnosticCodeTest.java#L46

pubudu91:
Wasn't there existing negative tests for this particular error? 🤔

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticHintCode.java
dulajdilshan: Yes. The earlier discussions were only to separate errors/warning/hints from the compiler side and the side of the extensions (having a different prefix for extensions). However, I'm +1 for having a different prefix for hints. (Ex: `BCH`). Also, this can be followed for warning too.

pubudu91: Shall we raise this in Slack and check with others as well?

gimantha: @dulajdilshan 

> Wasn't there existing negative tests for this particular error? 🤔

Yes. there are.. :) i will fix them.. thats why pr fails

gimantha:
> Wasn't there existing negative tests for this particular error? 🤔

Yes. Thats why the pr fails.. will fix them :)

PR #29719, opened 2021-03-30
Deprecate parameters() and introduce params() as the alternative

### LINKED ISSUE #29301: Verify whether we need to distinguish between the 'function' typedesc and regular function type descs
Need to evaluate and see whether we need to reconsider the design for the `FunctionTypeSymbol` with the recent introduction of the `function` typedesc. Internally, there's a slight difference in the symbols of these since the params list is deliberately made `null` for `function` typedesc. Currently we make no such distinctions in `FunctionTypeSymbol`. The params list is an empty list for the `function` typedesc as well. 

Need to see if there are any requirement to distinguish between the two types or whether that makes no difference from the Semantic API user POV.

### LINKED ISSUE #29714: [Semantic API] Support for included record params
Currently there's no way to identify included record params. For the other params we can use the `ParameterKind` to distinguish between different param types.

### LINKED ISSUE #29717: Wrong flags set to function type symbol when resolving the type
See https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java#L1528

Here it sets the flags of the type as the flags of the param. But this should be set using the mask of the flag set in the param node.

pubudu91:
## Purpose
This PR is mainly for introducing a new method for retrieving the params of a function/function type symbol which takes in to account the `function` typedesc. This also addresses a couple of other issues as well, which hindered the introduction of this method.

Fix #29301 
Fix #29714 
Fix #29717 

## Approach
- Introduce a new method: `Optional<List<ParameterSymbol>> params()`
- For regular cases, this will return a list of params as usual. If it is the `function` typedesc, this will return empty.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/AbstractTypeSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/AbstractTypeSymbol.java
@@ -134,7 +134,7 @@ public abstract class AbstractTypeSymbol implements TypeSymbol {
         List<FunctionSymbol> filteredFunctions = new ArrayList<>();
 
         for (FunctionSymbol function : functions) {
-            ParameterSymbol firstParam = function.typeDescriptor().parameters().get(0);
+            ParameterSymbol firstParam = function.typeDescriptor().params().get().get(0);
             BType firstParamType = ((AbstractTypeSymbol) firstParam.typeDescriptor()).getBType();
 
             if (types.isAssignable(internalType, firstParamType)) {
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionTypeSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionTypeSymbol.java
@@ -93,8 +93,7 @@ public class BallerinaFunctionTypeSymbol extends AbstractTypeSymbol implements F
             return Optional.of(this.params);
         }
 
-        // Becomes null for the function typedesc.
-        if (this.typeSymbol.params == null) {
+        if (Symbols.isFlagOn(this.typeSymbol.flags, Flags.ANY_FUNCTION)) {
             return Optional.empty();
         }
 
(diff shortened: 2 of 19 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/AbstractTypeSymbol.java
```diff
@@ -134,7 +134,7 @@ public BType getBType() {
         List<FunctionSymbol> filteredFunctions = new ArrayList<>();
 
         for (FunctionSymbol function : functions) {
-            ParameterSymbol firstParam = function.typeDescriptor().parameters().get(0);
+            ParameterSymbol firstParam = function.typeDescriptor().params().get().get(0);
```
IMS94, line 137: Can the `typeDescriptot().params()` return `empty`? According to `FunctionTypeSymbol` implementation, it can right?

IMS94 (changes requested):
Added a comment on checking if params returned an `Optional.empty()`.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/AbstractTypeSymbol.java
pubudu91: In this particular context, no. It only return empty when it represents the `function` typedesc. Since here, we are getting the function type symbol from a `FunctionSymbol` instance, it should never return empty.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionTypeSymbol.java
```diff
@@ -116,7 +153,7 @@ public String signature() {
 
         StringBuilder signature = new StringBuilder("function (");
         StringJoiner joiner = new StringJoiner(", ");
-        for (ParameterSymbol requiredParam : this.parameters()) {
+        for (ParameterSymbol requiredParam : this.params().get()) {
```
dulajdilshan, line 156: We need to have a `isPresent()` check

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/AbstractTypeSymbol.java
IMS94: No issue then :+1:

nadeeshaan (approved):
LGTM

PR #29755, opened 2021-04-01
Check all completions are nil for foreach

pcnfernando:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -1652,7 +1652,8 @@ public class Types {
                     break;
                 }
                 varType = streamType.constraint;
-                if (streamType.error != null) {
+                List<BType> completionType = getAllTypes(streamType.error);
+                if (completionType.stream().anyMatch(type -> type.tag != TypeTags.NIL)) {
                     BType actualType = BUnionType.create(null, varType, streamType.error);
                     dlog.error(foreachNode.collection.pos, DiagnosticErrorCode.INCOMPATIBLE_TYPES,
                             varType, actualType);
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/query/SimpleQueryExpressionWithDefinedTypeTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/query/SimpleQueryExpressionWithDefinedTypeTest.java
@@ -363,6 +363,13 @@ public class SimpleQueryExpressionWithDefinedTypeTest {
                 "deptAccess:\"XYZ\", address:{city:\"Colombo\", country:\"SL\"}}");
     }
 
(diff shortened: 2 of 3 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1652,7 +1652,8 @@ public void setForeachTypedBindingPatternType(BLangForeach foreachNode) {
                     break;
                 }
                 varType = streamType.constraint;
-                if (streamType.error != null) {
+                List<BType> completionType = getAllTypes(streamType.error);
+                if (completionType.stream().anyMatch(type -> type.tag != TypeTags.NIL)) {
```
MaryamZi, line 1656: The spec says for the collection in a foreach statement

> The static type of action-or-expr must be an iterable type with an **iteration completion type of nil**.

Since `never` is a subtype of `()`, I think this should work with `never` too. WDYT? We can change this separately though.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/query/SimpleQueryExpressionWithDefinedTypeTest.java
```diff
@@ -363,6 +363,13 @@ public void testQueryWithRecordVarInLetClause() {
                 "deptAccess:\"XYZ\", address:{city:\"Colombo\", country:\"SL\"}}");
     }
 
+    @Test(description = "Test looping a stream with foreach")
+    public void testForeachStream() {
+        BValue[] returnValues = BRunUtil.invoke(result, "testForeachStream");
+        Assert.assertNotNull(returnValues);
+        Assert.assertTrue(((BBoolean) returnValues[0]).booleanValue());
```
MaryamZi, line 370: We can assert this in b7a itself, right? We can maybe change this also separately.

MaryamZi (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
pcnfernando: Will handle this separately after clarifying. https://github.com/ballerina-platform/ballerina-lang/issues/29758 will be used to track this

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/query/SimpleQueryExpressionWithDefinedTypeTest.java
pcnfernando: This was done since the rest of the cases use the same approach.

PR #29792, opened 2021-04-02
Improve runtime anonymous type resolution

### LINKED ISSUE #28262: Dependencies can become incompatible due to anonymous types

grainier:
## Purpose
The purpose of this PR is to $title.

Fixes #28262

## Approach
In this PR, it'll code-gen `public Type getAnonType(int hash, String shapeInfo)` method into module init, and that will allow type resolution like shown below.

**generated method:**
```java
public Type getAnonType(int var1, String var2) {
    switch(var1) {
    case -1449513272:
        return $type$$anonType$_9;
    case -1413292641:
        return $type$$anonType$_14;
    case -899684319:
        return $type$$anonType$_8;
    case -270018685:
        return $type$$anonType$_0;
    case 18898473:
        return $type$$anonType$_15;
    case 124652424:
        return $type$$anonType$_4;
    case 1937814141:
        return $type$$anonType$_11;
    default:
        throw new ErrorValue(StringUtils.fromString("No such type: " + var2));
    }
}
```

**generated usage:**
```java
BObjectType var10006 = (BObjectType)(new $_init()).getAnonType(124652424, "Package: ballerina/lang.__internal:0.1.0, Shape: object { public isolated function next () returns ((record {| (any|error) value; |}|error?)); }");
```

In case, if the type couldn't be resolved, a runtime error similar to the following will be thrown.
```
error("No such type: Package: ballerina/lang.__internal:0.1.0, Shape: object { public isolated function next () returns ((record {| (any|error) value; |}|error?)); })
```

## Samples
n/a

## Remarks
- #28262

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


rdhananjaya:
I think we need to add the string equals test too to avoid hash collusions. Similar to `get` method of record value classes. 

```ballerina
type T record {|
    int i;
    string j;
|};
```

Generated (decompiled to java) `get` method
```java
        String var2 = ((BString)var1).getValue();
        switch(var2.hashCode()) {
        case 105:
            if (var2.equals("i")) {
                return this.i;
            }
            break;
        case 106:
            if (var2.equals("j")) {
                return this.j;
            }
        }

        return super.get(var1);
    }
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
```diff
@@ -1856,13 +1932,28 @@ private void loadTupleType(MethodVisitor mv, BTupleType bType) {
      * @param bType user defined type
      */
     private void loadUserDefinedType(MethodVisitor mv, BType bType) {
+        BTypeSymbol typeSymbol = bType.tsymbol.isTypeParamResolved ? bType.tsymbol.typeParamTSymbol : bType.tsymbol;
+        BType typeToLoad = bType.tsymbol.isTypeParamResolved ? typeSymbol.type : bType;
```
rdhananjaya, line 1936: I think this and above line can be merged `BType typeToLoad = bType.tsymbol.isTypeParamResolved ? bType.tsymbol.typeParamTSymbol.type : bType;`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
```diff
@@ -1856,13 +1932,28 @@ private void loadTupleType(MethodVisitor mv, BTupleType bType) {
...
+        PackageID packageID = typeSymbol.pkgID;
+        String typeOwner = JvmCodeGenUtil.getPackageName(packageID) + MODULE_INIT_CLASS_NAME;
+        String fieldName = getTypeFieldName(toNameString(typeToLoad));
 
-        PackageID packageID = bType.tsymbol.pkgID;
+        if (fieldName.contains(BLangAnonymousModelHelper.ANON_PREFIX)) {
```
rdhananjaya, line 1941: We only need to do this when loading annon types from other modules right, I think we need to add a check to see if the loaded type is not from the current package being compiled.

rdhananjaya: @warunalakshitha @manuranga Is it possible to get the information about current module being compiled here?

manuranga: We could try adding it as the field to current class (JvmTypeGen) and provide that from the constructor. Have to try it can see.

manuranga:
+1 for @rdhananjaya 's suggestion, but since current hash calculator (`UniqueTypeVisitor<Integer>`) returns an Integer, I am not sure this is possible to do without changing it.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
grainier: Can we do `&& packageID != PackageID.DEFAULT`?

grainier:
> +1 for @rdhananjaya 's suggestion, but since current hash calculator (`UniqueTypeVisitor<Integer>`) returns an Integer, I am not sure this is possible to do without changing it.

@manuranga, @rdhananjaya I'll add the hash validation in a separate PR. (issue to track this: https://github.com/ballerina-platform/ballerina-lang/issues/29928)

KRVPerera (commented):
This means for the similar anonymous types we have the same signature. Is this expected?

```Ballerina
public type Address object {
    public string city;
    public string country;

    public function value() returns string;
};

public type Address3 object {
    public string city;
    public string country;

    public function value() returns string;
};

public function main() {

var address = object Address {
                     // In object constructor expressions, the `init` method does not take any parameters.
                     public function init() {
                         self.city = "Colombo";
                         self.country = "Sri Lanka";
                     }

                     public function value() returns string {
                         return self.city + ", " + self.country;
                     }
                 };


var address2 = object Address3 {
                     // In object constructor expressions, the `init` method does not take any parameters.
                     public function init() {
                         self.city = "Colombo";
                         self.country = "Sri Lanka";
                     }

                     public function value() returns string {
                         return self.city + ", " + self.country;
                     }
                 };
}
```
Here for both **$type$$anonType$_1** and **$type$$anonType$_0** we have same signature.

```
final BObjectType duplicate = ((BObjectType)new $_init().getAnonType(-203469200, "Package: $_init, TypeName: $type$$anonType$_0, Shape: object { public string city; public string country; public function value () returns (string); }")).duplicate();


final BObjectType duplicate2 = ((BObjectType)new $_init().getAnonType(-203469200, "Package: $_init, TypeName: $type$$anonType$_1, Shape: object { public string city; public string country; public function value () returns (string); }")).duplicate();
```

```
@Override
    public Type getAnonType(final int n, final String str) {
        switch (n) {
            case -203469200: {
                return $_init.$type$$anonType$_0;
            }
            default: {
                throw new ErrorValue(StringUtils.fromString("No such type: " + str));
            }
        }
    }
```

grainier:
@KRVPerera yes. We have only considered the shape. So, such situations can happen. Do you think this can lead to errors?

KRVPerera:
Think it is okay. We are taking type ids also to the hash. Sorry for the noise.

grainier:
> Think it is okay. We are taking type ids also to the hash. Sorry for the noise.

I really wasn't aware whether there can be any issues by treating similar shapes as one. That's why. :) 

Anyway, we didn't use the type name, because if there's an anon type name and if we capture that as part of the shape, there's a possibility for that anon name to change later on (since we are using incremental anon type name generator, like in the root issue), and cause the hash to change as well.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
```diff
@@ -501,6 +511,74 @@ private void addTypeId(MethodVisitor mv, BTypeIdSet.BTypeId typeId, boolean isPr
...
+        mv.visitCode();
+
+        int hashParamRegIndex = 1;
+        int shapeParamRegIndex = 2;
+        Label defaultCaseLabel = new Label();
+
+        // filter anon types and sorts them before generating switch case.
+        Set<BIRTypeDefinition> typeDefSet = new TreeSet<>(TYPE_HASH_COMPARATOR);
+        for (BIRTypeDefinition t : typeDefinitions) {
+            if (t.internalName.value.contains(BLangAnonymousModelHelper.ANON_PREFIX)) {
```
MaryamZi, line 530: I think there are some cases where we generate types/names but they don't have the anon prefix. Shall we create an issue to track such types and maybe use the ANONYMOUS flag here?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/internal/TypeHashComparator.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.wso2.ballerinalang.compiler.bir.model.BIRNode.BIRTypeDefinition;
+import org.wso2.ballerinalang.compiler.semantics.analyzer.TypeHashVisitor;
+
+import java.util.Comparator;
+
+/**
+ * BType hash comparator to sort BIRTypeDefinitions according its type hash code.
+ *
+ * @since Swan Lake
```
MaryamZi, line 28:
```suggestion
 * @since 2.0.0
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeHashVisitor.java
```diff
@@ -0,0 +1,574 @@
...
+import java.util.Optional;
+import java.util.Stack;
+import java.util.stream.Collectors;
+
+import static java.util.Objects.hash;
+
+/**
+ * BType hash visitor to visit and generate a hash code for a given BType.
+ *
+ * @since Swan Lake
```
MaryamZi, line 72:
```suggestion
 * @since 2.0.0
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeHashVisitor.java
```diff
@@ -0,0 +1,574 @@
...
+
+    @Override
+    public Integer visit(BErrorType type) {
+        if (isVisited(type)) {
+            return visited.get(type);
+        }
+        if (isCyclic(type)) {
+            return 0;
+        }
+        Integer hash = hash(baseHash(type), visit(type.detailType));
```
KRVPerera, line 312: @rdhananjaya Do we need to add type Id set as a precaution?

KRVPerera (commented):
LGTM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
grainier: created: https://github.com/ballerina-platform/ballerina-lang/issues/29948

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
```diff
@@ -501,6 +511,74 @@ private void addTypeId(MethodVisitor mv, BTypeIdSet.BTypeId typeId, boolean isPr
...
+            mv.visitFieldInsn(GETSTATIC, typeOwnerClass, fieldName, String.format("L%s;", TYPE));
+            mv.visitInsn(ARETURN);
+        }
+
+        createDefaultCase(mv, defaultCaseLabel, shapeParamRegIndex, "No such type: ");
+        mv.visitMaxs(typeDefSet.size() + 10, typeDefSet.size() + 10);
+        mv.visitEnd();
+    }
+
+    private Map<String, Label> createLabelsForAnonTypeHashSwitch(MethodVisitor mv,
```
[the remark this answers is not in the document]
grainier: it'll require re-factoring both methods since the logic is different.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeHashVisitor.java
grainier: Type id set hash included in error type hash and object type hash.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
grainier: Fixed in: 3da69b3

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java
grainier: fixed in: 9a611ab

hasithaa (approved):
Approving since review suggestions are added.

PR #29807, opened 2021-04-02
Fix imported non-read only records used in configurable variables

HindujaB:
## Purpose
$subject

## Approach
Added loading the parent package of the record.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
@@ -355,12 +355,16 @@ public class TomlProvider implements ConfigProvider {
     private BMap<BString, Object> retrieveRecordValues(TomlNode tomlNode, String variableName, Type type) {
         RecordType recordType;
         String recordName;
+        Module recordPkg;
         if (type.getTag() == TypeTags.RECORD_TYPE_TAG) {
             recordName = type.getName();
             recordType = (RecordType) type;
+            recordPkg = recordType.getPackage();;
         } else {
-            recordName = ((BIntersectionType) type).getConstituentTypes().get(0).getName();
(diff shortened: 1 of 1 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
```diff
@@ -355,12 +355,16 @@ private BArray retrieveArrayValues(TomlNode tomlValue, String variableName, Arra
...
         RecordType recordType;
         String recordName;
+        Module recordPkg;
         if (type.getTag() == TypeTags.RECORD_TYPE_TAG) {
             recordName = type.getName();
             recordType = (RecordType) type;
+            recordPkg = recordType.getPackage();;
         } else {
-            recordName = ((BIntersectionType) type).getConstituentTypes().get(0).getName();
+            Type originalType = ((BIntersectionType) type).getConstituentTypes().get(0);
```
manuranga, line 364: Can non-readonly part come as `.get(1)` ?

warunalakshitha, line 364: I think we can bring this logic to codegen side. May be we can do that in master branch

HindujaB: no. when I check with front-end we always have an order for intersectionType
https://github.com/ballerina-platform/ballerina-lang/blob/930969f1e3a43fff8a5ec2f74d8484b7df4bd0e9/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/util/ImmutableTypeCloner.java#L820

PR #29814, opened 2021-04-04
Improve diagnostics and support for `never` type

### LINKED ISSUE #29663: Address suggestions for #29361
**Description:**
$title

**Describe your task(s)**

lasinicl:
## Purpose
$title

Fixes #29663

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1903,8 +1903,8 @@ public void visit(BLangExpressionStmt exprStmtNode) {
         SymbolEnv stmtEnv = new SymbolEnv(exprStmtNode, this.env.scope);
         this.env.copyTo(stmtEnv);
         BType bType = typeChecker.checkExpr(exprStmtNode.expr, stmtEnv, symTable.noType);
-        if (bType != symTable.nilType && bType != symTable.neverType && bType != symTable.semanticError &&
-                exprStmtNode.expr.getKind() != NodeKind.FAIL) {
+        if (bType != symTable.nilType && !types.isNeverTypeOrStructureTypeWithARequiredNeverMember(bType) &&
```
pcnfernando, line 1906: Better to have the least overhead checks first in a &&. Move  isNeverTypeOrStructureTypeWithARequiredNeverMember () to last. If the prior condition is false, rest of them won't be evaluated

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -683,6 +683,11 @@ private boolean isAssignable(BType source, BType target, Set<TypePair> unresolve
             return false;
         }
 
+        if (isNeverTypeOrStructureTypeWithARequiredNeverMember(source) &&
+                isNeverTypeOrStructureTypeWithARequiredNeverMember(target)) {
+            return true;
+        }
```
MaryamZi, line 689: The following fails at the L648 mutability check.

```ballerina
function func() returns record {| never x; |} {
    panic error("error!");
}

function () returns never x = func;
```

We can maybe repeat this check within that block or make this one of the first checks.

One other option is to analyze the structure type and if it is equivalent to never, set the read-only flag. This would have to be done once all the fields are defined but before semantic analysis I guess.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -683,6 +683,11 @@ private boolean isAssignable(BType source, BType target, Set<TypePair> unresolve
             return false;
         }
 
+        if (isNeverTypeOrStructureTypeWithARequiredNeverMember(source) &&
```
MaryamZi, line 686: Need to introduce these changes at runtime too.

```ballerina
function func() returns never {
    panic error("error!");
}

function () returns record {| never x; |} x = func;

public function main() {
    boolean b = func is function () returns record {| never x; |}; // HINT [foo.bal:(8:17,8:66)] unnecessary condition: expression will always evaluate to 'true'
    panic error(string `${b}`); // but evaluates to false
}
```

lasinicl:
This PR needs to be updated once #29905 PR is merged.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/AnnotationUtils.java
```diff
@@ -109,7 +109,7 @@ private static void processObjectMethodLambdaAnnotation(MapValue globalAnnotMap,
     public static void processFPValueAnnotations(FPValue fpValue, MapValue globalAnnotMap, String name) {
         BAnnotatableType type = (BAnnotatableType) fpValue.getType();
         BString nameKey = StringUtils.fromString(name);
-        if (globalAnnotMap.containsKey(nameKey)) {
+        if (globalAnnotMap != null && globalAnnotMap.containsKey(nameKey)) {
```
MaryamZi, line 112: How does this become null now?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -552,6 +552,10 @@ public static boolean checkIsType(Type sourceType, Type targetType) {
 
     @Deprecated
     public static boolean checkIsType(Type sourceType, Type targetType, List<TypePair> unresolvedTypes) {
+        if (checkIsNeverTypeOrStructureTypeWithARequiredNeverMember(sourceType)) {
```
MaryamZi, line 555: I think we should move this to after the same type check, since the chances of this being true is less.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/AnnotationUtils.java
lasinicl: This issue is already there in the master. Even for the following, I get the NPE.
```
function func() returns int {
    return 1;
}
boolean b = func is function () returns int;
panic error(string `${b}`);
```

MaryamZi:
```ballerina
function func() returns int {
    return 1;
}
public function main() {
    boolean b = func is function () returns int;
    panic error(string `${b}`);
}
```
seems to work for me.

lasinicl: This doesn't work for me. Not sure why. I have reverted this change anyway. Will check whether the checks pass without this change

lasinicl: The build fails with an NPE without this change. Hence added back the null check.

MaryamZi: Doesn't that mean this is becoming null due to a change in this PR? Shall we try and figure out why?

lasinicl: Since this happens only when writing unit tests, need to properly investigate the cause for this. Created https://github.com/ballerina-platform/ballerina-lang/issues/30017 to track this. To avoid the null check for annotation map, a cast was added for the runtime tests to avoid expressions that always evaluate to true.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BFunctionType.java
```diff
@@ -133,6 +133,11 @@ public boolean equals(Object o) {
         if (!Arrays.equals(paramTypes, that.paramTypes)) {
             return false;
         }
+
+        if (retType == null) {
```
MaryamZi, line 137: AFAIK, the absence of a return type is always desuaged to be nil (and it is not never). The only exception I believe is with `function`, for which we don't have a return type, but we shouldn't get here due to the checks at L115 and L119. Can you share the sample for which this became null here?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -652,6 +656,10 @@ private static boolean checkIsType(Object sourceVal, Type sourceType, Type targe
         int sourceTypeTag = sourceType.getTag();
         int targetTypeTag = targetType.getTag();
 
+        if (checkIsNeverTypeOrStructureTypeWithARequiredNeverMember(sourceType)) {
```
MaryamZi, line 659: Shouldn't we do this before L690 instead?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -1916,6 +1924,33 @@ private static boolean checkArrayEquivalent(Type actualType, Type expType) {
...
 
+    private static boolean checkIsNeverTypeOrStructureTypeWithARequiredNeverMember(Type type) {
+        switch (type.getTag()) {
+            case TypeTags.NEVER_TAG:
+                return true;
+            case TypeTags.RECORD_TYPE_TAG:
+                for (Field field : ((BRecordType) type).getFields().values()) {
+                    // skip check for fields with self referencing type and not required fields.
+                    if (SymbolFlags.isFlagOn(field.getFlags(), SymbolFlags.REQUIRED) &&
+                            !isSameType(type, field.getFieldType()) &&
```
MaryamZi, line 1935: Instead of the `isSameType` check here (which seems to do some additional checks), IMO, it would be better to use a set with the already visited types passed as a a param for checkIsNeverTypeOrStructureTypeWithARequiredNeverMember.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -1916,6 +1924,33 @@ private static boolean checkArrayEquivalent(Type actualType, Type expType) {
...
+                            checkIsNeverTypeOrStructureTypeWithARequiredNeverMember(field.getFieldType())) {
+                        return true;
+                    }
+                }
+                return false;
+            case TypeTags.TUPLE_TAG:
```
MaryamZi, line 1941: Shouldn't we consider a closed array of type `never` to be `never`? e.g., `never[3]`?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -1916,6 +1924,33 @@ private static boolean checkArrayEquivalent(Type actualType, Type expType) {
...
     }
 
+    private static boolean checkIsNeverTypeOrStructureTypeWithARequiredNeverMember(Type type) {
+        switch (type.getTag()) {
+            case TypeTags.NEVER_TAG:
+                return true;
+            case TypeTags.RECORD_TYPE_TAG:
+                for (Field field : ((BRecordType) type).getFields().values()) {
+                    // skip check for fields with self referencing type and not required fields.
+                    if (SymbolFlags.isFlagOn(field.getFlags(), SymbolFlags.REQUIRED) &&
```
MaryamZi, line 1934: AFAIK, for fields with default values neither the REQUIRED flag nor the OPTIONAL flag is set. You may have to do something like
`SymbolFlags.isFlagOn(field.getFlags(), SymbolFlags.REQUIRED) || !SymbolFlags.isFlagOn(field.getFlags(), SymbolFlags.OPTIONAL)`

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never-type-runtime.bal
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
MaryamZi, line 1: Let's use underscores for file names.

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never-type-runtime.bal
```diff
@@ -0,0 +1,98 @@
...
+    panic error("error!");
+}
+
+function func2() returns record {| never x; |} {
+    panic error("error!");
+}
+
+function test1() {
+    boolean b = <any> func1 is function () returns record {| never x; |};
+    panic error("Error value is:" + string `${b}`);
```
MaryamZi, line 53: Can't we just assert `assertEquality(true, b)` instead of panicking here and introducing a separate test function?

Let's also add some negative tests - target has only never, source has something that's not never.

MaryamZi (approved):

MaryamZi:
@lasinicl, let's address the suggestions in a separate PR.

PR #28712, opened 2021-02-17
Set BType for function related BInvokableTypeSymbol

sanjana:
Fixes #28583 

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #29859, opened 2021-04-06
Improve the Return Type Descriptor in Function Type Symbol

### LINKED ISSUE #27059: Incorrect/incomplete variable types with "Create Local Variable"  code actions for iterators
**Description:**
$title.

**Steps to reproduce:**
For

```ballerina
public function main() {
    int[] x = [1, 2, 3];

    x.iterator();

    x.iterator().next();
}
```

adds

```ballerina
public function main() {
    int[] x = [1, 2, 3];

     object {function () returns {| any| value |}|()} iteratorResult = x.iterator();

    function () nextResult = x.iterator().next();
}
```

but it should be

```ballerina
public function main() {
    int[] x = [1, 2, 3];

    object {public function next() returns record {| int value; |}?;} iteratorResult = x.iterator();

    record {| int value; |}? nextResult = x.iterator().next();
}
```

**Related issues:**
https://github.com/ballerina-platform/ballerina-lang/issues/24379

**Affected Versions:**
slp5

sanjana:
Fixes #27059

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/test/resources/codeaction/create-variable/config/variableAssignmentRequiredCodeAction42.json
```diff
@@ -16,7 +16,7 @@
               "character": 4
             }
           },
-          "newText": "record {| int value; |}? next = "
+          "newText": "object {public isolated function next() returns record {| int value; |}? ;} iterator = "
```
nadeeshaan, line 19: We have to fix this code action test case. We have edited the `createVariable7.bal` source and the cursor position of this test should be changed to "line": 3

nadeeshaan (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionTypeSymbol.java
```diff
@@ -133,7 +133,12 @@ public BallerinaFunctionTypeSymbol(CompilerContext context, ModuleID moduleID,
     public Optional<TypeSymbol> returnTypeDescriptor() {
         if (returnType == null) {
             TypesFactory typesFactory = TypesFactory.getInstance(this.context);
-            this.returnType = typesFactory.getTypeDescriptor(this.typeSymbol.returnType);
+            if (this.getBType().getReturnType() != null) {
+                this.returnType = typesFactory.getTypeDescriptor(this.getBType().getReturnType());
+            } else {
+                this.returnType = typesFactory.getTypeDescriptor(this.typeSymbol.returnType);
```
dulajdilshan, line 139: Shouldn't this be fixed in the beginning when we creating the `BallerinaFunctionTypeSymbol`? WDYT @pubudu91?

dulajdilshan, line 139: We can improve this one by extracting `this.getBType().getReturnType()` to a variable. (minor)

language-server/modules/langserver-core/src/test/resources/codeaction/create-variable/config/variableAssignmentRequiredCodeAction42.json
sanjana: @nadeeshaan a new test case for "line": 3 has been added in [language-server/modules/langserver-core/src/test/resources/codeaction/create-variable/config/variableAssignmentRequiredCodeAction43.json](https://github.com/ballerina-platform/ballerina-lang/pull/29859/files#diff-4e3bb2f57483b284225072e10691cc600eff09584a00ee9f453032e9c6758626).

Should the new test case be set in there and this config file be as it is (with only the line number changes)?

nadeeshaan: No issue. Then it is fine.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionTypeSymbol.java
sanjana: Fixed.

pubudu91: What's the issue here? Is `typeSymbol.returnType` becoming `null` in some cases? If so, should look at why that's happening. The only case where it can be `null` is for the `function` typedesc. In all other cases, it should not be `null`.

sanjana: Yes, for `function` typedesc it is set as null. Let me check and update.

sanjana: @pubudu91 seems like the `BType` issue was previously mentioned in the **Remarks** section of a workaround fix at #28712

dulajdilshan: @sanjana can't we fix the issue where we are getting the `null` for `typeSymbol.returnType` in this particular place?

sanjana: @dulajdilshan yes, I'm checking on it now. Seems like an issue in the TypeChecker.

sanjana: Seems like the `tSymbol.returnType` is being `null` by this stage. 

image

sanjana: Changed the approach.

nadeeshaan (dismissed):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/Symbols.java
```diff
@@ -156,7 +157,12 @@ public static BTypeSymbol createTypeSymbol(int symTag,
                                                Location pos,
                                                SymbolOrigin origin) {
         if (type != null && type.tag == TypeTags.INVOKABLE) {
-            return createInvokableTypeSymbol(symTag, flags, pkgID, type, owner, pos, origin);
+            BInvokableTypeSymbol invokableTypeSymbol =
+                    createInvokableTypeSymbol(symTag, flags, pkgID, type, owner, pos, origin);
+            if (type instanceof BInvokableType) {
```
pubudu91, line 162: This if check is unnecessary since we check the type tag above.

sanjana: @pubudu91 without this check, there would be a `SpotBugs` test failure. 

image

pubudu91: Let's add an exception for this warning in the `Symbols` class

sanjana: Added a rule to exclude the `BC_UNCONFIRMED_CAST` in `Symbols` class at SpotBugs excluding XML file.

nadeeshaan (approved):
LGTM

dulajdilshan (approved):
LGTM

Issue #29862, opened 2021-04-06
`ballerina/test` should not persist to `Dependencies.toml`


pramodya1994:
**Description:**
$Title
```
[[dependency]]
org = "ballerina"
name = "test"
version = "0.0.0"
```

- [ ] Need to change `isLangLib` to `isBuiltInPackage` and ignore `ballerinai` and `ballerina/test` packages.
- [ ] Also need to verify dependencies with test scope not packing to the bala.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #29875, opened 2021-04-06
Make built in packages not persist to `Dependencies.toml`

pramodya1994:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/29862

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageDescriptor.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageDescriptor.java
@@ -72,9 +72,11 @@ public class PackageDescriptor {
         return Optional.ofNullable(repository);
     }
 
-    public boolean isLangLibPackage() {
+    public boolean isBuiltInPackage() {
         return (org().isBallerinaOrg() && packageName.value().startsWith(LANG_LIB_PACKAGE_NAME_PREFIX)) ||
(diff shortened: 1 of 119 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageDescriptor.java
```diff
@@ -72,9 +72,11 @@ public PackageVersion version() {
         return Optional.ofNullable(repository);
     }
 
-    public boolean isLangLibPackage() {
+    public boolean isBuiltInPackage() {
```
sameerajayasoma, line 75: This is an API change.  Should we do this now? 

Can we introduce a new method and deprecate the old method later?

sameerajayasoma (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageDescriptor.java
pramodya1994: Decided to add a new `isBuiltInPackage(...)` method and use it only when persisting dependencies to `Dependencies.toml`. This will resolve this.

sameerajayasoma (approved):

PR #29880, opened 2021-04-07
If the attach-point is not provided, the argument provided for the name parameter of attach method should be ()

### LINKED ISSUE #28848: Invalid argument passed for the attach method's second parameter when a service is declared with no attach point
**Description:**
$title.

The spec says 
> The attach-point determines the second argument passed to the attach method: if the attach-point is absent, then the argument is nil

but we seem to be passing an empty string array atm.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public class Listener {
    public isolated function 'start() returns error? { }

    public isolated function gracefulStop() returns error? { }

    public isolated function immediateStop() returns error? { }

    public isolated function detach(service object {} s) returns error? { }

    public isolated function attach(service object {} s, string? name) returns error? {
        io:println(name is ()); // false
        io:println(<any> name is string[]); // true, eventhough `name` is of type `string?`
    }
}

service on new Listener() {

}
```

**Affected Versions:**
slalpha2

gimantha:
Fixes #28848

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/services/ServiceDeclTest.java
```diff
@@ -43,6 +45,13 @@ public void testServiceNameLiteral() {
         BRunUtil.invoke(compileResult, "testServiceName");
     }
 
+    @Test
+    public void testAttachMethodParams() {
+        CompileResult compileResult = BCompileUtil.compile("test-src/services/service_attach_test.bal");
+        BValue[] values = BRunUtil.invoke(compileResult, "testAttachMethodParams");
+        Assert.assertTrue(((BBoolean)values[0]).booleanValue());
```
MaryamZi, line 52: Shall we do this assertion in b7a itself?

tests/jballerina-unit-test/src/test/resources/test-src/services/service_attach_test.bal
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+service "Service" on new Listener2() {
+
+}
+
+public function testAttachMethodParams() returns boolean {
+    lock {
+        return nameParamCheck[0] && nameParamCheck[1];
```
MaryamZi, line 61: Not directly related to this PR, but shall we add another service and test the scenario where the value is a string array if we don't have such a test already?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/services/ServiceDeclTest.java
gimantha: sure

tests/jballerina-unit-test/src/test/resources/test-src/services/service_attach_test.bal
gimantha: @MaryamZi Does that work in normal case? To test the array case, we need to have a resource path right? Or should that test go to http module?

MaryamZi: It's when the service attach-point is an absolute-resource-path with resource-path-segments right?

https://ballerina.io/ballerina-spec/spec.html#attach-point

```ballerina
service /foo/bar on new Listener() {
    resource function get . () {
        
    }    
}
```

MaryamZi (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ServiceDesugar.java
```diff
@@ -205,14 +206,20 @@ void rewriteServiceVariable(BLangService service, SymbolEnv env, BLangBlockStmt
             args.add(ASTBuilderUtil.createVariableRef(pos, service.serviceVariable.symbol));
 
             if (service.getServiceNameLiteral() == null) {
-                BLangListConstructorExpr.BLangArrayLiteral arrayLiteral =
-                        ASTBuilderUtil.createEmptyArrayLiteral(service.getPosition(), symTable.arrayStringType);
-                for (IdentifierNode path : service.getAbsolutePath()) {
-                    var literal = ASTBuilderUtil.createLiteral(path.getPosition(), symTable.stringType,
-                            path.getValue());
-                    arrayLiteral.exprs.add(literal);
+                if (service.getAbsolutePath().isEmpty()) {
```
MaryamZi, line 209: @rdhananjaya, hope this check here is sufficient.

tests/jballerina-unit-test/src/test/resources/test-src/services/service_attach_test.bal
```diff
@@ -0,0 +1,86 @@
...
+    public isolated function detach(service object {} s) returns error? { }
+
+    public isolated function attach(service object {} s, string[]? name) returns error? {
+        lock {
+            nameParamCheck[2] = name is string[] && name.length() == 2;
+        }
+    }
+}
+
+service /foo/bar on new Listener3() {
```
rdhananjaya, line 75: Shall we add another test case for service with just a '/`.
```
service / on new Listener4() {
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ServiceDesugar.java
rdhananjaya: Change looks good to me

tests/jballerina-unit-test/src/test/resources/test-src/services/service_attach_test.bal
gimantha: sure

rdhananjaya (approved):

tests/jballerina-unit-test/src/test/resources/test-src/services/service_attach_test.bal
```diff
@@ -0,0 +1,107 @@
...
+
+    public isolated function gracefulStop() returns error? { }
+
+    public isolated function immediateStop() returns error? { }
+
+    public isolated function detach(service object {} s) returns error? { }
+
+    public function attach(service object {} s, string[]? name) returns error? {
+            lock {
+                nameParamCheck[3] = name is string[] && name[0] == "/" && name.length() == 1;
```
MaryamZi, line 90: The spec says
> The attach-point determines the second argument passed to the attach method: if the attach-point is absent, then the argument is nil; if it is a string-literal, then the argument is a string; otherwise, **it is an absolute-resource-path and the argument is an array of strings, with one string for each resource-path-segment.**

In this case, is it correct for the array to have a single member "/" or should it be an empty array? :thinking:

rdhananjaya: Spec issue created as from I can remember, it was required to pass a string array with a single `/` member.
https://github.com/ballerina-platform/ballerina-spec/issues/810

rdhananjaya: Issue created to fix this spec deviation https://github.com/ballerina-platform/ballerina-lang/issues/29960

rdhananjaya: I think we have to remove the `if` block in line:209-212 of desugar class and align test cases with the change

gimantha: @rdhananjaya I did the necessary changes. But if we removed lines 209-212, we will not be handling the nil literal case right? Anyway did the changes, please review

rdhananjaya: Oh yes you are correct, my bad.

rdhananjaya (approved):

PR #29905, opened 2021-04-07
Fix subtyping rules with `never`

### LINKED ISSUE #29813: Fix subtyping rules with `never`
**Description:**
See https://github.com/ballerina-platform/ballerina-spec/issues/782

**Steps to reproduce:**
The following should be allowed.

```ballerina
function func() returns never {
    panic error("Error!");
}

function () returns int x = func;
```

**Affected Versions:**
slalpha4

### LINKED ISSUE #29822: Spread fields with optional fields of type `never` don't work as expected
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type Student record {
    string name;
};

type Grades record {|
    int physics;
    never name?;
    int...;
|};

public function main() {
    Grades grades = {
        physics: 75
    };

    Student john = {
        name: "John Doe",
        ...grades
    };
}
```

Results in a compilation error.
```cmd
ERROR [foo.bal:(18:12,18:18)] incompatible types: expected 'string' for field 'name', found 'never'
```

**Affected Versions:**
slalpha4-SNAPSHOT

lasinicl:
## Purpose
$title

Fixes #29813
Fixes #29822

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -645,6 +645,10 @@ private boolean isAssignable(BType source, BType target, Set<TypePair> unresolve
         int sourceTag = source.tag;
         int targetTag = target.tag;
 
+        if (isNeverTypeOrStructureTypeWithARequiredNeverMember(source) && isAssignable(target, symTable.anyType)) {
```
MaryamZi, line 648: Do we even need to check the target type?

Will this conflict with your other PR btw?

lasinicl: For any type `T`, the type `T|never` is the same as `T`. Therefore the target type can not be `error` type right? Therefore we do need to check the target type?

Nope. This check won't conflict with the other PR.

lasinicl: We don't need the target type it seems.
Yeah. then this check will conflict with the other PR. will update the other to use the same

grainier (commented):
LGTM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3844,7 +3848,7 @@ private BType createRecordIntersection(IntersectionContext diagnosticContext,
 
     private BType getConstraint(BRecordType recordType) {
         if (recordType.sealed) {
-            return symTable.neverType;
+            return symTable.noType;
```
grainier, line 3851: @MaryamZi can you please verify this change?

grainier (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
MaryamZi: @lasinicl can you share an example of why we needed this change, please?

lasinicl:
```
type ClosedRecordWithIntField record {|
    int i;
|};
ClosedRecordWithIntField val5 = {i: 100};
boolean b10 = val5 is record {| int i; string s; |}; //incompatible types: 'ClosedRecordWithIntField' will not be matched to 'record {| int i; string s; |}'
```
When there's `never` here, the above compilation error was not given.
https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/type-test-expr-negative.bal#L296

MaryamZi: This is happening due to the checks in org.wso2.ballerinalang.compiler.semantics.analyzer.Types#validateRecordField, right?

I think we should try and continue to use never here, and when validating the field explicitly check if the intersection type is never, and if not `Symbols.isOptional(origField.symbol)` return semantic error, or something similar to this.

Also, IMO, the behaviour should be the same for a closed record and an open record with a `never` rest field.

Both the type tests in the following example result in errors on alpha4, but only one results in an error on the current master.

```ballerina
type ClosedRecordWithIntField record {|
    int i;
|};

type RecordWithIntFieldAndNeverRestField record {|
    int i;
    never...;
|};

type Record record {| 
    int i; 
    string s; 
|};

public function main() {
    Record rec = {i: 1, s: ""};

    boolean b1 = rec is ClosedRecordWithIntField;
    boolean b2 = rec is RecordWithIntFieldAndNeverRestField; // no error
}
```

lasinicl: Okay. Will fix this with 29814 PR :)

Issue #29923, opened 2021-04-08
Fix module name & record variable name conflicts for configurable TOML


HindujaB:
**Description:**
https://github.com/ballerina-platform/ballerina-spec/blob/master/configurable/spec.md#toml-keys
A configurable variable can be specified in a TOML file in the following format,
```
    key:= [[orgName .] moduleName .] varName
```

At the same time, currently a record variable can be specified in the following format,

```
 key:= [[orgName .] moduleName .] varName [.fieldName ]*
```
This can cause conflicts, if the org/module name which can be optional and the variable/field name have the same name in a project. 

**Steps to reproduce:**

**Affected Versions:**
alpha4

**OS, DB, other environment details and versions:**

**Related Issues (optional):**
#28999 

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #29924, opened 2021-04-08
Add optional module-name support for configurable variables in root module

### LINKED ISSUE #28999: Support optional module names in configurable Toml
**Description:**
The module related validations for configurable TOML syntax is updated as this [1]. 
We need to modify the validation logics accordingly. 

[1] https://github.com/ballerina-platform/ballerina-spec/blob/master/configurable/spec.md#resolving-the-module-and-configurable-variable-from-a-toml-key

HindujaB:
## Purpose
$subject

Fixes #28999 

## Approach
As per spec, in TOML syntax for configurable variables, 
- The orgName and moduleName are optional only for configurable variables defined in the root module of the program
- The orgName is optional only for configurable variables defined in the root package of the program.
- For imported modules, the orgName, moduleName is mandatory
Addressed the orgName & moduleName ambiguities by validating them to have orgName & moduleName both.
 
## Remarks
The ambiguities between record names and module-names will be tracked in #29923

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


warunalakshitha:
Shall we have a code review for this PR.

bvm/ballerina-runtime/src/test/resources/config_files/ConfigClashingModule1.toml
```diff
@@ -0,0 +1,7 @@
+[rootOrg.test_module]
```
warunalakshitha, line 1: Lets add some tests with other format as well. ex rootOrg.test_module.intVar

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
```diff
@@ -234,48 +245,116 @@ private TomlNode getTomlNode(TomlNode tomlValue, String variableName, Type type)
...
-                throw new TomlConfigException(String.format(INVALID_MODULE_STRUCTURE, fullModuleName,
-                        fullModuleName), orgNode);
+        return table.map(Toml::rootNode).orElse(null);
+    }
+
+    private void analyseModuleAmbiguity() {
+        for (Module entry : moduleSet) {
+            String rootModuleName = rootModule.getName();
+            if (rootModuleName.startsWith(entry.getOrg())) {
+                hasModuleAmbiguity = true;
```
warunalakshitha, line 351: Will this make all the modules to be configured with org name if one module name has clashed with org name. Should we keep this per module.

bvm/ballerina-runtime/src/test/resources/config_files/ConfigClashingModule3.toml
```diff
@@ -0,0 +1,7 @@
+[rootOrg.test_module.util.foo]
```
warunalakshitha, line 1: Lets create issue for chek above cases with CLI values

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
```diff
@@ -234,48 +245,116 @@ private TomlNode getTomlNode(TomlNode tomlValue, String variableName, Type type)
...
     }
 
-    private TomlTableNode extractModuleNode(TomlTableNode orgNode, String moduleName, String fullModuleName,
-                                                   boolean hasRequired) {
-        if (orgNode == null) {
-            return null;
+    private TomlTableNode getRootModuleNode(Toml baseToml) {
+        String moduleName = rootModule.getName();
+        String moduleKey = getModuleKey(rootModule);
+        Optional<Toml> table = baseToml.getTable(moduleKey);
```
warunalakshitha, line 333: We need to support following case too.

abc =10
[root_module]
def = 10

so both abc and def variables should get value

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
HindujaB: This check considers the clash between org-name and root module-name. We mandate the org-name in this case. 
Therefore, if there's a clash then it mandates all the modules (as the non-default modules also starts with root module-name & org-name is always mandatory for imported packages)

So, I think we can keep this as a boolean.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
HindujaB: created issue #29989

bvm/ballerina-runtime/src/test/resources/config_files/ConfigClashingModule3.toml
HindujaB: created #29988

manuranga:
I created an issue regarding what we discussed in the code review. let's fix it in a separate pr.

manuranga (dismissed):

manuranga (approved):

PR #29964, opened 2021-04-09
Present the Original Value Symbols for Quoted Identifiers of Reserved Keywords

### LINKED ISSUE #29462: For variables named with ' prefix; ' is missed during suggestion
**Description:**
Screenshot 2021-03-23 at 15 42 10

Sample code:

```ballerina
function rate(Currency 'from, Currency to) {
    string fromCurrency = 'from;
    string toCurrency = to;
    string responseKey = fr
}
```

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

sanjana:
Fixes #29462

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


hasithaa:
Please add some test cases

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaMethodSymbol.java
```diff
@@ -60,7 +60,7 @@ public FunctionTypeSymbol typeDescriptor() {
 
     @Override
     public Optional<String> getName() {
-        return this.functionSymbol.getName();
+        return this.functionSymbol.getName().map(this::escapeReservedKeyword);
```
pubudu91, line 63: This is not necessary right? IINM, the name would already be escaped in the result returned by `functionSymbol.getName()`.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
```diff
@@ -65,7 +66,16 @@ protected BallerinaSymbol(String name, SymbolKind symbolKind, BSymbol symbol, Co
...
     @Override
     public Optional<String> getName() {
-        return Optional.ofNullable(this.name);
+        // In the langlib context, reserved keywords can be used as regular identifiers. Therefore, they will not be
+        // escaped.
+        if (getModule().isPresent() && getModule().get().id() != null &&
+                !getModule().get().id().moduleName().isEmpty() &&
+                getModule().get().id().moduleName().startsWith("lang.") &&
+                !getModule().get().id().orgName().isEmpty() &&
+                getModule().get().id().orgName().startsWith("ballerina")) {
```
pubudu91, line 75: Shall we refactor this a bit? Assign the returned value to a var and use, so that all these `getModule().get().id()` calls are simplified?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaMethodSymbol.java
sanjana: Fixed.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
sanjana: Updated.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
```diff
@@ -65,7 +67,16 @@ protected BallerinaSymbol(String name, SymbolKind symbolKind, BSymbol symbol, Co
...
+        // In the langlib context, reserved keywords can be used as regular identifiers. Therefore, they will not be
+        // escaped.
+        if (getModule().isPresent()) {
+            ModuleID moduleID = getModule().get().id();
+            if (moduleID != null && !moduleID.moduleName().isEmpty() && moduleID.moduleName().startsWith("lang.") &&
+                    !moduleID.orgName().isEmpty() && moduleID.orgName().startsWith("ballerina")) {
+                return Optional.ofNullable(this.name);
+            }
+        }
+        return Optional.ofNullable(escapeReservedKeyword(this.name));
```
pubudu91, line 79: The 2 conditions `moduleID != null && !moduleID.moduleName().isEmpty()` are unnecessary IINM. If there's a module symbol, there should always be a module ID. And since we are checking `startsWith()`, checking for an empty string is irrelevant right?

pubudu91: Another small point, shall we cache the escaped name? So that we don't have to compute it each time `getName()` is called. A symbol is immutable. So it won't change after creation. Refer to some of the other methods for examples of where this has been done.

sanjana: @pubudu91 thanks for the suggestion. 

Sure, that'd be a great idea to cache it.

sanjana: @pubudu91 used a caching mechanism. Also implemented it in the `BallerinaRecordFieldSymbol` and `BallerinaObjectFieldSymbol` classes as well.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
```diff
@@ -65,7 +68,21 @@ protected BallerinaSymbol(String name, SymbolKind symbolKind, BSymbol symbol, Co
...
     public Optional<String> getName() {
-        return Optional.ofNullable(this.name);
+        // In the langlib context, reserved keywords can be used as regular identifiers. Therefore, they will not be
+        // escaped.
+        if (this.escapedName != null) {
+            return Optional.of(this.escapedName);
+        }
+        if (getModule().isPresent()) {
+            ModuleID moduleID = getModule().get().id();
+            if (moduleID != null && moduleID.moduleName().startsWith("lang")
```
pubudu91, line 78: The `moduleID != null` check is also unnecessary since it's expected to be not null.

sanjana: Removed it.

pubudu91 (approved):

PR #29001, opened 2021-03-04
Modify Github workflow to publish code coverage reports for PRs.

anjaleeps:
## Purpose
$subject

Fixes #<Issue Number>

## Approach
Remove the separate jobs used for running integration and unit tests in the `CI_ubuntu.yml` and run the tests while building the module. After the build, generate Jacoco report and publish it to Codecov.

PR #30026, opened 2021-04-16
Fix ubuntu build failure

NipunaRanasinghe:
## Purpose
This PR fixes the intermittent CI build failures on ubuntu, which is caused by the code coverage report generation related recent changes.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/.github/workflows/CI_ubuntu.yml
+++ b/.github/workflows/CI_ubuntu.yml
@@ -13,7 +13,7 @@ jobs:
   ubuntu_build:
     name: Build with all tests on Ubuntu
     runs-on: ubuntu-latest
-    timeout-minutes: 75
+    timeout-minutes: 120
 
     steps:
       - uses: actions/checkout@v2


ThisaruGuruge (approved):

sameerajayasoma:
Do we know why we need to increase the timeout?

NipunaRanasinghe:
> Do we know why we need to increase the timeout?

@sameerajayasoma Yes we do :slightly_smiling_face: While troubleshooting the issue, I found out that its happening due to https://github.com/ballerina-platform/ballerina-lang/pull/29001, where we have merged all 3 previous ubuntu jobs into a single job and it has increased the job build time to ~75 mins (which is same as our build timeout threshold).

ThisaruGuruge:
> Do we know why we need to increase the timeout?

Previously, we had separate phases of the builds running parallel, so the build time for a given build is lesser.
But recently, as a part of the Code Coverage generation, we had to merge all these phases to a single one (Otherwise generating a CodeCov report was impossible). Therefore the build time has increased.

manuranga:
@sameerajayasoma 
I believe it actually it should takes *less* time after combining the steps, but maybe GitHub tracks time per step not per pr as a whole. @keizer619 what do you think?

PR #30043, opened 2021-04-18
Fix record hash mismatch

grainier:
## Purpose
$title

Fixes #<Issue Number>

## Approach
n/a

## Samples
n/a

## Remarks
n/a

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeHashVisitor.java
```diff
@@ -462,8 +462,8 @@ public Integer visit(BObjectType type) {
         List<Integer> typeInclHashes = getTypesHashes(type.typeInclusions);
         final int initFunctionHash = getFunctionHash(((BObjectTypeSymbol) type.tsymbol).initializerFunc);
         List<Integer> attachedFunctionsHashes = getFunctionsHashes(((BObjectTypeSymbol) type.tsymbol).attachedFuncs);
-        Integer hash = hash(baseHash(type), fieldsHashes, typeInclHashes, initFunctionHash,
-                attachedFunctionsHashes, type.typeIdSet);
+        Integer hash = hash(baseHash(type), fieldsHashes, typeInclHashes,
+                initFunctionHash, attachedFunctionsHashes, type.typeIdSet);
```
MaryamZi, line 466: The `init` function is not part of the shape. Should we remove this?

grainier: removed in: 9b8d292

manuranga (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeHashVisitor.java
manuranga: If two anon objects with different init function signatures are present, can't this create a runtime error?

MaryamZi: IIRC, our assumption was that this method wouldn't be used when creating values since we don't have default value information, right? If there were two different types with with the same fields/methods and init methods or no init method, I believe we return a single type anyway, so it can again return a value of an unexpected type?

But yes, if it is used to create a value at runtime, I guess it could result in an error. I think we can even consider the init method here, since a change to the init method can be considered a backward incompatible change.

manuranga: > wouldn't be used when creating values

Ya, It would be find in that case.

PR #30074, opened 2021-04-19
Add capability to manually close obs context

tharinduwijewardane:
## Purpose
We need to provide the lister with the capability to manually close the observer context after adding status code to resource span. It is required when the user has put a return statement instead of a caller->respond method in the resource function.


manuranga (approved):

nadundesilva (approved):
LGTM

Issue #29735, opened 2021-03-31
[MacOS] STEP_IN needs to be done twice for standard libraries


praveennadarajah:
**Description:**

We need to click the `STEP_IN` button twice if we want to go inside the standard libraries. This behaviour is only observed in MacOs and not in Linux.

**Steps to reproduce:**
- Consider the code segment:
```
import ballerina/io;

public function main() {
    io:println("Hello World!");
}
```
- Add a breakpoint on the line `io:println("Hello World!");`
- Run the program in debug mode.
- When the breakpoint gets hit, try to `STEP_IN` inside `io:println()` method.
- `STEP_IN` request will fail in the first attempt and it will be successful in the second attempt.

**Affected Versions:**
```
Ballerina Swan Lake Alpha 4
Language specification v2020-12-17
Update Tool 1.3.1
```

**OS, DB, other environment details and versions:**
```
macOS High Sierra
Version 10.13.6
```

PR #30094, opened 2021-04-20
Fix the double `STEP_IN` issue for some standard libraries in macOS

praveennadarajah:
## Purpose
This PR will fix the double `STEP_IN` issue for some standard libraries in macOS

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/29735

### DIFF
--- a/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
+++ b/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
@@ -299,7 +299,7 @@ public class JBallerinaDebugServer implements IDebugProtocolServer {
             // There are instances where we need to STEP_IN twice to actually step into a module.
             // In such case, we are comparing the previous breakpoint's top stack frame with the current breakpoint's
             // top stack frame, and if both are same we are sending a step request with stepType STEP_INTO.
-            StackFrame lastBreakpointTopStackFrame = context.getLastBreakpointTopStackFrame();
+            StackFrame lastBreakpointTopStackFrame = context.getLastDebugHitFrame();
             if (context.getLastInstruction() == DebugInstruction.STEP_IN
                 && validFrames.length > 0 && lastBreakpointTopStackFrame != null
                 && validFrames[0].getSource().getPath().equals(lastBreakpointTopStackFrame.getSource().getPath())
@@ -310,7 +310,7 @@ public class JBallerinaDebugServer implements IDebugProtocolServer {
                 context.setLastInstruction(null);
             }
             StackFrame currentBreakpointTopStackFrame = validFrames.length > 0 ? validFrames[0] : null;
-            context.setLastBreakpointTopStackFrame(currentBreakpointTopStackFrame);
+            context.setLastDebugHitFrame(currentBreakpointTopStackFrame);
             stackTraceResponse.setStackFrames(validFrames);
             return CompletableFuture.completedFuture(stackTraceResponse);
         } catch (JdiProxyException e) {
--- a/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
+++ b/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
@@ -142,7 +142,6 @@ public class JBallerinaDebugServer implements IDebugProtocolServer {
     private ThreadReferenceProxyImpl activeThread;
     private SuspendedContext suspendedContext;
     private boolean terminationRequestReceived = false;
-    private StackFrame previousTopFrame;
 
     private final AtomicLong nextVarReference = new AtomicLong();
     private final Map<Long, StackFrameProxyImpl> stackFramesMap = new HashMap<>();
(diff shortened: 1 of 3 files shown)


misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -296,6 +297,13 @@ ClientConfigHolder getClientConfigHolder() {
                     .filter(JBallerinaDebugServer::isValidFrame)
                     .toArray(StackFrame[]::new);
 
+            if (context.getLastInstruction() == DebugInstruction.STEP_IN
+                && validFrames[0].getSource().getPath().equals(topStackFrame.getSource().getPath())
```
NipunaRanasinghe, line 301: Potential `IndexOutOfBoundsException` as the `validFrames` can be an empty array

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -296,6 +297,13 @@ ClientConfigHolder getClientConfigHolder() {
...
+                && validFrames[0].getLine().intValue() == topStackFrame.getLine()) {
```
NipunaRanasinghe, line 302: Potential `IndexOutOfBoundsException` as the `validFrames` can be an empty array

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -296,6 +297,13 @@ ClientConfigHolder getClientConfigHolder() {
...
+                eventProcessor.sendStepRequest(args.getThreadId(), StepRequest.STEP_INTO);
+                context.setLastInstruction(null);
+            }
+            topStackFrame = validFrames[0];
```
NipunaRanasinghe, line 306: Potential `IndexOutOfBoundsException` as the `validFrames` can be an empty array

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -296,6 +297,13 @@ ClientConfigHolder getClientConfigHolder() {
                     .filter(JBallerinaDebugServer::isValidFrame)
                     .toArray(StackFrame[]::new);
 
+            if (context.getLastInstruction() == DebugInstruction.STEP_IN
+                && validFrames[0].getSource().getPath().equals(topStackFrame.getSource().getPath())
+                && validFrames[0].getLine().intValue() == topStackFrame.getLine()) {
+                eventProcessor.sendStepRequest(args.getThreadId(), StepRequest.STEP_INTO);
+                context.setLastInstruction(null);
```
NipunaRanasinghe, line 304: Why are we setting `null` as the last instruction?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -142,6 +142,7 @@
     private ThreadReferenceProxyImpl activeThread;
     private SuspendedContext suspendedContext;
     private boolean terminationRequestReceived = false;
+    private StackFrame topStackFrame;
```
NipunaRanasinghe, line 145: As this is basically contains the PREVIOUS breakpoint location information, shall we use a more meaningful name which indicates that?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Fixed

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Fixed

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Fixed

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Fixed

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Changed the name to `previousTopFrame`

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Here I'm making sure `STEP_IN` request is sent only once, and after the request is sent I'm making the last instruction to null. 

Alternatively, we can maintain a flag to control the number of `STEP_IN` request. WDYT?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
NipunaRanasinghe: Using the `topFrame` might still be misleading, when considering why we are using it. How about `lastLocationInfo`?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
NipunaRanasinghe: i"m still not clear how we are receiving multiple step in requests from the client for a single event.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Changed to `lastDebbugHitFrame`

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: We can get multiple `stackTrace` request for a single breakpoint event. Possible `stackTrace` requests are

1.  To represent debug hit navigation bar
2. To represent call stack information.

NipunaRanasinghe (approved):

NipunaRanasinghe (changes requested):
@praveennadarajah As per the offline discussion we had, let's move this logic to the `JDIEventProcessor` so that we can avoid additional stop event sent to the client and reduce the processing overhead.

NipunaRanasinghe (commented):
According to the review we had, let's hold this until we find a proper resolution for the regression issue (related to recursive functions).

Issue #30088, opened 2021-04-20
Arithmetic operations does not work for sub types


lasinicl:
**Description:**
$title

**Steps to reproduce:**
```ballerina
public const BT_NIL = 0;
public const BT_BOOLEAN = 1;
public const BT_INT = 2;
public const BT_STRING = 3;
public const BT_LIST = 4;

type BasicTypeCode BT_NIL|BT_BOOLEAN|BT_INT|BT_STRING|BT_LIST;
 
public function main() {
    int a = 10;
    BasicTypeCode b = 4;
    int x = a + b; // operator '+' not defined for 'int' and 'BasicTypeCode'
}
```
**Affected Versions:**
slalpha4

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Issue #30073, opened 2021-04-19
Fix error for types that are not yet supported with constant declaration


lasinicl:
**Description:**
$title
```ballerina
public const int:Signed32 BT_INT = 1;
```
ATM the above gives the following compilation error,
`constant cannot be defined with type 'int:Signed32', expected a simple basic types or a map of a simple basic type`
This needs to be updated.

**Describe your task(s)**

**Related Issues (optional):**
https://github.com/ballerina-platform/ballerina-lang/issues/21542

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

jclark:
There's no difference between:

```
public const int:Signed32 BT_INT = 1;
```

and

```
public const int BT_INT = 1;
```

and

```
public const BT_INT = 1;
```

See https://ballerina.io/spec/lang/master/#section_8.8

The type of a constant is always a singleton. In all the above cases, the type is `1`.

MaryamZi:
Agreed, @jclark. 

The issue with something like `public const int:Signed32 BT_INT = 1;` right now is that we haven't implemented support for defining constants with builtin subtypes yet, but the error we log is

```
ERROR [test.bal:(1:14,1:26)] constant cannot be defined with type 'int:Signed32', expected a simple basic types or a map of a simple basic type
```

even though this is valid according to the spec.

As the first step, we will be fixing this error to say it is not supported yet. This is also the case for some of the other types we don't support as constants yet.

Issue #30028, opened 2021-04-16
Avoid unnecessary casting with same types for relational expressions


lasinicl:
**Description:**
$title

**Describe your task(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Issue #29969, opened 2021-04-13
Shift operator does not work for int subtype


jclark:
This should work, but gets an error ` operator '<<' not defined for 'int' and 'BasicTypeCode'`

```
public const BT_NIL = 0;
public const BT_BOOLEAN = 1;
public const BT_INT = 2;
public const BT_STRING = 3;
public const BT_LIST = 4;

type BasicTypeCode BT_NIL|BT_BOOLEAN|BT_INT|BT_STRING|BT_LIST;

function includesAll(int bits, BasicTypeCode code) returns boolean {
  return (bits & (1 << code)) != 0;
}
```

jclark:
This problem seems to apply to other arithmetic operations too. Other subtypes of int don't seem to have this problem.

PR #30103, opened 2021-04-20
Fix several issues with binary operations

lasinicl:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/29969 Shift operator does not work for int subtype
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30088 Arithmetic operations does not work for sub types 
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30073 Fix error for types that are not yet supported with constant declaration
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30028 Avoid unnecessary casting with same types for relational expressions
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/28881 Implicit conversions between float and decimal should not be allowed
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30303 Operator '+' not defined for singleton and int

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -6528,6 +6540,26 @@ public void visit(BLangBinaryExpr binaryExpr) {
...
+                return;
+            }
+            if (TypeTags.isXMLTypeTag(rhsExprTypeTag) && !TypeTags.isXMLTypeTag(lhsExprTypeTag)) {
+                binaryExpr.lhsExpr = ASTBuilderUtil.createXMLTextLiteralNode(binaryExpr, binaryExpr.lhsExpr,
+                        binaryExpr.lhsExpr.pos, symTable.xmlType);
+                return;
+            }
+            binaryExpr.lhsExpr = createTypeCastExpr(binaryExpr.lhsExpr, binaryExpr.type);
+            binaryExpr.rhsExpr = createTypeCastExpr(binaryExpr.rhsExpr, binaryExpr.type);
+            return;
```
MaryamZi, line 6560: Shall we extract this out to a method?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -6528,6 +6540,26 @@ public void visit(BLangBinaryExpr binaryExpr) {
...
 
+        if (isArithmeticOperator) {
+            if ((TypeTags.isIntegerTypeTag(lhsExprTypeTag) && TypeTags.isIntegerTypeTag(rhsExprTypeTag)) ||
+                    (TypeTags.isStringTypeTag(lhsExprTypeTag) && TypeTags.isStringTypeTag(rhsExprTypeTag))) {
+                return;
+            }
+            if (TypeTags.isXMLTypeTag(lhsExprTypeTag) && !TypeTags.isXMLTypeTag(rhsExprTypeTag)) {
+                binaryExpr.rhsExpr = ASTBuilderUtil.createXMLTextLiteralNode(binaryExpr, binaryExpr.rhsExpr,
+                        binaryExpr.rhsExpr.pos, symTable.xmlType);
+                return;
```
MaryamZi, line 6551: Doesn't seem to be covered by tests?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -3250,7 +3250,14 @@ public void visit(BLangConstant constant) {
             dlog.error(constant.name.pos, DiagnosticErrorCode.UNDERSCORE_NOT_ALLOWED);
         }
         if (constant.typeNode != null && !types.isAllowedConstantType(constant.typeNode.type)) {
-            dlog.error(constant.typeNode.pos, DiagnosticErrorCode.CANNOT_DEFINE_CONSTANT_WITH_TYPE, constant.typeNode);
+            if (TypeTags.isIntegerTypeTag(constant.typeNode.type.tag) ||
+                    TypeTags.isStringTypeTag(constant.typeNode.type.tag)) {
```
MaryamZi, line 3254: Shall we do this for all the supported types from the spec?

>  The type-descriptor must specify a type that is a subtype of anydata and must not be never.

So basically, `types.isAssignable(type, symTable.anydataType) && !types.isNeverTypeOrStructureTypeWithARequiredNeverMember(type)`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: Please check other codecov comments too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3515,6 +3515,109 @@ private boolean equalityIntersectionExists(Set<BType> lhsTypes, Set<BType> rhsTy
...
+            case TypeTags.UNION:
+                BUnionType unionType = (BUnionType) type;
+                Set<BType> memberTypes = unionType.getMemberTypes();
+                BType firstTypeInUnion = memberTypes.iterator().next();
+                for (BType memType : memberTypes) {
+                    if (memType.tag == TypeTags.FINITE && firstTypeInUnion.tag == TypeTags.FINITE) {
+                        Set<BLangExpression> valSpace = ((BFiniteType) firstTypeInUnion).getValueSpace();
+                        BType baseExprType = valSpace.iterator().next().type;
+                        if (!checkValueSpaceHasSameType((BFiniteType) memType, baseExprType)) {
+                            return false;
```
MaryamZi, line 3532: `firstTypeInUnion` doesn't change after it is first set, right? Maybe we should switch the if check and the for loop, to avoid having to extract the value space/type and checking if the first member is finite for each member.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3515,6 +3515,109 @@ private boolean equalityIntersectionExists(Set<BType> lhsTypes, Set<BType> rhsTy
...
+                        return false;
+                    }
+                }
+                return true;
+            default:
+                return false;
+        }
+    }
+
+    private boolean isIntType(int firstTypeTag, int secondTypeTag) {
```
MaryamZi, line 3559: Shouldn't this be a method that accepts a single type, which gets called separately for `firstTypeTag` and `secondTypeTag`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3515,6 +3515,109 @@ private boolean equalityIntersectionExists(Set<BType> lhsTypes, Set<BType> rhsTy
...
+        }
+    }
+
+    boolean validStringTypeExists(BType type) {
+        if (TypeTags.isStringTypeTag(type.tag)) {
+            return true;
+        }
+        switch (type.tag) {
+            case TypeTags.XML:
+            case TypeTags.XML_TEXT:
```
MaryamZi, line 3598: Is this for interpolation? What about other kinds of XML?

I think we should also update the method name, since XML is not implicitly string.

lasinicl: This is for concatenation between `xml` and `string`. https://github.com/ballerina-platform/ballerina-lang/blob/616ea44722f3fd841f512345b9894f6e5b1cdf28/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolTable.java#L622

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -6535,6 +6543,10 @@ public void visit(BLangBinaryExpr binaryExpr) {
             return;
         }
 
+        if (isArithmeticOperator) {
```
MaryamZi, line 6546: Shouldn't we do this after L6558 and return?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -6545,32 +6557,94 @@ public void visit(BLangBinaryExpr binaryExpr) {
...
+        }
+
+        binaryExpr.lhsExpr = createTypeCastExpr(binaryExpr.lhsExpr, symTable.intType);
+        binaryExpr.rhsExpr = createTypeCastExpr(binaryExpr.rhsExpr, symTable.intType);
+    }
+
+    private void createTypeCastExprForRelationalExpr(BLangBinaryExpr binaryExpr, int lhsExprTypeTag,
+                                                                int rhsExprTypeTag) {
+        boolean isLhsIntegerType = TypeTags.isIntegerTypeTag(lhsExprTypeTag);
+        boolean isRhsIntegerType = TypeTags.isIntegerTypeTag(rhsExprTypeTag);
```
MaryamZi, line 6621: Can't/shouldn't we first check `isLhsIntegerType && isRhsIntegerType` and return?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -6545,32 +6557,94 @@ public void visit(BLangBinaryExpr binaryExpr) {
...
+        }
+
+        if (!isLhsIntegerType && isRhsIntegerType) {
+            binaryExpr.lhsExpr = createTypeCastExpr(binaryExpr.lhsExpr, symTable.intType);
+            return;
+        }
+
+        if (lhsExprTypeTag == TypeTags.BYTE || rhsExprTypeTag == TypeTags.BYTE) {
+            binaryExpr.rhsExpr = createTypeCastExpr(binaryExpr.rhsExpr, symTable.intType);
+            binaryExpr.lhsExpr = createTypeCastExpr(binaryExpr.lhsExpr, symTable.intType);
```
MaryamZi, line 6634: Do we need this even if both are byte?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: Same for string.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1120,6 +1120,10 @@ private void handleDeclaredWithVar(BLangVariable variable) {
             variable.type = symTable.semanticError;
             dlog.error(variable.pos, DiagnosticErrorCode.VARIABLE_DECL_WITH_VAR_WITHOUT_INITIALIZER);
         } else {
+            if (varRefExpr.getKind() == NodeKind.BINARY_EXPR &&
```
MaryamZi, line 1123: This is not an issue only when used with `var`. An additive expression `a + b` should fail if `a` and `b` are of different basic types.

Additive expression
> It is required that either
>
> - the static type of both operand expressions is a subtype of the same basic type, in which case this basic type will be the static type of the result, or
> - the static type of one operand expression must be a subtype of xml and of the other operand expression must be a subtype of string, in which case the static type of the result is xml

Multiplicative expression
> The static type of both operand expressions is required to be a subtype of the same basic type; this basic type will be the static type of the result.

```ballerina
const A = 10;
const B = 20;

type C A|B;

function testImplicitConversion() {
    float a = 1;
    decimal b = 1;
    any i1 = a + b; // No error, should fail
    var i2 = a + b; // ERROR: implicit type conversion not allowed here

    int d = 10;
    float e = 12.25;
    any i3 = d - e; // No error, should fail
    var i4 = d - e; // ERROR: implicit type conversion not allowed here
}
```

Also rather than saying "implicit type conversion not allowed here" I think we should stick to errors similar to the ones we have already "operator '+' not defined for 'float' and 'decimal'". IMO, this will be more helpful to the user.

misc/debug-adapter/modules/debug-adapter-runtime/src/main/ballerina/arithmetic_operations.bal
```diff
@@ -24,18 +24,6 @@ function add(any lhs, any rhs) returns any|error {
         result = trap (lhs + rhs); // float + float
     } else if (lhs is decimal && rhs is decimal) {
         result = trap (lhs + rhs); // decimal + decimal
-    } else if (lhs is int && rhs is float) {
```
lasinicl, line 27: @NipunaRanasinghe Please check the debugger related changes

tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/evaluation/ExpressionEvaluationTest.java
```diff
@@ -429,62 +429,28 @@ public void multiplicativeExpressionEvaluationTest() throws BallerinaTestExcepti
         ///////////////////////////////-----------multiplication----------------//////////////////////////////////////
         // int * int
         debugTestRunner.assertExpression(context, String.format("%s * %s", INT_VAR, INT_VAR), "400", "int");
-        // float * int
```
lasinicl, line 432: @NipunaRanasinghe Please check the debugger related changes

tests/jballerina-unit-test/src/test/resources/test-src/jvm/errors.bal
```diff
@@ -61,7 +61,7 @@ function testNestedCallsWithAllTraps(int y) {
 
 function validateError(int y, int|error x) {
     if (x is error) {
-        var expMessage = "reason foo " + y.toString();
+        string expMessage = "reason foo " + y.toString();
```
MaryamZi, line 64: Why do we need this change?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/binaryoperations/BitwiseShiftOperationTest.java
```diff
@@ -0,0 +1,92 @@
...
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+/**
+ * Test functionality of bitwise left shift, bitwise right shift and bitwise unsigned right shift operations.
+ *
+ * @since 2.0.0
+ */
+public class BitwiseShiftOperationTest {
```
MaryamZi, line 32: I think we already have tests under different tests. Shall we move them all to one file? We can do this separately.

tests/jballerina-unit-test/src/test/resources/test-src/jvm/errors.bal
lasinicl: We don't need this change. will revert this when addressing the remaining review suggestions 
Created https://github.com/ballerina-platform/ballerina-lang/issues/30313

MaryamZi (approved):

lasinicl:
Created #30313 to address the review suggestions

NipunaRanasinghe (dismissed):
Debugger related changes look good to me. However, we might need some further refactorings on the debugger side to disallow binary expression evaluations between unsupported union types. I've created https://github.com/ballerina-platform/ballerina-lang/issues/30314 to track the progress. Anyway, as this is more likely a debugger side limitation,  +1 to proceed with these fixes.

language-server/modules/langserver-core/src/test/java/org/ballerinalang/langserver/codeaction/TypeCastTest.java
```diff
@@ -55,7 +55,7 @@ public void negativeTest(String config, String source) throws IOException, Works
                 {"typeCast2.json", "typeCast.bal"},
                 {"typeCast3.json", "typeCast.bal"},
                 {"typeCast4.json", "typeCast.bal"},
-                {"typeCast5.json", "typeCast.bal"},
+                //{"typeCast5.json", "typeCast.bal"},
```
lasinicl, line 58: Created https://github.com/ballerina-platform/ballerina-lang/issues/30317

tests/jballerina-unit-test/src/test/resources/test-src/jvm/errors.bal
lasinicl: Fixed in 21f4087

MaryamZi (dismissed):

hasithaa (approved):

MaryamZi (approved):

Issue #30108, opened 2021-04-20
Enum Type Name Value is Incorrect


ThisaruGuruge:
**Description:**
When we call `getName()` or `getQualifiedName()` method on enum type the value returned is the full name of the type, unlike the other types.

Consider the following code:

```ballerina
type Person record {
    string name;
};

enum Color {
    RED,
    BLUE,
    GREEN
}
```

When we try to retrieve the name of these types in the native code as below:

```java
Type type = // get the type
String name = type.getName();
```

The value of the variable `name` will be `Person` for the record type, but the value of the `name` will be `<org_name>/<package_name>:<package_version>:<enum_name>`. This should return only the name.

ThisaruGuruge:
This is observed for UnionType as well. Consider the following (lousy) example:

```ballerina
type Person record {
    string name;
};

type Address record {
    string city;
};

type Information Person|Address;
```

The type name of the `Information` is also incorrect.

PR #30121, opened 2021-04-21
Fix name and qualified name for certain types

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30108

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BTupleType.java
```diff
@@ -135,11 +135,6 @@ public String toString() {
         return readonly ? toString + " & readonly" : toString;
     }
 
-    @Override
-    public String getName() {
-        return toString();
```
MaryamZi, line 140: @manuranga, @warunalakshitha, for certain types we seem to have returned the `toString()` representation as the name. If there is no type name, this would return the shape string. IMO, we can't consider the shape string as the type name, wdyt?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BType.java
```diff
@@ -133,11 +133,16 @@ public int hashCode() {
     }
 
     public String getName() {
-        return typeName;
+        return typeName == null ? "" : typeName;
     }
 
     public String getQualifiedName() {
-        return pkg == null ? typeName : pkg.toString() + ":" + typeName;
```
MaryamZi, line 140: In cases where the typeName is not present, this would return an invalid value, thus the change.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BType.java
```diff
@@ -133,11 +133,16 @@ public int hashCode() {
     }
 
     public String getName() {
-        return typeName;
+        return typeName == null ? "" : typeName;
```
MaryamZi, line 136: Need to decide if we should return null instead of "" if `typeName` is null.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BTupleType.java
warunalakshitha: +1. Shall we make this method final and remove the other override methods.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BType.java
warunalakshitha: Shall we make getQualifiedName also final?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BType.java
```diff
@@ -133,11 +133,16 @@ public int hashCode() {
...
     }
 
     public String getQualifiedName() {
-        return pkg == null ? typeName : pkg.toString() + ":" + typeName;
+        String name = getName();
+        if (name.isEmpty()) {
```
warunalakshitha, line 141: Instead of calling getName and do the empty name check again can we simply check typeName != null and return here?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BType.java
warunalakshitha: Can typeName can be null? Is that a bug?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BTupleType.java
MaryamZi: There seems to be one particular case where we need to override in `BMethodType` - io.ballerina.runtime.internal.types.BMethodType#getName

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BType.java
MaryamZi: We seem to have explicitly passed null in quite a few places, e.g., io.ballerina.runtime.internal.types.BTupleType#BTupleType(java.util.List<io.ballerina.runtime.api.types.Type>)

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BType.java
MaryamZi: I think there are places where empty string is explicitly set as the name too. org.wso2.ballerinalang.compiler.semantics.model.types.BType#BType(int, org.wso2.ballerinalang.compiler.semantics.model.symbols.BTypeSymbol)

warunalakshitha (approved):

Issue #27583, opened 2020-12-18
Checks don't run for local anonymous records that are part of another type


MaryamZi:
**Description:**
$title. Since local anonymous records aren't added as top level nodes until Desugar now, we need take to make sure the phases run for these local types.

**Steps to reproduce:**
The following results in a compilation error.
```ballerina
public function main() {
    record {|
        int i = "foo"; // incompatible types: expected 'int', found 'string'
    |} a;
}
```

But when the same record is a part of the element type of an array, there are **no compilation errors**.
```ballerina
public function main() {
    record {|
        int i = "foo"; // NO compilation error!
    |}[] b;
}
```

```cmd
$ ballerina run xyz.bal 

Compiling source
        xyz.bal
error: casting is not supported from 'string' to 'int'
```

**Affected Versions:**
slp8-SNAPSHOT pack

Issue #23847, opened 2020-06-09
Compiler doesn't complain about missing a variable with user-defined types


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type Foo record {|
    int i;
|};

public function main() {
    // int = int; // fails with invalid token '='
    // Foo = 1; // fails with expected 'typedesc<Foo>', found 'int'

    Foo = Foo; // doesn't fail!
}
```

Results in the runtime crashing.

ballerina-internal.log
```
[2020-06-09 16:21:46,526] SEVERE {b7a.log.crash} - org.ballerinalang.jvm.values.TypedescValue cannot be cast to org.ballerinalang.jvm.values.MapValue 
java.lang.ClassCastException: org.ballerinalang.jvm.values.TypedescValue cannot be cast to org.ballerinalang.jvm.values.MapValue
	at xyz.main(xyz.bal:9)
	at ___init.$lambda$main$(.)
	at org.ballerinalang.jvm.scheduling.SchedulerItem.execute(Scheduler.java:482)
	at org.ballerinalang.jvm.scheduling.Scheduler.run(Scheduler.java:263)
	at org.ballerinalang.jvm.scheduling.Scheduler.runSafely(Scheduler.java:231)
	at java.lang.Thread.run(Thread.java:745)
```

**Affected Versions:**
1.2.4

rdhananjaya:
Fixing this should potentially fix https://github.com/ballerina-platform/ballerina-lang/issues/30311
@lasinicl please test and close https://github.com/ballerina-platform/ballerina-lang/issues/30311 when you fix this

PR #30144, opened 2021-04-21
Fix issue with local anonymous records and no error when missing variable with user-defined types

lasinicl:
## Purpose
$title

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/27583 Checks don't run for local anonymous records that are part of another type
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23847 Compiler doesn't complain about missing a variable with user-defined types
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30145 Can't create an array of record with a closure type
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30170 Compiler doesn't complain when an uninitialized anonymous record has a closure type

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


chiranSachintha:
```Ballerina
public function main() {
    record {|
        int i = "foo";
    |} & readonly b;
}
``` 
This gives an error as ` casting is not supported from 'string' to 'int'`.  This also should be handled. shouldn't it?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1468,7 +1527,14 @@ public void visit(BLangAssignment assignNode) {
...
 
-        typeChecker.checkExpr(assignNode.expr, this.env, expType);
+        BType typeChecked = typeChecker.checkExpr(assignNode.expr, this.env, expType);
+
+        if (typeChecked != symTable.semanticError && varRef.getKind() == NodeKind.SIMPLE_VARIABLE_REF) {
+            BLangSimpleVarRef variableRef = (BLangSimpleVarRef) varRef;
+            if (variableRef.lhsVar && variableRef.type.tsymbol == null && variableRef.type.tag == TypeTags.TYPEDESC) {
+                dlog.error(variableRef.pos, DiagnosticErrorCode.INVALID_VARIABLE_REFERENCE);
+            }
+        }
```
chiranSachintha, line 1538:
```ballerina
type x record {|
|};
public function main() {
    x = 5;
}
```
I think this is not the correct fix. For the above code, It gives an error as `incompatible types: expected 'typedesc<x>', found 'int'` (Here, the issue is `x`(type definition) taken as the symbol of the variable reference `x`(x = 5). We need to fix it(#30311)). @MaryamZi

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -858,6 +869,54 @@ public void visit(BLangSimpleVariable varNode) {
         transferForkFlag(varNode);
     }
 
+    private void analyzeTypeNode(BLangType typeNode, SymbolEnv env) {
+        if (typeNode.getKind() == NodeKind.USER_DEFINED_TYPE) {
+            return;
+        }
+        if (typeNode.getKind() == NodeKind.RECORD_TYPE && !((BLangRecordTypeNode) typeNode).analyzed) {
+            analyzeDef(typeNode, env);
```
MaryamZi, line 877: Shouldn't we return here?

If `typeNode.getKind() == NodeKind.RECORD_TYPE` is true and `((BLangRecordTypeNode) typeNode).analyzed` is also true, we can still return here, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -858,6 +869,54 @@ public void visit(BLangSimpleVariable varNode) {
...
+        }
+        switch (typeNode.type.tag) {
```
MaryamZi, line 879: Don't we have to check error, future, function, and typedesc too?

I assume objects will work because of the module-level type definitions that are added, but let's add tests for them too, if we don't have already.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/varref/RecordVariableReferenceTest.java
```diff
@@ -215,16 +215,11 @@ public void testRecordVariablesSemanticsNegative() {
...
-        BAssertUtil.validateError(resultSemanticsNegative, ++i,
-                "invalid binding pattern, variable reference 'm[var2]' cannot be used with binding pattern", 198, 36);
-        BAssertUtil.validateError(resultSemanticsNegative, ++i, "cannot assign a value to final 's'", 213, 6);
-        BAssertUtil.validateError(resultSemanticsNegative, ++i, "cannot assign a value to final 'i'", 213, 9);
-        BAssertUtil.validateError(resultSemanticsNegative, ++i, "cannot assign a value to final 'f'", 213, 12);
-        BAssertUtil.validateError(resultSemanticsNegative, ++i, "cannot assign a value to final 's2'", 231, 6);
-        BAssertUtil.validateError(resultSemanticsNegative, ++i, "cannot assign a value to final 'iv'", 231, 19);
-        BAssertUtil.validateError(resultSemanticsNegative, ++i, "cannot assign a value to final 'b3'", 231, 23);
-        BAssertUtil.validateError(resultSemanticsNegative, ++i, "cannot assign a value to final 'm2'", 231, 31);
+                "invalid expr in assignment lhs", 186, 51);
```
MaryamZi, line 218: I don't think these tests belong here. Record variable reference tests test destructuring and record binding patterns IINM.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: Yeah I think we need to say we can't assign to `x` here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1468,7 +1527,14 @@ public void visit(BLangAssignment assignNode) {
         setTypeOfVarRefInAssignment(varRef);
         expType = varRef.type;
 
-        typeChecker.checkExpr(assignNode.expr, this.env, expType);
+        BType typeChecked = typeChecker.checkExpr(assignNode.expr, this.env, expType);
+
+        if (typeChecked != symTable.semanticError && varRef.getKind() == NodeKind.SIMPLE_VARIABLE_REF) {
```
MaryamZi, line 1532: The following crashes at runtime.

```ballerina
type x record {|
|};

public function main() {
    int a;
    [x, a] = [x, 1];
}
```

```log
[2021-05-04 00:06:49,954] SEVERE {b7a.log.crash} - class $typedesc$x cannot be cast to class io.ballerina.runtime.internal.values.MapValue ($typedesc$x and io.ballerina.runtime.internal.values.MapValue are in unnamed module of loader 'app') 
java.lang.ClassCastException: class $typedesc$x cannot be cast to class io.ballerina.runtime.internal.values.MapValue ($typedesc$x and io.ballerina.runtime.internal.values.MapValue are in unnamed module of loader 'app')
	at foo.main(foo.bal:6)
	at $_init.$lambda$main$(.)
	at io.ballerina.runtime.internal.scheduling.SchedulerItem.execute(Scheduler.java:571)
	at io.ballerina.runtime.internal.scheduling.Scheduler.run(Scheduler.java:301)
	at io.ballerina.runtime.internal.scheduling.Scheduler.runSafely(Scheduler.java:269)
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -858,6 +870,71 @@ public void visit(BLangSimpleVariable varNode) {
...
+                    analyzeTypeNode(errorType.detailType, env);
+                }
+                break;
+            case FUNCTION_TYPE:
+                if (typeNode.getKind() == NodeKind.FUNCTION_TYPE) {
+                    BLangFunctionTypeNode functionTypeNode = (BLangFunctionTypeNode) typeNode;
+                    List<BLangVariable> params = functionTypeNode.params;
+                    for (BLangVariable param : params) {
+                        analyzeTypeNode(param.typeNode, env);
+                    }
```
chiranSachintha, line 926: Here we can define defaultable parameters with various types. Is this the correct approach to handle all of those scenarios? 
Example.

```ballerina
public function main() {
    function(string b = 5) a;
    function(never b = 5) b;
    function(xml b = 5) c;
}
```

lasinicl: This fix was done considering only the local anonymous records. Can you create an issue for this?

lasinicl: For this particular fix, I was just focusing on local anonymous records. Shall we create an issue for this too?

chiranSachintha: Will create an issue.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1479,6 +1565,14 @@ public void visit(BLangAssignment assignNode) {
     public void visit(BLangTupleDestructure tupleDeStmt) {
         for (BLangExpression tupleVar : tupleDeStmt.varRef.expressions) {
             setTypeOfVarRefForBindingPattern(tupleVar);
+            if (tupleVar.getKind() == NodeKind.SIMPLE_VARIABLE_REF) {
+                BLangSimpleVarRef variableRef = (BLangSimpleVarRef) tupleVar;
+                if (variableRef.lhsVar && variableRef.symbol != null &&
+                        (variableRef.symbol.tag & SymTag.TYPE_DEF) == SymTag.TYPE_DEF) {
+                    dlog.error(tupleVar.pos, DiagnosticErrorCode.CANNOT_ASSIGN_VALUE_TO_TYPE_DEF);
+                    return;
```
MaryamZi, line 1573: Why are we returning here? Can't we continue the rest of the analysis? Same with other changes too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1479,6 +1565,14 @@ public void visit(BLangAssignment assignNode) {
     public void visit(BLangTupleDestructure tupleDeStmt) {
```
MaryamZi, line 1565: Shouldn't we do this for mapping and error too?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1479,6 +1565,14 @@ public void visit(BLangAssignment assignNode) {
...
         for (BLangExpression tupleVar : tupleDeStmt.varRef.expressions) {
             setTypeOfVarRefForBindingPattern(tupleVar);
+            if (tupleVar.getKind() == NodeKind.SIMPLE_VARIABLE_REF) {
```
MaryamZi, line 1568: What if it is not simple var ref but in the member var refs has an invalid assignment?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -858,6 +870,71 @@ public void visit(BLangSimpleVariable varNode) {
...
+                List<BLangType> memberTypeNodes = tupleTypeNode.memberTypeNodes;
+                for (BLangType memType : memberTypeNodes) {
+                    analyzeTypeNode(memType, env);
+                }
+                if (tupleTypeNode.restParamType != null) {
+                    analyzeTypeNode(tupleTypeNode.restParamType, env);
+                }
+                break;
+            case TABLE_TYPE:
+                analyzeTypeNode(((BLangTableTypeNode) typeNode).constraint, env);
```
MaryamZi, line 897: Can the key be a record?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1468,6 +1545,15 @@ public void visit(BLangAssignment assignNode) {
         setTypeOfVarRefInAssignment(varRef);
         expType = varRef.type;
 
+        if (varRef.getKind() == NodeKind.SIMPLE_VARIABLE_REF) {
+            BLangSimpleVarRef variableRef = (BLangSimpleVarRef) varRef;
+            if (variableRef.lhsVar && variableRef.symbol != null &&
+                    (variableRef.symbol.tag & SymTag.TYPE_DEF) == SymTag.TYPE_DEF) {
```
MaryamZi, line 1551: Need to do this for all module level final constructs like functions too. We can fix that separately maybe. Let's create an issue.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -858,6 +870,71 @@ public void visit(BLangSimpleVariable varNode) {
...
+                }
+                if (tupleTypeNode.restParamType != null) {
+                    analyzeTypeNode(tupleTypeNode.restParamType, env);
+                }
+                break;
+            case TABLE_TYPE:
+                analyzeTypeNode(((BLangTableTypeNode) typeNode).constraint, env);
+                break;
+            case INTERSECTION_TYPE_NODE:
+                List<BLangType> constituentTypeNodes = ((BLangIntersectionTypeNode) typeNode).constituentTypeNodes;
```
MaryamZi, line 900: Let's also analyze the effective type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -858,6 +870,71 @@ public void visit(BLangSimpleVariable varNode) {
...
+                BLangErrorType errorType = (BLangErrorType) typeNode;
+                if (errorType.detailType != null) {
+                    analyzeTypeNode(errorType.detailType, env);
+                }
+                break;
+            case FUNCTION_TYPE:
+                if (typeNode.getKind() == NodeKind.FUNCTION_TYPE) {
+                    BLangFunctionTypeNode functionTypeNode = (BLangFunctionTypeNode) typeNode;
+                    List<BLangVariable> params = functionTypeNode.params;
+                    for (BLangVariable param : params) {
```
MaryamZi, line 924: This doesn't include the rest param, does it?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
lasinicl: `BLangIntersectionTypeNode` does not contain information about the effective type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
lasinicl: Nope. A `key` should be a `field`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
lasinicl: Created https://github.com/ballerina-platform/ballerina-lang/issues/30564

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -830,6 +837,11 @@ public void visit(BLangSimpleVariable varNode) {
                         DiagnosticErrorCode.CONFIGURABLE_VARIABLE_CURRENTLY_NOT_SUPPORTED, lhsType);
             }
         }
+
+        if (varNode.typeNode != null) {
+            analyzeTypeNode(varNode.typeNode, env);
```
MaryamZi, line 842: IMO we should try and call analyzeDef for the type nodes or we aren't quite making use of the visitor pattern. I guess we can move the check for org.wso2.ballerinalang.compiler.tree.types.BLangRecordTypeNode#analyzed into the visit of `BLangRecordTypeNode` itself.

Let's do this separately.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1477,6 +1557,14 @@ public void visit(BLangAssignment assignNode) {
         setTypeOfVarRefInAssignment(varRef);
         expType = varRef.type;
 
+        if (varRef.getKind() == NodeKind.SIMPLE_VARIABLE_REF) {
+            BLangSimpleVarRef variableRef = (BLangSimpleVarRef) varRef;
```
MaryamZi, line 1561: Can't we call `checkInvalidTypeDef` here also?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1488,6 +1576,7 @@ public void visit(BLangAssignment assignNode) {
     public void visit(BLangTupleDestructure tupleDeStmt) {
         for (BLangExpression tupleVar : tupleDeStmt.varRef.expressions) {
             setTypeOfVarRefForBindingPattern(tupleVar);
+            checkInvalidTypeDef(tupleVar);
```
MaryamZi, line 1579: Shouldn't we check for the rest param also?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1543,6 +1635,7 @@ public void visit(BLangErrorDestructure errorDeStmt) {
             if (varRef.cause.getKind() != NodeKind.SIMPLE_VARIABLE_REF ||
                     names.fromIdNode(((BLangSimpleVarRef) varRef.cause).variableName) != Names.IGNORE) {
                 setTypeOfVarRefInErrorBindingAssignment(varRef.cause);
+                checkInvalidTypeDef(varRef.cause);
```
MaryamZi, line 1638: What about the detail bindings? Probably happens via org.wso2.ballerinalang.compiler.semantics.analyzer.SemanticAnalyzer#checkErrorVarRefEquivalency?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1557,13 +1557,7 @@ public void visit(BLangAssignment assignNode) {
         setTypeOfVarRefInAssignment(varRef);
```
MaryamZi, line 1557: Shall we add tests for the changes introduced in this commit?

MaryamZi (dismissed):
@lasinicl, shall we create high priority issues for the pending tasks?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
lasinicl: Created https://github.com/ballerina-platform/ballerina-lang/issues/30575

MaryamZi (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
lasinicl: Created https://github.com/ballerina-platform/ballerina-lang/issues/30577

PR #30159, opened 2021-04-22
Fix inconsistencies in `lang.xml:strip()`

### LINKED ISSUE #24200: Issues with `lang.xml:strip()`
**Description:**
The spec says

```
# Strips the insignificant parts of the an xml value.
# Comment items, processing instruction items are considered insignificant.
# After removal of comments and processing instructions, the text is grouped into
# the biggest possible chunks (i.e. only elements cause division into multiple chunks)
# and a chunk is considered insignificant if the entire chunk is whitespace.
#
# + x - the xml value
# + return - `x` with insignificant parts removed
public function strip(xml x) returns xml = external;
```

**Steps to reproduce:**
But `strip()` in the current implementation doesn't seem to actually strip. Nor does it panic when attempted on an immutable value (since it is supposed to modify the original value itself).

```ballerina
import ballerina/io;

public function main() {
    xml a = xml `<foo><bar/><!--Comment--></foo>`;
    xml b = a.strip();

    io:println(a); // <foo><bar/><!--Comment--></foo>
    io:println(b); // <foo><bar/><!--Comment--></foo> - NOT stripped!

    xml c = a.cloneReadOnly();
    xml d = c.strip(); // doesn't panic!
}
```

**Affected Versions:**
1.2.4, Current master.

suleka96:
Fixes #24200

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


rdhananjaya (commented):
As the function comment suggests, I think we need to check for the white spaces [1] and skip them.

[1] - [w3.org/TR/xml/#sec-white-space](w3.org/TR/xml/#sec-white-space)

rdhananjaya:
> As the function comment suggests, I think we need to check for the white spaces [1] and skip them.
> 
> [1] - [w3.org/TR/xml/#sec-white-space](w3.org/TR/xml/#sec-white-space)

Oh my bad, this is on the old function doc, in the current spec it doesn't say this.

langlib/langlib-test/src/test/resources/test-src/xmllib_test.bal
```diff
@@ -214,6 +214,35 @@ function testXMLCycleErrorInner() returns xml {
...
+    assertEquals(x2.toString(), "");
+    'xml:ProcessingInstruction x9 = xml `<?foo?>`;
+    x2 = x9.strip();
+    assertEquals(x2.toString(), "");
+
+    xml x5 = xml `<foo><!--Comment--></foo>text1 <?foo?>text2<foo><?foo?></foo>text3text4  <foo></foo><!--Comment-->  `;
+    x2 = x5.strip();
+    assertEquals(x2.toString(), "<foo></foo>text1 text2<foo></foo>text3text4  <foo></foo>");
+}
+
```
rdhananjaya, line 245: shall we add a test case that tries to perform this on a readonly xml value

rdhananjaya:
Plese take a look at https://github.com/ballerina-platform/ballerina-spec/issues/834

suleka96:
As per the clarification, I have removed the extra code that was put to look inside xml elements

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlSequence.java
```diff
@@ -321,6 +321,9 @@ public void addChildren(BXml xmlItem) {
      */
     @Override
     public XmlValue strip() {
+        if (this.isFrozen()) {
```
rdhananjaya, line 324: Strip create a new sequence out from this one, so we don't need to do this IMO

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
```diff
@@ -391,8 +391,8 @@ private boolean isLastChildIsTextNode(List<BXml> childList) {
      */
     @Override
     public XmlValue strip() {
-        children.strip();
-        return this;
+       this.children = (XmlSequence) children.strip();
```
rdhananjaya, line 394: Shouldn't this be an empty method, as calling strip on a single element should return it self

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlSequence.java
suleka96: does that mean that ```xml x5 = xml `<foo><bar/><?foo?></foo><?foo?>text1 text2<!--Com1--> <foo><bar/></foo>`;
  xml x2 = (x5.cloneReadOnly()).strip();``` should also be possible

rdhananjaya: Compilation error is wrong as `strip` function is does not change the original value, it just create a new value skipping unwanted things.

rdhananjaya:
Changes looks good to me

rdhananjaya (approved):

Issue #30169, opened 2021-04-23
Getting error when we configure a record type variable which has record fields with default values


warunalakshitha:
**Description:**
Please refer to the following record.

```ballerina
public type Person record {
    string name;
    int id;
    Address address = {};
};

public type Address record {
    string city = "Nugeoda";
};
```

In config.toml

```
[configProject.person]
name = "waruna"
id = 10
```

Getting following error 

error: {ballerina/lang.map}InherentTypeViolation {"message":"invalid value for record field 'address': expected value of type '(configProject:Address & readonly)', found 'configProject:Address'"}
        at waruna.configProject.0_1_0.Person:Person$gen$.<init>(main.bal:9)

warunalakshitha:
The fix is merged to the master hence closing the issue.

PR #30171, opened 2021-04-23
Fix issue with record field with default value

warunalakshitha:
## Purpose
Resolves https://github.com/ballerina-platform/ballerina-lang/issues/30169

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
@@ -477,8 +477,8 @@ public class TomlProvider implements ConfigProvider {
             recordName = type.getName();
             recordType = (RecordType) type;
         } else {
-            recordName = ((BIntersectionType) type).getEffectiveType().getName();
-            recordType = (RecordType) ((BIntersectionType) type).getEffectiveType();
+            recordType = (RecordType) ((BIntersectionType) type).getConstituentTypes().get(0);
+            recordName = recordType.getName();
         }
         if (tomlNode.kind() != getEffectiveTomlType(recordType, variableName)) {
             throw new TomlConfigException(String.format(INCOMPATIBLE_TYPE_ERROR_MESSAGE, variableName, recordType ,
--- a/tests/jballerina-integration-test/src/test/resources/configurables/configStructuredTypesProject/Config.toml
+++ b/tests/jballerina-integration-test/src/test/resources/configurables/configStructuredTypesProject/Config.toml
@@ -105,8 +105,17 @@ id=888
 [configStructuredTypes.person]
 name = "waruna"
 id = 10
-address.city="Colombo"
-address.country.name="Sri Lanka"
+address.city="San Francisco"
+address.country.name="USA"
+
+[configStructuredTypes.person2]
+name = "manu"
+id = 11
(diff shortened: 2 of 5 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MapUtils.java
```diff
@@ -82,16 +83,22 @@ public static void handleInherentTypeViolatingRecordUpdate(MapValue mapValue, BS
...
-
-                throw ErrorCreator.createError(
-                        getModulePrefixedReason(MAP_LANG_LIB, INHERENT_TYPE_VIOLATION_ERROR_IDENTIFIER),
-                        BLangExceptionHelper.getErrorMessage(RuntimeErrors.RECORD_INVALID_READONLY_FIELD_UPDATE,
-                                                             fieldName, recType));
+            if (SymbolFlags.isFlagOn(recField.getFlags(), SymbolFlags.READONLY)) {
+                if (initialValue) {
+                    recFieldType = recField.getFieldType();
+                    if (recFieldType.getTag() == TypeTags.INTERSECTION_TAG) {
+                        recFieldType = ((BIntersectionType) recField.getFieldType()).getConstituentTypes().get(0);
```
MaryamZi, line 90: I don't think this is correct. Does the following still fail?

```ballerina
type Foo record {|
    int[] & readonly x?;
    int[]...; 
|};

public function main() {
    string str = "x";
    int[] arr = [1, 2];
    Foo foo = {[str]: arr};
}
```

MaryamZi: The above will work as expected since it is not a readonly record. The following will probably fail (won't panic but it should panic).

```ballerina
type Person record {|
    int[] x = [];
|};

public function main() {
    Person & readonly y = {};
}
```

warunalakshitha: I have refactor the fix to address above two cases as well.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
```diff
@@ -477,8 +477,8 @@ private BArray getRecordArray(TomlNode tomlValue, String variableName, ArrayType
             recordName = type.getName();
             recordType = (RecordType) type;
         } else {
-            recordName = ((BIntersectionType) type).getEffectiveType().getName();
-            recordType = (RecordType) ((BIntersectionType) type).getEffectiveType();
+            recordType = (RecordType) ((BIntersectionType) type).getConstituentTypes().get(0);
```
MaryamZi, line 480: Shall we create an issue to revisit the assumption that this will always be the first member?

MaryamZi (approved):
Change LGTM, since `freezeDirect()` is called later on. 

@warunalakshitha, shall we add tests? Shall we also add one where an anonymous record is used in the intersection?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/toml/TomlProvider.java
warunalakshitha: Created https://github.com/ballerina-platform/ballerina-lang/issues/30178

PR #30181, opened 2021-04-23
Enable disabled identifier literal integration test

### LINKED ISSUE #26933: Enable Identifire test testPackageIDClash in Project API branch
**Description:**
$subject. The above test was disable in project API branch because the new project API doesn't support path dependencies. Creating this issue to track it so we can re enable once we support path API changes.

HindujaB:
## Purpose
$subject

Fixes #26933 

## Approach
Removed `path` configuration in `Dependencies.toml`.
Instead of that, `bal push --repository=local` command is used for importing packages.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


azinneera (approved):
Usage of Project API LGTM

tests/jballerina-integration-test/src/test/resources/identifier/PackageNameClashProject/main.bal
```diff
@@ -14,6 +14,10 @@
 // specific language governing permissions and limitations
 // under the License.
 
+import a_b/foo;
+import ballerina/test;
```
warunalakshitha, line 18: Previously we had imported from the same project module and different project modules. Why we have change the example?

import a_b/foo;
import a/main.b_foo;

tests/jballerina-integration-test/src/test/java/org/ballerinalang/test/identifier/IdentifierLiteralTest.java
```diff
@@ -56,17 +58,22 @@ public void testModuleIdentifierClash() throws BallerinaTestException {
...
-        bMainInstance.runMain("build", new String[0], new HashMap<>(), new String[0],
-                new LogLeecher[]{}, importProjectPath.toString());
-        bMainInstance.runMain("build", new String[0], new HashMap<>(), new String[0],
-                new LogLeecher[]{runLeecher}, projectPath.toString());
+        String importProjectPath = testFileLocation + File.separator + "testProject";
+        LogLeecher buildLeecher = new LogLeecher("target/bala/a_b-foo-any-0.1.0.bala");
+        LogLeecher pushLeecher =
+                new LogLeecher("Successfully pushed target/bala/a_b-foo-any-0.1.0.bala to 'local' repository.", ERROR);
+        LogLeecher runLeecher = new LogLeecher("Tests passed");
+        bMainInstance.runMain("build", new String[]{"-c"}, null, null, new LogLeecher[]{buildLeecher},
```
warunalakshitha, line 68: Previously we tested with ballerina test command and now have been changes to run command.

warunalakshitha: Shall we use the same approch?

tests/jballerina-integration-test/src/test/resources/identifier/PackageNameClashProject/main.bal
HindujaB: This test was disabled before the slp5 Project API changes, where we didn't have a root module concept.
Therefore it was checked using two different imported modules that are used in another module (`main`)
`a/b_foo` and `a_b/foo`
If we have a non-root module with name `b_foo`, it will always be `a/main.b_foo` and not cause a clash.
Therefore, I changed the root module name to `b_foo`

Added another non-root module with the same name in latest commit

warunalakshitha (approved):

Issue #28359, opened 2021-02-02
Error when using `cloneWithType` with a structure that has an immutable value


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/io;

type Map map<anydata>;

public function main() {
    map<int> & readonly m = {a: 1, b: 2};
    map<map<int>> n = {m}; // or `m: m.cloneReadOnly()` if `m` is mutable

    var o = checkpanic n.cloneWithType(Map);

    io:println(o["m"] === m); // Should be false.
    io:println(o["m"].isReadOnly()); // Should be false.
}
```
```cmd
$ bal run xyz.bal 

Compiling source
        xyz.bal

Running executable

error: {ballerina/lang.map}InvalidUpdate {"message":"Invalid map insertion: modification not allowed on readonly value"}
        at ballerina.lang.value.1_0_0:cloneWithType(value.bal:84)
           xyz:main(xyz.bal:9)
```

**Affected Versions:**
slalpha1

PR #30213, opened 2021-04-27
Fix error when using cloneWithType with a structure that has an immutable value 

warunalakshitha:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/28359

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


manuranga (dismissed):

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -828,12 +829,12 @@ private static boolean isFloatWithinIntRange(double doubleValue) {
...
         if (value instanceof ArrayValue && ((ArrayValue) value).getType().getTag() == TypeTags.ARRAY_TAG &&
                 !isDeepConversionRequiredForArray(((ArrayValue) value).getType())) {
-            return ((ArrayValue) value).getType();
+            return TypeCreator.createArrayType(((BArrayType) ((ArrayValue) value).getType()).getElementType());
         }
 
         if (value instanceof MapValue && ((MapValue) value).getType().getTag() == TypeTags.MAP_TAG &&
                 !isDeepConversionRequiredForMap(((MapValue) value).getType())) {
-            return ((MapValue) value).getType();
+            return TypeCreator.createMapType(((BMapType) ((MapValue) value).getType()).getConstrainedType());
```
MaryamZi, line 837: Shouldn't the type come from the target type?

The following fails atm.

```ballerina
type Map map<anydata>;

public function main() {
    map<int> & readonly m = {a: 1, b: 2};
    map<map<int>> n = {m};

    var o = checkpanic n.cloneWithType(Map);

    anydata p = o["m"];

    if p is map<anydata> {
        p["c"] = "foo";
    }
}
```

```cmd
error: {ballerina/lang.map}InherentTypeViolation {"message":"invalid map insertion: expected value of type 'int', found 'string'"}
        at foo:main(foo.bal:12)
```

This happens even when the original value is mutable (on slalpha5 also).

warunalakshitha: In this case what would be the type of o? AFAIU it should be map<map<int>> not map<anydata>. Based on that, type should be coming from value not target type.

As per spec When `v` is a structural value, the inherent type of the value to be constructed comes from `t`. When `t` is a union, it must be possible to determine which member of the union to use for the inherent type by following the same rules  that are used by list constructor expressions and mapping constructor expressions with the contextually expected type.

warunalakshitha: p["c"] = 123 will work

MaryamZi: Type of `o` should be `Map`, which is `map<anydata>`, right? 

```
# When `v` is a structural value, the inherent type of the value to be constructed
# comes from `t`.
```
`t` here is `map<anydata>` so the inherent type is `map<anydata>`?

```
# The `cloneWithType` operation is recursively applied to each member of `v` using
# the type descriptor that the inherent type requires for that member.
```
When we get to the member `m`, the type required for that is `anydata`, so shouldn't the result be the same as calling `m.cloneWithType(anydata)`?

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -833,6 +833,27 @@ function testCloneWithTypeWithInferredArgument() {
...
+   anydata p = o["m"];
+   assert(p is map<anydata>, true);
+   map<anydata> q = <map<anydata>>p;
+   q["c"] = "foo";
+
+   int[] & readonly array = [3, 4];
+   map<int[]> n2 = {array};
+   var o2 = checkpanic n2.cloneWithType(Map);
+   assert(o2["array"] === array, false);
+   assert(o2["array"].isReadOnly(), false);
```
MaryamZi, line 854: Shall we try adding a non-int value here also?

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -833,6 +833,27 @@ function testCloneWithTypeWithInferredArgument() {
...
+
+   map<int> & readonly m = {a: 1, b: 2};
+   map<map<int>> n = {m};
+   var o = checkpanic n.cloneWithType(Map);
+   assert(o["m"] === m, false);
+   assert(o["m"].isReadOnly(), false);
+   anydata p = o["m"];
+   assert(p is map<anydata>, true);
+   map<anydata> q = <map<anydata>>p;
+   q["c"] = "foo";
```
MaryamZi, line 848: Shouldn't we also assert the value of `q["c"]` and maybe the length of `q` and see?

MaryamZi (approved):

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
warunalakshitha: This will panic normally. BTW I'll add a simple check.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/stamp/RecordStampInbuiltFunctionTest.java
```diff
@@ -539,8 +539,8 @@ public void testStampComplexRecordToJSON() {
         Assert.assertEquals(((BMapType) employee0.getType()).getConstrainedType().getClass(), BAnydataType.class);
 
         Assert.assertEquals((employee0.get("marks").getType().getClass()), BArrayType.class);
-        Assert.assertEquals(((BArrayType) ((BArrayType) (employee0).get("marks").getType()).getElementType()).
-                getElementType().getTag(), TypeTags.INT_TAG);
+        Assert.assertEquals(((BArrayType) (employee0).get("marks").getType()).getElementType().getTag(),
```
MaryamZi, line 542: Not introduced by this PR, but I think the name of the test is misleading? Shall we fix it in a separate PR maybe?

MaryamZi (approved):

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/stamp/RecordStampInbuiltFunctionTest.java
warunalakshitha: I'll send a seperate PR

warunalakshitha: I think we still use "stamp" in our tests. ( even in test class names as well). That needs to change as well.

warunalakshitha: https://github.com/ballerina-platform/ballerina-lang/issues/30326

Issue #27471, opened 2020-12-12
Compiler Crashes When a Submodule Test has a Compilation Error


ThisaruGuruge:
**Description:**
I have a package, including a submodule. When there's a compilation error in the submodule test file, the Package build is crashing with an NPE.

Following is the log generated:

```
[2020-12-12 23:09:44,091] SEVERE {b7a.log.crash} - null 
java.lang.NullPointerException
	at org.wso2.ballerinalang.compiler.desugar.Desugar.desugarObjectTypeInit(Desugar.java:5145)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.visit(Desugar.java:5130)
	at org.wso2.ballerinalang.compiler.tree.expressions.BLangTypeInit.accept(BLangTypeInit.java:62)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.rewriteExpr(Desugar.java:6654)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.visit(Desugar.java:1228)
	at org.wso2.ballerinalang.compiler.tree.BLangSimpleVariable.accept(BLangSimpleVariable.java:53)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.rewrite(Desugar.java:6629)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.visit(Desugar.java:1353)
	at org.wso2.ballerinalang.compiler.tree.statements.BLangSimpleVariableDef.accept(BLangSimpleVariableDef.java:46)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.rewrite(Desugar.java:6629)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.rewrite(Desugar.java:6670)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.rewriteStmt(Desugar.java:6680)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.visit(Desugar.java:1136)
	at org.wso2.ballerinalang.compiler.tree.BLangBlockFunctionBody.accept(BLangBlockFunctionBody.java:58)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.rewrite(Desugar.java:6629)
	at org.wso2.ballerinalang.compiler.desugar.Desugar.visit(Desugar.java:1181)
	at org.wso2.ballerinalang.compiler.tree.BLangFunction.accept(BLangFunction.java:73)
[...]

PR #30234, opened 2021-04-28
Stop jbal codegen when dependencies have errors

rdhananjaya:
## Purpose
Not sure how to add an test case for this one.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/27471

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/projects/JBallerinaBackend.java
```diff
@@ -144,7 +144,10 @@ private void performCodeGen() {
 
         List<Diagnostic> diagnostics = new ArrayList<>();
         for (ModuleContext moduleContext : pkgResolution.topologicallySortedModuleList()) {
-            moduleContext.generatePlatformSpecificCode(compilerContext, this);
+            // We can't generate backend code when one of its dependencies have errors.
```
warunalakshitha, line 147: Will this interop validation errors? We already have a check before codegen after interop validation. https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/io/ballerina/projects/JBallerinaBackend.java#L275

rdhananjaya: If there are errors in dependencies they yes, this will not goto interpo validation

hasithaa: I think this is safe for now. We have this problem in General. Let's figure out a proper solution later. wdyt ?

hasithaa (dismissed):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/JBallerinaBackend.java
```diff
@@ -144,7 +144,10 @@ private void performCodeGen() {
...
+            if (diagnostics.isEmpty()) {
+                moduleContext.generatePlatformSpecificCode(compilerContext, this);
```
hasithaa, line 149: Shouldn't we check specifically, for errors here?. Having warnings shouldn't stop this.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/JBallerinaBackend.java
warunalakshitha: +1. BTW we can move the      if (diagnostics.isEmpty()) { check after line 151
 moduleContext.diagnostics().forEach(diagnostic ->
                    diagnostics.add(new PackageDiagnostic(diagnostic, moduleContext.moduleName())));

AFAIU first check will be always true since diagnostics isEmpty.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/JBallerinaBackend.java
rdhananjaya: Fixed

compiler/ballerina-lang/src/main/java/io/ballerina/projects/JBallerinaBackend.java
rdhananjaya: With fix for this comment https://github.com/ballerina-platform/ballerina-lang/pull/30234/files#r622735337 we can keep this as is right

hasithaa (approved):

PR #30251, opened 2021-04-29
Improve the formatting behavior of blocks

### LINKED ISSUE #30095: Support both expanded and inline formatting of blocks
**Description:**
Currently, the formatter makes a decision to expand blocks given below based on the number of fields (if the number of fields is greater than 2, then it would be expanded instead of formatting it as an inline block).

However, it would be more user-friendly if the formatter takes into account the existing formatting of the user (whether it is expanded or inline) when deciding whether it should be formatted as an inline-block or an expanded block.

The following block types should be supported with this improvement.
* `MappingConstructorExpressionNode`
* `ObjectTypeDescriptorNode`
* `RecordTypeDescriptorNode`

**Affected version:**
SL Alpha 4

**Describe your solution(s)**

IrushiL:
## Purpose
This PR improves the current behavior of formatting blocks.

**Previous Rule:** 
* expanding blocks based on the number of fields -> if the number of fields is greater than 2, then it would be expanded instead of formatting it as an inline-block

**Improved Rule:**
* expanding blocks based on the existing formatting of the user -> if a single newline is observed inside the block, the block will be expanded, else it would be formatted as an in-line block

The following block types are supported with this improvement.
* MappingConstructorExpressionNode
* ObjectTypeDescriptorNode
* RecordTypeDescriptorNode

Fixes #30095

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
+++ b/misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
@@ -237,7 +237,6 @@ import io.ballerina.compiler.syntax.tree.XMLStepExpressionNode;
 import io.ballerina.compiler.syntax.tree.XMLTextNode;
 import io.ballerina.compiler.syntax.tree.XmlTypeDescriptorNode;
 import io.ballerina.tools.text.LineRange;
-import io.ballerina.tools.text.TextRange;
 import org.slf4j.Logger;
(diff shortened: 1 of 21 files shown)


mohanvive (dismissed):
LGTM. Behaviour looks OK for me.

nadeeshaan (approved):
LGTM

PR #30256, opened 2021-04-29
Fix incorrect symbol returning issue when using `symbol()` for typesdescs

pubudu91:
## Purpose
This PR fixes the issue we had with `symbol()` always returning a type ref symbol when looking up a typedesc. 

- Case 1
```ballerina
<cursor>int foo = 10;
```
When calling `symbol()` for the above illustrated position, previously it returned a `TypeReferenceTypeSymbol` for `int` type. With this PR, it returns an instance of `IntTypeSymbol`.

- Case 2
```ballerina
type Foo int|float;

<cursor1>Foo <cursor2>foo = 10;
```
When calling `symbol()` for above illustrated positions, previously it returned the type symbol of the typedesc itself (in this case, a union). With this PR, both of the above cases will now return a type ref type symbol for `Foo`.

## Remarks
This modifies the behaviour of `symbol()` when it comes to type descs. Previously it used to return type ref for all typedescs. Now it returns the relevant typedesc symbol. 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
@@ -1163,6 +1163,10 @@ class SymbolFinder extends BaseVisitor {
         lookupNode(tableType.constraint);
         lookupNode(tableType.tableKeySpecifier);
         lookupNode(tableType.tableKeyTypeConstraint);
+
+        if (this.symbolAtCursor == null) {
+            this.symbolAtCursor = tableType.type.type.tsymbol;
+        }
     }
 
     @Override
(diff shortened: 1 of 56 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
```diff
@@ -1152,6 +1152,10 @@ public void visit(BLangConstrainedType constrainedType) {
     public void visit(BLangStreamType streamType) {
         lookupNode(streamType.constraint);
         lookupNode(streamType.error);
+
+        if (symbolAtCursor == null) {
+            this.symbolAtCursor = streamType.type.type.tsymbol;
+        }
```
mohanvive, line 1158: @pubudu91 If this change required for stream type, then similar change also requires to table. Isn't it?

pubudu91: Yeah, earlier there was an issue with supporting this. Hence why I originally didn't add it. Had another go at it and seems to be working now 🤔 So added with https://github.com/ballerina-platform/ballerina-lang/pull/30256/commits/7f7b3a6c482574008f690f52859c31d38e65ab15

mohanvive (approved):
LGTM

mohanvive (approved):

PR #30263, opened 2021-04-30
Add foreach completion item suggestion on iterable field access

### LINKED ISSUE #29983: Add foreach completion item suggestion for iterable filed accession
**Description:**

Currently `foreach` completion item is provided as snippets as depicted below.

The expected behavior is to provide the particular completion item on the accession of a iterable  field as depicted below. 

```ballerina
  public function main() {
      int[5] arr = [1,2,3,4,5];
  
      // shall provide the completion item on field accession ('arr.foreach')
      for int item in arr {
  
      }
}
```

**Steps to reproduce:**
Try the above code segment in VSCode.

malinthar:
## Purpose
Suggest foreach statement dynamically on iterable field access as follows.

Fixes #29983

## Approach
Add a Util class to provide for each completion item when the cursor is in `FieldAccess` context.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,189 @@
...
+     * @param ctx        completion context
+     * @param expr       expression node
+     * @param typeSymbol type symbol corresponding to the expression node
+     * @return a list of {@link LSCompletionItem}.
+     */
+    public static List<LSCompletionItem> getForeachCompletionItemForIterable(BallerinaCompletionContext ctx,
+                                                                             FieldAccessExpressionNode expr,
+                                                                             Optional<TypeSymbol> typeSymbol) {
+        List<LSCompletionItem> completionItems = new ArrayList<>();
+        if (isInBlockContext(expr) && expr.expression() instanceof SimpleNameReferenceNode) {
```
nadeeshaan, line 77: Let's avoid the `instanceof` check and instead use the `kind` check. Also, use the early return here.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,189 @@
...
+            completionItems.addAll(
+                    typeSymbol.stream().filter(tSymbol -> {
+                                if (tSymbol.typeKind() == TypeDescKind.TYPE_REFERENCE) {
```
nadeeshaan, line 80: Instead of this `if else`  clause to get the type descriptor conditionally, shall we use `CommonUtil.getRawType` API? We do this same check there.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * Given an iterable field access expression returns the foreach completion item for it.
+     *
+     * @param symbol {@link ArrayTypeSymbol} symbol
+     * @param ctx    completion context
+     * @param expr   expression node
+     * @return {@link LSCompletionItem}
+     */
+    public static LSCompletionItem getForEachCompletionItemForIterable(TypeSymbol symbol,
```
nadeeshaan, line 101: Do we have any public usages for this API? If not, shall we make this private?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,189 @@
...
+                                                                       BallerinaCompletionContext ctx,
+                                                                       FieldAccessExpressionNode expr) {
+        String symbolName = ((SimpleNameReferenceNode) expr.expression()).name().text();
+        String detail = "foreach snippet for iterable variable - " + symbolName;
+        String label = "foreach - " + symbolName;
+        String type = getTypeOfIteratorVariable(symbol);
+        StringBuilder snippet = new StringBuilder("foreach");
+        snippet.append(" ").append(type).append(" ")
+                .append(CommonUtil.getValidatedSymbolName(ctx, VAR_NAME)).append(" in ").append(symbolName)
+                .append(" ").append("{").append(CommonUtil.LINE_SEPARATOR).append("\t${1:}")
```
nadeeshaan, line 111: For the placeholder in the snippet, we don't need to have the `:` since we do not have a place holder name.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,189 @@
...
+     * Returns the type of the member type of an iterable type symbol.
+     *
+     * @param symbol iterable type symbol.
+     * @return signature of the member type.
+     */
+    public static String getTypeOfIteratorVariable(TypeSymbol symbol) {
+        String type;
+        switch (symbol.typeKind()) {
+            case STRING:
+                type = symbol.typeKind().getName();
```
nadeeshaan, line 137: What if we use `signature` API here also? Then we can combine with `xml` case.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,189 @@
...
+                type = ((TableTypeSymbol) symbol).rowTypeParameter().signature();
+                break;
+            case STREAM:
+                type = ((StreamTypeSymbol) symbol).typeParameter().signature();
+                break;
+            case XML:
+                type = symbol.signature();
+                break;
+            case TYPE_REFERENCE:
+                type = getTypeOfIteratorVariable(((TypeReferenceTypeSymbol) symbol).typeDescriptor());
```
nadeeshaan, line 169: You can avoid this by using the `CommonUtil.getRawType` API. Instead of doing the type check for the type symbol in this switch-case, let's do the check against the raw type.

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
malinthar: Done!

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,175 @@
...
+import org.eclipse.lsp4j.Range;
+import org.eclipse.lsp4j.TextEdit;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Dynamic foreach snippet construction for iterables.
```
IMS94, line 49: Let's add the `@since 2.0.0` line to javadoc as well

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,175 @@
...
+ */
+public class ForeachUtil {
```
IMS94, line 51: How about we rename this to `ForeachCompletionUtil` since this is a utility for completions?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,175 @@
...
+    private static final List<TypeDescKind> ITERABLES = Arrays.asList(
+            TypeDescKind.STRING, TypeDescKind.ARRAY, TypeDescKind.TUPLE,
+            TypeDescKind.MAP, TypeDescKind.RECORD, TypeDescKind.TABLE,
+            TypeDescKind.STREAM, TypeDescKind.XML);
+
+    private ForeachUtil() {
+    }
+
+    /**
+     * Given a expression node in a particular completion context,
```
IMS94, line 64:
```suggestion
     * Given an expression node in a particular completion context,
```

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,175 @@
...
+     * @return a list of {@link LSCompletionItem}.
+     */
+    public static List<LSCompletionItem> getForeachCompletionItemForIterable(BallerinaCompletionContext ctx,
+                                                                             FieldAccessExpressionNode expr,
+                                                                             TypeSymbol typeSymbol) {
+        List<LSCompletionItem> completionItems = new ArrayList<>();
+        if (isInBlockContext(expr) && expr.expression().kind() == SyntaxKind.SIMPLE_NAME_REFERENCE
+                && ITERABLES.contains(CommonUtil.getRawType(typeSymbol).typeKind())) {
+            completionItems.add(getForEachCompletionItemForIterable(typeSymbol, ctx, expr));
+            return completionItems;
```
IMS94, line 79: We don't need the return here because we anyway return the same list in the last line.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Given an iterable field access expression returns the foreach completion item for it.
+     *
+     * @param symbol {@link ArrayTypeSymbol} symbol
+     * @param ctx    completion context
+     * @param expr   expression node
+     * @return {@link LSCompletionItem}
+     */
+    private static LSCompletionItem getForEachCompletionItemForIterable(TypeSymbol symbol,
```
IMS94, line 92: Seems like this has the same signature as the above one but with parameters reordered. What if we change this method's name at least?

IMS94 (changes requested):
Logic seems good. I have added some comments related to coding style at several places. Please have a look.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachUtil.java
malinthar: Done!

IMS94 (approved):

nadeeshaan (approved):

PR #30309, opened 2021-05-03
Add type narrowing support for == and !=

### LINKED ISSUE #30165: == should narrow type
With slalpha4 the compiler rejects this program:

```
function foo(string? s) returns string {
  if s == () {
    return "";
  }
  else {
     return s;
  }
}
```

But it's correct because the spec says:

> If E has the form x == E1 or E1 == x where the static type of E1 is an expression whose static type is a singleton simple type T1, then
the narrowed type for x implied by truth of E is the intersection of Tx and T1
the narrowed type for x implied by falsity of E is Tx with T1 removed

So the type of s in the else clause is (string? - ()) which is string.

rdulmina:
## Purpose
$title

Fixes  #30165

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeNarrower.java
```diff
@@ -350,4 +366,49 @@ private SymbolEnv getTargetEnv(BLangNode targetNode, SymbolEnv env) {
...
+
+        BFiniteType finiteType = new BFiniteType(finiteTypeSymbol);
+        expr.type = symTable.getTypeFromTag(expr.type.tag);
+        finiteType.addValue(expr);
+        finiteTypeSymbol.type = finiteType;
+
+        return finiteType;
+    }
+
+    private void narrowTypeForEqualOrNotEqual(BLangBinaryExpr binaryExpr, BLangExpression lhsExpr, BLangExpression rhsExpr) {
```
sanjana, line 383: This line is longer than 120 characters.

rdulmina: updated thanx

rdulmina:
@nadeeshaan can u verify the changes in b8c498a

language-server/modules/langserver-core/src/test/resources/completion/statement_context/config/elseif_stmt_ctx_config1.json
```diff
@@ -605,19 +605,19 @@
...
       "sortText": "C",
-      "insertText": "testFunction()",
+      "insertText": "doSomeTask()",
       "insertTextFormat": "Snippet"
     },
     {
       "label": "testInt",
       "kind": "Variable",
-      "detail": "int",
+      "detail": "2",
```
nadeeshaan, line 614: In this scenario, for the the `testInt` variable, the type is shown as `2` instead of `int`, which was the earlier behaviour. Refer the following user experience.

Screenshot 2021-05-11 at 09 51 47

nadeeshaan (changes requested):

tests/jballerina-unit-test/src/test/resources/test-src/closures/closure.bal
```diff
@@ -666,20 +666,20 @@ function test30() returns int|string {
 function test31() returns int|string {
     int|string x = 5;
     var func1 = function () {
-        if (x is int && x > 3) {
-            x = x + 1;
-            if (x > 5) {
-               x = x + 10;
+        if (x is int && <int>x > 3) {
```
rdhananjaya, line 669: IMO we can remove these test cases as these test cases are testing the type being narrowed, but we don't narrow this any more. So I think we should remove these and create negative test cases

rdhananjaya:
We can't close https://github.com/ballerina-platform/ballerina-lang/issues/30445 with this PR right, shall we remove it from the `fixes` list

rdulmina:
> We can't close #30445 with this PR right, shall we remove it from the `fixes` list. Also shall we add some negative test cases for the resetting of type-narrowing for compound assignment.

We can close it. Issues with compound assignment are fixed with this

hasithaa:
Following is not working with your PR, 

```
int? v = ();

// Here the value of `n` annot be `()`.
int n = v == () ? 0 : v;

public function main() {
    // io:println(v);
}
```

rdulmina:
> Following is not working with this PR,
> 
> ```
> int? v = ();
> 
> // Here the value of `n` annot be `()`.
> int n = v == () ? 0 : v;
> 
> public function main() {
>     // io:println(v);
> }
> ```

This issue is with the type check expression as well. I think compiler is not considering type narrowing for ternary expression

hasithaa:
I had a chat with @rdulmina , My example is wrong. Spec says only about local variables and parameters.

hasithaa (approved):
Tested the PRs, looks good to me.

hasithaa:
@rdulmina We need to add negative test cases, could you please create an issue?

rdulmina:
> @rdulmina We need to add negative test cases, could you please create an issue?

Sure ayye

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -7934,6 +7926,10 @@ private BLangLiteral createByteLiteral(Location pos, Byte value) {
     }
 
     private BLangExpression createTypeCastExpr(BLangExpression expr, BType targetType) {
+        if (expr.type.tag == targetType.tag) {
```
rdhananjaya, line 7929: Shall we send a separate PR checking if this is a same type. There could be same tag for say user defined types but actual types are different.

PR #30340, opened 2021-05-04
Add back the key not found diagnostic code

lasinicl:
## Purpose
$title

Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
@@ -512,7 +512,8 @@ public enum DiagnosticErrorCode implements DiagnosticCode {
     EXPRESSION_IS_NOT_A_CONSTANT_EXPRESSION("BCE3504", "expression.is.not.a.constant.expression"),
     INVALID_CONST_EXPRESSION("BCE3505", "invalid.const.expression"),
     CONSTANT_EXPRESSION_NOT_SUPPORTED("BCE3506", "const.expression.not.supported"),
-    CONSTANT_DECLARATION_NOT_YET_SUPPORTED("BCE3507", "constant.declaration.not.yet.supported.for.type"),
+    KEY_NOT_FOUND("BCE3507", "key.not.found"),
+    CONSTANT_DECLARATION_NOT_YET_SUPPORTED("BCE3508", "constant.declaration.not.yet.supported.for.type"),
 
     // Anonymous functions related codes
     ARROW_EXPRESSION_MISMATCHED_PARAMETER_LENGTH("BCE3600", "arrow.expression.mismatched.parameter.length"),


compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
```diff
@@ -512,7 +512,8 @@
     EXPRESSION_IS_NOT_A_CONSTANT_EXPRESSION("BCE3504", "expression.is.not.a.constant.expression"),
     INVALID_CONST_EXPRESSION("BCE3505", "invalid.const.expression"),
     CONSTANT_EXPRESSION_NOT_SUPPORTED("BCE3506", "const.expression.not.supported"),
-    CONSTANT_DECLARATION_NOT_YET_SUPPORTED("BCE3507", "constant.declaration.not.yet.supported.for.type"),
+    KEY_NOT_FOUND("BCE3507", "key.not.found"),
```
MaryamZi, line 515: Where/how is this being used in stdlib? @lasinicl @niveathika

niveathika: Used in the file compiler plugin https://github.com/ballerina-platform/module-ballerina-file/blob/master/file-compiler-plugin/src/main/java/io/ballerina/stdlib/file/compiler/FileServiceValidator.java#L96

MaryamZi: I think we need to validate if this is the correct error code to use. That logic seems to be validating if the remote qualifier is present in a method.

Let's go with this change for now, but shall we create two issues for lang and the file module?

MaryamZi (approved):

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
MaryamZi: Also, I was under the impression that there wouldn't/shouldn't be usages of these codes outside the compiler. 

Looping in @pubudu91 and @hasithaa  too.

lasinicl: Created https://github.com/ballerina-platform/ballerina-lang/issues/30341

pubudu91: The expectation was that compiler plugin developers would come up with their own diagnostics and diagnostic codes. I can't really think of a scenario where a compiler plugin would need a diagnostic from the compiler. 🤔 I mean, if that's the case, then that check could've been done in the compiler itself right? 🤔

niveathika:
Closing this as we decided to use different error code in file for now

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
hasithaa: As Pubudu mentioned, the expectation was each compiler plugin must have its own diagnostics and diagnostic codes. Diagnostics should be searchable using diagnostic codes, so It is wrong to use compiler codes.

PR #30344, opened 2021-05-04
Set a default error message while throwing compiler exception

### LINKED ISSUE #30210: NPE when an old version is pulled from central as the import module
**Description:**
Getting an NPE when the import module is not available in the pack. In the sample, it's `ballerinax/java.jdbc`.
Better if we can have a proper error message.

```
Compiling source
        testBal.bal
Exception in thread "main" java.lang.NullPointerException
        at io.ballerina.cli.launcher.Main.main(Main.java:63)
```

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**
SL Alpha 5

**Steps to reproduce:**
```ballerina
import ballerina/lang.'transaction as transactions;
import ballerina/io;
import ballerinax/java.jdbc;

    // The JDBC Client for the H2 database.
    jdbc:Client dbClient =
                check new (url = "jdbc:h2:file:./local-transactions/accountdb",
                           user = "test", password = "test");

    // Create the database table and populate some records.
    _ = check dbClient->execute("CREATE TABLE IF NOT EXISTS ACCOUNT " +
                                "(ID INTEGER, BALANCE DECIMAL, " +
                                "PRIMARY KEY(id))");
    _ = check dbClient->execute("INSERT INTO ACCOUNT VALUES (1, 2500.0)");
    _ = check dbClient->execute("INSERT INTO ACCOUNT VALUES (2, 1000.0)");

    // This is a `transaction` statement block. It is a must to have either
    // a commit action or a rollback statement in it. The transaction scope ends after
    // the commit action or rollback statement.
    transaction {
        // Execute database operations within the transaction
        var creditResult = dbClient->execute(
                                       "UPDATE ACCOUNT " +
                                       "SET BALANCE=BALANCE+500.0 WHERE ID=1");
        var debitResult = dbClient->execute(
                                       "UPDATE ACCOUNT " +
                                       "SET BALANCE=BALANCE-500.0 WHERE ID=2");

        // Returns information about the current transactions.
        transactions:Info transInfo = transactions:info();
        io:println("Transaction Info: ", transInfo);

        // The `commit` action performs the commit operation of the current transaction.
        // The result of the commit action will be either `error` or `()`.
        var commitResult = commit;
        if commitResult is () {
            // Operations to be executed after the transaction are committed successfully.
            io:println("Transaction committed");
            io:println("Account Credit: ", creditResult);
            io:println("Account Debit: ", debitResult);
        } else {
            // Operations to be executed if the transaction commit failed.
            io:println("Transaction failed");
        }
    }

    _ = check dbClient->execute("DROP TABLE ACCOUNT");
    
    // Close the JDBC client.
    check dbClient.close();
}
```

**Related Issues:**

pcnfernando:
Fixes #30210

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
@@ -195,7 +195,8 @@ public class BIRPackageSymbolEnter {
             this.env = prevEnv;
             return pkgSymbol;
         } catch (Throwable e) {
-            throw new BLangCompilerException(e.getMessage(), e);
+            throw new BLangCompilerException("Error while loading the bir. " +
+                    (e.getMessage() != null ? e.getMessage() : ""), e);
         }
     }
 


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -195,7 +195,8 @@ private BPackageSymbol definePackage(PackageID packageId,
             this.env = prevEnv;
             return pkgSymbol;
         } catch (Throwable e) {
-            throw new BLangCompilerException(e.getMessage(), e);
+            throw new BLangCompilerException("Error while loading the bir. " +
+                    (e.getMessage() != null ? e.getMessage() : ""), e);
```
pubudu91, line 199: Since this is a `BLangCompilerException`, this will get printed with the `error:` prefix. So starting the message with `Error` may look like redundant. Also, shall we separate the messages for the 2 conditions (where error message is `null` and not `null`)? Something like,
- if the error message is there, `"failed to load the module '" + packageId.toString() + "' from its BIR due to: " + e.getMessage()`
- if the error message is `null`, `"failed to load the module '" + packageId.toString() + "' from its BIR"`

Also wondering if that 2nd case should include an additional hint like `Please ensure that the module version used is compatible with the language version in use` 🤔 

@MaryamZi thoughts?

MaryamZi: +1 to start with "failed to".

Re: the actual error though, is this purely due to not being able to load the module from the BIR? Or can it also happen maybe due to the BIR not being available for example?

 
> Also wondering if that 2nd case should include an additional hint like `Please ensure that the module version used is compatible with the language version in use` 

Same concern as above, is this the only reason that can cause the message to be null?

pubudu91: If the BIR is not available at all, it'll fail when trying to resolve the import right? Saying that it could not resolve the module? @azinneera please confirm.

The thing with the second one is a bit tricky. Version incompatibility may not be the only cause for NPEs. It could be a bug in the compiler too. But the common scenario I've seen for NPEs is the version incompatibility. Hence the suggestion to add that hint. The desirable situation would be to be able to distinguish between the error cases: failure due to version incompatibility (give an error message to the user), compiler bugs (crash with bad-sad). But can't really do that with the current design.

azinneera: We load the BIR only if it exists in the file system. Else we use the bala files to compile. https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/io/ballerina/projects/ModuleContext.java#L246. So this will only when the BIR is available, AFAIU.

azinneera: We load the BIR only if it exists in the file system. Else, we use the bala files to compile. https://github.com/ballerina-platform/ballerina-lang/blob/8be01a443677993bb55965af19b8d1843f474f6b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/ModuleContext.java#L246. 
https://github.com/ballerina-platform/ballerina-lang/blob/8be01a443677993bb55965af19b8d1843f474f6b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/repos/FileSystemCache.java#L63-L75
So this will only happen when the BIR is available, AFAIU.

pcnfernando: Thanks for the inputs. Shall we stick to the below suggestions since the versioning issue cannot be distinguished? Even though the hint will address the common case, it might be misleading for the rest. @pubudu91 @MaryamZi 
> if the error message is there, "failed to load the module '" + packageId.toString() + "' from its BIR due to: " + e.getMessage()
> if the error message is null, "failed to load the module '" + packageId.toString() + "' from its BIR"

pubudu91: +1

pubudu91 (approved):

PR #30354, opened 2021-05-05
Fix incorrect results with maps and arrays of int and string subtypes

### LINKED ISSUE #30264: Incorrect result for `type test expression` with maps and arrays of `int` and `string` subtypes
**Description:**

**Steps to reproduce:**
```ballerina
public function main() {
    map<string:Char> x1 = {};
    any a1 = x1;
    boolean b1 = a1 is anydata; 
    io:println(b1); //false, expected true

    string:Char[] x2 = [];
    any a2 = x2;
    boolean b2 = a2 is anydata; 
    io:println(b2); //false, expected true

    map<int:Signed32> x3 = {};
    any a3 = x3;
    boolean b3 = a3 is anydata; 
    io:println(b3); //false, expected true

    int:Signed32[] x4 = [];
    any a4 = x4;
    boolean b4 = a4 is anydata; 
    io:println(b4); //false, expected true
}
```
The following work as expected
```ballerina
public function main() {
    map<string> x1 = {};
    any a1 = x1;
    boolean b1 = a1 is anydata; 
    io:println(b1); //true

    string[] x2 = [];
    any a2 = x2;
    boolean b2 = a2 is anydata; 
    io:println(b2); //true

    map<int> x3 = {};
    any a3 = x3;
    boolean b3 = a3 is anydata; 
    io:println(b3); //true

    int[] x4 = [];
    any a4 = x4;
    boolean b4 = a4 is anydata; 
    io:println(b4); //true
}
```

**Affected Versions:**
slalpha5

gabilang:
fix the incorrect results occurred for type test expression with maps and arrays of int and string sub types.
include the built-in subtypes of `xml` to the `io.ballerina.runtime.TypeChecker#hasFillerValue` check. 

Fixes 
#30264
#30524
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/TypeTags.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/TypeTags.java
@@ -24,35 +24,37 @@ package io.ballerina.runtime.api;
  */
 public class TypeTags {
 
-    // Value types
     public static final int INT_TAG = 1;
     public static final int BYTE_TAG = INT_TAG + 1;
     public static final int FLOAT_TAG = BYTE_TAG + 1;
     public static final int DECIMAL_TAG = FLOAT_TAG + 1;
     public static final int STRING_TAG = DECIMAL_TAG + 1;
     public static final int BOOLEAN_TAG = STRING_TAG + 1;
+    public static final int JSON_TAG = BOOLEAN_TAG + 1;
+    public static final int XML_TAG = JSON_TAG + 1;
+    public static final int TABLE_TAG = XML_TAG + 1;
+    public static final int NULL_TAG = TABLE_TAG + 1;
 
-    // Subtypes of value types (int and string)
-    public static final int SIGNED32_INT_TAG = BOOLEAN_TAG + 1;
+    // subtypes
+    public static final int SIGNED32_INT_TAG = NULL_TAG + 1;
     public static final int SIGNED16_INT_TAG = SIGNED32_INT_TAG + 1;
     public static final int SIGNED8_INT_TAG = SIGNED16_INT_TAG + 1;
     public static final int UNSIGNED32_INT_TAG = SIGNED8_INT_TAG + 1;
     public static final int UNSIGNED16_INT_TAG = UNSIGNED32_INT_TAG + 1;
     public static final int UNSIGNED8_INT_TAG = UNSIGNED16_INT_TAG + 1;
     public static final int CHAR_STRING_TAG = UNSIGNED8_INT_TAG + 1;
+    public static final int XML_ELEMENT_TAG = CHAR_STRING_TAG + 1;
+    public static final int XML_PI_TAG = XML_ELEMENT_TAG + 1;
+    public static final int XML_COMMENT_TAG = XML_PI_TAG + 1;
+    public static final int XML_TEXT_TAG = XML_COMMENT_TAG + 1;
+    public static final int NEVER_TAG = XML_TEXT_TAG + 1;
 
-    // branded types
-    public static final int JSON_TAG = CHAR_STRING_TAG + 1;
-    public static final int XML_TAG = JSON_TAG + 1;
-    public static final int TABLE_TAG = XML_TAG + 1;
-    public static final int NULL_TAG = TABLE_TAG + 1;
-    public static final int ANYDATA_TAG = NULL_TAG + 1;
+    public static final int ANYDATA_TAG = NEVER_TAG + 1;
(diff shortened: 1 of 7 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/types/anydata/anydata_test.bal
```diff
@@ -900,6 +900,29 @@ function testRuntimeIsAnydata() {
...
+    assertTrue(a3 is anydata);
+
+    int:Signed32[] x4 = [];
+    any a4 = x4;
+    assertTrue(a4 is anydata);
+
+    map<never> x5 = {};
+    any a5 = x5;
+    assertTrue(a5 is anydata);
+}
```
riyafa, line 924: Shall we break the tests to multiple test cases? In this case we can be more specific on what we are testing. We can use a data provider on the java side to call the functions.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/TypeTags.java
```diff
@@ -33,7 +34,22 @@
...
-    public static final int ANYDATA_TAG = NULL_TAG + 1;
+
+    // subtypes
+    public static final int SIGNED32_INT_TAG = NULL_TAG + 1;
+    public static final int SIGNED16_INT_TAG = SIGNED32_INT_TAG + 1;
+    public static final int SIGNED8_INT_TAG = SIGNED16_INT_TAG + 1;
+    public static final int UNSIGNED32_INT_TAG = SIGNED8_INT_TAG + 1;
+    public static final int UNSIGNED16_INT_TAG = UNSIGNED32_INT_TAG + 1;
+    public static final int UNSIGNED8_INT_TAG = UNSIGNED16_INT_TAG + 1;
+    public static final int CHAR_STRING_TAG = UNSIGNED8_INT_TAG + 1;
```
manuranga, line 45: lets add a comment break here
```suggestion
    // subtypes with no filler value
    public static final int CHAR_STRING_TAG = UNSIGNED8_INT_TAG + 1;
```

manuranga: looks like we can't say this, only no-filler-value is char

gabilang: Since the built-in subtypes of `xml` have filler values, we can't say that there. Anyway current implementation does not consider `built-in Xml subtypes` for filler values. It looks like a bug. 
e.g.
```
import ballerina/io;

public function main() {

    'xml:Text x1 = 'xml:createText("text 1");
    'xml:Text x2 = 'xml:createText("text 2");
    'xml:Text x3 = 'xml:createText("text 3");

    'xml:Text[] x = [x1, x2];
    insertListValue(x, 4, x3);
    io:println(x);
} 

function insertListValue('xml:Text[] list, int pos, 'xml:Text value) {
    list[pos] = value;
}
```
Above code panics abruptly with the following error message
`error: {ballerina/lang.array}IllegalListInsertion {"message":"array of length 2 cannot be expanded into array of length 5 without filler values"}`
But should allow the insertion with filler values according to the [spec](https://ballerina.io/spec/lang/master/#FillMember).

riyafa: Shall we create an issue for this if you haven't already and resolve it from this PR since I see that you have tried to fix it here

warunalakshitha (approved):
Review changes are fixed. LGTM

warunalakshitha (changes requested):
This  PR moves CHAR_STRING_TAG above  TypeTags.JSON_TAG . 
we do <  TypeTags.JSON_TAG in many places need to check all those places are now correct. Also need to verify the change with spec too.

gabilang:
> This PR moves CHAR_STRING_TAG above TypeTags.JSON_TAG .
> we do < TypeTags.JSON_TAG in many places need to check all those places are now correct. Also need to verify the change with spec too.

`< TypeTags.JSON_TAG` check has been used in mutability checks and string representation cases (in five places at runtime) and those places have been checked and verified. So, moving `TypeTags.CHAR_STRING_TAG` above `TypeTags.JSON_TAG` will not bring any issues.

warunalakshitha (approved):
LGTM

Issue #23263, opened 2020-05-14
Ballerina --debug command is not working when executing the module.jar


praveennadarajah:
**Description:**
Ballerina --debug command is not working when executing the module.jar.

Execute the following command:
`ballerina run --debug 5005 target/bin/module.jar`

Following message is displayed:
```
ballerina: invalid Ballerina source path. It should either be a name of a module in a Ballerina project, a file with a '.bal' extension, or an executable '.jar' file.

USAGE:
    ballerina run {<bal-file> | <module-name> | <executable-jar>}

For more information try --help
```

PR #30363, opened 2021-05-05
[Debugger] Add jar/executable debugging capability

praveennadarajah:
## Purpose
This PR will add jar/executable debugging cabilities.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/23263


distribution/zip/jballerina/bin/bal
```diff
@@ -262,6 +262,27 @@ elif [[ $1 == "run" ]] && [[ $2 == *.jar ]]; then # if 1st arg is "run" and 2nd
         -Dballerina.target=jvm \
         -Djava.command=$JAVACMD \
         -jar "${@:2}" # ignores "run"
+
+elif [[ $1 == "run" ]] && [[ $2 == "--debug" ]]; then # if 1st arg is "run" and 2nd arg is "--debug".
```
NipunaRanasinghe, line 266: This logic will break the `bal run` command for ballerina package/single file, as the conventional CLI `run` commands (i.e. `bal run --debug 5005 <PACKAGE_NAME>` will be matched to this condition(hence the Ballerina CLI launcher will not be executed). Therefore move these debug agent args to the above `elif` condition.

NipunaRanasinghe (approved):

NipunaRanasinghe (commented):
We also need to modify the windows batch file with the same set of changes.

distribution/zip/jballerina/bin/bal
praveennadarajah: IMO we can not move the debug agent args to the above `elif` condition because the above `elif` condition runs the jar in normal mode. (i.e. `bal run <EXECUTABLE_JAR>`). Therefore adding an additional condition (`[[ $4 == *.jar ]`) here would work fine.
i.e. `elif [[ $1 == "run" ]] && [[ $2 == "--debug" ]] && [[ $4 == *.jar ]]; then`

distribution/zip/jballerina/bin/bal
```diff
@@ -262,6 +262,28 @@ elif [[ $1 == "run" ]] && [[ $2 == *.jar ]]; then # if 1st arg is "run" and 2nd
...
+        -Dballerina.version=2.0.0-beta.1-SNAPSHOT \
+        -Djava.util.logging.config.class="org.ballerinalang.logging.util.LogConfigReader" \
+        -Djava.util.logging.manager="org.ballerinalang.logging.BLogManager" \
+        -Denable.nonblocking=false \
+        -Djava.security.egd=file:/dev/./urandom \
+        -Dfile.encoding=UTF8 \
+        -Dballerina.target=jvm \
+        -Djava.command=$JAVACMD \
+        -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=$3 \
+        -jar "${@:4}" # ignores "run"
```
NipunaRanasinghe, line 286: Let's move this logic to the above 'else if' condition, to reduce code duplications

distribution/zip/jballerina/bin/bal
```diff
@@ -262,6 +262,28 @@ elif [[ $1 == "run" ]] && [[ $2 == *.jar ]]; then # if 1st arg is "run" and 2nd
         -Dballerina.target=jvm \
         -Djava.command=$JAVACMD \
         -jar "${@:2}" # ignores "run"
+
+# running executable(jar) in debug mode
+# if 1st arg is "run", 2nd arg is "--debug", 3rd arg is <debugPort> and 4th arg is executable(jar).
+elif [[ $1 == "run" ]] && [[ $2 == "--debug" ]] && [[ $4 == *.jar ]]; then
```
NipunaRanasinghe, line 268: Let's add the required validations for the debug port, before proceeding with the final command.

praveennadarajah: Debug port validation is handled by `agentlib`. When we give an invalid port the following error is thrown:
```
ERROR: transport error 103: invalid port number specified
ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [:735]
```

The above behaviour is observed for `bal run --debug <InvalidPort>` and `bal run --debug <InvalidPort> <PATH_TO_JAR>`. Therefore, IMO we don't have to handle this explicitly. WDYT?

distribution/zip/jballerina/bin/bal
praveennadarajah: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/30363/commits/6a6cc870877c09e0f9dad8a6149cf726df15019c

distribution/zip/jballerina/bin/bal
NipunaRanasinghe: Yes +1

NipunaRanasinghe (commented):
@praveennadarajah Let's add the required integration test cases as we discussed.

praveennadarajah:
> We also need to modify the windows batch file with the same set of changes.

Changes added to Windows batch file.

NipunaRanasinghe (approved):

PR #30374, opened 2021-05-05
Fix non-isolated method scheduling to be sequential

### LINKED ISSUE #29322: Support runtime api for non isolated function scheduling
Provide an API such as `invokeMethodAsync` but to schedule functions with an exclusive way. That is, functions from the same object never run in parallel.

manuranga:
Fixes #29322

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
@@ -19,15 +19,22 @@ package io.ballerina.runtime.api;
 
 import io.ballerina.runtime.api.async.Callback;
 import io.ballerina.runtime.api.async.StrandMetadata;
+import io.ballerina.runtime.api.flags.SymbolFlags;
+import io.ballerina.runtime.api.types.MethodType;
+import io.ballerina.runtime.api.types.ObjectType;
+import io.ballerina.runtime.api.types.ResourceMethodType;
 import io.ballerina.runtime.api.types.Type;
 import io.ballerina.runtime.api.values.BFuture;
 import io.ballerina.runtime.api.values.BObject;
 import io.ballerina.runtime.internal.scheduling.Scheduler;
 import io.ballerina.runtime.internal.scheduling.Strand;
+import io.ballerina.runtime.internal.types.BServiceType;
 import io.ballerina.runtime.internal.values.FutureValue;
 
 import java.util.Map;
 import java.util.function.Function;
+
+import static io.ballerina.runtime.api.TypeTags.SERVICE_TAG;
 
 /**
  * External API to be used by the interop users to control Ballerina runtime behavior.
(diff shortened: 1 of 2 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
```diff
@@ -54,23 +61,46 @@ public static Runtime getCurrentRuntime() {
...
+        if (isIsolate) {
+            return scheduler.schedule(new Object[1], func, null, callback, strandName, metadata).result;
+        } else {
+            return scheduler.scheduleToObjectGroup(object, new Object[1], func, null, callback, strandName, metadata)
+                    .result;
+        }
+    }
+
+    private boolean isIsolated(ObjectType type, String methodName) {
+        for (MethodType method : type.getMethods()) {
```
warunalakshitha, line 89: Shall we keep the set of isolated object method/resource method names as a set inside ObjectType. We can get that list during compile time and set it during codegen. https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmTypeGen.java#L1154

sameerajayasoma: The complete service object has to be isolated. This should come from the compiler frontend. 

Something similar to `ObjectType.isolated()` or a flag in the `ObjectType`. Checking all resource methods is not enough

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
```diff
@@ -54,23 +61,46 @@ public static Runtime getCurrentRuntime() {
...
+            }
+        }
+        if (type.getTag() == SERVICE_TAG) {
+            for (ResourceMethodType method : ((BServiceType) type).getResourceMethods()) {
+                if (method.getName().equals(methodName)) {
+                    return SymbolFlags.isFlagOn(method.getFlags(), SymbolFlags.ISOLATED);
+                }
+            }
+        }
+        assert false : "object type does not contain method : " + methodName;
```
warunalakshitha, line 101: I think we do not need this check. We already codegen no such error method if we call non defined object method.

manuranga: it's `assert` so it will get removed by jvm. just here to catch code issues.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
manuranga: Modified to check both object and resource

warunalakshitha (approved):

Issue #28507, opened 2021-02-08
Incorrect error message for a service declaration which doesn't provide definitions for methods


pubudu91:
Consider the following:
```ballerina
import ballerina/io;

listener Listener lsn = new();

type Foo object {
    function foo() returns string;
};

service Foo / on lsn {

    public string greeting = "Hello World!";

    resource function get greet() returns json => { output: self.greeting };

    function createError() returns @tainted error? => ();
}

public class Listener {

    public function 'start() returns error? {
        io:println("Starting listener");
    }

    public function gracefulStop() returns error? {
        io:println("Gracefully stopping listener");
    }

    public function immediateStop() returns error? {
        io:println("Immediately stopping listener");
    }

    public function detach(service object {} s) returns error? {
    }

    public function attach(service object {} s, string[]? name = ()) returns error? {
    }
}
```
When compiling the above, the following error is given:
```
ERROR [test.bal:(9:1,16:2)] no implementation found for the method 'foo' of class 'object { public string greeting; resource function get greet () returns (json); function createError () returns (error?); function foo () returns (string); }'
```
It looks like this goes in the exact same path as for class defs, hence resulting in the above incorrect error. Either we should handle service decls differently or need to rephrase this error message to make it more general, so it can be used in service contexts as well.

rdhananjaya:
Yes, this goes though the same path. +1 for emitting a specific error message for services.

rdhananjaya:
Similar found with object constructors:
```

public function main() {

    O o = object O {
        string h = "hello";

    };
}

type O object {
    function foo();
};
```

output:
```
no implementation found for the method 'foo' of class 'object { string h; function foo () returns (); }'
error: compilation contains errors
```

PR #30396, opened 2021-05-06
Improve unimplemented method diagnostic error messages given for service-decl and object constructor 

rdhananjaya:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/28507

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/resources/compiler.properties
+++ b/compiler/ballerina-lang/src/main/resources/compiler.properties
@@ -921,8 +921,8 @@ error.invalid.function.pointer.invocation=\
 error.invalid.default.param.value=\
   invalid value for parameter ''{0}'': only simple literals allowed
 
-error.abstract.object.constructor=\
-  abstract object ''{0}'' cannot have a constructor method
+error.illegal.init.method.in.object.type.descriptor=\
+  object type descriptor ''{0}'' cannot have an init method
 
 error.cannot.initialize.abstract.object=\
   cannot initialize abstract object ''{0}''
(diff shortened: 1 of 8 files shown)


hasithaa (approved):

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -933,6 +933,12 @@ error.invalid.interface.of.non.abstract.object=\
 error.unimplemented.referenced.method.in.class=\
   no implementation found for the method ''{0}'' of class ''{1}''
 
+error.unimplemented.referenced.method.in.service.declaration=\
+  no implementation found for the method ''{0}'' of service declaration ''{1}''
```
pubudu91, line 937: Isn't this a bit misleading? The error is that the service decl hasn't implemented the method in the type descriptor it specifies. Judging by the error message in the test case for this, the type descriptor is passed to `{{1}}`. How about something like 
```
implementation not found for the method 'exec' of type 'ServType' referred in the service declaration
```
or something similar. Something that makes it clear that the issue is that the service decl has referred a type, but it hasn't provided an implementation for a method described in the type.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -933,6 +933,12 @@ error.invalid.interface.of.non.abstract.object=\
...
+
+error.unimplemented.referenced.method.in.object.constructor=\
+  no implementation found for the method ''{0}'' of object constructor ''{1}''
```
pubudu91, line 940: Similar comment as above for this. Gives the feeling that the method is in the object constructor, not that the object constructor is expected to be providing an implementation for the method described in the contextually expected type.

pubudu91: Something like, 
```
object constructor expected to provide an implementation for the method 'onMessage' of type 'SerT'
```

compiler/ballerina-lang/src/main/resources/compiler.properties
pubudu91: Or
```
service declaration expected to provide an implementation for the method 'exec' of type 'ServType'
```

rdhananjaya: No, the `type` of the type that is being implemented is passed as the second arg to error message.

PR #30414, opened 2021-05-07
Add typedesc value to create filler values of anonymous record type 

### LINKED ISSUE #30147: Default values don't work as expected in filler values for nested anonymous records
**Description:**
$title.

**Steps to reproduce:**
```ballerina
import ballerina/io;

function foo() returns record {| int i; |}[] {
    record {| int i = 101; |}[] x = [];
    return x;
}

public function main() {
    var m = foo();
    m[1] = {i: 202};
    io:println(m); // [{"i":0},{"i":202}], note 0 instead of 101
}
```

Works when there is a separate type definition though.
```ballerina
import ballerina/io;

type Foo record {| int i = 101; |};

function foo() returns record {| int i; |}[] {
    Foo[] x = [];
    return x;
}

public function main() {
    var m = foo();
    m[1] = {i: 202};
    io:println(m); // [{"i":101},{"i":202}]
}
```

Also, the following works as expected.
```ballerina
import ballerina/io;

function foo() returns record {| int i; |} {
    record {| int i = 101; |} x = {};
    return x;
}

public function main() {
    map<anydata> m = {};
    io:println(m.cloneWithType(typeof foo())); // {"i":101}
}
```

**Affected Versions:**
slalpha4

HindujaB:
## Purpose
$subject

Fixes #30147 

## Approach
We earlier used record value creator to create filler values for arrays of records. The default values provided for anonymous records are ignored as the value creator identifies the record type without its default values. Therfore, we modified it to use the  `typedesc` value of the record type to instantiate filler values.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
@@ -20,6 +20,7 @@ package io.ballerina.runtime.internal.types;
 
 import io.ballerina.runtime.api.Module;
 import io.ballerina.runtime.api.TypeTags;
+import io.ballerina.runtime.api.creators.ValueCreator;
 import io.ballerina.runtime.api.flags.SymbolFlags;
 import io.ballerina.runtime.api.flags.TypeFlags;
 import io.ballerina.runtime.api.types.Field;
@@ -29,11 +30,9 @@ import io.ballerina.runtime.api.types.Type;
 import io.ballerina.runtime.api.utils.IdentifierUtils;
 import io.ballerina.runtime.api.utils.StringUtils;
 import io.ballerina.runtime.api.values.BString;
-import io.ballerina.runtime.internal.scheduling.Scheduler;
 import io.ballerina.runtime.internal.values.MapValue;
 import io.ballerina.runtime.internal.values.MapValueImpl;
 import io.ballerina.runtime.internal.values.ReadOnlyUtils;
-import io.ballerina.runtime.internal.values.TypedescValue;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -50,7 +49,6 @@ public class BRecordType extends BStructureType implements RecordType {
     public int typeFlags;
     private final boolean readonly;
     private IntersectionType immutableType;
-    public TypedescValue typedescValue;
 
     /**
      * Create a {@code BRecordType} which represents the user defined record type.
@@ -111,7 +109,7 @@ public class BRecordType extends BStructureType implements RecordType {
 
     @Override
     public <V extends Object> V getZeroValue() {
-        return (V) typedescValue.instantiate(Scheduler.getStrand());
+        return (V) ValueCreator.createRecordValue(this.pkg, this.typeName);
     }
 
     @SuppressWarnings("unchecked")
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
@@ -20,7 +20,6 @@ package io.ballerina.runtime.internal.types;
 
 import io.ballerina.runtime.api.Module;
 import io.ballerina.runtime.api.TypeTags;
-import io.ballerina.runtime.api.creators.ValueCreator;
 import io.ballerina.runtime.api.flags.SymbolFlags;
 import io.ballerina.runtime.api.flags.TypeFlags;
 import io.ballerina.runtime.api.types.Field;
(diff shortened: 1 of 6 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array_fill_runtime_test.bal
```diff
@@ -144,6 +144,71 @@ function tesTwoDimensionalArrayWithConstantSizeReferenceFill() {
...
+
+    testAnonRecordAsFunctionReturnType();
+}
+
+function testAnonRecordAsFunctionReturnType() {
+    var resultArr = getAnonFromAnon();
+    resultArr[1] = {name: "Jack"};
+    assertEqualPanic(2, resultArr.length());
+    assertEqualPanic("Jane", resultArr[0].name);
+    assertEqualPanic("Jack", resultArr[1].name);
```
riyafa, line 163: We can move the common code to a separate function and invoke the function with different parameters

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
```diff
@@ -49,6 +50,7 @@
     public int typeFlags;
     private final boolean readonly;
     private IntersectionType immutableType;
+    public TypedescValue typedescValue;
```
manuranga, line 53: So far in our impl, `TypedescValue` knows about `BType` but `BType` doesn't have any reference to Typedesc. Since TypedescValue already has a reference to BType it can create a circular reference, technically it's ok, but it would be nice if we can avoid it.

Better if we can use TypedescValue directly instead of BType when creating filler values.

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array_fill_runtime_test.bal
```diff
@@ -144,6 +144,56 @@ function tesTwoDimensionalArrayWithConstantSizeReferenceFill() {
...
+
+function getAnonFromAnon() returns record {| string name; |}[] {
+    record {|
+        string name = "Jane";
+    |}[] arr = [];
+    return arr;
+}
+
+function getAnonFromPerson() returns record {| string name; |}[] {
+    Person[] arr = [];
```
warunalakshitha, line 181: Shall we add a test with Person & readonly array too

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
```diff
@@ -109,7 +111,7 @@ public BRecordType(String typeName, Module pkg, long flags, Map<String, Field> f
 
     @Override
     public <V extends Object> V getZeroValue() {
-        return (V) ValueCreator.createRecordValue(this.pkg, this.typeName);
+        return (V) typedescValue.instantiate(Scheduler.getStrand());
```
warunalakshitha, line 114: Can we get NPE if recordtype is created through api?

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array_fill_runtime_test.bal
HindujaB: As per spec, we don't have a filler value for intersection types. Therefore, `Person & readonly` will give an `IllegalListInsertion` error.
https://ballerina-platform.slack.com/archives/CAXS256RX/p1621243803134900

warunalakshitha (dismissed):
LGTM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeHashVisitor.java
```diff
@@ -475,7 +475,8 @@ public Integer visit(BRecordType type) {
         }
         List<Integer> fieldsHashes = getFieldsHashes(type.fields);
         List<Integer> typeInclHashes = getTypesHashes(type.typeInclusions);
-        Integer hash = hash(baseHash(type), type.sealed, fieldsHashes, typeInclHashes, visit(type.restFieldType));
+        Integer hash = hash(baseHash(type), type.sealed, fieldsHashes, typeInclHashes, visit(type.restFieldType),
```
grainier, line 478: This line is changed in `stage-slbeta1` branch. Therefore, this can conflict. Better if we can wait for `stage-slbeta1` to sync with master and update the PR.
https://github.com/ballerina-platform/ballerina-lang/blob/stage-slbeta1/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeHashVisitor.java#L479

tests/jballerina-unit-test/src/test/resources/test-src/statements/arrays/array_fill_runtime_test.bal
MaryamZi: The spec has been updated according to https://github.com/ballerina-platform/ballerina-spec/issues/850. Have we fixed this? Shall we create an issue if not?

HindujaB: created https://github.com/ballerina-platform/ballerina-lang/issues/30977

warunalakshitha (approved):

Issue #30411, opened 2021-05-07
Unnecessary warnings are printed if given configurable type is not supported via cli args and value is not provided via clia arg


warunalakshitha:
**Description:**
Steps to reproduce

Define array type configurable variable
```
configurable int[] & readonly array = ?;
```
Provide the value using toml.
```
array=[1,2,3]
```
Result

The program runs fine but there is an unnecessary warning.

```
warning: value for configurable variable 'array' with type 'int[] & readonly' is not supported as a command line argument
	at waruna/configProject:0.1.0(main.bal:9)
```

warunalakshitha:
If the user has not provided the value for a given variable that is not supported with cli, then only we should print the warning.

warunalakshitha:
Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/30415

PR #30415, opened 2021-05-07
Fix unnecessary warnings are printed if given configurable type is not supported via cli args 

warunalakshitha:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30411

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/cli/CliProvider.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/cli/CliProvider.java
@@ -172,18 +172,30 @@ public class CliProvider implements ConfigProvider {
 
     @Override
     public Optional<BArray> getAsArrayAndMark(Module module, VariableKey key) {
+        CliArg cliArg = getCliArg(module, key);
+        if (cliArg.value == null) {
+            return Optional.empty();
+        }
         Type effectiveType = ((IntersectionType) key.type).getEffectiveType();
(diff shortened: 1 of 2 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/cli/CliProvider.java
```diff
@@ -172,18 +172,30 @@ public boolean hasConfigs() {
 
     @Override
     public Optional<BArray> getAsArrayAndMark(Module module, VariableKey key) {
+        CliArg cliArg = getCliArg(module, key);
```
HindujaB, line 175: As the same lines are executed for methods `getAsArrayAndMark`, `getAsRecordAndMark` & `getAsTableAndMark` , Shall we move it to a separate method and call?

HindujaB (approved):

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/configurable/providers/cli/CliProvider.java
warunalakshitha: In other get methods also same. Only in this case we throw a not support error until supported and other places we throw a different error. Shall we keep this since its more clear for me?

Issue #29748, opened 2021-03-31
Transaction handler function name suggested along with ()


pcnfernando:
**Description:**

![trxHandler](https://user-images.githubusercontent.com/15815199/113147065-068de900-924e-11eb-83b7-c3bc943097e1.gif)

**Steps to reproduce:**
```ballerina
import ballerina/io;
import ballerina/lang.'transaction as trx;

public function main() {
    var onRollbackFunc = isolated function(transaction:Info? info, error? cause, boolean willTry) {
                             io:println(" trxAborted");
                         };

    var onCommitFunc = isolated function(transaction:Info? info) {
                           io:println(" trxCommited");
                       };

    transaction {
        trx:onCommit(onCommitFunc);
        checkpanic  commit;
    }
}
```

**Affected Versions:**
SL Alpha 4

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

mohanvive:
Closing the issue because it tracked at high level, https://github.com/ballerina-platform/ballerina-lang/issues/30078

PR #30432, opened 2021-05-07
[LS] Refactor FunctionCallExpressionNodeContext to provide sorted completion items in paramter context

### LINKED ISSUE #30078: Completion suggestions are not accurate for function pointers
**Description:**
Function pointers as parameters are suggested with "()". This is not a valid scenario. 
 
**Steps to reproduce:**
Apr-19-2021 17-06-49

malinthar:
## Purpose
**1. Fix function pointer not rendering issue**

  >Solution: Provide function pointer completion item in addition to the function completion item when the context type of the node at cursor is FUNCTION

Sample: 

**2. Refactor `FunctionCallExpressionNodeContext` to provide sorted completion items  when the cursor is positioned within the parameter (argument) context.**

> With this change the completion items for the parameter context  of a function call expression is sorted in the following precedence depending on the assignability as an argument. 
>   
>   1. Variable symbols with assignable TypeDescriptor ( Function pointers are also considered variables)
>   2. Function symbols with assignable TypeDescriptor
>   3. Other symbols - Default Sorting precedence

Consider the following samples

Sample 1

Sample 2

Fixes #30078, #29748, #30529 

## Approach
Filter out visible symbols based on expected parameter type descriptor.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/FunctionCallExpressionNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/FunctionCallExpressionNodeContext.java
@@ -15,17 +15,28 @@
  */
 package org.ballerinalang.langserver.completions.providers.context;
 
+import io.ballerina.compiler.api.symbols.FunctionSymbol;
+import io.ballerina.compiler.api.symbols.Symbol;
+import io.ballerina.compiler.api.symbols.SymbolKind;
+import io.ballerina.compiler.api.symbols.TypeSymbol;
+import io.ballerina.compiler.api.symbols.VariableSymbol;
 import io.ballerina.compiler.syntax.tree.FunctionCallExpressionNode;
 import io.ballerina.compiler.syntax.tree.QualifiedNameReferenceNode;
 import io.ballerina.compiler.syntax.tree.Token;
 import org.ballerinalang.annotation.JavaSPIService;
+import org.ballerinalang.langserver.common.utils.CommonUtil;
 import org.ballerinalang.langserver.common.utils.completion.QNameReferenceUtil;
 import org.ballerinalang.langserver.commons.BallerinaCompletionContext;
 import org.ballerinalang.langserver.commons.completion.LSCompletionException;
 import org.ballerinalang.langserver.commons.completion.LSCompletionItem;
+import org.ballerinalang.langserver.completions.SymbolCompletionItem;
+import org.ballerinalang.langserver.completions.util.ContextTypeResolver;
+import org.ballerinalang.langserver.completions.util.SortingUtil;
+import org.eclipse.lsp4j.CompletionItemKind;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 
 /**
  * Completion Provider for {@link FunctionCallExpressionNode} context.
@@ -34,6 +45,7 @@ import java.util.List;
  */
 @JavaSPIService("org.ballerinalang.langserver.commons.completion.spi.BallerinaCompletionProvider")
 public class FunctionCallExpressionNodeContext extends BlockNodeContextProvider<FunctionCallExpressionNode> {
+
     public FunctionCallExpressionNodeContext() {
         super(FunctionCallExpressionNode.class);
     }
@@ -50,7 +62,7 @@ public class FunctionCallExpressionNodeContext extends BlockNodeContextProvider<
         completionItems.addAll(this.expressionCompletions(ctx));
         // TODO: implement the following
 //        completionItems.addAll(this.getNewExprCompletionItems(ctx, node));
-
+        this.sort(ctx, node, completionItems);
         return completionItems;
     }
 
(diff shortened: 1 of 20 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/FunctionCallExpressionNodeContext.java
```diff
@@ -62,4 +88,155 @@ public boolean onPreValidation(BallerinaCompletionContext context, FunctionCallE
...
+        completionItems.add(new SnippetCompletionItem(context, Snippet.KW_FALSE.get()));
+        completionItems.add(new SnippetCompletionItem(context, Snippet.KW_CHECK.get()));
+        completionItems.add(new SnippetCompletionItem(context, Snippet.KW_CHECK_PANIC.get()));
+        completionItems.add(new SnippetCompletionItem(context, Snippet.KW_IS.get()));
+        completionItems.add(new SnippetCompletionItem(context, Snippet.EXPR_ERROR_CONSTRUCTOR.get()));
+        completionItems.add(new SnippetCompletionItem(context, Snippet.EXPR_OBJECT_CONSTRUCTOR.get()));
+        completionItems.add(new SnippetCompletionItem(context, Snippet.EXPR_BASE16_LITERAL.get()));
+        completionItems.add(new SnippetCompletionItem(context, Snippet.EXPR_BASE64_LITERAL.get()));
+
+        Optional<FunctionCallExpressionNode> exprNode = findFunctionCallExpressionNode(context.getNodeAtCursor());
```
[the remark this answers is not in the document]
malinthar: @IMS94 I have actually overridden that method for providing parameter context aware completions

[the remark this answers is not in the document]
IMS94: @malinthar Yes, noticed that. My concern is, if there's duplicated logic in parent method and this, we can reuse with the above approach. If that's not possible, no issue.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/FunctionCallExpressionNodeContext.java
```diff
@@ -62,4 +88,155 @@ public boolean onPreValidation(BallerinaCompletionContext context, FunctionCallE
...
+     * @param parameterTypeSymbol expected type symbol corresponding to the current argument position.
+     * @param visibleSymbols      visible function and variable symbols.
+     * @return {@link List<LSCompletionItem>}
+     */
+    private List<LSCompletionItem> getParameterContextCompletionItemList(BallerinaCompletionContext ctx,
+                                                                         TypeSymbol parameterTypeSymbol,
+                                                                         List<Symbol> visibleSymbols) {
+        List<LSCompletionItem> completionItems = new ArrayList<>();
+
+        List<Symbol> processedSymbols = new ArrayList<>();
```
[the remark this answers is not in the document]
malinthar: Done!

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/BallerinaCompletionContextImpl.java
```diff
@@ -82,4 +86,25 @@ public void addResolver(Node node) {
...
+        }
+        ContextTypeResolver contextTypeResolver = new ContextTypeResolver(this);
+        this.contextType = getNodeAtCursor().apply(contextTypeResolver);
+
+        NonTerminalNode node = getNodeAtCursor();
+        while (contextType == null && node.parent() != null) {
+            this.contextType = Optional.ofNullable(node.parent().apply(contextTypeResolver).get());
+        }
+        if (contextType == null) {
+            contextType = Optional.empty();
```
nadeeshaan, line 106: Revamp this logic according to the above comment

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ContextTypeResolver.java
```diff
@@ -149,7 +150,7 @@ public ContextTypeResolver(PositionedOperationContext context) {
                 && (symbol.kind() == SymbolKind.TYPE_DEFINITION || symbol.kind() == SymbolKind.CLASS)
                 && symbol.getName().get().equals(node.identifier().text());
         List<Symbol> moduleContent = QNameReferenceUtil.getModuleContent(context, node, predicate);
-        if (moduleContent.size() > 1) {
+        if (moduleContent.size() > 1 || moduleContent.isEmpty()) {
```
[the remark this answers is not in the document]
malinthar: > The size check also covers the `isEmpty()` check.

If the the list is empty we would get a IndexOutOfBounds Excepcion here in line 157
https://github.com/ballerina-platform/ballerina-lang/blob/9921e9b89f6f2fb3894299c72667a2981f8babd3/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ContextTypeResolver.java#L152-L157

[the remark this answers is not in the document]
nadeeshaan: Then we can check for `size != 1`

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/BallerinaCompletionContextImpl.java
malinthar: Done

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/BallerinaCompletionContextImpl.java
```diff
@@ -82,4 +87,30 @@ public void addResolver(Node node) {
...
+        }
+
+        this.isContextTypeFound = true;
+        if (this.getNodeAtCursor() == null) {
+            this.contextType = Optional.empty();
+            return this.contextType;
+        }
+
+        NonTerminalNode node = getNodeAtCursor();
+        do {
```
nadeeshaan, line 105: Is there any specific reason to use a loop here? The parent visit should be carry out by the visitor itself. Hence we can avoid using the loop here and depend on the `ContextTypeResolver`'s behaviour

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/BallerinaCompletionContextImpl.java
```diff
@@ -37,6 +40,8 @@
     private final List<Node> resolverChain = new ArrayList<>();
     private Token tokenAtCursor;
     private NonTerminalNode nodeAtCursor;
+    private boolean isContextTypeFound = false;
+    private Optional<TypeSymbol> contextType;
```
nadeeshaan, line 44: Usually we don't use Optional as a field's type. Shall we change it?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/BallerinaCompletionContextImpl.java
malinthar: There can be instances where the parent is not visited. The ExternalTreeList Node implementation is one such instance. (The apply method does not visit the transformer). Hense, the parent will not be visited. IMO, to handle such scenarios a loop is required since we can't guarantee that the ContextTypeResolver always resolves a context for a given node

nadeeshaan: +1. Let's proceed with this approach then. Better if we can have a discussion on getting a null value from the `ExternalTreeList`.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/BallerinaCompletionContextImpl.java
```diff
@@ -82,4 +87,32 @@ public void addResolver(Node node) {
...
+        do {
+            ContextTypeResolver contextTypeResolver = new ContextTypeResolver(this);
+            Optional<TypeSymbol> typeSymbol = node.apply(contextTypeResolver);
+            if (typeSymbol == null || typeSymbol.isEmpty()) {
+                this.contextType = null;
+            } else {
+                this.contextType = typeSymbol.get();
+            }
+            node = node.parent();
+        } while (this.contextType == null && node.parent() != null);
```
nadeeshaan, line 114: Shall we add a null check for the node as well?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/BallerinaCompletionContextImpl.java
malinthar: Done

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/BallerinaCompletionContextImpl.java
malinthar: done, Ideally, the null check should be for the node not the  node.parent()

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/BallerinaCompletionContextImpl.java
malinthar: >  Better if we can have a discussion on getting a null value from the `ExternalTreeList`.
+1

nadeeshaan (approved):
LGTM

PR #30462, opened 2021-05-10
Add syntax tree node to respective ast nodes

### LINKED ISSUE #27042: Store syntax diagnostics in respective AST Node
**Description:**

In the current Toml Parser Implementation, Syntax tree contains all the diagnostics in the correct fields, but when we are transforming the Syntax tree to AST, we append all the diagnostics to rootTable. This makes loss of values when querying diagnostics via inner child tables. As a solution we should store the appropriate syntax tree diagnostics in respective AST node.

### LINKED ISSUE #30220: Misleading extra diagnostic thrown when a value is not present in Ballerina.toml and Cloud.toml
**Description:**
When a value is not present in Ballerina.toml or Config.toml, 2 error messages are logged for the same location as shown below. 
```
ERROR [Ballerina.toml:(3:1,3:1)] incompatible type for key 'org': expected 'STRING', found 'BOOLEAN'
ERROR [Ballerina.toml:(3:1,3:1)] missing value
```
The first diagnostic shown above error is misleading. IMO, we should not default an empty value to boolean. We should continue with null or empty and avoid the first error.

**Steps to reproduce:**

Modify the `Ballerina.toml` of a project as follows:

```toml
[package]
org = 
name = "proj"
version = "0.1.0"
```

**Affected Versions:**
At least slalpha5

### LINKED ISSUE #30395: Avoid performing custom validations when toml syntax errors exists
**Description:**
Currently, Schma based validations will be done even when the node contains syntax diagnostics. This results in multiple diagnostics. We should avoid appending another diagnostic when syntax diagnostic exist in that specific node.

**Describe your problem(s)**

**Describe your solution(s)**

### LINKED ISSUE #30406: Toml Parser Syntax tree diagnostics should contain in respective AST nodes
**Description:**
$Subject

**Describe your problem(s)**

**Describe your solution(s)**

xlight05:
## Purpose
Earlier we only had DocumentNode as a reference to root syntax tree node. In this PR we add corresponding child nodes to each AST node. 
As each ast node contains respective syntax tree node now, diagnostics is based on that node. So the addtional step of appending sytnax diagnostics to ast nodes is not needed now. (Config API and Project API changes)
Adds addtional isMissingNode property to AST to check if the AST node is an injected one
TomlValidator now doesn't give type errors if the node is an injected one.

Resolves #30406
Resolves #30395
Resolves #30220
Resolves #27042

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlStringValueNode.java
```diff
@@ -20,20 +20,45 @@
...
     }
 
     @Override
     public void accept(TomlNodeVisitor visitor) {
         visitor.visit(this);
     }
+
+    @Override
+    public boolean isMissingNode() {
+        if (this.externalTreeNode().kind() == SyntaxKind.STRING_LITERAL) {
```
lochana-chathura, line 48: `this.externalTreeNode()` can be extracted to separate varaible as it is repeated.

lochana-chathura (approved):

PR #30487, opened 2021-05-11
[LS]  Change AnnotationUtil to avoid empty mapping constructor for annotatons with optional fields

### LINKED ISSUE #29355: Should we avoid adding the empty mapping constructor for annotations?
**Description:**
With https://github.com/ballerina-platform/ballerina-lang/pull/28938 the mapping constructor expression is now optional in an annotation if the empty mapping constructor is a valid constructor for that type.

```ballerina
annotation record {| int i?; int j = 10; |} x on function;

annotation record {| int i; |} y on function;

@x // valid - same as `@x {}`
@y // invalid, because `i` is a required field
function func() {

}
```

**Describe your problem(s)**
Currently, accepting the suggestion for such an annotation adds the empty mapping constructor expression too.

```ballerina
annotation record {| int i?; int j = 10; |} x on function;

@x<cursor>
function func() {

}
```

annotation

**Describe your solution(s)**
Should the default be to not add it instead? If the fields are optional, IMO, it's quite likely that the common case would be to not specify values for them?

malinthar:
## Purpose
$subject

ezgif com-gif-maker(4)

Fixes #29355 

## Approach
Check if the required fields of the record symbol is empty before adding the constructor snippet.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/AnnotationUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/AnnotationUtil.java
@@ -198,22 +198,24 @@ public class AnnotationUtil {
             if (resultType.isPresent() && (resultType.get().typeKind() == TypeDescKind.RECORD
                     || resultType.get().typeKind() == TypeDescKind.MAP)) {
                 List<RecordFieldSymbol> requiredFields = new ArrayList<>();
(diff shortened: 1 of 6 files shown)


nadeeshaan (approved):
LGTM

Issue #30271, opened 2021-05-01
JvmCodegen throws an error when lambda is returned from a function that returns a union


MaryamZi:
**Description:**
$title, results in an error at the backend.

**Steps to reproduce:**
```ballerina
public function main() returns error? {
    function () y = check getFunction();
}

function getFunction() returns function ()|error => function () {};
```

```cmd
$ bal run foo.bal 

Compiling source
        foo.bal
error: expected BInvokableType, found (function () returns ()|error)
```

**Affected Versions:**
slalpha5

rdhananjaya:
Turned out this had nothing to do with the `check`, just having the function returning the function is enough to re-produce this issue.

```ballerina
function getFunction() returns function ()|error => function () {};
```

PR #30494, opened 2021-05-11
Fix jvmCodegen throws an error when lambda is returned from a function that returns a union

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30271

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/evaluation/ExpressionEvaluationTest.java
```diff
@@ -139,7 +139,8 @@ public void variableReferenceEvaluationTest() throws BallerinaTestException {
         // error variable test
         debugTestRunner.assertExpression(context, ERROR_VAR, "SimpleErrorType", "error");
         // anonymous function variable test
-        debugTestRunner.assertExpression(context, ANON_FUNCTION_VAR, "function (string,string) returns (string)",
+        debugTestRunner.assertExpression(context, ANON_FUNCTION_VAR, "function isolated function (string,string) " +
```
MaryamZi, line 142: Is having function before the function type expected here?

MaryamZi (commented):
Isolated function related changes LGTM.

tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/evaluation/ExpressionEvaluationTest.java
NipunaRanasinghe: This happens because the `stringValue()` method of the `io.ballerina.runtime.internal.values.FPValue` adds a `function` prefix, in https://github.com/ballerina-platform/ballerina-lang/blob/865850e24ff2882178085c7b0ed3948248e5df8c/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/FPValue.java#L86.

I had a call with Waruna and he will try removing the prefix, which would solve the issue.

NipunaRanasinghe (commented):
Debugger related changes look good to me.

tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/evaluation/ExpressionEvaluationTest.java
MaryamZi: Sorry, missed this earlier. But I think this is intended. For values where the spec doesn't define a string representation, the implementation defines its own and we seem to have used the basic type-descriptor + type as the string representation.

For example, https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/TypedescValueImpl.java#L93. So I guess we have to keep this?

warunalakshitha: +1. I have reverted the change. @NipunaRanasinghe I think as per @MaryamZi comment, we need to fix it in the debugger side.

NipunaRanasinghe: @warunalakshitha @MaryamZi Ack.. I'll modify the debugger side to preprocess the type string before sending it to the frontend.

warunalakshitha: Created https://github.com/ballerina-platform/ballerina-lang/issues/30604

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -2833,7 +2833,7 @@ private void generateFPVarRef(BLangExpression fpVarRef, BInvokableSymbol funcSym
 
         setScopeAndEmit(
                 new BIRNonTerminator.FPLoad(fpVarRef.pos, funcSymbol.pkgID, funcName, lhsOp, params, new ArrayList<>(),
-                        funcSymbol.retType, funcSymbol.strandName, funcSymbol.schedulerPolicy));
+                        funcSymbol.type, funcSymbol.strandName, funcSymbol.schedulerPolicy));
```
MaryamZi, line 2836: FPLoad seems to be expecting the return type though, right? https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/model/BIRNonTerminator.java#L733

Since we have the params also, we can probably construct the function type by passing the flags also.

Wondering if the `returnType.tag != TypeTags.INVOKABLE` check in JvmInstructionGen L1578 was unnecessary to begin with.

MaryamZi: If we are using the function type, shouldn't we rename the field and maybe get rid of params too?

warunalakshitha: Actually, it should be a function type. Even we pass the type earlier we haven't used it during codegen. Only place used is https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/methodgen/LambdaGen.java#L427:L429

Based on the logic, it should still work. 

I will rename returnType -> type.

warunalakshitha: IMO We can still keep the returnType.tag != TypeTags.INVOKABLE check to catch any missing cases in the future.

MaryamZi (approved):

Issue #30403, opened 2021-05-06
No intersection error shown for a valid error intersection


IMS94:
**Description:**
Consider the following error intersection:
```
type Foo record {|
    int i;
|};

type Bar record {|
    int i;
|};

type FooError error<Foo> & error<Bar>;
```
It shows an error with message `invalid intersection type '$anonType$_276 & $anonType$_277': no intersection(BCE2026)`.

**Steps to reproduce:**
See description

**Affected Versions:**
Beta 1 SNAPSHOT

rdhananjaya:
Related? https://github.com/ballerina-platform/ballerina-lang/issues/30118

Issue #30401, opened 2021-05-06
Intersection of two records cause an NPE


IMS94:
**Description:**
Consider the following code snippet:
```
type Foo record {|
    int i;
|};

type Bar record {
    int i;
};

type FooBar Foo & Bar;
```

It cause an NPE with the following logs:
```
[Error - 5:03:23 PM] Operation 'ballerinaDocument/executorPositions' failed! {uri: '/home/imesha/Documents/WSO2/Ballerina/Projects/test_project/projects/lstest/intersection_types.bal', error: 'java.lang.NullPointerException, org.wso2.ballerinalang.compiler.semantics.analyzer.Types.getAllTypes(Types.java:4778)'}
java.lang.NullPointerException
	at org.wso2.ballerinalang.compiler.semantics.analyzer.Types.getAllTypes(Types.java:4778)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.Types.getTypeIntersection(Types.java:4069)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.Types.createRecordIntersection(Types.java:4341)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.Types.getIntersection(Types.java:4106)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.Types.getTypeIntersection(Types.java:4072)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolResolver.getPotentialIntersection(SymbolResolver.java:2094)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolResolver.computeIntersectionType(SymbolResolver.java:1914)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolResolver.visit(SymbolResolver.java:1109)
	at org.wso2.ballerinalang.compiler.tree.types.BLangIntersectionTypeNode.accept(BLangIntersectionTypeNode.java:48)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolResolver.resolveTypeNode(SymbolResolver.java:479)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolResolver.resolveTypeNode(SymbolResolver.java:470)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.visit(SymbolEnter.java:1369)
[...]

Issue #30118, opened 2021-04-21
No intersection with closed error detail record


jclark:
With slalpha4, the compiler rejects this:

```
type MyDetail record {|
    string s;
|};

type MyError error<MyDetail> & error;
```
with

```
invalid intersection type '$anonType$_0 & error': no intersection
```

It works if I leave the `& error` out.

PR #30511, opened 2021-05-12
Improve intersection type support

rdhananjaya:
## Purpose
Allow closed records to be intersected when possible. And improve the error message for intersection between types that we don't handle at the moment.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30118
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30403
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30401

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -1919,6 +1919,12 @@ private BType computeIntersectionType(BLangIntersectionTypeNode intersectionType
             isErrorIntersection = true;
         }
 
+        if ((typeOne.tag == typeTwo.tag && typeOne.tag != TypeTags.ERROR)
+            || (!hasReadOnlyType && (typeOne.tag != TypeTags.ERROR || typeTwo.tag != TypeTags.ERROR))) {
```
MaryamZi, line 1923: Can't we use `!isErrorIntersection && !hasReadOnlyType && (typeOne.tag != TypeTags.ERROR || typeTwo.tag != TypeTags.ERROR)` instead?

MaryamZi: Also, shouldn't this result in one of two errors? One for intersections that are not yet supported vs one for invalid intersection types (DiagnosticErrorCode.INVALID_INTERSECTION_TYPE)?

`int & string x = 1;` also results in `unsupported intersection 'int & string'` with these changes rather than `invalid intersection type 'int & string': no intersection`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangUserDefinedType.java
```diff
@@ -71,9 +71,18 @@ public void accept(BLangNodeVisitor visitor) {
 
     @Override
     public String toString() {
+        String typeName = this.typeName.value;
+
+        if (typeName.startsWith("$") && this.type != null) {
```
MaryamZi, line 76: Can we check for the anon flag instead?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangUserDefinedType.java
```diff
@@ -71,9 +71,18 @@ public void accept(BLangNodeVisitor visitor) {
...
+            String tName = this.type.toString();
+            if (!tName.equals("other")) {
```
MaryamZi, line 78: When is it "other"? :thinking: Shall we add a comment too?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangUserDefinedType.java
rdhananjaya: Since this is `toString` method I think we need to make sure we can call this method at any point in compiler phases.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangUserDefinedType.java
rdhananjaya: Will add explaining comment.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
rdhananjaya:
```ballerina
type A int & int;
```
This crash for me in SL Alpha-5

MaryamZi: Shall we create an issue? We need to fix the error message for Swan Lake anyway, right?

MaryamZi:
The build seems to have failed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
rdhananjaya: Issue created https://github.com/ballerina-platform/ballerina-lang/issues/31101

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -1931,19 +1931,22 @@ private BType computeIntersectionType(BLangIntersectionTypeNode intersectionType
             isErrorIntersection = true;
         }
 
+        if (!(hasReadOnlyType || isErrorIntersection)) {
+            dlog.error(intersectionTypeNode.pos,
+                    DiagnosticErrorCode.UNSUPPORTED_TYPE_INTERSECTION, intersectionTypeNode);
+            return symTable.noType;
```
MaryamZi, line 1937: Shouldn't this be `symTable.semanticError`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangUserDefinedType.java
rdhananjaya: No, the type is `SymbolTable.noType` (and symbol in null) since I don't have access to an instance of that class, only way I can check for that is to use `toString` and see if it's `other`.

MaryamZi: Can't we check the type tag?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangUserDefinedType.java
```diff
@@ -71,9 +73,20 @@ public void accept(BLangNodeVisitor visitor) {
...
+        String typeName = this.typeName.value;
+
+        BType thisType = this.getBType();
+        if (typeName.startsWith("$") && thisType != null) {
+
+            // When there are errors in this type, we can't just use the tName.
+            // Fall back to pkgAlias:typeName pattern
+            if (thisType.tag != TypeTags.NONE) {
+                return thisType.toString();
+            }
```
MaryamZi, line 85:
```suggestion
        // When there are errors in this type, we can't just use the shape.
        // Fall back to pkgAlias:typeName pattern.
        if (typeName.startsWith("$") && thisType != null && thisType.tag != TypeTags.NONE) {
            return thisType.toString();
```

MaryamZi (approved):

dulajdilshan (approved):
LGTM

MaryamZi (approved):

PR #30523, opened 2021-05-12
Add support for in-built compiler plugins

### LINKED ISSUE #30351: Support for built-in extentions
**Description:**
Currently a compiler extension is distributed in a package and you need to import it to activate the extension. But in some cases you need to ship a compiler extension with the distribution itself and they should be engaged by default. 

As a solution we can implement a concept of "built in extensions" where the extension will be discover and leaded via a java SPI. The extensions can be added to the distribution as an when needed. 

Couple of use cases this will be useful includes
* Data validator extension used for data mapper
* CLI option parser validator 

Basically the above extensions will add additional compile time validations which cannot be included in the compiler.

azinneera:
Fixes #30351

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/projects/CompilerPluginManager.java
```diff
@@ -53,10 +56,34 @@ static CompilerPluginManager from(PackageCompilation compilation) {
         DependencyGraph<ResolvedPackageDependency> dependencyGraph = packageResolution.dependencyGraph();
         List<Package> directDependencies = getDirectDependencies(rootPkgNode, dependencyGraph);
         List<CompilerPluginInfo> compilerPlugins = loadEngagedCompilerPlugins(directDependencies);
+        List<CompilerPluginInfo> inBuiltCompilerPlugins = loadInBuiltCompilerPlugins(rootPkgNode.packageInstance());
+        compilerPlugins.addAll(inBuiltCompilerPlugins);
         List<CompilerPluginContextIml> compilerPluginContexts = initializePlugins(compilerPlugins);
         return new CompilerPluginManager(compilation, compilerPluginContexts);
     }
 
+    private static List<CompilerPluginInfo> loadInBuiltCompilerPlugins(Package rootPackage) {
```
sameerajayasoma, line 65: I can suggest the following changes. 
1) The built-in compiler plugins do not change across compilations, but package-provided plugins may change per compilation. Therefore we could load and cache built-in compiler plugins. Use plugins from the cache in subsequent compilations. 
2) Move the class loading logic to the `CompilerPlugins.java` class. You can cache the built-in plugin list in that class. Perhaps you could load the plugins in a static block even.

sameerajayasoma (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/CompilerPluginManager.java
azinneera: Addressed in 40a3bf6

compiler/ballerina-lang/src/main/java/io/ballerina/projects/CompilerPluginInfo.java
```diff
@@ -34,6 +35,12 @@
     private final PackageDescriptor packageDesc;
     private final CompilerPluginDescriptor compilerPluginDesc;
 
+    CompilerPluginInfo(CompilerPlugin compilerPlugin) {
```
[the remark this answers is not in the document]
azinneera: Addressed in 40a3bf6

sameerajayasoma (approved):

PR #30619, opened 2021-05-17
Initialize static fields at code-gen level

### LINKED ISSUE #26139: Inconsistent behavior when generating byte code for ballerina constant definitions
**Description:**
As a part of adding debugging support for ballerina global scope variables (https://github.com/ballerina-platform/ballerina-lang/issues/25275), debugger implementation is accessing the JVM runtime (`$_init` class), in order to fetch and display ballerina global variable/constant declarations. But as per the current codegen implementation, ballerina constants definition information for primitive types are not available in the generated byte code (defined, but not initialized). 

**Steps to reproduce:**
Decompile and observe the `$_init` class of the generated bytecode of the code sample below, to see how these constants are represented in the runtime, 

```ballerina 
const nameWithoutType = "Ballerina";    // Not accessible in runtime
const string nameWithType = "Ballerina"; // Not accessible in runtime
const map<string> gv02_nameMap = {"asd":"abc"};   // Accessible in runtime

public function main() {
    
}
```

**Affected Versions:**

### LINKED ISSUE #28884: [Debugger] Incorrect enum variable values in global variable scope
**Description:**
Enum variables are not visible in debug mode.

**Steps to reproduce:**
- Consider the below code segment:
```
enum Language {
    EN = "English",
    TA = "Tamil",
    SI = "Sinhala"
}

public function main() {
    int x = 5;
}
```
- Run the above program in debug mode by adding a breakpoint at the last line of `main()` method.
- We could observe that the enum variables are not visible in debug mode

image

grainier:
## Purpose
$title, so that initialized constant fields will be visible at runtime (i.e for debugger).

@NipunaRanasinghe, @praveennadarajah please verify. 

Fixes #26139
Fixes #28884

## Approach
n/a

## Samples
Before:
Screen Shot 2021-05-17 at 7 02 21 PM

After the fix:
Screen Shot 2021-05-17 at 7 08 44 PM

## Remarks
-  #26139, #28884

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmCodeGenUtil.java
```diff
@@ -577,6 +577,20 @@ public static String cleanupFunctionName(String functionName) {
                 "$" + JVM_RESERVED_CHAR_SET.matcher(functionName).replaceAll("_") : functionName;
     }
 
+    public static boolean isSimpleBasicType(BType bType) {
```
riyafa, line 580: If this function is only used in the `JvmPackageGen` class we can move it there. This glass is normally used only if we are writing a function that is used in multiple places

NipunaRanasinghe (commented):
@grainier I've tested locally from the debugger side and the fix seems to be working fine for all ballerina constant types. LGTM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmCodeGenUtil.java
warunalakshitha: Or we can move to TypeTags class

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmPackageGen.java
```diff
@@ -235,25 +235,46 @@ private static void generateLockForVariable(ClassWriter cw) {
...
+        JvmCodeGenUtil.generateStrandMetadata(mv, className, birPackage.packageID, asyncDataCollector);
         mv.visitInsn(RETURN);
         mv.visitMaxs(0, 0);
         mv.visitEnd();
     }
 
+    private static void setConstantFields(MethodVisitor mv, BIRPackage birPackage, String className,
+                                          JvmBStringConstantsGen stringConstantsGen) {
+        for (BIRNode.BIRConstant constant : birPackage.constants) {
+            if (constant != null) {
```
warunalakshitha, line 262: Can this become null?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmPackageGen.java
```diff
@@ -235,25 +235,46 @@ private static void generateLockForVariable(ClassWriter cw) {
...
+    private static void generateStaticInitializer(ClassWriter cw, String className, BIRPackage birPackage,
+                                                  boolean isInitClass, boolean serviceEPAvailable,
+                                                  AsyncDataCollector asyncDataCollector,
+                                                  JvmBStringConstantsGen stringConstantsGen) {
         if (!isInitClass && asyncDataCollector.getStrandMetadata().isEmpty()) {
             return;
         }
         MethodVisitor mv = cw.visitMethod(ACC_STATIC, "<clinit>", "()V", null, null);
         if (isInitClass) {
+            setConstantFields(mv, birPackage, className, stringConstantsGen);
```
warunalakshitha, line 247: This might get method too large issue if we have around 6000 constants in the module since we initialize all constants in <clinit> method. Please refer JvmBStringConstantsGen class. Shall we test that too?

warunalakshitha: A better way would be to use separate constant gen-class and initialize the constants there without using init class static initializer.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmCodeGenUtil.java
manuranga: I think we already have`isSimpleBasicType` functions maybe it's possible to reuse, let's do this refactoring after beat1.

manuranga: #30639

manuranga (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmPackageGen.java
manuranga: We are alrady initilizing at some point, right? So the issue is with ide is not picking it up.

warunalakshitha: Earlier we used to replace the constants with exact values. So constants are not initialized.

warunalakshitha: Created https://github.com/ballerina-platform/ballerina-lang/issues/30648

PR #30625, opened 2021-05-17
Fix syntaxTree gen tool to capture endpoint outside a function

NipunaMarcus:
## Purpose
  - Fix syntaxTree gen tool to capture endpoint outside a function
  - Add tests for diagram util

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/diagram-util/src/test/resources/mainHttpCall/mainHttpCall.bal
```diff
@@ -0,0 +1,8 @@
+import ballerina/http;
+
+http:Client myclient = check new ("http://abcd");
```
nadeeshaan, line 3: Is it okay to have standard library here? Does this import get resolved during the test phase?

misc/diagram-util/src/main/java/org/ballerinalang/diagramutil/SyntaxTreeMapGenerator.java
```diff
@@ -236,6 +240,19 @@ private JsonObject visibleEP(Node node, TypeSymbol typeSymbol) {
                 symbolMetaInfo.addProperty("orgName", orgName);
                 symbolMetaInfo.addProperty("moduleName", moduleName);
             }
+        } else if (node.kind() == SyntaxKind.MODULE_VAR_DECL) {
+            JsonObject metaInfoForModuleVar = new JsonObject();
+            ModuleVariableDeclarationNode variableDeclarationNode = (ModuleVariableDeclarationNode) node;
+            CaptureBindingPatternNode captureBindingPatternNode =
+                    (CaptureBindingPatternNode) variableDeclarationNode.typedBindingPattern().bindingPattern();
+            metaInfoForModuleVar.addProperty("name", captureBindingPatternNode.variableName().text());
+            metaInfoForModuleVar.addProperty("isCaller", "Caller".equals(typeSymbol.getName()
```
nadeeshaan, line 249: Type name cannot always be caller. Hence we have to check the semantic properties to verify whether it is a caller or not. Otherwise this is going to be a specific case being handled.

nadeeshaan (changes requested):

misc/diagram-util/src/main/java/org/ballerinalang/diagramutil/SyntaxTreeMapGenerator.java
NipunaMarcus: it is a specific case being handled here.... This is very specific to HTTP:Caller

misc/diagram-util/src/test/resources/mainHttpCall/mainHttpCall.bal
NipunaMarcus: seems no hence the coverage issue... i will create my own client and use it ...

misc/diagram-util/src/main/java/org/ballerinalang/diagramutil/SyntaxTreeMapGenerator.java
nadeeshaan: Since this mapper is a public API and can be used by other parties as well, shall we use the semantic API to get the client information? Also, in language point of view, are going to refer a client?

NipunaMarcus: this is very specific to the diagram ... this Json generator is basically built for that... and what API we can use to get the client information? ... this is very very specific to identifying caller in HTTP resource.

nadeeshaan (approved):
LGTM

hevayo (approved):

PR #30654, opened 2021-05-18
Clear compiler package cache when Dependencies.toml is edited

### LINKED ISSUE #30359: Modules are not updated for Dependencies.toml edit
**Description:**
When the Dependencies.toml is edited, the modules should be reset.

**Steps to reproduce:**
Create two versions of the same package with a signature change in a function and check diagnostics in VScode by switching the version.

1. Create a new lib package, build and push to local repo.
    $ bal new -t lib proj1
   **_proj1.bal_**
   ```bal
   import ballerina/io;

   public function hello() {
       io:println("Hello World!");
   }
   ```
   _**Ballerina.toml**_
   ```toml
    [package]
    org = "testOrg"
    name = "proj1"
    version = "0.1.0"
    ```
    
2. Edit the main.bal and the Ballerina.toml of the package as follows:
   Edit the `proj1.bal` file to add a parameter to the `hello` function
**_proj1.bal_**
   ```bal
   import ballerina/io;

   public function hello(string p1) {
       io:println("Hello World!");
   }
   ```
   Edit the Ballerina.toml to change the version
   _**Ballerina.toml**_
   ```toml
    [package]
    org = "testOrg"
    name = "proj1"
    version = "0.1.1"
    ```
3. Build and push the updated package to local repo.
4. Create a new project that uses the first package an open in VSCode
   $ bal new projTest
   _**main.bal**_
   ```bal
    import testOrg/proj1;
    
    public function main() {
        proj1:hello();
        
    }
   ``` 
    _**Dependencies.toml**_
   ```toml
    [[dependency]]
    org = "testOrg"
    name = "proj1"
    version = "0.1.0"
    ```
5. Change the version in `Dependencies.toml` to `0.1.1`

**Expected behavior:**
    Show diagnostic `missing required parameter 'p1' in call to 'hello()'`
    Go-to-def would direct to the file in version `0.1.1`

**Current behavior**
   No diagnostics shown
   Go-to-def would direct to the file in version `0.1.0`

azinneera:
Fixes #30359

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/projects/Package.java
```diff
@@ -412,8 +415,22 @@ private Package createNewPackage() {
...
                     this.compilerPluginTomlContext, this.packageMdContext,  this.compilationOptions,
-                    this.moduleContextMap, this.pkgDependencyGraph);
+                    this.moduleContextMap, DependencyGraph.emptyGraph());
             this.project.setCurrentPackage(new Package(newPackageContext, this.project));
+
+            List<ResolvedPackageDependency> newSortedList = this.project.currentPackage().getResolution()
+                    .dependencyGraph().toTopologicallySortedList();
+            List<ResolvedPackageDependency> oldSortedList = this.dependencyGraph.toTopologicallySortedList();
+
+            for (int i = 0; i < newSortedList.size(); ++i) {
```
hevayo, line 425: We can compare the sizes before comparing each item. will save bit of time when a new dependency is added. But ideally when a new dependency is added we do not need to clear the cache.

sameerajayasoma: +1. We can improve this logic a bit more. 

So we have two dependency graphs here. Old graph O and new graph N. Can we figure out the number of node additions and the number of node deletions that we need to apply to graph O to get graph N? We don't worry about the structure of the graph; we just need to additions and deletions. 

If there are deletions, we clear the cache, otherwise keep the cache as it is.

sameerajayasoma (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/Package.java
azinneera: In the case of Dependencies.toml edit, there is only a version change in nodes. The number of nodes will stay consistent since the imports are resolved from the source files and only the version is captured from the Dependencies.toml. This improvement will be applicable for document edits (since there can be a change in import statements) but can we apply this for this case?

sameerajayasoma: Yes, what we need is to figure out deletions. Deletions should tell you whether you need to clear the cache or not.

azinneera: Addressed in 1c11376

compiler/ballerina-lang/src/main/java/io/ballerina/projects/DependencyGraph.java
```diff
@@ -60,6 +60,30 @@ private DependencyGraph(Map<T, Set<T>> dependencies) {
...
     }
 
+    /**
+     * Compares two instances of {@code DependencyGraph}.
+     *
+     * @param other other dependency graph to compare with
+     * @return DependencyGraph.Compatibility.COMPATIBLE if graphs are compatible;
+     *         DependencyGraph.Compatibility.INCOMPATIBLE otherwise
+     */
+    public Compatibility compareTo(DependencyGraph<T> other) {
```
sameerajayasoma, line 70: Let `O` be the set of nodes in the old dependency graph and `N` be the new dependency graph set. So you are checking whether `O-N` is empty to figure out whether there are any deleted nodes. The logic is correct; however, I don't think this method name and return type make sense.

1) Shall we not make this a public method for now? This is only used for internal implementation at the moment
2) We can rename this method as `difference` and change the return signature to return a collection of nodes.

sameerajayasoma (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/DependencyGraph.java
azinneera: Addressed in 7b14ab4

sameerajayasoma (dismissed):

hevayo (approved):

PR #30692, opened 2021-05-20
[LS] Add custom sorting logic to variable declaration node context

### LINKED ISSUE #30671: [LS] Improve completion item sorting in  VariableDeclarationNodeContext
**Description:**
$subject to provide sorted completion items based on contextual information.

**Describe your problem(s)**

For instance, in the following code segment the the desired completion item order at the cursor position is as follows.

```ballerina 
function getString() returns string {
  return "hello";
}
public function main() {
  int var1 = 10;
  string var2 = "var2";
  int var3 = 20;

  string var4  =  <cursor>
}
```
       

>  var2
>         getString()
>         Other visible symbols and snippets
>         ...

malinthar:
## Purpose
$subject to provide sorted completion items in variable declaration node context.

With this change the completion items are sorted in the following precedence. 

>  1. Variable symbols with assignable TypeDescriptor ( Function pointers are also considered variables)
>  2. Function symbols with assignable TypeDescriptor
>  3. Other symbols - Default Sorting precedence

Fixes #30671 

## Approach
Add custom sorting logic based on the type symbol of the L.H.S.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/VariableDeclarationNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/VariableDeclarationNodeContext.java
@@ -15,15 +15,23 @@
  */
 package org.ballerinalang.langserver.completions.providers.context;
 
+import io.ballerina.compiler.api.symbols.FunctionSymbol;
+import io.ballerina.compiler.api.symbols.Symbol;
+import io.ballerina.compiler.api.symbols.TypeSymbol;
 import io.ballerina.compiler.syntax.tree.VariableDeclarationNode;
 import io.ballerina.tools.text.TextRange;
 import org.ballerinalang.annotation.JavaSPIService;
+import org.ballerinalang.langserver.common.utils.SymbolUtil;
 import org.ballerinalang.langserver.commons.BallerinaCompletionContext;
 import org.ballerinalang.langserver.commons.completion.LSCompletionException;
 import org.ballerinalang.langserver.commons.completion.LSCompletionItem;
+import org.ballerinalang.langserver.completions.SymbolCompletionItem;
+import org.ballerinalang.langserver.completions.util.SortingUtil;
+import org.eclipse.lsp4j.CompletionItemKind;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 /**
  * Completion provider for {@link VariableDeclarationNode} context.
@@ -47,7 +55,7 @@ public class VariableDeclarationNodeContext extends VariableDeclarationProvider<
         List<LSCompletionItem> completionItems
                 = this.initializerContextCompletions(context, node.typedBindingPattern().typeDescriptor());
         this.sort(context, node, completionItems);
-        
+
         return completionItems;
     }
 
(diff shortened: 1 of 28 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/VariableDeclarationNodeContext.java
```diff
@@ -60,4 +68,49 @@ public boolean onPreValidation(BallerinaCompletionContext context, VariableDecla
...
+                SymbolCompletionItem symbolCompletionItem = (SymbolCompletionItem) completionItem;
+                Optional<Symbol> completionSymbol = symbolCompletionItem.getSymbol();
+                if (completionSymbol.isEmpty()) {
+                    completionItem.getCompletionItem().setSortText(SortingUtil.genSortText(rank));
+                    return;
+                }
+                Optional<TypeSymbol> evalTypeSymbol = SymbolUtil.getTypeDescriptor(completionSymbol.get());
+                CompletionItemKind completionItemKind = completionItem.getCompletionItem().getKind();
+                if (evalTypeSymbol.isPresent()) {
+                    switch (completionItemKind) {
```
nadeeshaan, line 94: Can't we handle it without considering the item kind?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/VariableDeclarationNodeContext.java
```diff
@@ -60,4 +68,49 @@ public boolean onPreValidation(BallerinaCompletionContext context, VariableDecla
...
+                        case Variable:
+                            if (evalTypeSymbol.get().assignableTo(symbol)) {
+                                rank = 1;
+                            }
+                            break;
+                        case Method:
+                        case Function:
```
nadeeshaan, line 101: Missing the default case

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/VariableDeclarationNodeContext.java
malinthar: Since we have added the function pointer as an completion item that is of variable kind we require this implementation. So we can't just rely on symbol kind  in this case.

nadeeshaan: What if we add a new CompletionItemType here? Is it possible?

malinthar: IMO, the correct completion item type for function pointer is variable. What is your suggestion as the type?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/VariableDeclarationNodeContext.java
malinthar: fixed

nadeeshaan (approved):
LGTM

PR #30694, opened 2021-05-20
Fix non-isolated method scheduling to be sequential - part 2

### LINKED ISSUE #29322: Support runtime api for non isolated function scheduling

manuranga:
Fixes #29322
Part of the fix was missing in #30374 this pr address that

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
@@ -81,8 +81,8 @@ public class Runtime {
          if (isIsolate) {
              return scheduler.schedule(new Object[1], func, null, callback, strandName, metadata).result;
          } else {
-             return scheduler.scheduleToObjectGroup(object, new Object[1], func, null, callback, strandName, metadata)
-                     .result;
+             return scheduler.scheduleToObjectGroup(object, new Object[1], func, null, callback, null,
+                                                    PredefinedTypes.TYPE_NULL, strandName, metadata).result;
          }
      }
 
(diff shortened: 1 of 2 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
```diff
@@ -81,8 +81,8 @@ public Object invokeMethodAsync(BObject object, String methodName, String strand
          if (isIsolate) {
              return scheduler.schedule(new Object[1], func, null, callback, strandName, metadata).result;
          } else {
-             return scheduler.scheduleToObjectGroup(object, new Object[1], func, null, callback, strandName, metadata)
-                     .result;
+             return scheduler.scheduleToObjectGroup(object, new Object[1], func, null, callback, null,
```
warunalakshitha, line 84: Shall we call the this invokeMethodAsync method with other function.

ex
      */
     public Object invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
                                     Callback callback, Object... args) {
         return invokeMethodAsync(object, methodName, strandName, metadata, callback, null,
                           PredefinedTypes.TYPE_NULL, args).getResult();

     }

warunalakshitha: Maybe we can do as a separate PR to master. Then we do not need to duplicate  if (isIsolate) { logic .

manuranga: I think two functions are duplicates, shall we remove one of them altogether ?

PR #30709, opened 2021-05-21
Fix readonly-ness of slice of readonly arrays

### LINKED ISSUE #30353: slice of readonly array should not be readonly
This program

```
public function main() {
   readonly & string[] roNames = ["x"];
   string[] rwNames = roNames.slice(0);
   rwNames[0] = "y";
}
```

gets a panic at runtime:

```
{ballerina/lang.array}InvalidUpdate {"message":"modification not allowed on readonly value"}
```

The spec isn't as clear as it should be here. But the declaration of `slice` does not say that the result should get its readonly-ness from the input. If the docs don't say to create something as readonly, then in general you shouldn't.

gabilang:
Fixes #30353

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ArrayValueImpl.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ArrayValueImpl.java
@@ -31,7 +31,6 @@ import io.ballerina.runtime.api.values.BValue;
 import io.ballerina.runtime.internal.CycleUtils;
 import io.ballerina.runtime.internal.TypeChecker;
 import io.ballerina.runtime.internal.types.BArrayType;
-import io.ballerina.runtime.internal.types.BIntersectionType;
 import io.ballerina.runtime.internal.util.exceptions.BLangExceptionHelper;
 import io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons;
 import io.ballerina.runtime.internal.util.exceptions.BallerinaException;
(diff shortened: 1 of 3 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ArrayValueImpl.java
```diff
@@ -801,27 +802,33 @@ public ArrayValueImpl slice(long startIndex, long endIndex) {
...
                 System.arraycopy(floatValues, (int) startIndex, slicedArray.floatValues, 0, slicedSize);
                 break;
             case TypeTags.STRING_TAG:
             case TypeTags.CHAR_STRING_TAG:
-                slicedArray = new ArrayValueImpl(new BString[slicedSize], arrayType.isReadOnly());
+                slicedArray = new ArrayValueImpl(new BString[slicedSize], false);
                 System.arraycopy(bStringValues, (int) startIndex, slicedArray.bStringValues, 0, slicedSize);
                 break;
             default:
+                if (this.elementType.getTag() == TypeTags.INTERSECTION_TAG) {
```
HindujaB, line 826: Shall we add this as a new case?
```java
case TypeTags.INTERSECTION_TAG:
```

warunalakshitha (approved):

PR #30755, opened 2021-05-24
Add an API for checking if an object type is compatible with `Listener`

pubudu91:
## Purpose
This PR introduces a new method to `ObjectTypeSymbol` for checking whether the object type can be used in a listener declaration context.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/ObjectTypeSymbol.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/ObjectTypeSymbol.java
@@ -50,4 +50,12 @@ public interface ObjectTypeSymbol extends TypeSymbol, Qualifiable {
      * @return The list of included object types
      */
     List<TypeSymbol> typeInclusions();
+
+    /**
+     * Checks whether this object type can be used in a listener declaraion. i.e., whether this object type is a subtype
+     * of the builtin `Listener` type.
+     *
+     * @return Returns true if this type is a subtype of `Listener`
+     */
+    boolean isCompatibleListenerType();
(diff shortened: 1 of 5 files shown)


mohanvive (dismissed):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/ObjectTypeSymbol.java
```diff
@@ -50,4 +50,12 @@
...
      */
     List<TypeSymbol> typeInclusions();
+
+    /**
+     * Checks whether this object type can be used in a listener declaraion. i.e., whether this object type is a subtype
+     * of the builtin `Listener` type.
+     *
+     * @return Returns true if this type is a subtype of `Listener`
+     */
+    boolean isCompatibleListenerType();
```
hasithaa, line 60: Another option is we can introduce another utility Class to check this. I guess we have similar validation in other places too.

pubudu91: True, but the issue is, where would such a util fit in? This is sort of like the same discussion we had about the check for assignability and lang lib methods.

IMS94 (approved):

pubudu91:
Closing this since we decided not to go with the approach taken in this PR to address this issue.

Issue #28567, opened 2021-02-11
Single quoted variables are missing the quote in the name


nadeeshaan:
**Description:**
Consider the following source snippet,

`string 'string = "Hello World!";`

If we get the symbol name, the name does not include the `'`, instead the literal is returned.

**Related Issues**
#18156

jclark:
What's wrong with that? The symbol name is `string`, and you should quote it when printing.

These are the same:

```
string 'string;
string \u{73}tring;
```

These are also the same:

```
int x;
int 'x;
```

Same thing with `x\-y`: the symbol name is `x-y`, but on output it should be converted to `x\-y`.

Same concept as string escaping.

sameerajayasoma:
I guess we need to have a flag in the symbol name in the implementation which indicates that the identifier is a quoted identifier or we can preserve the original quoted identifier.

jclark:
@sameerajayasoma Are you going to preserve that when it is written as `\u{73}string`? You need a list of reserved keywords, then when you have an identifier that is a reserved keyword, you quote it. Just as with a string, you store the result of processing escapes (things starting with a backslash) and then add backslashes when you output as necessary.

sanjana:
Related to #29462

pubudu91:
@nadeeshaan Isn't this the same as #29462? If so, shall we close this as well? With the fix done for #29462, we now get the quoted identifier from the symbols names when looking them up through the Semantic API.

nadeeshaan:
> @nadeeshaan Isn't this the same as #29462? If so, shall we close this as well? With the fix done for #29462, we now get the quoted identifier from the symbols names when looking them up through the Semantic API.

@pubudu91 yes, this is the same one. Let's close this as well.

pubudu91:
Closing this as per the above comment. This was fixed using the approach James suggested above.

PR #30778, opened 2021-05-25
Skip escape-character of method name in method-call-expressions

### LINKED ISSUE #30242: No completions are provided for field access experssions
**Description:**
Consider the following source snippet and no completions are provided at the cursor position.

`string header = "+ ".'join('join(" + ", headers)).j<cursor>` Completions are not provided

**Affected Versions:**
sl-alpha5 at least

dulajdilshan:
## Purpose
$title

With this PR `'join` lang-lib method will be available as a completion
Screenshot from 2021-05-25 15-35-30

Fixes #30242 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/FieldAccessCompletionResolver.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/FieldAccessCompletionResolver.java
@@ -84,7 +84,7 @@ public class FieldAccessCompletionResolver extends NodeTransformer<Optional<Type
     @Override
     public Optional<TypeSymbol> transform(SimpleNameReferenceNode node) {
         Optional<Symbol> symbol = this.getSymbolByName(context.visibleSymbols(context.getCursorPosition()),
-                node.name().text());
+                unescape(node.name().text()));
 
         if (symbol.isEmpty()) {
             return Optional.empty();
@@ -102,7 +102,7 @@ public class FieldAccessCompletionResolver extends NodeTransformer<Optional<Type
         if (fieldName.kind() != SyntaxKind.SIMPLE_NAME_REFERENCE) {
             return Optional.empty();
         }
-        String name = ((SimpleNameReferenceNode) fieldName).name().text();
+        String name = unescape(((SimpleNameReferenceNode) fieldName).name().text());
         List<Symbol> visibleEntries = this.getVisibleEntries(typeSymbol.orElseThrow(), node.expression());
         Optional<Symbol> filteredSymbol = this.getSymbolByName(visibleEntries, name);
 
@@ -122,10 +122,7 @@ public class FieldAccessCompletionResolver extends NodeTransformer<Optional<Type
         }
         Predicate<Symbol> predicate = symbol -> symbol.kind() == SymbolKind.METHOD
                 || symbol.kind() == SymbolKind.FUNCTION;
-        String methodName = ((SimpleNameReferenceNode) nameRef).name().text();
-        if (methodName.startsWith(IDENTIFIER_LITERAL_PREFIX)) {
-            methodName = IdentifierUtils.unescapeUnicodeCodepoints(methodName.substring(1));
-        }
+        String methodName = unescape(((SimpleNameReferenceNode) nameRef).name().text());
         List<Symbol> visibleEntries = this.getVisibleEntries(exprTypeSymbol.orElseThrow(), node.expression());
 
         FunctionSymbol symbol = (FunctionSymbol) this.getSymbolByName(visibleEntries, methodName, predicate);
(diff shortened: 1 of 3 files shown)


IMS94 (commented):
@dulajdilshan Do we have a test case to cover this scenario? If not, will be better if we can add one. The same scenario will need to be addressed in other nodes in future as well. Therefore, this will be a good starting point.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/FieldAccessCompletionResolver.java
```diff
@@ -121,6 +123,9 @@ public FieldAccessCompletionResolver(PositionedOperationContext context, boolean
         Predicate<Symbol> predicate = symbol -> symbol.kind() == SymbolKind.METHOD
                 || symbol.kind() == SymbolKind.FUNCTION;
         String methodName = ((SimpleNameReferenceNode) nameRef).name().text();
+        if (methodName.startsWith(IDENTIFIER_LITERAL_PREFIX)) {
+            methodName = IdentifierUtils.unescapeUnicodeCodepoints(methodName.substring(1));
+        }
```
pubudu91, line 128: Instead of special casing this in one place, shall we follow a similar approach to the node transformer and fix all the places in this visitor? i.e., Move this name extraction logic to a method and use that method anywhere a name is accessed in this visitor

dulajdilshan: @pubudu91 But, usually we don't change the attributes of the nodes in the Syntax tree, which we get from the parser. Right? Here, In the `FieldAccessCompletionResolver.java`, we get the method name from a `MethodCallExpressionNode`. :thinking:

pubudu91: We aren't changing anything. But since the compiler keeps all identifiers in unescaped form, need to use the unescaped form when comparing the names we get from the syntax tree nodes with names taken from symbols. Basically the idea is, each place in the `FieldAccessCompletionResolver` where we take a name from a syntax tree node, we unescape it first. This fix only unescapes method call expressions.

dulajdilshan: Now I got the idea. Thanks @pubudu91

IMS94 (dismissed):
LGTM!

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/FieldAccessCompletionResolver.java
```diff
@@ -288,4 +290,11 @@ private boolean withValidAccessModifiers(Node exprNode, MethodSymbol methodSymbo
         return isPublic || (!isPrivate && objModuleId.moduleName().equals(currentModule.moduleName())
                 && objModuleId.orgName().equals(currentPackage.packageOrg().value()));
     }
+
+    private String unescape(String name) {
+        if (name.startsWith(IDENTIFIER_LITERAL_PREFIX)) {
+            return IdentifierUtils.unescapeUnicodeCodepoints(name.substring(1));
+        }
+        return name;
+    }
```
nadeeshaan, line 299: Why do we need to handle this specifically here? Shouldn't the value be escaped/ unescaped before consuming in LS?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/FieldAccessCompletionResolver.java
pubudu91: Since this is a syntax tree visitor, it's getting the values as-is. It's up to the visitor to decide whether to unescape or not right?

nadeeshaan: My concern is, that we have some information preserved in the syntax tree and those information is not available in the semantic models. If we are going to have a similar implementation in other places, we do have to do the unescaping.

dulajdilshan: Yeah.

Talked with @lochana-chathura, It is okay to have this `unescape()` function inside `IdentifierToken` and retrieve the unescaped form of the text of the name. So we can use it any other place we have a similar problem. WDYT?

nadeeshaan: What's the possibility of preserving the actual value (with the quote) in the symbol as well? This is because if we get the  symbol after escaping or not, the symbol which is being visible to the will have the name without the quote

pubudu91: Storing the unescaped actual value shouldn't be an issue I think. When @sanjana was working on the quoted identifier issue with symbols, we followed the approach James mentioned here: https://github.com/ballerina-platform/ballerina-lang/issues/28567. The issue with this approach is this only covers adding the quote back to reserved words. We can check the feasibility of preserving the original value the others and see.

nadeeshaan: IMHO if we address the quoted identifier issue, this issue will be resolved automatically and will not be needing a custom fix.

dulajdilshan:
Invalid

PR #30801, opened 2021-05-26
Introduce several changes/fixed for `isolated`

### LINKED ISSUE #30412: Multiple error messages for the same `isolated` related error
Here is the code example:
```
import ballerina/http;
import ballerina/log;
import ballerinax/twilio;

// configurable variables are implicitly final and readonly
final configurable string accountSId = ?;
final configurable string authToken = ?;
final configurable string fromMobile = ?;
final configurable string toMobile = ?;

// Mutable module-level variables has to be declared as isolated,
//  if you are planning to use them within an isolated service
isolated twilio:Client twilioClient = new ({
    accountSId: accountSId,
    authToken: authToken
});

    isolated resource function post foo() returns error? {
        twilio:SmsResponse smsResp;
        lock {
            smsResp = check twilioClient->sendSms(fromMobile, toMobile, "Hi!");
        }
        log:printDebug(msg = "SMS send successful", id = smsResp.accountSid, body = smsResp.body);
    }
}
```

Compiler output:
```
Compiling source
        sameera/services:0.1.0
ERROR [main.bal:(22:29,22:41)] invalid attempt to transfer out a value from a 'lock' statement with restricted variable usage
ERROR [main.bal:(22:51,22:61)] invalid attempt to transfer out a value from a 'lock' statement with restricted variable usage
ERROR [main.bal:(22:63,22:71)] invalid attempt to transfer out a value from a 'lock' statement with restricted variable usage
```

### LINKED ISSUE #30800: Isolation inference doesn't happen for methods in a service object constructor
**Description:**
$title. Happens when it's part of a service declaration though.

**Affected Versions:**
slbeta1-rc4

### LINKED ISSUE #30802: Isolated transfer out analysis should use a top down approach
**Description:**
$title. 

Right now when we have something like `return [1, x, 2]`, the analysis starts from the members and looks back until the list constructor to see if it part of an isolated expression, if required. But the spec mentions this starting from the entire expression in the return statement.

AFAICT the current approach doesn't result in the invariant being violated, but this seems to be allowing scenarios that are not considered as isolated expressions by the spec (like `from var f in self.array select f`).

**Affected Versions:**
slbeta1-rc4

MaryamZi:
## Purpose
This PR
- logs a warning for non-isolated services and non-isolated remote/resource methods - This would currently result in warnings even in scenarios where the spec says a service will be inferred as isolated but the implementation doesn't infer so yet.
- fixes #30800 - isolation inference doesn't happen for methods in a service object constructor
- improves some isolation transfer in/out errors. Fixes #30412. 
- fixes #30802 - isolated transfer out analysis should use a top down approach

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/IsolationAnalyzer.java
```diff
@@ -3014,6 +3015,47 @@ private boolean inferredIsolated(BInvokableSymbol symbol, IsolationInferenceInfo
         return true;
     }
 
+    private void logServiceIsolationWarnings(List<BLangClassDefinition> classDefinitions) {
```
MaryamZi, line 3018: This would currently result in warnings even in scenarios where the spec says a service/method will be inferred as isolated but the implementation doesn't infer so yet.

hasithaa (approved):

PR #30808, opened 2021-05-27
Allow using `constant-expr` as table key field `value-expr`

### LINKED ISSUE #30723: Table key value should allow `const-expr`s
**Description:**
$title since the spec says,
> For every field-name in the key sequence of the inherent type every mapping-constructor-expr in the row-list must include a specific-field that has a `value-expr` that is a `const-expr`.

We only allow `literals` ATM. 

**Steps to reproduce:**
```ballerina
const s = "foo";

type Row record {
   readonly string k;
   int value;
};

table<Row> key(k) t = table [
   { k: s, value: 17 }
];
```
`ERROR [main.bal:(11:4,11:23)] field 'k' used in key specifier must have a literal value`

**Affected Versions:**
slalpha5

lasinicl:
## Purpose
$title

Fixes #30723

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1097,12 +1102,69 @@ private boolean validateTableType(BTableType tableType, List<BLangRecordLiteral>
...
+            case GROUP_EXPR:
+                return isConstExpression(((BLangGroupExpr) expression).expression);
+            default:
+                return false;
+        }
+    }
+
+    private BLangRecordKeyValueField getRecordKeyValueField(BLangRecordLiteral recordLiteral,
+                                                            String fieldName) {
+        for (RecordLiteralNode.RecordField recordField : recordLiteral.fields) {
```
gimantha, line 1158: cant we use streams here? we need to see if there is a performance issue using java streams

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/DataflowAnalyzer.java
```diff
@@ -914,6 +914,39 @@ public Integer hash(Node node) {
         } else if (node.getKind() == NodeKind.SIMPLE_VARIABLE_REF) {
             BLangSimpleVarRef simpleVarRef = (BLangSimpleVarRef) node;
             result = simpleVarRef.variableName.hashCode();
+        } else if (node.getKind() == NodeKind.STRING_TEMPLATE_LITERAL) {
```
gimantha, line 917: What will happen if two tables have different containers as keys, but the expressions are same? will there be a key collision? for example one table, string template and other table, list constructor? there can be a key collision as we take exprs only to calculate the hash?

gimantha (approved):

Issue #30818, opened 2021-05-28
Revisit org.wso2.ballerinalang.compiler.bir.codegen.JvmDesugarPhase#encodeModuleIdentifiers


MaryamZi:
**Description:**
$title. With type definition instances potentially shared across modules of the same package, replacing the symbol's name with the encoded name lead to issues such as https://github.com/ballerina-platform/ballerina-lang/issues/30814.

MaryamZi:
The following is another scenario that fails based on just quoted identifiers (as opposed to readonly intersections in https://github.com/ballerina-platform/ballerina-lang/issues/30814).

In one package have a sub-module `mod` and the default module.

Sub-module `mod`
```
public type Bar 'Will\.be\&encoded|boolean;

public type 'Will\.be\&encoded record {|
    int i;
    Bar j;
|};
```

Default module
```ballerina
import pkg.mod;

public type Foo record {|
    mod:Bar bar;
|};
```

Import the default module in another module and use it as follows.
```ballerina
import test/pkg;

pkg:Foo x = {bar: {i: 0, j: true}};
```

Fails on beta1-rc5 and even with #30821 with the following error because the name is encoded and noType is set as the union member type.
```cmd
ERROR [main.bal:(4:19,4:34)] a type compatible with mapping constructor expressions not found in type '(other|boolean)'
```

PR #30858, opened 2021-05-29
Replace encoding identifier names after codegen

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30818

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/CodeGenerator.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/CodeGenerator.java
@@ -30,6 +30,8 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.util.HashMap;
+
 /**
  * JVM byte code generator from BIR model.
  *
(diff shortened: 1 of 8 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/CodeGenerator.java
```diff
@@ -83,10 +85,15 @@ private CompiledJarFile generate(BPackageSymbol packageSymbol) {
...
         //Rewrite identifier names with encoding special characters
-        JvmDesugarPhase.encodeModuleIdentifiers(packageSymbol.bir, Names.getInstance(this.compilerContext));
+        HashMap<String, String> originalIdentifierMap = JvmDesugarPhase
+                .encodeModuleIdentifiers(packageSymbol.bir, Names.getInstance(this.compilerContext));
 
         // TODO Get-rid of the following assignment
         packageSymbol.compiledJarFile = jvmPackageGen.generate(packageSymbol.bir, true);
+
+        //Revert encoding identifier names
+        JvmDesugarPhase.replaceEncodedModuleIdentifiers(packageSymbol.bir, Names.getInstance(this.compilerContext),
```
warunalakshitha, line 95: There are two ways to fix this issue.

1) Decode all encoded identifier names after codegen.
2) Track the original identifier names before encoding and then replace encoded identifier names after codegen with original strings.

1 option didn't work since for some string original string and decoded string retrieved via encoded string was not the same.
ex.If you encode "Will.be&encoded", then decode it you will get "Will\.be\&encoded"

warunalakshitha: @MaryamZi @HindujaB

MaryamZi: @warunalakshitha, but doesn't that mean that our decoding logic is not sound? Couldn't this be an issue when decoding at runtime?

HindujaB: This is because we have 3 different representations for an identifier for different purposes. 
1. identifiers that user defined before encoding (with escape characters) - Eg. `'Will\.be\&encoded`
2. identifiers that the user intended to have (without escaped characters) - Eg. `Will.be&encoded` 
     - Used for member access, etc. 
3. identifiers that are encoded (for jvm back-end) - Eg. `Will$0046be&encoded` 

The decode function will always return the identifier to its original state (with escaped characters)
Therefore, +1 to the approach.

HindujaB (approved):
LGTM

PR #30957, opened 2021-06-02
[Shell] Remove initial import statements

### LINKED ISSUE #30922: [Shell] Import for `jballerina.java` shown in the output for `/imports`
It'd be good if we can hide this if the user hasn't explicitly imported this.
```
=$ /imports                                                                                                                                                                                                         
| ('java) import ballerina/'jballerina.'java as 'java;
```

DamithaSenevirathne:
Fixes #30922

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/ClassLoadInvoker.java
+++ b/ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/ClassLoadInvoker.java
@@ -45,6 +45,7 @@ import io.ballerina.shell.utils.StringUtils;
 
 import java.io.PrintStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -539,9 +540,13 @@ public class ClassLoadInvoker extends ShellSnippetsInvoker {
     @Override
     public List<String> availableImports() {
         // Imports with prefixes
+        List<String> initialImports = getInitialImports();
         List<String> importStrings = new ArrayList<>();
         for (QuotedIdentifier prefix : importsManager.prefixes()) {
-            importStrings.add(String.format("(%s) %s", prefix, importsManager.getImport(prefix)));
+            String importStatement = String.format("(%s) %s", prefix, importsManager.getImport(prefix));
+            if (!initialImports.contains(importStatement)) {
+                importStrings.add(importStatement);
+            }
         }
         return importStrings;
     }
@@ -579,5 +584,13 @@ public class ClassLoadInvoker extends ShellSnippetsInvoker {
     @Override
     protected PrintStream getErrorStream() {
         return System.err;
+    }
+
+    /**
+     * @return Initial imports.
+     */
+    private List<String> getInitialImports() {
+        return Arrays.
+                asList("(\'java) import ballerina/\'jballerina.\'java as \'java;");
     }
 }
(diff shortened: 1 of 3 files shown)


ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/ClassLoadInvoker.java
```diff
@@ -539,9 +540,13 @@ private ClassLoadContext createVariablesExecutionContext(
...
     public List<String> availableImports() {
         // Imports with prefixes
+        List<String> initialImports = getInitialImports();
         List<String> importStrings = new ArrayList<>();
         for (QuotedIdentifier prefix : importsManager.prefixes()) {
-            importStrings.add(String.format("(%s) %s", prefix, importsManager.getImport(prefix)));
+            String importStatement = String.format("(%s) %s", prefix, importsManager.getImport(prefix));
+            if (!initialImports.contains(importStatement)) {
+                importStrings.add(importStatement);
+            }
```
pubudu91, line 549: Can't this be handled in the `ImportsManager` itself? i.e., have separate APIs in `ImportsManager` which will return the imports with and without this import?

@kdsuneraavinash thoughts?

kdsuneraavinash: If  I understand correctly, this implementation would hide `jballerina.java import` even if it is explicitly imported, no?

So I think the better way would be to modify `storeImport` in imports manager. 

https://github.com/ballerina-platform/ballerina-lang/blob/c5efa989cd35b9631916e6cf5de1fc399c94ba2c/ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/ImportsManager.java#L160-L164

However, simply hiding this import can cause several other issues,
- users cant use `java` import prefix because there is this implicit import with the same prefix. So they will be shown an error if they try to add `import ballerina/jballerina/java` explicitly.
- users still be able to use `java` prefix in code.

So these cases also need to be handled. But I think it is possible to handle these cases just by modifying imports manager.

kdsuneraavinash (approved):

ballerina-shell/modules/shell-core/src/main/java/io/ballerina/shell/invoker/classload/ImportsManager.java
```diff
@@ -303,5 +317,12 @@ private void storeAnonImplicitPrefixes() {
     private void storeImport(QuotedIdentifier quotedPrefix, QuotedImport moduleName) {
         this.imports.put(quotedPrefix, moduleName);
         this.reverseImports.put(moduleName, quotedPrefix);
+        if (!quotedPrefix.toString().equals(JAVA_PREFIX)) {
+            this.userImports.put(quotedPrefix, moduleName);
+        } else {
+            if (this.initialized.get()) {
+                this.userImports.put(quotedPrefix, moduleName);
+            }
+        }
```
pubudu91, line 326: Can't this be simplified to the following? 🤔 
```java
if (!quotedPrefix.toString().equals(JAVA_PREFIX) || this.initialized.get()) {
     this.userImports.put(quotedPrefix, moduleName);
}
```

pubudu91: Will go ahead and merge for now. If the above is possible, let's address in a separate PR.

pubudu91 (approved):

PR #30962, opened 2021-06-02
Fix Syntax Tree packaging issue when ballerina run is used

### LINKED ISSUE #27555: Ballerina run with observability-included is not working
**Description:**
The following command doesnot work.
`ballerina run <package-path> --observability-included --b7a.observability.enabled=true --b7a.observability.provider=choreo`

**Steps to reproduce:**

**Affected Versions:**
SLP8-snapshot

nadundesilva:

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/JarResolver.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/JarResolver.java
@@ -22,9 +22,6 @@ import org.wso2.ballerinalang.compiler.semantics.analyzer.ObservabilitySymbolCol
 import org.wso2.ballerinalang.compiler.spi.ObservabilitySymbolCollector;
 import org.wso2.ballerinalang.compiler.util.CompilerContext;
 
-import java.io.BufferedOutputStream;
-import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.net.MalformedURLException;
@@ -36,10 +33,6 @@ import java.security.PrivilegedAction;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-import java.util.UUID;
-import java.util.jar.Attributes;
-import java.util.jar.JarOutputStream;
-import java.util.jar.Manifest;
 
 import static io.ballerina.projects.util.ProjectConstants.ANON_ORG;
 import static io.ballerina.projects.util.ProjectConstants.DOT;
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/projects/JarResolver.java
```diff
@@ -80,6 +94,33 @@
...
+            manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0");
+
+            try {
+                String packageName = getPackageName(rootPackageContext);
+                Path jarPath = Path.of(System.getProperty("java.io.tmpdir"),
+                        packageName.replaceAll(File.separatorChar == '\\' ? "\\\\" : File.separator, "-")
+                                + "-" + UUID.randomUUID() + ".jar");
+                JarOutputStream jarOutputStream = new JarOutputStream(new BufferedOutputStream(
+                        new FileOutputStream(jarPath.toFile())), manifest);
+                jarOutputStream.close();
```
tharinduwijewardane, line 110: What is the purpose of creating a jarOutputStream and closing it without writing any entries?

nadundesilva: This is to generate a proper empty (still it will contain the manifest) Jar file which can be used by the ObservabilitySymbolCollector to store the syntax tree.

nadundesilva: Added a comment in fe87a0066595c9fa06d4693306192fa1e3d80386 to improve readability

compiler/ballerina-lang/src/main/java/io/ballerina/projects/JarResolver.java
```diff
@@ -80,6 +94,35 @@
...
+        // TODO: Move to a compiler extension once Compiler revamp is complete
+        // 4) Add the Observability Symbols Jar
+        if (rootPackageContext.compilationOptions().observabilityIncluded()) {
+            Manifest manifest = new Manifest();
+            manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0");
+
+            try {
+                // Generating an empty Jar which can be used by the Observability Symbol Collector
+                String packageName = getPackageName(rootPackageContext);
+                Path jarPath = Path.of(System.getProperty("java.io.tmpdir"),
```
hevayo, line 106: You can use Files.createTempFile here. Also can we create this in target so we can use it for caching ?

hevayo (commented):
please see inline

compiler/ballerina-lang/src/main/java/io/ballerina/projects/JarResolver.java
azinneera: Shall we move the jar creation to [ProjectUtils](https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/io/ballerina/projects/util/ProjectUtils.java) class?

azinneera: @hevayo ATM persisting the thin jars is handled by the `CompilationCache` and there is no way to check if observability is enabled in the CompilationCache class. 
https://github.com/ballerina-platform/ballerina-lang/blob/0ed6999d21d053407ab3759b8e8ddf3cc9431364/compiler/ballerina-lang/src/main/java/io/ballerina/projects/repos/FileSystemCache.java#L100-L102

Another way is to use the `Target` model but that would need a refactor to include the package descriptor info in our `Target` model since the jar caches path needs the `org`, `package name`, `version`, and `target platform` to create the directory structure. So I think having it in a temp location would suffice for now.

nadundesilva: Moved to ProjectUtils and changed the temp file creation to use `Files.createTempFile()` in b49b3969bdad31895d83daa58ac52f03e03524df

azinneera (approved):
LGTM

PR #30986, opened 2021-06-03
Cleanup code

pubudu91:
## Purpose
Cleans up the unused/unnecessary code and modules.


settings.gradle
```diff
@@ -34,7 +34,6 @@ include(':stdlib')
 //include(':ballerina-crypto')
```
MaryamZi, line 34: We should probably remove what's commented out too, right? Maybe we can do that in a separate PR.

hasithaa (dismissed):

stdlib/http/build.gradle
```diff
@@ -1,122 +0,0 @@
...
-    interopImports project(':ballerina-runtime-api')
-    interopImports project(':ballerina-mime')
-
-    // transitive
-    balaImplementation project(path: ':ballerina-task', configuration: 'balaImplementation')
-
-    interopImports project(':ballerina-io')
-
-    implementation 'io.netty:netty-codec-http'
-    implementation ('org.wso2.transport.http:org.wso2.transport.http.netty') {
```
warunalakshitha, line 91: There are some netty dependencies in io.netty:netty-codec-http javaProject.gradle file. Shall we remove them too?

pubudu91: Ack. Will remove and see.

warunalakshitha: We can fix with a separate PR as well.

compiler/ballerina-lang/src/test/java/org/wso2/ballerinalang/compiler/BLangNodeTransformerTest.java
```diff
@@ -280,16 +246,12 @@ private static JsonObject fileContentAsObject(String filePath) {
...
-                }
-            }, true, "UTF-8");
-        }
+    private static Project loadProject(String sourceFilePath) {
+        Path sourcePath = Paths.get(sourceFilePath);
+        String sourceFileName = sourcePath.getFileName().toString();
+        Path sourceRoot = RES_DIR.resolve(sourcePath.getParent());
+        Path projectPath = Paths.get(sourceRoot.toString(), sourceFileName);
+        BuildOptionsBuilder buildOptionsBuilder = new BuildOptionsBuilder();
+        return ProjectLoader.loadProject(projectPath, buildOptionsBuilder.build());
```
azinneera, line 255: We can use `ProjectLoader.loadProject(projectPath)`. It will set the default build options.

tests/jballerina-integration-test/src/test/java/org/ballerinalang/test/util/SQLDBUtils.java
```diff
@@ -18,6 +18,7 @@
 package org.ballerinalang.test.util;
```
MaryamZi, line 18: IINM, none of the tests that use this class seem to be running. Shall we create an issue to check if they are still valid/belong in the lang repo and enable/remove these.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/packageimport/PackageImportTest.java
```diff
@@ -93,7 +93,7 @@ public void testUnusedImports() {
         BAssertUtil.validateError(result, i, "unused import module 'ballerina/jballerina.java as otherJAVA'", 2, 1);
     }
 
-    @Test
+    @Test(enabled = false)
```
MaryamZi, line 96: Why are we disabling this?

stdlib/http/build.gradle
pubudu91: I think we'll need to take a closer look at these. Seems to be in used in tests and test utils.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/packageimport/PackageImportTest.java
pubudu91: Need to think of a different way to write this. This uses `ballerina/io` to print a message in module init function and then capture that to a buffer and compare the outputs to verify the behaviour. With the `ballerina/io` module removed, can't do it that way. So disabled it for now, will create an issue for this.

MaryamZi: We can probably use interop similar to https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/jballerina-unit-test/src/test/resources/test-src/execution/proj8/main.bal#L18.

Shall we create an issue if we are disabling this?

tests/jballerina-integration-test/src/test/java/org/ballerinalang/test/util/SQLDBUtils.java
pubudu91: Ack. Will create an issue for this.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/packageimport/PackageImportTest.java
pubudu91: Yeah, a util like this should work fine. Will address this separately.

azinneera (approved):

MaryamZi (approved):

PR #31020, opened 2021-06-07
Change type parameters to MapType in lang.table

### LINKED ISSUE #30608: Fix incorrect `table` type usages in `lang.table`
**Description:**
Refer https://github.com/ballerina-platform/ballerina-spec/issues/849

gimantha:
Fixes #30608

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -273,6 +273,10 @@ private BType createBuiltInType(BType type, Name name, long flags) {
                 anydataType.name = name;
                 anydataType.flags |= flags;
                 return anydataType;
+            case TypeTags.MAP:
```
MaryamZi, line 276: Not sure if we can consider map as a built-in type in this context based on the other types we are handling here. Do we know where it breaks if we do not add this here?

gimantha: yes. we can remove it. i ll remove it

langlib/lang.table/src/main/ballerina/table.bal
```diff
@@ -160,7 +166,7 @@ public isolated function reduce(table<Type> t, @isolatedParam function(Type1 acc
 # + t - the table
 # + k - the key
 # + return - the member of parameter `t` that had key parameter `k`
-public isolated function remove(table<Type> key<KeyType> t, KeyType k) returns Type = @java:Method {
+public isolated function remove(table<MapType> key<KeyType> t, KeyType k) returns MapType = @java:Method {
```
lasinicl, line 169:
```ballerina
type MapType map<any|error>;
type KeyType anydata;

public function main() {
    table<MapType> key<KeyType> t = table [];
}
```
Specifying a `key-type-constraint` for a `table` constrained with a `map` type results in `table with constraint of type map cannot have key specifier or key type constraint` compilation error. Shouldn't the same be applicable here when such `table` type is given as a function param?

lasinicl:
```ballerina
function foo(table<MapType> key<KeyType> t) {}
```
This should not be allowed right? This doesn't result in a compilation error ATM.

gimantha: hmm. yeah. I think we will have to handle it specifically, by checking if it is not a langlib function. will send the fix with this pr

lasinicl (approved):

PR #31023, opened 2021-06-07
Handle int add, sub, mul overflow

### LINKED ISSUE #20799: Should use java.lang.Math.addExact() for runtime int add operation
**Description:**

$subject.
Currently its a LADD instruction and doesn't handle byte overflow.

Nadeeshan96:
Previously int addition, subtraction, and multiplication overflowed
without giving a runtime error (panic). Now throws a panic and a 
message when int overflow happens. Added 3 unit tests for 
checking int addition, subtraction, and multiplication overflow.

Fixes #20799

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
@@ -23,7 +23,7 @@ import io.ballerina.runtime.api.values.BString;
 import io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons;
 
 /**
- * Common utility methods used for arithmatic operations.
+ * Common utility methods used for arithmetic operations.
  *
  * @since 1.0
  */
@@ -33,12 +33,16 @@ public class MathUtils {
 
     private static final BString INT_RANGE_OVERFLOW_ERROR = StringUtils.fromString(" int range overflow");
 
+    private static void throwIntRangeOverflowError() {
+        throw ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+    }
+
     public static long divide(long numerator, long denominator) {
         try {
             if (numerator == Long.MIN_VALUE && denominator == -1) {
                 // a panic will occur on division by zero or overflow,
                 // which happens if the first operand is -2^63 and the second operand is -1
-                throw ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+                throwIntRangeOverflowError();
             }
             return numerator / denominator;
         } catch (ArithmeticException e) {
@@ -65,26 +69,29 @@ public class MathUtils {
     }
 
     public static long addExact(long num1, long num2) {
-        long r = num1 + num2;
-        if (((num1 ^ r) & (num2 ^ r)) < 0) {
-            throw  ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+        long result = num1 + num2;
+        // an overflow has happened on addition if the sign of the result is different from BOTH operands' sign
+        if (((num1 ^ result) & (num2 ^ result)) < 0) {
+            throwIntRangeOverflowError();
         }
-        return r;
+        return result;
     }
 
     public static long subtractExact(long num1, long num2) {
-        long r = num1 - num2;
-        if (((num1 ^ num2) & (num1 ^ r)) < 0) {
-            throw  ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+        long result = num1 - num2;
+        // an overflow has happened on subtraction when both operands have the same sign AND
+        // if the sign of the result is different from the first operand's sign
+        if (((num1 ^ num2) & (num1 ^ result)) < 0) {
+            throwIntRangeOverflowError();
         }
-        return r;
+        return result;
     }
 
     public static long multiplyExact(long num1, long num2) {
-        long r = num1 * num2;
-        if (num1 != 0 && r / num1 != num2) {
-            throw  ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+        long result = num1 * num2;
+        if ((num1 != 0 && result / num1 != num2) || (num2 != 0 && result / num2 != num1)) {
+            throwIntRangeOverflowError();
         }
-        return r;
+        return result;
     }
 }
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
@@ -63,4 +63,28 @@ public class MathUtils {
             }
         }
     }
+
+    public static long addExact(long num1, long num2) {
+        try {
+            return Math.addExact(num1, num2);
+        } catch (ArithmeticException e) {
+            throw ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+        }
+    }
+
+    public static long subtractExact(long num1, long num2) {
+        try {
+            return Math.subtractExact(num1, num2);
+        } catch (ArithmeticException e) {
+            throw ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+        }
+    }
+
+    public static long multiplyExact(long num1, long num2) {
+        try {
+            return Math.multiplyExact(num1, num2);
+        } catch (ArithmeticException e) {
+            throw ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+        }
+    }
 }
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
@@ -19,6 +19,7 @@ package io.ballerina.runtime.internal;
 
 import io.ballerina.runtime.api.creators.ErrorCreator;
 import io.ballerina.runtime.api.utils.StringUtils;
+import io.ballerina.runtime.api.values.BError;
 import io.ballerina.runtime.api.values.BString;
 import io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons;
 
@@ -33,8 +34,8 @@ public class MathUtils {
 
     private static final BString INT_RANGE_OVERFLOW_ERROR = StringUtils.fromString(" int range overflow");
 
-    private static void throwIntRangeOverflowError() {
-        throw ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+    public static BError getIntOverflowError() {
+        return ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
     }
 
     public static long divide(long numerator, long denominator) {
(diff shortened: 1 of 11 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
```diff
@@ -63,4 +63,28 @@ public static long remainder(long numerator, long denominator) {
             }
         }
     }
+
+    public static long addExact(long num1, long num2) {
+        try {
+            return Math.addExact(num1, num2);
```
warunalakshitha, line 69: Should we check overflow check here instead of directly calling addExact so we can get rid of the try catch?
ex

    public static long addExact(long x, long y) {
        long r = x + y;
        // HD 2-12 Overflow iff both arguments have the opposite sign of the result
        if (((x ^ r) & (y ^ r)) < 0) {
            throw  ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
        }
        return r;
    }

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
```diff
@@ -63,4 +63,28 @@ public static long remainder(long numerator, long denominator) {
             }
         }
     }
+
+    public static long addExact(long num1, long num2) {
+        try {
```
warunalakshitha, line 68: Should we check overflow check here instead of directly calling addExact so we can get rid of the try-catch?
ex
```java
    public static long addExact(long x, long y) {
        long r = x + y;
        // HD 2-12 Overflow iff both arguments have the opposite sign of the result
        if (((x ^ r) & (y ^ r)) < 0) {
            throw  ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
        }
        return r;
    }

```

Nadeeshan96: I will look into this.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
```diff
@@ -63,4 +63,28 @@ public static long remainder(long numerator, long denominator) {
...
+            throw  ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+        }
+        return r;
+    }
+
+    public static long subtractExact(long num1, long num2) {
+        long r = num1 - num2;
+        if (((num1 ^ num2) & (num1 ^ r)) < 0) {
+            throw  ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
+        }
```
riyafa, line 79: It would be good to move the common code to a new function

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
```diff
@@ -63,4 +63,28 @@ public static long remainder(long numerator, long denominator) {
...
+        return r;
+    }
+
+    public static long multiplyExact(long num1, long num2) {
+        long r = num1 * num2;
+        if (num1 != 0 && r / num1 != num2) {
+            throw  ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
```
riyafa, line 86: if we are throwing the same error in multiple places we can reuse code by moving it to a new method

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
```diff
@@ -63,4 +63,28 @@ public static long remainder(long numerator, long denominator) {
             }
         }
     }
+
+    public static long addExact(long num1, long num2) {
+        long r = num1 + num2;
```
riyafa, line 68: can't we say `result` instead of `r`? `r` appears not to have a meaning 🙂

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/add-operation.bal
```diff
@@ -2,6 +2,12 @@ function intAdd(int a, int b) returns (int) {
     return a + b;
 }
 
+public function overflowByAddition() {
```
HindujaB, line 5: The function need not to be `public` since it's used in a single file. 
Shall we change these like the other functions?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/subtract-operation.bal
```diff
@@ -2,6 +2,12 @@ function intSubtract(int a, int b) returns (int) {
     return a - b;
 }
 
+public function overflowBySubtraction() {
+ int val = -9223372036854775808;
+ int val1 = 1;
+ int k = val1 - val;
+}
+
```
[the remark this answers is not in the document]
Nadeeshan96: I will look into this.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
Nadeeshan96: I will look into this.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
Nadeeshan96: I will look into this.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/add-operation.bal
Nadeeshan96: I will look into this.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
Nadeeshan96: As the if condition is not same as the above one, shall I not move this to a new function. I will move the same error throwing part into a new method.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
```diff
@@ -63,4 +67,28 @@ public static long remainder(long numerator, long denominator) {
             }
         }
     }
+
+    public static long addExact(long num1, long num2) {
+        long result = num1 + num2;
+        if (((num1 ^ result) & (num2 ^ result)) < 0) {
```
riyafa, line 73: What does this condition check for? Can we add a comment or refactor the expression to add more clarity? This is because future readers of the code should be able to understand it.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
```diff
@@ -63,4 +67,28 @@ public static long remainder(long numerator, long denominator) {
...
+            throwIntRangeOverflowError();
+        }
+        return result;
+    }
+
+    public static long subtractExact(long num1, long num2) {
+        long result = num1 - num2;
+        if (((num1 ^ num2) & (num1 ^ result)) < 0) {
```
riyafa, line 81: Shall we add some clarity? Same as above

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
```diff
@@ -63,4 +67,28 @@ public static long remainder(long numerator, long denominator) {
...
+            throwIntRangeOverflowError();
+        }
+        return result;
+    }
+
+    public static long multiplyExact(long num1, long num2) {
+        long result = num1 * num2;
+        if ((num1 != 0 && result / num1 != num2) || (num2 != 0 && result / num2 != num1)) {
```
riyafa, line 89: why did we decide not to use the `Math.multiplyExact`? Is it to reduce function calls?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
Nadeeshan96: I will add a comment in the code as follows
// an overflow has happened on addition if the sign of the result is different from BOTH operands

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
Nadeeshan96: I will add a comment in the code as follows above the if condition
// an overflow has happened on subtraction when both operands have the same sign AND
// if the sign of the result is different from the first operand

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
Nadeeshan96: I checked for performance using hyperfine using 5000 runs of a collection of multiplications, this one gave the best performance out of 3 methods : this one : Math.multiplyExact , and checking whether
if ((num1 != 0 && result / num1 != num2) || (num1==-1L && num2==-9223372036854775808L))

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
Nadeeshan96: When I moved throwing the error to a function
`private static void throwIntRangeOverflowError() {
        throw ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
    }`
although it worked in this implementation, the Codecov patch report says that the function call line was not covered by tests (but it is actually called to throw the error).
Now as going to `Math.addExact(), Math.subractExact(), Math.multiplyExact()` methods are better, when I write them using try catch blocks, it doesn't allow me to write 
`public static long subtractExact(long num1, long num2) {
        try {
            return Math.subtractExact(num1, num2);
        } catch (ArithmeticException e) {
            throwIntRangeOverflowError();
        }
    }`
I can neglect that by introducing another variable , But may I remove the function and use the statement in all the relevant places

warunalakshitha (commented):
LGTM

riyafa:
Shall we have commit messages in the imperative format: https://chris.beams.io/posts/git-commit/#imperative
In general refer the [article](https://chris.beams.io/posts/git-commit/) on general instructions on how to write good commit messages:

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
```diff
@@ -63,4 +63,28 @@ public static long remainder(long numerator, long denominator) {
             }
         }
     }
+
+    public static long addExact(long num1, long num2) {
+        try {
+            return Math.addExact(num1, num2);
+        } catch (ArithmeticException e) {
+            throw ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
```
riyafa, line 71: what I normally would do to avoid duplication is. Write a function
```
public BError getNumberOverFlowError() {
   return ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW, INT_RANGE_OVERFLOW_ERROR);
}
```
and throw the error in the catch statement
```
throw getNumberOverFlowError();
```

HindujaB (commented):
LGTM

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/MathUtils.java
Nadeeshan96: I will look into these.

Nadeeshan96: When I try to do the above _jballerina-debugger-integration-test_ fails.

warunalakshitha: ErrorUtils class we already use create internal errors. ex.createNumericConversionError. We can add the method there and call. 

throw ErrorUtils.createNumberOverFlowError(...)

Nadeeshan96: I will look into this.

warunalakshitha (approved):

PR #31042, opened 2021-06-08
Refactor Hoverutil to filter private class fields and methods

### LINKED ISSUE #30832: [LS] Hover documentation for classes from stdlibs reveal private fields
**Description:**
$subject

Consider
```
    resource function get getResource(http:Caller caller, http:Request req) {

    }
}
```
Now, hover over `http:Request` (method param) and the fields will be displayed in the documentation popup. It contains private fields like `isDirty`, etc. Need to hide them

**Steps to reproduce:**
See description

**Affected Versions:**
Beta1 RC5

malinthar:
## Purpose
$subject so that internal fields and methods are not shown

Fixes #30832 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/hover/HoverUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/hover/HoverUtil.java
@@ -15,12 +15,16 @@
  */
 package org.ballerinalang.langserver.hover;
 
+import io.ballerina.compiler.api.ModuleID;
 import io.ballerina.compiler.api.SemanticModel;
+import io.ballerina.compiler.api.symbols.ClassFieldSymbol;
 import io.ballerina.compiler.api.symbols.ClassSymbol;
 import io.ballerina.compiler.api.symbols.Documentable;
 import io.ballerina.compiler.api.symbols.Documentation;
 import io.ballerina.compiler.api.symbols.FunctionSymbol;
 import io.ballerina.compiler.api.symbols.MethodSymbol;
+import io.ballerina.compiler.api.symbols.ModuleSymbol;
+import io.ballerina.compiler.api.symbols.ObjectFieldSymbol;
 import io.ballerina.compiler.api.symbols.ObjectTypeSymbol;
 import io.ballerina.compiler.api.symbols.ParameterSymbol;
 import io.ballerina.compiler.api.symbols.Qualifier;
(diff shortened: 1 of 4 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/hover/HoverUtil.java
```diff
@@ -119,22 +120,27 @@ private static Hover getObjectHoverMarkupContent(Documentation documentation, Ob
             params.add(header(3, ContextConstants.FIELD_TITLE) + CommonUtil.MD_LINE_SEPARATOR);
 
             params.addAll(classSymbol.fieldDescriptors().entrySet().stream()
-                    .map(fieldEntry -> {
+                    .filter(fieldEntry -> paramsMap.containsKey(fieldEntry.getKey())
+                            && !fieldEntry.getValue().qualifiers().contains(Qualifier.PRIVATE)).map(fieldEntry -> {
```
IMS94, line 124: I think we should show only the `public` fields. Because there can be fields with no qualifier which accounts to package-private I guess (the spec doesn't say that explicitly: https://ballerina.io/ballerina-spec/spec.html#object-field @pubudu91 @nadeeshaan can you confirm?). We have to check that. Ideally we should package private entries when the opened file is from the same package. We can do that later.

IMS94 (commented):
LGTM! We have to clarify whether we are going to `not show private fields/methods` or show `only public members`.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/hover/HoverUtil.java
nadeeshaan: That depends on the context. According to the context, the fields' visibility is changed. But in this case, I'm +1 to go with only the public fields, in order to maintain the consistancy.

malinthar: +1. Let me check on this

malinthar: Done

nadeeshaan (approved):
LGTM

PR #31058, opened 2021-06-09
Add qualifier based filtering to field access completion resolver

### LINKED ISSUE #31032: [LS]  Avoid from suggesting private fields of classes
**Description:**
consider the following code segment.

```ballerina
function helloFunction(Person p1) returns string {
    return "Hello".concat(p1.<cursor>);
}


class Person {
    # Myvar
    private string name;

    function init(string name) {
        self.name = name;
    }

    function getName() returns string{
        return self.name;
    }
}
```
The private filed `name` of the above class is suggested on field access as depicted  below. 

**Steps to reproduce:**
Follow above steps

malinthar:
## Purpose
$subject to suggest only  visible methods, class fields and object fields based on qualifiers.

Fixes #31032 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/FieldAccessCompletionResolver.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/FieldAccessCompletionResolver.java
@@ -25,6 +25,7 @@ import io.ballerina.compiler.api.symbols.MapTypeSymbol;
 import io.ballerina.compiler.api.symbols.MethodSymbol;
 import io.ballerina.compiler.api.symbols.ModuleSymbol;
 import io.ballerina.compiler.api.symbols.ObjectTypeSymbol;
+import io.ballerina.compiler.api.symbols.Qualifiable;
 import io.ballerina.compiler.api.symbols.Qualifier;
 import io.ballerina.compiler.api.symbols.RecordFieldSymbol;
 import io.ballerina.compiler.api.symbols.RecordTypeSymbol;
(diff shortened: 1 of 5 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/FieldAccessCompletionResolver.java
```diff
@@ -267,24 +275,45 @@ private Symbol getSymbolByName(List<Symbol> visibleSymbols, String name, @Nonnul
...
-        boolean isResource = methodSymbol.qualifiers().contains(Qualifier.RESOURCE);
-        boolean isPrivate = methodSymbol.qualifiers().contains(Qualifier.PRIVATE);
-        boolean isPublic = methodSymbol.qualifiers().contains(Qualifier.PUBLIC);
+
+        boolean isPrivate = false;
+        boolean isPublic = false;
+        boolean isResource = false;
+
+        switch (symbol.kind()) {
+            case CLASS_FIELD:
```
nadeeshaan, line 291: Can't we simplify this? I think we can keep a list to maintain the qualifiers list and then do a ingle contains logic at the end.

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/FieldAccessCompletionResolver.java
malinthar: Here we are not able to use the contains method as we have to use extended logic to decide the visibility based on the context.

nadeeshaan (approved):
LGTM

PR #31078, opened 2021-06-10
Fix string iterator return type issue

### LINKED ISSUE #30735: string iterator should use string:Char
In slalpha5, this correct program

```
type StringIterator object {
    public isolated function next() returns record {|
        string:Char value;
    |}?;
};

function foo(string str) returns StringIterator {
  return str.iterator();
}
```

gets an incompatible types error. Changing `string:Char` to `string` avoids the error.

SandaruJayawardana:
Fixes #30735

## Check List 

- [ ] changed necessary tests & libraries


langlib/langlib-test/src/test/resources/test-src/stringlib_test.bal
```diff
@@ -35,14 +35,14 @@ function testIterator() returns string[] {
     string str = "Foo Bar";
 
     object {
-         public isolated function next() returns record {| string value; |}?;
+         public isolated function next() returns record {| string:Char value; |}?;
```
gimantha, line 38: Shall we also add a new test with string(not Char) also? It should work as we return Char right?

MaryamZi (commented):
Shall we add some tests for the foreach statement?

The following should work now, right?

```ballerina
string foo = "foo";

foreach string:Char item in foo {
   // do something here and assert it after the foreach statement    
}
```

This is true already, but we can maybe add a simple test.
```ballerina
string foo = "foo";

foreach var item in foo {
    string str = item;
    assertTrue(str is string:Char);
}
```

langlib/langlib-test/src/test/resources/test-src/stringlib_test.bal
SandaruJayawardana: Okay. I'll add.

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibStringTest.java
```diff
@@ -75,13 +75,18 @@ public void testSubString() {
     @Test
     public void testIterator() {
         String[] expected = new String[]{"F", "o", "o", " ", "B", "a", "r"};
-        BValue[] returns = BRunUtil.invoke(compileResult, "testIterator");
-        BValueArray arr = (BValueArray) returns[0];
-        long size = arr.size();
 
-        for (int i = 0; i < size; i++) {
-            assertEquals(arr.getString(i), expected[i]);
+        for (int test = 0; test < 2; test++) {
```
gimantha, line 79: Can you do the assertion in ballerina file itself?, for examples refer` tablelib_test.bal`

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/statements/foreach/ForeachStringTypedBindingPatternsTests.java
```diff
@@ -25,12 +26,14 @@
...
 
+import static org.testng.Assert.assertEquals;
+
 /**
  * Tests for typed binding patterns in foreach.
  *
  * @since 0.985.0
  */
-@Test(enabled = false)
+@Test(enabled = true)
```
gimantha, line 36: default value of `enabled` is true, so you can remove the parameter, without explicitly making them true

langlib/langlib-test/src/test/resources/test-src/statements/foreach/foreach-string-typed-binding-patterns.bal
```diff
@@ -10,6 +10,19 @@ function concatIntAny(int i, any a) {
 
 // ---------------------------------------------------------------------------------------------------------------------
 
+function testStringWithSimpleVariableWithCharType() returns string {
+    output = "";
+
+    string sdata = "Ballerina";
+
+    int i = 0;
+    foreach string:Char s in sdata {
```
gimantha, line 19: Shall also add a test case where the foreach var type is string as well?

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibStringTest.java
```diff
@@ -74,14 +74,13 @@ public void testSubString() {
...
     @Test
     public void testIterator() {
-        String[] expected = new String[]{"F", "o", "o", " ", "B", "a", "r"};
-        BValue[] returns = BRunUtil.invoke(compileResult, "testIterator");
-        BValueArray arr = (BValueArray) returns[0];
-        long size = arr.size();
 
-        for (int i = 0; i < size; i++) {
-            assertEquals(arr.getString(i), expected[i]);
+        for (int test = 0; test < 2; test++) {
```
gimantha, line 78: move this logic to ballerina as well

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibStringTest.java
```diff
@@ -74,14 +74,13 @@ public void testSubString() {
...
+
+            BRunUtil.invoke(compileResult, "testIterator", new BValue[]{new BInteger(test)});
+
         }
+
```
gimantha, line 83: Remove extra newline

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibStringTest.java
```diff
@@ -74,14 +74,13 @@ public void testSubString() {
...
-        BValue[] returns = BRunUtil.invoke(compileResult, "testIterator");
-        BValueArray arr = (BValueArray) returns[0];
-        long size = arr.size();
 
-        for (int i = 0; i < size; i++) {
-            assertEquals(arr.getString(i), expected[i]);
+        for (int test = 0; test < 2; test++) {
+
+            BRunUtil.invoke(compileResult, "testIterator", new BValue[]{new BInteger(test)});
+
```
gimantha, line 81: Remove extra new line

langlib/langlib-test/src/test/resources/test-src/statements/foreach/foreach-string-typed-binding-patterns.bal
```diff
@@ -81,3 +109,34 @@ function testIterationOnEmptyString() returns string {
...
+function assertEquals(anydata expected, anydata actual) {
+    if (expected == actual) {
+        return;
+    }
+    typedesc<anydata> expT = typeof expected;
+    typedesc<anydata> actT = typeof actual;
+    string msg = "expected [" + expected.toString() + "] of type [" + expT.toString()
+                            + "], but found [" + actual.toString() + "] of type [" + actT.toString() + "]";
+    panic error(ASSERTION_ERROR_REASON, message = msg);
+}
```
gimantha, line 142: Add a newline here

gimantha (dismissed):
LGTM

langlib/langlib-test/src/test/resources/test-src/stringlib_test.bal
```diff
@@ -31,24 +31,39 @@ function testSubString() returns [string,string, string] {
...
-    while (elem is record {| string value; |}) {
-        chars[i] = elem.value;
-        elem = itr.next();
-        i += 1;
+    string:Char[] expected = ["F", "o", "o", " ", "B", "a", "r"];
+    int i;
+
+    foreach int test in 1 ..< 3 {
+        i = 0;
+        match test{
```
MaryamZi, line 41:
```suggestion
        match test {
```

langlib/langlib-test/src/test/resources/test-src/stringlib_test.bal
```diff
@@ -31,24 +31,39 @@ function testSubString() returns [string,string, string] {
...
+                    } itr = str.iterator();
+                record {| string value; |}|() elem = itr.next();
+                while (elem is record {| string:Char value; |}) {
+                    assertEquals(expected[i], elem.value);
+                    elem = itr.next();
+                    i+= 1;
+                }
+            }
+        }
     }
```
MaryamZi, line 66: Why do we need a foreach statement here. Can't we just do 

```ballerina
    object {
        public isolated function next() returns record {| string:Char value; |}?;
    } itr1 = str.iterator();
    record {| string:Char value; |}|() elem1 = itr1.next();
    while (elem1 is record {| string:Char value; |}) {
        assertEquals(expected[i], elem1.value);
        elem1 = itr1.next();
        i += 1;
    }
    assertEquals(7, i);

    i = 0;
    object {
        public isolated function next() returns record {| string value; |}?;
    } itr2 = str.iterator();
    record {| string value; |}|() elem2 = itr2.next();
    while (elem2 is record {| string:Char value; |}) {
        assertEquals(expected[i], elem2.value);
        elem2 = itr2.next();
        i += 1;
    }
    assertEquals(7, i);
```

We should also assert `i` to verify that the expected number of iterations happen. If there's eventually a change that breaks the type of the `value` field in the returned record to `string`, the current test wouldn't catch it since it'll simply not go into the while loops.

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/statements/foreach/ForeachStringTypedBindingPatternsTests.java
```diff
@@ -30,7 +30,7 @@
  *
  * @since 0.985.0
  */
-@Test(enabled = false)
+@Test
 public class ForeachStringTypedBindingPatternsTests {
```
MaryamZi, line 34: Let's also add a negative test in org.ballerinalang.langlib.test.statements.foreach.ForeachNegativeTests#testForeachVarTypeNegative.

For something like
```ballerina
    string str = "foo";

    foreach int s in str {
        
    }
```

we should now get

```
incompatible types: expected 'string:Char', found 'int'
```

gimantha (approved):

PR #31122, opened 2021-06-11
Infer type for int/float literals in binary expressions

### LINKED ISSUE #13904: [Spec Deviation] Incorrect type inferring for int-literals/floating-point-literals used in equality expressions
**Description:**

The spec defines the types of int-literals and floating-point-literals to be the contextually expected types.

```
The type of an int-literal depends on the contextually expected numeric type, where 
the contextually expected numeric type is the intersection of the contextually 
expected type with int|float|decimal:
● if the contextually expected numeric type is a subtype of float, then the type of the
int-literal is a float
● if the contextually expected numeric type is a subtype of decimal, then the type of 
the int-literal is decimal
● otherwise, the broad type of the int-literal is int, and the narrow type is a 
singleton int type

Similarly the type of a floating-point-literal is determined as follows:
● if the contextually expected numeric type is a subtype of decimal, the type of the
floating-point-literal is a decimal,
● otherwise, the type of the floating-point literal is float
```

Thus even in the equality expression, the type of the int-literal/floating-point-literal used as an operand would depend on the type of the other operand.

The following is the expected behaviour:
```ballerina
    float f = 1.0;
    _ = f == 2; // 2 should be considered float 2.0

    decimal d = 2.0;
    _ = d == 2; // 2 should be considered decimal 2.0

    int|float|decimal ifd = 1;
    _ = ifd == 1; // already allowed, 1 would be considered int 1, evals to true

    ifd = f; // f is float 1.0
    _ = ifd == 1; // evals to false, since 1 is considered int 1, and ifd is float 1.0
}
```

**Affected Versions:**
0.990.3

### LINKED ISSUE #31305: Infer correct type for int/float numerical literals
**Description:**
Following expressions should be supported 
`float a1 = 10 + 3.0 + 5.0; `
`float a2 = 5 + 3 + 10;`
`decimal a3 = 10.0 + 3 ;`
`decimal a4 = 10 + 5 + 3.0;`

**Describe your problem(s)**

**Describe your solution(s)**

rpjayasekara:
Fixes #31305

`float f = 2.0 + 2;`
`decimal d = 2.0 + 2`
`float a = 3 + 5;`

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -4012,15 +4013,56 @@ public void visit(BLangBinaryExpr binaryExpr) {
         checkDecimalCompatibilityForBinaryArithmeticOverLiteralValues(binaryExpr);
 
         SymbolEnv rhsExprEnv;
-        BType lhsType = checkExpr(binaryExpr.lhsExpr, env);
+        BType lhsType;
+        if (binaryExpr.lhsExpr instanceof BLangNumericLiteral) {
```
MaryamZi, line 4017: We can use the node kind instead.

```suggestion
        if (binaryExpr.lhsExpr.getKind() == NodeKind.NUMERIC_LITERAL) {
```

We usually try to avoid expensive instanceof checks.

MaryamZi, line 4017: Shall we extract `binaryExpr.lhsExpr` out to a variable since it is accessed in multiple places?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -4012,15 +4013,56 @@ public void visit(BLangBinaryExpr binaryExpr) {
...
+                    } else {
+                        rhsType = checkExpectedTypeCompatibility(binaryExpr.rhsExpr, lhsType, rhsExprEnv);
+                    }
+                    break;
+                default:
+                    rhsType = checkExpr(binaryExpr.rhsExpr, rhsExprEnv);
+            }
+        } else {
+            rhsType = checkExpr(binaryExpr.rhsExpr, rhsExprEnv);
+        }
```
MaryamZi, line 4065: Isn't this similar to L4017-L4035? Shall we check if we can extract this logic out to a method?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/subtract-operation.bal
```diff
@@ -69,6 +69,18 @@ function testSubtractSingleton() {
     assertEqual(a1 - a7, 8);
 }
 
+function testContextuallyExpectedTypeOfNumericLiteralInSubtract(){
```
MaryamZi, line 72:
```suggestion
function testContextuallyExpectedTypeOfNumericLiteralInSubtract() {
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
rpjayasekara: I think it can, I will extract the logic out to a method

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -4078,6 +4080,46 @@ public void visit(BLangBinaryExpr binaryExpr) {
         resultType = types.checkType(binaryExpr, actualType, expType);
     }
 
+    private BType checkAndGetType(BLangExpression expr, SymbolEnv env, BLangBinaryExpr binaryExpr) {
```
rdhananjaya, line 4083: 1. This function looks too nested. Shall we try to early exit when possible.
2. This function only seems to propagate the contextually expected type one level down. We need to support `float x = 1.0 + 2 + 3` too.

MaryamZi:
@rpjayasekara, as discussed, let's create a new issue for what we are fixing here and link that in the PR.

rdhananjaya:
@rpjayasekara can you please reword commit messages to be within 50 characters

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/NodeCloner.java
```diff
@@ -2286,4 +2312,12 @@ public void visit(BLangObjectConstructorExpression source) {
 
         source.cloneRef = clone;
     }
+
+    public int getCloneAttempt() {
```
rdhananjaya, line 2316: I think we can remove these methods now (both get and set)

rpjayasekara: Will remove those

rdhananjaya (approved):

PR #31151, opened 2021-06-14
Fix JSON lax typing issue

### LINKED ISSUE #28360: Lax typing rules don't seem to work for checking-expr with a json variable
**Description:**
$title.

**Steps to reproduce:**
Shouldn't the following be allowed according to https://github.com/ballerina-platform/ballerina-spec/issues/487#issuecomment-671652259?
```ballerina
    json j = 1;
    int i = check j;
}
```

```cmd
$ bal run xyz.bal 

Compiling source
        xyz.bal
ERROR [xyz.bal:(3:19,3:20)] incompatible types: expected '(int|error)', found 'json'
```

**Affected Versions:**
slalpha1

### LINKED ISSUE #30901: Converting json to simple types doesn't work as expected
**Description:**
$title

**Steps to reproduce:**
```ballerina
json j = { 
   x: { 
      y: { 
         z: "value" 
      }
   }
};

   string s = check j.x.y.z;
}
```
Above results in a compilation error
`incompatible types: expected '(string|error)', found '(json|error)'`

**Related Issues:**
https://github.com/ballerina-platform/ballerina-lang/issues/28360

**Affected Versions:**
slbeta1rc6

gimantha:
Fixes #28360
Fixes #30901

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/types/jsontype/json-value.bal
```diff
@@ -432,3 +432,59 @@ function testMapJsonToJsonArray() {
...
+function assertEquals(anydata expected, anydata actual) {
+    if (expected == actual) {
+        return;
+    }
+    typedesc<anydata> expT = typeof expected;
+    typedesc<anydata> actT = typeof actual;
+    string msg = "expected [" + expected.toString() + "] of type [" + expT.toString()
+                            + "], but found [" + actual.toString() + "] of type [" + actT.toString() + "]";
+    panic error(ASSERTION_ERROR_REASON, message = msg);
+}
```
rdhananjaya, line 490:
```suggestion
}

```

rdhananjaya:
Shall we add some negative test cases, since we are allowing `insureType` with non field access syntax that we didn't allowed before.

tests/jballerina-unit-test/src/test/resources/test-src/types/jsontype/json-value.bal
```diff
@@ -432,3 +432,59 @@ function testMapJsonToJsonArray() {
...
+          json j = {
+             x: {
+                y: {
+                   z: "value"
+                }
+             }
+          };
+
+         string val =  check j.x.y.z;
+         assertEquals("value", val);
```
lasinicl, line 458:
```suggestion
    json j = {
        x: {
            y: {
                z: "value"
            }
        }
    };
    
    string val =  check j.x.y.z;
    assertEquals("value", val);
```

tests/jballerina-unit-test/src/test/resources/test-src/types/jsontype/json-value.bal
gimantha: seems i cannot commit this suggestion, i will do the change

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2519,7 +2519,7 @@ public void visit(BLangFieldBasedAccess fieldAccessExpr) {
                     fieldAccessExpr.isCompoundAssignmentLValue;
         }
 
-        BType varRefType = types.getTypeWithEffectiveIntersectionTypes(getTypeOfExprInFieldAccess(containerExpression));
+        BType varRefType = types.getTypeWithEffectiveIntersectionTypes(checkExpr(containerExpression, env));
```
MaryamZi, line 2522: We can get rid of the `getTypeOfExprInFieldAccess` method with this change, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5319,6 +5304,26 @@ private void visitCheckAndCheckPanicExpr(BLangCheckedExpr checkedExpr) {
         resultType = types.checkType(checkedExpr, actualType, expType);
     }
 
+    private void rewriteWithEnsureTypeFunc(BLangCheckedExpr checkedExpr, BLangExpression exprWithCheckingKeyword,
+                                           BType typeOfExprWithCheckingKeyword) {
+        BType exprType = checkExpr(exprWithCheckingKeyword, env);
+        BType errorLiftedType = types.getSafeType(exprType, false, true);
+        if (types.isLax(errorLiftedType)) {
```
MaryamZi, line 5311: Shall we return early if `!types.isLax(errorLiftedType)`.

tests/jballerina-unit-test/src/test/resources/test-src/types/jsontype/json-value.bal
```diff
@@ -432,3 +432,59 @@ function testMapJsonToJsonArray() {
     json[] j = [2];
     map<json>[] p = <map<json>[]> j;
 }
+
+function testJsonLaxErrorLifting() returns error? {
+    check testSimpleVarRefJsonErrorLiftingInLaxTyping();
+    check testJsonMapErrorLifingInLaxTyping();
+    check testJsonReturnTypeErrorLifting();
+}
```
MaryamZi, line 440: Shall we break this into three different tests and call them separately from the Java side? 

Also since this may return error, shouldn't we have asserted something on the Java side anyway? Instead we could use checkpanic here, to make the test fail if the function returns error.

gimantha: Shouldn't we assert in Ballerina code itself?

MaryamZi: We should, but since `check` just returns the error and we are not asserting anything on the Java side, this test wouldn't fail if any of these functions returned error, right?

But if we do something like
```ballerina
function testJsonLaxAccess1() {
    checkpanic testSimpleVarRefJsonErrorLiftingInLaxTyping();
}

function testJsonLaxAccess2() {
    checkpanic testJsonMapErrorLifingInLaxTyping();
}

function testJsonLaxAccess3() {
    checkpanic testJsonReturnTypeErrorLifting();
}
```

for the positive cases (i.e., where they shouldn't return error), and call these `testJsonLaxAccessN` functions from Java, if the function incorrectly returns error the Java test will fail.

Or we should do
```ballerina
function testJsonLaxAccess1() {
    error? res = testSimpleVarRefJsonErrorLiftingInLaxTyping();
    assertTrue(res is ());
}
```

For the runtime negative cases (which I believe we haven't added yet), we have to do this anyway, and assert that `res is error`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5256,22 +5256,21 @@ private void visitCheckAndCheckPanicExpr(BLangCheckedExpr checkedExpr) {
...
-                typedescExpr.resolvedType = expType;
-                typedescExpr.setBType(typedescType);
-                argExprs.add(typedescExpr);
-                BLangInvocation invocation = ASTBuilderUtil.createLangLibInvocationNode(FUNCTION_NAME_ENSURE_TYPE,
-                                                                    argExprs, exprWithCheckingKeyword, checkedExpr.pos);
-                invocation.symbol = symResolver.lookupLangLibMethod(typeOfExprWithCheckingKeyword,
-                                                                    names.fromString(invocation.name.value));
-                invocation.pkgAlias = (BLangIdentifier) TreeBuilder.createIdentifierNode();
-                checkedExpr.expr = invocation;
+        if (checkedExpr.getKind() == NodeKind.CHECK_EXPR && types.isUnionOfSimpleBasicTypes(expType)) {
```
gimantha, line 5259: This switch block can be removed once https://github.com/ballerina-platform/ballerina-lang/pull/31189 merged.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5326,6 +5311,56 @@ private void visitCheckAndCheckPanicExpr(BLangCheckedExpr checkedExpr) {
         resultType = types.checkType(checkedExpr, actualType, expType);
     }
 
+    private void rewriteWithEnsureTypeFunc(BLangCheckedExpr checkedExpr, BLangExpression expr,
+                                           BType typeOfExprWithCheckingKeyword) {
+        BType rhsType = getCandidateType(checkedExpr);
+        if (rhsType == symTable.semanticError) {
+            rhsType = checkExpr(checkedExpr.expr, env);
+        }
+        BType errorLiftedType = types.getSafeType(rhsType, false, true);
```
MaryamZi, line 5320: `Types.getSafeType` doesn't seem to handle all cases. Right now, for error lifting it only removes the [builtin error](https://github.com/ballerina-platform/ballerina-lang/blob/e9d7c054e4561ba41890e00bbe5240552f60664d/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java#L4847). So the following doesn't work. Instead we should check by tag and remove.

```ballerina
public function main() returns Error? {
    () x = check func();
}

function func() returns json|Error {

}

type Error distinct error;
```

Also, although probably has no impact on this path, we don't seem to be checking if nil should be lifted before returning types without nil in https://github.com/ballerina-platform/ballerina-lang/blob/e9d7c054e4561ba41890e00bbe5240552f60664d/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java#L4823.

tests/jballerina-unit-test/src/test/resources/test-src/types/jsontype/json-value-negative.bal
```diff
@@ -24,6 +24,11 @@ public function testCloseRecordToMapJsonAssigment() returns [map<json>, map<json
     return [pm, m];
 }
 
+public function testLaxAccessJSONArray() {
+    json x = [23, 45, 21, 12];
+    int [] y = check x;
```
MaryamZi, line 29:
```suggestion
    int[] y = check x;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Had a discussion with @gimantha and this should be a compilation error because `func()` is not lax.

MaryamZi: Let's add a test for this too.

gimantha: added

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -4844,7 +4849,9 @@ public BType getSafeType(BType type, boolean liftNil, boolean liftError) {
         }
 
         if (liftError) {
-            errorLiftedType.remove(symTable.errorType);
+            memTypes = errorLiftedType.getMemberTypes().stream().filter(t -> t.tag != TypeTags.ERROR)
+                    .collect(Collectors.toCollection(LinkedHashSet::new));
```
MaryamZi, line 4853: We usually try to avoid using streams for trivial logic, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -4844,7 +4849,9 @@ public BType getSafeType(BType type, boolean liftNil, boolean liftError) {
...
+            errorLiftedType = BUnionType.create(null, memTypes);
```
MaryamZi, line 4854: We need to check the size of `memTypes` before creating a union type. We shouldn't be creating a union type when there are 0 or 1 members. 0 should probably be neverType or semanticError.

tests/jballerina-unit-test/src/test/resources/test-src/types/jsontype/json-value.bal
```diff
@@ -432,3 +432,88 @@ function testMapJsonToJsonArray() {
...
+    error? e1 = testSimpleVarRefJsonErrorLiftingInLaxTyping();
+    assertTrue(e1 is ());
+    error? e2 = testJsonMapErrorLifingInLaxTyping();
+    assertTrue(e2 is ());
+    error? e3 = testJsonReturnTypeErrorLifting();
+    assertTrue(e3 is ());
+    error? e4 = testLaxAccessForRuntimeErrors1();
+    assertTrue(e4 is error);
+    assertEquals((<error> e4).detail().toString(),
+        "{\"message\":\"incompatible types: 'string' cannot be cast to 'int'\"}");
```
MaryamZi, line 446: Shall we assert the message itself instead?
```suggestion
    assertEquals(<string> checkpanic (<error> e4).detail()["message"],
        "incompatible types: 'string' cannot be cast to 'int'");
```

tests/jballerina-unit-test/src/test/resources/test-src/types/jsontype/json-value.bal
```diff
@@ -432,3 +432,88 @@ function testMapJsonToJsonArray() {
...
+    error? e5 = testLaxAccessForRuntimeErrors2();
+    assertTrue(e5 is error);
+    assertEquals((<error> e5).detail().toString(),
+        "{\"message\":\"incompatible types: 'int' cannot be cast to 'string'\"}");
```
MaryamZi, line 450:
```suggestion
    assertEquals(<string> checkpanic (<error> e5).detail()["message"],
        "incompatible types: 'int' cannot be cast to 'string'";
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5326,6 +5311,55 @@ private void visitCheckAndCheckPanicExpr(BLangCheckedExpr checkedExpr) {
         resultType = types.checkType(checkedExpr, actualType, expType);
     }
 
+    private void rewriteWithEnsureTypeFunc(BLangCheckedExpr checkedExpr, BType type) {
+        BType rhsType = getCandidateType(checkedExpr, type);
+        if (rhsType == symTable.semanticError) {
+            rhsType = checkExpr(checkedExpr.expr, env);
+        }
```
MaryamZi, line 5318: Shouldn't we add an else block and call `checkExpr(checkedExpr.expr, env, type)`? We should type check the actual expr, since we type check a clone in `getCandidateType` right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
gimantha: @MaryamZi in the next few lines, we are checking if the union type has one type or not and then return only that. We might need to return neverType if 0. That logic is not there.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
gimantha: that type checking logic is there in the line number 5238

MaryamZi: Ah okay. But then, we should use a clone in L5346 also, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -4882,12 +4887,24 @@ public BType getSafeType(BType type, boolean liftNil, boolean liftError) {
...
+            memTypes = bTypes;
+            errorLiftedType = BUnionType.create(null, memTypes);
         }
 
         if (errorLiftedType.getMemberTypes().size() == 1) {
             return errorLiftedType.getMemberTypes().toArray(new BType[0])[0];
         }
+
+        if (errorLiftedType.getMemberTypes().size() == 0) {
+            return symTable.neverType;
```
MaryamZi, line 4905: For now, shall we return semanticError? I don't think we've properly introduced type checking with never yet.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
MaryamZi: Ah, missed that. But we can probably avoid creating the union type unnecessarily, right? We can do this separately though.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
gimantha: sure

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
gimantha: No right? If we do a clone, then original `checkExpr.expr` will be type checked on the normal path and it will report type incompatible error.

gimantha: ah yes. I got your point :)

MaryamZi (approved):

MaryamZi:
We should add info on this to the beta3 release note.

PR #31165, opened 2021-06-14
Deprecate `type()` and introduce `typeOf()` in its place

pubudu91:
## Purpose
The old `type()` method was buggy. Although it was meant to give the type of expressions, it could be used to get the type of other constructs as well (e.g., variables, type nodes etc.). Therefore, this PR introduces a couple of changes to address that and another issue where expressions' types were set to compile error if the overall statement was deemed to be semantically incorrect. 

## Approach
- Added a new field to `BLangNode` called `determinedType`. This field holds the actual type determined for the expression, irrespective of the final result of the type checking. e.g.,
```ballerina
int x = foo(); // foo() returns a string
```
Previously, the type for `foo()` would be returned as compile error since the overall statement is semantically incorrect. But in reality, the expression `foo()` has a determinable type, independent of the context. The newly introduced `typeOf()` method returns this actual determined type instead of the final type.

- Deprecated `type()` and introduced `typeOf()` instead due to the above mentioned reasons and since changing the behaviour of `type()` would potentially cause major impact in downstream projects.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -195,7 +195,9 @@ public class Types {
                            BType actualType,
                            BType expType,
                            DiagnosticCode diagCode) {
-        expr.setBType(checkType(expr.pos, actualType, expType, diagCode));
+        expr.setDeterminedType(actualType);
+        expr.setTypeCheckedType(checkType(expr.pos, actualType, expType, diagCode));
+
         if (expr.getBType().tag == TypeTags.SEMANTIC_ERROR) {
             return expr.getBType();
         }
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/SemanticModel.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/SemanticModel.java
@@ -131,7 +131,9 @@ public interface SemanticModel {
      *
      * @param range the text range of the expression
      * @return the type of the expression
+     * @deprecated This method will be removed in a later version. Use typeOf() instead.
      */
+    @Deprecated
     Optional<TypeSymbol> type(LineRange range);
 
     /**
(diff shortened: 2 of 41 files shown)


IMS94 (approved):
LGTM!

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -195,7 +195,9 @@ public BType checkType(BLangExpression expr,
                            BType actualType,
                            BType expType,
                            DiagnosticCode diagCode) {
-        expr.setBType(checkType(expr.pos, actualType, expType, diagCode));
+        expr.setDeterminedType(actualType);
+        expr.setTypeCheckedType(checkType(expr.pos, actualType, expType, diagCode));
```
rdhananjaya, line 199: Are there places that we set types that is not `symtable.semanticError` when something is not right with the type-checking of that statement/expression.
If it's always `semanticError` we can merge this 2 methods.

pubudu91: Yes. Supporting that scenario is the reason for this separation. For example, consider a function `foo()` which returns a string. And now consider the following:
```ballerina
int x = foo();
```
Here, since the LHS and RHS are incompatible, we set `semanticError` as the type of the function call expr `foo()`. But in reality, we can (and we do) determine a type correctly for `foo()`. Being able to get this type through the semantic API was the original requirement. But since we overwrite that type to `semanticError` due to the type incompatibility, previously this info was lost.

rdhananjaya: Oh I was thinking about checking for semanticError, but I don't think we have access to that here.
`this.determinedType != symtable.semanticError`

rdhananjaya: So I'm fine with this change

rdhananjaya (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
pubudu91: In this particular instance, yes, `this.determinedType != symtable.semanticError`. But there can be instances where the determined type is `semanticError`. For example, I think the following binary expr would be one such instance: `10 + 2.3`

PR #30975, opened 2021-06-03
For check expression, take error from the expectedType when possible

rdhananjaya:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30737

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #31189, opened 2021-06-15
Revert "For check expression, take error from the expectedType when possible"

rdhananjaya:
Reverts ballerina-platform/ballerina-lang#30975


MaryamZi (approved):
We're reverting this for now since the following does not work with these changes.

```ballerina
type Detail record {|
    int code;
|};

type E1 distinct error<Detail>;

function foo() returns int|E1|error => 0;

function bar() returns error? {
    int m = check foo();
}
```

PR #31202, opened 2021-06-16
Improve query expression intermediate clause recovery

### LINKED ISSUE #30612: Select clause is not suggested when there is more statements after query expression
Consider the bellow example
```bal
public function main() {
    int[*] sourceArray = [1, 2, 3, 4, 5, 6, 7];
    int[] selectedPersons = from var i in sourceArray
                            se<cursor>
    io:println(selectedPersons);

}
```
The select clause is not suggested when there are more statements after query expression. It shows select clause when there is no other statement follows it

### LINKED ISSUE #31103: Query Expression's Intermidiate-clause is not  recovered correctly
**Description:**
Consider the following source code.

```ballerina
public function main() {
    models:PET pet1 = new (1000,2);
    models:PET pet2 = new (2000,1);
    models:PET pet3 = new (900,6);

    models:PET[] allpets = [pet1,pet2,pet3];

    models:PET[] listResult = from var pet in allpets
        where pet.value > 1000 
        w<cursor>
        select pet;
}
```
Following is the syntax tree for the above source code.

Note that the `w` token is considered a part of the `binary expression`. In this scenario if the user's intention is to add another `intermediate-clause` we should suggest keywords such as `where`, `join`.  However, due to the above behavior the resolved context is `intermediate-clause` rather than `query-expression` context. Therefore LS can't provide accurate suggestions in this scenario. 
Therefore, a recovery solution is required to accommodate  the above requirement. 

**Steps to reproduce:**
Follow above steps

lochana-chathura:
## Purpose
$subject as per the LS requirement.

Fixes #31103
Fixes #30612 

## Approach
N/A

## Samples
Old recovery:
```java
models:PET[] listResult = from var pet in allpets
    where pet.value > 1000 MISSING[<] w//<cursor>
    select pet;
```
New recovery:
```java
models:PET[] listResult = from var pet in allpets
    where pet.value > 1000 MISSING[where] w//<cursor>
    select pet;
```

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/expressions/QueryExpressionTest.java
+++ b/compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/expressions/QueryExpressionTest.java
@@ -241,4 +241,9 @@ public class QueryExpressionTest extends AbstractExpressionsTest {
         testFile("query-expr/query_expr_source_65.bal", "query-expr/query_expr_assert_65.json");
         testFile("query-expr/query_expr_source_66.bal", "query-expr/query_expr_assert_66.json");
     }
+
+    @Test
+    public void testIntermediateClauseStartRecovery() {
+        testFile("query-expr/query_expr_source_68.bal", "query-expr/query_expr_assert_68.json");
+    }
 }
--- a/compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/expressions/QueryExpressionTest.java
+++ b/compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/expressions/QueryExpressionTest.java
@@ -246,4 +246,9 @@ public class QueryExpressionTest extends AbstractExpressionsTest {
     public void testIntermediateClauseStartRecovery() {
         testFile("query-expr/query_expr_source_68.bal", "query-expr/query_expr_assert_68.json");
     }
+
+    @Test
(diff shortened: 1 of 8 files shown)


compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/expressions/QueryExpressionTest.java
```diff
@@ -241,4 +241,9 @@ public void testIncompleteClauses() {
         testFile("query-expr/query_expr_source_65.bal", "query-expr/query_expr_assert_65.json");
         testFile("query-expr/query_expr_source_66.bal", "query-expr/query_expr_assert_66.json");
     }
+
+    @Test
+    public void testIntermediateClauseStartRecovery() {
+        testFile("query-expr/query_expr_source_68.bal", "query-expr/query_expr_assert_68.json");
```
rdulmina, line 247: `testFile` starts context with `TOP_LEVEL_NODE`. Let's add the codes inside a function

lochana-chathura: oops. thanks!

rdulmina (approved):
Parser changes LGTM

PR #31210, opened 2021-06-16
Add completion support for start actions

### LINKED ISSUE #30519: [LS] Completions for `self.<cursor>` is not shown for start action within a method inside a class
**Description:**
$title

Consider the following code:
```
public class WordCounter {

    public function countInText(string text) {
        string[] lines = regex:split(text, "\n");        
        future<map<int>> f = start self.<cursor>
    }

    function countWordsInLine(string line) returns map<int> {

    }
}
```

Consider the above code and the cursor is at `<cursor>`. At that position, completions for other methods inside the class are not shown.

**Steps to reproduce:**
See description

**Affected Versions:**
Beta 1 RC2

Language Server

TharushiJay:
## Purpose
$subject

Screenshot from 2021-06-16 16-42-53

Fixes #30519 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MethodCallExpressionNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MethodCallExpressionNodeContext.java
@@ -18,6 +18,7 @@ package org.ballerinalang.langserver.completions.providers.context;
 import io.ballerina.compiler.syntax.tree.ExpressionNode;
 import io.ballerina.compiler.syntax.tree.MethodCallExpressionNode;
 import io.ballerina.compiler.syntax.tree.NameReferenceNode;
+import io.ballerina.compiler.syntax.tree.Token;
 import org.ballerinalang.annotation.JavaSPIService;
 import org.ballerinalang.langserver.commons.BallerinaCompletionContext;
 import org.ballerinalang.langserver.commons.completion.LSCompletionException;
@@ -57,10 +58,13 @@ public class MethodCallExpressionNodeContext extends FieldAccessContext<MethodCa
         hierarchy and skip properly
         eg:
         (2) s<cursor>abc.def.testMethod()
+        (3) self.<cursor>Method()
          */
         int cursor = context.getCursorPositionInTree();
         NameReferenceNode nameRef = node.methodName();
+        Token dotToken = node.dotToken();
 
-        return cursor >= nameRef.textRange().startOffset() && cursor <= nameRef.textRange().endOffset();
+        return ((cursor >= nameRef.textRange().startOffset() && cursor <= nameRef.textRange().endOffset())
+                || (!dotToken.isMissing() && cursor > dotToken.textRange().startOffset()));
     }
 }
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/StartActionNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/StartActionNodeContext.java
@@ -32,7 +32,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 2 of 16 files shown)


IMS94 (dismissed):
Awesome set of tests @TharushiJay !!! :clap: LGTM!

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MethodCallExpressionNodeContext.java
```diff
@@ -57,10 +58,13 @@ public boolean onPreValidation(BallerinaCompletionContext context, MethodCallExp
...
         (2) s<cursor>abc.def.testMethod()
+        (3) self.<cursor>Method()
          */
         int cursor = context.getCursorPositionInTree();
         NameReferenceNode nameRef = node.methodName();
+        Token dotToken = node.dotToken();
 
-        return cursor >= nameRef.textRange().startOffset() && cursor <= nameRef.textRange().endOffset();
+        return ((cursor >= nameRef.textRange().startOffset() && cursor <= nameRef.textRange().endOffset())
+                || (!dotToken.isMissing() && cursor > dotToken.textRange().startOffset()));
```
nadeeshaan, line 68: What's the sample we are going to cover with this additional check?

TharushiJay: @nadeeshaan The sample would be `start self.<cursor>` . It failed on `onPreValidation()` in MethodCallExpression, and then picked up the provider incorrectly as `StartActionNodeContext` . Added this check to avoid that.

nadeeshaan: I was wondering whether this should direct to `FieldAccessExpressionNodeContext`. Can you add a comment in this place?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/StartActionNodeContext.java
```diff
@@ -0,0 +1,102 @@
...
+        Supports the following
+        (1) future<int> f = start <cursor>
+        (2) future<int> f = start a<cursor>
+         */
+        List<LSCompletionItem> completionItems = new ArrayList<>(this.getTypeDescContextItems(context));
+        
+        List<Symbol> visibleSymbols = context.visibleSymbols(context.getCursorPosition());
+        List<Symbol> filteredList = visibleSymbols.stream()
+                .filter(symbol -> (symbol instanceof VariableSymbol || symbol.kind() == FUNCTION
+                        || symbol.kind() == CLASS))
```
nadeeshaan, line 87: Why do we filter classes here? Class symbols are basically provided with `getTypeDescContextItems`, doing this will duplicate class entires.

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MethodCallExpressionNodeContext.java
TharushiJay: @nadeeshaan Yes, that did occur to me. But since we should only allow methods in the `start` action expression, I thought it would be alright directing to `MethodCallExpressionNodeContext` . I have added a comment in line [#61](https://github.com/ballerina-platform/ballerina-lang/pull/31210/files#diff-998e23560dea026640124f3c72c1a969d50f5e3cb847da96f3b72bc78a10e231R61) regarding this. Would it work?

nadeeshaan (approved):
LGTM

Issue #30911, opened 2021-06-01
Resolving packages without internet


hemikak:
**Description:**
Currently when trying to build a package having `import ballerina/jballerina.java` gives an error when internet is disconnected. Ideally this should not fail as `ballerina/jballerina.java` is already there in the distribution. This is the same with `ballerina/io` import.

**Steps to reproduce:**
Build the following without internet.
```ballerina
import ballerina/jballerina.java;

public function pi() returns float = @java:FieldGet {
   name:"PI",
   'class:"java/lang/Math"
} external;
```

**Affected Versions:**
ballerina-swan-lake-beta1

**OS, DB, other environment details and versions:**
Mac

IrushiL:
Blocking on implementation of package resolution diagnostics.

PR #31217, opened 2021-06-16
Resolve packages without internet

pramodya1994:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30911

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/repositories/RemotePackageRepository.java
```diff
@@ -133,6 +133,8 @@ public static RemotePackageRepository from(Environment environment, Path cacheDi
             }
         } catch (ConnectionErrorException e) {
             // ignore connect to remote repo failure
+            // set `offline` true
+            this.isOffline = true;
```
hevayo, line 137: Doing this might take the repo indefinitely offline. This might be an issue when using project api in a service. Can we make the offline only affect the current compilation

azinneera: IMO what we need to do is simply add a diagnostic and continue. But we would need to implement diagnostic reporting in Package Resolution phase for that.

azinneera: Shall we just print the error as a warning and continue the run?? @hevayo wdyt??

compiler/ballerina-lang/src/main/java/io/ballerina/projects/util/ProjectUtils.java
```diff
@@ -636,6 +647,15 @@ private static void addDependencyContent(StringBuilder content, String org, Stri
         return possiblePkgNames;
     }
 
+    private static boolean isBuiltInPackage(ImportModuleRequest importModuleRequest) {
```
hevayo, line 650: We have a isBuiltInPackage in package descriptor can we make both use a single API

hevayo (changes requested):
please see inline comments

compiler/ballerina-lang/src/main/java/io/ballerina/projects/util/ProjectUtils.java
pramodya1994: Added this separate method since in  L636 we don't have access to `PackageDescriptor`.

hevayo: then can package description use the Util function ?

hevayo (approved):

Issue #30987, opened 2021-06-03
Compiler crashes for query returning an anonymous function array


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
public function main() {
    int[] x = [1, 2];

    (function () returns int)[] fn = 
        from var m in x  
        select function () returns int => 0;
}
```

```log
[2021-06-03 22:35:26,940] SEVERE {b7a.log.crash} - Index 0 out of bounds for length 0 
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:459)
	at org.wso2.ballerinalang.compiler.desugar.QueryDesugar.visit(QueryDesugar.java:1284)
	at org.wso2.ballerinalang.compiler.tree.expressions.BLangLambdaFunction.accept(BLangLambdaFunction.java:59)
	at org.wso2.ballerinalang.compiler.desugar.QueryDesugar.visit(QueryDesugar.java:1372)
	at org.wso2.ballerinalang.compiler.tree.statements.BLangAssignment.accept(BLangAssignment.java:80)
	at org.wso2.ballerinalang.compiler.desugar.QueryDesugar.lambda$visit$7(QueryDesugar.java:1289)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at org.wso2.ballerinalang.compiler.desugar.QueryDesugar.visit(QueryDesugar.java:1288)
	at org.wso2.ballerinalang.compiler.tree.expressions.BLangLambdaFunction.accept(BLangLambdaFunction.java:59)
	at org.wso2.ballerinalang.compiler.desugar.QueryDesugar.addSelectFunction(QueryDesugar.java:644)
	at org.wso2.ballerinalang.compiler.desugar.QueryDesugar.buildStream(QueryDesugar.java:360)
	at org.wso2.ballerinalang.compiler.desugar.QueryDesugar.desugar(QueryDesugar.java:251)
[...]

PR #31242, opened 2021-06-17
Fix issue with query expressions having anonymous functions

lasinicl:
## Purpose
$title

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30987

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
```diff
@@ -904,6 +904,12 @@ public void visit(BLangLambdaFunction bLangLambdaFunction) {
         SymbolEnv symbolEnv = env.createClone();
         bLangLambdaFunction.capturedClosureEnv = symbolEnv;
         BLangFunction enclInvokable = (BLangFunction) symbolEnv.enclInvokable;
+        // Since lambda functions enclosed in a query lambda are not included in the
+        // package node lambda list need to visit them here.
+        if (enclInvokable.flagSet.contains(Flag.QUERY_LAMBDA)) {
```
MaryamZi, line 909: Any particular reason why we do not model anonymous functions generated for query expressions similar to normal anonymous functions?

I believe there's some improvements we can do for normal anonymous functions (like local anonymous functions not having to be available at module-level until desugar) but desugar and after, shouldn't have much of a difference, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
```diff
@@ -1299,16 +1303,28 @@ private BRecordTypeSymbol getFrameTypeSymbol() {
...
+            // module-level anonymous function has an expr function body.
+            if (function.getBody().getKind() == NodeKind.EXPR_FUNCTION_BODY) {
+                BLangExprFunctionBody anonLambdaBody = (BLangExprFunctionBody) function.getBody();
+                anonLambdaBody.expr.accept(this);
+            } else {
+                BLangBlockFunctionBody anonLambdaBody = (BLangBlockFunctionBody) function.getBody();
+                List<BLangStatement> stmts = new ArrayList<>(anonLambdaBody.getStatements());
+                stmts.forEach(stmt -> stmt.accept(this));
+            }
+        }
```
MaryamZi, line 1327: Can be made an else-if and an else instead?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/query/QueryExprWithAnonFunctionExprs.java
```diff
@@ -0,0 +1,61 @@
...
+import org.ballerinalang.test.BRunUtil;
+import org.ballerinalang.test.CompileResult;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+/**
+ * This contains methods to test query expressions with anonymous function expressions.
+ *
+ * @since Swan Lake
```
MaryamZi, line 31:
```suggestion
 * @since 2.0.0
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
lasinicl: This was done to ensure that the lambda functions created for query clauses would be visited prior to the local anonymous functions in query clauses so that the closures would be resolved correctly. Updated the query desugar to handle it from the query side itself.

tests/jballerina-unit-test/src/test/resources/test-src/query/query-expr-with-anon-func-exprs.bal
```diff
@@ -0,0 +1,383 @@
...
+    f1 = fn[1];
+    value = f1();
+    assertEqual(value, 20);
+}
+
+function testQueryWithAnonFuncExprInWhereClause() {
+    int[] x = [1, 2];
+
+    int[] arr = from int m in x
+                let function () returns int func = function () returns int => m * y
```
pcnfernando, line 51: Try out a test with function that returns a function
eg:
`
function dummyFunc () returns (int) {
    return 1 + 2;
}

from...
let function () returns (function () returns (int)) func = function () returns (function () returns (int)) => dummyFunc
`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
```diff
@@ -1299,16 +1304,30 @@ private BRecordTypeSymbol getFrameTypeSymbol() {
...
+        if (function.flagSet.contains(Flag.QUERY_LAMBDA)) {
+            currentFrameSymbol = function.requiredParams.get(0).symbol;
+            identifiers = new HashMap<>();
+            currentLambdaBody = (BLangBlockFunctionBody) function.getBody();
+            currentLambdaEnv = SymbolEnv.createFuncBodyEnv(currentLambdaBody, env);
+            List<BLangStatement> stmts = new ArrayList<>(currentLambdaBody.getStatements());
+            stmts.forEach(stmt -> stmt.accept(this));
+            currentFrameSymbol = null;
+            identifiers = null;
+            currentLambdaBody = null;
```
pcnfernando, line 1316: Can you try out a nested lambda function example? Probably you'll have to maintain previous state and set at the end

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -7097,7 +7097,9 @@ public void visit(BLangTypeConversionExpr conversionExpr) {
     @Override
     public void visit(BLangLambdaFunction bLangLambdaFunction) {
         // Collect all the lambda functions.
-        env.enclPkg.lambdaFunctions.add(bLangLambdaFunction);
+        if (!env.enclPkg.lambdaFunctions.contains(bLangLambdaFunction)) {
```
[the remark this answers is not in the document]
lasinicl: Added the lambda functions in the query and the lambda functions we create for the clauses when we are visiting the lambda functions in `QueryDesugar` [1]. This was done to ensure that they are added in the order they are visited so that the closures are resolved properly. Hence those should not be added to the `env.enclPkg.lambdaFunctions` when they are being visited in `Desugar`.
[1] https://github.com/ballerina-platform/ballerina-lang/blob/740c7ec80a28b255773fa79052418996b73b96b9/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java#L1329

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
```diff
@@ -1300,14 +1304,30 @@ private BRecordTypeSymbol getFrameTypeSymbol() {
...
+            currentQueryLambdaBody = (BLangBlockFunctionBody) function.getBody();
+            List<BLangStatement> stmts = new ArrayList<>(currentQueryLambdaBody.getStatements());
+            stmts.forEach(stmt -> stmt.accept(this));
+            currentFrameSymbol = null;
+            identifiers = null;
+            currentQueryLambdaBody = null;
+        } else {
+            boolean prevWithinLambdaFunc = withinLambdaFunc;
+            withinLambdaFunc = true;
+            if (function.getBody().getKind() == NodeKind.EXPR_FUNCTION_BODY) {
```
pcnfernando, line 1319: Shouldn't these visits be added and the body be visited thereafter?

tests/jballerina-unit-test/src/test/resources/test-src/query/query-expr-with-anon-func-exprs.bal
```diff
@@ -0,0 +1,489 @@
...
+                                               string hello = "Hello ";
+                                               return hello + firstName + " " + lastName;
+                                           };
+
+type Card record {|
+   string msg;
+   int noOfEmp;
+|};
+
+(function () returns Card)[] globalFn5 = from var person in personList
```
pcnfernando, line 389: Trying to use the argument within the arrow function crashed at runtime. We can handle it separately if needed.
```ballerina
(function () returns string)[] globalFn5 = ttttFunc("sdasd");

function ttttFunc(string abc) returns (function () returns string)[] {
    var res = from var person in personList
                                             from var empFunc in (from var {firstName, lastName, dept} in employeeList
                                             select function () returns string => abc + " " + dept)
                                             select empFunc;
    return res;
}
```

lasinicl: It seems this is not specific to argument being used in an anonymous function in the query. Using an argument in an inner query in a nested query crashes at runtime. The following crashes at runtime.
```ballerina
function foo(string msg) returns string[] {
    var res = from var person in personList
              from var emp in (from var {firstName, lastName, dept} in employeeList
                               select msg + " " + firstName)
               select emp;
    return res;
}
```
Created https://github.com/ballerina-platform/ballerina-lang/issues/31629. @pcnfernando Shall we fix this separately?

pcnfernando: Tried with a `join` instead of a nested `from`. Couldn't reproduce. +1 let's handle it separately

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
```diff
@@ -1349,6 +1378,11 @@ public void visit(BLangFieldBasedAccess fieldAccessExpr) {
         fieldAccessExpr.expr.accept(this);
     }
 
+    @Override
+    public void visit(BLangFieldBasedAccess.BLangStructFunctionVarRef structFunctionVarRef) {
+        structFunctionVarRef.expr.accept(this);
+    }
```
pcnfernando, line 1384: Uncovered visits within QueryDesugar mean we don't have tests with such statements being used inside a query action/expression. Better to create an issue to track this adding tests

pcnfernando (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
lasinicl: Created https://github.com/ballerina-platform/ballerina-lang/issues/31658

PR #31251, opened 2021-06-17
Fix not showing the escaped-name in symbols with compiler API

### LINKED ISSUE #29494: Functions with ' prefix is missed in the suggestions
Screenshot 2021-03-23 at 18 33 29

**Steps to reproduce:**

**Affected Versions:**

### LINKED ISSUE #30242: No completions are provided for field access experssions

dulajdilshan:
## Purpose
$title

### Note: 
Here, we introduce an API to compare the name of the symbol against a given name, according to the semantics of identifiers in the spec

In `io.ballerina.compiler.api.symbols.Symbol`

```Java
boolean nameEquals(String name);
```

Fixes #30242 
Fixes #29494 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/ballerinalang/model/symbols/Symbol.java
```diff
@@ -35,6 +35,8 @@
 
     Name getName();
 
+    Name getOriginalName();
```
pubudu91, line 38: This is not what we discussed right? IIRC, the suggestion was to change `getName()` to return the original name.

dulajdilshan: I think we have discussed to keep the original-name as a new attribute inside the `ballerinalang.model.symbols.Symbol`. In compiler-api symbol, it will return the original name when we called `getName()` (if there is a one) :thinking:

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionSymbol.java
```diff
@@ -102,6 +102,7 @@ public boolean deprecated() {
         protected List<Qualifier> qualifiers = new ArrayList<>();
         protected List<AnnotationSymbol> annots = new ArrayList<>();
         protected FunctionTypeSymbol typeDescriptor;
+        protected boolean replaceOriginalName = false;
```
pubudu91, line 105: Why is this needed? 🤔

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
```diff
@@ -223,5 +233,13 @@ public SymbolBuilder(String name, SymbolKind symbolKind, BSymbol bSymbol, Compil
...
          */
         public abstract BallerinaSymbol build();
+
+        /**
+         * Build the Ballerina Symbol.
+         *
+         * @param  replaceOriginalName     whether to replace the original name or not
+         * @return {@link BallerinaSymbol} built
+         */
+//        public abstract BallerinaSymbol build(boolean replaceOriginalName);
```
pubudu91, line 243: Unused code?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
```diff
@@ -70,18 +76,18 @@ protected BallerinaSymbol(String name, SymbolKind symbolKind, BSymbol symbol, Co
     public Optional<String> getName() {
         // In the langlib context, reserved keywords can be used as regular identifiers. Therefore, they will not be
         // escaped.
-        if (this.escapedName != null) {
-            return Optional.of(this.escapedName);
-        }
```
pubudu91, line 75: Removing this means the name will be computed each time `getName()` is called. What's the reasoning behind the change in this section?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
dulajdilshan: Will remove this

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/LangLibrary.java
```diff
@@ -140,7 +140,8 @@ public static LangLibrary getInstance(CompilerContext context) {
 
     private void populateMethodList(List<FunctionSymbol> list, Map<String, BInvokableSymbol> langLib) {
         for (Map.Entry<String, BInvokableSymbol> entry : langLib.entrySet()) {
-            FunctionSymbol method = symbolFactory.createFunctionSymbol(entry.getValue(), entry.getKey());
+            String name = entry.getValue().getOriginalName().getValue();
+            FunctionSymbol method = symbolFactory.createFunctionSymbol(entry.getValue(), name);
```
pubudu91, line 144: If the key is not required, can consider looping through `values()` of the map, instead of the entry set.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFactory.java
```diff
@@ -375,7 +364,7 @@ public ParameterSymbol createBallerinaParameter(BVarSymbol symbol, ParameterKind
         if (symbol == null) {
             return null;
         }
-        String name = symbol.getName().getValue().isBlank() ? null : symbol.getName().getValue();
+        String name = symbol.getOriginalName().getValue().isBlank() ? null : symbol.getOriginalName().getValue();
```
pubudu91, line 367: Do we set empty string to the original name as well for this? 🤔  If we don't, we don't need this check.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFactory.java
```diff
@@ -525,8 +514,9 @@ public BallerinaConstantSymbol createConstantSymbol(BConstantSymbol constantSymb
      */
     public BallerinaAnnotationSymbol createAnnotationSymbol(BAnnotationSymbol symbol) {
         BallerinaAnnotationSymbol.AnnotationSymbolBuilder symbolBuilder =
-                new BallerinaAnnotationSymbol.AnnotationSymbolBuilder(symbol.name.getValue(), symbol,
-                                                                      this.context);
+                new BallerinaAnnotationSymbol.AnnotationSymbolBuilder(symbol.getOriginalName().getValue(),
+                        symbol,
+                        this.context);
```
pubudu91, line 519: Let's format this properly

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionSymbol.java
```diff
@@ -134,7 +134,8 @@ public FunctionSymbolBuilder withAnnotation(AnnotationSymbol annot) {
         @Override
         public BallerinaFunctionSymbol build() {
             return new BallerinaFunctionSymbol(this.name, this.qualifiers, this.annots,
-                                               this.typeDescriptor, (BInvokableSymbol) this.bSymbol, this.context);
+                    this.typeDescriptor, (BInvokableSymbol) this.bSymbol,
+                    this.context);
```
pubudu91, line 138: This as well.

pubudu91: Btw, this seems to be just a formatting change?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/Symbol.java
```diff
@@ -35,6 +35,13 @@
      */
     Optional<String> getName();
 
+    /**
+     * Retrieves the original name of the symbol if it is associated with an identifier.
+     *
+     * @return The name of the symbol if applicable
+     */
+    Optional<String> getOriginalName();
```
pubudu91, line 43: What's the need for this? IIRC, the original discussion was to change `getName()` to return the original name 🤔

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaFunctionSymbol.java
dulajdilshan: I'll correct it. This happened due to two commits including a revert commit

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/Symbol.java
dulajdilshan: In lang libs, Since we are changing the `getName()` to return the unquoted names for quoted names (Ex: `'map()`), we may ended up having same [error](https://github.com/ballerina-platform/ballerina-lang/issues/30242) for `'map()`. 

Please try this example
```ballerina
public function main() {
    string[] hello = ["a", "b"];
    int[] mapResult = hello.'map(function (string a) returns int {
        return 12;
    }).<cursor>
}
```

We need two APIs to retrieve the original name (with the quote) and the changed name (which we get from `getName()`

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFactory.java
dulajdilshan: We may get an empty string

dulajdilshan:
Waiting for this https://github.com/ballerina-platform/ballerina-lang/issues/31936

mohanvive:
@dulajdilshan @pubudu91 shall we merge this PR?

pubudu91:
> @dulajdilshan @pubudu91 shall we merge this PR?

Going through this now. Will check and merge.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/Symbol.java
```diff
@@ -66,4 +66,12 @@
...
      */
     Optional<Location> getLocation();
+
+    /**
+     * Check whether the given name equals to the name of the symbol.
+     *
+     * @param name The name that needs to be compared
+     * @return true if the name is similar to the symbol's name and false otherwise
+     */
+    boolean nameEquals(Optional<String> name);
```
pubudu91, line 76: Shall we not use `Optional` in the param here? Anyone who wants to compare a name would have to first wrap it in `Optional`. Plus use of `Optional` in params is not the intended use for it.

pubudu91: Also, let's elaborate a bit more in the API docs on what exactly this method does: comparing the symbol's name against the given name, according to the semantics defined in the spec regarding identifiers.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
```diff
@@ -126,6 +131,22 @@ public Location location() {
         return Optional.ofNullable(this.position);
     }
 
+    @Override
+    public boolean nameEquals(Optional<String> name) {
+        Optional<String> symbolName = this.getName();
+        Optional<String> symbolOriginalName = Optional.of(this.name);
```
pubudu91, line 137: No need to wrap this in `Optional` IINM

tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/LangLibFunctionTest.java
```diff
@@ -338,11 +338,11 @@ public void testIntersectionType() {
...
-                                            "remove", "push", "filter", "pop", "lastIndexOf", "iterator", "removeAll",
-                                            "setLength", "slice", "enumerate", "unshift", "map", "indexOf",
-                                            "cloneWithType", "cloneReadOnly", "toBalString", "toJson", "isReadOnly",
-                                            "fromJsonWithType", "mergeJson", "clone", "ensureType", "toString",
-                                            "toJsonString");
+                "remove", "push", "filter", "pop", "lastIndexOf", "iterator", "removeAll",
+                "setLength", "slice", "enumerate", "unshift", "map", "indexOf",
+                "cloneWithType", "cloneReadOnly", "toBalString", "toJson", "isReadOnly",
+                "fromJsonWithType", "mergeJson", "clone", "ensureType", "toString",
+                "toJsonString");
```
pubudu91, line 345: This file seems to  contain a lot of formatting changes? Better if the IDE is set to format only the modified sections of a file

pubudu91 (commented):
Shall we add tests to cover the scenario(s) we originally set out to solve? Also need tests for the `nameEquals()` method

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/AbstractTypeSymbol.java
```diff
@@ -102,6 +102,11 @@ public boolean assignableTo(TypeSymbol targetType) {
         return types.isAssignable(this.bType, getTargetBType(targetType));
     }
 
+    @Override
+    public boolean nameEquals(String name) {
+        return false;
```
pubudu91, line 107: This will have to be overridden in some type symbols. e.g., type ref, builtin subtypes

Let's maybe go with this for now as the initial implementation and address it in a separate PR.

dulajdilshan: I kept this like this because I didn't see any requirement for alternative implementation in built-in subtypes. Do we have any?

dulajdilshan: I mean the subtypes of `AbstractTypeSymbol`

pubudu91: Consider a type ref type symbol. Those are named. But this will always return `false` for type ref type symbols too. Same goes for other type symbols in which the `getName()` method implementation of `AbstractTypeSymbol` is overridden.

dulajdilshan: Ah got the idea :+1:

dulajdilshan: Let's add this change. I'll update the PR

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSymbol.java
```diff
@@ -126,6 +131,23 @@ public Location location() {
...
+        }
+        String symName = symbolName.get();
+        if (name.equals(symName)) {
+            return true;
+        }
+        String unEscapedName = unEscapeIfEscaped(name);
+        if (unEscapedName.equals(symName)) {
+            return true;
+        }
+        return name.equals(unEscapeIfEscaped(symName));
```
pubudu91, line 148: This doesn't look quite right. For example, consider something like the following:
```ballerina
string 'string;
```
And you try to compare the above var's name with the string `\u{73}tring`. This should return `true`.

dulajdilshan: Oops. My bad. Just corrected. Plese check

tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolAtCursorTest.java
```diff
@@ -207,36 +208,39 @@ public void testQuotedIdentifiers(int line, int column, String expSymbolName) {
...
+                {85, 7, "'int", "int"},
+                {86, 8, "'from", "from"},
+                {86, 19, "'from", "from"},
+                {88, 4, "'Example", "Example"},
+                {88, 13, "'new", "new"},
+                {89, 4, "'new", "new"},
+                {89, 9, "'anydata", "anydata"},
+                {90, 11, "'foo", "foo"},
+                {91, 15, "'string", "string"},
+                {92, 12, "'intVal", "\\u{69}ntVal"}
```
pubudu91, line 243: Shall we add a few more unicode identifier related tests like this? Different combinations

tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolAtCursorTest.java
```diff
@@ -191,13 +191,14 @@ public void testMissingConstructs(int line, int column) {
...
-    public void testQuotedIdentifiers(int line, int column, String expSymbolName) {
+    public void testQuotedIdentifiers(int line, int column, String expSymbolName, String expSymbolNameUnQuoted) {
         Project project = BCompileUtil.loadProject("test-src/symbol_at_cursor_quoted_identifiers_test.bal");
         SemanticModel model = getDefaultModulesSemanticModel(project);
         Document srcFile = getDocumentForSingleSource(project);
 
         Optional<Symbol> symbol = model.symbol(srcFile, LinePosition.from(line, column));
-        symbol.ifPresent(value -> assertEquals(value.getName().get(), expSymbolName));
+        symbol.ifPresent(value ->
+                assertTrue(value.nameEquals(expSymbolName) && value.nameEquals(expSymbolNameUnQuoted)));
```
pubudu91, line 201: Shall we add this as 2 different assert statements? So that if there's a failure we can easily recognize which comparison failed

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFactory.java
```diff
@@ -133,9 +133,9 @@ public Symbol getBCompiledSymbol(BSymbol symbol, String name) {
                     if (Symbols.isFlagOn(symbol.flags, Flags.RESOURCE)) {
                         return createResourceMethodSymbol((BInvokableSymbol) symbol);
                     }
-                    return createMethodSymbol((BInvokableSymbol) symbol);
+                    return createMethodSymbol((BInvokableSymbol) symbol, name);
                 }
-                return createFunctionSymbol((BInvokableSymbol) symbol, name);
+                return createFunctionSymbol((BInvokableSymbol) symbol, symbol.getOriginalName().getValue());
```
pubudu91, line 138: Why do we need a different approach just for this?

dulajdilshan: This is a mistake, I'll send the change

pubudu91 (approved):

PR #31276, opened 2021-06-18
Fix SO issue of BObject and BIntersection toString

### LINKED ISSUE #30001: Runtime stack overflow from cloneReadOnly on array of recursive read-only objects
This gets a StackOverflowError at runtime:

```
type Obj readonly & object {
    function foo(Obj t) returns Obj;
};

public function main() {
   Obj[] v = [];
   any x = v.cloneReadOnly();
}
```

KRVPerera:
Fixes #30001

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BIntersectionType.java
```diff
@@ -81,6 +84,20 @@ public int getTag() {
 
     @Override
     public String toString() {
+        if (resolving) {
```
MaryamZi, line 87: Can you share an example of when this fails for the intersection type rather than the effective type?

tests/jballerina-unit-test/src/test/resources/test-src/object/readonly_objects.bal
```diff
@@ -305,6 +305,16 @@ public function testReadOnlyClassIntersectionWithValidQualifiers() {
     assertEquality(9876, getBits(new Corge()));
 }
 
+type Obj readonly & object {
+    function foo(Obj t) returns Obj;
+};
+
+public function testRecursiveObjectArrayReadonlyClone() {
```
MaryamZi, line 312: Let's add relevant tests for each type that has been changed in this PR.

MaryamZi (commented):
@KRVPerera, there are test failures too.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BIntersectionType.java
```diff
@@ -81,6 +84,20 @@ public int getTag() {
...
+            return "";
+        }
+        resolving = true;
+        computeStringRepresentation();
+        resolving = false;
+        return cachedToString;
+
```
rdulmina, line 94:
```suggestion
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BIntersectionType.java
KRVPerera: I have to rework the code experiencing some issues in debugging. Almost like race conditions.
Check the same example in the original problem. The return type of the method and parameter is the intersection type of the object.
So this stack overflow loops over Object and Intersection both. Intersection type having a collection of other types is easily susceptible to this issue. That is why I decided to fix it there too.

Screenshot 2021-06-21 at 1 27 40 PM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -441,20 +419,41 @@ public BIntersectionType getImmutableType() {
...
+            return;
+        }
+        if (tsymbol != null && !tsymbol.getName().getValue().isEmpty()) {
+            String typeName = tsymbol.getName().getValue();
+            String packageId = tsymbol.pkgID.toString();
+            boolean isTypeParam = Symbols.isFlagOn(flags, Flags.TYPE_PARAM);
+            // improve readability of cyclic union types
+            if (isCyclic && (pCloneable.matcher(typeName).matches() ||
+                    (isTypeParam && pCloneableType.matcher(typeName).matches()))) {
+                System.out.printf("433");
```
gimantha, line 437: Is this for testing purposes?

KRVPerera: Done 31673ea

gimantha (approved):
LGTM

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BIntersectionType.java
MaryamZi: But shouldn't this get fixed automatically when the object case is fixed?

KRVPerera: Missed the comment.
What if this also happens with records?

KRVPerera: Missed the comment.
What if this also happens with records?

KRVPerera: Missed the comment.
What if this also happens with records?

KRVPerera: Missed the comment.
What if this also happens with records?

KRVPerera: I will check if this is redundant and revert the change in another PR.

Issue #31239, opened 2021-06-17
Typechecker getType passes even we provide java String


warunalakshitha:
**Description:**
Need to make sure the runtime does not have java strings.

https://github.com/ballerina-platform/ballerina-lang/blob/master/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java#L388

Issue #31282, opened 2021-06-18
Runtime throw Class cast error when interop function return invalid value


warunalakshitha:
**Description:**
Ballerina
```ballerina
public function acceptIntAnydataReturn() returns anydata= @java:Method {
    'class:"xyz",
    name:"acceptIntUnionReturn"
} external;
```

Java
```java

    public static Object acceptIntUnionReturn() {
                return new Module("ww","ww");
        }
    }

```
Error

error: java.lang.ClassCastException {"message":"class io.ballerina.runtime.api.Module cannot be cast to class io.ballerina.runtime.api.values.BValue (io.ballerina.runtime.api.Module and io.ballerina.runtime.api.values.BValue are in unnamed module of loader 'app')"}
        at waruna.project2.0_1_0:acceptIntAnydataReturn(main.bal:15)
           waruna.project2.0_1_0:interopWithAnydataReturn(main.bal:8)
           waruna.project2.0_1_0:main(main.bal:4)
```


```

warunalakshitha:
We can provide a error  io.ballerina.runtime.api.Module can not be assigned to anydata.

PR #31283, opened 2021-06-18
Remove java String check from type checker

warunalakshitha:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/31239
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


riyafa:
I have been wondering if we should include a test that would throw and exception if a String is used instead of a BString?

warunalakshitha:
@riyafa String is also the same as other java classes with these changes. ATM if you return a String for any data we will get exception like https://github.com/ballerina-platform/ballerina-lang/issues/31282. I think we need to properly fix that

riyafa:
can't we check for the `java.lang.ClassCastException` for now in a test and then fix the test later once the above issue is fixed?

riyafa (approved):

riyafa (approved):

Issue #30688, opened 2021-05-20
User defined error type owner is set to `lang.annotations`


IMS94:
**Description:**
Consider a user defined error type like below residing in say `mod1`:
```
public type Mod1Error error;
```
the `owner` of the user defined error type is set to `lang.annotations`. This doesn't happen to errors like `type Err error<MyRec>`. The `owner` is used by the `ReferenceFinder` to determine references to module `mod1`.

**Steps to reproduce:**
See description

**Affected Versions:**
Beta1 snapshot

**Related Issues (optional):**
#29276

IMS94:
This cause `goto defs` to not work for user defined error types with type reference `error` (ex: `type Err error`). Module of such user defined type symbols are set to `lang.annotations` too. 

Blocks #29276

IMS94:
Relates to #29232

mohanvive:
Relates to https://github.com/ballerina-platform/ballerina-lang/issues/26284

mohanvive:
@pubudu91 Shall we prioritise and check this?

pubudu91:
> @pubudu91 Shall we prioritise and check this?

The correct module info gets returned now with the fix done for #26284.

pubudu91:
Checked again in 2201.0.0, the correct module gets returned for this. Hence closing this.

Issue #26284, opened 2020-10-08
Type aliases for builtin types gets effectively discarded


pubudu91:
Consider the following:
```ballerina
type Foo int;
Foo x = 10;
```
For the above, a type `Foo` gets defined and there's a type symbol. But since it's an alias for a builtin type, it gets handled differently when resolving. When defining the variable, we resolve the type node, `Foo`. But when resolving, we don't keep track of the symbol it resolved to. Only the type of the symbol. Since the type is `int` and it's builtin, whenever we refer to `type.tsymbol`, it returns the type symbol for the builtin type `int`, instead of the type symbol for `Foo`. To fix this, we may need to refactor how the symbol resolver works. Currently it returns a type, although it's named SymbolResolver. I think we can resolve this issue if we return the symbol itself from it.

Issue: #26283 is sorta similar to this, but the difference is it's dealing with more complex and user defined types, therefore we create individual type symbols for each of those. Here we can't change the type symbol for `int`.

MaryamZi:
This was also raised in https://github.com/ballerina-platform/ballerina-lang/issues/26470#issuecomment-725295493

rdhananjaya:
Probably the root cause is same for this one too https://github.com/ballerina-platform/ballerina-lang/issues/25742

pubudu91:
To fix this issue, we need to do the following:
- Introduce a separate BSymbol (e.g., BTypeDefinitionSymbol) for type definitions instead of relying on BTypeSymbol for type definitions. 
- Introduce the type reference type descriptor for representing usages of types defined using type definitions, constants etc. This means introduction of a new type tag and modifying the usages of the type tag.

pcnfernando:
Compiler changes added to https://github.com/ballerina-platform/ballerina-lang/tree/type_ref_type_impl

PR #31297, opened 2021-06-19
Fix enum type shadowing

### LINKED ISSUE #30805: Type definition shadows Enum symbol
Observed in SL-Beta1-RC4.

```ballerina
public enum Foo {
    foo, 
    bar
}

type Foo int; // Should be an error.

public function main() {
    Foo x = "foo"; // Not working.
}
```

gimantha:
Fixes #30805 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -313,7 +313,10 @@ private boolean isDistinctSymbol(BSymbol symbol, BSymbol foundSym) {
 
     private boolean hasSameOwner(BSymbol symbol, BSymbol foundSym) {
         // check whether the given symbol owner is same as found symbol's owner
-        if (foundSym.owner == symbol.owner) {
+        if (foundSym.owner == symbol.owner ||
+                (foundSym.owner.tag & SymTag.PACKAGE) == SymTag.PACKAGE &&
+                (symbol.owner.tag & SymTag.PACKAGE) == SymTag.PACKAGE &&
+                foundSym.pkgID.equals(symbol.pkgID)) {
```
MaryamZi, line 319: Isn't this issue due to/related to https://github.com/ballerina-platform/ballerina-lang/issues/26284, https://github.com/ballerina-platform/ballerina-lang/issues/30688. etc. 

```ballerina
public enum Foo {
    foo, 
    bar
}

type Foo int;
```

The owner of the symbol for `type Foo int` is what is incorrect here (lang.annotations), right? I think we need to fix the type aliasing issues before/instead of introducing these changes here.

@pubudu91, thoughts please?

pubudu91: This check is ok I guess. But wondering how this fixed the issue though. Like Maryam said, in the type def's symbol the owner would be the `lang.annotations` module. 🤔

MaryamZi: We do have a symbol for `Foo` in `type Foo int;` here and the `pkgID` is the current package's ID. It's just that the owner of that is the lang.annotations module.

We shouldn't need this check though, right? I think we should be able to rely on the owner. IINM, there shouldn't be scenarios where two symbols' package IDs are the same but the owner package symbols are different?

pubudu91: Ah I see. Yeah, with the fix for type aliases, ideally shouldn't need this. IIRC, we don't create package symbols like we do for other kinds of symbols. But for now, don't see any harm in this though.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -313,7 +313,11 @@ private boolean isDistinctSymbol(BSymbol symbol, BSymbol foundSym) {
 
     private boolean hasSameOwner(BSymbol symbol, BSymbol foundSym) {
         // check whether the given symbol owner is same as found symbol's owner
-        if (foundSym.owner == symbol.owner) {
+        if (foundSym.owner == symbol.owner ||
+                //TODO: remove this, once #26284, #30688 are fixed
```
MaryamZi, line 317: Shall we create an issue too?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/natives/BuiltinLoadingTest.java
```diff
@@ -31,17 +31,18 @@
     @Test
```
MaryamZi, line 31: Sorry, missed this previously, but should we add the new test to `org.ballerinalang.test.variable.shadowing.ShadowingNegativeTest` instead?

tests/jballerina-unit-test/src/test/resources/test-src/natives/builtin-symbol-negative.bal
```diff
@@ -9,3 +9,10 @@ type 'int int;
 type 'never never;
 type 'readonly readonly;
 type 'xml xml;
+
+public enum Foo {
```
MaryamZi, line 13: Shall we add a positive test to `org.ballerinalang.test.variable.shadowing.ShadowingTest` also?

e.g.,
```ballerina
type Foo int;

public function main() {
    string Foo = "hello";
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
gimantha: sure

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/natives/BuiltinLoadingTest.java
gimantha: sure, will move the test

tests/jballerina-unit-test/src/test/resources/test-src/natives/builtin-symbol-negative.bal
gimantha: sure

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
gimantha: created https://github.com/ballerina-platform/ballerina-lang/issues/31433

MaryamZi (approved):

PR #31342, opened 2021-06-21
[LS] Fix annotations not showing without attached node

### LINKED ISSUE #29420: @test:Config is not suggested in auto completion
**Description:**
$subject
Screenshot from 2021-03-23 11-35-47

**Suggested Labels:**

**Suggested Assignees:**

**Affected Product Version:**

**Steps to reproduce:**

**Related Issues:**

### LINKED ISSUE #30913: No suggestions for annotations in http module when typeing at EOF
**Description:**
when I type `@http:` at the end of the file no suggestions are shown. It shows suggestions if there is a top-level construct that follows it. Should be a parser issue.

IMS94:
## Purpose
$subject

Fixes #29420
Fixes #30913

## Approach
Updated the logic in annotation related completion provider to check for the `no attached node` diagnostic.
Also updated the completion config json replacing logic to preserve order

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/AnnotationNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/AnnotationNodeContext.java
@@ -87,7 +87,7 @@ public class AnnotationNodeContext extends AbstractCompletionProvider<Annotation
         List<Symbol> visibleSymbols = ctx.visibleSymbols(ctx.getCursorPosition());
         return visibleSymbols.stream()
                 .filter(symbol -> symbol.kind() == SymbolKind.ANNOTATION
-                        && this.matchingAnnotation((AnnotationSymbol) symbol, attachedNode))
+                        && this.matchingAnnotation((AnnotationSymbol) symbol, attachedNode, ctx))
                 .map(symbol -> AnnotationUtil.getAnnotationItem((AnnotationSymbol) symbol, ctx))
                 .collect(Collectors.toList());
     }
@@ -110,7 +110,7 @@ public class AnnotationNodeContext extends AbstractCompletionProvider<Annotation
 
         return moduleEntry.orElseThrow().allSymbols().stream()
                 .filter(symbol -> symbol.kind() == SymbolKind.ANNOTATION
-                        && this.matchingAnnotation((AnnotationSymbol) symbol, attachedNode))
+                        && this.matchingAnnotation((AnnotationSymbol) symbol, attachedNode, context))
                 .map(symbol -> AnnotationUtil.getAnnotationItem((AnnotationSymbol) symbol, context))
                 .collect(Collectors.toList());
     }
@@ -123,8 +123,14 @@ public class AnnotationNodeContext extends AbstractCompletionProvider<Annotation
         return node.parent();
     }
 
-    private boolean matchingAnnotation(AnnotationSymbol symbol, Node attachedNode) {
+    private boolean matchingAnnotation(AnnotationSymbol symbol, Node attachedNode, BallerinaCompletionContext context) {
         if (symbol.attachPoints().isEmpty()) {
+            return true;
+        }
+        // This is to handle a special case where the user is typing an annotation in the module context
+        // before writing the attached node. The parser detects the attached node as a module var declaration
+        // and mark it's parts as missing.
+        if (context.getCursorPositionInTree() == attachedNode.textRange().endOffset()) {
             return true;
         }
         switch (attachedNode.kind()) {
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/AnnotationNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/AnnotationNodeContext.java
@@ -28,6 +28,7 @@ import io.ballerina.compiler.syntax.tree.QualifiedNameReferenceNode;
 import io.ballerina.compiler.syntax.tree.SyntaxKind;
 import io.ballerina.projects.Module;
 import io.ballerina.projects.Project;
+import io.ballerina.tools.diagnostics.Diagnostic;
 import org.ballerinalang.annotation.JavaSPIService;
 import org.ballerinalang.langserver.common.utils.AnnotationUtil;
 import org.ballerinalang.langserver.common.utils.CommonUtil;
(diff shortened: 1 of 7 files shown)


nadeeshaan:
@IMS94 As per the related parser fix, there should be a diagnostic attached to the annotation node. What if we depend on that since it is more reliable when compared to the dummy node's positioning details because it has a chance to change. WDYT?

lochana-chathura:
> In a related note, the particular diagnostic cannot be found in the node. @lochana-chathura @rdulmina any clue?

I tried the two test samples. They both have the diagnostic(`ERROR_METADATA_NOT_ATTACHED_TO_A_TOP_LEVEL_CONSTRUCT`) aiye. Note that, the diagnostic is attached to the module-var-decl node. not in the attentions itself. 
<details>
  <summary><b>syntax-tree-with-diagnostics</b>  (<i>click to expand</i>)</summary>

```json
{
          "kind": "MODULE_VAR_DECL",
          "hasDiagnostics": true,
          "diagnostics": [
            "ERROR_METADATA_NOT_ATTACHED_TO_A_TOP_LEVEL_CONSTRUCT"
          ],
          "children": [
            {
              "kind": "METADATA",
              "hasDiagnostics": true,
              "children": [
                {
                  "kind": "LIST",
                  "hasDiagnostics": true,
                  "children": [
                    {
                      "kind": "ANNOTATION",
                      "hasDiagnostics": true,
                      "children": [
                        {
                          "kind": "AT_TOKEN",
                          "leadingMinutiae": [
                            {
                              "kind": "END_OF_LINE_MINUTIAE",
                              "value": "\n"
                            }
                          ]
                        },
                        {
                          "kind": "QUALIFIED_NAME_REFERENCE",
                          "hasDiagnostics": true,
                          "children": [
                            {
                              "kind": "IDENTIFIER_TOKEN",
                              "value": "module1"
                            },
                            {
                              "kind": "COLON_TOKEN",
                              "trailingMinutiae": [
                                {
                                  "kind": "WHITESPACE_MINUTIAE",
                                  "value": " "
                                }
                              ]
                            },
                            {
                              "kind": "IDENTIFIER_TOKEN",
                              "isMissing": true,
                              "hasDiagnostics": true,
                              "diagnostics": [
                                "ERROR_MISSING_IDENTIFIER"
                              ]
                            }
                          ]
                        }
                      ]
                    }
                  ]
                }
```
</details>

nadeeshaan:
> > In a related note, the particular diagnostic cannot be found in the node. @lochana-chathura @rdulmina any clue?
> 
> I tried the two test samples. They both have the diagnostic(`ERROR_METADATA_NOT_ATTACHED_TO_A_TOP_LEVEL_CONSTRUCT`) aiye. Note that, the diagnostic is attached to the module-var-decl node. not in the annotation itself.
> 
> syntax-tree-with-diagnostics (click to expand)

@lochana-chathura yes, the particular diagnostic is attached to the module-variable-declaration node. When we capture the annotation node shouldn't the attachment node/ the metadata node should contain the particular diagnostic? Metadata/ annotation it-self is the one who is isolated.

rdulmina:
> > > In a related note, the particular diagnostic cannot be found in the node. @lochana-chathura @rdulmina any clue?
> > 
> > 
> > I tried the two test samples. They both have the diagnostic(`ERROR_METADATA_NOT_ATTACHED_TO_A_TOP_LEVEL_CONSTRUCT`) aiye. Note that, the diagnostic is attached to the module-var-decl node. not in the annotation itself.
> > syntax-tree-with-diagnostics (click to expand)
> 
> @lochana-chathura yes, the particular diagnostic is attached to the module-variable-declaration node. When we capture the annotation node shouldn't the attachment node/ the metadata node should contain the particular diagnostic? Metadata/ annotation it-self is the one who is isolated.

+1

lochana-chathura:
@nadeeshaan Quoting from our original conversation back then(https://github.com/ballerina-platform/ballerina-lang/issues/29420#issuecomment-810731155).  Following was the requirement and instead of `isMissing` property, we went ahead with the diagnostic in the parent. 

> When we get the annotation node, we extract the parent of it, and based on the parent kind we filter the annotations. In this particular case, the parent would be module-no-init-var-decl. In such case, we should be able to do a validation against the parent and ignore. If we are going with this approach, I would say that we should make sure the parent is set with the isMissing to true.

Instead, we can do the following.
`modVarDeclrNode.metadata().annotations()` is a list node, to which we cannot attach diagnostics. We can loop through the list and add the diagnostic to each `AnnotationNode`. And also to the `markdownDocumentationNode`. Please create an issue if we need that requirement.

nadeeshaan:
> @nadeeshaan Quoting from our original conversation back then([#29420 (comment)](https://github.com/ballerina-platform/ballerina-lang/issues/29420#issuecomment-810731155)). Following was the requirement and instead of `isMissing` property, we went ahead with the diagnostic in the parent.
> 
> > When we get the annotation node, we extract the parent of it, and based on the parent kind we filter the annotations. In this particular case, the parent would be module-no-init-var-decl. In such case, we should be able to do a validation against the parent and ignore. If we are going with this approach, I would say that we should make sure the parent is set with the isMissing to true.
> 
> Instead, we can do the following.
> `modVarDeclrNode.metadata().annotations()` is a list node, to which we cannot attach diagnostics. We can loop through the list and add the diagnostic to each `AnnotationNode`. And also to the `markdownDocumentationNode`. Please create an issue if we need that requirement.

@lochana-chathura yes, as per our original discussion, we initiate the decision making based on the annotation's properties. In this particular case, the attached node. My concern is, the particular diagnostic should be the one which we should use to decide processing, in the sense it is illogical to use the parent's/ moduleVar's diagnostic to decide the annotation's behaviour.

lochana-chathura:
> @lochana-chathura yes, as per our original discussion, we initiate the decision making based on the annotation's properties. In this particular case, the attached node. My concern is, the particular diagnostic should be the one which we should use to decide processing, in the sense it is illogical to use the parent's/ moduleVar's diagnostic to decide the annotation's behaviour.

Created #31356 to address this.

nadeeshaan:
> > @lochana-chathura yes, as per our original discussion, we initiate the decision making based on the annotation's properties. In this particular case, the attached node. My concern is, the particular diagnostic should be the one which we should use to decide processing, in the sense it is illogical to use the parent's/ moduleVar's diagnostic to decide the annotation's behaviour.
> 
> Ack. Created #31356 to address this.

@IMS94 shall we wait for #31356 then?

IMS94:
@nadeeshaan sure. Marking this as blocked by #31356

nadeeshaan (approved):
LGTM

Issue #31262, opened 2021-06-17
Central API client needs to print error messages coming wit 503 status code


hemikak:
**Description:**
$subject. 503 comes when services goes into maintenance mode.

hemikak:
Subject was wrong earlier. It should 503 status code. Not 502. Hence changing the subject.

PR #31389, opened 2021-06-23
Handle service unavailable (503) error from central

pramodya1994:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/31262

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
+++ b/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
@@ -42,8 +42,14 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
-import javax.net.ssl.HttpsURLConnection;
-
+import static java.net.HttpURLConnection.HTTP_BAD_GATEWAY;
+import static java.net.HttpURLConnection.HTTP_BAD_REQUEST;
+import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
+import static java.net.HttpURLConnection.HTTP_MOVED_TEMP;
+import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
+import static java.net.HttpURLConnection.HTTP_NO_CONTENT;
+import static java.net.HttpURLConnection.HTTP_OK;
+import static java.net.HttpURLConnection.HTTP_UNAUTHORIZED;
 import static org.ballerinalang.central.client.CentralClientConstants.ACCEPT;
 import static org.ballerinalang.central.client.CentralClientConstants.ACCEPT_ENCODING;
 import static org.ballerinalang.central.client.CentralClientConstants.APPLICATION_OCTET_STREAM;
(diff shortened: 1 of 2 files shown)


cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
```diff
@@ -136,8 +142,9 @@ public Package getPackage(String orgNamePath, String packageNamePath, String ver
                     }
     
                     // If request sent is wrong or error occurred at remote repository
-                    if (getPackageResponse.code() == HttpsURLConnection.HTTP_BAD_REQUEST ||
-                        getPackageResponse.code() == HttpsURLConnection.HTTP_INTERNAL_ERROR) {
+                    if (getPackageResponse.code() == HTTP_BAD_REQUEST ||
```
azinneera, line 145: Instead of throwing the same exception for all errors will it be better to throw different errors for different error ranges. E.g. one for http client errors (code: 400-499) and another for http server errors (500-599). I think when troubleshooting, this will make it easier to identify where the issue lies. @hemikak @hevayo WDYT?

hemikak: In central API implementation level, we usually return the same error structure but with different error messages. The error message should be sufficient enough to identify the underlying cause.

cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
```diff
@@ -136,8 +142,9 @@ public Package getPackage(String orgNamePath, String packageNamePath, String ver
...
+                        getPackageResponse.code() == HTTP_INTERNAL_ERROR ||
+                        getPackageResponse.code() == HTTP_BAD_GATEWAY) {
```
hemikak, line 147: I see that 502(HTTP_BAD_GATEWAY) and 503(HTTP_SERVICE_UNAVAILABLE) can be caused at the LB level as well, in which it returns an html payload without going to . That would cause an error when doing 
```java
Error error = new Gson().fromJson(body.get().string(), Error.class);
```

A sample response from LB:
```html
<html>
<head><title>404 Not Found</title></head>
<body>
<center><h1>404 Not Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

Also 503 is not captured here.

cli/central-client/src/test/java/org/ballerinalang/central/client/TestCentralApiClient.java
```diff
@@ -284,6 +284,32 @@ public void testGetPackageWithBadRequest() throws IOException, CentralClientExce
...
+        String resString = "{message: \"package pushing is temporarily disabled. please try again later.\"}";
+
+        Request mockRequest = new Request.Builder()
+                .get()
+                .url("https://localhost:9090/registry/packages/bar/" + WINERY + "/v2")
+                .build();
+        Response mockResponse = new Response.Builder()
+                .request(mockRequest)
+                .protocol(Protocol.HTTP_1_1)
+                .code(HttpURLConnection.HTTP_BAD_REQUEST)
```
hemikak, line 300: This should be bad gateway ?

cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
pramodya1994: @hemikak Here we are only checking the 502, So if we add another if for 503, It will solve this issue right?
All 502 responses should have an error message, Is that the case?

cli/central-client/src/test/java/org/ballerinalang/central/client/TestCentralApiClient.java
pramodya1994: Added

cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
pramodya1994: Added 503

hemikak (approved):

azinneera (approved):

PR #31390, opened 2021-06-23
Enable indirect recursions

### LINKED ISSUE #30101: Runtime Bad, Sad
This program gets a bad, sad when executed:

```
type XType XNil|XBoolean|XInt|XString|
           XUnion|XIntersection|
	   XNever|XAny|
	   XListRef|XFunctionRef;

const XNil = "nil";
const XBoolean = "boolean";
const XInt = "int";
const XString = "string";
const XNever = "never";
const XAny = "any";

type XUnion ["|", XType...];
type XIntersection ["&", XType...];

type ListDef XType[];
type FunctionDef XType[2];

type Defs record {|
   ListDef[] listDefs;
   FunctionDef[] functionDefs;
|};

type XListRef ["listRef", int];
type XFunctionRef ["functionRef", int];

XType test = ["|", "int", "nil"];
```

[ballerina-internal.log](https://github.com/ballerina-platform/ballerina-lang/files/6342904/ballerina-internal.log)

### LINKED ISSUE #30113: Subtype error with recursive rest tuple type
In slalpha4, this gets a compile-time error:

```
type MyRecType int|[MyRecType...];

public function main() {
  MyRecType d = [1, 2, 3];
  json j = d;
}
```

If `[MyRecType...]` is replaced by `MyRecType[]`, the error goes away.

suleka96:
Fixes #30101
Fixes #30113

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1033,6 +1037,11 @@ private boolean isTupleTypeAssignable(BType source, BType target, Set<TypePair>
             unresolvedTypes.add(pair);
         }
```
KRVPerera, line 1039: :recycle: Think we cache the result for `source.tag == TypeTags.TUPLE` since it is used in many places in this method. We can also populate `rhsTupleType` beforehand. 

:thinking: :question: I am wondering why the `tupleTypes` has to be empty for this scenario.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -1261,6 +1255,22 @@ private void checkErrors(SymbolEnv env, BLangNode unresolvedType, BLangNode curr
...
+        switch (unresolvedType) {
+            case OBJECT_TYPE:
+            case RECORD_TYPE:
+            case CONSTRAINED_TYPE:
+            case ARRAY_TYPE:
+            case TUPLE_TYPE_NODE:
+            case TABLE_TYPE:
+            case ERROR_TYPE:
+            case FUNCTION_TYPE:
+                return true;
```
KRVPerera, line 1268: ❓ Don't we need `STREAM_TYPE`?

:wrench: The only time we call this method we don't pass `OBJECT_TYPE` is it intentional. I assume `RECORD_TYPE` should also behave the same way. Perhaps we need a different method name here and some tests for these scenarios.

suleka96: Not passing `OBJECT_TYPE` was intentional because in [[1]](https://ballerina.io/ballerina-spec/spec.html#section_5.5.3) under `5.5.3.6 Inclusion and type-ids` it says:

> It is an error for an object-type-descriptor to directly or indirectly include itself.

`RECORD_TYPE` gets defined in the very first iteration and doesn't come into the `unresolved` list and therefore do we need to restrict that as well?

KRVPerera: That is for object inclusion but we can have a recursive type by having the object as a field. Yes, and objects also get defined in one iteration and field are added later. This is not necessarily the first iteration.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: Please revisit point 2. Logic needs changes w.r.t. to JSON type. Probably need more test cases too.

suleka96:
Tracked in #31478

PR #31461, opened 2021-06-28
Refactor foreach statement for xml and string types

### LINKED ISSUE #31436: `foreach` statement added with broader types
**Description:**
$title. `foreach` statement for a `string` variable gets added with `string` instead of `string:Char` as the iteration type. This change was recently introduced in the FE - https://github.com/ballerina-platform/ballerina-lang/pull/31078.

**Steps to reproduce:**
```ballerina
public function main() {
    string s = "foo";

   s.f<cursor>
}
```

Accepting adds
```ballerina
    foreach string item in s {
        
    }
```

but we should add

```ballerina
    foreach string:Char item in s {
        
    }
```

**Affected Versions:**
slbeta2-rc3

malinthar:
## Purpose
$subject

Fixes #31436

## Samples
ezgif com-gif-maker(13)

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachCompletionUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachCompletionUtil.java
@@ -24,6 +24,8 @@ import io.ballerina.compiler.api.symbols.TableTypeSymbol;
 import io.ballerina.compiler.api.symbols.TupleTypeSymbol;
 import io.ballerina.compiler.api.symbols.TypeDescKind;
 import io.ballerina.compiler.api.symbols.TypeSymbol;
+import io.ballerina.compiler.api.symbols.UnionTypeSymbol;
+import io.ballerina.compiler.api.symbols.XMLTypeSymbol;
 import io.ballerina.compiler.syntax.tree.BlockStatementNode;
 import io.ballerina.compiler.syntax.tree.ExpressionStatementNode;
 import io.ballerina.compiler.syntax.tree.FieldAccessExpressionNode;
@@ -44,6 +46,7 @@ import org.eclipse.lsp4j.TextEdit;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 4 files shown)


IMS94 (approved):
LGTM!

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachCompletionUtil.java
```diff
@@ -183,8 +185,15 @@ public static String getTypeOfIteratorVariable(BallerinaCompletionContext ctx, T
         TypeSymbol rawType = CommonUtil.getRawType(symbol);
         switch (rawType.typeKind()) {
             case STRING:
+                type = CommonUtil.getModifiedTypeName(ctx, rawType) + ":Char";
```
nadeeshaan, line 188: For this, do we need to call the `getModifiedTypeName` since this is straight forward? What if we use `string:Char` directly?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachCompletionUtil.java
```diff
@@ -183,8 +185,15 @@ public static String getTypeOfIteratorVariable(BallerinaCompletionContext ctx, T
...
+                break;
             case XML:
-                type = CommonUtil.getModifiedTypeName(ctx, rawType);
+                Optional<TypeSymbol> xmlTypeParam = ((XMLTypeSymbol) symbol).typeParameter();
+                if (xmlTypeParam.isEmpty() || xmlTypeParam.get().typeKind() == TypeDescKind.UNION) {
```
nadeeshaan, line 192: Why do we need to special case the union here?
I believe it is for the following,

```
function test(xml<xml:Element | xml:Comment> xx) {
    foreach xml:Element | xml:Comment item in xx {

    }
}
```

Even in that case, it is okay to check the empty type-parameter and get the modified type parameter in other cases.

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ForeachCompletionUtil.java
malinthar: If we have the following 
  xml x1 = xml `<book>The Lost World</book>`;

we get the type-parameter as follows.

xml:Element|xml:Comment|xml:ProcessingInstruction|xml:Text

adding this union would be not suitable IMHO.

nadeeshaan: Basically we don't need to do this explicitly, and when we consider the language specification as well as the semantic API, the type parameter is optional. In the sense we should show the type parameters only when the user has added it in the source. If the user has added all four elements in the source, we can show it (which is a rare case).

nadeeshaan (approved):
LGTM

PR #31466, opened 2021-06-28
Introduce API to retrieve stored properties in the strand

chamil321:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Environment.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Environment.java
@@ -21,6 +21,7 @@ import io.ballerina.runtime.api.async.StrandMetadata;
 import io.ballerina.runtime.internal.scheduling.State;
 import io.ballerina.runtime.internal.scheduling.Strand;
 
+import java.util.Map;
 import java.util.Optional;
 
 /**
@@ -107,7 +108,7 @@ public class Environment {
      * @param value value to be store in the strand
      */
     public void setStrandLocal(String key, Object value) {
-        strand.setProperty(key, value);
+        strand.getProperties().put(key, value);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


manuranga:
If people can't write ballerina programs without having access to low level access Ballerina internal, it's big issue. We should try and get rid of Strand hacks, not improve them. Better to discuss this kind of thing with a lager audience, how about ballerina spec repo?

manuranga:
Maybe what this shows is ballerina language need support for thread locals ?

chamil321:
> If people can't write ballerina programs without having access to low level access Ballerina internal, it's big issue. We should try and get rid of Strand hacks, not improve them. Better to discuss this kind of thing with a lager audience, how about ballerina spec repo?

Agree to consider this as a ballerina requirement. Because thread local variable are quite important aspect.

Ideally there should be a proper way from the runtime to propagate thread local variable from the parent strand to the child strand. The PR ballerina-platform/ballerina-lang#31466 is a new API to access the complete property map and pass it over. That will reduce the cost of access each property and setting it back like the improvement done in the ballerina-platform/module-ballerina-http#496

But it's better if they provide propagation from the runtime.
case 1: When ballerina thread start, i.e invoke resource function from java, properties has to be passed
case 2: When another thread start after a parent thread execution, parent properties are propagated internally

However I'll create an issue on the spec repo.

manuranga:
I understand that there are issues with current implementation, maybe we can improve it until we have a proper solution, but I don't think we should make it even bigger. If you look at the ballerina spec, it doesn't even have a concept of Strand parent, so I feel like what we have done is a spec violation.

So we first need a spec issues justifying having a parent strand concept and being able pass though it. Let's describe the basic requirement without getting into our implementation.

anupama-pathirage:
@chamil321 / @shafreenAnfar Do we need this PR? Looks like it got automatically closed without any attention. If you really don't need this PR anymore, please remove the `stale` label from the PR. Then it won't get added to the `[Stale PR] Ballerina Stale PR Details` mail in the future.

Issue #30740, opened 2021-05-24
[Debugger] `STEP_OVER` from the last line of object's `init` method doesn't hit the next line of the object initialization step.


praveennadarajah:
**Description:**
Currently, if we `STEP_OVER` from the last line of object's `init` method, debugger doesn't hit the next line of the object initialization step, instead it hits the last line of `class Person`.

**Steps to reproduce:**
Consider the below code segment:
```
class Person {
    public int age;
    public string firstName;
    public string lastName;

    function init(int age, string firstName, string lastName) {
        self.age = age;
        self.firstName = firstName;
        self.lastName = lastName;
    }

    function getFullName() returns string {
        return self.firstName + " " + self.lastName;
    }
}

public function main() {
    Person p1 = new (5, "John", "Doe");
}
```

- Add debug point in the last line of Person class `init()` method and start the program in debug mode.
- Once the debug point gets hit, send a `STEP_OVER` request.
- Debugger hits the last line of `class Person` instead of hitting the next line of the object initialization step.

**Affected Versions:**
```
Ballerina Swan Lake Beta 1
Language specification v2020-12-17
Update Tool 1.3.5
```

PR #31506, opened 2021-06-30
Fix `STEP_OVER` from the last line of object `init()` method doesn't hit the next line of the object initialization step issue

praveennadarajah:
## Purpose
This PR will fix `STEP_OVER` from the last line of the object `init()` method doesn't hit the next line of the object initialization step issue.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30740

### DIFF
--- a/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
+++ b/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
@@ -283,9 +283,12 @@ public class JBallerinaDebugServer implements IDebugProtocolServer {
             if (loadedThreadFrames.containsKey(activeThread.uniqueID())) {
                 stackTraceResponse.setStackFrames(loadedThreadFrames.get(activeThread.uniqueID()));
             } else {
+                // if the active thread's top stack frame name is `$init$`, then send a `STEP_INTO` request.
+                // This is to avoid `$init$` stack frame information represented on the top of the call stack,
+                // as this stack frame does not give any useful information to the user.
                 if (activeThread.frames().get(0).location().method().name().equals(METHOD_INIT)) {
-                    prepareFor(DebugInstruction.STEP_OVER);
-                    eventProcessor.sendStepRequest(args.getThreadId(), StepRequest.STEP_OVER);
+                    prepareFor(DebugInstruction.STEP_IN);
+                    eventProcessor.sendStepRequest(args.getThreadId(), StepRequest.STEP_INTO);
                 }
                 StackFrame[] validFrames = activeThread.frames().stream()
                         .map(this::toDapStackFrame)
--- a/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
+++ b/misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
@@ -138,7 +138,7 @@ public class JBallerinaDebugServer implements IDebugProtocolServer {
     private static final String SCOPE_NAME_LOCAL = "Local";
     private static final String SCOPE_NAME_GLOBAL = "Global";
     private static final String VALUE_UNKNOWN = "unknown";
-    private static final String METHOD_INIT = "$init$";
+    private static final String J_INIT_FRAME_NAME = "$init$";
     private static final String COMPILATION_ERROR_MESSAGE = "error: compilation contains errors";
 
     public JBallerinaDebugServer() {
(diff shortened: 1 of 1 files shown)


misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -282,6 +283,10 @@ ClientConfigHolder getClientConfigHolder() {
             if (loadedThreadFrames.containsKey(activeThread.uniqueID())) {
                 stackTraceResponse.setStackFrames(loadedThreadFrames.get(activeThread.uniqueID()));
             } else {
+                if (activeThread.frames().get(0).location().method().name().equals(METHOD_INIT)) {
+                    prepareFor(DebugInstruction.STEP_OVER);
+                    eventProcessor.sendStepRequest(args.getThreadId(), StepRequest.STEP_OVER);
+                }
```
NipunaRanasinghe, line 289: I suppose we should do a `step-in` request in here. Am I missing something?

NipunaRanasinghe: Also better to add a comment explaining **why** we are doing this, for the future reference.

NipunaRanasinghe:
@praveennadarajah Don't we need to modify our integration test suite to cover the above scenario?

praveennadarajah:
> @praveennadarajah Don't we need to modify our integration test suite to cover the above scenario?

@NipunaRanasinghe Created an issue https://github.com/ballerina-platform/ballerina-lang/issues/31507 to add debugger integration tests covering all possible scenarios related to the ballerina `object`.

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/31506/commits/e88b66976d90596c9a5915accfdad27204e71e8b

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -138,6 +138,7 @@
     private static final String SCOPE_NAME_LOCAL = "Local";
     private static final String SCOPE_NAME_GLOBAL = "Global";
     private static final String VALUE_UNKNOWN = "unknown";
+    private static final String METHOD_INIT = "$init$";
```
NipunaRanasinghe, line 141:
```suggestion
    private static final String J_INIT_FRAME_NAME = "$init$";
```

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -282,6 +283,13 @@ ClientConfigHolder getClientConfigHolder() {
             if (loadedThreadFrames.containsKey(activeThread.uniqueID())) {
                 stackTraceResponse.setStackFrames(loadedThreadFrames.get(activeThread.uniqueID()));
             } else {
+                // if the active thread's top stack frame name is `$init$`, then send a `STEP_INTO` request.
+                // This is to avoid `$init$` stack frame information represented on the top of the call stack,
+                // as this stack frame does not give any useful information to the user.
```
NipunaRanasinghe, line 288:
```suggestion
                // If the active thread's top stack frame name is equal to `J_INIT_FRAME_NAME `,  need to abort the processing and send a `step-in` request to resume the VM to stop at its next state.
                // This is because the above state is related to java-level object initialisation and therefore suspending the VM at that intermediate state does not provide any useful information to the user.
```

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
```diff
@@ -282,6 +283,13 @@ ClientConfigHolder getClientConfigHolder() {
...
+                if (activeThread.frames().get(0).location().method().name().equals(METHOD_INIT)) {
```
NipunaRanasinghe, line 289: Don't we need to handle NPEs and other exceptions when directly accessing these stack frame information?

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/31506/commits/c351d222a78a9ae0aad3971a59c71a74e7c42dfc

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/31506/commits/c351d222a78a9ae0aad3971a59c71a74e7c42dfc

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/JBallerinaDebugServer.java
praveennadarajah: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/31506/commits/c351d222a78a9ae0aad3971a59c71a74e7c42dfc

NipunaRanasinghe (approved):

NipunaRanasinghe (approved):

PR #31547, opened 2021-07-02
Add extestion for retrieving typedata from expression

CharukaK:
## Purpose
$subject


anupama-pathirage:
@CharukaK / @kaviththiranga Do we need this PR? Looks like it got automatically closed without any attention. If you really don't need this PR anymore, please remove the `stale` label from the PR. Then it won't get added to the `[Stale PR] Ballerina Stale PR Details` mail in the future.

CharukaK:
This is no longer needed, We can remove the stale label
I don't have permission for it

PR #31604, opened 2021-07-06
Add parameter related APIs for remote functions

### LINKED ISSUE #31581: Retrieve the Parameter Names from a Remote Method at Runtime
**Description:**
GraphQL module is going to introduce support for mutation operations, through ballerina remote methods.

Similar to the resource functions, there is a requirement for reading the parameter names of the remote methods. Can we have an API to retrieve the parameter names of a remote method (something similar to `resourceMethodType.getParamNames()`)?

Related: #27417, #30106

If there is a default value assigned to a function, we need to know the default value of that function too.

HindujaB:
## Purpose
$subject

Fixes #31581 

## Approach
Added the following APIs to `BRemoteMethodType` & `BResourceMethodType`
- `Parameter[] getParameters()`
The `Parameter` object contains the parameter name and the parameter defaultability values in the fields `name` and `isDefault`.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/FunctionParameter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/FunctionParameter.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *  Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
  *
  *  WSO2 Inc. licenses this file to you under the Apache License,
  *  Version 2.0 (the "License"); you may not use this file except
@@ -15,21 +15,20 @@
  *  specific language governing permissions and limitations
  *  under the License.
  */
-package io.ballerina.runtime.api.types;
 
-import io.ballerina.runtime.api.FunctionParameter;
+package io.ballerina.runtime.api;
 
 /**
- * {@code ResourceFunctionType} represents a resource function in Ballerina.
+ * {@code {@link FunctionParameter} represents the parameter of a function in ballerina.
  *
  * @since 2.0
  */
-public interface ResourceMethodType extends MethodType {
-    @Deprecated
-    String[] getParamNames();
-    String getAccessor();
-    String[] getResourcePath();
-    @Deprecated
-    Boolean[] getParamDefaultability();
-    FunctionParameter[] getParameters();
+public class FunctionParameter {
+    public final String name;
+    public final boolean isDefault;
+
+    public FunctionParameter(String name, Boolean isDefault) {
+        this.name = name;
+        this.isDefault = isDefault;
+    }
 }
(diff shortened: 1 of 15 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/RemoteMethodType.java
```diff
@@ -27,4 +27,8 @@
     ObjectType getParentObjectType();
 
     FunctionType getType();
+
+    String[] getParamNames();
+
+    Boolean[] getParamDefaultability();
 }
```
riyafa, line 34: Shall we return a class with the ParamName and the defaultability status? This is not nice to have two functions to return the details about Paramters

ThisaruGuruge: +1. But in that case, it's better to change the resource functions too. That'll be a kinda big change though.

warunalakshitha: Since we need default values +1 to have that change in both resource and remote methods.
ex. 
Param[] getParameters();

Param can contain name and defaultability + default value if there is any

HindujaB:
Created issue https://github.com/ballerina-platform/ballerina-standard-library/issues/1646 to remove standard lib usage of deprecated APIs.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/FunctionParameter.java
```diff
@@ -0,0 +1,34 @@
+/*
+ *  Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
warunalakshitha, line 2:
```suggestion
 *  Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/nativeimpl/jvm/runtime/api/tests/Values.java
```diff
@@ -50,4 +59,40 @@ public static BObject getObject(BString objectName) {
...
+        if (remoteType == null) {
+            return ValueCreator.createArrayValue(TypeCreator.createArrayType(PredefinedTypes.TYPE_STRING, 0), 0);
+        }
+        FunctionParameter[] parameters = remoteType.getParameters();
+        int len = parameters.length;
+        BString[] params = new BString[len];
+        for (int i = 0; i < len; i++) {
+            params[i] = StringUtils.fromString(parameters[i].name);
+        }
+        return ValueCreator.createArrayValue(params);
```
riyafa, line 74: Can't we return an array of a pair(tuple), with the paramName and the defaultability together? Then we wouldn't need two methods? We can have a single function called `getParmeters`?

warunalakshitha (approved):

Issue #31394, opened 2021-06-23
Incorrect code generated with const


jclark:
slbeta1 miscompiles this:

```
import ballerina/io;

const TAG_SHIFT = 56;
const TAG_MASK = 0xff;

public function main() {
  int n = ~(TAG_MASK << TAG_SHIFT);
  io:println(n.toHexString());
}
```

Output is

```
ffffff
```
Should of course be:

```
ffffffffffffff
```

jclark:
Similar case:

```
import ballerina/io;

const TAG_SHIFT = 56;
const TAG_MASK = 0xff;

public function main() {
  int n = TAG_MASK << TAG_SHIFT;
  io:println(n.toHexString());
}
 ```

Output is

```
ffffffffff000000
```

Should be

```
ff00000000000000
```

In both cases changing `const` to `int` makes the problem go away.

PR #31616, opened 2021-07-07
Fix incorrect code generated with const

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/31394
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/31867

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/bytetype/BByteValueTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/bytetype/BByteValueTest.java
@@ -564,7 +564,15 @@ public class BByteValueTest {
         invokeUnsignedRightShiftOperatorTestFunction(a, i, j);
     }
 
-    private void invokeUnsignedRightShiftOperatorTestFunction(byte a, long i, long j) {
+    @Test(description = "Test bitwise unsigned right shift operator 2")
+    public void testBitwiseUnsignedRightShiftOperator3() {
+        int a = 0xff;
+        long i = -23445834;
+        long j = 5;
+        invokeUnsignedRightShiftOperatorTestFunction(a, i, j);
+    }
+
+    private void invokeUnsignedRightShiftOperatorTestFunction(int a, long i, long j) {
         BValue[] args = {new BByte(a), new BInteger(i), new BInteger(j)};
         BValue[] returns = BRunUtil.invoke(result, "testBitwiseUnsignedRightShiftOperator", args);
         Assert.assertEquals(returns.length, 2);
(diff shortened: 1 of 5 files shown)


HindujaB (approved):
LGTM

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/bytetype/BByteValueTest.java
```diff
@@ -564,7 +564,15 @@ public void testBitwiseUnsignedRightShiftOperator2() {
         invokeUnsignedRightShiftOperatorTestFunction(a, i, j);
     }
 
-    private void invokeUnsignedRightShiftOperatorTestFunction(byte a, long i, long j) {
+    @Test(description = "Test bitwise unsigned right shift operator 2")
+    public void testBitwiseUnsignedRightShiftOperator3() {
+        int a = 0xff;
+        long i = -23445834;
+        long j = 5;
+        invokeUnsignedRightShiftOperatorTestFunction(a, i, j);
```
riyafa, line 572: shall we test when it is a `const` as in the issue?

riyafa: Shall we also test what happens when it is a byte shifted by a large value? We need to see if it becomes zero.

riyafa: Shall we test the left shift cases also?

warunalakshitha: Added test for byte shift with large value and becomes 0. Left shift cases are already added through PR. 

For constants, IMO since constants are compile-time, the constant operation should be work the same as the underlying type operation at runtime. Any issue with constant operation should be reproducible by a given type of operation. Adding a test would be redundant for me.

riyafa: The issue says `In both cases changing const to int makes the problem go away.` That's why I thought it might be an issue with the `const`. Don't you think it would be better to have atleast the test from the issue https://github.com/ballerina-platform/ballerina-lang/issues/31394

riyafa (approved):

Issue #28872, opened 2021-02-24
Documentation errors are not shown when there's semantic errors


rasika:
**Description:**
Documentation errors are not shown when there's semantic errors. This is a major blocker for language server since language server cannot provide quick-fixes related to documentation issues.

![no-docs](https://user-images.githubusercontent.com/1448489/109005934-844f5980-76d0-11eb-8358-3599ad890140.gif)

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

rasika:
Related discussion can be found in https://github.com/ballerina-platform/ballerina-lang/discussions/27938.

PR #31621, opened 2021-07-07
[LS] Fix documentation code action issues in resource functions

### LINKED ISSUE #29495: Resource function path parameters not documenting using code action
**Description:**
> 1. Issue when the resource function has path parameters (below example int id), it won't document in the  API documentation.

Screenshot from 2021-03-23 18-22-36

> 2. When clicking code action for documentation more than one time it will generate same API documentation more than one.

Screenshot from 2021-03-23 18-21-57

**Describe your problem(s)**

**Describe your solution(s)**

TharushiJay:
## Purpose

1. Document resource function path parameters using `Document this` code action 

Screenshot from 2021-07-07 15-33-31

2. Avoid adding duplicated documentation for already documented resource functions and service declarations

Fixes #29495

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


nadeeshaan (changes requested):
If there is no change required for the documentation, shall we avoid showing the documentation code action?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/docs/DocumentationGenerator.java
```diff
@@ -125,7 +127,9 @@ public static boolean hasDocs(NonTerminalNode node) {
         switch (node.kind()) {
             case FUNCTION_DEFINITION:
             case OBJECT_METHOD_DEFINITION:
+            case RESOURCE_ACCESSOR_DEFINITION:
             case METHOD_DECLARATION:
+            case SERVICE_DECLARATION:    
 //            case SERVICE_DECLARATION: {
 //                ServiceDeclarationNode serviceDeclrNode = (ServiceDeclarationNode) node;
 //                return semanticModel.symbol(fileName, serviceDeclrNode.typeDescriptor().map(s->s.lineRange()
```
TharushiJay, line 135: Can I remove the commented out code segment for `SERVICE_DECLARATION`?

nadeeshaan: Seems so. So what's the reason for enabling the service declaration?

TharushiJay: Otherwise it returned an `Optional.empty()` from `getDocumentableSymbol()`, which resulted in already documented service declarations (with its own custom documentation) being overwritten by the default documentation.
Eg: `# This is a service` was overwritten by `# Description`

Maybe if we're going to avoid showing the documentation code action when there is no change required for the documentation, there's no need to enable that?

nadeeshaan: Yeah. That would be a better approach. @IMS94 WDYT?

IMS94: @nadeeshaan +1 for the idea. But if we do not show `document this` for already documented nodes, we have to identify when the docs are mismatched with the node (added/removed parameter, added object field, etc). For that we have to use diagnostics and use the `update documentation` code action. 

Due to https://github.com/ballerina-platform/ballerina-lang/issues/28872, when the code has semantic errors, documentation diagnostics are not shown. This may cause use to not show `update documentation` code action for most cases. IMO, one thing we can do is, instead of showing `Document this`, we can show `Update documentation` for already documented nodes. WDYT?

nadeeshaan: @IMS94 no no, I meant not to show the documentation code action for the already documented ones. Not to show the documentation code action, when there is no mismatch in the documentation. IIRC even now we do that check to update the documentation upon the document this code action.

IMS94: @nadeeshaan Checking if there's a mismatch will need to go into depths like checking:
1. For functions -> checking parameters, return type changes; deprecated annotation changes etc
2. For services -> object field changes, deprecated annotation changes etc
3. For resource functions -> Checking for path param, query param changes
4. ...
IMO, we have to do the same processing done by type checker in our code action. And will require putting some effort. Is it worth the effort?

nadeeshaan: Even now we do that right? Even now we update the documentation upon the `Document This`. Isn't it a matter of reusing the same logic?

IMS94: We currently do not explicitly check if there's a mismatch. It's the `DocAttachmentInfo` class which check and update the documentation. And doc attachment info class is involved in the command execution step, not in code action stage.

nadeeshaan: So the bottom line is, we are going to cover up a limitation from LS side and to do that, we are going to introduce inconsistent user experiences, and for solving those we have altered the logic right? If that is the situation, I'm preferred with an approach to live with the limitation, instead of introducing a confusion of the behaviour. I think we had this discussion to update the documentation with the `Document This` code action and now it seems to be problematic.

IMS94: As discussed offline, we will proceed with this PR for now. Next, we need to update the `add documentation` code action to show the `update documentation` as code action title for nodes that has documentation. @TharushiJay please create an issue to track that.

IMS94 (approved):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/docs/DocumentationGenerator.java
TharushiJay: Created issue #31678 to track above.

nadeeshaan (approved):
LGTM

Issue #31452, opened 2021-06-28
The Below Program Results in BAD SAD


shafreenAnfar:
```Ballerina
import ballerina/http;
import ballerina/graphql;

service /graphql on new graphql:Listener(9000) {

    final http:Client covid_19 = check new("https://disease.sh/v3/covid-19");

    resource function get foo() returns WorldStatus|error? {
        return check new WorldStatus(self.covid_19);               
    }
}

service class WorldStatus {

    final WorldStatusData worldStatusData;

    public function init(http:Client covid_19) returns error? {
        self.worldStatusData = check covid_19->get("/all");
    }

    resource function get cases () returns int|error {
        return self.worldStatusData.cases;
    }

    resource function get deaths () returns int|error {
        return self.worldStatusData.deaths;
    }

    resource function get recovered () returns int|error {
        return self.worldStatusData.recovered;
    }
}

type WorldStatusData record {
    int cases;
    int deaths;
    int recovered;
    int active;
};
```
Error
```
[2021-06-28 11:16:51,987] SEVERE {b7a.log.crash} - null 
java.lang.NullPointerException
	at org.wso2.ballerinalang.compiler.semantics.analyzer.CodeAnalyzer.visit(CodeAnalyzer.java:3781)
	at org.wso2.ballerinalang.compiler.tree.expressions.BLangCheckedExpr.accept(BLangCheckedExpr.java:58)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.CodeAnalyzer.analyzeExpr(CodeAnalyzer.java:3983)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.CodeAnalyzer.visit(CodeAnalyzer.java:2521)
	at org.wso2.ballerinalang.compiler.tree.BLangSimpleVariable.accept(BLangSimpleVariable.java:53)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.CodeAnalyzer.analyzeNode(CodeAnalyzer.java:394)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.CodeAnalyzer.visit(CodeAnalyzer.java:422)
	at org.wso2.ballerinalang.compiler.tree.BLangClassDefinition.accept(BLangClassDefinition.java:106)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.CodeAnalyzer.analyzeNode(CodeAnalyzer.java:394)
[...]

Issue #31655, opened 2021-07-09
Reordering doesn't seem to happen with object constructor field initializers


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
final string|error notString = error("error!");

var x = object {
    string|error b = notString;

    function init() returns error? {
    }
};

public function main() {
    var y = checkpanic x;

    string|error yb = y.b;
    if yb is string {
        // panic error("should be error!");
        panic error(string `y.b is string '${yb}'`);
    }
}
```

The `b` field should hold an error value.

**Affected Versions:**
slbeta2, current master

KRVPerera:
- [ ] Test case also has a crash now
Probably due to the same issue
```LogTalk
[2021-09-27 13:12:22,289] SEVERE {b7a.log.crash} - null 
java.lang.NullPointerException
        at io.ballerina.runtime.internal.values.BmpStringValue.concat(BmpStringValue.java:47)
        at s1.main(s1.bal:16)
        at $_init.$lambda$main$(.)
        at io.ballerina.runtime.internal.scheduling.SchedulerItem.execute(Scheduler.java:608)
        at io.ballerina.runtime.internal.scheduling.Scheduler.run(Scheduler.java:333)
        at io.ballerina.runtime.internal.scheduling.Scheduler.runSafely(Scheduler.java:301)
        at java.base/java.lang.Thread.run(Thread.java:834)
 
 ```

PR #31656, opened 2021-07-09
Validate usage of `check` in record field default values and object field initializers

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/31452

This PR introduces the following changes.
- `check` cannot be used in the default value of a record field (unless within a function that may return error). 

```ballerina
final string x = "not an int";

type Foo record {|
    int i = check int:fromString(x); // compilation error
|};
```

- `check` cannot be used in the initializer of an object constructor/class field (unless within a function that may return error) if there is no `init` method or the return type of the `init` method does not contain the error that the expression may evaluate to. 

```ballerina
final string x = "not an int";

type MyError distinct error;

class Bar {
    int i = check int:fromString(x); // compilation error, no `init` method
}

class Baz {
    int i = check int:fromString(x); // compilation error, `init` method `error` return type mismatch

    function init() returns MyError? {
        
    }
}
```

- `check` can be used in the initializer of an object constructor/class field if the error the expression may evaluate to belongs to the return type of the `init` method. 

```ballerina
final string x = "not an int";

class Qux {
    int i = check int:fromString(x); // allowed now

    function init() returns error? {

    }
}
```

The compiler crashed for all of the above previously.

## Remarks
https://github.com/ballerina-platform/ballerina-lang/issues/31655

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java
```diff
@@ -542,7 +548,14 @@
 
         /* the body can be null in the case of Object type function declarations */
         if (funcNode.body != null) {
+
+            DefaultValueState prevDefaultValueState = this.defaultValueState;
+            if (defaultValueState == DefaultValueState.RECORD_FIELD_DEFAULT ||
```
rdhananjaya, line 553: shall we use `this.defaultValueState` in these 2 intances too, looks like `this.` is used through out the function

rdhananjaya (dismissed):

rdhananjaya (dismissed):

rdhananjaya (approved):

PR #31695, opened 2021-07-13
Improve parser completions at close brace

### LINKED ISSUE #28492: Remove unused parameters passed to parser error recovery
**Description:**
With the latest change of error recovery, we no longer need to pass arguments of the calling method to the recovery. Need to remove and refactor the code.

### LINKED ISSUE #31314: Improve recovery solution logic at EOF recovery
**Description:**
As a precaution at EOF recovery, we simply insert the token as the error handler solution.
When doing so, we get the insert token kind directly from the context as per the below lines.

https://github.com/ballerina-platform/ballerina-lang/blob/98c1a3eaf404810e3be30faa5a282e0ba8dc52a4/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/AbstractParserErrorHandler.java#L87-L88

This could cause a problem(as we insert `NONE` token) when the context is an alternative path 
or if it is just an intermediate context that does not refer to any token
(e.g. In `OBJECT_TYPE_DESCRIPTOR` -> `OBJECT_TYPE_START` rule, the first ctx is just an intermediate ctx)

We can modify the `getExpectedTokenKind()` logic as follows. 
If the context is an alternative path, we can get the expected token from the first alternative. If the first alternative itself an alternative path, we can consider its first alternative. Like that we can do this recursively.

Also re-evaluate the current insert token approach at the EOF recovery, as this could lead to different recovery when the cursor is at the end of file, which is a problem for LS.

### LINKED ISSUE #31628: Improve closing brace removal logic in the parser
**Description:**
Consider the following screen capture

https://user-images.githubusercontent.com/1329674/124863036-07b07e00-dfd4-11eb-9f3f-f347cd9717e8.mp4

**Steps to reproduce:**
Check where the missing closing brace diagnostic is shown (line 14). Also, we cannot determine whether the cursor is within the body of the class since the immediate close brace has been removed and evaluated as missing.

**Affected Versions:**
swanlake beta2 at least

lochana-chathura:
## Purpose
$subject.

Fixes #31628
Fixes #31314 
Fixes #28492

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


lochana-chathura:
Will fix the build failure

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -5052,6 +5470,30 @@ private boolean isParameter(ParserRuleContext ctx) {
...
+            return new Solution(Action.INSERT, ctx, kind, ctx.toString());
+        }
+        
+        if (hasAlternativePaths(ctx)) {
+            ctx = getShortestAlternative(ctx);
+            return getInsertSolution(ctx);
+        }
+        
+        ctx = getNextRule(ctx, 1);
+        return getInsertSolution(ctx);
```
rdulmina, line 5494: Why do we need this? Can't we just do this as we did in `XMLParserErrorHandler`

lochana-chathura: Resolving as per our offline discussion.

rdulmina (approved):

PR #31705, opened 2021-07-14
Fix type narrowing issue in multiple if-else statement

### LINKED ISSUE #31327: Type narrowing doesn't work properly in multiple if else blocks
Please consider the code line here: https://github.com/ballerina-platform/nballerina/blob/main/compiler/modules/front/codeGen.bal#L136

image

In line number 128, we are checking if `curBlock` is (). So in the `else if` coming next, we shouldn't have to cast it to `bir:BasicBlock`.  In the first  else if block, the cast is not used (which is correct), but in other `else if` blocks, it gives a compile error if we dont have a cast (which is wrong). 
`[codeGen.bal:(136:51,136:59)] incompatible types: expected 'wso2/nballerina.bir:0.1.0:BasicBlock', found 'wso2/nballerina.bir:0.1.0:BasicBlock?'`

Could not easily reproduce with a simple if else blocks.

gimantha:
Fixes #31327

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/ifelse/TypeGuardTest.java
```diff
@@ -167,9 +167,9 @@ public void testTypeGuardSemanticsNegative() {
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found " +
                 "'(int|string|boolean)'", 293, 25);
         BAssertUtil.validateError(negativeResult, i++, "incompatible types: expected 'int', found " +
-                "'(int|string|boolean)'", 296, 29);
+                "'(int|string)'", 296, 29);
```
MaryamZi, line 170: Is this correct? Shouldn't it be reset to the original type of the variable? For example, in https://github.com/ballerina-platform/ballerina-lang/blob/522385842fc3e961aa99e0dac090f0e193f208c9/tests/jballerina-unit-test/src/test/resources/test-src/statements/ifelse/type-guard-semantics-negative.bal#L292 we could set a boolean value to `x` right? In which case the error message isn't quite right.

MaryamZi: The following compiles with these changes and results in an incorrect value at runtime.

```ballerina
public function main() {
    int|string|boolean x = 8;
    int|string|boolean y = 8;

    if (x is int|string) {
        int|string a = x;
        if (x is int) {
            int b = x;
            if (x > 5) {
                int c = x;
                x = true;

                if (y is int) {
                    int|string e = x;
                    panic error(string `e's defined type: int|string, actual type: ${(typeof e).toString()}, value: ${e}`);
                }
            }
        }
    }
}
```

```cmd
e's defined type: int|string, actual type: typedesc boolean, value: true
```

MaryamZi:
There's a typo in the title. nultiple -> multiple

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2183,6 +2183,7 @@ public void visit(BLangIf ifNode) {
         Map<BVarSymbol, BType.NarrowedTypes> prevNarrowedTypeInfo = this.narrowedTypeInfo;
 
         SymbolEnv ifEnv = typeNarrower.evaluateTruth(ifNode.expr, ifNode.body, env);
+        typeNarrower.addHigherEnvTypeNarrowedSymbols(ifEnv);
```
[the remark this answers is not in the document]
gimantha: I will check on this

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/ifelse/TypeGuardTest.java
gimantha: > Is this correct? Shouldn't it be reset to the original type of the variable? For example, in
> 
> https://github.com/ballerina-platform/ballerina-lang/blob/522385842fc3e961aa99e0dac090f0e193f208c9/tests/jballerina-unit-test/src/test/resources/test-src/statements/ifelse/type-guard-semantics-negative.bal#L292
> 
> we could set a boolean value to `x` right? In which case the error message isn't quite right.

But in this case we have set a string to x right? so it is within the scope starting from line 285. So do we still need to reset back to the original type at line no 282?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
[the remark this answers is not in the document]
gimantha: I checked with this fix. The below type resetting issue is also fixed with this. Thanks

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/type-test-expr.bal
```diff
@@ -1227,6 +1227,28 @@ function testBuiltInSubTypeTypeTestAgainstFiniteType() {
     assertFalse(b is Ints);
 }
 
+public type SomeRecord record {
+    int intField;
+};
+
+public function testIfElseWithTypeTest() {
```
MaryamZi, line 1234: Shall we add a few more tests? For example, scenarios where multiple variables are reset, different variables are reset in different blocks, different nesting combinations (even though they are effectively an if block and else block with else-ifs also in the else), etc,? I think we should add some negative tests too (e.g., trying to use it as the narrowed type after resetting).

gimantha: will add

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/binaryoperations/TypeTestExprTest.java
```diff
@@ -47,6 +47,22 @@ public void testTypeTestExprSemanticsNegative() {
         BAssertUtil.validateError(negativeResult, 0, "unknown type 'C'", 35, 14);
     }
 
+    @Test
+    public void testTypeTestExprTypeNarrowingNegative() {
+        CompileResult negativeResult =
+                BCompileUtil.compile("test-src/expressions/binaryoperations/type-test-type-narrowing-negative.bal");
```
MaryamZi, line 53: We use underscores for bal file names.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/binaryoperations/TypeTestExprTest.java
```diff
@@ -47,6 +47,22 @@ public void testTypeTestExprSemanticsNegative() {
...
+        int i = 0;
+        BAssertUtil.validateError(negativeResult, i++,
+                "incompatible types: expected 'SomeRecord', found 'SomeRecord?'", 49, 17);
+        BAssertUtil.validateError(negativeResult, i++,
+                "incompatible types: expected '(SomeRecord|int)', found '(SomeRecord|int)?'", 52, 21);
+        BAssertUtil.validateError(negativeResult, i++,
+                "incompatible types: expected '()', found 'SomeRecord'", 63, 13);
+        BAssertUtil.validateError(negativeResult, i++,
+                "incompatible types: expected '(SomeRecord|int)', found '(SomeRecord|int)?'", 69, 13);
```
MaryamZi, line 62: Is this correct? Isn't the type of `g` `SomeRecord|int` in this block? And since `w` is `SomeRecord|int`, type narrowing shouldn't be reset in the current master?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/type-test-expr.bal
```diff
@@ -1227,6 +1227,126 @@ function testBuiltInSubTypeTypeTestAgainstFiniteType() {
...
+    if c is () {
+
+    } else if x is int {
+        c = f;
+    } else if (x is boolean) {
+        foo(c);
+    }
+}
+
+public function testIfElseWithTypeTestMultipleVariables() {
```
MaryamZi, line 1248: Shouldn't this and the next two functions be called?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/type-test-expr.bal
```diff
@@ -1227,6 +1227,126 @@ function testBuiltInSubTypeTypeTestAgainstFiniteType() {
...
+    int intField;
+};
+
+public function testIfElseWithTypeTest() {
+    SomeRecord? c = {intField: 10};
+    SomeRecord? f = ();
+    any x = true;
+
+    if c is () {
+
```
MaryamZi, line 1240: Shouldn't the test fail if we reach here? If not, we're not testing the complete runtime behaviour right? Same applies to the other functions introduced.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/type-test-expr.bal
gimantha: I was only checking the FE changes. I will add invokes

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2349,6 +2349,7 @@ public void visit(BLangIf ifNode) {
         }
 
         Map<BVarSymbol, BType.NarrowedTypes> prevNarrowedTypeInfo = this.narrowedTypeInfo;
+        Map<BVarSymbol, BType.NarrowedTypes> narrowedTypes = new HashMap<>();
```
MaryamZi, line 2352: Can we use a name here that reflects what this is used for? Along with a comment maybe?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/binaryoperations/TypeTestExprTest.java
```diff
@@ -48,6 +48,22 @@ public void testTypeTestExprSemanticsNegative() {
         BAssertUtil.validateError(negativeResult, 0, "unknown type 'C'", 35, 14);
     }
 
+    @Test
```
MaryamZi, line 51: Shouldn't this tests be added in org.ballerinalang.test.statements.ifelse.TypeGuardTest instead?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/type_test_type_narrowing_negative.bal
```diff
@@ -0,0 +1,86 @@
...
+
+}
+
+function goo(SomeRecord|int aa) {
+
+}
+
+function yoo(SomeRecord|int|() aa) {
+
+}
```
MaryamZi, line 86: Shall we also add negative tests like

```ballerina
public function main() {
    int|string? a = 1;
    int|string? b = 1;

    if a is int|string && b is int|string {
        if a is int {

        } else {
            if b is int {
                a = 1;
                b = 1;
            } else {
                string c = a; // OK now
                string d = b;
            }
        }

        int|string e = a; // error
        int|string f = b; // error
    }
}
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/binaryoperations/TypeTestExprTest.java
gimantha: moved

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2350,6 +2350,10 @@ public void visit(BLangIf ifNode) {
 
         Map<BVarSymbol, BType.NarrowedTypes> prevNarrowedTypeInfo = this.narrowedTypeInfo;
 
+        // This map keeps the narrowed types of inner if statements and propagate the false types to else block when
+        // flow goes from the if statement to else block in compile time.
```
MaryamZi, line 2354: Isn't it for after the else block completes rather than for the else block?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/binaryoperations/TypeTestExprTest.java
```diff
@@ -736,4 +736,24 @@ public void testTypeTestExpression(String functionName) {
                 "testIntSubtypes"
         };
     }
+
+    @Test
```
MaryamZi, line 740: These tests should also go in `TypeGuardTest`?

gimantha: moved

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
gimantha: I changed the comment, saying that we want to keep track of the false types after the else blocks, which we might need in the outer block

MaryamZi (approved):

PR #31708, opened 2021-07-14
Fix incorrect byte comparison

### LINKED ISSUE #31700: Bad code generated for byte comparison
This prints `false` when it should print `true`.

```
import ballerina/io;

public function main() {
    io:println(testByte(0x20));
}

function testByte(byte b) returns boolean {
    if b >= 0x20 {
       return true;
    }
    return false;
}
```

Changing `byte` to `int` or changing `0x20` to decimal both make the problem go away.

Nadeeshan96:
Fixes #31700

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/greater-less-than-operation.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/greater-less-than-operation.bal
@@ -24,6 +24,31 @@ function testFloatRanges(float a) returns (int) {
     return retunType;
 }
 
+function compareByteValues(byte lowValue, byte highValue) {
+    test:assertTrue(lowValue < highValue);
+    test:assertTrue(lowValue <= highValue);
+
+    test:assertTrue(highValue > lowValue);
+    test:assertTrue(highValue >= lowValue);
+
+    test:assertFalse(highValue < lowValue);
+    test:assertFalse(highValue <= lowValue);
+    test:assertFalse(lowValue > highValue);
+    test:assertFalse(lowValue >= highValue);
+
+    test:assertFalse(lowValue > lowValue);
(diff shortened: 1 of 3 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/greater-less-than-operation.bal
```diff
@@ -24,6 +24,27 @@ function testFloatRanges(float a) returns (int) {
     return retunType;
 }
 
+function testByteComparison() {
+    byte lowValue = 0x20;
```
warunalakshitha, line 28: Shall we same test with byte value with decimals.

Nadeeshan96: I ran the same test with byte value with decimals. It was fine. Do I need to add those lines here also.

warunalakshitha: I mean lets keep the hexa decimal value as well.

something like

function testByteComparison() {
compareByteValues(0x20, 0x21);
compareByteValues(32, 45);
compare ByteValues(0x20, 35);
}

compare ByteValues(lowValue, highValue ) {// your compare test}
}

Nadeeshan96: I will look into this.

warunalakshitha (approved):
LGTM

PR #31718, opened 2021-07-15
Fix completion item sorting within type descriptor contexts

### LINKED ISSUE #31357: Object Type descriptor context completions include false positives
**Description:**
$title.
In the particular context we suggest the private keyword, which is not allowed.
```
public type ObjectType isolated service object {
    <cursor>
};
```

**Affected Versions:**
swanlake beta1 at least

### LINKED ISSUE #31699: No completions are provided when cursor is above an enum declaration
**Description:**
$title

```
import ballerina/http;
import ballerina/io;

    resource function get getResource(http:Caller caller, http:Request req) {
        
    }
}

<cursor>

enum enumName {
    ENUM_FIELD
}

public function main() {
    io:println("Hello world");
}
```

**Steps to reproduce:**
Try to invoke the completions at the given cursor position.

**Affected Versions:**
swanlake beta2 at least

nadeeshaan:
## Purpose
With this,
Generalise the completion item sorting within the map type descriptor
Generalise the completion item sorting within the tuple type descriptor

Fixes #31699
Fixes #31357

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SnippetGenerator.java
```diff
@@ -35,7 +35,7 @@
  */
 public class SnippetGenerator {
 
-    private static final String FILTER_TEXT_SEPARATOR = "|";
+    private static final String FILTER_TEXT_SEPARATOR = "_";
```
malinthar, line 38: Just wanted to know why have we removed "|".

malinthar (approved):
LGTM

IMS94 (approved):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SnippetGenerator.java
nadeeshaan: That is because, if we have non-word symbols, then the completion's filter text will honour only the first component. for example, consider,
`service|http` as the filter text. Then, even when we are trying to type `http`, the http service snippet should get filtered. When we are having the pipe in between, for the `http` part it won't show the completion.

In order to make this happen, we have to separate these components with a word character. so went with the `_`

PR #31723, opened 2021-07-15
Fix Error Message for Accessing Incorrect Field Name of Closed Records

### LINKED ISSUE #30873: Misleading error message for incorrect field name of closed record
In beta1 rc5, this

```
type R record {|
  int x;
|};

function foo(R r) returns int {
  return r.y; // wrong field name
}
```

gets the error message: ` type 'R' does not support field access for non-required field 'y'`.

The error message should be ` type 'R' does not have a field 'y'`.

I run into this ALL the time.

sanjana:
Fixes #30873

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7031,6 +7045,27 @@ private BType checkFieldAccessExpr(BLangFieldBasedAccess fieldAccessExpr, BType
         return actualType;
     }
 
+    private boolean checkRecordFieldExistence(BRecordType recordVarRefType, Name fieldName) {
```
rdhananjaya, line 7048: Shall we re-name this to something like `isRecordFieldExists` or something?

rdhananjaya:
I think we need to rethink the error message.
```ballerina
type R record {
    int i;
};

public function main() {
    R r = {i: 0, "j": 22};
    anydata jr = r["j"];
    anydata ir = r["i"]; 
    anydata j = r.j; // undefined field 'j' in record 'R' // this is wrong: this field is defined.
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6970,6 +6970,20 @@ private BType checkFieldAccessExpr(BLangFieldBasedAccess fieldAccessExpr, BType
...
+                dlog.error(fieldAccessExpr.pos, DiagnosticErrorCode.UNDEFINED_FIELD_IN_RECORD, fieldName, varRefType);
+                return actualType;
+            }
+
+            // Check if the field accessed is present in the fields of a record which is a member of a union type.
+            if (varRefType.tag == TypeTags.UNION &&
+                    !checkFieldExistenceInUnionRecordMembers(((BUnionType) varRefType).getMemberTypes(), fieldName)) {
+                dlog.error(fieldAccessExpr.pos, DiagnosticErrorCode.UNDEFINED_FIELD_IN_RECORD, fieldName, varRefType);
+                return actualType;
+            }
```
pubudu91, line 6985: Is this the best way to handle this? These 2 fixes placed here seem rather ad-hoc. The logic related to this is in the `checkRecordFieldAccessExpr()` method. Shouldn't this logic be incorporated somewhere in that method?

MaryamZi: I also think we can improve how we handle this. But `checkRecordFieldAccessExpr` doesn't log errors. I think we can let this go on the original path and handle the errors from L6994 onward.

If we do it this way we'll be doing some checks unnecessarily. For example where the field is valid, both the introduced checks and  `checkRecordFieldAccessExpr` will check for existence.

MaryamZi: Basically if `!fieldAccessExpr.isLValue` iin L6994, do these checks and use the relevant message.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7031,6 +7045,27 @@ private BType checkFieldAccessExpr(BLangFieldBasedAccess fieldAccessExpr, BType
...
     }
 
+    private boolean checkRecordFieldExistence(BRecordType recordVarRefType, Name fieldName) {
+        return recordVarRefType.fields.containsKey(fieldName.getValue());
+    }
+
+    private boolean checkFieldExistenceInUnionRecordMembers(LinkedHashSet<BType> memberTypes, Name fieldName) {
+        boolean membersHaveRecordsTypes = false;
+        for (BType memberType : memberTypes) {
+            if (memberType.tag == TypeTags.RECORD) {
```
MaryamZi, line 7055: Given that for the record tag, isSubTypeOfBaseType specifically checks for records, does this become false for any case atm?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/access/FieldAccessTest.java
```diff
@@ -52,8 +52,7 @@ public void testNegativeCases() {
         int i = 0;
         validateError(negativeResult, i++, "invalid operation: type 'Employee' does not support field access " +
                               "for non-required field 'id'", 32, 9);
-        validateError(negativeResult, i++, "invalid operation: type 'Employee' does not support field access " +
-                              "for non-required field 'salary'", 33, 9);
+        validateError(negativeResult, i++, "undefined field 'salary' in record 'Employee'", 33, 9);
```
MaryamZi, line 55: Shall we add negative tests for unions if we don't have them already.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7031,6 +7045,27 @@ private BType checkFieldAccessExpr(BLangFieldBasedAccess fieldAccessExpr, BType
         return actualType;
     }
 
+    private boolean checkRecordFieldExistence(BRecordType recordVarRefType, Name fieldName) {
+        return recordVarRefType.fields.containsKey(fieldName.getValue());
```
MaryamZi, line 7049: Do we need a method for this? IMO we can do this in-line since it's straightforward, self-explanatory, and the only place this is used.

sanjana:
The types of scenarios where the Record Field Access Expression may log diagnostics, are listed down here: 
https://docs.google.com/document/d/16tQz2zM3BwU2RKt3i_jSD1puLET-rDDcS1ocLtiiZBo/edit?usp=sharing 

The solution would align accordingly.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
sanjana: Updated the approach.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
sanjana: Updated.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/access/FieldAccessTest.java
sanjana: Added negative test cases for union of records as well.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
sanjana: Removed with the new approach.

sanjana:
> I think we need to rethink the error message.
> 
> ```ballerina
> type R record {
>     int i;
> };
> 
> public function main() {
>     R r = {i: 0, "j": 22};
>     anydata jr = r["j"];
>     anydata ir = r["i"]; 
>     anydata j = r.j; // undefined field 'j' in record 'R' // this is wrong: this field is defined.
> }
> ```

@rdhananjaya 

With the new approach, the error message for this would be: `invalid field access: 'j' is not a required field in record 'R', use member access to access a field that may have been specified as a rest field`.

rdhananjaya (approved):

PR #31730, opened 2021-07-15
Add a default return value when generating a function with `Implement method` code action

### LINKED ISSUE #31608: Add a default return value when generating a function with `implement method code action`
**Description:**

When a function with a return value is created using `Implement method`/`Implement all` code actions, there is a semantic error in the generated code segment.

Screenshot from 2021-07-07 10-49-11

Add a default return value to get rid of the semantic error.

Screenshot from 2021-07-07 10-49-05

**Steps to reproduce:**

```
type Foo object {
    function abc() returns int;
};

object {} x = object Foo {

                  function abc() returns int {
                  }
};
```

**Affected Versions:**

TharushiJay:
## Purpose
When a function with a return value is created using `Implement method`/`Implement all` code actions, there is a semantic error in the generated code segment. This PR adds a default return value to avoid the semantic error.

Fixes #31608

## Samples
Screenshot from 2021-07-15 15-57-01

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/AbstractImplementMethodCodeAction.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/AbstractImplementMethodCodeAction.java
@@ -158,8 +158,7 @@ public abstract class AbstractImplementMethodCodeAction extends AbstractCodeActi
         String returnStmt = "";
         if (unimplMethod.typeDescriptor().returnTypeDescriptor().isPresent()) {
             TypeSymbol returnTypeSymbol = unimplMethod.typeDescriptor().returnTypeDescriptor().get();
-            Optional<String> returnType = FunctionGenerator.getReturnTypeAsString(context, returnTypeSymbol);
-            if (returnType.isPresent()) {
+            if (returnTypeSymbol.typeKind() != TypeDescKind.COMPILATION_ERROR) {
                 returnStmt = "return " + CommonUtil.getDefaultValueForType(returnTypeSymbol) +
                         CommonKeys.SEMI_COLON_SYMBOL_KEY;
             }
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/AbstractImplementMethodCodeAction.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/AbstractImplementMethodCodeAction.java
@@ -30,6 +30,7 @@ import io.ballerina.tools.diagnostics.Diagnostic;
 import io.ballerina.tools.text.LinePosition;
 import org.apache.commons.lang3.StringUtils;
 import org.ballerinalang.langserver.common.ImportsAcceptor;
+import org.ballerinalang.langserver.common.utils.CommonKeys;
 import org.ballerinalang.langserver.common.utils.CommonUtil;
 import org.ballerinalang.langserver.common.utils.FunctionGenerator;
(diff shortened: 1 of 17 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/AbstractImplementMethodCodeAction.java
```diff
@@ -153,9 +154,33 @@ public AbstractImplementMethodCodeAction(List<CodeActionNodeType> nodeTypes) {
         ImportsAcceptor importsAcceptor = new ImportsAcceptor(context);
         String typeName = FunctionGenerator.processModuleIDsInText(importsAcceptor, unimplMethod.signature(), context);
         List<TextEdit> edits = new ArrayList<>(importsAcceptor.getNewImportTextEdits());
-        String editText = offsetStr + typeName + " {" + LINE_SEPARATOR + offsetStr + "}" + LINE_SEPARATOR;
+
+        String returnStmt = "";
+        if (unimplMethod.typeDescriptor().returnTypeDescriptor().isPresent()) {
+            TypeSymbol returnTypeSymbol = unimplMethod.typeDescriptor().returnTypeDescriptor().get();
+            Optional<String> returnType = FunctionGenerator.getReturnTypeAsString(context, returnTypeSymbol);
```
IMS94, line 161: What if we simply do 
```
returnStmt = "return " + CommonUtil.getDefaultValueForType(returnTypeSymbol) +
                        CommonKeys.SEMI_COLON_SYMBOL_KEY;
```
without invoking `returnTypeAsString()` here? Only use case of invoking that is to avoid compilation error type symbols. WDYT?

IMS94 (commented):
LGTM! There's a small comment on an alternative approach.

nadeeshaan:
Regarding the generated return statement, what is the default return statement for the optional types? For example, as shown in the image reference, we have added `return error("");` for `error?` as the return type. If we have an optional as the return type, let's not add the return statement.

TharushiJay:
> Regarding the generated return statement, what is the default return statement for the optional types? For example, as shown in the image reference, we have added `return error("");` for `error?` as the return type. If we have an optional as the return type, let's not add the return statement.

@nadeeshaan As per the current implementation in `CommonUtil.getDefaultValueForType()`, when there is a union of types, the default value is obtained as the value of its first `memberTypeDescriptor` where index is 0. This causes `string|()` to return `“”` and `()|string` to return `()`. Therefore as in the image reference, considering return type `error?` to be the union of `error` & `()`, it returns `error(“”)` .
Whenever there is a nil type descriptor in a union type, would you recommend not adding the return statement? Or would that behavior introduce an inconsistency with the existing implementation of obtaining the default value for a given type? WDYT?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/AbstractImplementMethodCodeAction.java
TharushiJay: +1

TharushiJay:
> Regarding the generated return statement, what is the default return statement for the optional types? For example, as shown in the image reference, we have added `return error("");` for `error?` as the return type. If we have an optional as the return type, let's not add the return statement.

Will be tracked under https://github.com/ballerina-platform/ballerina-lang/issues/31790

IMS94 (approved):

PR #31772, opened 2021-07-19
Change implementation of outStream for Central API Client functions

### LINKED ISSUE #31711: `CentralAPIClient` writing to stdout breaks language server in online mode
**Description:**
When the `CentralAPIClient` is pulling a module, it prints the success message (and progress) to stdout. Since language server and vscode language client communicates over the `stdin/stdout` using json RPC, above behavior cause non-json content to be written into stdout. This breaks the language server protocol and cause LS to not work there onward.

Changing vscoe - LS to communicate over a different stream is a large refactoring + any component inside LS (including project API, compiler, etc) should not be writing to `stdout` when they are invoked via the LS.

Therefore, need a way to change the `outputStream` in `CentralAPIClient` for LS use cases.

**Steps to reproduce:**
See description

**Affected Versions:**
SL Beta 3 SNAPSHOT

#30757

Ibaqu:
## Purpose
Add option to disable 'Printing to console' for Central API Client functions.

Inorder to fix the following issue :

> When the `CentralAPIClient` is pulling a module, it prints the success message (and progress) to stdout. Since language server and vscode language client communicates over the stdin/stdout using JSON RPC, above behavior cause non-json content to be written into stdout. This breaks the language server protocol and causes LS to not work there onward.

Fixes #31711

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
@@ -26,6 +26,7 @@ import io.ballerina.projects.PackageCompilation;
 import io.ballerina.projects.Project;
 import io.ballerina.projects.ProjectException;
 import io.ballerina.projects.directory.SingleFileProject;
+import org.ballerinalang.central.client.CentralClientConstants;
 
 import java.io.PrintStream;
 
@@ -60,6 +61,8 @@ public class CompileTask implements Task {
         }
         // Print the source
         this.out.println("\t" + sourceName);
+
+        System.setProperty(CentralClientConstants.ENABLE_OUTPUT_STREAM, "false");
 
         try {
             long start = 0;
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
@@ -62,7 +62,7 @@ public class CompileTask implements Task {
         // Print the source
         this.out.println("\t" + sourceName);
 
-        System.setProperty(CentralClientConstants.ENABLE_OUTPUT_STREAM, "false");
+        System.setProperty(CentralClientConstants.ENABLE_OUTPUT_STREAM, "true");
 
         try {
             long start = 0;
--- a/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
+++ b/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
@@ -352,8 +352,8 @@ public class CentralAPIClient {
     }
 
     public void pullPackage(String org, String name, String version, Path packagePathInBalaCache,
-                            String supportedPlatform, String ballerinaVersion, boolean isBuild) throws CentralClientException {
-
+                            String supportedPlatform, String ballerinaVersion, boolean isBuild)
+            throws CentralClientException {
         pullPackage(org, name, version, packagePathInBalaCache, supportedPlatform, ballerinaVersion, isBuild, false);
     }
 
--- a/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
+++ b/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
@@ -280,7 +280,7 @@ public class CentralAPIClient {
 
             // Successfully pushed
             if (packagePushResponse.code() == HTTP_NO_CONTENT) {
-                if (!enableOutputStream) {
+                if (enableOutputStream) {
                     this.outStream.println(packageSignature + " pushed to central successfully");
                 }
                 return;
--- a/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
+++ b/cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
@@ -243,7 +243,14 @@ public class CentralAPIClient {
     }
 
     /**
-     * Pushing a package to registry.
+     * Push a package to registry.
+     *
+     * @param balaPath              The path to the bala file.
+     * @param org                   The organization of the package.
+     * @param name                  The name of the package.
+     * @param version               The version of the package.
+     * @param supportedPlatform     The supported platform.
+     * @param ballerinaVersion      The ballerina version.
      */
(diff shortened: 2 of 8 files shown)


Ibaqu:
Push package and Pull package which makes use of outStream has been modified as such

A boolean `disableOutStream` has been added which will disable all prints to the console.

```
public void pushPackage(Path balaPath, String org, String name, String version, String supportedPlatform,
                            String ballerinaVersion, boolean disableOutStream)
                            
public void pullPackage(String org, String name, String version, Path packagePathInBalaCache,
                            String supportedPlatform, String ballerinaVersion, boolean isBuild,
                            boolean disableOutStream)                            
```

Encased all out.println() with a check to prevent it from being printed. Both functions are overloaded with the defaults as `false` to prevent other features from breaking

cli/central-client/src/main/java/org/ballerinalang/central/client/Utils.java
```diff
@@ -274,6 +280,23 @@ private static void writeAndHandleProgress(InputStream inputStream, FileOutputSt
...
+        byte[] buffer = new byte[1024];
+
+        try {
+            while ((count = inputStream.read(buffer)) > 0) {
+                outputStream.write(buffer, 0, count);
+            }
+        } catch (IOException e) {
+            PrintStream printStream = System.out;
+
+            printStream.println(logFormatter.formatLog(fullPkgName + "pulling the package from central failed"));
```
hevayo, line 295: At this level this should be an exception and can be handled at a higher level. Then the exception can be converted to a diagnostic at the Package resolution

cli/central-client/src/main/java/org/ballerinalang/central/client/Utils.java
```diff
@@ -274,6 +280,23 @@ private static void writeAndHandleProgress(InputStream inputStream, FileOutputSt
...
+
+        }
+    }
+
     /**
      * Create nightly build meta file.
      *
```
hevayo, line 302: Can we add some tests how can you disable printing when using ProjectAPI ?

hevayo (changes requested):

cli/central-client/src/main/java/org/ballerinalang/central/client/Utils.java
```diff
@@ -274,6 +278,17 @@ private static void writeAndHandleProgress(InputStream inputStream, FileOutputSt
         }
     }
 
+    private static void writeAndHandleProgressWithoutLogging(InputStream inputStream, FileOutputStream outputStream)
+            throws IOException {
+        int count;
+        byte[] buffer = new byte[1024];
+
+        while ((count = inputStream.read(buffer)) > 0) {
+            outputStream.write(buffer, 0, count);
```
azinneera, line 287: We should not write to the output stream. The idea is that we do the pull operation quietly. If the pull fails due to any interruption, then we should throw an exception and log it as a diagnostic.

cli/central-client/src/main/java/org/ballerinalang/central/client/Utils.java
Ibaqu: Added tests. Will add tests to distribution as well.

Ibaqu:
Changes : 

- Change `DISABLE_OUTPUT_STREAM` -> `ENABLE_OUTPUT_STREAM` 
- Added `System.setProperty(CentralClientConstants.ENABLE_OUTPUT_STREAM, "false");` to `CompileTask` disabling output stream for all commands
- Added tests

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
```diff
@@ -61,6 +62,8 @@ public void execute(Project project) {
         // Print the source
         this.out.println("\t" + sourceName);
 
+        System.setProperty(CentralClientConstants.ENABLE_OUTPUT_STREAM, "false");
```
azinneera, line 65: The progress bar should show in the stdout in the case of commands.

Ibaqu: Changed to true as discussed

cli/central-client/src/main/java/org/ballerinalang/central/client/CentralAPIClient.java
```diff
@@ -342,8 +347,14 @@ public void pushPackage(Path balaPath, String org, String name, String version,
         }
     }
 
+    /**
+     * Pulling a package.
```
azinneera, line 351: Shall we refactor the method comment to use proper tense and add the parameters and exceptions thrown?

Ibaqu: Addressed in https://github.com/ballerina-platform/ballerina-lang/commit/a26d8a97e757d634eba17907656536dfe134e2d5

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
```diff
@@ -61,6 +62,8 @@ public void execute(Project project) {
         // Print the source
         this.out.println("\t" + sourceName);
 
+        System.setProperty(CentralClientConstants.ENABLE_OUTPUT_STREAM, "true");
```
azinneera, line 65: We need to set this in `PullCommand` and `PushCommand` as well

cli/central-client/src/test/java/org/ballerinalang/central/client/TestCentralApiClient.java
```diff
@@ -526,4 +527,91 @@ private void setBallerinaHome() {
             System.setProperty(ballerinaInstallDirProp, String.valueOf(Paths.get("build")));
         }
     }
+
+    @Test(description = "Test pull package without Logs", enabled = false)
```
azinneera, line 531: We can move these tests to distribution without having it disabled here.

Ibaqu: Removed disabled tests

Ibaqu:
Test : https://github.com/ballerina-platform/ballerina-distribution/pull/2070

azinneera (approved):

hevayo (approved):

PR #31779, opened 2021-07-19
Fix issues with tuple type conversion

### LINKED ISSUE #31747: `cloneWithType` does not work with Tuple types properly
**Description:**
Conversion with `cloneWithType` is not working as expected for Tuple type conversions. 

**Steps to reproduce:**
Following should succeed but currently fails. 
```
type T1 record {
    int x = 1;
    int y = 2;
};
type T2 record {|
    int x = 5;
    int y = 6;
    int z = 7;
|};
type T1_T2 [T1, T2]; 
public function main() {
    T1 a = {};
    T1 b = {};
    [T1, T1] x = [a, b];
    anydata y = x;
    var result = y.cloneWithType(T1_T2);
    io:println(result);
}
```
**Affected Versions:**

Team/jBallerina

gabilang:
Fixes #31747

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
@@ -455,14 +455,9 @@ public class TypeConverter {
                 return true;
             }
         }
-        Set<Type> convertibleTypes;
+
         for (int i = 0; i < source.size(); i++) {
-            convertibleTypes = getConvertibleTypes(source.get(i), targetTypeElementType, unresolvedValues);
-            if (convertibleTypes.isEmpty()) {
-                return false;
-            }
-            if (convertibleTypes.size() != 1 && !convertibleTypes.contains(TypeChecker.getType(source.get(i)))
-                    && !hasIntegerSubTypes(convertibleTypes)) {
+            if (!isConvertibleToArrayInstance(source.get(i), targetTypeElementType, unresolvedValues)) {
                 return false;
             }
         }
@@ -488,28 +483,29 @@ public class TypeConverter {
             return false;
         }
 
-        Set<Type> convertibleTypes;
-
         for (int i = 0; i < targetTypeSize; i++) {
-            convertibleTypes = getConvertibleTypes(source.getRefValue(i), targetTypes.get(i), unresolvedValues);
-            if (convertibleTypes.isEmpty()) {
-                return false;
-            }
-            if (convertibleTypes.size() > 1 && !convertibleTypes.contains(TypeChecker.getType(source.getRefValue(i)))
-                    && !hasIntegerSubTypes(convertibleTypes)) {
+            if (!isConvertibleToArrayInstance(source.getRefValue(i), targetTypes.get(i), unresolvedValues)) {
                 return false;
             }
         }
 
         for (int i = targetTypeSize; i < sourceTypeSize; i++) {
-            convertibleTypes = getConvertibleTypes(source.getRefValue(i), targetRestType, unresolvedValues);
-            if (convertibleTypes.isEmpty()) {
+            if (!isConvertibleToArrayInstance(source.getRefValue(i), targetRestType, unresolvedValues)) {
                 return false;
             }
-            if (convertibleTypes.size() > 1 && !convertibleTypes.contains(TypeChecker.getType(source.getRefValue(i)))
-                    && !hasIntegerSubTypes(convertibleTypes)) {
-                return false;
-            }
+        }
+        return true;
+    }
+
+    private static boolean isConvertibleToArrayInstance(Object sourceElement, Type targetType,
+                                                        List<TypeValuePair> unresolvedValues) {
+        Set<Type> convertibleTypes = getConvertibleTypes(sourceElement, targetType, unresolvedValues);
+        if (convertibleTypes.isEmpty()) {
+            return false;
+        }
+        if (convertibleTypes.size() > 1 && !convertibleTypes.contains(TypeChecker.getType(sourceElement))
+                && !hasIntegerSubTypes(convertibleTypes)) {
+            return false;
         }
         return true;
     }
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
@@ -38,6 +38,7 @@ import io.ballerina.runtime.internal.types.BIntersectionType;
 import io.ballerina.runtime.internal.types.BMapType;
 import io.ballerina.runtime.internal.types.BRecordType;
 import io.ballerina.runtime.internal.types.BTableType;
+import io.ballerina.runtime.internal.types.BTupleType;
 import io.ballerina.runtime.internal.types.BUnionType;
 import io.ballerina.runtime.internal.util.exceptions.BLangExceptionHelper;
 import io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons;
@@ -264,6 +265,11 @@ public class TypeConverter {
                     convertibleTypes.add(targetType);
                 }
                 break;
+            case TypeTags.TUPLE_TAG:
+                if (isConvertibleToTupleType(inputValue, (BTupleType) targetType, unresolvedValues)) {
+                    convertibleTypes.add(targetType);
+                }
+                break;
             case TypeTags.RECORD_TYPE_TAG:
                 if (isConvertibleToRecordType(inputValue, (BRecordType) targetType, false, unresolvedValues)) {
                     convertibleTypes.add(targetType);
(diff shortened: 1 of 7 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -463,6 +469,51 @@ private static boolean isConvertibleToArrayType(Object sourceValue, BArrayType t
...
+        for (int i = 0; i < targetTypeSize; i++) {
+            convertibleTypes = getConvertibleTypes(source.getRefValue(i), targetTypes.get(i), unresolvedValues);
+            if (convertibleTypes.isEmpty()) {
+                return false;
+            }
+            if (convertibleTypes.size() > 1 && !convertibleTypes.contains(TypeChecker.getType(source.getRefValue(i)))
+                    && !hasIntegerSubTypes(convertibleTypes)) {
+                return false;
+            }
+        }
```
HindujaB, line 502: Shall we move this logic to a separate method?
Since we the only difference is the `targetTypes.get(i)` part, we can pass it as a parameter

tests/jballerina-unit-test/src/test/resources/test-src/expressions/conversion/native-conversion.bal
```diff
@@ -855,14 +857,22 @@ public class O1 {
   public int y = 0;
 }
 
-function testTupleConversion1() returns [T1, T1]|error {
+function testTupleConversion1() {
     T1 a = {};
     T2 b = {};
-    [T1, T2] x = [a, b];
+    [T1, T2] x1 = [a, b];
```
HindujaB, line 863: Shall we use the already defined types `T1_T1` & `T1_T2` here and similar cases?
We can re-use the type.

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -919,6 +919,74 @@ function testCloneWithTypeIntArrayToUnionArray() {
...
+
+    [float|decimal, int|byte...]|error e = arr.cloneWithType();
+    assert(e is error, true);
+    err = <error> e;
+    message = err.detail()["message"];
+    messageString = message is error ? message.toString() : message.toString();
+    assert(err.message(), "{ballerina/lang.typedesc}ConversionError");
+    assert(messageString, "'int[]' value cannot be converted to '[(float|decimal),(int|byte)...]'");
+
+    [string, string:Char, string|string:Char]|error f = arr.cloneWithType();
```
HindujaB, line 955: Shall we also have some tests with structured type as a member type?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -463,6 +469,51 @@ private static boolean isConvertibleToArrayType(Object sourceValue, BArrayType t
...
+
+        for (int i = 0; i < targetTypeSize; i++) {
+            convertibleTypes = getConvertibleTypes(source.getRefValue(i), targetTypes.get(i), unresolvedValues);
+            if (convertibleTypes.isEmpty()) {
+                return false;
+            }
+            if (convertibleTypes.size() > 1 && !convertibleTypes.contains(TypeChecker.getType(source.getRefValue(i)))
+                    && !hasIntegerSubTypes(convertibleTypes)) {
+                return false;
+            }
```
riyafa, line 501: we can combine the if the two if statements using an `||` and move the boolean checks to a separate method

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -463,6 +469,51 @@ private static boolean isConvertibleToArrayType(Object sourceValue, BArrayType t
         return true;
     }
 
+    private static boolean isConvertibleToTupleType(Object sourceValue, BTupleType targetType,
```
riyafa, line 472: Most of the code in this function seems to be similar to the previous function `isConvertibleToArrayType`. Shall we try to reduce duplication by reusing code?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/conversion/native-conversion.bal
gabilang: Addressed with https://github.com/ballerina-platform/ballerina-lang/pull/31779/commits/93a63a8b643f9e14f56219f8cc2aba8a54280939

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
gabilang: Addressed with https://github.com/ballerina-platform/ballerina-lang/pull/31779/commits/93a63a8b643f9e14f56219f8cc2aba8a54280939

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
gabilang: Following part has been moved to a separate method where we can reuse the code for the method `isConvertibleToArrayType` also. 
```
convertibleTypes = getConvertibleTypes(source.getRefValue(i), targetTypes.get(i), unresolvedValues);
if (convertibleTypes.isEmpty()) {
    return false;
}
if (convertibleTypes.size() > 1 && !convertibleTypes.contains(TypeChecker.getType(source.getRefValue(i)))
        && !hasIntegerSubTypes(convertibleTypes)) {
    return false;
}
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
gabilang: The both `if` checks have been moved to a separate method and `convertibleTypes.isEmpty()` has been kept as an early separate check to prevent generating multiple instructions on this particular failing case.

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -919,6 +919,82 @@ function testCloneWithTypeIntArrayToUnionArray() {
     assert(messageString, "'float[]' value cannot be converted to '(lang.int:Signed16|lang.int:Unsigned8|decimal)[]'");
 }
 
+function testCloneWithTypeArrayToTuple() {
```
riyafa, line 922: can we split this test to multiple functions so that it is more readable? We can make the array  `arr` to be a global variable if needed. Also it would be better if we can move the newly created tests to a separate file as this file seems to have become really long

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -919,6 +919,82 @@ function testCloneWithTypeIntArrayToUnionArray() {
...
+    [map<int>, [string, int]...]|error g = arr.cloneWithType();
+    assert(g is error, true);
+    err = <error> g;
+    message = err.detail()["message"];
+    messageString = message is error ? message.toString() : message.toString();
+    assert(err.message(), "{ballerina/lang.typedesc}ConversionError");
+    assert(messageString, "'int[]' value cannot be converted to '[map<int>,[string,int]...]'");
+}
+
+function testCloneWithTypeTuple() {
```
riyafa, line 972: We can split this function also.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -449,20 +455,61 @@ private static boolean isConvertibleToArrayType(Object sourceValue, BArrayType t
...
+    private static boolean isConvertibleToArrayInstance(Object sourceElement, Type targetType,
+                                                        List<TypeValuePair> unresolvedValues) {
+        Set<Type> convertibleTypes = getConvertibleTypes(sourceElement, targetType, unresolvedValues);
+        if (convertibleTypes.isEmpty()) {
+            return false;
+        }
+        if (convertibleTypes.size() > 1 && !convertibleTypes.contains(TypeChecker.getType(sourceElement))
+                && !hasIntegerSubTypes(convertibleTypes)) {
+            return false;
+        }
```
riyafa, line 509: I think the following is possible
```suggestion
        if (convertibleTypes.isEmpty() || (convertibleTypes.size() > 1 && !convertibleTypes.contains(TypeChecker.getType(sourceElement))
                && !hasIntegerSubTypes(convertibleTypes))) {
            return false;
        }
```
We can move the boolean checks to a separate function. I mean what is inside the brackets of the `if` statement

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -449,20 +455,61 @@ private static boolean isConvertibleToArrayType(Object sourceValue, BArrayType t
...
+        int sourceTypeSize = source.size();
+        int targetTypeSize = targetTypes.size();
+        Type targetRestType = targetType.getRestType();
+
+        if (sourceTypeSize < targetTypeSize) {
+            return false;
+        }
+        if (targetRestType == null && sourceTypeSize > targetTypeSize) {
+            return false;
+        }
```
riyafa, line 484: we can combine these if statements as well:
```suggestion
        if (sourceTypeSize < targetTypeSize || (targetRestType == null && sourceTypeSize > targetTypeSize)) {
            return false;
        }
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
gabilang: Yeah, it is possible, but we are going to create too many instructions for a failing case right? I was just thinking about for an early return when the convertible types set is empty. WDYT?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -449,20 +455,62 @@ private static boolean isConvertibleToArrayType(Object sourceValue, BArrayType t
...
+            if (!isConvertibleToArrayInstance(source.get(i), targetTypeElementType, unresolvedValues)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private static boolean isConvertibleToTupleType(Object sourceValue, BTupleType targetType,
+                                                    List<TypeValuePair> unresolvedValues) {
+        if (!(sourceValue instanceof ArrayValue)) {
```
warunalakshitha, line 469: Do we need this line. I think it will be always ArrayValue

gabilang: When the target type is tuple type, the execution will reach to this place and we can't guarantee the source type here is always an instance of `ArrayValue`. This check will ensure the success of `Object` to `ArrayValue` casting. So, IMO we need this check here.

warunalakshitha (approved):
LGTM

PR #31882, opened 2021-07-26
Fix string:char iterator for large codepoints

### LINKED ISSUE #31767: Iterating over strings using `foreach` loop for uni-code characters
**Description:**
Iterating over strings using `foreach` loop will lead to different results for certain uni-code characters.

**Steps to reproduce:**
```
import ballerina/io;

string test = "￿"; // utf-8 codepoint `EFBFBF`
string test2 = "𐀀"; // utf-8 codepoint `F0908080`

function loop1(string s) {
    foreach var char in s {
        io:println("$", char);
    }
}

function loop2(string s) {
    foreach int i in 0 ..< s.length() {
        string char = s[i];
        io:println("#", char);
    }
}

public function main() {
	loop1(test);
	loop2(test);
	io:println("------");
	loop1(test2);
	loop2(test2);
}
```
Result from `loop2` and `loop1` are different
**Affected Versions:**

### LINKED ISSUE #31775: toCodePointInts() and fromCodePointInt() have different outputs
Consider the following code

```ballerina
    string str  = check string:fromCodePointInt(0x10FFFF);
    foreach var ch in str {
        io:println(ch.toCodePointInts());
    }
}
```
Output is as followed.

```
[56319]
[57343]
```

### LINKED ISSUE #31917: String concat() method for non-bmp strings provide wrong surrogates
**Description:**
This is due to a wrong array copy happening in `concat()` method for non-bmp string value.
https://github.com/ballerina-platform/ballerina-lang/blob/faacd2f4a7ce041fd6e9904e7d660f7ec222f007/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/NonBmpStringValue.java#L76

**Steps to reproduce:**
```ballerina
string str = "a☃😀";
str = str + "bc😀";
```

When initialising the string,

error-surrogate

Note that the surrogates are given as `[2,2]` where they should be `[2, 5]` in the resulting non-bmp string.

This applies for the bmp string + non-bmp string case as well.

**Affected Versions:**
beta2

ubuntu 20.04, java 11

HindujaB:
## Purpose
$subject

Fixes #31775
Fixes #31767
Fixes #31917

## Approach
Use a natively implemnted character iterator instead of `StringCharacterIterator` which does not consider the 4 byte encoded unicode values.

## Samples
```ballerina
    string str  = check string:fromCodePointInt(0x10FFFF);
    foreach var ch in str {
        io:println(ch.toCodePointInts());
    }
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/CharIterator.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/CharIterator.java
@@ -1,0 +1,70 @@
+/*
+ *  Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  WSO2 Inc. licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+package io.ballerina.runtime.internal.values;
+
+import io.ballerina.runtime.api.values.BString;
+
+/**
+ * {@code {@link CharIterator }} provides iterator implementation for Ballerina string values.
+ *
+ * @since 2.0
+ */
+public class CharIterator implements IteratorValue {
+
+    BString value;
+    boolean isNonBmp;
+    long cursor = 0;
+    long length;
+    String stringValue;
+
+    CharIterator(BString value) {
+        this.value = value;
+        this.length = value.length();
+        this.isNonBmp = value instanceof NonBmpStringValue;
+        this.stringValue = value.getValue();
+    }
+
+    @Override
+    public Object next() {
+        long cursor = this.cursor++;
+        if (cursor == length) {
+            return null;
+        }
+        if (isNonBmp) {
+            int offset = (int) cursor;
+            for (int surrogate : ((NonBmpStringValue) value).getSurrogates()) {
+                if (surrogate < cursor) {
+                    offset++;
+                } else if (surrogate > cursor) {
+                    break;
+                } else {
+                    return new String(new char[]{stringValue.charAt(offset), stringValue.charAt(offset + 1)});
+                }
+            }
+            return String.valueOf(stringValue.charAt(offset));
+        }
+        return String.valueOf(stringValue.charAt((int) cursor));
+    }
+
+    @Override
+    public boolean hasNext() {
+        return cursor < length;
+    }
+
+}
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/CharIterator.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/CharIterator.java
@@ -18,8 +18,6 @@
 
 package io.ballerina.runtime.internal.values;
 
-import io.ballerina.runtime.api.values.BString;
-
 /**
  * {@code {@link CharIterator }} provides iterator implementation for Ballerina string values.
  *
@@ -27,12 +25,12 @@ import io.ballerina.runtime.api.values.BString;
  */
 public class CharIterator implements IteratorValue {
 
-    BString value;
+    StringValue value;
     long cursor = 0;
     long length;
     String stringValue;
 
-    CharIterator(BString value) {
+    CharIterator(StringValue value) {
         this.value = value;
         this.length = value.length();
         this.stringValue = value.getValue();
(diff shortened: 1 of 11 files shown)


langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/Next.java
```diff
@@ -35,24 +33,30 @@
...
-            String s = ((BString) m.get(StringUtils.fromString("m"))).getValue();
-            stringCharacterIterator = new StringCharacterIterator(s);
-            m.addNativeData("&iterator&", stringCharacterIterator);
+        String s = ((BString) m.get(StringUtils.fromString("m"))).getValue();
+        BreakIterator breakIterator = (BreakIterator) m.getNativeData("&iterator&");
+        if (breakIterator == null) {
+            breakIterator = BreakIterator.getCharacterInstance();
+            breakIterator.setText(s);
+            m.addNativeData("&iterator&", breakIterator);
         }
```
riyafa, line 47: Shall we have meaningful names for `m` and `s`?

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/Next.java
```diff
@@ -35,24 +33,30 @@
...
+    }
+
     //TODO: refactor hard coded values
     public static Object next(BObject m) {
-        StringCharacterIterator stringCharacterIterator = (StringCharacterIterator) m.getNativeData("&iterator&");
-        if (stringCharacterIterator == null) {
-            String s = ((BString) m.get(StringUtils.fromString("m"))).getValue();
-            stringCharacterIterator = new StringCharacterIterator(s);
-            m.addNativeData("&iterator&", stringCharacterIterator);
+        String s = ((BString) m.get(StringUtils.fromString("m"))).getValue();
```
riyafa, line 41: we can move this call after the null check as earlier

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibStringTest.java
```diff
@@ -148,7 +147,7 @@ public void testGetCodepointNegative() {
     public void testToCodepointInts(String st1, int[] expected) {
         BValue[] args = {new BString(st1)};
         BValue[] returns = BRunUtil.invoke(compileResult, "testToCodepointInts", args);
-        assertEquals(((BValueArray) returns[0]).size(), expected.length);
+        assertEquals((returns[0]).size(), expected.length);
```
riyafa, line 150: unnecessary bracket
```suggestion
        assertEquals(returns[0].size(), expected.length);
```

HindujaB:
Checked the performance of char iterator with bmp & non-bmp strings with 1 000 000 iterations. 
The results (execution time) are as follows
- For `java.text.BreakIterator` - 37.773180343s
- For the native iterator - 32.894702807s

Therefore, I'll be going ahead with the native iterator implementation.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/CharIterator.java
```diff
@@ -0,0 +1,70 @@
...
+    BString value;
+    boolean isNonBmp;
+    long cursor = 0;
+    long length;
+    String stringValue;
+
+    CharIterator(BString value) {
+        this.value = value;
+        this.length = value.length();
+        this.isNonBmp = value instanceof NonBmpStringValue;
```
warunalakshitha, line 39: Shall we creata String flag here. So BMPString =0  nonBMPString 1

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/CharIterator.java
```diff
@@ -0,0 +1,70 @@
...
+        this.stringValue = value.getValue();
+    }
+
+    @Override
+    public Object next() {
+        long cursor = this.cursor++;
+        if (cursor == length) {
+            return null;
+        }
+        if (isNonBmp) {
```
warunalakshitha, line 49: Shall we add a comment here.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/AbstractStringValue.java
```diff
@@ -0,0 +1,80 @@
...
+
+import io.ballerina.runtime.api.values.BLink;
+import io.ballerina.runtime.api.values.BString;
+
+/**
+ * Abstract class representing ballerina strings.
+ *
+ * @since 2.0
+ */
+public abstract class AbstractStringValue implements StringValue {
```
warunalakshitha, line 29: Why we need abastract class here. Can we moved to StringValue

warunalakshitha (commented):
LGTM

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/StringValue.java
```diff
@@ -28,20 +29,75 @@
  *
  * @since 1.0.5
  */
-public interface StringValue extends BString, SimpleValue {
+public abstract class StringValue implements BString, SimpleValue {
+
+    final String value;
+    final int nonBmpFlag;
```
riyafa, line 35: If we are using a flag, it is the best practice to use an `enum`. But since we have only two types we can use a `boolean` flag

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/CharIterator.java
```diff
@@ -0,0 +1,66 @@
...
+        this.value = value;
+        this.length = value.length();
+        this.stringValue = value.getValue();
+    }
+
+    @Override
+    public Object next() {
+        long currentIndex = this.cursor++;
+        // For non-bmp strings, we should consider the surrogate pairs to return non-bmp unicode characters
+        if (((StringValue) value).nonBmpFlag == 1) {
```
riyafa, line 45: If we are going to cast it to `StringValue`, then in the `CharIterator` we know that we are expecting a `StringValue`, so why not simply take the `StringValue` via the constructor?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/CharIterator.java
```diff
@@ -0,0 +1,66 @@
...
+    CharIterator(BString value) {
+        this.value = value;
+        this.length = value.length();
+        this.stringValue = value.getValue();
+    }
+
+    @Override
+    public Object next() {
+        long currentIndex = this.cursor++;
+        // For non-bmp strings, we should consider the surrogate pairs to return non-bmp unicode characters
```
riyafa, line 44: We can move what is inside the if statement to a separate method, in which case maybe the comment might not be necessary 🙂

riyafa (approved):

Issue #26924, opened 2020-11-14
Unit tests for runtime apis


warunalakshitha:
**Description:**
We have set of apis exposed by ballerina runtime. Even their internal already has been used by core and covered by unit tests, we not have unit tests for public interfaces.

PR #31900, opened 2021-07-27
Decode typename at compile time

riyafa:
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/creators/TypeCreator.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/creators/TypeCreator.java
@@ -502,4 +502,6 @@ public class TypeCreator {
         return new BFiniteType(typeName, values, typeFlags);
     }
 
+    private TypeCreator() {
+    }
 }
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/ValueUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/ValueUtils.java
@@ -109,7 +109,7 @@ public class ValueUtils {
      * @return value of the object.
      */
     public static BObject createObjectValue(Module packageId, String objectTypeName, Object... fieldValues) {
-        Strand currentStrand = getStrand();
+        Strand currentStrand = Scheduler.getStrandNoException();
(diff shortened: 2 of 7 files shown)


manuranga:
Are any of the encoded names visible to the uses? eg via stack trace, via typedesc, value creator? if that is the cases we need to keep encoded names in separate variables.

@HindujaB can you remember if these are kept un-encoded for a reason?

riyafa:
If I am not mistaken that is the reason(to avoid encoded name being displayed) the `typeName` was decoded at the `BType` constructor. This PR moves all the required decoding to the compile time.

riyafa:
The code coverage reduction is due to changing the source code of some of the public apis which are not covered by tests as in https://github.com/ballerina-platform/ballerina-lang/issues/26924

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/creators/TypeCreator.java
```diff
@@ -279,7 +281,7 @@ public static StreamType createStreamType(Type constraint, Type completionType)
      */
     public static StreamType createStreamType(String typeName, Type constraint,
                                               Type completionType, Module modulePath) {
-        return new BStreamType(typeName, constraint, completionType, modulePath);
+        return new BStreamType(IdentifierUtils.decodeIdentifier(typeName), constraint, completionType, modulePath);
```
warunalakshitha, line 284: Should we always expect decoded type name from api. Do we need this. In other places as well in TypeCreator?

HindujaB: I think we need this decoding if we are to remove it in BType constructor.
Because, we use the decoded names in all places other than the definitions in generated classes. Eg. call methods/ switch cases etc.
So it may cause issues if we don't use the decoded names as type names at runtime.

riyafa: I have removed these

warunalakshitha (approved):

PR #31909, opened 2021-07-27
Fix scope finding logic in `EnvironmentResolver`

pubudu91:
## Purpose
This PR fixes an issue we had in the `EnvironmentResolver` which caused the correct scope found to be overwritten.

## Approach
Added an additional check before proceeding to create the symbol env and setting it. This check basically checks whether the current node is a narrow scope than the narrowest scope found so far. If not, we skip visiting it. 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
@@ -435,8 +435,7 @@ public class EnvironmentResolver extends BaseVisitor {
 
     @Override
     public void visit(BLangQueryAction queryAction) {
-        if (!PositionUtil.withinRightInclusive(this.linePosition, queryAction.getPosition())
-                || !isNarrowerEnclosure(queryAction.getPosition())) {
+        if (!PositionUtil.withinRightInclusive(this.linePosition, queryAction.getPosition())) {
             return;
         }
         for (BLangNode clause : queryAction.queryClauseList) {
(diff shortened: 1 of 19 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/EnvironmentResolver.java
```diff
@@ -1333,4 +624,14 @@ private void acceptNode(BLangNode node, SymbolEnv env) {
...
+
+    private boolean isNarrowerEnclosure(Location nodePosition) {
+        // Have to special case the pkg node since its position is (0,0,0,0). Plus any other node would for sure will
+        // be a narrower enclosure than the package.
+        if (this.scope == null || this.scope.node.getKind() == NodeKind.PACKAGE) {
+            return true;
+        }
+
+        return PositionUtil.withinRange(nodePosition.lineRange(), this.scope.node.getPosition());
+    }
```
dulajdilshan, line 636: Seems like this is called with `PositionUtil.withInBlock` together many times, What if we add this inside `PositionUtil.withInBlock()` method ?

pubudu91: This check is specific to the env resolver though. If there's a requirement from other components as well for this, can consider moving it to `PositionUtil`.

mohanvive (approved):

PR #31933, opened 2021-07-28
Fix invalid token diagnostic locations

### LINKED ISSUE #24765: Incorrect position is given for an invalid node
**Description:**

Sample:
```ballerina
private function foo() {
}
```

The compiler error for the above is:
```
error: .::testMain.bal:1:9: invalid token 'private'
```

Note the column-position is given as `9`, where the actual position of the invalid token is `1`. This is the position of the valid node to which the invalid-node-minutiae is attached.

VSCode shows the error like this:

Screenshot 2020-07-15 at 18 37 17

---
Currently below tests are disabled due to this (in JBallerina-unit-tests)
- FunctionSignatureInBaloTest.testNegativeFunctionInvocations
- FunctionSignatureNegativeTest
- ClientObjectTest.testRemoteBasicsNegative
- ServiceTest.testServiceBasicsNegative
- WorkerFailTest.invalidUsagesOfDefault

lochana-chathura:
## Purpose
$subject.

Fixes #24765 

## Approach
Store invalid token diagnostics in the invalid token itself.

## Samples
Old invalid node diagnostic locations:

Screenshot from 2021-07-30 12-37-20

New invalid node diagnostic locations:

Screenshot from 2021-07-30 12-38-06

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -1970,8 +1970,8 @@ private STNode parseParamList(boolean isParamNameOptional) {
                     paramsList.add(param);
                 } else {
                     paramOrderErrorPresent = true;
-                    updateLastNodeInListWithInvalidNode(paramsList, paramEnd, paramOrderError);
-                    updateLastNodeInListWithInvalidNode(paramsList, param, null);
+                    updateLastNodeInListWithInvalidNode(paramsList, paramEnd, null);
```
rdulmina, line 1973: Not related to this PR bt it would be better if we can overload this method without `DiagnosticCode`.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/Token.java
```diff
@@ -82,9 +84,22 @@ public Token modify(MinutiaeList leadingMinutiae, MinutiaeList trailingMinutiae)
...
+
+    private List<Diagnostic> collectDiagnostics() {
+        List<Diagnostic> diagnosticList = new ArrayList<>();
+
+        // Collect diagnostics inside invalid token minutiae
+        leadingInvalidTokens().forEach(token -> token.diagnostics().forEach(diagnosticList::add));
+        trailingInvalidTokens().forEach(token -> token.diagnostics().forEach(diagnosticList::add));
+
+        // Collect token diagnostics
+        internalNode.diagnostics().stream()
```
rdulmina, line 98: Should we use forEach instead of the stream? Since we are now replacing stream with foreach

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
lochana-chathura: I think the reason for not having an overloaded method is, if anyone wants to skip the diagnostic error code, they have to explicitly give the parameter as null.
 
If we create an overloaded method, one could easily forget to add the diagnostics when updating with invalid nodes.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/Token.java
lochana-chathura: shall we stick the same way as the nonterminal node?
https://github.com/ballerina-platform/ballerina-lang/blob/a67edec2f67a3df1f966f4ae87396cd3aa2af817/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/NonTerminalNode.java#L305-L307

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
rdulmina: +1

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/Token.java
rdulmina: Should we update that as well then? :)

rdulmina (approved):
LGTM

PR #31973, opened 2021-07-31
Add locations of wrapped error causes to stack trace

### LINKED ISSUE #31554: Wrapped error's cause location is missing
**Description:**
``` ballerina
public function main() {
    panic bar();
}

function bar() returns error {
    error b = error("b", foo());
    return b;
}

function foo() returns error {
    error a = error("a");
    return a;
}
```
Above prints 
```
error: b cause: a
        at main:bar(main.bal:6)
           main:main(main.bal:2)
```

Cause happens in line 12 but it is missing in the printed message. This makes it very hard to pinpoint nested errors.

gabilang:
## Purpose
This PR will introduce an improvement to add the wrapped error's cause locations to the stack trace. 
e.g: For the following code snippet
```
public function main() {
    panic bar();
}

function bar() returns error {
    error b = error("b", foo());
    return b;
}

function foo() returns error {
    error a = error("a", foobar());
    return a;
}

function foobar() returns error {
    error c = error("c");
    return c;
}
```
the printed error message will be,
```
error: b
        at main:bar(main.bal:6)
           main:main(main.bal:2)
cause: a
        at main:foo(main.bal:11)
cause: c
        at main:foobar(main.bal:16)
```

Fixes #31554

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ErrorValue.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ErrorValue.java
@@ -37,12 +37,10 @@ import io.ballerina.runtime.internal.types.BTypeIdSet;
 import java.io.PrintStream;
 import java.io.PrintWriter;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.StringJoiner;
 
 import static io.ballerina.runtime.api.PredefinedTypes.TYPE_MAP;
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ErrorValue.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ErrorValue.java
@@ -37,10 +37,12 @@ import io.ballerina.runtime.internal.types.BTypeIdSet;
 import java.io.PrintStream;
 import java.io.PrintWriter;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.StringJoiner;
 
 import static io.ballerina.runtime.api.PredefinedTypes.TYPE_MAP;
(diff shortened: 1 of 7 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ErrorValue.java
```diff
@@ -302,9 +302,20 @@ public String getPrintableStackTrace() {
...
+        addErrorCauseWithLocation(sb, this.cause);
         return sb.toString();
     }
 
+    public void addErrorCauseWithLocation(StringBuilder sb, BError cause) {
+        if (cause != null) {
+            sb.append("\ncause: ")
+                    .append(cause.getMessage())
+                    .append("\n\tat ");
+            printStackElement(sb, cause.getStackTrace()[0], "");
```
manuranga, line 314: can there be any cases where `cause.getStackTrace()` return an empty array?

manuranga (changes requested):
I see that you are printing only one element from the `cause`, what if cause is thrown from deep within another function. We need to print all the lines in the `cause`.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ErrorValue.java
```diff
@@ -302,9 +302,20 @@ public String getPrintableStackTrace() {
...
             printStackElement(sb, stackTrace[i], "\n\t   ");
         }
+        addErrorCauseWithLocation(sb, this.cause);
         return sb.toString();
     }
 
+    public void addErrorCauseWithLocation(StringBuilder sb, BError cause) {
+        if (cause != null) {
+            sb.append("\ncause: ")
+                    .append(cause.getMessage())
```
warunalakshitha, line 312: Shall we try to reduce this code from getPrintableStackTrace. May be having common method.

warunalakshitha:
We need to test with recursive functions as well.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ErrorValue.java
```diff
@@ -302,9 +302,20 @@ public String getPrintableStackTrace() {
         for (int i = 1; i < stackTrace.length; i++) {
             printStackElement(sb, stackTrace[i], "\n\t   ");
         }
+        addErrorCauseWithLocation(sb, this.cause);
```
warunalakshitha, line 305: Should we fix this in place we print the stacktrace. getPrintableStackTrace is for given error stack.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ErrorValue.java
gabilang: When the cause is not `null` we will always have elements in `StackTraceElement[]`. So IMO, do not need to have a separate check for that.

HindujaB (approved):
LGTM

manuranga:
Please create and link the issue to remove duplicate stack lines to this before merging.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/error/ErrorTest.java
```diff
@@ -388,6 +388,33 @@ public void testErrorBindingPattern() {
...
+        } catch (Exception e) {
+            expectedException = e;
+        }
+
+        Assert.assertNotNull(expectedException);
+        String message = expectedException.getMessage();
+        Assert.assertEquals(message, "error: error1\n" +
+                "\tat error_test:foo(error_test.bal:470)\n" +
+                "\t   error_test:testStackTraceWithErrorCauseLocation(error_test.bal:466)\n" +
+                "cause: error2\n" +
```
warunalakshitha, line 405: Shall we create separate issue for stacktrace duplication?

warunalakshitha (approved):
LGTM

manuranga (approved):

PR #32022, opened 2021-08-05
Fix class cast exceptions thrown by runtime

### LINKED ISSUE #31282: Runtime throw Class cast error when interop function return invalid value

gabilang:
Fixes #31282

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
@@ -258,6 +258,10 @@ public class TypeChecker {
      * @return true if the value belongs to the given type, false otherwise
      */
     public static boolean checkIsType(Object sourceVal, Type targetType) {
+        if (!(sourceVal == null || sourceVal instanceof Number || sourceVal instanceof BString ||
+                sourceVal instanceof Boolean || sourceVal instanceof BValue)) {
+            throw ErrorUtils.createTypeAssignableError(sourceVal.getClass(), targetType);
+        }
         return checkIsType(sourceVal, getType(sourceVal), targetType);
     }
 
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
@@ -258,10 +258,6 @@ public class TypeChecker {
      * @return true if the value belongs to the given type, false otherwise
      */
     public static boolean checkIsType(Object sourceVal, Type targetType) {
-        if (!(sourceVal == null || sourceVal instanceof Number || sourceVal instanceof BString ||
-                sourceVal instanceof Boolean || sourceVal instanceof BValue)) {
-            throw ErrorUtils.createTypeAssignableError(sourceVal.getClass(), targetType);
-        }
         return checkIsType(sourceVal, getType(sourceVal), targetType);
     }
 
@@ -3089,6 +3085,14 @@ public class TypeChecker {
             }
         }
         return true;
+    }
+
+    public static Object handleAnydataValues(Object sourceVal, Type targetType) {
+        if (!(sourceVal == null || sourceVal instanceof Number || sourceVal instanceof BString ||
+                sourceVal instanceof Boolean || sourceVal instanceof BValue)) {
+            throw ErrorUtils.createJToBTypeCastError(sourceVal.getClass(), targetType);
+        }
+        return sourceVal;
     }
 
     private TypeChecker() {
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
@@ -3088,8 +3088,8 @@ public class TypeChecker {
     }
 
     public static Object handleAnydataValues(Object sourceVal, Type targetType) {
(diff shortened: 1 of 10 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -258,6 +258,10 @@ public static boolean anyToJBoolean(Object sourceVal) {
      * @return true if the value belongs to the given type, false otherwise
      */
     public static boolean checkIsType(Object sourceVal, Type targetType) {
```
warunalakshitha, line 260: This will impact on all checkIsType calls. Given issue only occurs for interop methods returning union types. Can we handle onlly those.

warunalakshitha (changes requested):

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/javainterop/RefTypeTests.java
```diff
@@ -240,11 +240,18 @@ public void testGetInvalidIntegerAsMixType() {
...
     @Test(expectedExceptions = {BLangRuntimeException.class},
-            expectedExceptionsMessageRegExp = "error: java.lang.ClassCastException.*")
+            expectedExceptionsMessageRegExp = "error: 'class java.lang.String' cannot be assigned to type 'anydata'.*")
     public void testInteropWithJavaStringReturn() {
         BRunUtil.invoke(result, "interopWithJavaStringReturn");
     }
 
+    @Test(expectedExceptions = {BLangRuntimeException.class},
+            expectedExceptionsMessageRegExp = "error: 'class io.ballerina.runtime.api.Module' cannot be " +
+                    "assigned to type 'anydata'.*")
```
warunalakshitha, line 250: Shall we assert full stack trace here.

gabilang: Addressed with 1cf6f59

tests/jballerina-unit-test/src/test/java/org/ballerinalang/nativeimpl/jvm/tests/StaticMethods.java
```diff
@@ -173,6 +174,10 @@ public static Object acceptNothingInvalidAnyReturn() {
         return "invalid java string";
     }
 
+    public static List<String> acceptNothingInvalidAnydataReturn() {
```
warunalakshitha, line 177: shall we add a test which return unions of anydata ( int|string[]) and return int[] )

bvm/ballerina-runtime/src/main/resources/MessagesBundle_en_US.properties
```diff
@@ -160,3 +160,4 @@ tuple.index.out.of.range = tuple index out of range: index: {0}, size: {1}
 illegal.array.insertion = array of length {0} cannot be expanded into array of length {1} without filler values
 java.null.reference = handle refers to Java null
 illegal.tuple.insertion = tuple of length {0} cannot be expanded into tuple of length {1} without filler values
+incompatible.type.assignment = ''{0}'' cannot be assigned to type ''{1}''
```
warunalakshitha, line 163: We do not need to put this in here except any difference to MessagesBundle.properties message.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -3087,6 +3087,14 @@ private static boolean containsType(Set<Object> valueSpace, Type type) {
         return true;
     }
 
+    public static Object handleAnydataValues(Object sourceVal, Type targetType) {
+        if (!(sourceVal == null || sourceVal instanceof Number || sourceVal instanceof BString ||
```
warunalakshitha, line 3091: Shall we invert if condition

bvm/ballerina-runtime/src/main/resources/MessagesBundle_en_US.properties
gabilang: But most of the previously added messages also same in the both files. We are maintaining two message bundle properties for English and English(US) and they should reflect the native way of expressions of both languages. But US one is not being updated recently for some newly added error messages. Since it is there for a reason (might be for language selection), INS not adding to this is a bug.

warunalakshitha: IMO we need this only if we need change different message to English(US) . If it is not there MessagesBundle.properties will used. Here we have same message in both files then no need.

warunalakshitha (approved):
LGTM

PR #32036, opened 2021-08-06
Fix visibility of let expression's inner variables in the debugger view

### LINKED ISSUE #28883: [Debugger] `let` expression inner variables are visible outside the expression
**Description:**
Currently `let` expression variables are visible in debug mode. Those variables should not be visible.

**Steps to reproduce:**
- Consider the code segment:
```
public function main() {
    string greeting = let string hello = "Hello ",
                          string ballerina = "Ballerina!"
                      in hello + ballerina;
}
```
- Run the above program in debbubg mode.
- We can see the `greeting`, `hello` and `ballerina` are visible in the local variables panel.
- Expected behaviour is only `greeting` variable should be visible. Other variables (`hello` and `ballerina`) should not be visible

image

gabilang:
Fixes #28883

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
@@ -116,6 +116,7 @@ import org.wso2.ballerinalang.compiler.tree.expressions.BLangInvocation;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangIsAssignableExpr;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangIsLikeExpr;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangLambdaFunction;
+import org.wso2.ballerinalang.compiler.tree.expressions.BLangLetExpression;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangListConstructorExpr;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangListConstructorExpr.BLangArrayLiteral;
 import org.wso2.ballerinalang.compiler.tree.expressions.BLangListConstructorExpr.BLangJSONArrayLiteral;
@@ -1155,7 +1156,8 @@ public class BIRGen extends BLangNodeVisitor {
     @Override
     public void visit(BLangSimpleVariableDef astVarDefStmt) {
         VarKind kind;
-        if (astVarDefStmt.var.symbol.origin == SymbolOrigin.VIRTUAL) {
+        if (astVarDefStmt.var.symbol.origin == SymbolOrigin.VIRTUAL ||
+                astVarDefStmt.parent instanceof BLangLetExpression) {
             kind = VarKind.SYNTHETIC;
         } else {
             kind = VarKind.LOCAL;
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
@@ -1156,9 +1156,10 @@ public class BIRGen extends BLangNodeVisitor {
     @Override
     public void visit(BLangSimpleVariableDef astVarDefStmt) {
         VarKind kind;
-        if (astVarDefStmt.var.symbol.origin == SymbolOrigin.VIRTUAL ||
-                astVarDefStmt.parent instanceof BLangLetExpression) {
+        if (astVarDefStmt.var.symbol.origin == SymbolOrigin.VIRTUAL) {
             kind = VarKind.SYNTHETIC;
+        } else if (astVarDefStmt.parent instanceof BLangLetExpression) {
+            kind = VarKind.LET;
         } else {
             kind = VarKind.LOCAL;
         }
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/methodgen/MethodGen.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/methodgen/MethodGen.java
@@ -946,6 +946,17 @@ public class MethodGen {
                 if (localVar.endBB != null) {
                     endLabel = labelGen.getLabel(funcName + endBB.id.value + "beforeTerm");
                 }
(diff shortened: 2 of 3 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1155,7 +1156,8 @@ public void visit(BLangFail failNode) {
     @Override
     public void visit(BLangSimpleVariableDef astVarDefStmt) {
         VarKind kind;
-        if (astVarDefStmt.var.symbol.origin == SymbolOrigin.VIRTUAL) {
+        if (astVarDefStmt.var.symbol.origin == SymbolOrigin.VIRTUAL ||
```
warunalakshitha, line 1159: Shall we use separate VarKind for let expression? Should we sepearate out expression variable in bir @manuranga

manuranga: I don't this is right. `SYNTHETIC` kind is for variables that user didn't write themself. But if you look the original issue, both `hello` and `ballerina` are coming from the source, not desugered.  So why should they be marked `SYNTHETIC`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/methodgen/MethodGen.java
```diff
@@ -946,6 +946,17 @@ private void createLocalVariableTable(BIRFunction func, BIRVarToJVMIndexMap inde
                 if (localVar.endBB != null) {
                     endLabel = labelGen.getLabel(funcName + endBB.id.value + "beforeTerm");
                 }
+            } else if (localVar.kind == VarKind.SYNTHETIC && isCompilerAddedVars(localVar.metaVarName)) {
+                if (localVar.startBB != null) {
+                    startLabel = labelGen.getLabel(funcName + SCOPE_PREFIX + localVar.insScope.id);
+                }
+                for (int j = i + 1; j < func.localVars.size(); j++) {
```
manuranga, line 953: does this var not have a `endBB`? If so, isn't that an issue.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
gabilang: We need to have a way to separately identify let expression inner variables among the function local variables. Currently let expression inner variables also marked with the variable kind `LOCAL` and we can't differentiate them from other local variables to separately handle their visibility scopes. Yes `SYNTHETIC` kind has been introduced for the variables generated at desugar (usually these variable names starts with specific symbols like $). So, using `SYNTHETIC` kind for  let expression inner variable might not be a good approach. Is it okay to add a new variable kind for expression inner variables?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/methodgen/MethodGen.java
gabilang: We need to end the visibility of inner variables once we came out the expression. `Let` expression does not have any terminator instructions and so we can't keep the end label based on the `endBB`. 
eg: For the following ballerina code, 
```
public function main() {

    int a = 1;

    string greeting = let string hello = "Hello ", string ballerina = "Ballerina!" 
                        in hello + ballerina;

    int b = 5;
}
```
BBs are generated like this,
```
bb0 {
        %1 = ConstLoad 1;
        %4 = ConstLoad Hello ;
        %6 = ConstLoad Ballerina!;
        %8 = %4 + %6;
        %3 = %8;
        %13 = ConstLoad 5;
        %0 = ConstLoad 0;
        GOTO bb2;
    }
    bb1 {
        GOTO bb2;
    }
    bb2 {
        return;
    }
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/methodgen/MethodGen.java
```diff
@@ -946,6 +946,17 @@ private void createLocalVariableTable(BIRFunction func, BIRVarToJVMIndexMap inde
                 if (localVar.endBB != null) {
                     endLabel = labelGen.getLabel(funcName + endBB.id.value + "beforeTerm");
                 }
+            } else if (localVar.kind == VarKind.LET) {
+                if (localVar.startBB != null) {
+                    startLabel = labelGen.getLabel(funcName + SCOPE_PREFIX + localVar.insScope.id);
+                }
+                for (int j = i + 1; j < func.localVars.size(); j++) {
+                    BIRVariableDcl syntheticVar = func.localVars.get(j);
+                    if (syntheticVar.kind == VarKind.SYNTHETIC && !isCompilerAddedVars(syntheticVar.metaVarName)) {
```
warunalakshitha, line 955: What happen if there are not variables after let expression variable. 
public function main() {
    string greeting = let string hello = "Hello ", string ballerina = "Ballerina!" 
                        in hello + ballerina;
}

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/methodgen/MethodGen.java
manuranga: I believe the correct thing to do is to spit `bb0` to two bb's in the frontend (we also have to make sure our optimizer doesn't merge them)
eg:
```
    bb0 {
        %1 = ConstLoad 1;
        %4 = ConstLoad Hello ;
        %6 = ConstLoad Ballerina!;
        %8 = %4 + %6;
        %3 = %8;
    }
    bb1 {
        %13 = ConstLoad 5;
        %0 = ConstLoad 0;
        GOTO bb3;
    }
    bb2 {
        GOTO bb3;
    }
    bb3 {
        return;
    }
```

then we can use above `bb1` as the end bb.

gabilang:
Closing this PR since the solution for this issue is needed to be discussed further.

Issue #31744, opened 2021-07-16
Provide default values of methods through runtime API


HindujaB:
**Description:**
$subject

**Describe your problem(s)**

**Describe your solution(s)**

**Related Issues (optional):**
https://github.com/ballerina-platform/ballerina-lang/issues/31581

HindujaB:
blocked on https://github.com/ballerina-platform/ballerina-lang/pull/31589

PR #32045, opened 2021-08-06
Improve runtime function parameter API 

### LINKED ISSUE #31887: Improve Function Parameter APIs
**Description:**
Previously, we had three separate APIs for retrieving parameter information of a (resource) function.
1. `getParamNames()`
2. `getParamDefaultability()`
3. `getParameterTypes()`

With #31604, the first two of the above three APIs are deprecated, and a new API is introduced `getParameters()`, which will return `Parameter[]`.  

But the `getParameterTypes()` API is still there, and we have to get the Parameter type using that API, although we get the `Parameter[]` from the new API. It is better if we can get the `Type` of the `Parameter` using the `Parameter` class.

Additionally, it is better to make the `name` and `isDefault` variables private and provide getters for these variables, along with a `getType()` API.

HindujaB:
## Purpose
$subject
Fixes #31887

## Approach
- Moved `getParameters()` API to FunctionType from Resource & Remote method types.
- Deprecated `getParamTypes()` API as now we can improve `getParameter()`  for the same purpose

## Remarks
TODO: Add doc comments for deprecated annotations, add more tests for the moved API. 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/Parameter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/Parameter.java
@@ -26,9 +26,11 @@ package io.ballerina.runtime.api.types;
 public class Parameter {
     public final String name;
     public final boolean isDefault;
+    public Type type;
 
-    public Parameter(String name, Boolean isDefault) {
+    public Parameter(String name, Boolean isDefault, Type type) {
         this.name = name;
         this.isDefault = isDefault;
+        this.type = type;
     }
 }
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/Parameter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/Parameter.java
@@ -16,7 +16,7 @@
  *  under the License.
  */
 
-package io.ballerina.runtime.api;
+package io.ballerina.runtime.api.types;
 
 /**
  * {@code {@link Parameter } represents the parameter of a function in ballerina.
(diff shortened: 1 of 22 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BResourceMethodType.java
```diff
@@ -53,9 +50,8 @@ public String toString() {
         }
         StringJoiner sj = new StringJoiner(",", "resource function " + accessor + " " + rp.toString() +
                 "(", ") returns (" + type.retType + ")");
-        Type[] types = type.paramTypes;
-        for (int i = 0; i < types.length; i++) {
-            Type type = types[i];
+        for (int i = 0; i < parameters.length; i++) {
+            Type type = parameters[i].type;
```
warunalakshitha, line 54: Can we cover this with test cases?

warunalakshitha (approved):

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/Parameter.java
```diff
@@ -26,9 +26,11 @@
 public class Parameter {
     public final String name;
     public final boolean isDefault;
+    public Type type;
```
ThisaruGuruge, line 29: Shouldn't these be private variables and provide getters?

@warunalakshitha @HindujaB ?

HindujaB: I'll address this when fixing https://github.com/ballerina-platform/ballerina-lang/issues/31744

ThisaruGuruge: Great. Thank you.

PR #32046, opened 2021-08-06
[Semantic API] Add Support for Markdown Documentation Parameter Symbol Lookup

### LINKED ISSUE #31964: Symbol of a documentation parameter comes as empty
**Description:**
The symbol of a markdown documentation parameter results in an empty `Optional<Symbol>`.

**Steps to reproduce:**
```
# Description
#
# + x - Field Description  
public type Coord record {
    readonly int x;
    # + y - Field Description
    int y;
};
```
Get the symbol of `x` documentation parameter.

**Affected Versions:**

sanjana:
- Type definitions 
   - Record type definitions
   - Object type definitions
- Class definitions
- Function definitions

Fixes #31964

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
```diff
@@ -283,6 +285,10 @@ public void visit(BLangFunction funcNode) {
             return;
         }
 
+        if (funcNode.getMarkdownDocumentationAttachment() != null) {
+            lookupNodes(funcNode.markdownDocumentationAttachment.parameters);
+        }
```
pubudu91, line 290: Better if these are also handled in the normal flow of the visitor, instead of adding null checks everywhere and checking.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
```diff
@@ -1595,6 +1609,19 @@ private boolean isWithinNodeMetaData(TopLevelNode node) {
             }
         }
 
+        if (node instanceof DocumentableNode) {
```
pubudu91, line 1612: Need to rethink this. **This is only working by chance, not design**.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
sanjana: Replaced with `lookupNode()` for `markdownDocumentationAttachment` property.

tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolPositionTest.java
```diff
@@ -147,6 +147,11 @@ public void testTypeDefPositions(int sLine, int sCol, String expSymbolName, int
             return new Object[][] {
                     {88, 4, "Module", 78, 12, 78, 18},
                     {80, 5, "TypeDef", 80, 5, 80, 12},
+                    {96, 4, "abc", 98, 17, 98, 20},
```
pubudu91, line 150: Why add tests as position tests for these? 🤔

sanjana: When looking up a symbol of a documentation parameter, the actually defined position of the symbol is different from the cursor position. 
Therefore, in order to verify that info as well, decided to use the Position Tests for these.

The first two parameters are the cursor position, followed by the symbol name, and the position where the symbol is actually defined.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
sanjana: Updated the logic which prevented from reaching this code segment.

tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolPositionTest.java
pubudu91: I see. I think it'd be best to move these tests to their own test class since what we are primarily interested in is whether `symbol()` works for doc params as well. This position test is for testing particular scenarios where we need to check whether the symbol's position is correct.

sanjana: @pubudu91 The https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/ballerina-compiler-api-test/src/test/resources/test-src/documentation_test.bal source file contains a couple of compilation errors that block the `CompilerPhaseRunner` from proceeding after the Type Checking phase. 

Therefore, would it be better if a separate source file is created if the `DocumentationTest` class is to be used for these tests as well?

sanjana: Moved to `DocumentationTest` class.

pubudu91 (approved):

PR #32080, opened 2021-08-10
[WIP] Fix `typeof` expression providing wrong outputs

### LINKED ISSUE #15278: [Spec Deviation] The `typedesc` returned by `typeof` is not in line with the spec
Pls refer https://github.com/ballerina-platform/ballerina-spec/issues/106

Currently typedesc only contain widen type, this also need to include the narrow type.

typedesc a = typeof 1;

Internally 'a' should contain both 'int' (1 is a int at this context) and singleton type '1' as the narrow type.

HindujaB:
## Purpose
As per the spec, the `typeof` expression must return the type as the following,
- For basic simple value, 
       - provide a type with single shape
       - For each `typeof` call, produce a new typedesc value.
       
- For reference value, 
       - provide a type with single shape - if the value is immutable
       - Or else provide the typedesc that is used to construct the value
       - For each `typeof` call, provide identical typedesc values.

Fixes #15278

## Approach
- Create and return new singleton typedesc values for simple basic type values.
- Store the typedesc information of the reference values which can be returned at the `typeof` call.
- Store the singleton type as the value type when creating an immutable value
 
## Samples
```ballerina

    // Basic Simple Type 
    int i = 34;
    test:assertFalse(typeof i === typeof i); 

    // Structural types 
    int[] arr = [1, 2, 3];
    test:assertTrue(typeof arr === typeof arr); 

    map<string> mapVal = {s: "test"};
    test:assertTrue(typeof mapVal === typeof mapVal); 

    // Behavioral types - Object
    Obj0 obj = new ("abc", 123);
    test:assertTrue(typeof obj === typeof obj); 

    // Behavioral types - Error
    error err = error("This is an error!");
    test:assertTrue(typeof err === typeof err); 
```

## Remarks
- TODO: Fix immutable values

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BMap.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BMap.java
@@ -192,7 +192,5 @@ public interface BMap<K, V> extends BRefValue, BCollection {
 
     Object merge(BMap v2, boolean checkMergeability);
 
-    BTypedesc getTypedesc();
-
     void populateInitialValue(K key, V value);
 }
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BRefValue.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/values/BRefValue.java
@@ -94,4 +94,6 @@ public interface BRefValue extends BValue {
             throw new BallerinaException("error occurred while serializing data", e);
         }
     }
+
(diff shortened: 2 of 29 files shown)


HindujaB:
Some changes need to be done from debugger side. Hence, closing this and creating a new PR

PR #32105, opened 2021-08-11
Fix `typeof` expression providing wrong outputs

### LINKED ISSUE #15278: [Spec Deviation] The `typedesc` returned by `typeof` is not in line with the spec

### LINKED ISSUE #32101: Fix `typeof` usage in debugger evaluator error messages
**Description:**
The ballerina debugger evaluatior uses the `typeof` expression and the result's string value for the error messages. 
But, with the change after #32105, the `typeof` will return the singleton type instead of its basic type for simple basic values. Therefore the error messages will become unclear. 

Eg: 
```
operator '<' not defined for 'int' and 'string'
```
this error will become 
```
operator '<' not defined for '20' and 'foo'
```
We need to change this behavior. 
This can be changed to use the `getType()` API from TypeChecker via interop instead of the `typeof`. 

**Steps to reproduce:**
https://scans.gradle.com/s/xyvp55sc4wcmw/tests/failed

**Affected Versions:**

#15278

### LINKED ISSUE #32102: Fix string index providing the wrong value for debugger evaluator
**Description:**
The following debugger evaluator test  is failing when we changed it after #32105.
 https://github.com/ballerina-platform/ballerina-lang/blob/a2fb574e9f8576863e22c9b1aac0bffa4581bfa3/tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/evaluation/ExpressionEvaluationTest.java#L415
 
 The change, 
 ```java
 debugTestRunner.assertExpression(context, String.format("typeof %s[0]", STRING_VAR), "f", "typedesc");
 ```

It should pass, as the typeof evaluation is done on the string value `foo` with index 0. 

This passes correctly in `bal run` with the same `typeof` changes
```ballerina
string s = "foo";
io:println(typeof s[0]); // typedesc f
```

**Affected Versions:**
beta3-SNAPSHOT

#15278

### LINKED ISSUE #32130: Fix `typeof` usage in debugger type representation for immutable values
**Description:**
The ballerina debugger uses the `typeof` expression and the result's string value for type representations. 
But, with the change after #32105, the `typeof` will return the singleton type instead of its basic type for structural immutable values (array, tuple, map, record and table). This can result in a wrong representation for the types of values. 

Eg: 
For the following immutable map value (As it is a constant),
```ballerina
const map<string> nameMap = {"name":"John"};
```
The `getTypedesc()` will provide a singleton type of `{"name":"John"}`. 

Therefore, the type of the constant will become `{"name":"John"}` instead of `map<string>`.
This can be unclear to the users.

**Steps to reproduce:**
https://scans.gradle.com/s/szyxni77qvqb2/tests/failed

#15278

HindujaB:
## Purpose
As per the spec, the `typeof` expression must return the type as the following,
- For basic simple value, 
       - provide a type with single shape
       - For each `typeof` call, produce a new typedesc value.
       
- For reference value, 
       -  if the value is immutable & a container (structural value),  provide a type with single shape
       - Or else provide the typedesc that is used to construct the value
       - For each `typeof` call, provide identical typedesc values.

Fixes #15278
Fixes #32101
Fixes #32102
Fixes #32130

## Approach
- Create and return new singleton typedesc values for simple basic type values.
- Store the typedesc information of the reference values which can be returned at the `typeof` call.
- Store the singleton type as the value type when creating an immutable value
 
## Samples
```ballerina

    // Basic Simple Type 
    int i = 34;
    io:println(typeof i); // typedesc 34
    test:assertFalse(typeof i === typeof i); 

    // Structural types 
    int[] arr = [1, 2, 3];
    test:assertTrue(typeof arr === typeof arr); 

    map<string> mapVal = {s: "test"};
    test:assertTrue(typeof mapVal === typeof mapVal); 

    // Behavioral types - Object
    Obj0 obj = new ("abc", 123);
    test:assertTrue(typeof obj === typeof obj); 

    // Behavioral types - Error
    error err = error("This is an error!");
    test:assertTrue(typeof err === typeof err); 

   // Immutable Values
    int[] & readonly arr = [1, 2, 3];
    test:assertTrue(typeof arr === typeof arr);
    io:println(typeof arr); // typedesc [1,2,3]
```

## Remarks
- Duplicate of PR #32080

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/FutureValue.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/FutureValue.java
@@ -21,6 +21,7 @@ package io.ballerina.runtime.internal.values;
  import io.ballerina.runtime.api.types.Type;
  import io.ballerina.runtime.api.values.BFuture;
  import io.ballerina.runtime.api.values.BLink;
+ import io.ballerina.runtime.api.values.BTypedesc;
  import io.ballerina.runtime.internal.scheduling.Strand;
  import io.ballerina.runtime.internal.types.BFutureType;
 
@@ -39,7 +40,9 @@ package io.ballerina.runtime.internal.values;
  */
  public class FutureValue implements BFuture, RefValue {
 
-     public Strand strand;
+    private final BTypedesc typedesc;
+
+    public Strand strand;
 
      public Object result;
 
@@ -58,7 +61,8 @@ package io.ballerina.runtime.internal.values;
          this.strand = strand;
          this.callback = callback;
          this.type = new BFutureType(constraint);
-     }
+        this.typedesc = new TypedescValueImpl(this.type);
+    }
 
      @Override
      public String stringValue(BLink parent) {
@@ -93,7 +97,12 @@ package io.ballerina.runtime.internal.values;
          throw new UnsupportedOperationException();
      }
 
-     public void cancel() {
+    @Override
+    public BTypedesc getTypedesc() {
+        return typedesc;
+    }
+
+    public void cancel() {
          this.strand.cancel = true;
      }
 
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/FutureValue.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/FutureValue.java
@@ -15,7 +15,7 @@
   *  specific language governing permissions and limitations
   *  under the License.
   */
-package io.ballerina.runtime.internal.values;
+ package io.ballerina.runtime.internal.values;
 
  import io.ballerina.runtime.api.async.Callback;
  import io.ballerina.runtime.api.types.Type;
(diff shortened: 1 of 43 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/FutureValue.java
```diff
@@ -21,6 +21,7 @@
  import io.ballerina.runtime.api.types.Type;
```
Nadeeshan96, line 21: In this file, there is a slight indentation to the right by one space in some places. Shall we fix that

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlValue.java
```diff
@@ -51,6 +52,7 @@
 public abstract class XmlValue implements RefValue, BXml, CollectionValue {
 
     Type type = PredefinedTypes.TYPE_XML;
+    protected TypedescValue typedesc = new TypedescValueImpl(type);
```
warunalakshitha, line 55: IMO behavior typeof operation with immutable xml is not covered here . Shall we call the value utils to create the typedesc here.

warunalakshitha (approved):

warunalakshitha (approved):

Issue #32117, opened 2021-08-11
Move concurrent execution of resource method to Listener side            


warunalakshitha:
**Description:**
ATM runtime decide whether to  execute resource method concurrently or sequentially based on  object isolation. But Isolation does not 100% guarantee concurrent safety since what ever the arguments pass to parameters can be non concurrent safe or may be not. Its depend to caller side, in here Listener implementation. So we need to provide proper runtime api so object method caller can decide whether to execute method concurrently or sequentially.

PR #32118, opened 2021-08-11
Move concurrent execution of resource method to Listener side

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/32117

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
@@ -61,6 +61,36 @@ public class Runtime {
         return new Runtime(strand.scheduler);
     }
 
+    /**
+     * Invoke Object method asynchronously. This will schedule the function and block the strand.
+     *
+     * @param object     Object Value.
+     * @param methodName Name of the method.
+     * @param strandName Name for newly creating strand which is used to execute the function pointer. This is
+     *                   optional and can be null.
+     * @param metadata   Meta data of new strand.
+     * @param callback   Callback which will get notify once method execution done.
+     * @param isIsolated Ensures function can be called preserving isolation
+     * @param properties Set of properties for strand
+     * @param returnType Expected return type of this method
+     * @param args       Ballerina function arguments.
+     * @return           {@link FutureValue} containing return value of executing this method.
+     */
+    public BFuture invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
+                                     boolean isIsolated, Callback callback, Map<String, Object> properties,
+                                     Type returnType, Object... args) {
+        if (object == null) {
+            throw new NullPointerException();
+        }
+        Function<?, ?> func = o -> object.call((Strand) (((Object[]) o)[0]), methodName, args);
+        if (isIsolated) {
+            return scheduler.schedule(new Object[1], func, null, callback, properties, returnType, strandName,
+                    metadata);
+        } else {
+            return scheduler.scheduleToObjectGroup(object, new Object[1], func, null, callback, properties,
+                    returnType, strandName, metadata);
+        }
+    }
 
      /**
       * Invoke Object method asynchronously. This will schedule the function and block the strand.
@@ -72,7 +102,9 @@ public class Runtime {
       * @param metadata   Meta data of new strand.
       * @param callback   Callback which will get notify once method execution done.
       * @param args       Ballerina function arguments.
-      * @return the result of the function invocation
+      * @return           the result of the function invocation
+      * @deprecated       Use {@link #invokeMethodAsync(BObject, String, String, StrandMetadata, boolean,
+      *                   Callback, Map, Type, Object...)} providing isolation of method call.
       */
      public Object invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
                                      Callback callback, Object... args) {
@@ -119,6 +151,8 @@ public class Runtime {
      * @param returnType Expected return type of this method
      * @param args       Ballerina function arguments.
      * @return           {@link FutureValue} containing return value of executing this method.
+     * @deprecated       Use {@link #invokeMethodAsync(BObject, String, String, StrandMetadata, boolean,
+     *                   Callback, Map, Type, Object...)} providing isolation of method call.
      */
     public BFuture invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
                                      Callback callback, Map<String, Object> properties,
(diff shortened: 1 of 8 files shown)


warunalakshitha:
Code coverage will get drop since deprecated apis are no longer used in ballerina lang

manuranga:
Looks good

manuranga:
Looks good to me.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
```diff
@@ -61,6 +54,37 @@ public static Runtime getCurrentRuntime() {
...
+     * @param properties Set of properties for strand
+     * @param returnType Expected return type of this method
+     * @param args       Ballerina function arguments.
+     * @return           {@link FutureValue} containing return value of executing this method.
+     */
+    public BFuture invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
+                                     boolean isIsolated, Callback callback, Map<String, Object> properties,
+                                     Type returnType, Object... args) {
+        if (object == null) {
+            throw new NullPointerException();
```
riyafa, line 77: I think the more suitable  exception would be `IllegalArgumentException` because we do not want the argument `object` to be null. Otherwise we can choose to use it an `Optional` argument as input: https://www.oracle.com/technical-resources/articles/java/java8-optional.html

riyafa: I think we could also have a test to cover this line.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
```diff
@@ -72,40 +96,21 @@ public static Runtime getCurrentRuntime() {
...
       * @param callback   Callback which will get notify once method execution done.
       * @param args       Ballerina function arguments.
-      * @return the result of the function invocation
+      * @return           the result of the function invocation
+      * @deprecated       Use {@link #invokeMethodAsync(BObject, String, String, StrandMetadata, boolean,
+      *                   Callback, Map, Type, Object...)} providing isolation of method call.
       */
      public Object invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
                                      Callback callback, Object... args) {
          Function<?, ?> func = o -> object.call((Strand) (((Object[]) o)[0]), methodName, args);
```
manuranga, line 105: can remove most of this method's body by calling `invokeMethodAsync(BObject, String, String, StrandMetadata, boolean, Callback, Map, Type, Object...)` here.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
```diff
@@ -61,6 +54,37 @@ public static Runtime getCurrentRuntime() {
...
+     * @param callback   Callback which will get notify once method execution done.
+     * @param isIsolated Ensures method can be called preserving isolation with no data race is possible for the
+     *                   mutable state freely reachable from the arguments that it passes to the function.
+     * @param properties Set of properties for strand
+     * @param returnType Expected return type of this method
+     * @param args       Ballerina function arguments.
+     * @return           {@link FutureValue} containing return value of executing this method.
+     */
+    public BFuture invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
+                                     boolean isIsolated, Callback callback, Map<String, Object> properties,
```
MaryamZi, line 74: Shouldn't we use something like callConcurrently instead of isIsolated?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BObjectType.java
```diff
@@ -82,11 +85,37 @@ public String getAnnotationKey() {
...
     public MethodType[] getMethods() {
         return methodTypes;
     }
 
+    @Override
+    public boolean isIsolated() {
+        return SymbolFlags.isFlagOn(getFlags(), SymbolFlags.ISOLATED);
+    }
+    @Override
+    public boolean isIsolated(String methodName) {
```
MaryamZi, line 99: Shouldn't this method be exposed via the method type instead?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
```diff
@@ -61,6 +62,39 @@ public static Runtime getCurrentRuntime() {
...
+     */
+    public BFuture invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
+                                     boolean callConcurrently, Callback callback, Map<String, Object> properties,
+                                     Type returnType, Object... args) {
+        if (object == null) {
+            throw ErrorCreator.createError(StringUtils.fromString("object cannot be null"));
+        }
+        if (methodName == null) {
+            throw ErrorCreator.createError(StringUtils.fromString("method name cannot be null"));
+        }
```
riyafa, line 88: Can't we have a generic exceptions for all null pointers? Or since this is a java null can we throw the java `NullPointerException` or more correct to throw `IlegalArgumentException`? Saying `object cannot be null` in a Ballerina error is strange because we only have `nils` in Ballerina

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
```diff
@@ -119,16 +129,20 @@ private boolean isIsolated(ObjectType type, String methodName) {
...
     public BFuture invokeMethodAsync(BObject object, String methodName, String strandName, StrandMetadata metadata,
                                      Callback callback, Map<String, Object> properties,
                                      Type returnType, Object... args) {
         if (object == null) {
-            throw new NullPointerException();
+            throw ErrorCreator.createError(StringUtils.fromString("object cannot be null"));
+        }
+        if (methodName == null) {
+            throw ErrorCreator.createError(StringUtils.fromString("method name cannot be null"));
         }
```
riyafa, line 143: same as above I think we would need a generic error from Ballerina for this

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/Runtime.java
warunalakshitha: since this java api I'll change to IlegalArgumentException

warunalakshitha: created https://github.com/ballerina-platform/ballerina-lang/issues/32225 recheck on this.

PR #32128, opened 2021-08-11
Fix CCE when value:ToJson fails

### LINKED ISSUE #32001: CCE when `value:toJson` fails
**Description:**
$title.

**Steps to reproduce:**
```ballerina
public function main() {
    anydata[] x = [];
    x.push(x);
    json y = x.toJson();
}
```

```cmd
$ bal run pr_test.bal 
Compiling source
        pr_test.bal

Running executable

error: java.lang.ClassCastException {"message":"class io.ballerina.runtime.internal.util.exceptions.BallerinaException cannot be cast to class io.ballerina.runtime.api.values.BValue (io.ballerina.runtime.internal.util.exceptions.BallerinaException and io.ballerina.runtime.api.values.BValue are in unnamed module of loader 'app')"}
        at ballerina.lang.value.1_0_0:toJson(value.bal:179)
           pr_test:main(pr_test.bal:4)
```

**Affected Versions:**
slbeta2

gabilang:
Fixes #32001

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
+++ b/langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
@@ -1892,6 +1892,18 @@ function testToStringOnCycles() {
      assert(x.toString(), "{\"ee\":3,\"1\":{\"qq\":5,\"1\":...,\"2\":[2,3,5,...]}}");
 }
 
+function testToJsonWithCyclicParameter() {
+    anydata[] x = [];
+    x.push(x);
+    json|error y = trap x.toJson();
+    assert(y is error, true);
+    error err = <error> y;
+    var message = err.detail()["message"];
+    string messageString = message is error ? message.toString() : message.toString();
+    assert(err.message(), "{ballerina/lang.value}CyclicValueReferenceError");
+    assert(messageString, "'anydata[]' value has cyclic reference");
+}
+
 function assert(anydata actual, anydata expected) {
     if (expected == actual) {
         return;
--- a/langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
+++ b/langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
@@ -14,6 +14,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
+import ballerina/jballerina.java;
 import ballerina/lang.'value as value;
 
 type Address record {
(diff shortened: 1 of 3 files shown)


langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -1892,6 +1892,18 @@ function testToStringOnCycles() {
...
 }
 
+function testToJsonWithCyclicParameter() {
+    anydata[] x = [];
+    x.push(x);
+    json|error y = trap x.toJson();
+    assert(y is error, true);
+    error err = <error> y;
+    var message = err.detail()["message"];
+    string messageString = message is error ? message.toString() : message.toString();
```
warunalakshitha, line 1902: can we use message.toString()

gabilang: Since the `err.detail()["message"]` can be a `string` or `error`, we need the `is error` check in this case.

riyafa: Why do we need the check if we are assigning the same value to the `messageString`?

gabilang: Otherwise we can't use `toString()` method directly to `message`. Instead of this check we can use `<string> checkpanic err.detail()["message"]` to assert the error detail. I'll update with that.

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -1892,6 +1892,16 @@ function testToStringOnCycles() {
...
 }
 
+function testToJsonWithCyclicParameter() {
+    anydata[] x = [];
+    x.push(x);
+    json|error y = trap x.toJson();
+    assert(y is error, true);
+    error err = <error> y;
+    assert(err.message(), "{ballerina/lang.value}CyclicValueReferenceError");
+    assert(<string> checkpanic err.detail()["message"], "'anydata[]' value has cyclic reference");
```
riyafa, line 1902: you can do the following without casting:
```suggestion
    assert(err.detail()["message"].toString(), "'anydata[]' value has cyclic reference");
```

gabilang: It gave compilation errors. Since `err.detail()["message"]` is `string|error` we can't directly use `.toString()` on it right?

riyafa (approved):

PR #32179, opened 2021-08-16
Fix incorrect results for the equality of floating points with non-simple types

### LINKED ISSUE #32173: Wrong float `==` behavior when type is not simple
```
import ballerina/io;

type Num float|int;

public function main() {
    Num n1 = 0.0;
    Num n2 = -0.0;
    io:println(n1 == n2);
}
```

gabilang:
```
import ballerina/io;

type Num float|int;

public function main() {
    Num n1 = 0.0;
    Num n2 = -0.0;
    Num n3 = 0.0/0.0;
    Num n4 = -0.0/0.0;
    Num n5 = 2.0;
    Num n6 = 2;

    io:println(n1 == n2);  // true
    io:println(n3 == n4);  // true
    io:println(n5 == n6);  // false

    io:println(n1 === n2);  // false
    io:println(n3 === n4);  // true
    io:println(n5 === n6);  // false
}
```

Fixes #32173

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
@@ -2708,7 +2708,6 @@ public class TypeChecker {
 
         switch (lhsValTypeTag) {
             case TypeTags.STRING_TAG:
-            case TypeTags.FLOAT_TAG:
             case TypeTags.DECIMAL_TAG:
             case TypeTags.BOOLEAN_TAG:
                 return lhsValue.equals(rhsValue);
@@ -2722,6 +2721,14 @@ public class TypeChecker {
                     return false;
                 }
                 return ((Number) lhsValue).byteValue() == ((Number) rhsValue).byteValue();
+            case TypeTags.FLOAT_TAG:
+                if (rhsValTypeTag != TypeTags.FLOAT_TAG) {
+                    return false;
+                }
+                if (Double.isNaN((Double) lhsValue) && Double.isNaN((Double) rhsValue)) {
+                    return true;
+                }
+                return ((Number) lhsValue).doubleValue() == ((Number) rhsValue).doubleValue();
             case TypeTags.XML_TAG:
                 // Instance of xml never
                 if (lhsValue instanceof XmlText) {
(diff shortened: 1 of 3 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -467,8 +467,30 @@ public static boolean isReferenceEqual(Object lhsValue, Object rhsValue) {
         Type lhsType = getType(lhsValue);
         Type rhsType = getType(rhsValue);
 
-        if (isSimpleBasicType(lhsType) && isSimpleBasicType(rhsType)) {
-            return isEqual(lhsValue, rhsValue);
+        switch(lhsType.getTag()) {
+            case TypeTags.INT_TAG:
+                if (rhsType.getTag() != TypeTags.BYTE_TAG || rhsType.getTag() != TypeTags.INT_TAG) {
```
warunalakshitha, line 472: What happen if we have integer sub types

warunalakshitha: what is the difference between isReferenceEqual and isEqual for simple basic types?

gabilang: We have used `getType` method to get the types of values. So the types of integer sub types also will be given as the type integer. Equality and exact equality checks are not same for `float` and `decimal` types. So we can't simply use `isEqual()` check of simple basic types for `isReferenceEqual()` check. Or else we need to exclude the `float` and `decimal` cases and we can handle them separately.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -467,8 +467,30 @@ public static boolean isReferenceEqual(Object lhsValue, Object rhsValue) {
...
+                    return false;
+                }
+                return lhsValue.equals(((Number) rhsValue).longValue());
```
HindujaB, line 475: Shall we have some tests for both values with same type, in order to cover these lines?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
warunalakshitha: Lets handle float and decimal diffrently and other types use same isEqual which will minimise code duplication.

warunalakshitha (approved):

PR #32184, opened 2021-08-16
Set textRange information to BLangNode

### LINKED ISSUE #32176: [Semantic API] function symbols do not contain text range information
**Description:**
All the function symbols retrieved using the semantic API seem to be missing text range information, although the line range information is correct. Therefore, apparently there's no way to resolve the corresponding syntax tree node for a given symbol, as the `io.ballerina.compiler.syntax.tree.NonTerminalNode.findNode()` method of the syntax API only accept a text range.

**Steps to reproduce:**

**Affected Versions:**

rdulmina:
## Purpose
$title

Fixes #32176

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/diagnostic/BLangDiagnosticLocation.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/diagnostic/BLangDiagnosticLocation.java
@@ -36,6 +36,7 @@ public class BLangDiagnosticLocation implements Location {
     private LineRange lineRange;
     private TextRange textRange;
 
+    @Deprecated
     public BLangDiagnosticLocation(String filePath, int startLine, int endLine, int startColumn, int endColumn) {
         this.lineRange = LineRange.from(filePath, LinePosition.from(startLine, startColumn),
                 LinePosition.from(endLine, endColumn));
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/diagnostic/BLangDiagnosticLocation.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/diagnostic/BLangDiagnosticLocation.java
@@ -42,6 +42,13 @@ public class BLangDiagnosticLocation implements Location {
         this.textRange = TextRange.from(0, 0);
     }
 
+    public BLangDiagnosticLocation(String filePath, int startLine, int endLine, int startColumn, int endColumn,
+                                   int startOffset, int length) {
+        this.lineRange = LineRange.from(filePath, LinePosition.from(startLine, startColumn),
+                LinePosition.from(endLine, endColumn));
+        this.textRange = TextRange.from(startOffset, length);
+    }
+
     @Override
     public LineRange lineRange() {
         return lineRange;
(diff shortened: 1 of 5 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
```diff
@@ -545,8 +549,10 @@ private Location getPosition(Node startNode, Node endNode) {
         }
         LinePosition startPos = startNode.lineRange().startLine();
         LinePosition endPos = endNode.lineRange().endLine();
+        TextRange startNodeTextRange = startNode.textRange();
+        int length = startNodeTextRange.length() + endNode.textRange().length();
         return new BLangDiagnosticLocation(currentCompUnitName, startPos.line(), endPos.line(),
-                                           startPos.offset(), endPos.offset());
+                startPos.offset(), endPos.offset(), startNodeTextRange.startOffset(), length);
```
dulajdilshan, line 555: Shall we keep the previous formating here

dulajdilshan: Seems like we can't wrap it within the length of a line. so, please ignore this

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/diagnostic/BLangDiagnosticLocation.java
```diff
@@ -42,6 +42,13 @@ public BLangDiagnosticLocation(String filePath, int startLine, int endLine, int
         this.textRange = TextRange.from(0, 0);
     }
```
pubudu91, line 43: Do we need this previous constructor? Will this end up confusing users/making things error prone?

rdulmina: There are places where the `TextRange` information is not available when constructing `BLangDiagnosticLocation`. But we can pass `0`s as the values to TextRange for those cases and remove the old constructor. Should we proceed with that?

pubudu91: @dulajdilshan WDYT? Is `BLangDiagnosticLocation` only used in ballerina-lang?

dulajdilshan: Yes, there are places we use `BlangDiagnosticLocation` outside of ballerina-lang. If we remove the previous one, the extensions also needed to be changed suddenly.  I think it would be better if we can have both constructors for now. Let's deprecate the previous one.

rdulmina: Updated

dulajdilshan (approved):
LGTM

pubudu91 (approved):

Issue #32177, opened 2021-08-16
Key fields in a record of a table created from a query aren't read-only


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type User record {
    readonly int id;
    string firstName;
    string lastName;
    int age;
};

public function main() returns error? {
    table<User> key(id) users = table [
        {id: 1, firstName: "John", lastName: "Doe", age: 25}
    ];

    record {
        int id;
        string firstName;
        string lastName;
        int age;
    }? r = users[1];

    // Panics - "cannot update 'readonly' field 'id' in record of type 'User'"
    // if !(r is ()) {
    //     r.id = 1;
    // }


    var result = check table key(id, name) 
        from var user in users
        where user.age > 21 && user.age < 60
        select {id: user.id, name: user.firstName, user};

    var r2 = <record {
        int id;
        string name;
        User user;
    }> result[1, "John"];

    r2.id = 1; // Doesn't panic
}
```

Even when the expected type is
```ballerina
    table<record {
        int id;
        string name;
        User user;
    }> key(id, name)
```

**Affected Versions:**
slbeta2

**Related Issues (optional):**
https://github.com/ballerina-platform/ballerina-lang/issues/32049
https://github.com/ballerina-platform/ballerina-lang/issues/30985

Issue #32049, opened 2021-08-07
Inline declaration of table type doesn't validate rowtype parameter


rdhananjaya:
**Description:**
```ballerina
function name(table<any> p) { // this doesn't emit an error message
}

type T table<any>; // invalid constraint type. expected subtype of 'map<any|error>' but found 'any'
```

**Steps to reproduce:**

**Affected Versions:**
Swan Lake beta-2
**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Issue #30985, opened 2021-06-03
Table keys are not validated in no-init var declaration


rdulmina:
**Description:**
Consider the bellow sample
```bal
type Person record {|
    int id;
    string name;
|};

public function main() {
    table<Person> key(id) tableVar; // `id` is not readonly, no error at compile time
    table<Person> key(invalidField) tableVar2; // No field name `invalidField`, no error at compile time
}
```
It does validate for var declarations with initialization

Issue #30575, opened 2021-05-14
Move analysis of the type nodes to `BLangRecordTypeNode` visitor itself


lasinicl:
**Description:**
Refer https://github.com/ballerina-platform/ballerina-lang/pull/30144#discussion_r632462259

PR #32203, opened 2021-08-17
Fix several table type related issues

lasinicl:
## Purpose
$title

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/32177 Key fields in a record of a table created from a query aren't read-only
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/32049 Inline declaration of table type doesn't validate rowtype parameter
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30985 Table keys are not validated in no-init var declaration
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/30575

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -185,6 +185,7 @@ import org.wso2.ballerinalang.compiler.tree.statements.BLangWhile;
 import org.wso2.ballerinalang.compiler.tree.statements.BLangWorkerSend;
 import org.wso2.ballerinalang.compiler.tree.statements.BLangXMLNSStatement;
 import org.wso2.ballerinalang.compiler.tree.types.BLangArrayType;
+import org.wso2.ballerinalang.compiler.tree.types.BLangBuiltInRefTypeNode;
 import org.wso2.ballerinalang.compiler.tree.types.BLangConstrainedType;
 import org.wso2.ballerinalang.compiler.tree.types.BLangErrorType;
 import org.wso2.ballerinalang.compiler.tree.types.BLangFiniteTypeNode;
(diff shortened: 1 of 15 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -749,6 +756,79 @@ public void visit(BLangErrorType errorType) {
...
+        for (BLangType memberType : unionTypeNode.memberTypeNodes) {
+            analyzeDef(memberType, env);
+        }
+    }
+
+    @Override
+    public void visit(BLangStreamType streamType) {
+        analyzeNode(streamType.constraint, env);
+        if (streamType.error != null) {
+            analyzeDef(streamType.error, env);
```
pcnfernando, line 792: Handle the null check inside analyze function so that we don't need to repeat the check

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1303,7 +1299,9 @@ private boolean validateTableConstructorExpr(BLangTableConstructorExpr tableCons
     }
 
     public void validateMapConstraintTable(BLangTableConstructorExpr tableConstructorExpr, BType expType) {
-        if (((BTableType) expType).fieldNameList != null || ((BTableType) expType).keyTypeConstraint != null) {
+        if (expType != null && (((BTableType) expType).fieldNameList != null ||
+                ((BTableType) expType).keyTypeConstraint != null) &&
+                !expType.tsymbol.owner.getFlags().contains(Flag.LANG_LIB)) {
```
pcnfernando, line 1304: Why are we filtering out the lang_lib. Is there a way to handle the map constraint in a generic manner?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
lasinicl: analyze function returns a `BType`. if we were to handle the null check there we'll have to return something for it too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
lasinicl: In lang.table there are some places that have specified a `key-type-constraint` for table constrained with map [1], which results in `table with constraint of type map cannot have key specifier or key type constraint` compilation error. Therefore it's handled here with the langlib check here.
[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/langlib/lang.table/src/main/ballerina/table.bal#L77

PR #32222, opened 2021-08-19
Add simple value tests related to nil, boolean and int

SandaruJayawardana:
Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [x] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/ballerina-spec-conformance-tests/src/test/resources/5-values-types-and-variables/boolean.balt
+++ b/tests/ballerina-spec-conformance-tests/src/test/resources/5-values-types-and-variables/boolean.balt
@@ -1,19 +1,3 @@
-//  Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
-//
-//  WSO2 Inc. licenses this file to you under the Apache License,
-//  Version 2.0 (the "License"); you may not use this file except
-//  in compliance with the License.
-//  You may obtain a copy of the License at
-//
-//    http://www.apache.org/licenses/LICENSE-2.0
-//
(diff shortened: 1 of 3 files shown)


tests/ballerina-spec-conformance-tests/src/test/resources/5-values-types-and-variables/int.balt
```diff
@@ -0,0 +1,60 @@
...
+}
+
+Test-case: error
+Label: int
+Author: SandaruJayawardana
+
+public function main() {
+    int x = 9223372036854775808;
+    io:println(x); // @error <Integer '9223372036854775808' too large>
+}
```
MaryamZi, line 60: There's potentially few more tests like
- `DecimalNumber` cannot start with `0` unless the value itself is `0`. e.g., `01` should result in an error.
- `HexIntLiteral` must have both the `HexIndicator` and `HexNumber`

But these could be parser tests too. Let's discuss during the review and decide whether or not we should be adding such tests here.

tests/ballerina-spec-conformance-tests/src/test/resources/5-values-types-and-variables/boolean.balt
```diff
@@ -0,0 +1,11 @@
+Test-case: succeed
```
chiranSachintha, line 1: @MaryamZi  Here `succeed` should be changed to `output` . Right?

PR #32237, opened 2021-08-19
Add APIs to find references within the given document

### LINKED ISSUE #32136: Add an API for finding symbol references in a given document only
**Description:**
Currently, in order to find the references we access all the constructs in a given BLangPackage and when it comes to a larger project, we iterate through around hundreds of constructs in a single run.
Since we use the references API to implement the semantic token feature, we spend cycles to find the references in unnecessary documents and related constructs as well. Exposing an API to capture the references in a given document we can improve the performance.

**Describe your solution(s)**
Introduce an API to find the references in a single document

dulajdilshan:
## Purpose
$title

Fixes #32136 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
@@ -187,6 +187,22 @@ public class BallerinaSemanticModel implements SemanticModel {
     }
 
     @Override
+    public List<Location> references(Symbol symbol, Document targetDocument, boolean withDefinition) {
+        Optional<Location> symbolLocation = symbol.getLocation();
+
+        // Assumption is that the location will be null for regular type symbols
+        if (symbolLocation.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        BLangNode node = new NodeFinder(false)
+                .lookup(getCompilationUnit(targetDocument), symbolLocation.get().lineRange());
+
+        ReferenceFinder refFinder = new ReferenceFinder(withDefinition);
+        return refFinder.findReferences(node, getInternalSymbol(symbol));
+    }
+
+    @Override
     public List<Location> references(Symbol symbol, boolean withDefinition) {
         Optional<Location> symbolLocation = symbol.getLocation();
 
@@ -200,6 +216,27 @@ public class BallerinaSemanticModel implements SemanticModel {
 
         ReferenceFinder refFinder = new ReferenceFinder(withDefinition);
         return refFinder.findReferences(node, getInternalSymbol(symbol));
+    }
+
+    @Override
+    public List<Location> references(Document sourceDocument,
+                                     Document targetDocument,
+                                     LinePosition position,
+                                     boolean withDefinition) {
+        BLangCompilationUnit sourceCompilationUnit = getCompilationUnit(sourceDocument);
+        SymbolFinder symbolFinder = new SymbolFinder();
+        BSymbol symbolAtCursor = symbolFinder.lookup(sourceCompilationUnit, position);
+
+        if (symbolAtCursor == null) {
+            return Collections.emptyList();
+        }
+
+        BLangCompilationUnit targetCompilationUnit = getCompilationUnit(targetDocument);
+        BLangNode node = new NodeFinder(false)
+                .lookup(targetCompilationUnit, symbolAtCursor.pos.lineRange());
+
+        ReferenceFinder refFinder = new ReferenceFinder(withDefinition);
+        return refFinder.findReferences(node, symbolAtCursor);
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -175,45 +175,86 @@ public BallerinaSemanticModel(BLangPackage bLangPackage, CompilerContext context
...
+            symbolAtCursor = getInternalSymbol(symbol);
+        } else {
+            BLangCompilationUnit sourceCompilationUnit = getCompilationUnit(sourceDocument);
+            SymbolFinder symbolFinder = new SymbolFinder();
+            symbolAtCursor = symbolFinder.lookup(sourceCompilationUnit, position);
+            if (symbolAtCursor == null) {
+                return Collections.emptyList();
+            }
+            symbolLocation = Optional.of(symbolAtCursor.getPosition());
+        }
```
pubudu91, line 243: Is this the best way to handle this? 🤔  Feel like this method is trying to do too many things. Sometimes an indicator of this is having if-else statements like this one, where we do different different things based on some condition. Instead of doing that, shall we try to do the following:
- Identify the core, high level flow of steps this method is supposed to do to find the references
- Identify the input required for those steps. 
- Identify which of these input need to be passed down to this method. 
- Make those inputs params of the method and move the logic for deriving those input to the relevant overloaded version which calls this method.
e.g., In this if case, you are handling the 2 cases where in one you have the symbol and the other where you have to look it up. Instead of doing that, maybe pass the symbol as a param?

dulajdilshan: Aren't we introduce unnecessary methods with this approach?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -183,40 +183,84 @@ public BallerinaSemanticModel(BLangPackage bLangPackage, CompilerContext context
...
+        return getReferences(symbolAtCursor, node, true);
     }
 
     @Override
     public List<Location> references(Symbol symbol, boolean withDefinition) {
+        BSymbol symbolAtCursor = getInternalSymbol(symbol);
         Optional<Location> symbolLocation = symbol.getLocation();
+        if (symbolAtCursor == null || symbolLocation.isEmpty()) {
+            return Collections.emptyList();
+        }
```
pubudu91, line 201: A couple of points. 
- Can check whether there's a location as the first check since there's no point in getting the internal symbol if there isn't a location.
- I think there's no need for a null check for `symbolAtCursor`. Every `Symbol` instance should have an associated internal symbol.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -183,40 +183,84 @@ public BallerinaSemanticModel(BLangPackage bLangPackage, CompilerContext context
...
 
-        ReferenceFinder refFinder = new ReferenceFinder(withDefinition);
-        return refFinder.findReferences(node, getInternalSymbol(symbol));
+    @Override
+    public List<Location> references(Symbol symbol, Document targetDocument, boolean withDefinition) {
+        BSymbol symbolAtCursor = getInternalSymbol(symbol);
+        Optional<Location> symbolLocation = symbol.getLocation();
+        if (symbolAtCursor == null || symbolLocation.isEmpty()) {
+            return Collections.emptyList();
+        }
```
pubudu91, line 225: Same comment as above for this.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -183,40 +183,84 @@ public BallerinaSemanticModel(BLangPackage bLangPackage, CompilerContext context
...
+        return symbolFinder.lookup(sourceCompilationUnit, linePosition);
+    }
+
+    private BLangNode findEnclosingContainerNode(LineRange lineRange, Document targetDocument) {
+        return targetDocument == null ?
+                new NodeFinder(false)
+                        .lookupEnclosingContainer(this.bLangPackage, lineRange) :
+                new NodeFinder(false)
+                        .lookupEnclosingContainer(getCompilationUnit(targetDocument), lineRange);
+    }
```
pubudu91, line 259: This method is not required right? This is basically a one-liner, the case to be used depends on the particular `references()` version

mohanvive (approved):
LGTM.

mohanvive:
@pubudu91 shall we merge this?

pubudu91 (approved):

nadeeshaan (approved):
LGTM

Issue #1616, opened 2017-02-08
Connectors does not properly place in the service and resource.


hemikak:
Add multiple connectors to the service and resource level. See attachment.
<img width="1226" alt="screen shot 2017-02-08 at 1 49 15 pm" src="https://cloud.githubusercontent.com/assets/4398820/22729468/e09039c6-ee08-11e6-9c6d-da0ad41613c9.png">

PR #32270, opened 2021-08-22
Implement uniqueId for `ServiceDeclarationSymbol`

ayeshLK:
## Purpose
As an effort to support REST API development we are implementing an introspection endpoint for any ballerina HTTP service and that particular introspection endpoint will be serving a OpenAPI Doc for that particular service. But in-order to achieve this we need to identify service declaration via a uniqueId which could also be accessed at runtime.

Related to [#1616](https://github.com/ballerina-platform/ballerina-standard-library/issues/1616)
Related to [#1776](https://github.com/ballerina-platform/ballerina-standard-library/issues/1776)
Fixes [#32324](https://github.com/ballerina-platform/ballerina-lang/issues/32324)

## Approach
We will take following steps to implement this feature:
- Override `hashCode` and `equals` methods in `BallerinaServiceDeclarationSymbol`.
- Construct that same hashCode value for service-class-definition in annotation-desugar phase and attach new annotation with required information to that class-definition.
- Newly introduced annotation has following format.
```
# Contains the configurations related to service introspection. This is used by HTTP module.
#
# + name - The generated Open API spec document name
type ServiceIntrospectionDocConfig record {|
    string name;
|};

# The annotation which is used to configure a HTTP service.
annotation ServiceIntrospectionDocConfig IntrospectionDocConfig on service;
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
```diff
@@ -216,4 +218,42 @@ public BallerinaServiceDeclarationSymbol build() {
...
+    }
+
+    @Override
+    public boolean nameEquals(String name) {
+        String symbolName = this.getInternalSymbol().getName().getValue();
+        if (name.equals(symbolName)) {
+            return true;
+        }
+        return unescapedUnicode(name).equals(unescapedUnicode(symbolName));
+    }
```
pubudu91, line 250: This method is for comparing the actual name of a symbol. Since we don't have a public name for service symbols, let's not override this.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
```diff
@@ -216,4 +218,42 @@ public BallerinaServiceDeclarationSymbol build() {
...
+            return true;
+        }
+
+        if (!(obj instanceof ServiceDeclarationSymbol)) {
+            return false;
+        }
+
+        BallerinaServiceDeclarationSymbol symbol = (BallerinaServiceDeclarationSymbol) obj;
+        return nameEquals(symbol.getInternalSymbol().getName().getValue())
+                && isSameLocation(this.getLocation(), symbol.getLocation());
```
pubudu91, line 240: Checking `isSameLocation()` is not enough. We have to check whether the module is the same too since the location is given relative to the module directory. Hence, there can be instances where locations are equal even though the 2 symbols are from different modules.

Also, let's change the visibility of the `isSameLocation()` and `isSameModule()` to `protected` and reuse them.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaServiceDeclarationSymbol.java
```diff
@@ -216,4 +218,42 @@ public BallerinaServiceDeclarationSymbol build() {
                                                          this.qualifiers, this.annots, this.bSymbol, this.context);
         }
     }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(this.getInternalSymbol().getName().getValue(),
+                this.getLocation().map(Location::lineRange).orElse(null));
```
pubudu91, line 225: For the hash code also best to consider the module as well.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
```diff
@@ -191,6 +199,84 @@ private void defineClassAnnotations(BLangPackage pkgNode, SymbolEnv env, BLangFu
...
+    }
+
+    private void addIntrospectionInfoAnnotation(BLangClassDefinition serviceClass) {
+        Location position = serviceClass.pos;
+
+        // Create Annotation Attachment.
+        BLangAnnotationAttachment annoAttachment = (BLangAnnotationAttachment) TreeBuilder.createAnnotAttachmentNode();
+        final SymbolEnv pkgEnv = symTable.pkgEnvMap.get(serviceClass.symbol.getEnclosingSymbol());
+        BSymbol annSymbol = symResolver.lookupSymbolInAnnotationSpace(
+                pkgEnv, names.fromString(SERVICE_INTROSPECTION_INFO_ANN));
```
MaryamZi, line 215: Given that we know where this annotation is defined, wondering if we should directly do

```suggestion
        symResolver.lookupSymbolInAnnotationSpace(symTable.pkgEnvMap.get(symTable.rootPkgSymbol), 
                                                  names.fromString(SERVICE_INTROSPECTION_INFO_ANN));
```

MaryamZi: @pubudu91 thoughts please?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
```diff
@@ -191,6 +199,84 @@ private void defineClassAnnotations(BLangPackage pkgNode, SymbolEnv env, BLangFu
...
+        annoAttachment.annotationName = identifierNode;
+        annoAttachment.annotationName.value = SERVICE_INTROSPECTION_INFO_ANN;
+        annoAttachment.pos = position;
+        annoAttachment.annotationName.pos = position;
+
+        // create record literal related to annotation node
+        BLangRecordLiteral literalNode = (BLangRecordLiteral) TreeBuilder.createRecordLiteralNode();
+        annoAttachment.expr = literalNode;
+        BLangIdentifier pkgAlias = (BLangIdentifier) TreeBuilder.createIdentifierNode();
+        pkgAlias.setValue(BUILTIN_PKG_KEY);
```
MaryamZi, line 230: We don't have a builtin package now. We need to use the correct module.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
```diff
@@ -191,6 +199,84 @@ private void defineClassAnnotations(BLangPackage pkgNode, SymbolEnv env, BLangFu
         }
     }
 
+    private boolean isServiceDeclaration(BLangClassDefinition serviceClass) {
+        return Objects.nonNull(serviceClass.getFlags())
```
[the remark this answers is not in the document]
ayeshLK: Nope, will remove unwanted validation.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
pubudu91: Yeah, I guess we can do that

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
```diff
@@ -191,6 +199,84 @@ private void defineClassAnnotations(BLangPackage pkgNode, SymbolEnv env, BLangFu
...
+        if (valueLiteral != null) {
+            descriptorKeyValue.valueExpr = valueLiteral;
+        }
+
+        // add generated annotation to the service definition
+        serviceClass.addAnnotationAttachment(annoAttachment);
+    }
+
+    private String generateServiceHashCode(BLangClassDefinition serviceClass) {
+        String serviceName = "service$" + serviceClass.name.getValue();
```
pubudu91, line 274: Is this name you generate and the name you use in the hash code impl of `BallerinaServiceDeclarationSymbol` the same? 🤔

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
```diff
@@ -166,6 +171,9 @@ private void defineClassAnnotations(BLangPackage pkgNode, SymbolEnv env, BLangFu
             }
 
             BLangClassDefinition classDefinition = (BLangClassDefinition) topLevelNode;
+            if (isServiceDeclaration(classDefinition)) {
+                addIntrospectionInfoAnnotation(classDefinition);
```
pubudu91, line 175: @rdhananjaya Usually when annotations are added to a service decl, it seems it gets added to both the `BLangService` node and the corresponding `BLangClassDefinition` node. But here it seems the annotation is added only to the generated class. Is this ok? 🤔

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
ayeshLK: Yes, `serviceClass.name.getValue()` just have the `$anonType$_` + `<auto-incremented-value>` value. So, we prepend `service$` part.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
ayeshLK: For compiler generated annotations (`DefaultableArgs` annotation etc.) `ballerina/builtin` package has been used. Since, this is also a compiler generated one we have added it.

MaryamZi: I think that just hasn't been updated with the builtin change.

ayeshLK: Since we are attaching this annotation for any service-declarations what would be the correct package name to be used ?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
rdhananjaya: This should be OK as they only need this at runtime. I think with this we will miss this annotation in BIR.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/AnnotationDesugar.java
MaryamZi: This pkgAlias seems to be for the module in which the annotation is defined. So in this case I guess it would have to be the lang.annotations module.

ayeshLK:
Since this feature is required for `SL-Beta3` we have sent a separate PR ([#32328](https://github.com/ballerina-platform/ballerina-lang/pull/32328)) to `2.0.0-beta3-dev` branch with this change.

PR #32275, opened 2021-08-23
[Semantic API] Create bounded symbols for lang lib functions 

### LINKED ISSUE #18150: VSCode Hover suggestions on Langlib method invocation needs to be improved with bound type
**Description:**
Right now it shows actual type.  e.g:
```
string[] a = [];
a.push();   // <- Suggestions say any|error, but actually it should be a string.  
```

pubudu91:
## Purpose
Previously, when getting the lang lib functions associated with a type, we used to return the general function symbol. i.e., the type params were not resolved. With this PR, we are addressing this issue to resolve the type param associated with the first param in functions which can be called using a method call expression.

Fixes #18150

## Approach
For each of the `BInvokableSymbol` instances of a lang lib, the following is done to obtain a type param resolved instance (a copy) of it. 
- Given the type for which the lang libs are required, get the component of it which would be the resolved type value of the type param containing component.
- Then use the `TypeParamFinder` to find the type param component of the first param's type.
- If the first param's type is a type param type, proceed with the type resolving and symbol duplication.
- For each of the params and return type, use the `TypeParamResolver` to resolve each param's type's type params by binding them to the previously derived bound type. Note that this binding is done rather blindly. If there are type params, new types are created with the type params resolved and new var symbols are created as well for the params. 
- These params and return type are set to the duplicated function symbol and returned.

type-param

## Remarks
- This only resolves a single type param. In typedescs which have multiple type params, only the type param for the more common case is considered e.g., stream and table typedescs
- Complex types such as object and error types are not considered for the moment
- `Cloneable` type related type params are not supported yet. Need to figure out how to handle them due to their cyclic nature. Attempting to allow them with the current design results in stack overflow errors.
- Some of the tests for the function typed params of lang lib functions are commented due to #32398

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/ballerina-compiler-api-test/src/test/resources/test-src/type_param_bound_langlib_functions.bal
```diff
@@ -0,0 +1,37 @@
...
+type PersonObj object {
+    string name;
+
+    function getName() returns string;
+};
+
+type Person record {|
+    int id;
+    string name;
+|};
```
mohanvive, line 37: Add a line break.

mohanvive (dismissed):
Overall PR looks good as the first step. Shall we create an issue and list down the todo items for the future? WDYT? @pubudu538

mohanvive (approved):

Issue #32317, opened 2021-08-24
Add basic class structure reflecting nballerina semtype implementation


rdhananjaya:
**Description:**
$title.

**Describe your task(s)**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #32318, opened 2021-08-24
Add initial code structure for semtype port

rdhananjaya:
## Purpose
$title

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/32317

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/semtypes/src/main/java/io/ballerina/semtype/Env.java
+++ b/semtypes/src/main/java/io/ballerina/semtype/Env.java
@@ -17,5 +17,23 @@
  */
 package io.ballerina.semtype;
 
-public interface SemType {
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class Env {
+    private final HashMap<AtomicType, Atom> atomTable;
+    private final ArrayList<ListAtomicType> recListAtoms;
+    private final ArrayList<MappingAtomicType> recMappingAtoms;
+    private final ArrayList<FunctionAtomicType> recFunctionAtoms;
+
+    public Env() {
+        this.atomTable = new HashMap<>();
+        this.recListAtoms = new ArrayList<>();
+
+        this.recMappingAtoms = new ArrayList<>();
+
+        this.recFunctionAtoms = new ArrayList<>();
+    }
+
+
 }
--- a/semtypes/src/main/java/io/ballerina/semtype/PredefinedType.java
+++ b/semtypes/src/main/java/io/ballerina/semtype/PredefinedType.java
@@ -1,0 +1,73 @@
+/*
+ *  Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  WSO2 Inc. licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package io.ballerina.semtype;
+
+import io.ballerina.semtype.subtypedata.IntSubtype;
+
+// todo: Should we rename this to `Core`?
+public class PredefinedType {
+    public static final UniformTypeBitSet NEVER = uniformTypeUnion(0);
+    public static final UniformTypeBitSet NIL = uniformType(UniformTypeCode.UT_NIL);
+    public static final UniformTypeBitSet BOOLEAN = uniformType(UniformTypeCode.UT_BOOLEAN);
+    public static final UniformTypeBitSet INT = uniformType(UniformTypeCode.UT_INT);
+    public static final UniformTypeBitSet FLOAT = uniformType(UniformTypeCode.UT_FLOAT);
+    public static final UniformTypeBitSet DECIMAL = uniformType(UniformTypeCode.UT_DECIMAL);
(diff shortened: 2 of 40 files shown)


semtypes/src/main/java/io/ballerina/semtype/PredefinedType.java
```diff
@@ -0,0 +1,73 @@
...
+    public static final UniformTypeBitSet SIMPLE_OR_STRING =
+            uniformTypeUnion((1 << UniformTypeCode.UT_NIL)
+                    | (1 << UniformTypeCode.UT_BOOLEAN)
+                    | (1 << UniformTypeCode.UT_INT)
+                    | (1 << UniformTypeCode.UT_FLOAT)
+                    | (1 << UniformTypeCode.UT_DECIMAL)
+                    | (1 << UniformTypeCode.UT_STRING));
+    public static final SemType BYTE = IntSubtype.intWidthUnsigned(8);
+
+    private static UniformTypeBitSet uniformTypeUnion(long bitset) {
```
jclark, line 62: Should be int

semtypes/src/main/java/io/ballerina/semtype/RecAtom.java
```diff
@@ -0,0 +1,30 @@
...
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package io.ballerina.semtype;
+
+public class RecAtom implements Atom {
+    long index;
```
jclark, line 21: Should be int

semtypes/src/main/java/io/ballerina/semtype/UniformTypeCode.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Represent bit field that indicate which uniform type a semType belongs to.
+ * Regular types are divided longo mutable part and immutable part and these parts are called an uniform type.
+ * 5th bit indicate mutability; 0 immutable, 1 mutable
+ *
+ * @since 2.0.0
+ */
+public class UniformTypeCode {
+    // Inherently immutable
+    public static final long UT_NIL = 0x00;
```
jclark, line 29: These should all be int. UniformTypeBitSet is a Unsigned32, but int is fine too.

semtypes/src/main/java/io/ballerina/semtype/Env.java
```diff
@@ -0,0 +1,39 @@
...
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package io.ballerina.semtype;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class Env {
+    private final HashMap<AtomicType, Atom> atomTable;
```
jclark, line 24: Should be `HashMap<AtomicType,TypeAtom>`

jclark (commented):
This all looks good to me. Most (all?) of the longs should be ints (since UniformTypeBitSet is 32 bits and indices in Java have to be 32-bit).

semtypes/src/main/java/io/ballerina/semtype/UniformTypeOps.java
```diff
@@ -0,0 +1,27 @@
...
+ *  under the License.
+ */
+package io.ballerina.semtype;
+
+public interface UniformTypeOps {
+    SubtypeData union(SubtypeData t1, SubtypeData t2);
+    SubtypeData intersect(SubtypeData t1, SubtypeData t2);
+    SubtypeData diff(SubtypeData t1, SubtypeData t2);
+    SubtypeData complement(SubtypeData t);
+    boolean isEmpty(TypeCheckContext tc, SubtypeData t);
```
jclark, line 25: I suspect it will work out better to have an interface with just the first 4 and then have another interface that adds isEmpty. Then you can have an abstract class that implements the first interface for Bdd.

jclark (dismissed):

semtypes/src/main/java/io/ballerina/semtype/RecAtom.java
rdhananjaya: Java `int` is 32bit, so this should be `long` I think.

manuranga: Only 23 bits are used https://github.com/ballerina-platform/nballerina/blob/fb8c9d506b1c09b267f7aec0ace364f709c969bf/compiler/modules/types/core.bal#L6
so int should work.

semtypes/src/main/java/io/ballerina/semtype/UniformTypeOps.java
rdhananjaya: We already have a similar hierarchy here (just without interface segregation) https://github.com/ballerina-platform/ballerina-lang/blob/9cb56bbe2640a68c7a3124f8a53a04a04cfd6245/semtypes/src/main/java/io/ballerina/semtype/typeops/CommonOps.java#L23

rdhananjaya: Oh please disregards the comment, I get what you meant.

semtypes/src/main/java/io/ballerina/semtype/RecAtom.java
manuranga: Ahh this is a different long, I was commenting about UniformTypeCode's long.

jclark: Number of members in a Java collection cannot exceed what can be represented in an int, so I think it should be int.

rdhananjaya: ahh yes, agreed. Thank you.

PR #32385, opened 2021-08-26
[LS] Refactor default value handling of optional types

### LINKED ISSUE #31790: Refactor default value handling of optional types
**Description:**

Modify `CommonUtil.getDefaultValueForType()` method to return an `Optional<String>` where it will return `Optional.empty()` when the type is an optional one.

Also add a new API to return a default value regardless of the type being an optional type.

**Steps to reproduce:**

**Affected Versions:**

TharushiJay:
## Purpose

Fixes #31790

## Approach
Modify `CommonUtil.getDefaultValueForType()` method to return an `Optional<String>`.

## Samples

As a result, it adds no return statement when generating a function with `Implement method` code action where the default return value is `()`. But for record fields where the default value is `()`, it adds `fieldName : ()`.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
@@ -248,15 +248,15 @@ public class CommonUtil {
      * @param bType Type descriptor to get the default value
      * @return {@link String}   Default value as a String
      */
-    public static String getDefaultValueForType(TypeSymbol bType) {
+    public static Optional<String> getDefaultValueForType(TypeSymbol bType) {
         return getDefaultValueForType(bType, 1);
     }
     
-    private static String getDefaultValueForType(TypeSymbol bType, int depth) {
+    private static Optional<String> getDefaultValueForType(TypeSymbol bType, int depth) {
         String typeString;
         
         if (bType == null) {
-            return "";
+            return Optional.empty();
         }
 
         TypeSymbol rawType = getRawType(bType);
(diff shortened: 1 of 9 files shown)


mohanvive (approved):

mohanvive:
@imesh waiting on your review. Can you please check and merge.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
```diff
@@ -303,15 +304,15 @@ private static String getDefaultValueForType(TypeSymbol bType, int depth) {
                 break;
             case OBJECT:
                 if (depth > MAX_DEPTH) {
-                    return "";
+                    return Optional.of("");
```
malinthar, line 307: I think we should send `new ()` in this case as we have done for record type, WDYT?

malinthar (commented):
Changes looks good. Have a minor comment

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
TharushiJay: In record types we return a `{}` to denote a complex record beyond level 1 and it's inserted after `Fieldname : ` (line 297). Hence it is necessary. But in the case of an object, `new()` is separately added in line 317/ 319 to the empty string returned from here. Returning `new ()` here explicitly will result in `new(new())`. Hence it is not necessary.

IMS94: Makes sense. Shall we do it? @TharushiJay

IMS94 (approved):
LGTM! Let's address @malinthar 's comment and merge this.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
IMS94: Sorry, missed @TharushiJay's reply. If that's the case, let's let this be as it is.

PR #32449, opened 2021-08-30
Fix issue with invalid qualifier order

### LINKED ISSUE #32445: Invalid text range for token
**Description:**
Consider the following source code.
```
function myFunc {
    isolated client 
}
```
The text range of client token is as follows.
tec

Irrespective of where the above code segment is used. The text range of client token always seem to be as depicted above.

**Steps to reproduce:**
Mentioned above

lochana-chathura:
## Purpose
$subject.

Fixes #32445 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
+++ b/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
@@ -4618,7 +4618,8 @@ public class BallerinaParser extends AbstractParser {
     }
 
     private STNode modifyNodeWithInvalidTokenList(List<STNode> qualifiers, STNode node) {
-        for (STNode qualifier : qualifiers) {
+        for (int i = qualifiers.size() - 1; i >= 0; i--) {
+            STNode qualifier = qualifiers.get(i);
             node = SyntaxErrors.cloneWithLeadingInvalidNodeMinutiae(node, qualifier);
         }
         return node;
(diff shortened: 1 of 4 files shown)


compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -4618,7 +4618,8 @@ private STNode createMissingSimpleObjectField(STNode metadata, List<STNode> qual
     }
 
     private STNode modifyNodeWithInvalidTokenList(List<STNode> qualifiers, STNode node) {
-        for (STNode qualifier : qualifiers) {
+        for (int i = qualifiers.size() - 1; i >= 0; i--) {
+            STNode qualifier = qualifiers.get(i);
```
rdulmina, line 4622: R we having the same issue with `reportInvalidQualifierList` as well?

lochana-chathura: Inside `reportInvalidQualifierList()` we call `addInvalidNodeToNextToken()` for each token, which keeps the tokens in a stack. Since stack is FILO order we get the right order there.

rdulmina: Cool

rdulmina (approved):

PR #32248, opened 2021-08-20
Handle immutable typedesc within toStream

pcnfernando:
Fixes #32244

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #32469, opened 2021-08-31
Fix langlib toStream() method issue on tuples

### LINKED ISSUE #32265: Issue with `toStream()` on tuples and finite (singleton) types
**Description:**
$title.

**Steps to reproduce:**
**Case I**
```ballerina
public function main() {
    [int, string] x = [1, "hello"];

    stream<int|string> y = x.toStream();
    stream<int|string> z = <stream<int|string>> y; // Panics - incompatible types: 'stream<()>' cannot be cast to 'stream<(int|string)>'
}
```

**Case II**
```ballerina
type Type [1, "hello"]|[2];

public function main() {
    Type x = [1, "hello"];

    stream<1|"hello"|2> y = x.toStream();
    stream<1|"hello"|2> z = <stream<1|"hello"|2>> y; // Panics - incompatible types: 'stream<()>' cannot be cast to 'stream<>'
}
```

**Affected Versions:**
slbeta2, slbeta3-SNAPSHOT

Nadeeshan96:
$title

Fixes #32265

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.__internal/src/main/java/org/ballerinalang/langlib/internal/GetElementType.java
```diff
@@ -42,16 +43,23 @@ public static BTypedesc getElementType(Object td) {
...
+            case TypeTags.STREAM_TAG:
+                return ValueCreator.createTypedescValue(((StreamType) type).getConstrainedType());
+            case TypeTags.FINITE_TYPE_TAG:
+                if (((FiniteType) type).getValueSpace().size() == 1) {
+                    return getTypeDescValue(
+                            ((BValue) (((FiniteType) type).getValueSpace().iterator().next())).getType());
+                }
+                break;
+            default:
+                break;
```
warunalakshitha, line 61: Should this be an error?

langlib/lang.__internal/src/main/java/org/ballerinalang/langlib/internal/GetElementType.java
```diff
@@ -42,16 +43,23 @@ public static BTypedesc getElementType(Object td) {
     }
 
     private static BTypedesc getTypeDescValue(Type type) {
```
pcnfernando, line 45: Shall we rename the method name as `getElementTypeDescValue`. ref:https://github.com/ballerina-platform/ballerina-lang/pull/32248#discussion_r692872819

Nadeeshan96: I will do this.

langlib/lang.__internal/src/main/java/org/ballerinalang/langlib/internal/GetElementType.java
Nadeeshan96: Due to `public isolated function getElementType(typedesc<Type[]> | typedesc<stream<Type, CompletionType>> td)` I think only the above 4 types are allowed to come inside the function. So `return ValueCreator.createTypedescValue(PredefinedTypes.TYPE_NULL);` is never executed anyway. Should I throw an error instead of that line like https://github.com/ballerina-platform/ballerina-lang/pull/32257/files#r697291327 or may I remove the line?

Nadeeshan96: @pcnfernando Can you please tell why the `(((FiniteType) type).getValueSpace().size() == 1)` is required. In the samples I tested, the cases which came under the `FINITE_TYPE_TAG` had this property. Are there any case which does not have this property?

What I am trying to do is to increase the code coverage, change the `case TypeTags.FINITE_TYPE_TAG:` to the `default case` and only put `return getElementTypeDescValue(((BValue) (((FiniteType) type).getValueSpace().iterator().next())).getType());` there. I would like to put `assert ((((FiniteType) type).getValueSpace().size() == 1) && type.getTag() == TypeTags.FINITE_TYPE_TAG);` but that does not get executed without the `-enableassertions` java cli option.

Nadeeshan96: @pcnfernando Could you please tell why the `if (((FiniteType) type).getValueSpace().size() == 1)` condition is required. In the samples I tested which came under the case `FINITE_TYPE_TAG`, all of them had this property. Is there any case which does not have that property?

What I am trying to do here is to increase the code coverage changing the `case TypeTags.FINITE_TYPE_TAG:` to `default case` and put `return getElementTypeDescValue(
                            ((BValue) (((FiniteType) type).getValueSpace().iterator().next())).getType());` under that. I would like to put `assert ((((FiniteType) type).getValueSpace().size() == 1) && type.getTag() == TypeTags.FINITE_TYPE_TAG);` before that but that does not get executed without `-enableassertions` java clip option.

pcnfernando: Finite case itself was added to handle the change introduced for typedesc with immutable values in https://github.com/ballerina-platform/ballerina-lang/pull/32105. Hence, we can assume its value space is always 1. Please add a comment for finite case saying that it's reached only for immutable values.
Since this method should accept array/tuple and streams alone through ballerina, no need to throw the error. Use stream type as the default case.

Nadeeshan96: I will make these changes. Thank you!

langlib/lang.__internal/src/main/java/org/ballerinalang/langlib/internal/GetElementType.java
```diff
@@ -36,22 +36,27 @@
...
+        switch (type.getTag()) {
+            case TypeTags.ARRAY_TAG:
+                return ValueCreator.createTypedescValue(((ArrayType) type).getElementType());
+            case TypeTags.TUPLE_TAG:
+                return ValueCreator.createTypedescValue(
+                        TypeCreator.createUnionType(((TupleType) type).getTupleTypes()));
+            case TypeTags.FINITE_TYPE_TAG:
+                // this is reached only for immutable values
+                return getElementTypeDescValue(
+                        ((BValue) (((FiniteType) type).getValueSpace().iterator().next())).getType());
```
Nadeeshan96, line 57: It seems that due to the recent changes to the beta 3 branch read-only values also goes to other 3 cases.

langlib/lang.__internal/src/main/java/org/ballerinalang/langlib/internal/GetElementType.java
```diff
@@ -36,22 +34,23 @@
...
-    private static BTypedesc getTypeDescValue(Type type) {
-        if (type.getTag() == TypeTags.ARRAY_TAG) {
-            return ValueCreator.createTypedescValue(((ArrayType) type).getElementType());
-        } else if (type.getTag() == TypeTags.STREAM_TAG) {
-            return ValueCreator.createTypedescValue(((StreamType) type).getConstrainedType());
-        } else if (type.getTag() == TypeTags.TABLE_TAG) {
-            return ValueCreator.createTypedescValue(((TableType) type).getConstrainedType());
-        } else if (type.getTag() == TypeTags.FINITE_TYPE_TAG && ((FiniteType) type).getValueSpace().size() == 1) {
-            return getTypeDescValue(((BValue) (((FiniteType) type).getValueSpace().iterator().next())).getType());
+    private static BTypedesc getElementTypeDescValue(Type type) {
```
pcnfernando, line 45: Is the immutable case already covered in the tests? If not let's add a test

Nadeeshan96: It seems it is already covered in the test you have written.
https://github.com/Nadeeshan96/ballerina-lang/blob/0446326d33708a48f644fc1b2a5385028f29074f/langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/LangLibArrayTest.java#L576 
https://github.com/Nadeeshan96/ballerina-lang/blob/0446326d33708a48f644fc1b2a5385028f29074f/langlib/langlib-test/src/test/resources/test-src/arraylib_test.bal#L1296

pcnfernando: Let's add one for the tuple case

Nadeeshan96: I will do it.

pcnfernando (approved):

PR #32479, opened 2021-08-31
[WIP] Narrow the inherent type of immutable container values 

### LINKED ISSUE #13189: [Spec Deviation] Freezing a container should narrow its inherent type to a type that consists of just its current shape
**Description:**
According to the spec (v0.990:2019-01-16 - page 17), "A frozen container value belongs to a type if and only if the type contains the shape of the value. Thus after a container value is frozen, its inherent type does not provide additional information that cannot be derived from the value. In other words, freezing a container **narrows its inherent type to a type that consists of just its current shape**"

The `is` check works as expected after freezing a value:
```ballerina
import ballerina/io;

    map<string|int> m1 = { str: "string only" };
    anydata a = m1;  
    io:println(a is map<string>); // false

    _ = a.freeze();
    io:println(a is map<string>); // true
}
```

But when trying to use such a value in the context of the narrowed type, the original type is still considered.
```ballerina
import ballerina/io;

    map<string|int> m1 = { str: "string only" };
    anydata a = m1;

    _ = a.freeze();
    io:println(a is map<string>); // true

    map<map<string>> m2 = {};
    insertMemberToMap(m2, "strMap", a); // panics with {ballerina}InherentTypeViolation {"message":"invalid map insertion: expected value of type 'map<string>', found 'map<string|int>'"}
}

public function insertMemberToMap(map<any> mapVal, string index, any member) {
    mapVal[index] = member;
}
```

Casting/assertion also fails similarly.

**Affected Versions:**
0.990.2

HindujaB:
## Purpose
$subject

Fixes #13189 

## Approach
Implemented methods to create a singleton type containing the immutable values. 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/SingletonUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/SingletonUtils.java
@@ -1,0 +1,265 @@
+/*
+ *  Copyright (c) 2021, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ *  WSO2 Inc. licenses this file to you under the Apache License,
+ *  Version 2.0 (the "License"); you may not use this file except
+ *  in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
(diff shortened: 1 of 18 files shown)


HindujaB:
This fix need to be revisited once we finalised the implementation for singleton type.

PR #32540, opened 2021-09-02
Fix syntax error in source generated by type guard code action

### LINKED ISSUE #32313: Code Action generates syntactically incorrect source
**Description:**
Refer the following screen capture. Code action executed against the `nBallerina` repo.

https://user-images.githubusercontent.com/1329674/130611780-b1cd4b9e-c3dd-47cb-8d13-1269c8973903.mp4

TharushiJay:
## Purpose
Fixes generating a syntactically incorrect source by the type guard code action.

Fixes #32313 

## Samples
Before :
Screenshot 2021-09-02 at 10 59 10

After : 
Screenshot 2021-09-02 at 10 58 45

## Check List 
- [ ] Updated Change Log
- [X] Added necessary tests
  - [X] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
@@ -70,11 +70,13 @@ import org.eclipse.lsp4j.Range;
 import org.eclipse.lsp4j.TextEdit;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.StringJoiner;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static org.ballerinalang.langserver.common.utils.CommonUtil.LINE_SEPARATOR;
@@ -398,7 +400,14 @@ public class CodeActionUtil {
             }
             List<String> memberTypes = new ArrayList<>();
             for (TypeSymbol tMember : tMembers) {
-                memberTypes.add(CodeActionUtil.getPossibleType(tMember, edits, context).orElseThrow());
+                String possibleType = CodeActionUtil.getPossibleType(tMember, edits, context).orElseThrow();
+                if (tMember.typeKind() == TypeDescKind.UNION || tMember.typeKind() == TypeDescKind.SINGLETON) {
+                    String[] parts = possibleType.split(Pattern.quote("|"));
+                    possibleType = Arrays.stream(parts)
+                            .map(s -> "\"" + s + "\"")
+                            .collect(Collectors.joining("| "));
+                }
+                memberTypes.add(possibleType);
             }
             if (addErrorTypeAtEnd) {
                 memberTypes.add("error");
--- a/language-server/modules/langserver-core/src/test/resources/codeaction/type-guard/source/typeGuard2.bal
+++ b/language-server/modules/langserver-core/src/test/resources/codeaction/type-guard/source/typeGuard2.bal
@@ -1,0 +1,10 @@
+type Token SingleCharDelim|MultiCharDelim|Keyword;
+type SingleCharDelim "+" | "-" ;
+type MultiCharDelim "<<" | ">>" | CompoundAssignDelim;
+type CompoundAssignDelim "+=" | "-=" ;
+type Keyword KEY | "string";
+const KEY = "int";
+
+public function main(string... args) {
+    Token? cur = getToken();
+}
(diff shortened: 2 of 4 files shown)


language-server/modules/langserver-core/src/test/resources/codeaction/type-guard/source/typeGuard2.bal
```diff
@@ -0,0 +1,10 @@
+type Token SingleCharDelim|MultiCharDelim|Keyword;
+type SingleCharDelim "+" | "-" ;
+type MultiCharDelim "<<" | ">>" | CompoundAssignDelim;
+type CompoundAssignDelim "+=" | "-=" ;
+type Keyword KEY | "string";
```
IMS94, line 5: Let's change this to `type Keyword KEY | "string" | string | 100;`

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
```diff
@@ -398,7 +400,14 @@ public static DiagBasedPositionDetails computePositionDetails(SyntaxTree syntaxT
             }
             List<String> memberTypes = new ArrayList<>();
             for (TypeSymbol tMember : tMembers) {
-                memberTypes.add(CodeActionUtil.getPossibleType(tMember, edits, context).orElseThrow());
+                String possibleType = CodeActionUtil.getPossibleType(tMember, edits, context).orElseThrow();
+                if (tMember.typeKind() == TypeDescKind.UNION || tMember.typeKind() == TypeDescKind.SINGLETON) {
```
IMS94, line 404: @pubudu91 In this case, we have a union type like below:
```
type t1 string| int | t2;
type t2 "int" | "string" | 10;

function fn() {
   t1? t = 10;<cursor>
}
```
At cursor we suggest type guard and we need to make sure the generated type guards use proper types/values like:
```
if t is string {
} else if t is "string" {
} else if t is 10 {
}...
```
Can you confirm if there's any issue in this approach?

IMS94 (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
pubudu91, line 404: For what sort of scenarios would the `tMember` be a union type? 🤔

IMS94:
```
type t t1|t2;
type t1 "a"|"b"|"c";
type t2 "p"|"q"|"r";
```
If we are looking at `t`, I we get `"a"|"b"|"c"` as a union type. @TharushiJay can you confirm?

TharushiJay: Consider the scenario, 
type t "0"|"1"|t1;
type t1 "a"|"b"|"c";

TharushiJay: Consider the scenario,
```
type t "0"|"1"|t1;
type t1 "a"|"b"|"c";
```
Screenshot 2021-09-02 at 20 40 52

Here, we don't get `t` in `tMembers`. Instead we get `"0"|"1"` as a union type. That is, when the type definition has both, more than 1 terminal and at least 1 non-terminal.

malinthar:
Shall we address the following as well.

https://user-images.githubusercontent.com/35211477/131904269-0de55836-3174-4fc4-b213-109a38c989cf.mp4

Please be aware of this issue when you check out the above as well. 
https://github.com/ballerina-platform/ballerina-lang/issues/32556

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
pubudu91: Hmm. That's an issue. This should ideally have been a union type containing 3 members: 2 singleton types and a type ref. Will check this and see.

IMS94:
@malinthar +1. I think that will be part of what @pubudu91 mentioned in the above comment. We should be consistent when introducing the type guard, either use the top level member types or simplified member types (desugared).

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
IMS94: If that's the case, we need to get that fixed. @TharushiJay shall we create an issue?

language-server/modules/langserver-core/src/test/resources/codeaction/type-guard/source/typeGuard2.bal
TharushiJay: Current fix doesn't work for this example (adds quotation marks around 100 as well). Therefore, blocked until #32569 is fixed.

nadeeshaan:
> Shall we address the following as well.
> 
>  demo-article_00007.mp4 
> Please be aware of the following when you check out the above issue
> #32556

@malinthar The concern with #32556 is because of the syntax tree. The parser is providing an invalid parent node for the field access expression.

PR #32541, opened 2021-09-02
Improve accuracy of completions using diagnostic state by filtering out invalid variables

### LINKED ISSUE #32038: Improve accuracy of completions by filtering invalid variables using diagnostic state
**Description:**
With #31072, a new API was added to get the visible symbols filtered by their diagnostic state (valid, redeclared, unknown). Need to incorporate this API in LS completions to improve completions.

This can be used to remove suggesting variables being declared in completions. For example:
```
int count = <cursor>
```
Right now, at `<cursor>` we suggest, `count` which is being declared. We can avoid showing that using the above API.

**Describe your task(s)**
See description

#30081
PR: #31072

gayalkuruppu:
Fixes #32038 

## Samples
Consider the following example.
```import ballerina/module1;

function testFunction() {
int value2= 12;
    var testVar = <cursor>
    int value1 = 12;
}

# This is a test function
#
# + return - Return Value Description
function functionWithReturn() returns Record1 {
    return {};
}

type Record1 record {
    module1:TestRecord1 testField = {};
    int testOptional?;
};
```
With this fix we remove completion suggestion of `testVar` at Cursor 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
@@ -17,10 +17,8 @@
  */
 package org.ballerinalang.langserver.completions.providers;
 
-import io.ballerina.compiler.api.SemanticModel;
 import io.ballerina.compiler.api.symbols.ClassSymbol;
 import io.ballerina.compiler.api.symbols.ConstantSymbol;
-import io.ballerina.compiler.api.symbols.DiagnosticState;
 import io.ballerina.compiler.api.symbols.FunctionSymbol;
 import io.ballerina.compiler.api.symbols.MethodSymbol;
 import io.ballerina.compiler.api.symbols.ModuleSymbol;
@@ -48,7 +46,6 @@ import io.ballerina.projects.Module;
 import io.ballerina.projects.Package;
 import io.ballerina.projects.Project;
 import io.ballerina.projects.ProjectKind;
-import io.ballerina.tools.text.LinePosition;
 import org.ballerinalang.langserver.LSPackageLoader;
 import org.ballerinalang.langserver.common.utils.CommonKeys;
 import org.ballerinalang.langserver.common.utils.CommonUtil;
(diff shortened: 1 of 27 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -484,7 +487,10 @@ protected LSCompletionItem getImplicitNewCItemForClass(ClassSymbol symbol, Balle
     }
 
     protected List<LSCompletionItem> expressionCompletions(BallerinaCompletionContext context) {
-        List<Symbol> visibleSymbols = context.visibleSymbols(context.getCursorPosition());
```
IMS94, line 487: We should be updating the `DocumentServiceContext.visibleSymbols()` method instead of using the `semanticModel` directly. In `DocumentServiceContext`, we should overload, `visibleSymbols()` method to take in `DiagnosticState` (vararg) as a second parameter and update the existing `visibleSymbols()` method to pass in `VALID` as the diagnostic state.

IMS94 (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
malinthar: Should we add the same logic to WorkSpaceServiceContext as well? I'm not sure about it. @IMS94 WDYT?

IMS94: Yes. +1. It as only one usage (in CreateFunctionExecutor) right now. But, we should do the same there as well.

nadeeshaan (approved):
LGTM

malinthar (dismissed):
LGTM!

nadeeshaan (dismissed):
LGTM

IMS94 (approved):

PR #32571, opened 2021-09-03
Handle readBuildJson exceptions in PackageResolution to disable sticky flag

### LINKED ISSUE #32543: [ProjectAPI] Handle read exceptions for Target Build File
**Description:**
Handle the exception in ProjectUtils and disable stickys as necessary

**Describe your task(s)**

Ibaqu:
## Purpose
Handles the exceptions thrown by readBuildJson to disable sticky flag at PackageResolution

Fixes #32543

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/test/java/io/ballerina/projects/ProjectUtilsTests.java
```diff
@@ -52,23 +54,35 @@ public void setUp() throws IOException {
...
+            Assert.assertEquals(buildJson.lastBuildTime(), 1629359520);
+            Assert.assertEquals(buildJson.lastUpdateTime(), 1629259520);
+        } catch (Exception e) {
+            Assert.fail("Reading Build Json failed");
+        }
     }
 
     @Test(expectedExceptions = ProjectException.class,
-            expectedExceptionsMessageRegExp = "Failed to read the 'build' file")
+            expectedExceptionsMessageRegExp = "Failed to read the 'build' file", enabled = false)
```
azinneera, line 67: Why do we need to disable some tests?

Ibaqu: These tests actually call the `readBuildJson` function directly and assert for the exception message being thrown. With the change, the specific exception messages are not thrown so these tests fail. 

Should I remove them entirely or modify them in a certain way?

azinneera: We can assert for the original IO exceptions here, isnt it?

Ibaqu: Changed tests to assert for specific exceptions

compiler/ballerina-lang/src/test/java/io/ballerina/projects/ProjectUtilsTests.java
```diff
@@ -52,23 +53,29 @@ public void setUp() throws IOException {
...
+        }
     }
 
-    @Test(expectedExceptions = ProjectException.class,
-            expectedExceptionsMessageRegExp = "Failed to read the 'build' file")
+    @Test()
     public void testReadBuildJsonForNonExistingBuildFile() {
         Path buildFilePath = PROJECT_UTILS_RESOURCES.resolve("xyz").resolve(ProjectConstants.BUILD_FILE);
-        ProjectUtils.readBuildJson(buildFilePath);
+        Assert.assertThrows(IOException.class, () -> {
```
pramodya1994, line 68: Can we assert the exception message here?

compiler/ballerina-lang/src/test/java/io/ballerina/projects/ProjectUtilsTests.java
```diff
@@ -52,23 +53,29 @@ public void setUp() throws IOException {
...
+        });
     }
 
-    @Test(expectedExceptions = ProjectException.class,
-            expectedExceptionsMessageRegExp = "Invalid 'build' file format")
+    @Test()
     public void testReadBuildJsonForInvalidBuildFile() {
         Path buildFilePath = PROJECT_UTILS_RESOURCES.resolve("invalid-build");
-        ProjectUtils.readBuildJson(buildFilePath);
+        Assert.assertThrows(JsonSyntaxException.class, () -> {
```
pramodya1994, line 76: Here also:
Can we assert the exception message here?

compiler/ballerina-lang/src/test/java/io/ballerina/projects/ProjectUtilsTests.java
Ibaqu: After changing the implementation, the function does not throw any particular exceptions with set messages. ie. the exception message can vary, so we decided to assert the exception type directly

azinneera (commented):
Shall we add an integration test to verify the behavior when the `build` file is invalid?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/BuildProject.java
```diff
@@ -207,18 +208,30 @@ public void save() {
...
+                        // update build json file
+                        writeBuildFile(buildFilePath);
+                    } else {
+                        // only update build time
+                        buildJson.setLastBuildTime(System.currentTimeMillis());
+                        ProjectUtils.writeBuildFile(buildFilePath, buildJson);
+                    }
+                } else {
+                    Files.deleteIfExists(buildFilePath);
+                    createBuildFile(buildFilePath);
```
Ibaqu, line 228: @azinneera This particular line fails on windows with an IO Exception as specified in the build scan. 

```
at io.ballerina.projects.directory.BuildProject.createBuildFile(BuildProject.java:404)
at io.ballerina.projects.directory.BuildProject.save(BuildProject.java:228)
at io.ballerina.projects.test.PackageResolutionTests.testProjectSaveWithEmptyBuildFile(PackageResolutionTests.java:166)
```

So if an empty build file exists, we delete it, create a new one and rewrite it. Is this an acceptable way to proceed with such a case?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/BuildProject.java
```diff
@@ -207,18 +208,28 @@ public void save() {
...
+                        writeBuildFile(buildFilePath);
+                    } else {
+                        // only update build time
+                        buildJson.setLastBuildTime(System.currentTimeMillis());
+                        ProjectUtils.writeBuildFile(buildFilePath, buildJson);
+                    }
+                } else {
+                    writeBuildFile(buildFilePath);
+                }
+            } catch (JsonSyntaxException e) {
```
sameerajayasoma, line 229: Shouldn't we ignore or report a warning instead of an exception here?

sameerajayasoma (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/BuildProject.java
azinneera: Since this is not important to the user, I think we can ignore and rewrite the file. Also, we should be able to add a no-location diagnostics to report a warning which we don't have atm.

hevayo: +1 to ignore since this file should be transparent to user

Ibaqu: With 864788ea19e77a6292976f2b999a8c933034371d we ignore the JSONSyntaxException and simply rewrite the build file

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/BuildProject.java
```diff
@@ -207,18 +208,28 @@ public void save() {
...
                 writeBuildFile(buildFilePath);
-            } else {
-                // only update build time
-                buildJson.setLastBuildTime(System.currentTimeMillis());
-                ProjectUtils.writeBuildFile(buildFilePath, buildJson);
+            } catch (IOException e) {
+                throw new ProjectException("Failed to read the '" + BUILD_FILE + "' file");
```
azinneera, line 232: This exception shouldn't be thrown either.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/BuildProject.java
```diff
@@ -207,18 +208,26 @@ public void save() {
...
+                    } else {
+                        // only update build time
+                        buildJson.setLastBuildTime(System.currentTimeMillis());
+                        ProjectUtils.writeBuildFile(buildFilePath, buildJson);
+                    }
+                } else {
+                    writeBuildFile(buildFilePath);
+                }
+            } catch (JsonSyntaxException | IOException e) {
                 writeBuildFile(buildFilePath);
```
azinneera, line 230: Looks like this can end up on a ProjectException from the ProjectUtils.writeBuildFile method. Shall we modify the util to throw the IOException, catch and ignore here?

azinneera (approved):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/BuildProject.java
```diff
@@ -201,24 +203,30 @@ public void save() {
...
         } else {
-            // build file exists
-            BuildJson buildJson = readBuildJson(buildFilePath);
+            BuildJson buildJson = null;
+            try {
+                buildJson = readBuildJson(buildFilePath);
+            } catch (JsonSyntaxException e) {
+                // ignore
+            } catch (IOException e) {
+                return;
```
azinneera, line 218: Returning here will stop writing the Dependencies.toml. We should not do that. We should ignore both exceptions and continue.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -161,11 +163,16 @@ private boolean getSticky(PackageContext rootPackageContext) {
         if (rootPackageContext.project().kind() == ProjectKind.BUILD_PROJECT) {
             Path buildFilePath = this.rootPackageContext.project().sourceRoot().resolve(TARGET_DIR_NAME)
                     .resolve(BUILD_FILE);
-            if (buildFilePath.toFile().exists()) {
-                BuildJson buildJson = readBuildJson(buildFilePath);
-                if (!buildJson.isExpiredLastUpdateTime()) {
-                    this.autoUpdate = false;
-                    return true;
+            if (buildFilePath.toFile().exists() && buildFilePath.toFile().length() > 0) {
```
hevayo, line 166: We could use `Files.exists(path)` it will check if the file has read access too. Also can we check the behavior where build file without permission

Ibaqu: Changed to `Files.exists()`

I added a test for NoReadPermissions as well and it creates a new file if the file has no read permissions.

azinneera (approved):

Issue #32470, opened 2021-08-31
Remove versions of langlibs in compiler


azinneera:
**Description:**
As per the current implementation, the lang repo has the lang library versions hardcoded https://github.com/ballerina-platform/ballerina-lang/blob/79e484fee14324cfa2f73324170ae621981fa853/compiler/ballerina-lang/src/main/java/org/ballerinalang/model/elements/PackageID.java#L66-L110

Since the latest language spec is updated and that we now do not have versions to lang libraries (ref: https://github.com/ballerina-platform/ballerina-spec/commit/eff622461edb56250d597efbcba136aea3793f50) we should remove the hardcoded versions.

azinneera:
This is a prerequisite for #32420

hasithaa:
Can we create a package without a version? IIRC, it will default to 0.1.0.

azinneera:
@hasithaa As per the offline discussion with @sameerajayasoma we decided to set the version  of langlibs to `0.0.0`

PR #32588, opened 2021-09-06
Remove langlib versions

azinneera:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/32470

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/ResolutionEngine.java
```diff
@@ -195,13 +196,38 @@ private void mergeGraph(PackageDescriptor rootNode,
                             PackageDependencyGraphBuilder graphBuilder) {
         Collection<PackageDescriptor> directDependencies = dependencyGraph.getDirectDependencies(rootNode);
         for (PackageDescriptor directDep : directDependencies) {
+            DependencyGraph<PackageDescriptor> dependencyGraphFinal;
+            if (directDep.isBuiltInPackage()) {
```
hevayo, line 200: Do we need to change the resolution engine can we enforce this from package resolver

azinneera: AFAIU, if we are to handle this in the package resolver, we would need to create the dependency graph there. Since the dependency graph is constructed in ResolutionEngine, we have to do it here.

sameerajayasoma: This solution is good as we discussed offline.

sameerajayasoma (approved):

PR #32598, opened 2021-09-07
Fix `symbol()` API for Computed Keys in Mapping Constructors

### LINKED ISSUE #31803: `symbol()` not working for computed keys in mapping constructors
Consider the following where `foo` is a function. 
```ballerina
function test(string name, int age) {
    Address adrs = {city: "Colombo 3", country: "Sri Lanka"};

    Person p = {
        name: name,
        age,
        [<cursor>foo()]: "FOO",
        ...adrs
    };
}
```
If I use `symbol()` at the illustrated position, it returns empty.

sanjana:
Fixes #31803

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
@@ -1532,7 +1532,7 @@ class SymbolFinder extends BaseVisitor {
 
     @Override
     public void visit(BLangRecordLiteral.BLangRecordKey recordKey) {
-        if (setEnclosingNode(recordKey.fieldSymbol, recordKey.pos)) {
+        if (recordKey.fieldSymbol != null && setEnclosingNode(recordKey.fieldSymbol, recordKey.pos)) {
             return;
         }
 
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
@@ -1532,7 +1532,7 @@ class SymbolFinder extends BaseVisitor {
 
     @Override
     public void visit(BLangRecordLiteral.BLangRecordKey recordKey) {
-        if (recordKey.fieldSymbol != null && setEnclosingNode(recordKey.fieldSymbol, recordKey.pos)) {
+        if (!recordKey.computedKey && setEnclosingNode(recordKey.fieldSymbol, recordKey.pos)) {
             return;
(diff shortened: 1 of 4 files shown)


mohanvive (approved):
LGTM

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
```diff
@@ -1532,7 +1532,7 @@ public void visit(BLangRecordLiteral.BLangRecordKeyValueField recordKeyValue) {
 
     @Override
     public void visit(BLangRecordLiteral.BLangRecordKey recordKey) {
-        if (setEnclosingNode(recordKey.fieldSymbol, recordKey.pos)) {
+        if (recordKey.fieldSymbol != null && setEnclosingNode(recordKey.fieldSymbol, recordKey.pos)) {
```
pubudu91, line 1535: Why is this null check needed?

mohanvive: Yes, I also wanted to check from Anjana. Do we need this change with other fix you have done here. @sanjana

sanjana: @pubudu91 @mohanvive 

The null check here is essential. 

If the field has a computed key, the `fieldSymbol` is set to `null`.

https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java#L6556-L6570

Therefore, this null check is required to bypass this `if` condition.

sanjana: Another possible approach would be as following: 

`if (!recordKey.computedKey && setEnclosingNode(recordKey.fieldSymbol, recordKey.pos))`

pubudu91: Ah I see. I think the second approach you suggested would be better since it's more clear as to what we are checking there.

sanjana: Updated.

mohanvive (approved):

PR #32667, opened 2021-09-09
Fix visibility of let expression inner variables

### LINKED ISSUE #28883: [Debugger] `let` expression inner variables are visible outside the expression

### LINKED ISSUE #32477: [Debugger] Variables defined inside the conditional blocks and loops are also visible outside the blocks
**Description:**

The variables defined inside the `if` statements and, `while` and `foreach` loops are visible at the outside of the blocks and loops. Ideally they should not visible outside. 

Consider the following code example:
```
public function main() {
    int i = 1;
    if (i == 1) {
        int j = i;
    }
}
```
The variable `j` is visible outside the `if` block also.
 
ifblock

**Affected Versions:**

https://github.com/ballerina-platform/ballerina-lang/issues/28883

gabilang:
Fixes #28883
Fixes #32477

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
@@ -1579,7 +1579,7 @@ public class BIRGen extends BLangNodeVisitor {
                             astIfStmt.elseStmt.pos.lineRange().endLine().line(),
                             astIfStmt.elseStmt.pos.lineRange().endLine().line(),
                             astIfStmt.elseStmt.pos.lineRange().endLine().offset(),
-                            astIfStmt.elseStmt.pos.lineRange().endLine().offset());
+                            astIfStmt.elseStmt.pos.lineRange().endLine().offset(), 0, 0);
                     this.env.enclBB.terminator = new BIRTerminator.GOTO(newLocation, nextBB, this.currentScope);
                 } else {
                     this.env.enclBB.terminator = new BIRTerminator.GOTO(null, nextBB, this.currentScope);
@@ -1635,7 +1635,18 @@ public class BIRGen extends BLangNodeVisitor {
         astWhileStmt.body.accept(this);
         this.env.unlockVars.pop();
         if (this.env.enclBB.terminator == null) {
-            this.env.enclBB.terminator = new BIRTerminator.GOTO(null, whileExprBB, this.currentScope);
+            if (astWhileStmt.body.pos != null) {
+                Location newLocation = new BLangDiagnosticLocation(
+                        astWhileStmt.body.pos.lineRange().filePath(),
+                        astWhileStmt.body.pos.lineRange().endLine().line(),
+                        astWhileStmt.body.pos.lineRange().endLine().line(),
+                        astWhileStmt.body.pos.lineRange().endLine().offset(),
+                        astWhileStmt.body.pos.lineRange().endLine().offset(), 0, 0
+                );
+                this.env.enclBB.terminator = new BIRTerminator.GOTO(newLocation, whileExprBB, this.currentScope);
+            } else {
+                this.env.enclBB.terminator = new BIRTerminator.GOTO(null, whileExprBB, this.currentScope);
+            }
         }
 
         this.env.enclBasicBlocks.add(whileEndBB);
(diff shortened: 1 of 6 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1635,7 +1645,18 @@ public void visit(BLangWhile astWhileStmt) {
         astWhileStmt.body.accept(this);
         this.env.unlockVars.pop();
         if (this.env.enclBB.terminator == null) {
-            this.env.enclBB.terminator = new BIRTerminator.GOTO(null, whileExprBB, this.currentScope);
+            if (astWhileStmt.body.pos != null) {
```
warunalakshitha, line 1648: Can we get this null?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmObservabilityGen.java
```diff
@@ -616,6 +618,7 @@ private void rewriteObservableFunctionInvocations(BIRFunction func, BIRPackage p
                         BIRBasicBlock observeEndBB = insertBasicBlock(func, eeTargetIndex + 1);
                         BIRBasicBlock newTargetBB = insertBasicBlock(func, eeTargetIndex + 2);
                         swapBasicBlockContent(errorEntry.targetBB, newTargetBB);
+                        resetLocalVarEndBBs(func, errorEntry.targetBB, newTargetBB);
```
nadundesilva, line 621: Does it make more sense to move this into the `swapBasicBlockContent` method itself ?
(Also there is another `swapBasicBlockTerminator` method as well where we only swap the terminator.)

nadundesilva: I think when we have an observability terminator (eg:- HTTP Client call), the the immediate local vars might not be visible when not swapped. But it's not the exact bug we are trying to fix I think.

gabilang: Yes, adding this to when we do the terminator swap only is not be relevant to the fix for this particular issue. IMO reassigning endBBs of local variables after the `Return` or `Panic` terminator swap may not be useful. As you mentioned, when the terminator instruction is `Call` and it is observable, the endBBs may need to be reassigned after the terminator swap. I'll make changes accordingly.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
gabilang: We can remove the null check since the position never takes null here. Addressed with https://github.com/ballerina-platform/ballerina-lang/pull/32667/commits/d6c32a58981fff76246a855c20aef34596fead4f

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmObservabilityGen.java
gabilang: Addressed with https://github.com/ballerina-platform/ballerina-lang/pull/32667/commits/3d12d91fd7257fcb7f4417157a159d0be72b84cd

nadundesilva: Yes for return and panic it's probably not required 👍🏻

nadundesilva (approved):
Observability related changes LGTM

warunalakshitha (approved):
LGTM

PR #32672, opened 2021-09-09
Fix `typeof` expression for immutable values

### LINKED ISSUE #32531: `typeof` expression returns wrong value for immutable container values
**Description:**
As per the spec, 
```
When v is a reference value,
- For containers, this is the same as the inherent type
- when the container is immutable, it will be a singleton type.  
```

The current implementation deviates from this specifications. 

**Steps to reproduce:**
```ballerina
int[] & readonly arr = [1, 2, 3];
io:println(typeof arr); // typedesc int[] & readonly -  should be typedesc [1,2,3]
```

**Affected Versions:**
beta3

**Remarks:**
related issues - #15278 , #13189

HindujaB:
## Purpose
This is to complete the fix done on `typeof` implementation for immutable values. 

Fixes #32531

## Approach
The `typeof` will provide a Finite type that contains the immutable value. 
This is to comply with the spec,
```
When v is a reference value,
- For containers, this is the same as the inherent type
- when the container is immutable, it will be a singleton type.  
```
## Samples
```ballerina
int[] & readonly arr = [1, 2, 3];
io:println(typeof arr); // typedesc int[] & readonly -  should be typedesc [1,2,3]
```
## Remarks
This implementation was earlier reverted due to some issues found  in `langlib:value`.  https://github.com/ballerina-platform/ballerina-lang/pull/32530
This PR fixes them and redo the earlier change.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
@@ -2152,7 +2152,8 @@ public class TypeChecker {
             case TypeTags.ANYDATA_TAG:
                 return checkIsLikeAnydataType(sourceValue, sourceType, unresolvedValues, allowNumericConversion);
             case TypeTags.FINITE_TYPE_TAG:
-                return checkFiniteTypeAssignable(sourceValue, sourceType, (BFiniteType) targetType);
+                return checkFiniteTypeAssignable(sourceValue, sourceType, (BFiniteType) targetType,
+                 unresolvedValues, allowNumericConversion);
             case TypeTags.XML_ELEMENT_TAG:
                 if (sourceTypeTag == TypeTags.XML_TAG) {
                     XmlValue xmlSource = (XmlValue) sourceValue;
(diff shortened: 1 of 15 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -2609,7 +2610,16 @@ private static boolean checkIsLikeTableType(Object sourceValue, BTableType targe
         return true;
     }
 
-    private static boolean checkFiniteTypeAssignable(Object sourceValue, Type sourceType, BFiniteType targetType) {
+    private static boolean checkFiniteTypeAssignable(Object sourceValue, Type sourceType, BFiniteType targetType,
+                                                     List<TypeValuePair> unresolvedValues,
+                                                     boolean allowNumericConversion) {
+        Type firstValueType = getType(targetType.valueSpace.iterator().next());
```
warunalakshitha, line 2616: Should we do this only targetType.valueSpace.size() == 1

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -308,7 +310,13 @@ static boolean isConvertibleToFloatingPointTypes(Object value) {
                 convertibleTypes.addAll(getConvertibleTypes(inputValue, effectiveType, unresolvedValues));
                 break;
             case TypeTags.FINITE_TYPE_TAG:
-                for (Object valueSpaceItem : ((BFiniteType) targetType).valueSpace) {
+                BFiniteType finiteType = (BFiniteType) targetType;
+                Type firstValueType = getType(finiteType.valueSpace.iterator().next());
```
warunalakshitha, line 314: IMO first we can check targetType.valueSpace.size() == 1

warunalakshitha (approved):

Issue #32648, opened 2021-09-09
In-Built compiler plugins should have an order when executing


ayeshLK:
**Description:**
In-Built compiler plugins should have an order when executing.

**Problem:**
Currently there are several in-built compiler plugins available in ballerina distribution. And we have added a new plugin for `open-api-doc-generation` which will generate open-api doc for http-services. Since this doc-generation should work with both local/docker builds we are updating both the executable-jar and the thin-jar with generated resources at compile time.
Updating the `thin-jar` is required for ballerina docker build since it will use the `thin-jar` when generating the docker artifacts. 

Ballerina docker build happens in `c2c-plugin` which is again a built-in plugin. And since we have not defined any execution order for built-in compiler plugins `c2c-plugin` executes before `open-api-plugin`. So, currently our solution does not work properly with docker builds.

PR #32674, opened 2021-09-09
 Add a hack to run c2c codeGenCompleted task at last

azinneera:
Hack for #32648

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


sameerajayasoma (changes requested):
Can we simply run the C2C plugin as the last one? This will guarantees that the C2C plugin runs as the last one? 

Otherwise we will have to introduce more hacks

azinneera:
> Can we simply run the C2C plugin as the last one? This will guarantees that the C2C plugin runs as the last one?
> 
> Otherwise we will have to introduce more hacks

Then we might need to iterate the keyset and find the c2c by the classname since at this point what we have is a map of <LifeCycleListenerInfo, List<CodeGenerationCompletedTask>>

Is this ok to do?

sameerajayasoma (approved):

PR #32712, opened 2021-09-13
Fix escaping chars in XML texts twice

### LINKED ISSUE #32673: Printing XML CDATA section gives incorrect output
**Description:**
Consider the bellow sample of XML CDATA added via #32564
```bal
public function main() {
    xml x7 = xml `<![CDATA[ abc --> <!-- --> some more text ]]>`;
    io:println(x7);
}
```
This prints `abc --&amp;gt; &amp;lt;!-- --&amp;gt; some more text` where it should print `abc --&gt; &lt;!-- --&gt; `some more text

**Affected Versions:**
Current master

HindujaB:
## Purpose
In the implementation for xml `stringValue()` we use `XMLStreamWriter` to convert the xml texts from byte array stream to string. 
The stream writer instance is created from different classes when we run unit tests and when we externally run using `bal run`. 

Fixes #32673 

## Approach
When the unit tests are running,  the stream writer uses the default configurations where it escapes the characters in  the string output which is already escaped by the native implementation.  
https://github.com/ballerina-platform/ballerina-lang/blob/243aa0746732976f520802677cba1f257f096f0f/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/BallerinaXmlSerializer.java#L158

This PR unsets the `escapeCharacters` property for default stream writers to fix this error.

## Samples
Need to be tested within unit tests.
```ballerina
public function main() {
    xml x7 = xml `<![CDATA[ abc --> <!-- --> some more text ]]>`;
    assertTrue(x7.toString() == " abc --&gt; &lt;!-- --&gt; some more text ");  
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/BallerinaXmlSerializer.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/BallerinaXmlSerializer.java
@@ -45,6 +45,8 @@ import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
 import javax.xml.stream.XMLStreamWriter;
 
+import static io.ballerina.runtime.api.values.BXmlItem.XMLNS_NS_URI_PREFIX;
+
 /**
  * XML Serializer for Ballerina XML value trees.
  *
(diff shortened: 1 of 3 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/BallerinaXmlSerializer.java
```diff
@@ -141,8 +147,9 @@ private void writeXMLComment(XmlComment xmlValue) throws XMLStreamException {
     }
 
     private void writeXMLText(XmlText xmlValue) throws XMLStreamException {
-        // No need to escape xml text when they are within xml element, it's handle from xml stream writer.
-        if (this.withinElement) {
+        // No need to escape xml text when they are within xml element or if it is not from default stream writer.
+        // It's handled by xml stream  writer.
+        if (this.withinElement && !isDefaultFactory) {
```
warunalakshitha, line 152: Can we use xmlOutputFactory.getProperty("escapeCharacters") != null. and equal to true here

HindujaB: If we use the `xmlOutputFactory.getProperty("escapeCharacters")` in this check, it  gives an `IllegalStateException` for the non-default stream writer (Since it is not a predefined property for `WstxOutputFactory`).

warunalakshitha (approved):

PR #32734, opened 2021-09-14
Fix allowing concurrent calls from non-isolated resource functions

### LINKED ISSUE #32652: Execution of non isolated methods with async/yield code
**Description:**
Hi,
Please refer to the following example.

```ballerina
import ballerina/log;
import ballerina/http;

int a = 5;

service / on new http:Listener(8090) {
    resource function post .(@http:Payload string textMsg) returns string {
        log:printInfo(textMsg);
        int _ = a;
        runtime:sleep(10);
        return textMsg;
    }
}
```

Output for 3 requests( Hello1, Hello2, Hello3)

```
time = 16:41:59 message = "Hello1"
time = 16:42:01 message = "Hello2"
time = 16:42:04 message = "Hello3" **// Executed to one by one but each does not wait 10s one after other.(Refer to the time)**
```

The given resource method is non-isolated. Because of that, requests should be executed sequentially. 
Here we have sleep, so once request execution comes to that line, the strand will wait for sleep to end. 
Since there is no active execution related to the same service, the next request will get executed and yield again. 

So there is no concurrent requests execution but still, there can be requests on the waiting stage for the same service. That's the current behavior of the scheduler. 

What would be the expected behavior?
Case 1) Should the scheduler wait until the first request is made to execute the next request? 
Case 2) Should subsequent requests execute sequentially even if there are requests for the same service in yield state?

HindujaB:
## Purpose
Currently, the strands used to execute the resource functions are mapped according to their objects in the scheduler. 
This allows concurrent execution of resource functions if they are non-isolated. This can cause errors if multiple functions access same entity. 

Fixes #32652

## Approach
Fixed by not mapping the strand according to service object;

## Remarks
TODO: add a unit test to test this scenario.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/scheduling/Scheduler.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/scheduling/Scheduler.java
@@ -32,17 +32,16 @@ import io.ballerina.runtime.internal.values.ChannelDetails;
 import io.ballerina.runtime.internal.values.FutureValue;
 
 import java.io.PrintStream;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.Stack;
-import java.util.WeakHashMap;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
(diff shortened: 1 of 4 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/scheduling/Scheduler.java
```diff
@@ -169,22 +168,21 @@ public FutureValue scheduleLocal(Object[] params, BFunctionPointer<?, ?> fp, Str
...
-            if (groupInMap == null) {
-                return new ItemGroup(item);
-            } else {
-                groupInMap.add(item);
-                return groupInMap;
-            }
-        });
+        ItemGroup group = objectGroup.get();
+        if (group == null) {
+            group = new ItemGroup(item);
```
HindujaB, line 181: The `objectGroup` need to initialised here as we can't create an `ItemGroup` without a `ScedulerItem`

warunalakshitha (approved):

PR #32737, opened 2021-09-14
Avoid central calls when resolved from file system

### LINKED ISSUE #32675: Avoid querying the central when exact match is needed and the version is available in filesystem
**Description:**

When the versions are available in Dependencies.toml and a non-expired `target/build` is present,  internally it should be considered as `sticky == true, and locking mode == HARD` which means we should match the exact version. If that version is available in the filesystem repos, we should not query the central.

**Steps to Reproduce:**

1. Create a package with a few dependencies and build with `--dump-build-time` flag and note the package resolution time
      This will create the Dependencies.toml and the `target/build` file
2. Re-run the build command and compare the package resolution times

*_Expected Behavior:_*
The package resolution time for the 2nd attempt should be much lesser than the first time since it is expected to resolve from the filesystem and so no central querying should take place.

*_Actual Behavior:_*
Both times are almost equal

**Affected Versions:**
Beta.3-SNAPSHOT

Dilhasha:
## Purpose
Avoid central calls when resolved from file system.

Fixes #32675

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/repositories/RemotePackageRepository.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/repositories/RemotePackageRepository.java
@@ -243,23 +243,24 @@ public class RemotePackageRepository implements PackageRepository {
         if (options.offline()) {
             return cachedPackages;
         }
+        List<ResolutionRequest> updatedRequests = new ArrayList<>(requests);
         // Remove the already resolved requests when the locking mode is hard
         for (PackageMetadataResponse response : cachedPackages) {
             if (response.packageLoadRequest().packageLockingMode().equals(PackageLockingMode.HARD)
                     && response.resolutionStatus().equals(ResolutionResponse.ResolutionStatus.RESOLVED)) {
-                requests.remove(response.packageLoadRequest());
+                updatedRequests.remove(response.packageLoadRequest());
             }
         }
         // Resolve the requests from remote repository if there are unresolved requests
-        if (!requests.isEmpty()) {
+        if (!updatedRequests.isEmpty()) {
             try {
-                PackageResolutionRequest packageResolutionRequest = toPackageResolutionRequest(requests);
+                PackageResolutionRequest packageResolutionRequest = toPackageResolutionRequest(updatedRequests);
                 PackageResolutionResponse packageResolutionResponse = client.resolveDependencies(
                         packageResolutionRequest, JvmTarget.JAVA_11.code(),
                         RepoUtils.getBallerinaVersion(), true);
 
                 Collection<PackageMetadataResponse> remotePackages =
-                        fromPackageResolutionResponse(requests, packageResolutionResponse);
+                        fromPackageResolutionResponse(updatedRequests, packageResolutionResponse);
                 // Merge central requests and local requests
                 // Here we will pick the latest package from remote or local
                 return mergeResolution(remotePackages, cachedPackages);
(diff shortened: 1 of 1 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/repositories/RemotePackageRepository.java
```diff
@@ -242,27 +243,35 @@ private PackageNameResolutionRequest toPackageNameResolutionRequest(List<ImportM
...
-            throw new ProjectException(e.getMessage());
+        // Remove the already resolved requests when the locking mode is hard
+        for (PackageMetadataResponse response : cachedPackages) {
+            if (response.packageLoadRequest().packageLockingMode().equals(PackageLockingMode.HARD)
+                    && response.resolutionStatus().equals(ResolutionResponse.ResolutionStatus.RESOLVED)) {
+                requests.remove(response.packageLoadRequest());
+            }
+        }
+        // Resolve the requests from remote repository if there are unresolved requests
+        if (requests.size() > 0) {
```
lnash94, line 254:
```suggestion
        if (requests.isEmpty()) {
```

Dilhasha: @lnash94 it should be `if (!requests.isEmpty()) {`

Dilhasha: Fixed with 9113a96f8a8d84844aae0b5b4cb4e4aa61df0f50

azinneera (approved):

sameerajayasoma (commented):
Is there a way to test this change?

azinneera:
> Is there a way to test this change?

We can add a test case to distribution

compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/repositories/RemotePackageRepository.java
```diff
@@ -242,27 +243,35 @@ private PackageNameResolutionRequest toPackageNameResolutionRequest(List<ImportM
...
-        } catch (ConnectionErrorException e) {
-            // ignore connect to remote repo failure
-            // TODO we need to add diagnostics for resolution errors
-        } catch (CentralClientException e) {
-            throw new ProjectException(e.getMessage());
+        // Remove the already resolved requests when the locking mode is hard
+        for (PackageMetadataResponse response : cachedPackages) {
+            if (response.packageLoadRequest().packageLockingMode().equals(PackageLockingMode.HARD)
+                    && response.resolutionStatus().equals(ResolutionResponse.ResolutionStatus.RESOLVED)) {
+                requests.remove(response.packageLoadRequest());
```
sameerajayasoma, line 250: A better approach would be to create a new list here. Modifying the method parameters should be avoided as much as possible.

sameerajayasoma (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/repositories/RemotePackageRepository.java
Dilhasha: fixed with 68e2b272179cba27dae47408c84cb84d3e85adbc

sameerajayasoma (approved):

PR #32796, opened 2021-09-17
Fix completions within method call expr args

### LINKED ISSUE #19612: Invalid suggestions for lang lib function calls
Refer to the screen grab below:
Peek 2019-10-28 12-07

As it can be seen above, when you press `Ctrl+space` after typing the function, it shows the receiver in the suggestions, instead of showing the actual parameter (which is a function pointer). I'm guessing the plugin is simply picking the first param in the param list. In this particular case, the actual parameter the user is concerned about is the second param in the param list.

### LINKED ISSUE #21540: Add tooling support for XML navigation expressions
**Description:**
$title.

**Steps to reproduce:**

**Affected Versions:**

Original issue https://github.com/ballerina-platform/ballerina-lang/issues/20667

### LINKED ISSUE #32825: Introduce a way to disable cancel checker while debugging
**Description:**
With the introduction of cancellation support (#32419), Requests get cancelled while doing remote debugging with VSCode. This will be problematic if we are to debug a released/RC ballerina pack.

**Describe your problem(s)**
See description

**Describe your solution(s)**
Need to implement a mechanism to disable cancellation support via a system property or an environment variable. (An environment variable will be better.)

#32311

nadeeshaan:
## Purpose
Fixes #19612
Fixes #32825
Fixes #21540

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
@@ -143,7 +143,7 @@ class BallerinaTextDocumentService implements TextDocumentService {
                     position.getPosition());
             try {
                 return LangExtensionDelegator.instance()
-                        .completion(position, context, this.serverContext, cancelChecker);
+                        .completion(position, context, this.serverContext, null);
             } catch (CancellationException ignore) {
                 // Ignore the cancellation exception
             } catch (Throwable e) {
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
@@ -143,7 +143,7 @@ class BallerinaTextDocumentService implements TextDocumentService {
                     position.getPosition());
             try {
                 return LangExtensionDelegator.instance()
-                        .completion(position, context, this.serverContext, null);
+                        .completion(position, context, this.serverContext, cancelChecker);
             } catch (CancellationException ignore) {
                 // Ignore the cancellation exception
             } catch (Throwable e) {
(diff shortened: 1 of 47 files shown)


malinthar (approved):
Changes Looks Good! Shall we ensure that we have addressed the Remote Method Call Action Node as well.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
```diff
@@ -143,7 +143,7 @@ void setClientCapabilities(LSClientCapabilities clientCapabilities) {
                     position.getPosition());
             try {
                 return LangExtensionDelegator.instance()
-                        .completion(position, context, this.serverContext, cancelChecker);
+                        .completion(position, context, this.serverContext, null);
```
mohanvive, line 146: This change is not required, right?

mohanvive: I think we should find some other option to handle this case when debugging. There is a high possibility this temp change get committed. Worst case, a test case maybe :)

mohanvive (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/BallerinaTextDocumentService.java
nadeeshaan: @mohanvive as @IMS94 suggested, let's add a system property. I'm working on the fix and we can merge it along with this PR.

IMS94: Created #32825 to address that. May be we can close it with this PR.

IMS94 (dismissed):
LGTM!

mohanvive (approved):

PR #32812, opened 2021-09-18
Add module level components to package service

prabushi:
## Purpose
- Add components support for record, objects, types, constants, enums, classes, listeners, and module levelVariables.
- Update the package components API implementation to use the syntax tree.

Related issue https://github.com/wso2/ballerina-plugin-vscode/issues/118

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/packages/DocumentComponentsVisitor.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/packages/DocumentComponentsVisitor.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright (c) 2021, WSO2 Inc. (http://wso2.com) All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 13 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/extensions/ballerina/packages/DocumentComponentsVisitor.java
```diff
@@ -0,0 +1,152 @@
...
+    private boolean isInsideClass;
+    private boolean isInsideService;
+
+    DocumentComponentsVisitor(JsonObject module) {
+        this.module = module;
+    }
+
+    public void getDocumentComponents(Node node) {
+        visitSyntaxNode(node);
+    }
```
nadeeshaan, line 52: Here we are maintaining a state in between the construct visits which is not a good approach with a visitor like this. Specially dynamically setting variables like `isInsideClass` and `isInsideService` might not be a good approach since we might end up with filling invalid information.

Instead of the visitor, let's use a node transformer.

nadeeshaan (changes requested):

PR #32867, opened 2021-09-23
Filter Field Symbols from Visible Symbols in the Context of Fields

### LINKED ISSUE #29382: Inconsistent symbol in scope behaviour among different kinds of fields
```ballerina
public class Foo {
    bar:Unknown c = new();
    f<cursor1>
}

type Person record {|
    bar:Name name;
    f<cursor2>
|};

type PersonObj object {
    bar:Name name;
    f<cursor3>
};
```

If we check the symbols visible at each of the above illustrated points, both first and second cases return 4 symbols (including the `c` and `name` fields). But in cursor3 position, only 3 symbols are returned (just the type defs). What should be the correct behaviour in these 3 cases?

sanjana:
Fixes #29382

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
@@ -474,12 +474,23 @@ public class BallerinaSemanticModel implements SemanticModel {
             } else {
                 compiledSymbol = symbolFactory.getBCompiledSymbol(symbol, symbol.getOriginalName().getValue());
             }
-            if (compiledSymbol == null || compiledSymbols.contains(compiledSymbol)) {
+            if (compiledSymbol == null || compiledSymbols.contains(compiledSymbol) || isFieldSymbol(compiledSymbol)) {
                 return;
             }
             compiledSymbols.add(compiledSymbol);
         }
         addToCompiledSymbols(compiledSymbols, scopeEntry.next, cursorPos, name, states);
+    }
+
+    private boolean isFieldSymbol(Symbol compiledSymbol) {
+        switch (compiledSymbol.kind()) {
+            case CLASS_FIELD:
+            case OBJECT_FIELD:
+            case RECORD_FIELD:
+                return true;
+            default:
+                return false;
+        }
     }
 
     private boolean isServiceDeclSymbol(BSymbol symbol) {
--- a/tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolLookupTest.java
+++ b/tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolLookupTest.java
@@ -493,6 +493,37 @@ public class SymbolLookupTest {
         }
     }
 
+    @Test(dataProvider = "FieldSymbolPosProvider")
+    public void testSymbolLookupInFields(int line, int column, int expSymbols, List<String> expSymbolNames) {
+        Project project = BCompileUtil.loadProject("test-src/symbol_lookup_in_fields.bal");
+        Package currentPackage = project.currentPackage();
+        ModuleId defaultModuleId = currentPackage.getDefaultModule().moduleId();
+        PackageCompilation packageCompilation = currentPackage.getCompilation();
+        SemanticModel model = packageCompilation.getSemanticModel(defaultModuleId);
+        Document srcFile = getDocumentForSingleSource(project);
+
+        BLangPackage pkg = packageCompilation.defaultModuleBLangPackage();
+        ModuleID moduleID = new BallerinaModuleID(pkg.packageID);
+
+        Map<String, Symbol> symbolsInFile = getSymbolsInFile(model, srcFile, line, column, moduleID);
+        assertEquals(symbolsInFile.size(), expSymbols);
+
+        for (String symName : expSymbolNames) {
+            assertTrue(symbolsInFile.containsKey(symName), "Symbol not found: " + symName);
+        }
+    }
+
+    @DataProvider(name = "FieldSymbolPosProvider")
+    public Object[][] getFieldSymbolPositions() {
+        return new Object[][]{
+                {2, 4, 4, asList("Foo", "Bar", "Person", "PersonObj")},
+                {8, 4, 4, asList("Foo", "Bar", "Person", "PersonObj")},
+                {16, 8, 8, asList("Foo", "Bar", "Person", "PersonObj", "init", "inc", "self", "x")},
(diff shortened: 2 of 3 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -123,9 +123,9 @@ public BallerinaSemanticModel(BLangPackage bLangPackage, CompilerContext context
         for (Map.Entry<Name, List<Scope.ScopeEntry>> entry : scopeSymbols.entrySet()) {
             Name name = entry.getKey();
             List<Scope.ScopeEntry> scopeEntries = entry.getValue();
-
             for (Scope.ScopeEntry scopeEntry : scopeEntries) {
-                addToCompiledSymbols(compiledSymbols, scopeEntry, cursorPos, name, statesSet);
+                boolean isChildOfCurrentNode = symbolEnv.scope.owner.equals(scopeEntry.symbol.owner);
```
pubudu91, line 127: This may be problematic since we don't have an `equals()` implementation for symbols and our symbol implementation doesn't guarantee that the same symbol will always be in use.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
```diff
@@ -474,12 +474,24 @@ private void addToCompiledSymbols(Set<Symbol> compiledSymbols, Scope.ScopeEntry
...
-        addToCompiledSymbols(compiledSymbols, scopeEntry.next, cursorPos, name, states);
+        addToCompiledSymbols(compiledSymbols, scopeEntry.next, cursorPos, name, isChildOfCurrentNode, states);
+    }
+
+    private boolean isFieldSymbol(Symbol compiledSymbol, boolean isChildOfCurrentNode) {
+        switch (compiledSymbol.kind()) {
+            case CLASS_FIELD:
+            case OBJECT_FIELD:
+            case RECORD_FIELD:
+                return isChildOfCurrentNode;
```
pubudu91, line 491: This is a bit weird/misleading. Basically attempting to do 2 things and misleadingly named.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
sanjana: In that case, we need to perform different equivalency checks for the `pkgID`, `name`, and the `pos` of these owner symbols. 
That would derive that the two symbols are equal. 

@pubudu91, WDYT?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
sanjana: Here the basic intention is to identify if the symbol we've encountered is a field symbol, which is also a field of the current node which the cursor is positioned at.

Those two can be broken down into two different checks. 
Also, if the current node's owner symbol can be passed down to this stage, we could check the owner symbols' equivalency checks once it is verified that it is a field symbol. 

@pubudu91, any other suggestions to improve this proposed filtering mechanism?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
pubudu91: Hmm yeah. Would have to try something like that I guess 🤔  Also, see if deriving this info using the available node info is a viable approach

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
pubudu91: The high level filtering idea isn't an issue. It's how it's implemented that's problematic. It needs to be refactored in a way that it's not misleading for someone reading it.

language-server/modules/langserver-core/src/test/resources/completion/class_def/config/config25.json
```diff
@@ -493,24 +493,6 @@
...
       "insertTextFormat": "Snippet"
-    },
-    {
-      "label": "new(int field1, int field2)",
-      "kind": "Function",
-      "detail": "()",
-      "documentation": {
-        "right": {
-          "kind": "markdown",
-          "value": "**Package:** _ballerina/module1:0.1.0_  \n  \n  \n**Params**  \n- `int` field1  \n- `int` field2"
```
nadeeshaan, line 504: This language server's completions have been failed. We need to check how this compiler changes have been affected

nadeeshaan (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
sanjana: Updated with the above-mentioned approach.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/BallerinaSemanticModel.java
sanjana: Refactored the filtering implementation.

language-server/modules/langserver-core/src/test/resources/completion/class_def/config/config25.json
sanjana: @nadeeshaan isn't this due to the changes made in the filtering mechanism at `visibleSymbols()` method of the Semantic API?

pubudu91:
The issue @nadeeshaan mentioned is still there isn't it?

sanjana:
> The issue @nadeeshaan mentioned is still there isn't it?

Yes, and blocked by #33089

language-server/modules/langserver-core/src/test/resources/completion/class_def/config/config25.json
sanjana: Updated with a fix.

mohanvive:
@sanjana can you please resolve the conflicts?

sanjana:
> @sanjana can you please resolve the conflicts?

@mohanvive, resolved the merge conflicts and fixed a few other failing LS test cases as well.

PR #32873, opened 2021-09-23
Fix incorrect member types for union type definition

### LINKED ISSUE #32569: Incorrect member types for union type definition
**Description:**

Consider the scenario,

```
type t "0"|"1"|t1;
type t1 "a"|"b"|"c";
```
In `memberTypeDescriptors()`,  we get `"0|1"` as a union type.

Refer to https://github.com/ballerina-platform/ballerina-lang/pull/32540#discussion_r701611819

image

**Steps to reproduce:**

**Affected Versions:**

dulajdilshan:
## Purpose
$title

Fixes #32569

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaUnionTypeSymbol.java
```diff
@@ -70,7 +71,17 @@ public BallerinaUnionTypeSymbol(CompilerContext context, ModuleID moduleID, BFin
                 TypesFactory typesFactory = TypesFactory.getInstance(this.context);
 
                 for (BType memberType : ((BUnionType) this.getBType()).getMemberTypes()) {
-                    members.add(typesFactory.getTypeDescriptor(memberType));
+                    if (typesFactory.isTypeReference(memberType, memberType.tsymbol, false)
+                            || memberType.getKind() != TypeKind.FINITE) {
+                        members.add(typesFactory.getTypeDescriptor(memberType));
```
pubudu91, line 76: Why do we have to check whether it's a type ref? Isn't checking whether it's not a finite type isn't enough? 🤔

dulajdilshan: Let's say we have 
```ballerina
type T2 1|2|3;
type T4 "1"|"2"|T2;
```
If we get members of `T4` we get two `FiniteType`s (`T2` and `"1"|"2"|`). Therefore, we need to check whether `T2` is a type ref or not.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaUnionTypeSymbol.java
```diff
@@ -70,7 +71,17 @@ public BallerinaUnionTypeSymbol(CompilerContext context, ModuleID moduleID, BFin
...
+                        continue;
+                    }
+
+                    BFiniteType finiteType = (BFiniteType) memberType;
+                    for (BLangExpression value : finiteType.getValueSpace()) {
+                        ModuleID moduleID = getModule().isPresent() ? getModule().get().id() : null;
+                        members.add(new BallerinaSingletonTypeSymbol(this.context, moduleID, value, value.getBType()));
```
pubudu91, line 83: Not sure this is the correct way to do this. Doesn't `value.getBType()` return the broader type of that expression? IINM, we'll have to create individual `BFiniteType` types here for each of the member values.

pubudu91 (approved):

PR #32891, opened 2021-09-24
Remove adding unnecessary any cast when creating tuple values

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/32892

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
@@ -5833,10 +5833,22 @@ public class Desugar extends BLangNodeVisitor {
             result = rewriteExpr(typedescExpr);
             return;
         }
-        tupleLiteral.exprs.forEach(expr -> {
+        List<BLangExpression> exprs = tupleLiteral.exprs;
+        BTupleType tupleType = (BTupleType) tupleLiteral.getBType();
+        List<BType> tupleMemberTypes = tupleType.tupleTypes;
+        int tupleMemberExprSize = tupleMemberTypes.size();
+        int tupleExprSize = exprs.size();
+        for (int i = 0; i < tupleExprSize; i++) {
+            BLangExpression expr = exprs.get(i);
             BType expType = expr.impConversionExpr == null ? expr.getBType() : expr.impConversionExpr.getBType();
-            types.setImplicitCastExpr(expr, expType, symTable.anyType);
-        });
+            BType targetType;
+            if (i < tupleMemberExprSize) {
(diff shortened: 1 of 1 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5833,10 +5833,22 @@ public void visit(BLangTupleLiteral tupleLiteral) {
             result = rewriteExpr(typedescExpr);
             return;
         }
-        tupleLiteral.exprs.forEach(expr -> {
+        List<BLangExpression> exprs = tupleLiteral.exprs;
+        BTupleType tupleType = (BTupleType) tupleLiteral.getBType();
+        List<BType> tupleMemberTypes = tupleType.tupleTypes;
+        int tupleMemberExprSize = tupleMemberTypes.size();
```
MaryamZi, line 5839: Shall we say `tupleMemberTypeSize` instead?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5833,10 +5833,22 @@ public void visit(BLangTupleLiteral tupleLiteral) {
...
+            BLangExpression expr = exprs.get(i);
             BType expType = expr.impConversionExpr == null ? expr.getBType() : expr.impConversionExpr.getBType();
-            types.setImplicitCastExpr(expr, expType, symTable.anyType);
-        });
+            BType targetType;
+            if (i < tupleMemberExprSize) {
+                targetType = tupleMemberTypes.get(i);
+            } else {
+                targetType = tupleType.restType;
+            }
```
MaryamZi, line 5849: Can use a ternary operator here also?

MaryamZi (commented):
Even with this fix, if every (or most) members need to be boxed, it will still result in the error, right?

Shall/can we also add tests?

warunalakshitha:
> Even with this fix, if every (or most) members need to be boxed, it will still result in the error, right?
> 
> Shall/can we also add tests?

I can add tests but wondering do we need? . Existing tests should cover the code. If members need to box it will box since types.setImplicitCastExpr(expr, expType, targetType) will handle it.

MaryamZi (approved):

Issue #32894, opened 2021-09-24
Cleanup unnecessary temp variables in frame classes


warunalakshitha:
**Description:**
If given temp bir variables are not used after the same bir block, those might not need to be store in the frame classes. We need to check and remove those from frame classes.

PR #32923, opened 2021-09-27
Remove unnecessary temp variables from method frames

warunalakshitha:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/32894

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/optimizer/BIROptimizer.java
```diff
@@ -312,86 +316,70 @@ private void reuseTempVariables(BIRFunction birFunction, OptimizerEnv funcOpEnv)
...
+                    removableOpsMap.put(birOperand, reusableVar);
+                    reusableList.remove(reusableVar);
+                    reusableList.remove(variableDcl);
+                    replaceableVarSet.add(reusableVar);
                     continue;
                 }
-                secondLoadedSet.add(variableDcl);
                 BIRVariableDcl reusableVar = removableVarVsReplacementVarMap.get(variableDcl);
-                // Same var can be used by multiple operands. we need to collect all and then replace the it
+                Set<BIRVariableDcl> reusableList = typeVsReusableVarsMap.get(variableDcl.type);
```
HindujaB, line 346: Shall we have different name for the `reusableList` since we have another `reusableList` in the same function?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/optimizer/BIROptimizer.java
```diff
@@ -312,86 +316,70 @@ private void reuseTempVariables(BIRFunction birFunction, OptimizerEnv funcOpEnv)
...
+            }
         }
 
         @Override
         public void visit(BIRNode.BIRBasicBlock birBasicBlock) {
+            this.env.currentBB = birBasicBlock;
             this.env.newInstructions = new ArrayList<>();
             birBasicBlock.instructions.forEach(i -> this.optimizeNonTerm(i, this.env));
             birBasicBlock.instructions = this.env.newInstructions;
+            this.env.isTerminator = true;
```
HindujaB, line 380: Shall we move this assignment inside the `optimizeTerm` method? Changing it before & after the `optimizeTerm()` call seems a bit unclear.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/optimizer/BIROptimizer.java
warunalakshitha: It actually the same list.  This list is inside a type vs reusable list map. At first, we check if there is any reusable var with the same type to reuse. The second time when the variable is used, we can add that variable to the reusable list so any other variable can use.

HindujaB (approved):
LGTM

PR #32941, opened 2021-09-28
Set Appropriate Positions of `BLangXMLNavigationAccess` expressions.

### LINKED ISSUE #32835: [Semantic API] `symbol()` not working for `xml-step-extend`
Consider the following:
```ballerina
xmlns "foo" as ns;

int indx = 0;

function testXMLStepExpr() {
    xml x1 = xml `<ns:root><ns:child></ns:child></ns:root>`;
    xml x2 = x1/<ns:child>[<cursor1>indx];
    string s = x1/**/<ns:child>.<cursor2>toString();
}
```
For the illustrated cursor positions, we aren't getting any symbols.

sanjana:
Fixes #32835

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
```diff
@@ -2221,10 +2221,11 @@ public BLangNode transform(IndexedExpressionNode indexedExpressionNode) {
             indexBasedAccess.expr = ((BLangGroupExpr) expression).expression;
             BLangGroupExpr group = (BLangGroupExpr) TreeBuilder.createGroupExpressionNode();
             group.expression = indexBasedAccess;
-            group.pos = getPosition(indexedExpressionNode);
+            group.pos = indexBasedAccess.pos;
```
mohanvive, line 2224: @sanjana Do we need this change to resolve this issue?

sanjana: @mohanvive the `indexBasedAccess.pos` is declared at https://github.com/ballerina-platform/ballerina-lang/pull/32941/files#diff-f19078e211ec6eaa361839b611670e633460cbe1f9193a55e6e11ffbce0409ecR2193 

I reused its value here in order to avoid re-computing `getPosition(indexedExpressionNode)`. 

It is not directly required for this issue. But, an optimization that I came across.

pubudu91: Isn't this the wrong position to be setting for this? 🤔 IINM, the above is a case like the following: `<1>(expr)<2>[1]<3>`. For the group expr, the position should be `<1>` to `<2>`. The above is setting `<1>` to `<3>` for it

tests/ballerina-compiler-api-test/src/test/resources/test-src/symbol_at_cursor_basic_test.bal
```diff
@@ -164,3 +164,13 @@ function test(string name, int age) {
...
 }
+
+xmlns "foo" as ns;
+
+int indx = 0;
+
+function testXMLStepExpr() {
+    xml x1 = xml `<ns:root><ns:child></ns:child></ns:root>`;
+    xml x2 = x1/<ns:child>[indx];
+    string s = x1/**/<ns:child>.toString();
```
pubudu91, line 175: These tests aren't needed. Only have to enable the tests in `SymbolsInXMLNavigationExprsTest`

sanjana: Updated.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
sanjana: Consider the example code `int a = <int> (dataMap)["a"];`, where the `group` is `(dataMap)["a"]`, and `group.expression`'s position is set for `(dataMap)`. 
Therefore, the above position setting can be considered as accurate. 
In that case, the position of `group` shall be `<1>` to `<3>`.

@lochana-chathura WDYT?

lochana-chathura: As per L2223 `group` seem to present the whole indexed access expr here. so should the position?

Here the AST representation is confusing to me. Why would we create a `BLangGroupExpr` node here, when the actually node is not in the form of `(` `expression` `)`.  Ideally we need to create `BLangIndexBasedAccess` node with `bLangIndexBasedAccess.expr` being a `BLangGroupExpr`.

pubudu91: Exactly. Basically that whole if block is unnecessary right?

lochana-chathura: yeah, unnecessary.

PR #32943, opened 2021-09-28
[LS] Add ModuleVarToListnerDecl code action

### LINKED ISSUE #30384: Code action to make a module listener variable a module listener declaration
**Description:**
Can we do $title?

```ballerina
Listener ln = new; // Code action to make this `listener Listener ln = new;`?

service on ln { // ERROR: invalid listener attachment 
    
}

class Listener {
    public function attach(service object {} s, string|string[]? name = ()) returns error? {
    }

    public function detach(service object {} s) returns error? {
    }

    public function 'start() returns error? {
    }

    public function gracefulStop() returns error? {
    }

    public function immediateStop() returns error? {
    }
}
```

malinthar:
## Purpose
$subject as depicted below.

https://user-images.githubusercontent.com/35211477/135050565-6e2c425b-9291-4164-aa67-625aa78af1bb.mp4

Fixes #30384 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/ModVarToListenerDeclCodeAction.java
```diff
@@ -0,0 +1,137 @@
...
+        if (!(DiagnosticErrorCode.INVALID_LISTENER_ATTACHMENT.diagnosticId()
+                .equals(diagnostic.diagnosticInfo().code()))) {
+            return Collections.emptyList();
+        }
+
+        Node matchedNode = positionDetails.matchedNode();
+        if (matchedNode == null) {
+            return Collections.emptyList();
+        }
+        Optional<Pair<CaptureBindingPatternNode, String>> captureBindingPatternNodeFilePathPair =
```
nadeeshaan, line 70: variable name is too long

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/ModVarToListenerDeclCodeAction.java
```diff
@@ -0,0 +1,137 @@
...
+        TypedBindingPatternNode typedBindingPatternNode =
+                (TypedBindingPatternNode) captureBindingPatternNodeFilePathPair.get().getLeft().parent();
+        List<CodeAction> actions = new ArrayList<>();
+        List<TextEdit> textEdits = new ArrayList<>();
+        Position pos = CommonUtil.toRange(typedBindingPatternNode.lineRange()).getStart();
+        Position insertPos = new Position(pos.getLine(), pos.getCharacter());
+        textEdits.add(new TextEdit(new Range(insertPos, insertPos),
+                SyntaxKind.LISTENER_KEYWORD.stringValue() + " "));
+        String commandTitle = String.format(CommandConstants.CONVERT_MODULE_VAR_TO_LISTENER_DECLARATION,
+                matchedNode.toSourceCode());
```
nadeeshaan, line 86: when we get the value from the `sourceCode` API for displaying purposes, make sure you trim the value.

nadeeshaan, line 86: when we get the value from the `sourceCode` API for displaying purposes, make sure you trim the value.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/ModVarToListenerDeclCodeAction.java
```diff
@@ -0,0 +1,137 @@
...
+                return Optional.empty();
+            }
+            Path filePath = project.get().sourceRoot()
+                    .resolve("modules").resolve(symbol.get().getModule().get().id().modulePrefix())
+                    .resolve(location.get().lineRange().filePath());
+            Optional<SyntaxTree> syntaxTree = context.workspace().syntaxTree(filePath);
+            if (syntaxTree.isEmpty()) {
+                return Optional.empty();
+            }
+            foundNode = CommonUtil.findNode(symbol.get(), syntaxTree.get());
```
nadeeshaan, line 116: Shall we refactor the `CommonUtil.findNode` API to return an optional? WDYT?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/ModVarToListenerDeclCodeAction.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public int priority() {
+        return super.priority();
+    }
+
+    @Override
+    public String getName() {
+        return null;
```
nadeeshaan, line 135: Why do we return `null` here?

nadeeshaan (changes requested):
Here, we do not check whether the variable is with the type of listener. It is necessary to do this check and avoid modifying the variable and also should not suggest the code action at the first place.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/ModVarToListenerDeclCodeAction.java
malinthar: Have missed this

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/ModVarToListenerDeclCodeAction.java
malinthar: That's better. Will do

malinthar:
> Here, we do not check whether the variable is with the type of listener. It is necessary to do this check and avoid modifying the variable and also should not suggest the code action at the first place.

I have added the check for variable symbol. The `INVALID_LISTENER_ATTACHMENT` diagnostic is added in the semantic analyzer for variable symbol  only in the following case where the listener flag is on. 
https://github.com/ballerina-platform/ballerina-lang/blob/f907ee42200ba068b56e2938bfee663e67c291a7/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java#L3594
Therefore if  a simple var has the listener keyword, this diagnostic will not be pushed. 

Do we have to check for the listener keyword?

nadeeshaan (changes requested):

nadeeshaan:
> > Here, we do not check whether the variable is with the type of listener. It is necessary to do this check and avoid modifying the variable and also should not suggest the code action at the first place.
> 
> @nadeeshaan , I have added the check for variable symbol. The `INVALID_LISTENER_ATTACHMENT` diagnostic is added in the semantic analyzer for variable symbol only in the following case where the listener flag is on.
> 
> https://github.com/ballerina-platform/ballerina-lang/blob/f907ee42200ba068b56e2938bfee663e67c291a7/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java#L3594
> 
> 
> Therefore if a simple var has the listener keyword, this diagnostic will not be pushed.
> Do we have to check for the listener keyword? But then, it is safe to add the check. Could you suggest a method to do that? Will that be a qualifier of the symbol or do we have to traverse the children of the variable node and check for the existence of the listener keyword.

This is not regarding the listener keyword. Consider the following scenario where the variable is a record type. We get the same diagnostic.
Screenshot 2021-10-04 at 08 19 49

malinthar:
> This is not regarding the listener keyword. Consider the following scenario where the variable is a record type. We get the same diagnostic. Screenshot 2021-10-04 at 08 19 49

Done

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/ModVarToListenerDeclCodeAction.java
```diff
@@ -0,0 +1,130 @@
...
+    public List<CodeAction> getDiagBasedCodeActions(Diagnostic diagnostic,
+                                                    DiagBasedPositionDetails positionDetails,
+                                                    CodeActionContext context) {
+        if (!(DiagnosticErrorCode.INVALID_LISTENER_ATTACHMENT.diagnosticId()
+                .equals(diagnostic.diagnosticInfo().code()))) {
+            return Collections.emptyList();
+        }
+
+        Node matchedNode = positionDetails.matchedNode();
+        if (matchedNode == null) {
```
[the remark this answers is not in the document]
malinthar: Done

nadeeshaan:
Please add a test case where the listener is from an external library/ module. You can use module1 in the language server (language-server-stdlib)

nadeeshaan (dismissed):
LGTM

malinthar:
> Please add a test case where the listener is from an external library/ module. You can use module1 in the language server (language-server-stdlib)

Done

PR #32970, opened 2021-09-29
Add anon function snippets to the expression completion list

### LINKED ISSUE #25182: Add anon function snippets to the expression completion list
**Description:**
Currently we only support the `function` keyword in the expression contexts. We need to support the anon function snippets for this context.

**Steps to reproduce:**
Ex: return <cursor>

Here the cursor is within the expression context where we can suggest the anon function expression.

**Affected Versions:**
SwanLake-Preview3-SNAPSHOT

### LINKED ISSUE #25490: Add snippets for anonymous function expression deriving from the RHS
**Description:**
Currently, anonymous function context suggestions are provided for individual components such as parameter context, return context and etc. We can improve the usability by providing snippets analysing the LHS of an expression

TharushiJay:
## Purpose
This PR adds support for anon function snippets in the expression context.

Fixes #25182
Fixes #25490

## Samples

Screen Recording 2021-09-29 at 13 23 03 (1)

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -676,4 +679,74 @@ protected boolean onSuggestionsAfterQualifiers(BallerinaCompletionContext contex
...
+            int argIndex = 1;
+
+            Set<String> visibleSymbolNames = visibleSymbols
+                    .stream()
+                    .map(Symbol::getName)
+                    .filter(Optional::isPresent)
+                    .map(Optional::get)
+                    .collect(Collectors.toSet());
+                
+            if (functionTypeSymbol.params().isEmpty()) {
```
malinthar, line 711: Shall we check the case where we don't have any parameters? For example, 
```
function myFunc returns function  {
   <cursor>
}
```

nadeeshaan (changes requested):
Shall we consider the other expression contexts where the annon function is a valid expression? Specially, the function argument context.

malinthar (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -676,4 +679,74 @@ protected boolean onSuggestionsAfterQualifiers(BallerinaCompletionContext contex
...
+                TypeSymbol parameterTypeSymbol = CommonUtil.getRawType(parameterSymbol.typeDescriptor());
+                varName = CommonUtil.generateParameterName(varName, argIndex, parameterTypeSymbol, 
+                        visibleSymbolNames);
+                args.add(FunctionGenerator.getParameterTypeAsString(context, parameterTypeSymbol) + " " + varName);
+                visibleSymbolNames.add(varName);
+                argIndex++;
+            }
+
+            Optional<TypeSymbol> returnTypeSymbol = functionTypeSymbol.returnTypeDescriptor();
+            if (returnTypeSymbol.isEmpty()) {
```
malinthar, line 726: Similarly, shall we check what happens when there is a NIL return type

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -676,4 +679,74 @@ protected boolean onSuggestionsAfterQualifiers(BallerinaCompletionContext contex
...
         return lastQualifier.textRange().endOffset() < cursor;
     }
+
+    /**
+     * Get completions for anonymous function definition.
+     *
+     * @param context completion context.
+     * @return completion item
+     */
+    protected List<LSCompletionItem> getAnonFunctionDefCompletions(BallerinaCompletionContext context) {
```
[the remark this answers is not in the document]
TharushiJay: Yes, changed to `Optional<SnippetCompletionItem>`.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -676,4 +679,74 @@ protected boolean onSuggestionsAfterQualifiers(BallerinaCompletionContext contex
...
+            if (returnType.isPresent()) {
+                // returns clause
+                returnsClause = "returns " + returnType.get();
+                // return statement
+                Optional<String> defaultReturnValue = CommonUtil.getDefaultValueForType(returnTypeSymbol.get());
+                if (defaultReturnValue.isPresent()) {
+                    returnStmt = "return " + defaultReturnValue.get() + CommonKeys.SEMI_COLON_SYMBOL_KEY;
+                }
+            }
+            String snippet = "function(" + String.join(", ", args) + ") " + returnsClause + " {" +
```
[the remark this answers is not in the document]
TharushiJay: Appended `returnsClause` inside `returnType.isPresent()` check to avoid this.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -676,4 +679,74 @@ protected boolean onSuggestionsAfterQualifiers(BallerinaCompletionContext contex
...
+                // returns clause
+                returnsClause = "returns " + returnType.get();
+                // return statement
+                Optional<String> defaultReturnValue = CommonUtil.getDefaultValueForType(returnTypeSymbol.get());
+                if (defaultReturnValue.isPresent()) {
+                    returnStmt = "return " + defaultReturnValue.get() + CommonKeys.SEMI_COLON_SYMBOL_KEY;
+                }
+            }
+            String snippet = "function(" + String.join(", ", args) + ") " + returnsClause + " {" + 
+                    CommonUtil.LINE_SEPARATOR + "\t" + returnStmt + CommonUtil.LINE_SEPARATOR + "}${1}";
```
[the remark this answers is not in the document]
TharushiJay: Added `returnStmt` and `paddingStr` to `body`.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
TharushiJay: Generated variable names only if `functionTypeSymbol.params().isPresent()` to support this.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
TharushiJay: Works as expected,
```
function () returns () testVar = function () returns () {
    
};
```

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ItemResolverConstants.java
```diff
@@ -60,6 +60,7 @@
     public static final String DEFAULT = "default";
     public static final String ROLLBACK = "rollback";
 
+    public static final String ANON_FUNCTION = "anon function";
```
nadeeshaan, line 63: Shall we use something like `function (..) {..}` as the label? Also, make sure to use function as the filter text, instead of the label itself.

nadeeshaan (changes requested):
We have to fix the sorting as well. we should increase the priority of the snippet in valid contexts. I would suggest,
1 - Variables - for the variables with the function type
2 - anon function snippet

malinthar:
> > Changes look good. I have few minor suggestions. When we substitute the anon func def as a function argument shouldn't we insert that in a single line? WDYT?
> 
> This is a valid use case. But, I think this will be a bit complex since we will have to consider the context. For example, when assigning a variable, annon function can span multiple lines. But when assigning a function call's parameter, we have to make it a one liner. IMO, we should let this be as it is since given the usage of annon functions will be less and we have lot of other spec conformance tasks to complete before GA. WDYT? @malinthar @nadeeshaan

That's true. Let's proceed with the current implementation. +1

Issue #33097, opened 2021-10-06
Disallow in match guards calling a function that may mutate the value


hasithaa:
Spec issue https://github.com/ballerina-platform/ballerina-spec/issues/895

PR #33219, opened 2021-10-13
Disallow in match guards function/method calls that may mutate the value being matched

### LINKED ISSUE #33180: Code analysis doesn't happen for the match guard
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type Foo record {
    int x;
};

public function main() {
    Foo f1 = {x: 1, y: 2}; // error: invalid key 'y': identifiers cannot be used as rest field keys, expected a string literal or an expression

    int i = 1;

    match i {
        1 if (let Foo f2 = {x: 1, y: 2} in f2.x == 2) => { // no error for `y: 2`

        }
    }
}
```

**Affected Versions:**
slbeta3

### LINKED ISSUE #33186: No semantic error on invalid static type of match guard
**Description:**
$title. Crashes at the backend instead.

**Steps to reproduce:**
```ballerina
isolated function fn() {
}

public function main() {
    int i = 1;

    match i {
        1 if fn() => {

        }
    }
}
```

```cmd
$ bal run x.bal 
Compiling source
        x.bal
error: casting is not supported from '()' to 'boolean'
```

```ballerina
public function main() {
    int i = 1;

    match i {
        1 if 1 + 2 => {

        }
    }
}
```

```cmd
$ bal run x.bal 
Compiling source
        x.bal
error: casting is not supported from 'int' to 'boolean'
```

**Affected Versions:**
slbeta3

MaryamZi:
## Purpose
$title.

A function/method call will be allowed in a match guard only if
1. the matched expression's type is a subtype of `readonly` or
2. the function/method is `isolated` and all of the arguments have types that are subtypes of `readonly`

```ballerina
public function main() {
    int x = 1;

    match x {
        0 if isValid(x) =>  { // OK since the type of the expr being matched is a subtype of `readonly`
        }
    }
}

int val = 0;

function isValid(int i) returns boolean => i > val;
```

```ballerina
public function main() {
    record {} x = {};

    match x {
        var {a, b} if removeA(x) =>  { // no longer allowed, since the type of the expr being matched and the type of the argument are not subtypes of `readonly`
        }
        var {a, c} =>  {
        }
    }
}

isolated function removeA(record {} rec) returns boolean {
    if rec.hasKey("a") {
        _ = rec.remove("a");
        return true;
    }
    return false;
}
```

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/33097

This PR also fixes the following.
Fixes #33186 - No semantic error on invalid static type of match guard
Fixes #33180 - Code analysis doesn't happen for the match guard

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


hasithaa:
@MaryamZi @gimantha We need to capture this as a breaking change and inform the team.

PR #33277, opened 2021-10-18
Add type reference completions items to type test context

### LINKED ISSUE #32607: Suggestions should be context-aware
I am using Ballerina Swan Lake Beta2.  Let me describe the requirement using the following image. 
Screen Shot 2021-09-06 at 8 35 11 PM

IMO, the first two completion items should be `http:Response` and `http:ClientError`.

malinthar:
## Purpose
Add type reference completion items to type test(type guard) context and fix sorting to show expression's types at the top. 

Fixes #32607 

new pic

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/TypeTestExpressionNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/TypeTestExpressionNodeContext.java
@@ -15,6 +15,7 @@
  */
 package org.ballerinalang.langserver.completions.providers.context;
 
+import io.ballerina.compiler.api.ModuleID;
 import io.ballerina.compiler.api.symbols.Symbol;
 import io.ballerina.compiler.api.symbols.TypeDescKind;
 import io.ballerina.compiler.api.symbols.TypeReferenceTypeSymbol;
@@ -24,7 +25,9 @@ import io.ballerina.compiler.syntax.tree.QualifiedNameReferenceNode;
 import io.ballerina.compiler.syntax.tree.Token;
 import io.ballerina.compiler.syntax.tree.TypeTestExpressionNode;
 import io.ballerina.projects.Module;
+import io.ballerina.projects.ModuleId;
 import org.ballerinalang.annotation.JavaSPIService;
+import org.ballerinalang.langserver.common.utils.CommonUtil;
 import org.ballerinalang.langserver.common.utils.completion.QNameReferenceUtil;
 import org.ballerinalang.langserver.commons.BallerinaCompletionContext;
 import org.ballerinalang.langserver.commons.completion.LSCompletionException;
@@ -35,6 +38,7 @@ import org.ballerinalang.langserver.completions.providers.AbstractCompletionProv
 import org.ballerinalang.langserver.completions.util.SortingUtil;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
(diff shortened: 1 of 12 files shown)


gayalkuruppu (commented):
Is there a way that we can give priority to keywords such as `true` and `false` (and `null` which is a missing keyword) which are not `type`s
Screenshot 2021-12-06 at 14 32 39

malinthar:
> Is there a way that we can give priority to keywords such as `true` and `false` (and `null` which is a missing keyword) which are not `type`s Screenshot 2021-12-06 at 14 32 39

The true, false, () are values. Those are not type completion items. Hence, the only approach is special casing said values. But it would not be ideal in terms of generality. Therefore, shall we not prioritize these? @gayalkuruppu

gayalkuruppu:
> > Is there a way that we can give priority to keywords such as `true` and `false` (and `null` which is a missing keyword) which are not `type`s Screenshot 2021-12-06 at 14 32 39
> 
> The true, false, () are values. Those are not type completion items. Hence, the only approach is special casing said values. But it would not be ideal in terms of generality. Therefore, shall we not prioritize these? @gayalkuruppu

Agree +1

nadeeshaan:
Are we only considering the types coming from the modules? If so shall we consider the type in a generic manner? Because this should be supported for types coming from the modules, user defined types in the same module and the builtin types as well.

malinthar:
> Are we only considering the types coming from the modules? If so shall we consider the type in a generic manner? Because this should be supported for types coming from the modules, user defined types in the same module and the builtin types as well.

We have considered all type-defs. Refer the following image.
typedef

The two fold sorting order I have followed here is as follows. 
1. Types of the considered variable (expression of the type test context)
2. Other completion items

Each of the above categories again go through a second fold of sorting in which we use the following 
sorting method for type-desc-context. 
https://github.com/ballerina-platform/ballerina-lang/blob/68db4af5146e84529a7a71c6fee4a16fd3f7f374/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SortingUtil.java#L183

@nadeeshaan

nadeeshaan:
> > Are we only considering the types coming from the modules? If so shall we consider the type in a generic manner? Because this should be supported for types coming from the modules, user defined types in the same module and the builtin types as well.
> 
> We have considered all type-defs. Refer the following image. typedef
> 
> The two fold sorting order I have followed here is as follows.
> 
> 1. Types of the considered variable (expression of the type test context)
> 2. Other completion items
> 
> Each of the above categories again go through a second fold of sorting within the catregory in which we use the following sorting method for type-desc-context.
> 
> https://github.com/ballerina-platform/ballerina-lang/blob/68db4af5146e84529a7a71c6fee4a16fd3f7f374/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SortingUtil.java#L183
> 
> @nadeeshaan

@malinthar please refer the following

Screenshot 2021-12-13 at 14 55 05

malinthar:
> > > Are we only considering the types coming from the modules? If so shall we consider the type in a generic manner? Because this should be supported for types coming from the modules, user defined types in the same module and the builtin types as well.
> > 
> > 
> > We have considered all type-defs. Refer the following image. typedef
> > The two fold sorting order I have followed here is as follows.
> > 
> > 1. Types of the considered variable (expression of the type test context)
> > 2. Other completion items
> > 
> > Each of the above categories again go through a second fold of sorting within the catregory in which we use the following sorting method for type-desc-context.
> > https://github.com/ballerina-platform/ballerina-lang/blob/68db4af5146e84529a7a71c6fee4a16fd3f7f374/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SortingUtil.java#L183
> > 
> > @nadeeshaan
> 
> @malinthar please refer the following
> 
> Screenshot 2021-12-13 at 14 55 05

This is due to a limitation in the semantic API. 

```ballerina
type Type1 string | int;
type Type2 string;


public function main() {
      Type1 | Type2 |int | mod1:MyType myvar;
      if myvar is 
}
```

Here the member types of myvar is as follows. 

memeber types

In the following image the Type1 has expanded to string,int members. Due to this we do not recognize Type1 as a subtype of the type of myvar. Hense it's not given a higher precedence.

IMS94 (approved):
LGTM!

gayalkuruppu:
LGTM

PR #33302, opened 2021-10-20
Fix string unescaping with `\u{5C}`

### LINKED ISSUE #33051: Using numeric escape for backslash does not work in string literal
**Description:**
Consider the bellow sample
```bal
function init() {
    string c = "A \u{5C} B";
    io:println(c); // @output A \ B
}
```

This should print `A \ B` instead of  `A  B` 

**Affected Versions:**
Current master

lochana-chathura:
## Purpose
$subject.

Fixes #33051 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeTransformer.java
```diff
@@ -5336,7 +5336,7 @@ private BLangLiteral createSimpleLiteral(Node literal, SyntaxKind sign, boolean
 
             if (type != SyntaxKind.TEMPLATE_STRING && type != SyntaxKind.XML_TEXT_CONTENT) {
                 try {
-                    text = IdentifierUtils.unescapeJava(IdentifierUtils.unescapeUnicodeCodepoints(text));
+                    text = IdentifierUtils.unescapeJava(IdentifierUtils.unescapeUnicodeCodepoints(text, true));
```
rdhananjaya, line 5339: Shall we create a new method in `IdentifierUtils` which perform the operation done by both of these function calls. the reason for this is only public use case for `unescapeUnicodeCodepoints` with 2 param is this one. So let's make `unescapeUnicodeCodepoints` private and let's give a meaningful name to the new method.

lochana-chathura: Gave the name as `unescapeBallerina` similar to `unescapeJava` method in the same class.

rdhananjaya (approved):

PR #33376, opened 2021-10-25
Check invalid contextually expected type for error ctor

### LINKED ISSUE #33175: No error when a union type that does not include an error type is used as the contextually-expected type for an error constructor
**Description:**
In below mentioned case compilation finishes without failure even though its expected to fail because given and required return types are not similar

**Steps to reproduce:**

```
import ballerina/io;

function resolveTestSemtype(string|int t) returns TestType {
    if t is string {
        io:println("test");
    }
    else {
        return t;
    }
    return error("unreachable");
}

type TestType int|string;
```

KRVPerera:
Fixes #33175

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/errorconstructorexpr/ErrorConstructorExprTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/errorconstructorexpr/ErrorConstructorExprTest.java
@@ -99,14 +99,14 @@ public class ErrorConstructorExprTest {
         validateError(negativeSemanticResult, i++, "cannot infer type of the error from '(error|ErrorB)'", 51, 20);
         validateError(negativeSemanticResult, i++, "unknown type 'Blah'", 60, 27);
         validateError(negativeSemanticResult, i++, "cannot create a new error value from 'ErrorU1'", 62, 19);
-        validateError(negativeSemanticResult, i++, "compatible type for error constructor expression not found" +
-                " in type 'error'", 63, 13);
+        validateError(negativeSemanticResult, i++, "a type compatible with error constructor expressions not found " +
+                "expected 'error', found 'ErrorU2'", 63, 13);
         validateError(negativeSemanticResult, i++, "cannot create a new error value from 'ErrorU2'", 63, 19);
         validateError(negativeSemanticResult, i++, "undefined error type descriptor 'ErrorU3'", 64, 19);
         validateError(negativeSemanticResult, i++, "incompatible types: expected '(int|string)', found 'error'",
                 66, 18);
-        validateError(negativeSemanticResult, i++, "compatible type for error constructor expression not found" +
-                " in type '(int|string)'", 67, 16);
+        validateError(negativeSemanticResult, i++, "a type compatible with error constructor expressions not found" +
+                " expected '(int|string)', found 'error'", 67, 16);
         Assert.assertEquals(negativeSemanticResult.getErrorCount(), i);
     }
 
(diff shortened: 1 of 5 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/expressions/errorconstructorexpr/error-constructor-expr-negative.bal
```diff
@@ -62,3 +62,6 @@ type ErrorU3 map<int>|map<Blah>|map<string>;
 error uc1 = error ErrorU1("Hello");
 error uc2 = error ErrorU2("Hello");
 error uc3 = error ErrorU3("Hello");
+
+int|string uc4 = error ErrorA("msg");
```
MaryamZi, line 66: For this particular case, I think the previous error may be more accurate. When we say `error E(...)` we are saying the type of the constructed error is `E` right? So I think we should continue to say `expected 'T' found 'E'` here. WDYT?

KRVPerera: Agree. Check the latest commit.
70b674a78d76dede5031e3fa2882de65f808a48d

MaryamZi: But does the `found 'error'` part make sense when we say `a type compatible with error constructor expressions not found`?

Can't we do
1. if the user provides a `error-type-reference`, use the "incompatible types: expected 'T', found 'E'" error
2. else, use "a type compatible with error constructor expressions not found in type 'T'"

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/errorconstructorexpr/ErrorConstructorExprTest.java
```diff
@@ -99,8 +99,13 @@ public void testErrorConstructorExprNegative() {
         validateError(negativeSemanticResult, i++, "cannot infer type of the error from '(error|ErrorB)'", 51, 20);
         validateError(negativeSemanticResult, i++, "unknown type 'Blah'", 60, 27);
         validateError(negativeSemanticResult, i++, "cannot create a new error value from 'ErrorU1'", 62, 19);
+        validateError(negativeSemanticResult, i++, "incompatible types: expected 'error', found 'ErrorU2'", 63, 13);
```
MaryamZi, line 102: This isn't correct, right? The RHS expression itself is erroneous, so we can't mention RHS is `ErrorU2`. IMO just the next error (cannot create a new error value from 'ErrorU2') is enough here. WDYT?

We can fix this separately too.

KRVPerera: Actually, one error is issued from here https://github.com/ballerina-platform/ballerina-lang/blob/8adfc0ac5b12d60c70d33893a15e49e7eba2920e/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java#L2799.

And we return from that point. So currently when we provide `typeRef` there will be duplicates.

MaryamZi (approved):

PR #33556, opened 2021-11-01
Fix never field type checking in record variable 

### LINKED ISSUE #33010: Type checking bug with optional record fields of type `never`
**Description:**
$title, at both compile-time and runtime.

**Steps to reproduce:**
```ballerina
import ballerina/io;

type Pair record {
    int x;
};

type PairRest record {
    never x?;
};

public function main() {
    Pair p = {
        x: 1,
        "y": 2,
        "color": "blue"
    };

    var {x: _, ...rest} = p;


    record { never x?; } a = rest; // OK

    record { never x?; never y?; } y = a; // no compilation error!
    io:println(y is record { never x?; never y?; }); // true!
    io:println(y.hasKey("y")); // true!

    record { never x?; never y?; } z = rest; // no compilation error!
    io:println(z is record { never x?; never y?; }); // true!
    io:println(z.hasKey("y")); // true!
}
```

**Affected Versions:**
slbeta3-rc3

HindujaB:
## Purpose
Currently compile time & runtime type checkers do not consider the never type field when assigning values to records. 
This PR fixes it
 
Fixes #33010 

## Approach
Added required validations to compile-time & runtime `TypeChecker`

## Samples
```ballerina
import ballerina/io;

public function main() {
    record {} a = {"x": 11};
    record {never x?;} y2 = a; // this should fail at compile time

    record {} r1 = {"x": 2, "color": "blue"};
    io:println(r1 is record {never x?;});  // should give false
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/types/never/never_type_runtime.bal
```diff
@@ -133,6 +133,20 @@ function testNeverWithAnyAndAnydataRuntime() {
     assertEquality(true, c4);
 }
 
+function testNeverFieldTypeCheck() {
```
MaryamZi, line 136: Let's also add some runtime tests that cover isLike checks.

Due to https://github.com/ballerina-platform/ballerina-lang/issues/33561 you may have to do something like

```ballerina
record { never i?; int j; } x = {j: 1};
anydata y = (<anydata> x).cloneReadOnly();
```

to get the immutable value at runtime.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -1386,6 +1391,22 @@ private static boolean checkIsRecordType(MapValue sourceRecordValue, BRecordType
         return true;
     }
 
+    private static boolean containsInvalidNeverField(String fieldName, Field sourceField, MapValue sourceRecordValue) {
+        if (sourceField != null) {
```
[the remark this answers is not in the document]
HindujaB: As the method logic is changed when addressing other suggestions, I think we don't need to inverse the if condition. Because the else part is more complex than the if part.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -1386,6 +1391,31 @@ private static boolean checkIsRecordType(MapValue sourceRecordValue, BRecordType
...
+
+    private static boolean containsNeverType(Type fieldType) {
+        int fieldTag = fieldType.getTag();
+        if (fieldTag == TypeTags.NEVER_TAG) {
+            return true;
+        }
+        if (fieldTag == TypeTags.UNION_TAG) {
+            for (Type member : ((BUnionType) fieldType).getOriginalMemberTypes()) {
+                if (member.getTag() == TypeTags.NEVER_TAG) {
+                    return true;
```
warunalakshitha, line 1412: What if happen it contain union of union with never type?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -1386,6 +1391,31 @@ private static boolean checkIsRecordType(MapValue sourceRecordValue, BRecordType
...
+        return !containsNeverType(sourceRecordType.getRestFieldType());
+    }
+
+    private static boolean containsNeverType(Type fieldType) {
+        int fieldTag = fieldType.getTag();
+        if (fieldTag == TypeTags.NEVER_TAG) {
+            return true;
+        }
+        if (fieldTag == TypeTags.UNION_TAG) {
+            for (Type member : ((BUnionType) fieldType).getOriginalMemberTypes()) {
```
warunalakshitha, line 1410: Shall we move this to seperate variable ((BUnionType) fieldType).getOriginalMemberTypes()

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
HindujaB: That will be still passed as a single union type with members. 
I added some unit tests for this case.

warunalakshitha (approved):
LGTM

PR #33663, opened 2021-11-05
Add JSON-Record improvement

tharindulak:
Fixes: https://github.com/wso2-enterprise/choreo/issues/8564


misc/ls-extensions/modules/json-to-record-converter/build.gradle
```diff
@@ -38,6 +38,7 @@ dependencies {
         exclude group: 'com.google.guava', module: 'guava'
     }
     implementation 'com.google.guava:guava:30.0-jre'
+    implementation group: 'com.google.code.gson', name: 'gson', version: '2.8.5'
```
tharindu-nw, line 41: We were previously using Jackson-Databind library for json related tasks in the module. Is it possible to replace the gson use cases with that library also? To avoid adding another dependency

misc/ls-extensions/modules/json-to-record-converter/src/main/java/io/ballerina/converters/JsonToRecordConverterService.java
```diff
@@ -45,13 +45,15 @@
...
                 String jsonString = request.getJsonString();
-                String codeBlock = JsonToRecordConverter.convert(jsonString);
-                response.setCodeBlock(codeBlock);
+                String recordName = request.getRecordName();
+                boolean isRecordTypeDesc = request.getIsRecordTypeDesc();
+                boolean isClosed = request.getIsClosed();
+                response = JsonToRecordConverter.convert(jsonString, recordName, isRecordTypeDesc, isClosed);
             } catch (IOException | JsonToRecordConverterException | FormatterException e) {
-                response.setCodeBlock("");
+                response = new JsonToRecordResponse();
```
tharindu-nw, line 56: Can we also set an empty string as the codeblock here; as the vscode plugin checks for an empty string in case of an error.

tharindulak: This is fixed

misc/ls-extensions/modules/json-to-record-converter/build.gradle
tharindulak: This is fixed

misc/ls-extensions/modules/json-to-record-converter/src/main/java/io/ballerina/converters/JsonToRecordConverterService.java
tharindulak: Resolved in https://github.com/ballerina-platform/ballerina-lang/pull/33663/commits/bcf590f93a9af08bae304c79d752b7a8a45e29a0

misc/ls-extensions/modules/json-to-record-converter/build.gradle
tharindulak: Resolved in https://github.com/ballerina-platform/ballerina-lang/pull/33663/commits/bcf590f93a9af08bae304c79d752b7a8a45e29a0

tharindu-nw (approved):

PR #33668, opened 2021-11-05
Fix "Type param analysis doesn't seem to be happening properly for infer-anonymous-function-expr"

### LINKED ISSUE #33601: Type param analysis doesn't seem to be happening properly for infer-anonymous-function-expr
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type Employee record {
    int id;
    string name;
};

public function main() {
    table<Employee> tb = table [];

    // // No longer allowed after fixing https://github.com/ballerina-platform/ballerina-lang/issues/33420.
    // var _ = tb.map(function(Employee r) returns int {
    //     return r.id;
    // });

    // Still no error.
    var _ = tb.map(r => r.id);
}
```

**Affected Versions:**
slbeta4-rc1, langlib-fixes

gimantha:
fixes #33601

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6425,6 +6425,7 @@ private void checkTypeParamExpr(Location pos, BLangExpression arg, SymbolEnv env
             BType expType = typeParamAnalyzer.getMatchingBoundType(expectedType, env);
             BType inferredType = checkExpr(arg, env, expType);
             typeParamAnalyzer.checkForTypeParamsInArg(pos, inferredType, this.env, expectedType);
+            types.checkType(arg.pos, inferredType, expectedType, DiagnosticErrorCode.INCOMPATIBLE_TYPES);
```
MaryamZi, line 6428: The underlying issue (why this isn't caught in L6426) IINM is that we use `symTable.noType` for the return type, right? 

I'm not quite sure if using `symTable.noType` is correct there in the first place. For example, this also results in the following printing true, but I'm not sure if that's correct or if we should use `map<any|error>` as the contextually-expected type for the return type.

```ballerina
import ballerina/io;

type Employee record {
    int id;
    string name;
};

public function main() {
    table<Employee> tb = table [];
    
    var a = tb.map(r => {x: r.id});
    io:println(a is table<record {| int x; |}>); // true
}
```

gimantha: Should we not derive the return type from the typeparams? the arrow function's return type is `record {| int x; |}>` So the map should return `table<record {| int x; |}>` right?

MaryamZi: The infer-anonymous-function-expr's types are derived from the contextually-expected type, right?

https://ballerina.io/ballerina-spec/spec.html#infer-anonymous-function-expr

> The static type of the infer-anonymous-function-expr will be a function type whose return type is the static type of the expression in expr-function-body. If the contextually expected type for the anonymous-function-expr is a function type with return type T, then the contextually expected type for expression in expr-function-body is T.

My concern is if we consider this to be a scenario where there is no contextually-expected type (i.e., effectively using symTable.noType) or if we consider `function(MapType val) returns MapType1` as the contextually-expected type, where `MapType1` is considered as `map<any|error>`.

Depending on what we use, the inherent type used for `record {| int x; |}` will differ.

gimantha: Had a chat with @hasithaa  and @rdhananjaya It should print `true` for 

`io:println(a is table<record {| int x; |}>);`

MaryamZi (approved):

hasithaa (approved):

PR #33698, opened 2021-11-09
Fix `int:abs()` to panic for overflow values

### LINKED ISSUE #32590: Incorrect behavior of `lang.int:abs` function
**Description:**
$title

**Steps to reproduce:**
```ballerina
import ballerina/io;
public function main() {
    int a1 = -9223372036854775807 - 1;
    int b1 = a1.abs(); // should panic

    io:println(b1); 
}
```
```ballerina
import ballerina/io;
public function main() {
    int a1 = (-9223372036854775807 - 1).abs(); // should panic
    io:println(a1);
}
```
These print `-9223372036854775808`, but should panic at runtime since `9223372036854775808` would overflow

**Affected Versions:**
slbeta2

HindujaB:
## Purpose
$subject

Fixes #32590

## Approach
Added a validation for `int:abs()` implementation. 

## Samples
```ballerina
public function main() {
    int a1 = -9223372036854775808;
    int b1 = a1.abs(); // should panic

    int a2 = (-9223372036854775807 - 1).abs(); // should panic
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/langlib/lang.int/src/main/java/org/ballerinalang/langlib/integer/Abs.java
+++ b/langlib/lang.int/src/main/java/org/ballerinalang/langlib/integer/Abs.java
@@ -18,6 +18,14 @@
 
 package org.ballerinalang.langlib.integer;
 
+import io.ballerina.runtime.api.creators.ErrorCreator;
+import io.ballerina.runtime.internal.util.exceptions.BLangExceptionHelper;
+import io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons;
+import io.ballerina.runtime.internal.util.exceptions.RuntimeErrors;
+
+import static io.ballerina.runtime.api.constants.RuntimeConstants.INT_LANG_LIB;
+import static io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons.getModulePrefixedReason;
+
 /**
  * Native implementation of lang.int:abs(int).
  *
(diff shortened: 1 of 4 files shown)


hasithaa (approved):
Looks good to me

langlib/lang.int/src/main/java/org/ballerinalang/langlib/integer/Abs.java
```diff
@@ -32,6 +40,11 @@
 public class Abs {
 
     public static long abs(long n) {
+        if (n <= Long.MIN_VALUE) {
+            throw ErrorCreator.createError(getModulePrefixedReason(INT_LANG_LIB,
```
warunalakshitha, line 44: I think we can move this to a constant?

HindujaB: But, if we keep it in a constant, the error will be initialised even if there's no error. Wouldn't that be unnecessary?

warunalakshitha (approved):
LGTM

PR #33714, opened 2021-11-09
Add withTests and targetDir to BuildManifest

### LINKED ISSUE #33700: Adding targetDir and withTests as a build option is not supported
**Description:**
Specifying the path of the targetDir in the build options does not work as expected

**Steps to reproduce:**

**Affected Versions:**

Ibaqu:
Fixes #33700

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/ManifestBuilder.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/ManifestBuilder.java
@@ -384,8 +384,8 @@ public class ManifestBuilder {
 
         BuildOptions.BuildOptionsBuilder buildOptionsBuilder = BuildOptions.builder();
 
-        Boolean skipTests =
-                getBooleanFromBuildOptionsTableNode(tableNode, BuildOptions.OptionName.SKIP_TESTS.toString());
+        Boolean withTests = toBooleanFalseIfNull(
+                getBooleanFromBuildOptionsTableNode(tableNode, BuildOptions.OptionName.WITH_TESTS.toString()));
         Boolean offline = getBooleanFromBuildOptionsTableNode(tableNode, CompilerOptionName.OFFLINE.toString());
         Boolean experimental =
                 getBooleanFromBuildOptionsTableNode(tableNode, CompilerOptionName.EXPERIMENTAL.toString());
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/ManifestBuilder.java
```diff
@@ -384,8 +384,8 @@ private BuildOptions setBuildOptions(TomlTableNode tomlTableNode) {
 
         BuildOptions.BuildOptionsBuilder buildOptionsBuilder = BuildOptions.builder();
 
-        Boolean skipTests =
-                getBooleanFromBuildOptionsTableNode(tableNode, BuildOptions.OptionName.SKIP_TESTS.toString());
+        Boolean withTests = toBooleanFalseIfNull(
```
azinneera, line 387: Why do we need to introduce a new method here? Handling the null is done in the BuildOption getters. I think we can use `getBooleanFromBuildOptionsTableNode`

Ibaqu: I made this change because we still make use of the old setSkipTests function. 
https://github.com/ballerina-platform/ballerina-lang/pull/33714/files#diff-7e5ce460862aa48a7763a0055e868a61f7ba3af7745dccbff6a2ced886f04440R414

This is not an idea way of inverting the read boolean and passing it to setSkipTests(). What we are basically doing is, if `--with-tests` flag is used , or the build option is set, we are implying that `skipTests` is 'false'. Which is why null doesnt work in this case.

Ibaqu: I thik the better way of doing this would be to completely remove all instances of `skipTests` and replace it with `withTests` instead of juggling this boolean around

azinneera:
Please create an issue to remove `skipTests` from BuildOptions and use `withTests` throughout.

azinneera (approved):

Ibaqu:
Closing since we will be removing --with-tests entirely. 
The targetDir has been added to the following PR : https://github.com/ballerina-platform/ballerina-lang/pull/33743

Issue #33719, opened 2021-11-09
If distribution version does not match, should make `sticky = false` even `build.json` has not expired


pramodya1994:
**Description:**
$Title

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #33723, opened 2021-11-09
Make `sticky = false` when distribution version does not match

pramodya1994:
- Fixes https://github.com/ballerina-platform/ballerina-lang/issues/33719

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -160,6 +161,15 @@ private boolean getSticky(PackageContext rootPackageContext) {
             this.autoUpdate = false;
             return true;
         }
+
+        // if distribution is not same, we anyway return sticky as false
+        String distVersion = rootPackageContext.packageManifest().ballerinaVersion();
```
azinneera, line 166: This will end up discarding the build file for every run when the dist version in Ballerina.toml is an old one. I think we need to persist the dist version to the `target/build` file. Not having the dist version should only be a one-time thing. We should persist the dist version to the build file when updating it.

azinneera: @hevayo wdyt??

hevayo: Yes it should be in build file

pramodya1994: Added

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -160,14 +161,18 @@ private boolean getSticky(PackageContext rootPackageContext) {
...
         if (rootPackageContext.project().kind() == ProjectKind.BUILD_PROJECT) {
             Path buildFilePath = this.rootPackageContext.project().sourceRoot().resolve(TARGET_DIR_NAME)
                     .resolve(BUILD_FILE);
             if (Files.exists(buildFilePath) && buildFilePath.toFile().length() > 0) {
                 try {
                     BuildJson buildJson = readBuildJson(buildFilePath);
-                    if (buildJson != null && !buildJson.isExpiredLastUpdateTime()) {
+                    // if distribution is not same, we anyway return sticky as false
+                    if (buildJson != null &&
+                            buildJson.distributionVersion().equals(RepoUtils.getBallerinaShortVersion()) &&
```
azinneera, line 174: dist version can be null if `target/build` is generated with `beta3`. Is this handled?

project-api/project-api-test/src/test/resources/old_dist_version_project/Ballerina.toml
```diff
@@ -0,0 +1,5 @@
+[package]
+org = "sameera"
+name = "myproject"
+version = "0.1.0"
+distribution = "slbeta1"
```
azinneera, line 5: We are adding the distribution version from `beta4`. Shall we update this test case to not have the distribution version? This is related to https://github.com/ballerina-platform/ballerina-lang/pull/33723/files#r746322422

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
pramodya1994: Added null check

azinneera: Shall we add a test case for this?

azinneera (approved):

PR #33752, opened 2021-11-10
Introduce new methods to `NodeParser`

### LINKED ISSUE #32901: Implement methods to generate nodes via NodeParser
**Description:**
`parseExpression()`, `parseStatements()` is implemented with PR #32806 as per shell, debugger requirements.

There is a use-case(bindgen, open api, grpc, etc) if we add `parseStatement()`, `parseModuleMemberDeclaration()`, `parseModuleMemberDeclarations()`, `parseTypeDescriptor()` etc.

For example, say one wants to generate a statement with a specific name; `int foo = 10;`
`NodeParser.parseStatement(String.format("int %s = 10;", "foo"))`

will give the generated node without having to build it from scratch using NodeFactory API. 
Also, any changes to ST parsing are instantly reflected through API.

### LINKED ISSUE #33287: Ballerina Shell requirements related to NodeParser
**Description:**

Consider following two cases.

1.  `"abc->method(arg1)"`;

if we parse this as an expression, expected node is `RemoteMethodCallActionNode`.

In the current shell implementation, similar cases handled by parsing  as a statement as shown below.

`node = ((AssignmentStatementNode) NodeParser.parseStatements("a = " + source ) .get(0)).expression();`

2. "`@strand { thread:"any" } worker w1 { }`"

if we parse this as a statement, expected node is `NamedWorkerDeclarationNode`.

Need a method to identify above two nodes.

### LINKED ISSUE #33593: Remove the deprecated and marked for removal syntax APIs for beta5/GA
**Description:**
$subject.

https://github.com/ballerina-platform/ballerina-lang/blob/8a99c1aa8c0e6379ca404760ae7453adf7c58441/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/SyntaxTree.java#L68-L94

lochana-chathura:
## Purpose
$subject.

Following are introduced:

- parseStatement()
- parseActionOrExpression()
- parseModuleMemberDeclaration()
- parseImportDeclaration()
- parseTypeDescriptor()
- parseBindingPattern()
- parseFunctionBodyBlock()

Fixes #32901 Fixes #33287

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -100,6 +100,23 @@ public STNode parse() {
...
+     *
+     * @return Parsed node
+     */
+    public STNode parseAsStatement() {
+        STNodeList stmtNodeList = (STNodeList) parseAsStatements();
+        STNode stmt = stmtNodeList.get(0);
+
+        for (int i = 1; i < stmtNodeList.size(); i++) {
+            stmt = SyntaxErrors.cloneWithTrailingInvalidNodeMinutiae(stmt, stmtNodeList.get(i),
+            DiagnosticErrorCode.ERROR_SYNTAX_ERROR);
```
lochana-chathura, line 114: 1. Here used `parseAsStatements()` and then invalidate the extra statements. The reason for using it instead of `parseStatement()` is, there are validations we do on top of `parseStatement()` and we may not repeat them in another place?

2. Notice that here we have not given a specific error code, used the general error. Since this error is specific to this case and this does not go on compiler phases and log errors no point in declaring a new error? 

Although we declared a new one, we better not mix it with other errors. As we eventually give a doc or something to the users for diagnostic id vs description. It is odd to have NodeParser specific error mixed up among regular parser errors.

lochana-chathura: Refactored to use `parseStatement()` inside with b45e13b

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/NodeParser.java
```diff
@@ -21,21 +21,32 @@
...
+    }
+
+    /**
+     * Parses the input as a binding pattern.
+     *
+     * @param text the input
+     * @return a {@code BindingPatternNode}
+     */
+    public static BindingPatternNode parseBindingPattern(String text) {
+        BallerinaParser parser = ParserFactory.getParser(text);
```
rdulmina, line 48: Can we keep one global instance of `BallerinaParser` for `NodeParser` class? We need to empty the context stack after each parse if we are doing so

rdulmina (approved):

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/NodeParser.java
lochana-chathura: Resolving as we discussed offline.

sameerajayasoma (approved):

Issue #33162, opened 2021-10-11
Field that contains a default value in a record is not initialized with default value when there is readonly specific field in mapping cons expr


rdulmina:
**Description:**
Consider the bellow sample 
```bal
function init() {
    record {int fieldA; string fieldB = "default";} b = {readonly fieldA: 2}; //
    printResult(b.fieldB); // prints null
}
```
The above program prints null where it should print `default`. It works when we remove readonly from the mapping constructor

Issue #32012, opened 2021-08-04
Overriding a field included with record-type-inclusion doesn't remove the default value closure for that particular field being overriden


rdhananjaya:
**Description:**
```ballerina
type R0 record {
    int|string f = 1;
};

type R1 record {
    *R0;
    string f;
};

public function main() {
    R1 r1 = { f : "hello"};
}
````

This results in:
```
error: {ballerina/lang.map}InherentTypeViolation {"message":"invalid value for record field 'f': expected value of type 'string', found 'int'"}
        at $value$R0:R0$gen$.<init>(main.bal:2)
           main:main(main.bal:11)
```

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

**Comment by rdhananjaya** [comment]
The culprit for this issue is when we initialize R1, it automatically calling the init of referenced types.
```ballerina
import ballerina/io;
type R record {|
    int hello = inc();
|};

type RR record {|
    *R;
    int hello = inc();
|};

public function main() {
    R|RR rr = <RR>{};
    lock {
        io:println(i); // prints 2 [inc method have invoked twise]
    }
}

isolated int i = 0;

isolated function inc() returns int {
    lock {
        i = i + 1;
        return i;
    }
}
```

MaryamZi:
We desugar the initialization logic of all the fields into the generated init method of the record. This initializes all of the fields in the records for which default values have been specified. 

With type inclusion, I believe including records call the init methods of the included records to initialize fields that are initialized in the included records. But  this'll result in errors for examples like in the issue. Just removing the init call of the included record isn't an option because for some fields the default value will still come from the included record.

```ballerina
type R0 record {
    int|string f = 1;
    int? g = 1;
};

type R1 record {
    *R0;
    string f; 
    // g is initialized in R0
};
```

To properly support inclusion like in the example IMO we need to be able to selectively call the initialization logic for specific fields. We may need to introduce the concept of default value closures attached to each field.

warunalakshitha:
ATM calling the included record types in it calls generated during codegen. We can set a skipped field set during included record initialization but might cause more issues if we try to do that at runtime. Perfect solution would be to have default value closure for each field as @MaryamZi  said. So we can independently initialize each record.

Issue #24077, opened 2020-06-13
Intersection type validation should look at the default value specified for a field


MaryamZi:
**Description:**
$title.

See the suggested rule at https://github.com/ballerina-platform/ballerina-spec/issues/544#issuecomment-643069011.

**Steps to reproduce:**
The following should result in a compilation error since the default value of `details` is of type `Details` and is mutable (i.e., not `Details & readonly`).

```ballerina
type Details record {
    string name;
};

Details det = {name: "Maryam"};

type Employee record {
    Details details = det;
};

public function main() {
    Employee & readonly emp = {}; // `Employee & readonly` is not a valid intersection type
}
```

This currently fails at runtime when creating the value, but should actually fail at compile-time.

```cmd
$ ballerina run xyz.bal 
Compiling source
        xyz.bal
Running executables

error: {ballerina/lang.map}InherentTypeViolation message=invalid value for record field 'details': expected value of type '(Details & readonly)', found 'Details'
        at xyz:main(xyz.bal:12)
```

MaryamZi:
This is currently not possible since we don't have any details about the default value in the type/symbol. We need at least the type of the default value.

MaryamZi:
On hold until https://github.com/ballerina-platform/ballerina-spec/issues/544#issuecomment-655671466 is decided upon.

hasithaa:
This issue has been marked as "Deferred" and will be revisited in the future when resources are available. We will attend this once the spec issue is resolved.

gimantha:
Now we are giving the following error for the default value
`ERROR [main.bal:(8:23,8:26)] invalid access of mutable storage in the default value of a record field`
![Uploading image.png…]()

MaryamZi:
I guess the code should now be 

```ballerina
isolated function getDetails() returns Details => {name: "Maryam"};

type Employee record {
    Details details = getDetails();
};
```

We now use `cloneReadOnly` with the default value to make this work.

PR #33767, opened 2021-11-11
Fix defaults from the applicable contextually-expected type not being used when a mapping constructor has `readonly` fields

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/33162

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -60,6 +60,7 @@ import org.wso2.ballerinalang.compiler.semantics.model.symbols.BOperatorSymbol;
 import org.wso2.ballerinalang.compiler.semantics.model.symbols.BPackageSymbol;
 import org.wso2.ballerinalang.compiler.semantics.model.symbols.BRecordTypeSymbol;
 import org.wso2.ballerinalang.compiler.semantics.model.symbols.BSymbol;
+import org.wso2.ballerinalang.compiler.semantics.model.symbols.BTypeSymbol;
 import org.wso2.ballerinalang.compiler.semantics.model.symbols.BVarSymbol;
 import org.wso2.ballerinalang.compiler.semantics.model.symbols.BXMLNSSymbol;
 import org.wso2.ballerinalang.compiler.semantics.model.symbols.SymTag;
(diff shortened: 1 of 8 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1809,12 +1811,24 @@ private BType getEffectiveMappingType(BLangRecordLiteral recordLiteral, BType ap
...
         BLangRecordTypeNode recordTypeNode = TypeDefBuilderHelper.createRecordTypeNode(recordType, pkgID, symTable,
-                                                                                       recordLiteral.pos);
+                                                                                       pos);
         recordTypeNode.initFunction = TypeDefBuilderHelper.createInitFunctionForRecordType(recordTypeNode, env,
                                                                                            names, symTable);
         TypeDefBuilderHelper.addTypeDefinition(recordType, recordSymbol, recordTypeNode, env);
 
-        if (applicableMappingType.tag == TypeTags.MAP) {
+        if (applicableMappingType.tag == TypeTags.RECORD) {
+            BRecordType applicableRecordType = (BRecordType) applicableMappingType;
```
MaryamZi, line 1820: Since it is not possible to call record field default closures individually atm, the fix done for the time-being is adding the original type as an included type to make its init method (that initiates the fields) get called.

However this will cause issues for something like 

```ballerina
type Foo record {
    string[] x = ["foo", "bar"];
};

// Panics now with `invalid value for record field 'x': expected value of type 'string[] & readonly', found 'string[]'`,
// since `x` is initialized using the mutable value from `Foo` before the mapping constructor's value.
Foo _ = {readonly x: [""]};
```

This does not panic in Beta3/Beta4 but will panic with this fix.

Related to https://github.com/ballerina-platform/ballerina-lang/issues/24077 in a way.

```ballerina
type Foo record {
    int[] x = [];
};

Foo & readonly _ = {};
```

Also panics at runtime atm.

But if we don't fix this, defaults won't be used in cases like

```ballerina
import ballerina/io;

type Foo record {
    int i;
    string s = "default";
};

public function main() {
    Foo f = {readonly i: 1};
    io:println(f); // `{"i":1,"s":null}` and not `{"i":1,"s":"default"}`
}
```

@hasithaa, thoughts please?

hasithaa: +1 for now. will @chiranSachintha fix for closures address this? I guess not.

hasithaa (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: We need to introduce similar closures for record default values, right? Created https://github.com/ballerina-platform/ballerina-lang/issues/33771 to track fixing it. https://github.com/ballerina-platform/ballerina-lang/issues/32012 also depends on it.

Created https://github.com/ballerina-platform/ballerina-lang/issues/33773 to track fixing this specific case.

Issue #33297, opened 2021-10-19
Deriving the type for a table constructor doesn't correctly handle members having fields of the same name but of different types


MaryamZi:
**Description:**
The [spec](https://ballerina.io/ballerina-spec/spec.html#table-constructor-expr) says
> If there is no contextually expected type, then the member type of the inherent type is derived from the the static type of the expressions for the members: the member type will be the smallest record type that is a supertype of the static types of all the expressions in the row-list.

**Steps to reproduce:**
Consider

```ballerina
public function main() {
    var v1 = table [
        {a: 1},
        {a: "str", b: 2}
    ];
}
```

Shouldn't the type be inferred as `table<record {|int|string a; int b?;|}>`?

Currently results in a compilation error.

```cmd
cannot infer the member type from table constructor. field 'a' type is ambiguous
```

**Affected Versions:**
slbeta3

**Related Issues (optional):**
https://github.com/ballerina-platform/ballerina-lang/issues/33296

lochana-chathura:
> Same for
> 
> ```ballerina
> function init() {
>     record {string a; int b?;} m = {a: "str", b: 2};
>     string s1 = "a";
>     string s2 = "b";
> 
>     var v1 = table [
>         {a: 1},
>         {...m},
>         {[s1]: true, c: 2, [s2]: false}
>     ];
>     table<record {int|string|boolean a?; int|string b?; anydata c?;}> t1 = v1;
> }
> ```

@MaryamZi just to be clear, 
here for v1, inferred type should be `table<record {| (int|string) a?; int b?;  int c?; boolean...; |}>` right?

MaryamZi:
@lochana-chathura it should be

```ballerina
table<record {int|string|boolean a?; int|boolean b?; anydata c?;}>
```

right?

- `a` and `b` can be `boolean` because of the `boolean` fields with computed keys in the third member
- `c` can be `anydata` because `m` is open by `anydata`
- rest descriptor has to be anydata (or the record has to be an inclusive one) also because `m` is open

The original type I used in https://github.com/ballerina-platform/ballerina-lang/issues/33297#issuecomment-946764504 is wrong.

lochana-chathura:
@MaryamZi Sorry I missed the open record part.
If we consider inferred record types for each `mapping-constructor-expr`, that would be,

1. `record {| int a; |}`
2. `record {| string a; int b?; anydata...; |}`
3. `record {| int c; boolean...; |}` 

[...]

PR #33774, opened 2021-11-11
Fix ignoring rest field in record literals while inferring table constraint type

### LINKED ISSUE #33295: Invalid type derived for table constructor with mapping constructor members that have computed keys and spread fields
**Description:**
$title, when there's no contextually-expected type.

**Steps to reproduce:**
```ballerina
public function main() {
    string s1 = "id";
    string s2 = "employed";

    var v1 = table [
        {name: "Jo"},
        {[s1]: 2}, // crashes at runtime: {ballerina/lang.map}KeyNotFound {"message":"invalid field access: field 'id' not found in record type 'record {| string name?; |}'"}
        {[s2]: false}
    ];
    // int i = v1; // incompatible types: expected 'int', found 'table<record {| string name?; |}>'

    table<record {|string name?;|}> t1 = v1; // no compilation error!
}
```

**Related Issues:**
https://github.com/ballerina-platform/ballerina-lang/issues/32707

**Affected Versions:**
slbeta3

pcnfernando:
Fixes #33295

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/types/table/record-constraint-table-value.bal
```diff
@@ -492,6 +493,19 @@ public function testSpreadFieldInConstructor() {
...
+function testTableConstructExprVar() {
+    string s1 = "id";
+    string s2 = "employed";
+
+    var v1 = table [
+        {name: "Jo"},
+        {[s1]: 2},
+        {[s2]: false}
+    ];
+
```
gimantha, line 505: Tried the following code, seems it gives a runtime error

```ballerina
map<int> m = {name: 1, b: 2};
    var v2 = table [
        {name: "Jo"},
        {...m}
    ];
```

Should the table's constraint type be `record{|string|int name; ...int|} `here?

gimantha: We have a separate issue for that one. Can fix separately
https://github.com/ballerina-platform/ballerina-lang/issues/33297

gimantha (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1112,8 +1122,15 @@ private BRecordType createTableConstraintRecordType(Set<BField> allFieldSet, Loc
...
-        recordType.sealed = true;
-        recordType.restFieldType = symTable.noType;
+
+        if (restFieldTypes.isEmpty()) {
+            recordType.sealed = true;
+            recordType.restFieldType = symTable.noType;
+        } else if (restFieldTypes.size() == 1) {
+            recordType.restFieldType = restFieldTypes.get(0);
+        } else {
+            recordType.restFieldType = BUnionType.create(null, restFieldTypes.toArray(new BType[0]));
```
lochana-chathura, line 1132: Here we convert the list to an array then again convert back to a list. HBT?
```suggestion
            recordType.restFieldType = BUnionType.create(null, new LinkedHashSet<>(restFieldTypes));
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1044,8 +1044,17 @@ private BType inferTableMemberType(List<BType> memTypes, BLangTableConstructorEx
...
         Set<BField> allFieldSet = new LinkedHashSet<>();
+        List<BType> restFieldTypes = new ArrayList<>();
+
         for (BType memType : memTypes) {
-            allFieldSet.addAll(((BRecordType) memType).fields.values());
+            BRecordType member = (BRecordType) memType;
+            allFieldSet.addAll(member.fields.values());
+            if (!member.sealed) {
+                BType restFieldType = member.restFieldType;
+                if (isUniqueType(restFieldTypes, restFieldType)) {
```
lochana-chathura, line 1054: Just for my understanding, 
is there a difference if we use a `LinkedHashSet<BType>` instead of `isUniqueType()` check here?

pcnfernando: If we use a hashset, two instances of say `new BType(TypeTags.INT)` will not be considered as duplicates.  isUniqueType() checks through the `BSameTypeVisitor`

lochana-chathura: I see. Thanks for the clarification. :)

Say we have types, `(int|string)` and `int`, where two `int`s are from two `new BType(TypeTags.INT)` instances.
Now we do the `isUniqueType()` check here and conclude they are unique. 

Then we create the union with `BUnionType.create(null, types);`. Inside the `BUnionType.create()` we seem to be using `LinkedHashSet<BType>` and no `isUniqueType()` check for the flatten members. So the resulant union will be `(int|string|int)`.

Just wondering whether we should do the following here?
``` 
            if (restFieldType.tag == TypeTags.UNION) {
                ((BUnionType) restFieldType).getMemberTypes().forEach(type -> {
                    if (isUniqueType(restFieldTypes, type)) {
                        types.add(type);
                    }
                });
            } else if (isUniqueType(restFieldTypes, restFieldType)) {
                types.add(restFieldType);
            }

```

pcnfernando: `new BType(TypeTags.INT)` was provided as an example to say the object references would be different; anyway, the int would come from the symbol table; where the references would be anyway the same. let's have a offline chat regarding the exact requirement

gimantha (approved):

PR #33797, opened 2021-11-12
Fix completion sorting in remote method calls and checking expressions

### LINKED ISSUE #33624: Auto-completions spec conformance for the `CheckingExpression`
**Description:**
$title
https://ballerina.io/spec/lang/master/#checking-expr

### LINKED ISSUE #33625: Auto-completions spec conformance for the `RemoteMethodCallAction`
**Description:**
$title

### LINKED ISSUE #33672: Code completion for remote function can be sorted based on LHS of the expression
**Description:**
Screenshot 2021-11-05 at 21 02 16

Here sql:ProcedureCallResult[] is only returned for call method. This can be highlighted.

IMS94:
## Purpose
$subject

Fixes #33625
Fixes #33672
Fixes #33624

## Approach
Updated `ContextTypeResolver` and relevant completion providers to adjust the completion sorting.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionUtil.java
@@ -846,6 +846,7 @@ public class CodeActionUtil {
      */
     public static boolean hasErrorMemberType(UnionTypeSymbol unionTypeSymbol) {
         return unionTypeSymbol.memberTypeDescriptors().stream()
+                .map(CommonUtil::getRawType)
                 .anyMatch(member -> member.typeKind() == TypeDescKind.ERROR);
     }
 
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
@@ -759,7 +759,9 @@ public abstract class AbstractCompletionProvider<T extends Node> implements Ball
         Optional<TypeSymbol> contextType = context.getContextType();
         String sortText = "";
         // First we sort the assignable items above others, and then sort by the rank
-        if (contextType.isPresent() && SortingUtil.isCompletionItemAssignable(completionItem, contextType.get())) {
+        if (contextType.isPresent() &&
+                (SortingUtil.isCompletionItemAssignable(completionItem, contextType.get()) ||
+                        SortingUtil.isCompletionItemAssignableWithCheck(completionItem, contextType.get()))) {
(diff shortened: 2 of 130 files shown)


language-server/modules/langserver-core/src/test/java/org/ballerinalang/langserver/completion/ActionNodeContextTest.java
```diff
@@ -35,6 +35,10 @@ public void test(String config, String configPath) throws IOException, Workspace
     @Override
     public Object[][] dataProvider() {
         return this.getConfigsList();
+//        return new Object[][]{
```
malinthar, line 38: Shall we remove these lines

malinthar (changes requested):
Changes look good!

language-server/modules/langserver-core/src/test/java/org/ballerinalang/langserver/completion/ActionNodeContextTest.java
IMS94: Done. It was a mistake.

IMS94:
@nadeeshaan @malinthar @mohanvive With this PR, I have updated the assignability check as follows:
1. Items which are directly assignable to the context are sorted first
2. Items which are assignable after adding a `check` expression are sorted 2nd - This is similar to considering the `edit distance`. When we write code, we usually do `int varName = clientObj->methodCall()` and then add the `check` in between. That's why I thought of using this idea. Since remote method calls usually comes with an error union, this should be very useful.
3. Other items are sorted 3rd

WDYT?

nadeeshaan:
> @nadeeshaan @malinthar @mohanvive With this PR, I have updated the assignability check as follows:
> 
> 1. Items which are directly assignable to the context are sorted first
> 2. Items which are assignable after adding a `check` expression are sorted 2nd - This is similar to considering the `edit distance`. When we write code, we usually do `int varName = clientObj->methodCall()` and then add the `check` in between. That's why I thought of using this idea. Since remote method calls usually comes with an error union, this should be very useful.
> 3. Other items are sorted 3rd
> 
> WDYT?

This is very good thinking @IMS94.  We do sort within the items related check as well right?

IMS94:
> This is very good thinking @IMS94. We do sort within the items related check as well right?
@nadeeshaan yes, as of now we do default sorting within each group (mentioned above). Depending on the context, we can customize that as well.

nadeeshaan (approved):
LGTM

mohanvive (approved):

mohanvive:
> Changes look good!

@malinthar Need your approval to merge the PR.

malinthar (approved):
LGTM!

PR #33815, opened 2021-11-15
Fix `never` type rest field type checking

### LINKED ISSUE #33785: A record typedesc with a `never` rest descriptor should be considered the same as the record being a closed record
**Description:**
$title. Needs to be fixed at both compile time and runtime.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    record {| never...; |} a = {};

    // record {||} _ = a; // error: incompatible types: expected 'record {| |} & readonly', found 'record {| never...; |} & readonly'

    function () returns record {| never...; |} b = () => {};
    io:println(<any> b is function () returns record {| |}); // false
}
```

Should work both ways.

**Affected Versions:**
slbeta3, slbeta4-rc2

HindujaB:
## Purpose
$subject
Fixes #33785 

## Approach
Added the check for `never` type rest field in the compile-time & runtime type-checker.

## Samples
```ballerina
import ballerina/io;

public function main() {
    record {| never...; |} a = {};
    record {||} _ = a; 

    function () returns record {| never...; |} b = () => {};
    io:println(<any> b is function () returns record {| |});
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


warunalakshitha:
Any idea why patch coverage is less?

HindujaB:
> Any idea why patch coverage is less?

This happened in some other PRs as well. Not sure about the reason, since the added tests cover the fix and it didn't show any warnings as well. 
https://github.com/ballerina-platform/ballerina-lang/runs/4237450974

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1639,8 +1639,10 @@ private int getObjectFuncCount(BObjectTypeSymbol sym) {
     }
 
     public boolean checkRecordEquivalency(BRecordType rhsType, BRecordType lhsType, Set<TypePair> unresolvedTypes) {
-        // If the LHS record is closed and the RHS record is open, the records aren't equivalent
-        if (lhsType.sealed && !rhsType.sealed) {
+        // If the LHS record is closed and the RHS record is open and the rest field type of RHS is not a 'never'
+        // type, the records aren't equivalent
+        if (lhsType.sealed && !rhsType.sealed &&
+                (rhsType.restFieldType == null || rhsType.restFieldType.tag != TypeTags.NEVER)) {
```
MaryamZi, line 1645: IINM, rest field type is never null. It's usually noType. I think we can get rid of the null check.

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never_type_runtime.bal
```diff
@@ -245,6 +245,50 @@ function baz2() returns never[] {
     return [];
 }
 
+function testNeverRestFieldType() {
```
MaryamZi, line 248: Shall we add more tests with inclusive record type-descriptors and exclusive record type-descriptors with different non-never types?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/never/NeverTypeTest.java
```diff
@@ -326,7 +326,8 @@ public void testNeverRuntime(String functionName) {
                 "testNeverRuntime11",
                 "testNeverRuntime12",
                 "testNeverWithAnyAndAnydataRuntime",
-                "testNeverFieldTypeCheck"
+                "testNeverFieldTypeCheck",
```
MaryamZi, line 329: Also need to add negative compilation tests.

tests/jballerina-unit-test/src/test/resources/test-src/types/never/never_type_runtime.bal
HindujaB: This fix is specific to rest field types. Since, we can't have rest field descriptors insideinclusive record type-descriptors, IMO adding tests for them seems redundant. 
I added tests for exclusive record type-descriptors in latest PR.

warunalakshitha (approved):

Issue #33827, opened 2021-11-16
Blocked scheduler items intermittently not getting reschedule


warunalakshitha:
**Description:**
In the Jballerina runtime scheduler, we have scheduler groups and group items. The scheduler group contains a stack of scheduler items.

As an example, if you have 3 workers (except strand annotation with any) those will be added to the same scheduler group.

Each scheduler group is added to the runnable queue. Each os thread in scheduler there pool will assign a group( removed from runnable queue) and execute items.

Once execute each item can be done or blocked on another event like i/o operation, sleep. Then we will execute the next item in the group. If all items are done or blocked we will mark the group "scheduled" flag to be false. Then os thread will assign another group.

Same time blocked items can be done and once done the i/o thread will reschedule the given item. First, it will check the items' group "scheduled" flag and if it is still true means the given group is still running another thread. So we just added to the group items stack so the currently running scheduler thread will execute the item continue the next instructions.

If the group is already done ("scheduled flag false) i/o thread will add that group to the runnable queue.

The group scheduled status change is not an atomic operation. Hence there is a possibility i/o thread might think it is already an executable state and same time scheduler thread change the status to false, so that item gets lost without rescheduling.

So the item will be not done. This will make either program stuck or service requests time out.

PR #33832, opened 2021-11-16
Fix blocked scheduler items not getting reschedule

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/33827


chamil321:
Related HTTP issue https://github.com/ballerina-platform/ballerina-standard-library/issues/2035

sameerajayasoma:
@warunalakshitha Can you please explain the fix? :)

warunalakshitha:
In the Jballerina runtime scheduler, we have scheduler groups and group items. The scheduler group contains a stack of scheduler items. 

As an example, if you have 3 workers (except strand annotation with any) those will be added to the same scheduler group.

Each scheduler group is added to the runnable queue. Each os thread in scheduler there pool will assign a group( removed from runnable queue) and execute items. 

Once execute each item can be done or blocked on another event like i/o operation, sleep. Then we will execute the next item in the group. If all items are done or blocked we will mark the group "scheduled" flag to be false. Then os thread will assign another group.

Same time blocked items can be done and once done the i/o thread will reschedule the given item. First, it will check the items' group "scheduled" flag and if it is still true means the given group is still running another thread. So we just added to the group items stack so the currently running scheduler thread will execute the item continue the next instructions.

If the group is already done ("scheduled flag false) i/o thread will add that group to the runnable queue.

The group scheduled status change is not an atomic operation. Hence there is a possibility i/o thread might think it is already an executable state and same time scheduler thread change the status to false, so that item gets lost without rescheduling. 

So the item will be not done. This will make either program stuck or service requests time out.

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/scheduling/Scheduler.java
```diff
@@ -509,13 +516,15 @@ public String toString() {
...
      */
-    Stack<SchedulerItem> items = new Stack();
+    Stack<SchedulerItem> items = new Stack<>();
 
     /**
      * Indicates this item is already in runnable list/executing or not.
      */
     AtomicBoolean scheduled = new AtomicBoolean(false);
 
+    private final ReentrantLock groupLock = new ReentrantLock();
```
manuranga, line 526: Would it be more efficient to keep the a lock in the `ItemGroup`? So there is a minimal chance of locks waiting.

manuranga:
@sameerajayasoma just FYI, from what Waruna explained, in the main branch this is already caught via integration tests and fixed (maybe in a diffident way?).

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/scheduling/Scheduler.java
vinok88: +1 for the suggestion

warunalakshitha:
@manuranga I need to send the same fix in the main branch as well. In the main branch, there is a possibility to reschedule the same item twice. and during load test we got.

java.util.EmptyStackException
	at java.base/java.util.Stack.peek(Stack.java:102)
	at java.base/java.util.Stack.pop(Stack.java:84)
	at io.ballerina.runtime.internal.scheduling.ItemGroup.get(Scheduler.java:662)
	at io.ballerina.runtime.internal.scheduling.Scheduler.run(Scheduler.java:319)
	at io.ballerina.runtime.internal.scheduling.Scheduler.runSafely(Scheduler.java:291)

bvm/ballerina-runtime/src/main/java/org/ballerinalang/jvm/scheduling/Scheduler.java
warunalakshitha: Yes it is already inside the ItemGroup. Might confuse since ItemGroup is the inner class of the scheduler.

warunalakshitha:
Did a perf test as below.

```
Concurrency  = 100 Threads
Ramp up period = 1 second
time = 1 hour
Bal file with HTTP service contact 50 filters ( 20 filters have 1 ms sleep)

```

*Results*

```
Before fix

TPS = 2,216
Timeouts = 79

With Fix

TPS = 2521
Timeouts = 0

```

manuranga (approved):

PR #33874, opened 2021-11-17
[beta4] Fix blocked scheduler items not getting reschedule

### LINKED ISSUE #33827: Blocked scheduler items intermittently not getting reschedule

warunalakshitha:
## Purpose
Fixes #33827

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/scheduling/Scheduler.java
```diff
@@ -168,17 +168,7 @@ public FutureValue scheduleToObjectGroup(Object[] params, Function function, Str
         SchedulerItem item = new SchedulerItem(function, params, future);
         future.strand.schedulerItem = item;
         totalStrands.incrementAndGet();
-        ItemGroup group = objectGroup.get();
-        if (group == null) {
-            group = new ItemGroup(item);
-            objectGroup.set(group);
```
warunalakshitha, line 174: For nonisolated methods, we use this method. But we can initialize the object group during scheduler init. This is a minor improvemnt

HindujaB (dismissed):
LGTM

HindujaB (approved):
LGTM

PR #33936, opened 2021-11-22
Fix incorrect resulting type of `unary-plus` expression

### LINKED ISSUE #33881: Incorrect result type in `unary-plus` expression
**Description:**
According to the spec

`The unary + operator returns the value of its operand expression. The static type of the operand must be a number, and the static type of the result is the same as the static type of the operand expression. `

```
public function main() {
    int:Unsigned16 g = 256;
    int:Unsigned16 _ = +g; // currently results in an error. But it should allow, since the result type of the expression is same as the operand expression
}
```

**Steps to reproduce:**

**Affected Versions:**

rpjayasekara:
Fixes #33881

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolTable.java
```diff
@@ -626,7 +626,7 @@ private void defineIntegerUnaryOperations() {
         BType[] intTypes = {intType, byteType, signed32IntType, signed16IntType, signed8IntType, unsigned32IntType,
                 unsigned16IntType, unsigned8IntType};
         for (BType type : intTypes) {
-            defineUnaryOperator(OperatorKind.ADD, type, intType);
+            defineUnaryOperator(OperatorKind.ADD, type, type);
```
MaryamZi, line 629: Still doesn't work for user-defined subtypes.

```ballerina
type Foo -1|1;

public function main() {
    Foo g = -1;
    Foo _ = +g; // error: incompatible types: expected 'Foo', found 'int'
}
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/unaryoperations/unary-operation.bal
```diff
@@ -106,49 +106,49 @@ function testComplementOperator() {
 function testUnaryOperationsWithIntSubtypes() {
     int:Unsigned8 x1 = 7;
     int y1 = ~x1;
-    int:Unsigned8 x2 = +7;
```
MaryamZi, line 109: Why are we changing these? I think we need tests like these also.

Shall we leave these as is and add a new set of tests?

rpjayasekara: will add separate tests

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolTable.java
rpjayasekara: Will fix this

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -1942,6 +1942,12 @@ public BSymbol getUnaryOpsForTypeSets(OperatorKind opKind, BType type) {
         if (!validNumericTypeExists) {
             return symTable.notFoundSymbol;
         }
+        if (opKind == OperatorKind.ADD) {
+            if (type.isNullable()) {
+                return createUnaryOperator(opKind, type, BUnionType.create(null, type, symTable.nilType));
```
MaryamZi, line 1947: Why do we need to create a union type here? Can't we just return `createUnaryOperator(opKind, type, type)` if `opKind == OperatorKind.ADD`?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/unaryoperations/unary-operation.bal
MaryamZi: Shall we undo these changes?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/unaryoperations/unary-operation.bal
```diff
@@ -324,6 +324,39 @@ function testNullableUnaryExpressions() {
     assertEquality(a19, 10);
 }
 
+function testResultingTypeOfUnaryPlus() {
+    int:Unsigned8 a = 10;
+    a = +a;
```
MaryamZi, line 329: I would use a separate variable instead, in a way to differentiate between the operation being performed on the operand itself vs an operation that makes use of the value of the operand.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
rpjayasekara: Yeah, no need of creating it again. We can just return the type. Will change that

tests/jballerina-unit-test/src/test/resources/test-src/expressions/unaryoperations/unary-operation.bal
rpjayasekara: +1

MaryamZi (approved):
We need to add this to the release note also.

hasithaa (approved):

PR #33973, opened 2021-11-24
Fix inferred type in `table-constructor` for fields with same name but different types

### LINKED ISSUE #33297: Deriving the type for a table constructor doesn't correctly handle members having fields of the same name but of different types

### LINKED ISSUE #34019: Inferred table constraint type is incorrect for mapping-cons members that have spread fields which have optional fields
**Description:**
$subject. 

```
record {|string a; int b?;|} m = {a: "str", b: 2};

public function main() {
    var v1 = table [
            {...m},
            {a: "str", b: 2}
        ];

    io:println(typeof v1); // @output typedesc table<record {| string a; int b; |}>
    io:println(v1.add({...m})); // Gives "missing non-defaultable required record field 'b'"
}
```
Correct inferred type should be `table<record {| string a; int b?; |}>`

**Affected Versions:**
SLBeta4 and before

lochana-chathura:
## Purpose
$subject.

Fixes #33297 
Fixes #34019

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [x] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1098,66 +1098,135 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+            }
 
-            if (fieldNames.contains(fieldName)) {
-                dlog.error(tableConstructorExpr.pos,
-                        DiagnosticErrorCode.CANNOT_INFER_MEMBER_TYPE_FOR_TABLE_DUE_AMBIGUITY,
-                        fieldName);
-                return symTable.semanticError;
+            if (!member.sealed) {
+                BType restFieldType = member.restFieldType;
+                if (isUniqueType(restFieldTypes, restFieldType)) {
```
MaryamZi, line 1126: Should we directly use `org.wso2.ballerinalang.compiler.semantics.analyzer.Types#isSameType(org.wso2.ballerinalang.compiler.semantics.model.types.BType, org.wso2.ballerinalang.compiler.semantics.model.types.BType)` instead?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1098,66 +1098,135 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+     *
+     * @param fields the list of same name fields
+     * @return a {@code List<BType>}
+     */
+    private List<BType> getUniqueTypesFromSameNameFieldList(List<BField> fields) {
+        List<BType> uniqueTypes = new ArrayList<>();
+
+        for (BField field : fields) {
+            BType type = field.getType();
+            if (type.tag == TypeTags.UNION) {
```
MaryamZi, line 1183: We may have to update this change for the type reference type change.

For example,

```ballerina
type Foo int|string;

public function main() {
    Foo f = 1;

    var t = table [
        {a: f},
        {a: 1}
    ];
    int _ = t;
}
```

results in

```
incompatible types: expected 'int', found 'table<record {| (Foo|int) a; |}>'
```

although `Foo` already includes `int`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1098,66 +1098,135 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+                ((BUnionType) type).getMemberTypes().forEach(t -> {
```
MaryamZi, line 1184: Shall we use simple loops for trivial stuff?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1098,66 +1098,135 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+
+            for (BType memType : memTypes) {
+                BRecordType bMemType = (BRecordType) memType;
+                if (bMemType.sealed) {
+                    continue;
+                }
+
+                if (bMemType.fields.containsKey(fieldName)) {
+                    continue;
+                }
```
MaryamZi, line 1146:
```suggestion
                if (bMemType.sealed || bMemType.fields.containsKey(fieldName)) {
                    continue;
                }
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1098,66 +1098,135 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+
+                BType restFieldType = bMemType.restFieldType;
```
MaryamZi, line 1148: Wondering if we have to explicitly handle never-typed individual fields and rest fields.

```ballerina
type Foo record {|
    int i;
    never j?;
    never k?;
    never...;
|};

function fn(Foo f) {
    var t = table [
        {...f},
        {i: 1, j: 2, l: ""}
    ];
    int _ = t;
}
```

results in

```
incompatible types: expected 'int', found 'table<record {| int i; (never|int) j?; never k?; (string|never) l?; never...; |}>'
```

`T|never` is `T`, so wondering if we should be saying `never|int`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1098,66 +1098,135 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+    /**
+     * Replaces the {@code BField} type with a union type, derived from the given unique list of types.
+     *
+     * @param field       - bField of which, the type to be replaced
+     * @param uniqueTypes - list of unique types
+     * @return a {@code BField}
+     */
+    private BField replaceFieldType(BField field, List<BType> uniqueTypes) {
+        LinkedHashSet<BType> memberTypes = new LinkedHashSet<>(uniqueTypes);
+        BType unionType = BUnionType.create(null, memberTypes);
```
MaryamZi, line 1206: Length of `uniqueTypes` can be 1, right? If so we shouldn't create union types.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1098,66 +1098,135 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+    }
+
+    /**
+     * Replaces the {@code BField} type with a union type, derived from the given unique list of types.
+     *
+     * @param field       - bField of which, the type to be replaced
+     * @param uniqueTypes - list of unique types
+     * @return a {@code BField}
+     */
+    private BField replaceFieldType(BField field, List<BType> uniqueTypes) {
```
MaryamZi, line 1204: IMO, just looking at the name, it feels like this method replaces the type of `field`, whereas it is actually creating a new field (as it should).

How about something like
```suggestion
    private BField createFieldWithType(BField field, List<BType> uniqueTypes) {
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1098,66 +1098,135 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+
+        BVarSymbol originalSymbol = field.symbol;
+        BVarSymbol fieldSymbol = new BVarSymbol(originalSymbol.flags, originalSymbol.name, originalSymbol.pkgID,
+                unionType, originalSymbol.owner, originalSymbol.pos, VIRTUAL);
+
+        return new BField(field.name, field.pos, fieldSymbol);
+    }
+
+    private boolean hasOptionalFields(List<BField> fields) {
+        return fields.stream().anyMatch(field -> field.symbol.getFlags().contains(Flag.OPTIONAL));
```
MaryamZi, line 1216: We try to avoid streams for trivial checks due to the potential impact on performance.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1098,66 +1098,135 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
             if (isOptional) {
-                field.symbol.flags = Flags.asMask(EnumSet.of(Flag.OPTIONAL));
-            } else if (requiredFieldNames.contains(fieldName) && keySpecifierFieldNames.contains(fieldName)) {
-                field.symbol.flags = Flags.asMask(EnumSet.of(Flag.REQUIRED)) + Flags.asMask(EnumSet.of(Flag.READONLY));
-            } else if (requiredFieldNames.contains(fieldName)) {
-                field.symbol.flags = Flags.asMask(EnumSet.of(Flag.REQUIRED));
+                resultantField.symbol.flags = Flags.asMask(EnumSet.of(Flag.OPTIONAL));
+            } else if (keySpecifierFieldNames.contains(fieldName)) {
+                resultantField.symbol.flags = Flags.asMask(EnumSet.of(Flag.REQUIRED)) +
+                        Flags.asMask(EnumSet.of(Flag.READONLY));
```
MaryamZi, line 1161:
```suggestion
                    BVarSymbol symbol = resultantField.symbol;
                    symbol.flags |= Flags.REQUIRED;
                    symbol.flags |= Flags.READONLY;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
lochana-chathura: Then we seem to consider scenarios like `anydata`, `any` also at both places where union type is created.
```
public function main() {
    anydata f = 1;

    var t = table [
        {a: f},
        {a: 1}
    ];
    int _ = t;
}
```
`incompatible types: expected 'int', found 'table<record {| (anydata|int) a; |}>`

The problem is only when there is an error thrown at the user. Otherwise, with correct assignments, it works right?

Wondering whether this is affecting only changes introduced in this PR. What about all other places we create union type from a list. I kind of feel like there should be a common API to create a union while doing these validations.

Will update the PR.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
lochana-chathura: Is there any reason we consider previous flags into account here? 
If not isn't it better to not depend on them/assume them.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
lochana-chathura: Kindly check with the new changes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
lochana-chathura: Both cases are fixed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: My bad, misread it. Can't we just do

```suggestion
                resultantField.symbol.flags = Flags.REQUIRED | Flags.READONLY;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1097,76 +1097,202 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+
+                BType restFieldType = bMemType.restFieldType;
+                types.add(restFieldType);
+            }
+
+            BField resultantField = createFieldWithType(fields.get(0), types);
+            boolean isOptional = hasOptionalFields(fields) || fields.size() != memTypesSize;
+
+            if (isOptional) {
+                resultantField.symbol.flags = Flags.asMask(EnumSet.of(Flag.OPTIONAL));
```
MaryamZi, line 1156:
```suggestion
                resultantField.symbol.flags = Flags.OPTIONAL;
```

MaryamZi: Can update L1161 also?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1097,76 +1097,202 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+                    addEffectiveMemberTypes(eBTypes, ((BUnionType) memberType).getMemberTypes());
+                    continue;
+                case TypeTags.TYPEREFDESC:
+                    BType constraint = types.getReferredType(memberType);
+                    if (constraint.tag == TypeTags.UNION) {
+                        addEffectiveMemberTypes(eBTypes, ((BUnionType) constraint).getMemberTypes());
+                        continue;
+                    }
+                    bType = constraint;
+                    break;
```
MaryamZi, line 1225: There seems to be a codecov warning?

lochana-chathura: added a test case in 30420cb

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1097,76 +1096,201 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+        LinkedHashSet<BType> bTypes = new LinkedHashSet<>(bTypeList);
+
+        if (bTypeList.size() == 1) {
+            return bTypes;
+        }
+
+        LinkedHashSet<BType> eBTypes = new LinkedHashSet<>(bTypes.size());
+        addEffectiveMemberTypes(eBTypes, bTypes);
+
+        eBTypes = filterRedundantAnyAnydataSubtypes(eBTypes);
```
MaryamZi, line 1203: We can just use the `isAssignable` checks instead and filter out a type, right?

I think it would be something like
- org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker#getRepresentativeBroadType
- or the type logic in org.wso2.ballerinalang.compiler.semantics.analyzer.Types#inferRecordFieldType

Basically filtering out only the super types from the set while still representing the same value space.

That way we don't have to special case types and all types including user-defined types that are the same as any/anydata will also be covered.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -1097,76 +1096,159 @@ private BType inferTableMemberType(List<BType> memTypes, BType expType) {
...
+        for (BType memberType : bTypes) {
+            BType bType;
+            switch (memberType.tag) {
+                case TypeTags.NEVER:
+                    continue;
+                case TypeTags.UNION:
+                    addEffectiveMemberTypes(eBTypes, ((BUnionType) memberType).getMemberTypes());
+                    continue;
+                case TypeTags.TYPEREFDESC:
+                    BType constraint = types.getReferredType(memberType);
```
MaryamZi, line 1214: Doesn't `getRepresentativeBroadType` handle these as well? For type reference types we should probably update `getRepresentativeBroadType` itself to use `types.getReferredType(inferredTypeList.get(i))` as the `type` instead of `inferredTypeList.get(i)`.

lochana-chathura: But getRepresentativeBroadType does not do the flattening right?
e.g.
```bal
function testTableConstructorWithVar5() {
    int|never f = 1;

    var v1 = table [
            {a: f},
            {a: 0}
        ];
    int _ = v1;
}
```
`ERROR [test_parser.bal:(8:13,8:15)] incompatible types: expected 'int', found 'table<record {| (int|never) a; |}>'`
Basically what it does is eliminate a type if it is assignable to another type.

Also for type references `inferredTypeList.get(i)` and `types.getReferredType(inferredTypeList.get(i))` have no difference to the `types.isAssignable(otherType, type)` right? We have to loop again and modify the list replacing types with `types.getReferredType(inferredTypeList.get(i))`?

MaryamZi: I think the issue with `int|never` and `int` not being simplified to `int` in `getRepresentativeBroadType` is because we are passing unions as is right? 

What if for each member type `T` in the list passed as a parameter to `getResultantUnionOrNonUnionType`, we create a new set of members where

1. if `T` is a type reference type we get the referred type and use that as `T`
2. if `T` is a union add to the set of members the result of calling `org.wso2.ballerinalang.compiler.semantics.model.types.BUnionType#getMemberTypes`, if not we add `T` as is

The reason to call `getReferredType` is for this flattening, so it has to happen here (point 1.) instead of in `getRepresentativeBroadType` like I originally suggested.

MaryamZi (approved):
Suggested a minor improvement re: identifying the smallest super type, just to make sure we aren't duplicating type checking in two places and to make sure we are covering a wider range of combinations.

We can merge as is due to time constraints and introduce the changes separately also.

lochana-chathura:
> We can merge as is due to time constraints and introduce the changes separately also.
I'll address it separately. Thank you!.

gimantha (approved):

PR #33828, opened 2021-11-16
[LS] Add completion support for mapping match pattern

malinthar:
## Purpose
$subject as depicted below. 

capture

Fixes #33383 

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [x] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #33979, opened 2021-11-24
Add missing completions in multiple wait actions

### LINKED ISSUE #33886: Missing completions inside mapping constructor expression in multiple-wait-action
**Description:**

1. Missing completions for field names.
Screenshot 2021-11-17 at 15 46 58

2. Missing completions for workers.
Screenshot 2021-11-17 at 16 14 42

**Describe your task(s)**
```
type Result record {
 (int|error) a; 
 (string|error) b; 
 };

 public function main() {

    worker WA returns int|error {
        return 0;
    }
  
    worker WB returns string|error {
        return "";
    }

    Result r = wait {a:WA, b:WB};
}
```

Relates to #33632

TharushiJay:
## Purpose
This PR refactors the logic in `WaitFieldsListNodeContext` to add support for missing completions in multiple wait actions.

Fixes #33886 

## Samples
Screenshot 2021-12-02 at 12 57 12

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


malinthar (changes requested):
Changes Looks Good. However there are several methods that are identical to the methods in the MappingConstructorExpressionNodeContext. With this PR  #33828 I have moved the duplicated logic into an abstract class called MappingContextProvider. We can extend that abstract class here as well. WDYT?

TharushiJay:
> Changes Looks Good. However there are several methods that are identical to the methods in the MappingConstructorExpressionNodeContext. With this PR #33828 I have moved the duplicated logic into an abstract class called MappingContextProvider. We can extend that abstract class here as well. WDYT?

+1. Will refactor the code once it is merged.

gayalkuruppu (commented):
Is it okay to provide incorrect default values in the following scenario? WDYT @TharushiJay  @nadeeshaan  ?

https://user-images.githubusercontent.com/46120162/143392115-3d209cbb-36fb-4f7c-b251-1c3e7a4edf26.mov

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/VariableCompletionItemBuilder.java
```diff
@@ -49,6 +50,16 @@ public static CompletionItem build(VariableSymbol varSymbol, String label, Strin
         return item;
     }
 
+    public static CompletionItem build(WorkerSymbol workerSymbol, String label, String type) {
```
nadeeshaan, line 53: Shall we create a separate builder for the workers?

nadeeshaan (changes requested):

nadeeshaan:
> Is it okay to provide incorrect default values in the following scenario? WDYT @TharushiJay @nadeeshaan ?
> 
>  Screen.Recording.2021-11-25.at.12.03.20.mov

We should not provide the mapping expression related completions here, IMHO. We can stick to the workers and the futures.

TharushiJay:
> > Is it okay to provide incorrect default values in the following scenario? WDYT @TharushiJay @nadeeshaan ?
> > Screen.Recording.2021-11-25.at.12.03.20.mov
> 
> We should not provide the mapping expression related completions here, IMHO. We can stick to the workers and the futures.

@nadeeshaan We do provide futures and workers within the value expression. However, as the **completion insert text** of the field completion item, we get the default values (0, “ “). Since we cannot provide a default value for a future/worker, would you recommend to leave the completion insert text empty, in order to avoid the syntax errors?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/WaitFieldsListNodeContext.java
```diff
@@ -44,14 +54,50 @@ public WaitFieldsListNodeContext() {
...
+        List<Symbol> filteredList = visibleSymbols.stream() 
+                .filter(CommonUtil.getVariableFilterPredicate().or(symbol -> symbol.kind() == WORKER))
+                .collect(Collectors.toList());
+        return this.getCompletionItemList(filteredList, context);
+    }
+
+    protected List<LSCompletionItem> getExpressionsCompletionsForQNameRef(BallerinaCompletionContext context,
+                                                                          NonTerminalNode nodeAtCursor) {
+        QualifiedNameReferenceNode qNameRef = (QualifiedNameReferenceNode) nodeAtCursor;
+        Predicate<Symbol> filter = symbol -> symbol instanceof VariableSymbol
```
malinthar, line 82: Is there a particular reason to skip types. Would it be incorrect to use the `QnameReferenceUtil.getExpressionCompletions()` API

malinthar (commented):
Changes look good. I just have the following concern

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/WaitFieldsListNodeContext.java
TharushiJay: There was no particular reason. Will use it instead 👍

malinthar (dismissed):
LGTM

language-server/modules/langserver-core/src/test/resources/completion/statement_context/source/wait_action_ctx_source16.bal
```diff
@@ -0,0 +1,25 @@
...
+    }
+  
+    worker WB returns string|error {
+        return "";
+    }
+    
+    future<int> fint = start getInt();
+    future<string> fstr = start getString(); 
+
+    map<string|error> myVar = wait {fieldName: module1: };
```
IMS94, line 16: Can we expect qualified name references for workers? IMO, no. Only usage will be like this:
```
map<string|error> myVar = wait {fieldName: start module1:<function_name>() };
```
Therefore, I think we don't need the qualified name reference part in the completion context (or return an empty list in that case). WDYT?

IMS94 (changes requested):
Changes look good. Have concern regarding usage of qualified name references in wait action.

language-server/modules/langserver-core/src/test/resources/completion/statement_context/source/wait_action_ctx_source16.bal
TharushiJay: Agreed about the usage. Returned an empty list instead.

nadeeshaan (approved):
LGTM

PR #33985, opened 2021-11-24
Fix TypeGuardTest

MaryamZi:
## Purpose
$title.

### DIFF
--- a/language-server/modules/langserver-core/src/test/resources/codeaction/type-guard/config/typeGuardVariableCodeAction3.json
+++ b/language-server/modules/langserver-core/src/test/resources/codeaction/type-guard/config/typeGuardVariableCodeAction3.json
@@ -16,7 +16,7 @@
                             "character": 28
                         }
                     },
-                    "newText": "\n    if cur is SingleCharDelim {\n\n    } else if cur is CompoundAssignDelim {\n\n    } else if cur is \"\u003c\u003c\" {\n\n    } else if cur is \"\u003e\u003e\" {\n\n    } else if cur is \"int\" {\n\n    } else if cur is \"string\" {\n\n    } else if cur is 100 {\n\n    } else if cur is \"200\" {\n\n    } else {\n\n    }\n"
+                    "newText": "\n    if cur is Token {\n\n    } else {\n\n    }\n"
                 }
             ]
         }


language-server/modules/langserver-core/src/test/resources/codeaction/type-guard/config/typeGuardVariableCodeAction3.json
```diff
@@ -16,7 +16,7 @@
                             "character": 28
                         }
                     },
-                    "newText": "\n    if cur is SingleCharDelim {\n\n    } else if cur is CompoundAssignDelim {\n\n    } else if cur is \"\u003c\u003c\" {\n\n    } else if cur is \"\u003e\u003e\" {\n\n    } else if cur is \"int\" {\n\n    } else if cur is \"string\" {\n\n    } else if cur is 100 {\n\n    } else if cur is \"200\" {\n\n    } else {\n\n    }\n"
+                    "newText": "\n    if cur is Token {\n\n    } else {\n\n    }\n"
```
MaryamZi, line 19: The source here is

```ballerina
type Token SingleCharDelim|MultiCharDelim|Keyword;
type SingleCharDelim "+" | "-";
type MultiCharDelim "<<" | ">>" | CompoundAssignDelim  ;
type CompoundAssignDelim "+=" | "-=";
type Keyword KEY | "string" | 100 | "200" ;
const KEY = "int";

    Token? cur = getToken();
}

function getToken() returns Token? {
    return ();
}
```

IINM, this change is due to the type reference type change we did on master, and union types not being flattened. From a user's perspective, I don't have a strong opinion on which one the user may prefer.

But @pcnfernando and I had an offline discussion regarding having a flattened representation also. If/when that is available, we can use the relevant set of types for the code action here IMO (if we want the previous behaviour).

@TharushiJay, @IMS94, @nadeeshaan, for now, hope this change is OK.

MaryamZi: Looping in @hasithaa also.

IMS94: @MaryamZi This is actually better (and correct) than what was suggested earlier. +1 from me.

IMS94 (approved):
LGTM!

language-server/modules/langserver-core/src/test/resources/codeaction/type-guard/config/typeGuardVariableCodeAction3.json
MaryamZi: Thanks, @IMS94.

For cases like error unions, I also think the new way may generally be better.

```ballerina
type Foo distinct error;
type Bar distinct error;
type Baz distinct error;

type Error Foo|Bar|Baz;

public function main() {
   fn(); // using the type guard here   
}

function fn() returns Error? {  
}
```

But my concern is that there can be scenarios where the user wants the type test against the individual members too. I'm not sure if there can be like another code action that splits it up further?

Will go ahead and merge this PR now btw, since it is a blocker to sync the beta4 branch to master.

IMS94: Sure. When we consider ` Token? cur = getToken();` line, going beyond `if Token {} else {}` type guard is out of scope IMO. Such cases should be handled manually by the user. @nadeeshaan WDYT?

PR #33987, opened 2021-11-24
Support generating resources with compiler plugins code generators

### LINKED ISSUE #32648: In-Built compiler plugins should have an order when executing

### LINKED ISSUE #32717: `CompilerLifecycleTask` is not executed when we use `bal run` command
**Description:**
We have developed a in-built compiler plugin for ballerina OpenAPI to generate OpenAPI docs for HTTP services in built time. In this plugin we have used `CompilerLifecycleTask` provided in ballerina project API to inject resources to generated thin-jar and executable jar. 

When we use `bal build` command, `CompilerLifecycleTask` is invoked and resource injection will be successful. But if we use `bal run` command then `CompilerLifecycleTask` is not invoked and resource injection will be unsuccessful. 

**Affected Versions:**
- Ballerina Swan Lake Beta3 RC3

Related to [#1616](https://github.com/ballerina-platform/ballerina-standard-library/issues/1616)

### LINKED ISSUE #33651: Improve code generator plugin tasks to support resource file generation
I introduced code generators to the current compiler plugin design. You can find more details here: 
- https://github.com/ballerina-platform/ballerina-lang/issues/29283
- https://github.com/ballerina-platform/ballerina-lang/pull/33649

But in this version, I didn't support resource file generation. This is because the Project API does not represents resources in the API.

azinneera:
Fixes #33651
Fixes #32648
Fixes #32717

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CompileTask.java
```diff
@@ -72,6 +73,15 @@ public void execute(Project project) {
                 BuildTime.getInstance().packageResolutionDuration = System.currentTimeMillis() - start;
                 start = System.currentTimeMillis();
             }
+
+            // run built-in code generator compiler plugins
+            if (project.kind().equals(ProjectKind.BUILD_PROJECT)) {
```
xlight05, line 78: This could be a problem for Azure functions/AWS lambda modules. as those modules contain single file projects. I understand from the compiler POV Single File projects can't have more than one sources but from users point of view it can be bit confusing as they are not aware of the internals. 
In this case we might have to turn the single file project into build project if theres additional generated sources. I think this can be very complicated task :/
If we can't fix this issue, i think we have to force users to always use Projects instead of Single file projects when using AWS lambda and Azue functions modules.

@sameerajayasoma  @azinneera

sameerajayasoma: I think It is better to restrict your feature only to build projects

sameerajayasoma (dismissed):

pramodya1994 (approved):

PR #34057, opened 2021-11-30
Return temp path for targetDir() of a single file project

### LINKED ISSUE #33660: targetDir() method returns a wrong path for single file projects
**Description:**
Please refer https://github.com/ballerina-platform/ballerina-lang/blob/stage-slbeta4/compiler/ballerina-lang/src/main/java/io/ballerina/projects/Project.java#L86

We will have to check and handle this implementation for single file projects.
Currently it returns `singlefile.bal/target` as the target dir path.

**Steps to reproduce:**

**Affected Versions:**

Ibaqu:
Fixes #33660 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/Project.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/Project.java
@@ -18,13 +18,11 @@
 package io.ballerina.projects;
 
 import io.ballerina.projects.environment.ProjectEnvironment;
-import io.ballerina.projects.util.ProjectConstants;
 import org.wso2.ballerinalang.compiler.PackageCache;
 import org.wso2.ballerinalang.compiler.util.CompilerContext;
 import org.wso2.ballerinalang.compiler.util.CompilerOptions;
 
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Optional;
 
 import static org.ballerinalang.compiler.CompilerOptionName.PROJECT_DIR;
@@ -81,13 +79,7 @@ public abstract class Project {
         return this.sourceRoot;
     }
 
-    public Path targetDir() {
-        if (this.buildOptions.getTargetPath() == null) {
-            return this.sourceRoot.resolve(ProjectConstants.TARGET_DIR_NAME);
-        } else {
-            return Paths.get(this.buildOptions().getTargetPath());
-        }
-    }
+    public abstract Path targetDir();
 
     protected void setCurrentPackage(Package currentPackage) {
         // TODO Handle concurrent read/write to the currentPackage variable
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/bala/BalaProject.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/bala/BalaProject.java
@@ -126,4 +126,9 @@ public class BalaProject extends Project {
         }
         return true;
     }
+
+    @Override
+    public Path targetDir() {
+        throw new UnsupportedOperationException("Target Dir is not supported for BalaProject");
+    }
 }
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/bala/BalaProject.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/bala/BalaProject.java
@@ -129,6 +129,6 @@ public class BalaProject extends Project {
 
     @Override
     public Path targetDir() {
-        throw new UnsupportedOperationException("Target Dir is not supported for BalaProject");
+        throw new UnsupportedOperationException("target directory is not supported for BalaProject");
     }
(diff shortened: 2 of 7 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/projects/Project.java
```diff
@@ -82,6 +84,13 @@ public Path sourceRoot() {
     }
 
     public Path targetDir() {
+        if (this.projectKind == ProjectKind.SINGLE_FILE_PROJECT) {
+            try {
+                return Files.createTempDirectory("ballerina-cache" + System.nanoTime());
```
azinneera, line 89: This will create temp directories each time we call the function. We can improve this as follows: 
1. Make `Project.target` abstract and implement it in SingleFileProject and BuildProjects. BalaProject should throw an unsupported operation exception.
2. In SingleFileProject, declare a class level variable (e.g. targetDir) and create and assign the temp location at the project creation
3. BuildProject can contain the existing implementation

azinneera: After this change, shall we try to refactor the tasks in CLI to use this method?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/SingleFileProject.java
```diff
@@ -34,6 +36,8 @@
  */
 public class SingleFileProject extends Project {
 
+    Path targetDir;
```
azinneera, line 39: This can be private final

compiler/ballerina-lang/src/main/java/io/ballerina/projects/bala/BalaProject.java
```diff
@@ -126,4 +126,9 @@ private boolean isFilePathInProject(Path filepath) {
         }
         return true;
     }
+
+    @Override
+    public Path targetDir() {
+        throw new UnsupportedOperationException("Target Dir is not supported for BalaProject");
```
azinneera, line 132: I have second thoughts on this. Should we rather return `null` and add it in javadocs for the method? @sameerajayasoma @hevayo wdyt?

azinneera: We'll go ahead with throwing the exception.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/directory/SingleFileProject.java
Ibaqu: I changed it to private Path targetDir. final might not work since the initialization is done inside a try block and we are ignoring the exception entirely. WDYT?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/bala/BalaProject.java
azinneera, line 132: Shall we reword this to `target directory`

Ibaqu: Changed to "Target directory is not supported for Balaproject"

azinneera:
Shall we add test cases for these? Or rather we can modify the existing tests in project-api-test to assert the target directory path.

project-api/project-api-test/src/test/java/io/ballerina/projects/test/TestSingleFileProject.java
```diff
@@ -74,6 +74,21 @@ public void testLoadSingleFile() {
...
+        SingleFileProject project = null;
+        try {
+            project = TestUtils.loadSingleFileProject(projectPath);
+        } catch (Exception e) {
+            Assert.fail(e.getMessage());
+        }
+
+        Path targetDirPath = project.targetDir();
+        Assert.assertNotNull(targetDirPath);
+        Assert.assertTrue(targetDirPath.toFile().exists());
```
azinneera, line 89: Lets add an assertion to make sure its a temp location

Ibaqu: Added extra assertion : `Assert.assertEquals(Path.of(System.getProperty("java.io.tmpdir")), targetDirPath.getParent());`

Ibaqu: Changed to us `Paths.get()`

azinneera (approved):

Issue #34066, opened 2021-11-30
NPE with strand groups when calling object methods via runtime api


warunalakshitha:
**Description:**
Observed NPE with strand group when calling methods using objects created through runtime API.

PR #34067, opened 2021-11-30
Fix NPE with strand groups when calling object methods via runtime api 

warunalakshitha:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/34066


HindujaB (approved):
LGTM

PR #34079, opened 2021-12-01
Disallow optional fields in record binding pattern

### LINKED ISSUE #33389: No error when binding patterns bind optional fields of a record
**Description:**
$title, see https://github.com/ballerina-platform/ballerina-spec/issues/445.

Additionally the type used is also incorrect.

Need to follow up on the spec discussion before we fix this (to decide if we are disallowing this or making the type `T?`).

**Steps to reproduce:**
```ballerina
import ballerina/io;

type Employee record {
    string name;
    int id;
    Employee lead?;
};

public function main() {
    Employee e = {name: "Jo", id: 1234};

    string eName;
    int eId;
    Employee eLead;
    {name: eName, id: eId, lead: eLead} = e; // Allowed!
    io:println(eLead is Employee); // false
    io:println(<any> eLead is ()); // true

    var {name, id, lead} = e;

    Employee ld = lead; // Allowed!
    io:println(ld is Employee); // false
    io:println(<any> ld is ()); // true
}
```

**Affected Versions:**
slbeta3

### LINKED ISSUE #33538: No error when a mapping binding pattern binds rest fields
**Description:**
$title. Where the field is of type `T`, the type seems to be considered `T?`. I don't think this is inline with the spec.

**Steps to reproduce:**
```ballerina
public function main() {
    record {| int i; int...; |} a = {i: 1};
    record {| int i; int...; |} {i: _, j} = a; // no error

    // int x = j; // incompatible types: expected 'int', found 'int?'
}
```

**Affected Versions:**
slbeta4-rc1

rpjayasekara:
Fixes #33389 
Fixes #33538

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/record-variable-reference-semantics-negative.bal
```diff
@@ -198,3 +198,23 @@ function testFieldAndIndexBasedVarRefs() returns [anydata, anydata] {
...
+   string eName;
+   int eId;
+   int eAge;
+
+   {name: eName, id: eId, age: eAge} = e;
+
+   var {name: name, id: id, age: age} = e;
+
+   var {name: nameNew, id: idNew, age: ageNew} = e;
+}
```
rdulmina, line 220: Missing new line

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1197,6 +1197,11 @@ public void visit(BLangRecordVariable varNode) {
             varNode.setBType(symResolver.resolveTypeNode(varNode.typeNode, env));
         }
 
+        if (isOptionalFieldAvailableInRecord(varNode.typeNode.getBType())) {
```
rdulmina, line 1200: This will disallow `{name: eName, id: eId} = e;` as well right? we need to allow this

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1749,6 +1749,9 @@ error.mismatched.visibility.qualifiers.in.object.field=\
 error.invalid.inclusion.of.object.with.private.members=\
   invalid object type inclusion with an object that has private fields or methods
 
+error.invalid.optional.field.in.record.binding.pattern=\
+  invalid record binding pattern; optional fields of records are not allowed in binding patterns
```
rdulmina, line 1753: shouldn't we be using the term `mapping binding pattern`  instead of `record binging pattern`? since there is no such binding pattern called `record binging pattern`? @MaryamZi WDYT?

tests/jballerina-unit-test/src/test/resources/test-src/statements/vardeclr/module_record_var_decl_negetive.bal
```diff
@@ -60,4 +60,18 @@ type ComplexRecord record {|
...
+    int age?;
+};
+
+function getEmployee() returns Employee {
+    return {name: "Jo", id: 1234, age: 12};
+}
+
+var {name: eName, id: eId, age: eAge} = getEmployee();
+
+Employee {name: eNameNew, id: eIdNew, age: eAgeNew} = getEmployee();
```
rdulmina, line 77: Missing new line

compiler/ballerina-lang/src/main/resources/compiler.properties
MaryamZi: Yeah, let's use mapping binding pattern. 

We initially had record and tuple binding patterns, before they were allowed with maps and arrays also. That's why there's usage of the terms record and tuple binding patterns.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1746,6 +1746,9 @@ error.mismatched.visibility.qualifiers.in.object.field=\
 error.invalid.inclusion.of.object.with.private.members=\
   invalid object type inclusion with an object that has private fields or methods
 
+error.invalid.optional.field.in.mapping.binding.pattern=\
+  invalid mapping binding pattern; optional fields of records are not allowed in mapping binding pattern
```
MaryamZi, line 1750: How about
```suggestion
  invalid mapping binding pattern; cannot bind optional fields
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1938,10 +1948,48 @@ public void visit(BLangRecordDestructure recordDeStmt) {
...
+            dlog.error(recordDeStmt.expr.pos, DiagnosticErrorCode.INVALID_OPTIONAL_FIELD_IN_MAPPING_BINDING_PATTERN);
+        }
         checkRecordVarRefEquivalency(recordDeStmt.pos, recordDeStmt.varRef, recordDeStmt.expr.getBType(),
                                      recordDeStmt.expr.pos);
     }
 
+    private boolean checkValidMappingBindingPatternForRecordVarRef(List<BLangRecordVarRefKeyValue> recordRefFields,
+                                                                   BType type) {
+        type = types.getReferredType(type);
+        if (type.tag != TypeTags.RECORD) {
```
MaryamZi, line 1962: We also need to handle maps, right?

https://ballerina.io/ballerina-spec/spec.html#mapping-binding-pattern

> a mapping-binding-pattern { f1: p1, f2: p2, ..., fn: pn, r } matches a mapping value m that includes fields f1, f2, ... , fn if pi matches the value of field fi for each i in 1 to n; if r is ...v then a successful match causes a new mapping value consisting of all the other fields to be assigned to v; a field-binding-pattern consisting of just a variable-name x is equivalent to a field-binding-pattern x: x;

So we can't really use a mapping binding-pattern in a destructuring assignment or variable declaration with a map as the expression?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1938,10 +1948,48 @@ public void visit(BLangRecordDestructure recordDeStmt) {
...
+            if (recordType.fields.containsKey(lhsField.variableName.value)) {
+                if (Symbols.isOptional(recordType.fields.get(lhsField.variableName.value).symbol)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    private boolean checkValidMappingBindingPatternForRecordVariable(List<BLangRecordVariable.
```
MaryamZi, line 1976: Logic seems similar to the previous method? You can probably use one method that accepts the field names and locations instead?

MaryamZi, line 1976: Should we rename this to something like `validateBindingPatternForNonRequiredFields`, reflecting the actual check?

tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/record-variable-reference-semantics-negative.bal
```diff
@@ -198,3 +198,23 @@ function testFieldAndIndexBasedVarRefs() returns [anydata, anydata] {
...
+function testOptionalFieldsInRecordBindingPattern(){
+   Employee e = {name: "Jo", id: 1234};
+
+   string eName;
+   int eId;
+   int eAge;
+
+   {name: eName, id: eId, age: eAge} = e;
+
+   var {name: name, id: id, age: age} = e;
```
MaryamZi, line 217: For all kinds, let's add some tests for when the field is just the variable-name also, and not `field-name : binding-pattern`.

```ballerina
var {name, id: id, age}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: Assuming we move the logging to this method.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1198,6 +1198,11 @@ public void visit(BLangRecordVariable varNode) {
             varNode.setBType(symResolver.resolveTypeNode(varNode.typeNode, env));
         }
 
+        if (checkValidMappingBindingPatternForRecordVariable(varNode.variableList, varNode.typeNode.getBType())) {
```
MaryamZi, line 1201: I think it'll be more useful to log the error for the exact field binding pattern, rather than the single error for the mapping binding pattern, so that it's easier to identify which field binding pattern(s) is invalid.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/record-variable-reference-semantics-negative.bal
MaryamZi: Also for when the field binding patterns are within other binding patterns.

e.g., for
```ballerina
type Employee record {
    string name;
    record {
        int id;
        int age?;
    } details?;
};
```

and

```ballerina
type Employee record {
    string name;
    record {
        int id;
        int age?;
    } details;
};
```

MaryamZi: And part of list and error binding patterns.

This doesn't fail atm.

```ballerina
type Employee record {
    string name;
    record {
        int id;
        int age?;
    } details?;
};

function fn() {
    Employee[2] e = [{name: "Jo", details: {id: 1234}}, {name: "Amy", details: {id: 1234}}];

    var [{name: nameNew, details: {id: idNew, age: ageNew}}, ...x] = e;
}
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/record-variable-reference-semantics-negative.bal
```diff
@@ -198,3 +198,23 @@ function testFieldAndIndexBasedVarRefs() returns [anydata, anydata] {
...
+
+   var {name: nameNew, id: idNew, age: ageNew} = e;
```
MaryamZi, line 219: Is this the correct file for these tests? Doesn't this file contain only destructuring assignments?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1938,10 +1948,48 @@ public void visit(BLangRecordDestructure recordDeStmt) {
...
+    private boolean checkValidMappingBindingPatternForRecordVarRef(List<BLangRecordVarRefKeyValue> recordRefFields,
+                                                                   BType type) {
+        type = types.getReferredType(type);
+        if (type.tag != TypeTags.RECORD) {
+            return false;
+        }
+        BRecordType recordType = (BRecordType) type;
+        for (BLangRecordVarRefKeyValue lhsField : recordRefFields) {
+            if (recordType.fields.containsKey(lhsField.variableName.value)) {
+                if (Symbols.isOptional(recordType.fields.get(lhsField.variableName.value).symbol)) {
```
MaryamZi, line 1968: Also need to disallow when the field corresponds to the rest field. The following doesn't fail.

```ballerina
type Employee record {
    string name;
};

function fn() {
    Employee e = {name: ""};
    Employee {name, id} = e;
}
```

MaryamZi: Related issue - https://github.com/ballerina-platform/ballerina-lang/issues/33538.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
rdulmina: We currently allow samples like below right?
```bal
map<string> mapvar = {a: "my"};

public function main() {
    string? b;
    {a: b} = mapvar;
}
```
I think the spec doesn't specifically mention allowing or disallowing this but this can be useful right?

MaryamZi: Similar scenario, right? A map doesn't have required fields, so using a map with a mapping binding pattern that has field binding patterns (in a variable declaration or destructuring assignment) is not allowed by the spec?

rpjayasekara: Yeah, right now this is allowed in this way,
```
    map<int> m = {
        "x": 1,
        "y": 2
    };

    var {x:x1, y:y1} = m;
```
Types of the `x1` and` y1` are` int?`. I think it is because map does not have required fields.

rpjayasekara: Discussed to disallow mapping binding patterns that has an expression of map

tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/record-variable-reference-semantics-negative.bal
rpjayasekara: Fixed this. But to handle the recursive records had to introduce two methods because data types are different in destructure expression and var expression

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: Aren't we fixing this?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: This isn't fixed? Basically for each non-rest binding pattern we check if it is not a required field in the RHS expression.

It is not a required field if the mapping value 
- is a map
- is a record with an optional field for the particular key
- is an open record which has a rest field that may include a field for the the mapping value (both `record {| T...; |}` and `record {}`)

This way we don't have to special case the error for maps also.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1521,6 +1524,18 @@ private void handleDeclaredWithVar(BLangVariable variable) {
                 BLangTupleVariable tupleVariable = (BLangTupleVariable) variable;
                 tupleVariable.setBType(rhsType);
 
+
+                if (rhsType.tag == TypeTags.ARRAY) {
```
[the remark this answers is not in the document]
MaryamZi: This check should probably happen wherever we check for never-present fields like

```ballerina
[record {| int i?; |}] y = [{}];
var [{j}] = y;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1198,6 +1198,9 @@ public void visit(BLangRecordVariable varNode) {
             varNode.setBType(symResolver.resolveTypeNode(varNode.typeNode, env));
         }
 
+        validateBindingPatternForVariableRecord (varNode.variableList,
```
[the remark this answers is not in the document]
rpjayasekara: We did, but since I had to introduce two methods for two data types I changed it again

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
rpjayasekara: Though you mentioned about the column index of where the error is logged. Earlier it was for the entire expression. Now it points out what are the optional fields

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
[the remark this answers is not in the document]
MaryamZi: We are basically validating a mapping binding pattern right? So IMO `validateBindingPatternForNonRequiredFields` is OK.

Record variable is an internal term we used when these were initially introduced, isn't quite valid now.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1938,10 +1943,75 @@ public void visit(BLangRecordDestructure recordDeStmt) {
...
+        for (BLangRecordVariable.BLangRecordVariableKeyValue lhsField : recordVar.variableList) {
+            if (recordType.fields.containsKey(lhsField.key.value)) {
+                if (Symbols.isOptional(recordType.fields.get(lhsField.key.value).symbol)) {
+                    dlog.error(lhsField.key.pos, DiagnosticErrorCode.INVALID_OPTIONAL_FIELD_IN_MAPPING_BINDING_PATTERN);
+                }
+            } else {
+                if (!recordType.sealed) {
+                    dlog.error(lhsField.key.pos, DiagnosticErrorCode.INVALID_OPTIONAL_FIELD_IN_MAPPING_BINDING_PATTERN);
+                }
+            }
```
MaryamZi, line 1975:
```suggestion
            } else if (!recordType.sealed) {
                dlog.error(lhsField.key.pos, DiagnosticErrorCode.INVALID_OPTIONAL_FIELD_IN_MAPPING_BINDING_PATTERN);
            }
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1938,10 +1943,75 @@ public void visit(BLangRecordDestructure recordDeStmt) {
             return;
         }
         typeChecker.checkExpr(recordDeStmt.expr, this.env);
+        validateBindingPatternForRecordDestructure(recordDeStmt.varRef.recordRefFields, recordDeStmt.expr.getBType(),
+                recordDeStmt.expr.pos);
         checkRecordVarRefEquivalency(recordDeStmt.pos, recordDeStmt.varRef, recordDeStmt.expr.getBType(),
                                      recordDeStmt.expr.pos);
     }
 
+    private void validateBindingPatternForVariableRecord(BLangRecordVariable recordVar,
```
MaryamZi, line 1952: Do we need this validation separately? Doesn't org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter#defineVariableList handle this?

I tried with 

```ballerina
type Foo record {|
    record {
        int b?;
    }[1] a?;
|};

function fn(Foo f) {
    var {a: [{b}]} = f;
}
``` 

The diagnostic for `a` was added twice (when debugging).

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2817,6 +2817,11 @@ private boolean defineVariableList(BLangRecordVariable recordVar, BRecordType re
                     defineMemberNode(value, env, restType);
                 }
                 continue;
+            } else {
```
MaryamZi, line 2820: The following doesn't result in an error for `b`. 

```ballerina
type Foo record {|
    map<int>[1] a;
|};

function fn(Foo f) {
    var {a: [{b}]} = f;
}
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/record-variable-reference-semantics-negative.bal
```diff
@@ -198,3 +198,47 @@ function testFieldAndIndexBasedVarRefs() returns [anydata, anydata] {
...
+  int ageOne;
+
+  {name: nameOne, details:{id: idOne, age: ageOne}} = e1;
+}
+
+function testMappingBindingPatternWithMap() {
+    map<string> stringMap = {"a":"Foo"};
+    string? foo;
+    {a:foo} = stringMap;
+}
```
MaryamZi, line 244:
```suggestion
}

```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/record-variable-reference-semantics-negative.bal
```diff
@@ -198,3 +198,47 @@ function testFieldAndIndexBasedVarRefs() returns [anydata, anydata] {
...
+  int idOne;
+  int ageOne;
+
+  {name: nameOne, details:{id: idOne, age: ageOne}} = e1;
+}
+
+function testMappingBindingPatternWithMap() {
+    map<string> stringMap = {"a":"Foo"};
+    string? foo;
+    {a:foo} = stringMap;
```
MaryamZi, line 243:
```suggestion
    {a: foo} = stringMap;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
rpjayasekara: Yeah, will remove this logic and handle from `SymbolEnter`.

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
```diff
@@ -413,6 +413,8 @@
           "cannot.use.type.inclusion.with.more.than.one.open.record.with.different.rest.descriptor.types"),
     INVALID_METHOD_CALL_EXPR_ON_FIELD("BCE2651", "invalid.method.call.expr.on.field"),
     INCOMPATIBLE_TYPE_WAIT_FUTURE_EXPR("BCE2652", "incompatible.type.wait.future.expr"),
+    INVALID_FIELD_BINDING_PATTERN_WITH_NON_REQUIRED_FIELD("BCE2653",
+            "invalid.optional.field.in.mapping.binding.pattern"),
```
MaryamZi, line 417: We need to update this also.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1938,10 +1938,43 @@ public void visit(BLangRecordDestructure recordDeStmt) {
...
+        }
+        BRecordType recordType = (BRecordType) type;
+        for (BLangRecordVarRefKeyValue lhsField : fields) {
+            if (recordType.fields.containsKey(lhsField.variableName.value)) {
+                if (Symbols.isOptional(recordType.fields.get(lhsField.variableName.value).symbol)) {
+                    dlog.error(lhsField.variableName.pos,
+                            DiagnosticErrorCode.INVALID_FIELD_BINDING_PATTERN_WITH_NON_REQUIRED_FIELD);
+                }
+            }
+            if (lhsField.variableReference.getKind() == NodeKind.RECORD_VARIABLE_REF) {
```
MaryamZi, line 1969: Do we still need this?

rpjayasekara: Yeah we need this to handle recordDestructure expressions.  An example like below
```
type Employee record {
    int a;
    record {
        int b?;
    } c;
};

public function main() {
    Employee emp = {a:4, c:{b:5}};
    int aa;
    int bb;
    {a:aa, c:{b:bb}} = emp;
}
```

MaryamZi: But the following then doesn't result in an error?

```ballerina
type Employee record {
    int a;
    record {
        int b?;
    }[1] c;
};

public function main() {
    Employee emp = {a: 4, c: [{b: 5}]};
    int aa;
    int bb;
    {a: aa, c: [{b: bb}]} = emp;
}
```

rpjayasekara: Fixed it. Remove the method and move it to `checkRecordVarRefEquivalency` where not available fields are being checked.

MaryamZi:
@rpjayasekara can you please rebase this PR on https://github.com/ballerina-platform/ballerina-lang/tree/stage-swan-lake.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1746,6 +1746,9 @@ error.mismatched.visibility.qualifiers.in.object.field=\
 error.invalid.inclusion.of.object.with.private.members=\
   invalid object type inclusion with an object that has private fields or methods
 
+error.invalid.field.binding.pattern.with.non.required.field=\
+  invalid mapping binding pattern; can only bind required fields
```
MaryamZi, line 1750:
```suggestion
  invalid field binding pattern; can only bind required fields
```

PR #34089, opened 2021-12-01
Fix completions within select clause context

### LINKED ISSUE #33934: On conflict clause not suggested after select clause in query expression context
**Description:**
onconflict

As depicted above on conflict keyword pair is not suggested at the depicted cursor position.

```ballerina
function findTop10GoldMedalsToAthletesRatio() returns error? {
    map<int> athletesByCountry = check findNumOfAthletesBycountry();
    stream<MedalStats, error?> medalStats = check getMedalStats();

    MyError onConflictError = error("Conflict");

    var ratios = from [string,int] entry in athletesByCountry.entries() 
    join var stats in medalStats on entry[0] equals stats.country 
    let float ratio = (<float>stats.gold/<float>entry[1]) * 100 order by ratio descending 
    limit 10
    select  {
        country: entry[0],
        ratio
    } o<cursor> ;

    io:println("Top 10 countries with gold medals/athletes ratio");
    
    foreach var ratio in ratios {
        io:println("\t", ratio.country, "\t", float:round(ratio.ratio), "%");
    }
}
```

Further, the sorting order in the on-conflict-clause context can be improved to order the expressions with `error` subtype at the top.

onConflict

nadeeshaan:
## Purpose
Fixes #33934 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


mohanvive (approved):

mohanvive:
One thing to highlight, on-conflict clause is used only when creating a table typed value (out put of the query expression is table) using query expressions.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/OnConflictClauseNodeContext.java
```diff
@@ -57,12 +65,37 @@ public OnConflictClauseNodeContext() {
...
+                }
+            }
+            String sortText = SortingUtil.genSortText(rank)
+                    + SortingUtil.genSortText(SortingUtil.toRank(context, lsCItem));
+            lsCItem.getCompletionItem().setSortText(sortText);
+        }
+    }
+
+    private boolean isVariableOfErrorType(LSCompletionItem lsCItem) {
+        Predicate<Symbol> symbolPredicate = CommonUtil.getVariableFilterPredicate();
```
malinthar, line 96: How about function call expressions of error type? We haven't considered them here.

malinthar (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/OnConflictClauseNodeContext.java
nadeeshaan: Done

malinthar (approved):
LGTM!

PR #34134, opened 2021-12-06
Fix function names in class symbol hover and hover over new expressions

### LINKED ISSUE #33670: Missing Hover for function initialisation after code completion
**Description:**

Screenshot 2021-11-05 at 21 03 38

Here there was a hovering object with the function definition, with the latest version this is missing.

Also, this hover object does not include default values.

### LINKED ISSUE #33671: Missing details in hover such as function name and stream completion type
**Description:**
Function names are not visible for object hover
Screenshot 2021-11-05 at 21 02 56

Stream completion type is incomplete `sql:Error?` is seen as `sql:?`
Screenshot 2021-11-05 at 21 04 07

IMS94:
## Purpose
$subject

Fixes #33670
Fixes #33671

## Approach
1. Updated `HoverUtil` to provide hover for expressions when a symbol is not found. 
2. Updated hover of class symbols to use function symbols' signature instead of function type symbols' signature.

Screenshot from 2021-12-06 10-14-05

Screenshot from 2021-12-06 10-14-11

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


gayalkuruppu (dismissed):
LGTM

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/hover/HoverUtil.java
```diff
@@ -81,6 +87,16 @@ public static Hover getHover(HoverContext context) {
         // Check for the cancellation after the time consuming operation
         context.checkCancelled();
         if (symbolAtCursor.isEmpty()) {
+            Range nodeRange = new Range(context.getCursorPosition(), context.getCursorPosition());
+            NonTerminalNode nodeAtCursor = CommonUtil.findNode(nodeRange, srcFile.get().syntaxTree());
+            if (nodeAtCursor != null) {
+                MatchedExpressionNodeResolver expressionResolver = new MatchedExpressionNodeResolver(nodeAtCursor);
```
nadeeshaan, line 93: I see that we are reusing the Code Action's node finder util here. Shall we make a note to introduce a common API/ visitor in the future?

nadeeshaan (dismissed):
LGTM

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/hover/HoverUtil.java
```diff
@@ -120,6 +136,52 @@ private static Hover getHoverForSymbol(Symbol symbol, HoverContext context) {
...
 
+    /**
+     * Get hover for expression nodes. Note that we are supplying hover for a selected set of expressions.
+     *
+     * @param context  Context
+     * @param exprNode Expression node
+     * @return Hover
+     */
+    private static Hover getHoverForExpression(HoverContext context, Node exprNode) {
+        switch (exprNode.kind()) {
```
malinthar, line 147: missing default case

IMS94: I did not add a default case to keep the code simple. If we add a default case, we will have to return the default hover object in the default case and at the end of the method both. Wdyt?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/hover/HoverUtil.java
IMS94: I think this visitor is useful for us for use cases like this. Rather than checking each expression kind (via syntax kind), this is clean. Is it ok if we use it everywhere?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/hover/HoverUtil.java
malinthar: We can leave the default case empty.(`default:`) I thought it is best to add it

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/hover/HoverUtil.java
nadeeshaan: The best option would be now, to take this out to a common place. Please create a separate task to refactor this.

nadeeshaan (approved):
LGTM

Issue #34081, opened 2021-12-01
`bal build` on the Simple Query BBE for the second time results in compilation failure


azinneera:
**Description:**
$subject with the following output:
```
Compiling source
	test.bal
error: failed to load the module 'ballerina/sql:1.1.0' from its BIR
```

**Steps to reproduce:**
1. Copy the [Simple Query BBE](https://ballerina.io/learn/by-example/jdbc-query-operation.html?is_ref_by_example=true) to a bal file and execute `bal build` command - builds successfully and generates the executable
2. Rerun `bal build` command - compilation failure

**Affected Versions:**
SL Beta.6-SNAPSHOT (51934f8fa305af61241470f7dd4a9de1fafeb60b)

azinneera:
When debugged the following NPE was seen as the cause of the BIR load failure
```
java.lang.NullPointerException
	at org.wso2.ballerinalang.compiler.semantics.model.types.BTypeReferenceType.isNullable(BTypeReferenceType.java:64)
	at org.wso2.ballerinalang.compiler.semantics.model.types.BUnionType.add(BUnionType.java:272)
	at org.wso2.ballerinalang.compiler.BIRPackageSymbolEnter$BIRTypeReader.readType(BIRPackageSymbolEnter.java:1274)
	at org.wso2.ballerinalang.compiler.BIRPackageSymbolEnter.readBType(BIRPackageSymbolEnter.java:599)
	at org.wso2.ballerinalang.compiler.BIRPackageSymbolEnter$BIRTypeReader.readTypeFromCp(BIRPackageSymbolEnter.java:1014)
	at org.wso2.ballerinalang.compiler.BIRPackageSymbolEnter$BIRTypeReader.readType(BIRPackageSymbolEnter.java:1461)
	at org.wso2.ballerinalang.compiler.BIRPackageSymbolEnter.readBType(BIRPackageSymbolEnter.java:599)
	at org.wso2.ballerinalang.compiler.BIRPackageSymbolEnter$BIRTypeReader.readTypeFromCp(BIRPackageSymbolEnter.java:1014)
	at org.wso2.ballerinalang.compiler.BIRPackageSymbolEnter$BIRTypeReader.readType(BIRPackageSymbolEnter.java:1215)
	at org.wso2.ballerinalang.compiler.BIRPackageSymbolEnter.readBType(BIRPackageSymbolEnter.java:599)
[...]

PR #34137, opened 2021-12-06
Fix NPE in BIR symbol enter

MaryamZi:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/34081

**Remarks:**
https://github.com/ballerina-platform/ballerina-lang/issues/34138

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
@@ -82,6 +82,7 @@ import org.wso2.ballerinalang.compiler.semantics.model.types.BTypeReferenceType;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BTypedescType;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BUnionType;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BXMLType;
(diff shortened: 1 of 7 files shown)


pubudu91:
Why would the `referredType` be `null`? 🤔  Ideally, it should never be `null` right? 🤔 Because we create type reference types for references of type defs and there's always a typedesc in a type def.

MaryamZi:
> Why would the `referredType` be `null`? thinking Ideally, it should never be `null` right? thinking Because we create type reference types for references of type defs and there's always a typedesc in a type def.

This happens with something like 

```ballerina
public type Foo object {
    Bar b;    
};

public type Bar object {
    Baz b;
};

type Baz record {
    Foo|Bar c;
};
```

While resolving the type definition `Foo`, we try to resolve its fields which tries to resolve `Bar` and then goes back to `Foo` in `Foo|Bar` and tries to do the `isNullable` check. At this point since the referredType is null (since `Foo` is still being resolved), it results in an NPE. Bar` and `Baz` aren't defined yet.

I think we need to revisit ordering/how the types are resolved to fix this properly. Created an issue (https://github.com/ballerina-platform/ballerina-lang/issues/34138) also.

hasithaa (approved):

PR #29282, opened 2021-03-17
Validate expr with check keyword in call stmt

chiranSachintha:
Fix the issue mentioned in https://github.com/ballerina-platform/ballerina-lang/issues/28769#issuecomment-791130489 also.

Fixes #28769 

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #34180, opened 2021-12-07
Improve `call-stmt` recovery for partially written `method-call-expr`

### LINKED ISSUE #34104: VSCode suggestions for streams did not appear with check statement
**Description:**
When I tried to iterate a stream, the following suggestions were given correctly when I only used stream without `check`.

Screenshot 2021-12-02 at 14 36 16

However, when I tried the same with a `check` statement, I didn’t give the previously given suggestions.

Screenshot 2021-12-02 at 14 36 38

**Steps to reproduce:**

Follow the aforementioned steps.

**Affected Versions:**
- Ballerina Swan Lake Beta 05
- VSCode Ballerina plugin `v2.1.1`
- VSCode version: 1.62.3
- macOS

lochana-chathura:
## Purpose
$subject.

Fixes #34104 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -8787,23 +8787,32 @@ private STNode validateCallExpression(STNode callExpr) {
...
+            funcOrMethodCall = STNodeFactory.createMethodCallExpressionNode(fieldAccessExpr.expression,
+                    fieldAccessExpr.dotToken, fieldAccessExpr.fieldName, openParenToken, arguments, closeParenToken);
+            funcOrMethodCall = SyntaxErrors.addDiagnostic(funcOrMethodCall,
+                    DiagnosticErrorCode.ERROR_INVALID_EXPRESSION_EXPECTED_CALL_EXPRESSION);
+        } else if (expr.kind == SyntaxKind.SIMPLE_NAME_REFERENCE || expr.kind == SyntaxKind.QUALIFIED_NAME_REFERENCE) {
+            STNode funcName = SyntaxErrors.addDiagnostic(expr,
                     DiagnosticErrorCode.ERROR_INVALID_EXPRESSION_EXPECTED_CALL_EXPRESSION);
+            funcOrMethodCall = STNodeFactory.createFunctionCallExpressionNode(funcName, openParenToken, arguments,
+                    closeParenToken);
         } else {
```
rdulmina, line 8806: There can be other cases that we have to special case like this if we r following this method. Instead of this can we create `INVALID_EXPRESSION_STATEMENT` for invalid call statements which are valid check expressions so that it will also provide `Create variable assignment` code action?

malinthar (commented):
LGTM!

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
lochana-chathura: IIRC, the reason we limited `call-stmt` to `function-call` and `method-call` was to avoid complications at compiler phases. (Introduced with #29282) So the tree has to have either a function-call or method-call node. Looping @chiranSachintha 

On the other hand if we consider `a()`, `a:b()`, `b.c()` only simple-name-ref, qualified-name-ref and field-access can be in the partially written state. 

For the create variable code action with `check-expr` in other `expr` scenarios we may have to consider this diagnostics separately.

lochana-chathura: Created this issue #34200

rdulmina (approved):

PR #34189, opened 2021-12-07
Improve performance of map langlib functions

### LINKED ISSUE #34128: Reducing a map performance worsens as the map gets bigger
**Description:**
Reducing a map (i.e. `(map<int>).reduce()`) decreases in performance as the size of map increases.

**Steps to reproduce:**

Build and run the following:

```
    map<int> m = {};

    foreach int i in 0 ..< 60000 {
        m[i.toString()] = i;
    }

    int sum = m.reduce(function (int carry, int value) returns int {
        return carry + value;
    }, 0);

    io:println(sum);
}
```

On my machine (Intel Core i7-1165G7) running this takes 20.44 seconds (as measured with `time java -jar demo.jar`)

For comparison, consider the following equivalent:
```
    map<int> m = {};

    foreach int i in 0 ..< 60000 {
        m[i.toString()] = i;
    }

    int sum = 0;
    foreach int j in m {
        sum += j;
    }
    
    io:println(sum);
}
```

This runs in 1.78 seconds (again, measured with `time java -jar demo.jar`), ~ 1 tenth of the `.map` speed.

The behaviour gets increasingly worse as 60000 is increased to higher values.
For example, with 100000 it takes 164.08 seconds, while the `foreach` variant requires only 2.21 seconds.

**Affected Versions:**

Swan Lake Beta 5

Linux: Fedora 34 (Linux fedora 5.15.5-100.fc34.x86_64 #1 SMP Fri Nov 26 00:52:21 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux)

openjdk 11.0.13 2021-10-19
OpenJDK Runtime Environment 18.9 (build 11.0.13+8)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.13+8, mixed mode, sharing)

N/A

N/A

N/A

HindujaB:
## Purpose
$subject 
Fixes #34128 

## Approach
The map langlib implementation uses `map.getKeys()` API to retrieve keys of a map. In most of the cases, this is called   per iteration of the map members, which is redundant. 
The `getKeys()` method contains the `Collections.toArray()` java method call which causes the performance impact.
This PR fixes it by removing the `getKeys()` call for each map member and improve the `getKeys()` implementation. 

CPU profiling of the sample code, shows the following hotspots.

performance

## Samples
```ballerina
import ballerina/io;

    map<int> m = {};

    foreach int i in 0 ..< 60000 {
        m[i.toString()] = i;
    }

    int sum = m.reduce(function(int carry, int value) returns int {
        return carry + value;
    }, 0);

    io:println(sum);
}
```
## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


HindujaB:
The results of runtime performance tests :
| Method    | Beta 5 - runtime (s) | After fix - runtime (s) |
|-----------|----------------------|-------------------------|
| reduce()  |               63.643 |                   0.697 |
| filter()  |                57.38 |                   0.684 |
| foreach() |               48.597 |                   0.669 |
| map()     |               96.004 |                   0.667 |

rpkamp:
FWIW it seems this altered the behavior of the methods changed. 

Before the change one could push elements onto whatever you were looping and they would be considered too. Now only the keys that were there when the loop began are considered. 

Not sure if that's good or bad, but it is different.

PR #34192, opened 2021-12-07
Fix handling integer overflows in `lang.int:sum` function

### LINKED ISSUE #32603: Inconsistent behavior with  `lang.int:sum`  when the sum leads to an integer overflow
**Description:**
$title

**Steps to reproduce:**
```ballerina
import ballerina/io;
public function main() {
    int a1 = -9223372036854775805;

    int a2 = a1.sum(-1, -3);
    io:println(a2); // prints 9223372036854775807

    int a3 = -9223372036854775805 + (-1) + (-3); // panics with int range overflow
}
```
```ballerina
import ballerina/io;
public function main() {
    int a1 = 9223372036854775805;

    int a2 = a1.sum(1, 2);
    io:println(a2); // prints -9223372036854775808

    a2 = 9223372036854775805 + 1 + 2; // panics with int range overflow
}
```
**Affected Versions:**
slbeta2

lasinicl:
## Purpose
$title

Fixes #32603

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/langlib-test/src/test/resources/test-src/intlib_test.bal
```diff
@@ -123,3 +123,113 @@ function testIntOverflow() {
     test:assertValueEqual("{ballerina/lang.int}NumberOverflow", err.message());
     test:assertValueEqual("int range overflow", <string>checkpanic err.detail()["message"]);
 }
+
+function testIntOverflowWithSum() {
+    int a1 = int:MIN_VALUE;
+    int|error result = trap a1.sum(-1);
```
HindujaB, line 129: Since, `a1.sum()` and `int:sum()` calls to the same internal function, checking for both cases for this fix can be redundant. Shall we have only one of these?

lasinicl: But it's two ways calling the same function right? I believe it's okay to have some tests with both.

HindujaB: Yeah. It's okay to have them both. +1
My concern is this fix is specific to the method implementation and the way of calling it does not get impacted.

lasinicl: We don't really have many tests covering the behavior of lang.int functions. That's why I decided to add tests for both. :)

langlib/lang.int/src/main/java/org/ballerinalang/langlib/integer/Sum.java
```diff
@@ -31,9 +39,14 @@
...
-        for (int i = 0; i < size; i++) {
-            sum += ns[i];
+        for (long n : ns) {
+            try {
+                sum = Math.addExact(sum, n);
+            } catch (ArithmeticException e) {
+                throw ErrorCreator.createError(getModulePrefixedReason(INT_LANG_LIB,
+                        BallerinaErrorReasons.NUMBER_OVERFLOW_ERROR_IDENTIFIER),
+                        BLangExceptionHelper.getErrorDetails(RuntimeErrors.INT_RANGE_OVERFLOW_ERROR));
+            }
```
MaryamZi, line 49: We are doing this instead of calling `io.ballerina.runtime.internal.MathUtils#addExact` because we need to use a different error message (previously reason), right? Shall we add a comment mentioning this?

Alternatively, wondering if we should introduce overloaded methods in `MathUtils` and `ErrorUtils` to accept the error message.

@warunalakshitha, thoughts please?

My concern is that this should be the same as the additive expression. While it's highly unlikely that the addition logic would change, it would be better to have the logic in a single place IMO.

langlib/langlib-test/src/test/resources/test-src/intlib_test.bal
MaryamZi: IMO when adding tests for langlib methods it is better to add tests for both ways of calling them. Sometimes we may come across issues with how the first argument is propagated/analyzed.

langlib/lang.int/src/main/java/org/ballerinalang/langlib/integer/Sum.java
warunalakshitha: +1

warunalakshitha: Also, check whether there are any other lang lib methods, that have duplicate logic. I think core logic should be inside runtime classes not in lang lib.

PR #34239, opened 2021-12-09
Fix `annotations()` API of `WorkerSymbol`

### LINKED ISSUE #27461: [Semantic API] annotations() not working for workers and type defs
The `annotations()` API currently returns empty lists for type defs and workers. The issue in supporting this for type defs is that we don't have an internal symbol for type defs. Therefore, if we are adding annotation symbols, we'll have to add them to the internal type symbol itself. This may not be desirable for case like `type Foo int;`.

The issue with workers is that the worker gets desugared right at the start. The required info for the worker symbol is therefore, spread across two symbols: the function symbol created for the lambda and the var symbol created for the var which holds the future for the start action for the lambda.

pubudu91:
## Purpose
The `annotations()` API of `WorkerSymbol` was broken due to the way we model a worker internally. This PR addresses this by introducing a new internal symbol for workers which also keeps track of the function symbol of the generated function for workers.

Fixes #27461 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
@@ -19,6 +19,7 @@ package org.wso2.ballerinalang.compiler.semantics.analyzer;
 
 import io.ballerina.compiler.api.symbols.DiagnosticState;
 import io.ballerina.tools.diagnostics.Location;
+import io.ballerina.tools.text.LineRange;
 import org.ballerinalang.compiler.CompilerOptionName;
 import org.ballerinalang.compiler.CompilerPhase;
 import org.ballerinalang.model.TreeBuilder;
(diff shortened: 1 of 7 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -5031,6 +5055,22 @@ private boolean isReadOnlyAndObjectIntersection(BIntersectionType referredType)
         return false;
     }
 
+    private boolean withinNode(LineRange srcRange, LineRange targetRange) {
```
dulajdilshan, line 5058: According to the implementation, the duty of this function is to compare ranges. IMO, the name should be `withinRange`.

dulajdilshan (commented):
Rest is LGTM

dulajdilshan (approved):
LGTM

PR #34251, opened 2021-12-10
Fix array conversion without filler values

### LINKED ISSUE #12838: constructFrom langlib method doesn't handle closed array properly.
### Description:

**Tuple/open array to closed array conversion doesn't look at the array size**
```ballerina
public function main() {
    (int, int) intTuple = (1, 2);
    int[3]|error valueArray = int[3].convert(intTuple);
}
```

```ballerina
public function main() {
    int[] intArr = [1, 2];
    int[3]|error valueArray = int[3].convert(intArr);
}
```

Currently the above conversions do not fail. These should fail, since the minimum number of elements required is not present in the source type.

HindujaB:
## Purpose
$subject
Fixes #12838

## Approach
Added a validation for invalid tuple to array conversion for larger indexes.
 
## Samples
```ballerina
    [TypeWithNoFillerValue, TypeWithNoFillerValue] tuple1 = [1, "foo"];
    TypeWithNoFillerValue[5] arr1 = check tuple1.cloneWithType(); // should panic
    io:println(arr1);
}

type TypeWithNoFillerValue 1|"foo";
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -551,16 +551,27 @@ private static boolean isConvertibleToArrayType(Object sourceValue, BArrayType t
...
+            Type sourceElementType = ((BArrayType) sourceType).getElementType();
             if (isNumericType(sourceElementType) && isNumericType(targetTypeElementType)) {
                 return true;
             }
         }
+        int targetSize = targetType.getSize();
+        long sourceSize = source.getLength();
+        boolean returnVal = true;
+        if (!TypeChecker.hasFillerValue(sourceType) && sourceSize < targetSize) {
+            addErrorMessage(0, errors, "array size '" + targetSize + "' is out of range for source type'" +
```
warunalakshitha, line 565: Shall we improve the error that source type does not have filler values as well.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -551,16 +551,27 @@ private static boolean isConvertibleToArrayType(Object sourceValue, BArrayType t
...
                 return true;
             }
         }
+        int targetSize = targetType.getSize();
+        long sourceSize = source.getLength();
+        boolean returnVal = true;
+        if (!TypeChecker.hasFillerValue(targetType) && sourceSize < targetSize) {
+            addErrorMessage(0, errors, "array cannot be expanded to size '" + targetSize + "' for target " +
+                    "type '" + targetType + "' without filler values");
+            returnVal = false;
```
Nadeeshan96, line 567: Because this will pass
```bal
public function main() {
    [int, int] intTuple = [1, 2];
    int[3] valueArray = checkpanic intTuple.cloneWithType();
}
```
and as the issue here is that the targetType not having filler values, I was thinking it is better to emphasise that by saying something like `array cannot be expanded to size '" + targetSize + "' because the target type '" + targetType + "' does not have a filler value."`

Another thing that needed to be fixed in this issue is in some cases the obtained filler values are not correct, according to the spec https://ballerina.io/spec/lang/master/#FillMember.
For example
```bal
public function main() {
    [string, string] tuple = ["aaa", "bbb"];
    string[3] arr = checkpanic tuple.cloneWithType(); // here arr[2] will become () where it should be ""
}
```

warunalakshitha (dismissed):
LGTM

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -551,16 +551,27 @@ private static boolean isConvertibleToArrayType(Object sourceValue, BArrayType t
...
+        long sourceSize = source.getLength();
+        boolean returnVal = true;
+        if (!TypeChecker.hasFillerValue(targetType) && sourceSize < targetSize) {
+            addErrorMessage(0, errors, "array cannot be expanded to size '" + targetSize + "' because, the target " +
+                    "type '" + targetType + "' does not have a filler value");
+            returnVal = false;
+            if (errors.size() >= MAX_CONVERSION_ERROR_COUNT + 1) {
+                return false;
+            }
+        }
```
Nadeeshan96, line 571:
```suggestion
        if (!TypeChecker.hasFillerValue(targetType) && sourceSize < targetSize) {
            addErrorMessage(0, errors, "array cannot be expanded to size '" + targetSize + "' because, the target " +
                    "type '" + targetType + "' does not have a filler value");
            return false;
        }
```
I think we can return here because going through each element and checking whether is it convertible is redundant here.

warunalakshitha (approved):

PR #34272, opened 2021-12-10
Introduce ResetStream error

chamil321:


anupama-pathirage:
@chamil321 / @shafreenAnfar Do we need this PR? Looks like it got automatically closed without any attention. If you really don't need this PR anymore, please remove the `stale` label from the PR. Then it won't get added to the `[Stale PR] Ballerina Stale PR Details` mail in the future.

PR #34287, opened 2021-12-13
Fix not retrieving class method members in `visibleSymbols()`

### LINKED ISSUE #33492: [Semantic API] Need to define the `visibleSymbols()` behaviour for init() method
Consider the following:
```ballerina
function testObjectConstructor() {
    int y = 0;

    var obj = object {
        string name = "";
        int age;

        function init() {
            self.age = 0;
            var s = self.getName(); 
        }

        function getName() returns string {
            return self.name;
        }
    };
}
```
In the init() method, the getName() method is not visible. But init() is visible in getName(). Need to evaluate the behaviours for methods in object constructors. And should evaluate and see classes too.

dulajdilshan:
## Purpose
$title

Fixes #33492

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


dulajdilshan:
Here, we need to handle several cases separately, hence, converting this to a draft PR for now.
Case 1: Cursor at a variable declaration?
```ballerina
class Vehicle {
	public string brand = <cursor>

	function getBrand() returns string {
		return "Toyota";
	}
}
```

Case 2: Cursor at within a function-body block
```ballerina
class Vehicle {
	public string brand = "Toyota";

	function getName returns string {
		string name = <cursor> 
	}

	function getBrand() returns string {
		return "Toyota";
	}
}
```

In Case 1, we shouldn't show members as visible-symbols

tests/ballerina-compiler-api-test/src/test/resources/testng.xml
```diff
@@ -161,6 +161,7 @@
 
             <!--Visible Symbols-->
             <class name="io.ballerina.semantic.api.test.visiblesymbols.VisibleSymbolsInCompoundStatementsTest" />
+            <class name="io.ballerina.semantic.api.test.visiblesymbols.VisibleSymbolsInClassesAndObjectsTest" />
```
pubudu91, line 164: Shall we reorder this alphabetically? The rest seems ok

mohanvive (dismissed):

dulajdilshan:
Working on remove completions in the following cases
Case 1:
```ballerina 
class Vehicle {
	public string brand = <cursor>

	function getBrand() returns string {
		return "Toyota";
	}
}
```
Case 2:
```ballerina
service ProcessingService / on lsn {

    public string magic = "The Somebody Else's Problem field";

    resource function get processRequest() returns json => { output: self.magic };

    function createError() returns @tainted error? => <cursor>;

    resource function get [int... rest] () returns string => "foo";
}
``

mohanvive (approved):

PR #34319, opened 2021-12-14
Persist the Type Inclusion Information for Record/Object Type

### LINKED ISSUE #33520: [Semantic API] Type inclusion info lost in the signature of a record/object type symbol
In the signature generated through the `signature()` method, the type inclusion info is lost since we don't really consider that when modelling the internal type. e.g.,
```
type Foo object {
    string name;
};

object {
    *Foo;
    int z;
}
```
If we print the signature of the above object typedesc, it'd print `object {string name; int z;}`. While semantically that's valid, this is not the same thing that the user wrote in the code. Need to see if we need to preserve the type inclusion details in the signature.

sanjana:
Fixes #33520

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaObjectTypeSymbol.java
```diff
@@ -148,6 +196,33 @@ public BallerinaObjectTypeSymbol(CompilerContext context, ModuleID moduleID, BOb
         return this.typeInclusions;
     }
 
+    public List<TypeSymbol> originalTypeInclusions() {
```
pubudu91, line 199: Why do we need such a method? There's no distinction to be made in type inclusions.

pubudu91, line 199: Why do we need such a method? There's no distinction to be made in type inclusions.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaObjectTypeSymbol.java
```diff
@@ -162,19 +237,21 @@ public String signature() {
...
-        // this.getObjectTypeReference()
-        //         .ifPresent(typeDescriptor -> fieldJoiner.add("*" + typeDescriptor.getSignature()));
-        for (ObjectFieldSymbol objectFieldDescriptor : this.fieldDescriptors().values()) {
-            fieldJoiner.add(objectFieldDescriptor.signature() + ";");
+        for (TypeSymbol typeInclusion : this.originalTypeInclusions()) {
+            fieldJoiner.add("*" + typeInclusion.signature() + ";");
+        }
+
+        for (ObjectFieldSymbol objectFieldSymbol : this.originalFieldDescriptors().values()) {
+            fieldJoiner.add(objectFieldSymbol.signature() + ";");
```
pubudu91, line 245: Shall we get rid of the `+`s and stick to using `add()`?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordTypeSymbol.java
```diff
@@ -67,6 +69,22 @@ public BallerinaRecordTypeSymbol(CompilerContext context, ModuleID moduleID, BRe
         return this.fieldSymbols;
     }
 
+    public Map<String, RecordFieldSymbol> originalFieldDescriptors() {
```
pubudu91, line 72: Same comment as for object fields

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaRecordTypeSymbol.java
```diff
@@ -103,11 +121,44 @@ public BallerinaRecordTypeSymbol(CompilerContext context, ModuleID moduleID, BRe
...
+        return this.originalTypeInclusions;
+    }
+
     @Override
     public String signature() {
         // Treating every record typedesc as exclusive record typedescs.
         StringJoiner joiner = new StringJoiner(" ", "{| ", " |}");
-        for (RecordFieldSymbol fieldSymbol : this.fieldDescriptors().values()) {
+        for (TypeSymbol typeInclusion : this.originalTypeInclusions()) {
+            String ballerinaTypeSignature = "*" + typeInclusion.signature() + ";";
```
pubudu91, line 157: Misleading name. Not a type signature. Alternatively can just simply use a chain of `add()` calls.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/writer/BIRTypeWriter.java
```diff
@@ -365,6 +366,15 @@ public void visit(BRecordType bRecordType) {
...
         }
 
+        buff.writeInt(bRecordType.originalFields.size());
+        for (BField field : bRecordType.originalFields.values()) {
+            BVarSymbol symbol = field.symbol;
+            buff.writeInt(addStringCPEntry(symbol.name.value));
+            buff.writeLong(symbol.flags);
+            writeMarkdownDocAttachment(buff, field.symbol.markdownDocumentation);
+            writeTypeCpIndex(field.type);
+        }
```
pubudu91, line 376: Do we really need to be duplicating everything like this? Should be able to just filter out the user specified fields from the fields list and add them to the cached list of user specified fields.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -7539,6 +7539,7 @@ private BLangClassDefinition desugarTemplateLiteralObjectTypedef(List<BLangLiter
         // Create a new concrete, class type for the provided abstract object type
         BObjectType objectClassType = new BObjectType(classTSymbol, classTSymbol.flags);
         objectClassType.fields = objectType.fields;
+        objectClassType.originalFields = objectType.fields;
```
pubudu91, line 7542: This kind of assignments are problematic since you're setting the same value to both fields. Changes done using `fields` will be visible through `originalFields` as well. Applies to all the other places as well.

pubudu91 (changes requested):
Is the approach taken in this the one we discussed with @gimantha (I can't exactly remember now)? Wouldn't just marking the user specified field symbols do? Why do we need to maintain a separate map for them?

mohanvive:
@sanjana can you please attend the review comments?

mohanvive:
> @sanjana can you please attend the review comments?

Reminder!

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaObjectTypeSymbol.java
gimantha: I don't exactly remember what we discussed, but did we not come to a conclusion that the necessary inclusion type info is already there in symbols, and all we have to do is some filtering to capture them from the semantic API side?

pubudu91:
Let's close this and reopen/resend when looking into this again

dulajdilshan:
@sanjana There are some conflicts and review comments. Can you please look into them?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaObjectTypeSymbol.java
sanjana: Made this a private internal method. Used to identify the user-specified type inclusions separately when generating the `.signature()`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/writer/BIRTypeWriter.java
```diff
@@ -414,6 +425,17 @@ private void writeObjectAndServiceTypes(BObjectType bObjectType) {
...
         }
+
+        buff.writeInt(bObjectType.originalFields.size());
+        for (BField field : bObjectType.originalFields.values()) {
+            buff.writeInt(addStringCPEntry(field.name.value));
+            // TODO add position
+            buff.writeLong(field.symbol.flags);
+            buff.writeBoolean(field.symbol.isDefaultable);
+            writeMarkdownDocAttachment(buff, field.symbol.markdownDocumentation);
+            writeTypeCpIndex(field.type);
```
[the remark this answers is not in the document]
sanjana: Optimized by adding a flag when writing the fields.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/writer/BIRTypeWriter.java
sanjana: Optimized by adding a flag when writing the fields.

sanjana:
> Let's close this and reopen/resend when looking into this again

Closing for now. And will be re-opening once started looking into this again.

PR #34324, opened 2021-12-14
Fix MatchedExpressionNodeResolver returning null

### LINKED ISSUE #34318: NPEs in LS for Invalid Syntax
**Description:**
Consider the following code:

```
type Foo object {
    string name;
};

type A object {
    *Foo;
    int z;
};

A x = object {123, <A>""<B>}
```
Once the `""` within `<A>` and `<B>` is typed, NPEs can be observed in the LS for `TypeCastCodeAction` and `AddCheckCodeAction`. 

**Steps to reproduce:**
NPE-LS-for-object

**Affected Versions:**
SL Beta RC6, [Current Master](https://github.com/ballerina-platform/ballerina-lang/tree/2f32deccc15795a3b0f38e6877e88dc853fcc7af)

IMS94:
## Purpose
$subject

The `apply()` method of `ExternalTreeNodeList` returns `null` causing the `Optional` to be null in `MatchedExpressionNodeResolver` case. Changed the behavior of `MatchedExpressionNodeResolver` to not expose this behavior to the outside.

Fixes #34318

## Approach
See description

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/MatchedExpressionNodeResolver.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/MatchedExpressionNodeResolver.java
@@ -36,6 +36,9 @@ import java.util.Optional;
 /**
  * Node Transformer to find the container expression node for a given node.
  *
+ * <strong>Note</strong>: Rather than doing {@code node.apply(expressionResolver)}, 
+ * please use {@link #findExpression(Node)} since the {@code node.apply()} method may return {@code null}.
+ *
  * @since 2.0.0
  */
 public class MatchedExpressionNodeResolver extends NodeTransformer<Optional<ExpressionNode>> {
@@ -46,6 +49,26 @@ public class MatchedExpressionNodeResolver extends NodeTransformer<Optional<Expr
         this.matchedNode = matchedNode;
     }
 
+    /**
+     * Given the node, this method returns the optional expression in which the provided node is located.
+     *
+     * @param node Node
+     * @return Optional enclosing expression node
+     */
+    public Optional<ExpressionNode> findExpression(Node node) {
+        if (node == null) {
+            return Optional.empty();
+        }
+
+        Optional<ExpressionNode> exprNode = node.apply(this);
+        // Due to the way apply() method is implemented in some cases, this can return null
+        if (exprNode == null) {
+            return Optional.empty();
+        }
+        
+        return exprNode;
+    }
+    
     @Override
     protected Optional<ExpressionNode> transformSyntaxNode(Node node) {
         if (node.parent() == null) {
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/MatchedExpressionNodeResolver.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/MatchedExpressionNodeResolver.java
@@ -62,11 +62,7 @@ public class MatchedExpressionNodeResolver extends NodeTransformer<Optional<Expr
 
         Optional<ExpressionNode> exprNode = node.apply(this);
         // Due to the way apply() method is implemented in some cases, this can return null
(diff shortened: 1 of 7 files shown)


malinthar (dismissed):
LGTM

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/MatchedExpressionNodeResolver.java
```diff
@@ -46,6 +49,26 @@ public MatchedExpressionNodeResolver(Node matchedNode) {
...
+     * @return Optional enclosing expression node
+     */
+    public Optional<ExpressionNode> findExpression(Node node) {
+        if (node == null) {
+            return Optional.empty();
+        }
+
+        Optional<ExpressionNode> exprNode = node.apply(this);
+        // Due to the way apply() method is implemented in some cases, this can return null
+        if (exprNode == null) {
```
nadeeshaan, line 65: Shouldn't we fix this bug in a more higher level? For me this feels like sort of a specific remedy we are trying to provide. This same can lead to the same bug in other places as well. WDYT?

IMS94: The issue is, `apply()` method in `Node` has the following signature. 
```
public abstract <T> T apply(NodeTransformer<T> transformer);
```
As per generics, `Optional` is just another type for `T`. That's why I thought we cannot handle this better. WDYT?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/MatchedExpressionNodeResolver.java
```diff
@@ -46,6 +49,26 @@ public MatchedExpressionNodeResolver(Node matchedNode) {
...
+            return Optional.empty();
+        }
+        
+        return exprNode;
```
sanjana, line 69:
```suggestion
        return exprNode.isPresent()? exprNode : Optional.empty();
```

Is this further simplification possible?

sanjana (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/MatchedExpressionNodeResolver.java
malinthar: This might generate NPE when exprNode is null

sanjana: Can use `return exprNode == null?  Optional.empty() : exprNode;` instead?

malinthar: +1

IMS94: +1. Done!

sanjana (approved):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/MatchedExpressionNodeResolver.java
nadeeshaan: Given the implementation, that makes sense. But I think, if we call the accept method inside the at the particular node without returning null, this could have been fixed. But, yeah let's proceed with the fix for this time and let's make sure to check for the proper fix if possible as well.

nadeeshaan (approved):
LGTM

PR #34360, opened 2021-12-16
Update version with semver

udda1996:
## Purpose
Update version with SemVer and remove short version

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,10 +2,9 @@ org.gradle.caching=true
 org.gradle.parallel=true
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 org.gradle.workers.max=3
-version=2.0.0-beta.7-SNAPSHOT
+version=-2201.0.0-SNAPSHOT
 group=org.ballerinalang
 bootstrappedOn=1.1.0-alpha
-shortVersion=slbeta7
 versionDisplayText=Beta 7
 specVersion=2021R1
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,10 +2,9 @@ org.gradle.caching=true
 org.gradle.parallel=true
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 org.gradle.workers.max=3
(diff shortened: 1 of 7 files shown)


keizer619 (dismissed):

gradle.properties
```diff
@@ -2,10 +2,9 @@ org.gradle.caching=true
 org.gradle.parallel=true
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 org.gradle.workers.max=3
-version=2.0.0-beta.7-SNAPSHOT
+version=-2201.0.0-SNAPSHOT
 group=org.ballerinalang
 bootstrappedOn=1.1.0-alpha
-shortVersion=slbeta7
 versionDisplayText=Beta 7
```
pubudu91, line 8: What about this one?

gradle.properties
```diff
@@ -2,11 +2,9 @@ org.gradle.caching=true
 org.gradle.parallel=true
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 org.gradle.workers.max=3
-version=2.0.0-beta.7-SNAPSHOT
+version=-2201.0.0-SNAPSHOT
```
azinneera, line 5: Why do we have a hyphen (`-`) at the beginning of the version?

azinneera:
The failing test ` io.ballerina.projects.FileSystemRepositoryValidationTests.testGetPackageVersions` can be fixed only after the version update. Please disable the test and go ahead. Will send the fix and enable the test afterwards.

udda1996:
> The failing test ` io.ballerina.projects.FileSystemRepositoryValidationTests.testGetPackageVersions` can be fixed only after the version update. Please disable the test and go ahead. Will send the fix and enable the test afterwards.

ACK. Will disable the test

udda1996:
Screenshot 2021-12-17 at 12 04 45
This is the `bal -v ` output with this fix.

udda1996:
This is how the "bal -v" output looks with this fix.
Screenshot 2021-12-17 at 12 16 11

keizer619 (approved):

PR #34364, opened 2021-12-16
Remove unnecessary info from JSON syntax tree 

kaviththiranga:
In an effort of reducing the JSON size for a
given Syntax Tree, we are removing unnecessary
information from the JSON.


kaviththiranga:
Working on fixing tests

PR #34365, opened 2021-12-16
Remove duplicates from JarList

### LINKED ISSUE #34357: Fix repetitive JAR loading in Ballerina running and testing phases
**Description:**
It seems when we have the same JAR in two different stdlibs, Ballerina running and testing phases repetitively load both the JARs and due to that there could be JAR resolution errors like [#2492](https://github.com/ballerina-platform/ballerina-standard-library/issues/2492).

Ibaqu:
## Purpose
Remove duplicates from JarList for building and testing

Fixes #34357 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/JarResolver.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/JarResolver.java
@@ -131,8 +131,14 @@ public class JarResolver {
         // Add all the jar library dependencies of current package (packageId)
         Collection<PlatformLibrary> otherJarDependencies = jBalBackend.platformLibraryDependencies(
                 packageContext.packageId(), scope);
+
+        List<String> fileNames = new ArrayList();
+        libraryPaths.stream().forEach(e -> fileNames.add(e.path().toFile().getName()));
+
         for (PlatformLibrary otherJarDependency : otherJarDependencies) {
-            libraryPaths.add(new JarLibrary(otherJarDependency.path(), scope, getPackageName(packageContext)));
+            if (!fileNames.contains(otherJarDependency.path().toFile().getName())) {
+                libraryPaths.add(new JarLibrary(otherJarDependency.path(), scope, getPackageName(packageContext)));
+            }
         }
     }
 


compiler/ballerina-lang/src/main/java/io/ballerina/projects/JarResolver.java
```diff
@@ -131,8 +131,14 @@ private void addPlatformLibraryPaths(PackageContext packageContext,
         // Add all the jar library dependencies of current package (packageId)
         Collection<PlatformLibrary> otherJarDependencies = jBalBackend.platformLibraryDependencies(
                 packageContext.packageId(), scope);
+
+        List<String> fileNames = new ArrayList();
+        libraryPaths.stream().forEach(e -> fileNames.add(e.path().toFile().getName()));
```
azinneera, line 136: AFAIU, this filter will not work as intended. The requirement is to stop adding a jar if a path with the same filename already exists in the jarList.

azinneera: This seems to be correct. Lets test and verify

azinneera (approved):

Ibaqu:
We tested the fix at 2 states in GRPC : at the latest commit and as well as the commit where the issue was first found (95de4d) 

For the latest commit, the netty changes were reverted (https://github.com/BuddhiWathsala/module-ballerina-grpc/commit/124c0e54bad05e76853e970b8eedec00ecbed0ed) and the fix was tested with on the latest stage branch. It should be noted that the ballerina-parser version 2.0.0-beta.6 was to be used to make sure the tests ran. 

For the commit (95de4d), the HTTP module was using an old tcnative-boringssl jar which had to be updated for the tests to run.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/JarResolver.java
```diff
@@ -131,8 +131,14 @@ private void addPlatformLibraryPaths(PackageContext packageContext,
         // Add all the jar library dependencies of current package (packageId)
         Collection<PlatformLibrary> otherJarDependencies = jBalBackend.platformLibraryDependencies(
                 packageContext.packageId(), scope);
+
```
sameerajayasoma, line 134: IMO, the better solution would be to make a `jarFiles` a `Set` and override the `equals` method of `JarLibrary`

azinneera: @sameerajayasoma is this ok to be addressed in a PR later? RC1 is waiting on this.

azinneera: @sameerajayasoma would it be fine to address this in a PR later. RC1 is waiting on this PR.

PR #34399, opened 2021-12-18
Fix BIR tests

MaryamZi:
## Purpose
$title.

### DIFF
--- a/docs/bir-spec/src/main/resources/kaitai/bir.ksy
+++ b/docs/bir-spec/src/main/resources/kaitai/bir.ksy
@@ -359,7 +359,7 @@ types:
         type: s4
         repeat: expr
         repeat-expr: constituent_types_count
-      - id: effective_type_count
+      - id: effective_type_cp_index
         type: s4
   type_xml:
     seq:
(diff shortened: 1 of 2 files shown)


docs/bir-spec/src/main/resources/kaitai/bir.ksy
```diff
@@ -359,7 +359,7 @@ types:
         type: s4
         repeat: expr
         repeat-expr: constituent_types_count
-      - id: effective_type_count
+      - id: effective_type_cp_index
```
MaryamZi, line 362: This was incorrect previously.

docs/bir-spec/src/main/resources/kaitai/bir.ksy
```diff
@@ -645,7 +645,7 @@ types:
             'type_tag_enum::type_tag_decimal': decimal_constant_info
             'type_tag_enum::type_tag_boolean': boolean_constant_info
             'type_tag_enum::type_tag_nil': nil_constant_info
-            'type_tag_enum::type_tag_map': map_constant_info
+            'type_tag_enum::type_tag_intersection': map_constant_info
```
MaryamZi, line 648: At the moment, if the type tag is intersection, the value is definitely a mapping value, so this will work for the time-being. But ideally we should look at the effective type and select what to do. We may have to change the BIR instead to use the effective type when writing the constant.

hasithaa (approved):

PR #34402, opened 2021-12-18
Fix debugger ExpressionEvaluationTest

MaryamZi:
## Purpose
$title.

Verified using VSCode also.

Where the source is

```ballerina
const map<int> M = {a: 1};

record {int i;} x = {i: 1};
```

Previously we got
```cmd
x
record {| int i; anydata...; |}
i:1
M
map (size = 1)
a:1
```

Now it is

```cmd
x
record {| int i; anydata...; |}
i:1
M
record {| readonly 1 a; |} & readonly
a:1
```

### DIFF
--- a/tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/evaluation/ExpressionEvaluationTest.java
+++ b/tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/evaluation/ExpressionEvaluationTest.java
@@ -172,7 +172,9 @@ public class ExpressionEvaluationTest extends ExpressionEvaluationBaseTest {
 
         debugTestRunner.assertExpression(context, "nameWithType", "\"Ballerina\"", "string");
         debugTestRunner.assertExpression(context, "nameWithoutType", "\"Ballerina\"", "string");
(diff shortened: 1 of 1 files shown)


tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/evaluation/ExpressionEvaluationTest.java
```diff
@@ -172,7 +172,9 @@ public void variableReferenceEvaluationTest() throws BallerinaTestException {
 
         debugTestRunner.assertExpression(context, "nameWithType", "\"Ballerina\"", "string");
         debugTestRunner.assertExpression(context, "nameWithoutType", "\"Ballerina\"", "string");
-        debugTestRunner.assertExpression(context, GLOBAL_VAR_03, "map (size = 1)", "map");
+        debugTestRunner.assertExpression(context, GLOBAL_VAR_03,
+                                         "(debug_test_resources/evaluation_tests:0:$anonType$_10 & readonly)",
```
MaryamZi, line 176: We should ideally be the shape. We may also not be able to rely on the number suffix being the same.

gimantha (approved):

Issue #34307, opened 2021-12-14
Type not narrowed as expected when nested and else if blocks don't complete normally


MaryamZi:
**Description:**
$title.

**Steps to reproduce:**

```ballerina
public function main() {
    int|boolean|string|() x = ();

    if x is int|boolean {
        if x is int {
            return;
        }
    } else {
        if x is string {
            return;
        }
    }

    boolean? _ = x; // error: incompatible types: expected 'boolean?', found '(int|boolean|string)?'
}
```

```ballerina
public function main() {
    int|boolean|string|() x = ();

    if x is int {
        return;
    } else if x is string { // Works if just `if`.
        return;
    }

    boolean? _ = x; // error: incompatible types: expected 'boolean?', found '(int|boolean|string)?'
}
```

```ballerina
public function main() {
    int|boolean|float x = 0.1;

    if x is int|boolean {
        if x is int {
            return;
        }
        boolean b = x;
    }

    boolean|float _ = x; // error: incompatible types: expected '(boolean|float)', found '(int|boolean|float)'
}
```

**Affected Versions:**
slbeta6

pcnfernando:
We should not narrow 
```ballerina
public function dummyTest() {
      int? a = 10;
      string? b = "100";
      if a is int {
          if (a == 5) {
              return;
          }
      }
      b = a;
}
```

lasinicl:
> We should not narrow
> 
> ```ballerina
> public function dummyTest() {
>       int? a = 10;
>       string? b = "100";
>       if a is int {
>           if (a == 5) {
>               return;
>           }
>       }
>       b = a;
> }
> ```

Type is not being narrowed right? Checked with beta6. 
```ballerina
public function dummyTest() {
      int? a = 10;
      string? b = "100";
      if a is int {
          if (a == 5) {
              return;
          }
      }
      b = a; // incompatible types: expected 'string?', found 'int?'
}
```

pcnfernando:
> Yes. I didn't get the incompatible type error. Seems related to https://github.com/ballerina-platform/ballerina-lang/pull/34449 change. Please ignore.

pcnfernando:
When wrapped with a do; doesn't get narrowed
```ballerina
public function dummyTest() {
    int? a = 10;
    string? b = "100";
    do {
        if a is int {
            return;
        }
    }

    b = a; // incompatible types: expected 'string?', found 'int?'
}
```

lochana-chathura:
[...]

PR #34449, opened 2021-12-22
Fix issue with associating symbols being type narrowed with envs 

### LINKED ISSUE #34366: Wrong type of the variable showed when hovering
**Description:**
Peek 2021-12-16 16-11

**Steps to reproduce:**
Hover the cursor to the `b` in the line `20 c = b` in the following ballerina code
```ballerina
function testReachableCodeWithBinaryCondition4() returns int {
    10|20 b = 10;

    if b == 10 {
        return 10;
    }

    20 c = b;

    if b  == 20 {
        return 20;
    }
}
```

**Affected Versions:**
Swan Lake beta 6

### LINKED ISSUE #34369: Local variables don't get defined under the correct environment
**Description:**
```ballerina
function testRun(string name) returns int {
    if !(name is "") {
        return 1;
    }

    string testvar = "toyota";
   <cursor>
}
```
Try getting the completions at the `<cursor>` position.

Here `testvar` will be visible if we remove `return 1` line

**Steps to reproduce:**

**Affected Versions:**

lasinicl:
## Purpose
$title

Fixes #34369

## Approach
```ballerina
public function main() {
    int? a = 10;
    if a is () {
        return;
    }
   // Types are narrowed following an `if` statement without an `else`, which is not completed normally. Therefore `a` should be 
   // `int` in this context. 
   int b = a;
   <curser> // here `b` should be suggested. As per #34369 this is not getting suggested ATM.
}
```
To fix this I have followed the following approach.
A new block is created internally to store the statements following an `if` statement without an `else`. This block's environment is updated with the narrowed symbols. 
```ballerina
public function main() {
    int? a = 10;
    if a is () {
        return;
    }
    {
        // Narrowing is applied to this block.
        int b = a;
        <curser> // `b` is being suggested here.
    }
}
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


pcnfernando:
Didn't get unreachable error for return.
```ballerina
function dummyTest() {
    int? a = 10;
    if a is () && !(a is ()) {
        return;
    }
    int b = a; //expected 'int', found 'int?'
}

```

lasinicl:
In the first example `int b = a;` this line results in an incompatible types error. Unreachability analysis is carried out after semantic analysis. If you comment out that line or correct it, you would see the unreachable error for the `return`.
```ballerina
function dummyTest() {
    int? a = 10;
    if a is () && !(a is ()) {
        return; //unreachable code
    }
    int? b = a; 
}
```
The second one is a bug. I haven't done any changes in the unreachability analysis in this PR. Can you create a separate issue? Will address it separately.

pcnfernando:
> * The second one is a bug. I haven't done any changes in the unreachability analysis in this PR. Can you create a separate issue? Will address it separately.

This is also due to the error at `int b = a;` If we remove the line, getting "string will not be matched to ()" at inner if.

lasinicl:
> This is also due to the error at `int b = a;` If we remove the line, getting "string will not be matched to ()" at inner if.

Ah yes. That's also a sematic error. :)

pcnfernando:
Shouldn't we narrow the else if cases?

```ballerina
public function dummyTest() {
    int|string? a = 10;
    if a is string {
        return;
    } else if a is () {
        return;
    }
    int x = a; // expected 'int', found '(int|string)?']
}
```

lasinicl:
> Shouldn't we narrow the else if cases?
> 
> ```ballerina
> public function dummyTest() {
>     int|string? a = 10;
>     if a is string {
>         return;
>     } else if a is () {
>         return;
>     }
>     int x = a; // expected 'int', found '(int|string)?']
> }
> ```

We should. Related issue https://github.com/ballerina-platform/ballerina-lang/issues/34307.

pcnfernando:
We'll have to add few visible symbol tests to test the behaviour. Are we handling it separately?

pcnfernando:
This should not be narrowed. Seems introduced with the PR
```ballerina
public function dummyTest() {
      int? a = 10;
      string? b = "100";
      if a is int {
          if (a == 5) {
              return;
          }
      }
      b = a;
}
```

lasinicl:
> We'll have to add few visible symbol tests to test the behaviour. Are we handling it separately?

Will add the tests separately.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
```diff
@@ -4950,26 +4951,98 @@ BLangValueType addValueType(Location pos, TypeKind typeKind) {
...
                 statements.add((BLangStatement) statement.apply(this));
             }
         }
         return statements;
     }
 
+    private void createBlockStatement(StatementNode currentStatement, int position,
+                                      NodeList<StatementNode> statementNodes,
+                                      List<BLangStatement> statements, Node endNode) {
+        for (int j = position; j < statementNodes.size(); j++) {
```
pcnfernando, line 5006: Can you check whether merging the logics in the two methods is feasible since we are doing the same thing.

pcnfernando: suggestions in generateAndAddBLangStatements() are applicable here as well

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -508,7 +508,24 @@ public void visit(BLangFunction funcNode) {
...
+        int stmtCount = -1;
         for (BLangStatement stmt : body.stmts) {
+            stmtCount++;
+            BLangStatement prevStatement = null;
+            if (stmtCount > 0) {
+                prevStatement = body.stmts.get(stmtCount - 1);
+            }
+            if (stmt.getKind() == NodeKind.BLOCK && prevStatement != null &&
+                    prevStatement.getKind() == NodeKind.IF && ((BLangIf) prevStatement).elseStmt == null &&
+            this.notCompletedNormally) {
```
pcnfernando, line 520: code not formatted

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -1823,7 +1840,26 @@ private void recursivelySetFinalFlag(BLangVariable variable) {
 
     public void visit(BLangBlockStmt blockNode) {
         env = SymbolEnv.createBlockEnv(blockNode, env);
-        blockNode.stmts.forEach(stmt -> analyzeStmt(stmt, env));
+        int stmtCount = -1;
+        for (BLangStatement stmt : blockNode.stmts) {
```
pcnfernando, line 1844: Extract out the common logic in func body and block node

pcnfernando (approved):
LGTM

hasithaa (approved):

hasithaa:
@pubudu91, @lasinicl is still working on the improvement we discussed. But let's go ahead with this PR since this will address the original issue.

PR #34559, opened 2022-01-07
Fix the issue related to match narrowing of intersection types

### LINKED ISSUE #34522: Match narrowing does not work for readonly in-lined types for certain cases
This issue can be related to #33694. Consider the following code.

```ballerina
import ballerina/io;

type T readonly & S;
type S [INT, string]|[STRING, string]|[FOO, string];

// It works if the type is inlined as below
// type T readonly & [INT, string]|[STRING, string]|[FOO, string];

const INT = 1;
const STRING = 2;
const FOO = 3;

public function main() {
    T t = [FOO, "hello"];
    match t {
        [STRING, var val] => {
            string s = val;
            io:println("str:" + s);
        }
        [INT, var val] => {
            string i = val;
            io:println("int:" + i.toString());
        }
        [_, var str] => {
            string varS = str; // error: type of str is any|error but it should be string
            io:println("foo:", varS);
        }
    }
}
```

Compiling the above code, gives the following error.
`
[main.bal:(25:27,25:30)] incompatible types: expected 'string', found '(any|error)'`

SandaruJayawardana:
Fixes #34522

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


pcnfernando (dismissed):

chiranSachintha:
For the following program, we get the error as `incompatible types: expected '(string|int)', found '(any|error)'`(due to same reason)
```ballerina
type T readonly & S;
type S string[2]|int[2];

// readonly & string[2]|int[2] k = ["", ""];
function listMatchPattern32(T t) returns string|int {
    string|int s = 10;

    match t {
        [_, var val] => {
            s = val; // error: incompatible types: expected '(string|int)', found '(any|error)'
        }
    }

    return s;
}
```
Shall we fix this issue also?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2734,10 +2734,16 @@ private void assignTypesToMemberPatterns(BLangMatchPattern matchPattern, BType b
                     ((BArrayType) restMatchPattern.getBType()).eType = restType;
                     return;
                 }
-                if (patternType.tag != TypeTags.TUPLE) {
+                BTupleType patternTupleType;
+                if (patternType.tag == TypeTags.TUPLE) {
+                    patternTupleType = (BTupleType) patternType;
+                } else if (patternType.tag == TypeTags.INTERSECTION &&
```
MaryamZi, line 2740: Wouldn't the same issue exist with mapping match patterns? Shouldn't we do something like

```java
        if (patternType.tag == TypeTags.INTERSECTION) {
            patternType = ((BIntersectionType) patternType).effectiveType;
        }
```

before the switch case?

tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/list-match-pattern.bal
```diff
@@ -883,6 +883,26 @@ function listMatchPattern31(T t) returns [string?, int?] {
...
 
+function testListMatchPattern32() {
+    T t1 = [STRING, "hello"];
+    T t2 = [INT, 1234];
+
+    assertEquals("hello", listMatchPattern32(t1));
+    assertEquals(1234, listMatchPattern32(t2));
+}
+
+function listMatchPattern32(T t) returns string|int {
```
MaryamZi, line 894: Let's add a test covering https://github.com/ballerina-platform/ballerina-lang/issues/34522#issuecomment-1009642734 also.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
SandaruJayawardana: Yeah. Will update. Thanks.

gimantha (dismissed):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -2692,6 +2692,9 @@ public void visit(BLangMappingMatchPattern mappingMatchPattern) {
     private void assignTypesToMemberPatterns(BLangMatchPattern matchPattern, BType bType) {
         BType patternType = this.types.getReferredType(bType);
         NodeKind matchPatternKind = matchPattern.getKind();
+        if (patternType.tag == TypeTags.INTERSECTION) {
```
MaryamZi, line 2695: Shall we add some tests for the non-list scenarios also?

tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/varbindingpatternmatchpattern/capture-binding-pattern.bal
```diff
@@ -101,6 +101,32 @@ function testCaptureBindingPattern4() {
     assertEquals("OTHER", captureBindingPattern4(4));
 }
 
+type A [int, string] & readonly;
+type B map<int> & readonly;
+type T A | B;
```
MaryamZi, line 106:
```suggestion
type T A|B;
```

tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/mapping-match-pattern.bal
```diff
@@ -598,6 +598,28 @@ function testMappingMatchPattern26() {
     assertEquals(1, mappingMatchPattern27({y: "hello world", x: 1}));
 }
 
+
+type T (map<string> & readonly) | map<int>;
+
+function mappingMatchPattern28(T v) returns string {
```
MaryamZi, line 604: Also works?

It is OK to have these tests also, but we need to make sure we have tests covering the previously uncovered path, and that those test would fail without this fix.

tests/jballerina-unit-test/src/test/resources/test-src/statements/matchstmt/const-pattern.bal
```diff
@@ -602,6 +602,28 @@ function testConstPattern18() {
     assertEquals("Default", result);
 }
 
+type A int[2] & readonly;
+
+function constPattern19(A x) returns int {
```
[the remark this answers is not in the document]
SandaruJayawardana: Yeah. Will remove this.

MaryamZi (approved):

PR #34582, opened 2022-01-10
Fix issue in toString() for decimal zero

### LINKED ISSUE #34472: Incorrect output of `toString` for decimal with tailing zeros
**Description:**
```
import ballerina/io;
    decimal d1 = 0.0d;
    decimal d2 = 0d;
    io:println(check decimal:fromString(d1.toString()) === check decimal:fromString(d2.toString())); // should be false
}
```
When there are tailing zeros after decimal point, `toString` does not consider them.

rdulmina:
## Purpose
Fix issue in toString() for decimal zero with different precisions

Fixes #34472

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/DecimalValue.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/DecimalValue.java
@@ -253,7 +253,7 @@ public class DecimalValue implements SimpleValue, BDecimal {
      * @param parent The link to the parent node
      */
     public String stringValue(BLink parent) {
-        if (this.valueKind != DecimalValueKind.OTHER) {
+        if (this.valueKind != DecimalValueKind.OTHER && this.valueKind != DecimalValueKind.ZERO) {
             return this.valueKind.getValue();
         }
         return value.toString();
--- a/langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
+++ b/langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
@@ -2259,27 +2259,27 @@ function testUsingCloneableReturnType() {
     assertEquality("dummyVal", caller.getAttribute("dummy"));
 }
 
-function testDecimalZeroToStringWithDifferentPrecisions() returns error? {
+function testDecimalZeroToStringWithDifferentPrecisions() {
     decimal d1 = 0.0d;
     decimal d2 = 0d;
 
     assertFalse(d1.toString() == d2.toString());
-    assertTrue(check decimal:fromString(d1.toString()) == check decimal:fromString(d2.toString()));
-    assertFalse(decimal:fromString(d1.toString()) === decimal:fromString(d2.toString()));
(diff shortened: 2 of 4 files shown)


KavinduZoysa (approved):

langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -2278,3 +2259,52 @@ function testUsingCloneableReturnType() {
...
+
+    assertFalse(d1.toString() == d2.toString());
+    assertTrue(check decimal:fromString(d1.toString()) == check decimal:fromString(d2.toString()));
+    assertFalse(decimal:fromString(d1.toString()) === decimal:fromString(d2.toString()));
+
+    d1 = 0.00000000d;
+    d2 = 0.0d;
+
+    assertFalse(d1.toString() == d2.toString());
+    assertTrue(check decimal:fromString(d1.toString()) == check decimal:fromString(d2.toString()));
```
MaryamZi, line 2274: If `check decimal:fromString(d1.toString())` results in the error being returned from this function, the test will not fail. IMO, it's better to handle the error here or use checkpanic.

rdulmina: replaced  with `checkpanic` thanx

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/DecimalValue.java
```diff
@@ -253,7 +253,7 @@ public Object frozenCopy(Map<Object, Object> refs) {
      * @param parent The link to the parent node
      */
     public String stringValue(BLink parent) {
-        if (this.valueKind != DecimalValueKind.OTHER) {
+        if (this.valueKind != DecimalValueKind.OTHER && this.valueKind != DecimalValueKind.ZERO) {
```
chiranSachintha, line 256: Consider these three cases,
**Case 1**
```ballerina
public function main() {
   decimal d1 = 0.0;
   decimal d2 = 0E-1;
   assertEquality(d1.toString(), d2.toString()); // true
}
```
**Case 2**

```ballerina
public function main() {
   decimal d1 = 1000;
   decimal d2 = 1E3;
   assertEquality(d1.toString(), d2.toString()); // false
}
```
**Case 3**
```ballerina
public function main() {
   decimal d1 = 0.1;
   decimal d2 = 1E-1;
   assertEquality(d1.toString(), d2.toString()); // true
}
```
Is the output of first and third cases correct?  According to the spec result of `ToString()` should be in the expression style(not considering the value). Isn't it? @MaryamZi @KavinduZoysa

rdulmina: As per the offline discussion created separate issue to address this #34648

chiranSachintha (approved):

KavinduZoysa (approved):

PR #34590, opened 2022-01-11
Fix few issues in annot processing in testerina TestProcessor

### LINKED ISSUE #34488: the `enable: false` fails to work if there is a comment on the line before it
Consider the following code. 
```ballerina
import ballerina/test;
@test:Config {
    //comment
    enable: false
}
public function testMe() {
    test:assertEquals("E", "E");
}
```

I have set `enable` as `false`. but for some reason, if I put a comment before that line inside the annotation, test will run.

SL-beta 6

lochana-chathura:
## Purpose
$subject.

Fixes #34488 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestProcessor.java
+++ b/misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestProcessor.java
@@ -23,6 +23,7 @@ import io.ballerina.compiler.api.symbols.Qualifier;
 import io.ballerina.compiler.api.symbols.Symbol;
 import io.ballerina.compiler.api.symbols.SymbolKind;
 import io.ballerina.compiler.syntax.tree.AnnotationNode;
+import io.ballerina.compiler.syntax.tree.BasicLiteralNode;
 import io.ballerina.compiler.syntax.tree.ExpressionNode;
 import io.ballerina.compiler.syntax.tree.FunctionDefinitionNode;
 import io.ballerina.compiler.syntax.tree.ListConstructorExpressionNode;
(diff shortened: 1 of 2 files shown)


misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/core/TestProcessor.java
```diff
@@ -518,4 +528,27 @@ private void processTestAnnotation(AnnotationNode annotationNode, String functio
...
+
+        if (fieldNameNode.kind() != SyntaxKind.STRING_LITERAL) {
+            return ((Token) fieldNameNode).text();
+        }
+
+        String fieldName = ((BasicLiteralNode) fieldNameNode).literalToken().text();
+        if (fieldName.length() > 1 && fieldName.charAt(fieldName.length() - 1) == '"') {
+            fieldName = fieldName.substring(1, fieldName.length() - 1);
+        } else {
+            // Missing end quote case
```
Dilhasha, line 548: A missing end quote would give a compilation error. So, I think we can simply omit this else block. WDYT?

lochana-chathura: +1

rdulmina (approved):

PR #34592, opened 2022-01-11
Fix issues in binding patterns in the AST visitors

pubudu91:
## Purpose
This PR fixes some bugs related to binding patterns which were there in the AST visitors of the Semantic API.

## Remarks
Came across #34606, #34615, #34617 while fixing this.

Also, #34642 was caused with this fix. Need to take a look at it separately.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
@@ -620,7 +620,7 @@ public class BLangNodeBuilder extends NodeTransformer<BLangNode> {
     public BLangNode transform(ModuleVariableDeclarationNode modVarDeclrNode) {
         TypedBindingPatternNode typedBindingPattern = modVarDeclrNode.typedBindingPattern();
         BindingPatternNode bindingPatternNode = typedBindingPattern.bindingPattern();
-        BLangVariable variable = getBLangVariableNode(bindingPatternNode);
+        BLangVariable variable = getBLangVariableNode(bindingPatternNode, getPositionWithoutMetadata(modVarDeclrNode));
 
         if (modVarDeclrNode.visibilityQualifier().isPresent()) {
             markVariableWithFlag(variable, Flag.PUBLIC);
@@ -633,7 +633,7 @@ public class BLangNodeBuilder extends NodeTransformer<BLangNode> {
         if (annotations != null) {
             variable.annAttachments = applyAll(annotations);
         }
-        variable.pos = getPositionWithoutMetadata(modVarDeclrNode);
+
         variable.markdownDocumentationAttachment =
                 createMarkdownDocumentationAttachment(getDocumentationString(modVarDeclrNode.metadata()));
         return variable;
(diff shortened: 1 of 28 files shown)


rdulmina:
BLangNodeBuilder changes are LGTM. Some langlib tests are failing

rdulmina (commented):
BLangNodeBuilder changes are LGTM. Some langlib tests are failing

dulajdilshan (approved):
LGTM

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/statements/foreach/ForeachNegativeTests.java
```diff
@@ -38,37 +38,37 @@ public void testForeachSemanticsNegative() {
         int index = 0;
         BAssertUtil.validateError(compile, index++,
                 "invalid list binding pattern: attempted to infer a list type, but found 'string'",
-                39, 17);
+                39, 13);
```
lochana-chathura, line 41: Old positions of these error messages seem to be correct. Now the error is given to the whole `typed-binding pattern` instead of `list-binding-pattern`. Should we correct it from where the error is logged?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
```diff
@@ -2800,7 +2800,7 @@ private VariableDefinitionNode createBLangVarDef(Location location,
                                                 Optional<io.ballerina.compiler.syntax.tree.ExpressionNode> initializer,
                                                 Optional<Token> finalKeyword) {
         BindingPatternNode bindingPattern = typedBindingPattern.bindingPattern();
-        BLangVariable variable = getBLangVariableNode(bindingPattern);
+        BLangVariable variable = getBLangVariableNode(bindingPattern, location);
```
lochana-chathura, line 2803: Related to the above comment. Looks like in the AST we either have to set typedBindingPattern pos or bindingPattern pos depending on the context. Not both positions. Which does not allow us to pick the preferred pos in later stages.

rdulmina: I think what we are doing in the nodebuilder is correct, since for the variable we need to set the full length pos. When we are logging the error I think we need to calculate the exact pos with the help of typedesc pos because we cannot hold binding pattern pos in blang node.

langlib/langlib-test/src/test/java/org/ballerinalang/langlib/test/statements/foreach/ForeachNegativeTests.java
rdulmina: Yeah old positions seams to be correct

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
pubudu91: Yes, previously there were instances where incorrect positions were set to the whole var. This lead to issues in the Semantic API. Let me check and see whether these would be a simple fix. Can see some places where the var node pos is taken instead of the type node's pos.

pubudu91: Shall we look into this separately? Just simply changing it to get the type node's position caused negative tests to fail in a different manner

rdulmina: +1

lochana-chathura: +1

pubudu91: Created #34642 to track this.

lochana-chathura (approved):

PR #34616, opened 2022-01-11
Check for referred expected type while resolving binary expression type

### LINKED ISSUE #34608: Invalid incompatible types error when a type reference type referring to a numeric type is used as the contextually-expected type for a multiplicative expression
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type Seconds decimal;

decimal _ = 3600 * 24; // OK
Seconds _ = 3600 * 24; // error: incompatible types: expected 'Seconds', found 'int'
```

**Affected Versions:**
slbeta6, stage-swan-lake

pcnfernando:
Fixes #34608

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -1306,7 +1306,7 @@ public class Types {
         return false;
     }
 
-    public BType getReferredType(BType type) {
+    public static BType getReferredType(BType type) {
         BType constraint = type;
         if (type.tag == TypeTags.TYPEREFDESC) {
             constraint = getReferredType(((BTypeReferenceType) type).referredType);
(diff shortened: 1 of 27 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1306,7 +1306,7 @@ public boolean isInherentlyImmutableType(BType type) {
         return false;
     }
 
-    public BType getReferredType(BType type) {
+    public static BType getReferredType(BType type) {
```
MaryamZi, line 1309: Is there a specific reason to make this a static method in this PR?

IMO, most (if not all) methods in this class can be static methods. But given what we've been doing should we leave this as is for now?

pcnfernando: The reason was there can be instances where we need to extract the referred type but the type's instance cannot be retrieved as CompilerContext is not there. for example in Symbols.java. As the util method doesn't depend on CompilerContext or such; made it static. Other solution would be to duplicate the util function in such cases

gimantha: This method has been used in classes like `Symbols`, which we cannot use the instance of `Types`

hasithaa (approved):

hasithaa:
Let's refactor this later. Create an issue to track this.

Issue #34629, opened 2022-01-12
Port SemType Tests from nBallerina to lang repo Tests


gimantha:
$Subject.
No of people: 2

KavinduZoysa:
As per the discussion had with @hasithaa, @rdhananjaya, @gimantha, @chiranSachintha, 
our plan is,
- write semtype test runner to check all subtype relations for given test file
- split the test available [here](https://github.com/ballerina-platform/nballerina/tree/main/compiler/testSuite/semtype) into small pieces and change the comment to include all possible relations in order to test against semtype test runner.

KavinduZoysa:
We have to write separate tests for type tests available in [here](https://github.com/ballerina-platform/nballerina/tree/main/compiler/testSuite/semtype). 
Let's restrict each test file to having a maximum of four relations. 
Any failing test case should be present in a single bal file and it should be renamed as `-f.bal`


- [ ] anydata-tv.bal   
- [ ] fixed-length-array-t.bal        
- [ ] not1-tv.bal   
- [ ] proj2-tv.bal  
- [ ] table-t.bal    
- [ ] tuple2-tv.bal  
- [ ] tuple4-tv.bal          
- [ ] xml-complex-rw-tv.bal  
- [ ] xml-readonly-tv.bal  
- [ ] xml-te.bal
- [ ] bdddiff1-tv.bal  
- [ ] fixed-length-array-tuple-t.bal  
- [ ] proj1-tv.bal  
- [ ] table2-t.bal  
- [ ] tuple1-tv.bal  
- [ ] tuple3-tv.bal  
- [ ] xml-complex-ro-tv.bal  
- [ ] xml-never-tv.bal       
- [ ] xml-sequence-tv.bal

KavinduZoysa:
@chiranSachintha, let's do our changes in this branch https://github.com/ballerina-platform/ballerina-lang/tree/semtype-tests

KavinduZoysa:
Identified issues:
https://github.com/ballerina-platform/ballerina-lang/issues/34696

chiranSachintha:
As discussed with @rdhananjaya We are going to ignore the following two kinds of test cases which are in the semtype test suite

1. [Intersection](https://github.com/ballerina-platform/nballerina/blob/f1560e776a3a3944e2a019c2f12678e22ce8e815/compiler/testSuite/semtype/fixed-length-array-t.bal#L40) related tests(Because ballerina doesn't support these kinds of scenarios)

[...]

PR #34699, opened 2022-01-18
Add semtype relations for anydata

KavinduZoysa:
## Purpose
Related to #34629


tests/jballerina-semtype-test/src/test/resources/test-src/anydata2.bal
```diff
@@ -0,0 +1,3 @@
+// TB<:A
```
gimantha, line 1: Shall we add some doc comments in Java Test class what these signs mean?

KavinduZoysa: Added

PR #34813, opened 2022-01-24
Fix bug in resolving direct deps which were previously indirect

### LINKED ISSUE #34809: Bug in version resolution when a transitive dependency becomes a direct dependency
**Description:**
When a transitive dependency (previous compilation) has become a direct dependency (this compilation), the resolution engine picks the latest minor version from the repos disregarding the version locked in the `Dependencies.toml`. This affects some unrelated dependencies also to update to the latest minor version which is not expected.

**Steps to Reproduce**
1. Create a package with the following imports and build with `slbeta3`:
```bal
import ballerina/http as _;
import ballerina/sql as _;
```
This will lock the following versions of the dependencies:
sql - `1.0.0`
http - `2.0.0`

2. Now import the `ballerina/log` package (this is already locked in the Dependencies.toml) as a transitive depenency:
```bal
import ballerina/http as _;
import ballerina/log as _;
import ballerina/sql as _;
```

Rebuild the package and check the `Dependencies.toml`. The `sql` package version should have updated to `1.2.0`

**Affected Versions**
At least SL beta 6

azinneera:
## Purpose
Fix bug in resolving direct deps which were previously indirect
Fixes #34809

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/ResolutionEngine.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/ResolutionEngine.java
@@ -169,10 +169,11 @@ public class ResolutionEngine {
 
     private Collection<PackageMetadataResponse> resolveDirectDependencies(Collection<DependencyNode> directDeps) {
         // Set the default locking mode based on the sticky build option.
-        PackageLockingMode lockingMode = resolutionOptions.sticky() ?
+        PackageLockingMode defaultLockingMode = resolutionOptions.sticky() ?
                 PackageLockingMode.HARD : PackageLockingMode.MEDIUM;
         List<ResolutionRequest> resolutionRequests = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


sameerajayasoma (changes requested):
Looks good, but can we add a test case using the resolution test framework?

azinneera:
> Looks good, but can we add a test case using the resolution test framework?

This behavior is intermittent. It depends on the order that the direct dependencies are added to the list. Was able to reproduce it with the changes done in c3b4ec1. Hope we can assume the order here is consistent.

sameerajayasoma (approved):

PR #32806, opened 2021-09-17
Implement new syntax APIs

lochana-chathura:
## Purpose
$subject.

- [x] Implement parseExpression() and parseStatements() APIs
- [ ] Implement isKeyword(), isIdentifier() APIs via SyntaxInfo class

Fixes #32720 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #34955, opened 2022-02-07
Update LS to use SyntaxInfo API and prevent renaming invalid qualified name references

### LINKED ISSUE #33249: Migrate to new parser APIs to identify keywords and identifiers
**Description:**
With #32806, parser API has introduced `SyntaxInfo.keywords(), SyntaxInfo.isKeyword(text), SyntaxInfo.isIdentifier(text)` methods. We should use them and remove the custom code we have written to perform/get above actions/results.

**Describe your task(s)**
See description

### LINKED ISSUE #34705: Renaming a symbol with invalid syntax to a valid state does not work
**Description:**

Rename `XYZ` to `io` in the following code snippet.

```
import ballerina/io;

public function main() {
    XYZ:print();
}
```

**Steps to reproduce:**

renaming-invalid-to-valid

**Affected Versions:**

### LINKED ISSUE #34785: Renaming Identifiers with escaped non-identifier characters are not working
**Description:**
Consider the following code:

```
import ballerina/io;

public function main() {
    string firstName = "Anjana";
    io:println(firstName);
}
```
Try to rename the symbol `firstName` to `first\ name`. 
Or manually change `firstName` to `first\ name`, and try renaming `first\ name` to `first\ name2`.

The following popup messages can be observed in the VSCode.
image

**Steps to reproduce:**
rename-escaped-identifiers

**Affected Versions:**

IMS94:
## Purpose
$subject

Fixes #33249
Fixes #34785
Fixes #34705

## Approach
With #32806, new API has been added to get keywords, check for validity of an identifier and to check if a string is a keyword. In this PR, we have updated LS to use those APIs.

Next, in `xyz:function1()`, if `xyz` is an invalid module prefix (no import present), we don't allow it to be renamed.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


lochana-chathura (dismissed):
LGTM.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
```diff
@@ -1445,25 +1426,7 @@ private static String getQualifiedModuleName(Module module) {
...
-                        } catch (IllegalAccessException e) {
-                            return null;
-                        }
-                    })
-                    .filter(Objects::nonNull)
-                    .collect(Collectors.toList());
-        } catch (ClassNotFoundException e) {
-            return Collections.emptyList();
-        }
+        return SyntaxInfo.keywords();
```
nadeeshaan, line 1429: Any specific reason to maintain this private method?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
IMS94: Inside `SyntaxInfo.keywords()`, it dynamically generates the keywords list. Therefore, pre-calculating the list of keywords and having them assigned to `BALLERINA_KEYWORDS` static variable is better for the performance. 

One thing we can do is, instead of using this private method, can do the following:
```
public static final List<String> BALLERINA_KEYWORDS = SyntaxInfo.keywords();
```
and remove the private method. WDYT?

nadeeshaan: Yeah, That was the same idea I was having. This particular method is only used for populating the data. Hence let's move to your suggestion.

IMS94: Done!

malinthar (approved):
LGTM!

nadeeshaan (approved):
LGTM

PR #34958, opened 2022-02-07
Fix no field completions within mapping constructor inside a new expression

### LINKED ISSUE #34841: Completions not working inside mapping constructor in new expression parameter
**Description:**
Consider the following code:
```ballerina
import ballerina/http;

    private MyClient myClient;

    public function init() returns error? {
        self.myClient = check new({<cursor>})
    }
}

type Config record {|
    int a;
|};

client class MyClient {

    function init(Config config) returns error? {

    }
}
```
At `<cursor>`, no completions are provided for the matching record fields. This happens only when initializing `client class` type variables. If `MyClient` was just a class, this works.

**Steps to reproduce:**
See description

**Affected Versions:**
Swan Lake 2201 RC2

IMS94:
## Purpose
$subject

Fixes #34841

## Approach
Updated `ContextTypeResolver` to first check the type of mapping constructor node before looking up the parent.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/test/resources/completion/expression_context/config/mapping_expr_ctx_config54.json
+++ b/language-server/modules/langserver-core/src/test/resources/completion/expression_context/config/mapping_expr_ctx_config54.json
@@ -6,14 +6,6 @@
   "source": "expression_context/source/mapping_expr_ctx_source54.bal",
   "items": [
     {
-      "label": "a",
-      "kind": "Variable",
-      "detail": "int",
-      "sortText": "AB",
-      "insertText": "a",
-      "insertTextFormat": "Snippet"
-    },
-    {
       "label": "CONST1",
       "kind": "Variable",
       "detail": "\"Test\"",
@@ -23,7 +15,7 @@
           "value": ""
         }
       },
-      "sortText": "AB",
+      "sortText": "F",
       "insertText": "CONST1",
       "insertTextFormat": "Snippet"
     },
@@ -47,7 +39,7 @@
       "label": "b",
       "kind": "Variable",
       "detail": "map<any>",
-      "sortText": "AB",
+      "sortText": "F",
       "insertText": "b",
       "insertTextFormat": "Snippet"
     }
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MappingConstructorExpressionNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MappingConstructorExpressionNodeContext.java
@@ -15,8 +15,6 @@
  */
 package org.ballerinalang.langserver.completions.providers.context;
 
-import io.ballerina.compiler.api.symbols.RecordFieldSymbol;
-import io.ballerina.compiler.api.symbols.RecordTypeSymbol;
 import io.ballerina.compiler.api.symbols.Symbol;
 import io.ballerina.compiler.api.symbols.SymbolKind;
 import io.ballerina.compiler.api.symbols.VariableSymbol;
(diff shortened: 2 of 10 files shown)


language-server/modules/langserver-core/src/test/resources/completion/expression_context/config/mapping_expr_ctx_config54.json
```diff
@@ -4,52 +4,5 @@
...
-    {
-      "label": "b",
-      "kind": "Variable",
-      "detail": "map<any>",
-      "sortText": "AB",
-      "insertText": "b",
-      "insertTextFormat": "Snippet"
-    }
-  ]
+  "items": []
```
nadeeshaan, line 7: Shall we check this test case? We should provide completions in this context

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/test/resources/completion/expression_context/config/mapping_expr_ctx_config54.json
IMS94: In this case, the source code is:
```ballerina
import ballerina/module1;

public const CONST1 = "Test";

public function testFunction(int param1, int param2) {
    int a = 5;
    map<any> b = {readonly a<cursor>};
}
```
At `<cursor>`, what user should be writing is the name of the key. And I'm not sure if `readonly` has a meaning inside a map. But we can write:
```ballerina
    map<any> b = {
        readonly key1: ""
    };
```

Therefore, AFAIU, not suggesting anything here is correct. WDYT?

nadeeshaan: IINM, the following is also allowed by the compiler according to the spec.
https://ballerina.io/spec/lang/master/#mapping-constructor-expr. Check the rule `[readonly] variableName`
```
public const CONST1 = "Test";

public function testFunction(int param1, int param2) {
    int a = 5;
    map<any> b = {readonly a};
}
```

gayalkuruppu: Isn't it a `mapping-binding-pattern`? https://ballerina.io/spec/lang/master/#mapping-binding-pattern

nadeeshaan: IINM, the mapping binding pattern come at the LHS of a variable def. That's also a possible use-case but we should capture it separately. Please correct me if I have been mistaken.

IMS94: We have to support `[readonly] variableName` inside mapping constructor. So, this test case need to fixed. I will look into that.

gayalkuruppu: Yeah agree, Context Type Resolver finds the context incorrectly for the scenario as `RecordType` where it has to be `AnyType`

IMS94: @gayalkuruppu that's expected. Because the RHS is a record in this scenario.

@nadeeshaan I had a look at this. Issue is, when providing variable completions, we look at the missing field names in the mapping constructor and suggest variables which match the names of missing fields. So, what we are getting here is correct. (Field `a` is already added in `map<any> b = {readonly a};`. RHS is a `record { int a; ...}`).

We have to suggest variable names regardless of them matching field names IMO. And we can address that in a new issue. WDYT?

nadeeshaan (approved):
LGTM

PR #34977, opened 2022-02-08
Fix create function in return statement and fail expression

### LINKED ISSUE #34767: Incorrect create function code action
**Description:**

In the first two scenarios we do not give the expected return type and its default value in the created function even though we identify the return type by the compiler. In the 3rd scenario, the number of parameters are doubled and has changed its name and type.

1) 
Screenshot 2022-01-21 at 11 22 10

2)
Screenshot 2022-01-21 at 11 23 16

3)
Screenshot 2022-01-21 at 11 24 23

**Steps to reproduce:**
1)
```
function test1() {
    fail getFail();
}
```
2)
```
function test2() returns int {
    return getReturn();
}
```
3)
```
function test3() {
    boolean a = true;
    a = getBool(boolean q, boolean w);
}
```

**Affected Versions:**
SL 2201.0.0 RC1

IMS94:
## Purpose
$subject

Fixes #34767

## Approach
Overridden required methods in the `FunctionCallExpressionTypeFinder`.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
@@ -35,8 +35,11 @@ import io.ballerina.compiler.syntax.tree.AssignmentStatementNode;
 import io.ballerina.compiler.syntax.tree.BinaryExpressionNode;
 import io.ballerina.compiler.syntax.tree.ErrorConstructorExpressionNode;
 import io.ballerina.compiler.syntax.tree.ExplicitNewExpressionNode;
+import io.ballerina.compiler.syntax.tree.FailStatementNode;
 import io.ballerina.compiler.syntax.tree.FunctionArgumentNode;
+import io.ballerina.compiler.syntax.tree.FunctionBodyBlockNode;
 import io.ballerina.compiler.syntax.tree.FunctionCallExpressionNode;
+import io.ballerina.compiler.syntax.tree.FunctionDefinitionNode;
 import io.ballerina.compiler.syntax.tree.IfElseStatementNode;
 import io.ballerina.compiler.syntax.tree.ImplicitNewExpressionNode;
 import io.ballerina.compiler.syntax.tree.LetExpressionNode;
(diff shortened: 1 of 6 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
```diff
@@ -345,6 +349,35 @@ public void visit(ImplicitNewExpressionNode implicitNewExpressionNode) {
...
+        this.semanticModel.typeOf(returnStatementNode).ifPresent(this::checkAndSetTypeResult);
+        if (resultFound) {
+            return;
+        }
+
+        // Get function type symbol and get return type descriptor from it
+        returnStatementNode.parent().accept(this);
+        if (resultFound && returnTypeSymbol.typeKind() == TypeDescKind.FUNCTION) {
+            FunctionTypeSymbol functionTypeSymbol = (FunctionTypeSymbol) returnTypeSymbol;
+            functionTypeSymbol.returnTypeDescriptor().ifPresentOrElse(this::checkAndSetTypeResult, this::resetResult);
```
malinthar, line 375: Isn't it better if we move this logic to the function definition node itself.  As the return type  of function represents  the particular  function, I feel that this logic does not belong here.

malinthar (changes requested):
Minor change request

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
IMS94: I thought of having this logic in the `FunctionDefinitionNode` visit method. Then decided not to. 
```ballerina
function myFunction(int x = fn1()) returns int {
   return undefinedFunction();
}
```
Can we guarantee that the `visit(FunctionDefinitionNode)` is being called by via `function call expr -> return statement -> function body -> function definition` hierarchy only?
What if we wanted to find the return type of `fn1()`? (For the case of `fn1()`, I think it is covered by a function parameter node visit method. Not sure. But it's the idea.) Therefore, to keep the logic clean, I thought of going with the current approach. WDYT?

malinthar (approved):
LGTM!

Issue #34847, opened 2022-01-26
[Debugger] Step over instruction navigates into function invocations in some contexts


NipunaRanasinghe:
**Description:**
Please refer to the below code segment extracted from nBallerina project. In similar contexts when doing `step-over` on the function invocation line, debugger navigates into the function definition instead of going to the next line.

```ballerina
function codeGenLogicalBinaryExpr(ExprContext cx, bir:BasicBlock bb, s:BinaryLogicalOp op, Position pos, s:Expr left, s:Expr right) returns CodeGenError|ExprEffect {
    var { result: lhs, block: block1, narrowing: lhsNarrowing } = check codeGenExprForBoolean(cx, bb, left);
    var [leftValue, leftFlags] = booleanOperandValue(lhs);
}
```

**Steps to reproduce:**
https://github.com/ballerina-platform/nballerina/blob/5bb20b2f179d86874feb5dfa823f316876f27edc/compiler/modules/front/expr.bal#L576

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

NipunaRanasinghe:
Closing since this particular scenario is fixed with https://github.com/ballerina-platform/ballerina-lang/pull/35261. However few other cases related to the step-ver instruction handling within recursive functions are yet to fixed and created https://github.com/ballerina-platform/ballerina-lang/issues/35402 to track the progress.

PR #34979, opened 2022-02-09
[Debugger] Fix `STEP_OVER` issue in recursion

praveennadarajah:
## Purpose
This PR will fix `STEP_OVER` issue in recursion function.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/34847


NipunaRanasinghe (commented):
It seems to me that this logic won't work when the program is having runtime errors and returns with a panic. Also I doubt that there's a fundamental problem in here since the source code is analysed (using syntax tree API) in order to execute a debug instruction. This will also result in performance degradations specially when debugging sources with high LOCs.

PR #35034, opened 2022-02-11
Use stack to adjust yield depth dynamically

### LINKED ISSUE #18444: Frames array size in jBallerina should be validated
**Description:**
ATM, this size is 100. This should be properly handled.

**Steps to reproduce:**

**Affected Versions:**
1.0-beta

https://github.com/ballerina-platform/ballerina-lang/issues/12807

### LINKED ISSUE #34671: Improve the runtime frames yield depth size to increase dynamically
**Description:** 
ATM max frames stack size is 256 but we can start with a small number and increase it dynamically on demand.

Nadeeshan96:
$title
Fixes #34671
Fixes #18444

Instead of an array of fixed size (previously 100), now a stack is used to store frames.

The performance test https://github.com/gabilang/ballerina-performance results are shown below for the previous array-based vs. stack-based implementation done by this PR.

* Common to both methods:
  * Number of concurrent users: 100, Loop count: 10000, Protocol: https, Ramp-up period: 60 s, Message size : 1024 bytes, No. of runs:3, error rate: 0%, performance test was done using k8s using rancher desktop in macOS Monterey 12.0.1
* Average throughput with array-based implementation: 1207.5 requests/s
* Average throughput with stack-based implementation: 1238.4 requests/s

Time taken for the below code snippet (tested with 100 runs): array-based = 670.5 ms ± 10.2 ms, stack-based = 672.5 ms ± 9.4 ms
```bal

public function main() {
   bar(0);
}

function bar(int x) {
    if (x == 250) {
        runtime:sleep(0.1);
        return;
    } else {
        bar(x + 1);
    }
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/expressions/invocations/function-stmt.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/expressions/invocations/function-stmt.bal
@@ -1,3 +1,21 @@
+// Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+//
+// WSO2 Inc. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
+// in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import ballerina/lang.runtime;
+
 function emptyFunction () {
 }
 
@@ -59,4 +77,17 @@ function addFloat4 (float x, float y) returns (float) {
 function addFloat5 (float x, float y) returns (float) {
     float z = x + y;
     return z;
+}
+
+function testRecursiveFunctionWhichYields() {
+   bar(0);
+}
+
+function bar(int x) {
+    if (x == 1000) {
+        runtime:sleep(0.01);
+        return;
+    } else {
+        bar(x + 1);
+    }
 }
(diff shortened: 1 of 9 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/expressions/invocations/function-stmt.bal
```diff
@@ -1,3 +1,21 @@
+// Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
warunalakshitha, line 1: Licence header should be 2022

tests/jballerina-unit-test/src/test/resources/test-src/expressions/invocations/function-stmt.bal
```diff
@@ -60,3 +78,16 @@ function addFloat5 (float x, float y) returns (float) {
     float z = x + y;
     return z;
 }
+
+function testRecursiveFunctionWhichYields() {
+   bar(0);
+}
+
+function bar(int x) {
```
warunalakshitha, line 86: Shall we add a test for stackoverflow. we can use same test with x == 10000 something

tests/jballerina-unit-test/src/test/resources/test-src/expressions/invocations/function-stmt.bal
Nadeeshan96: This file had been added in 2019 without the header. So I added the header mentioning the year as 2019.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/invocations/function-stmt.bal
Nadeeshan96: Will look into this.

warunalakshitha (approved):
LGTM

manuranga:
Was there a performance improvement or degradation from this?

Nadeeshan96:
> Was there a performance improvement or degradation from this?

Performance results are depicted in this PR's description: Remarks section. Results show there is no much difference in the performance, with the previous and the new method.

PR #35059, opened 2022-02-14
Fix a bug in `is-expr` with readonly arrays

### LINKED ISSUE #34879: `is-expr` gives incorrect results for the closed arrays
**Description:**
[Spec says](https://ballerina.io/spec/lang/2021R1/#is-expr): An is-expr is evaluated by evaluating the expression yielding a result v. If v belongs to the type denoted by type-descriptor, then the result of the is-expr is true; otherwise the result of the is-expr is false. 
Here, value belongs to a type if it looks like the type, and it will necessarily continue to look like the type no matter how the value is mutated.

Consider the following example,
```ballerina
anydata a = [2, 2, 2];
io:println(a.cloneReadOnly() is int[2]); // Output: true
// But according to spec it should be false since `a.cloneReadOnly()` is an immutable array of size 3.
```

**Affected Versions:**
2201.0.0 (Swan Lake) RC3

Nadeeshan96:
$title
Fixes #34879

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
@@ -2511,9 +2511,8 @@ public class TypeChecker {
             }
         }
 
-        int targetTypeSize = targetType.getSize();
         int sourceSize = source.size();
-        if ((targetTypeSize != -1) && (sourceSize != targetTypeSize)) {
+        if ((targetType.getState() != ArrayState.OPEN) && (sourceSize != targetType.getSize())) {
             return false;
         }
         for (int i = 0; i < sourceSize; i++) {
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
@@ -2511,7 +2511,12 @@ public class TypeChecker {
             }
         }
 
-        for (int i = 0; i < source.size(); i++) {
+        int targetTypeSize = targetType.getSize();
+        int sourceSize = source.size();
+        if ((targetTypeSize != -1) && (sourceSize != targetTypeSize)) {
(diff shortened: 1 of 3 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -2511,7 +2511,12 @@ private static boolean checkIsLikeArrayType(Object sourceValue, BArrayType targe
             }
         }
 
-        for (int i = 0; i < source.size(); i++) {
+        int targetTypeSize = targetType.getSize();
+        int sourceSize = source.size();
+        if ((targetTypeSize != -1) && (sourceSize != targetTypeSize)) {
```
warunalakshitha, line 2516: Do we need -1 check here?

Nadeeshan96: Need to consider the array size, only when the targetType array is a closed one (i.e. `[2, 2, 3] is int[]` is `true`). When it is an open one, its size is given as -1.

HindujaB: Shall we use `targetType.getState() != ArrayState.OPEN` instead of the -1 check?

Nadeeshan96: Addressed with e29f827.

warunalakshitha (approved):

warunalakshitha (changes requested):
Seems this is a breaking change since working one will be an error with this change

Nadeeshan96:
Closing this because a new PR will be sent to the master branch.

PR #35149, opened 2022-02-21
Add missing signature help for union typed expressions

### LINKED ISSUE #34391: No signature help triggered for new expression
**Description:**
Consider the following source snippet and the signature help is not provided at the cursor position. 

```
import ballerina/http;

function test() {
    http:Client cl = new(<cursor>)
}
```

malinthar:
## Purpose
$subject to add signature help when a new expression's type is a union. 

Fixes #34391 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
@@ -29,6 +29,7 @@ import io.ballerina.compiler.api.symbols.Symbol;
 import io.ballerina.compiler.api.symbols.SymbolKind;
 import io.ballerina.compiler.api.symbols.TypeDescKind;
 import io.ballerina.compiler.api.symbols.TypeSymbol;
+import io.ballerina.compiler.api.symbols.UnionTypeSymbol;
 import io.ballerina.compiler.api.symbols.VariableSymbol;
 import io.ballerina.compiler.syntax.tree.ChildNodeList;
 import io.ballerina.compiler.syntax.tree.ExplicitNewExpressionNode;
@@ -287,8 +288,8 @@ public class SignatureHelpUtil {
         // Add parameters and rest params
         List<ParameterSymbol> parameterSymbols = fnType.params().orElse(new ArrayList<>());
         parameters.addAll(parameterSymbols.stream()
-                                  .map(param -> new Parameter(param, false, false, context))
-                                  .collect(Collectors.toList()));
+                .map(param -> new Parameter(param, false, false, context))
+                .collect(Collectors.toList()));
         Optional<ParameterSymbol> restParam = fnType.restParam();
         restParam.ifPresent(parameter -> parameters.add(new Parameter(parameter, false, true, context)));
         boolean skipFirstParam = symbol.kind() == METHOD
@@ -506,9 +507,9 @@ public class SignatureHelpUtil {
                 QualifiedNameReferenceNode qNameRef = (QualifiedNameReferenceNode) nameReferenceNode;
                 funcName = (qNameRef).identifier().text();
                 filteredContent = QNameReferenceUtil.getModuleContent(context, qNameRef,
-                                                                      symbolPredicate
-                                                                              .and(symbol -> symbol.getName().orElse("")
-                                                                                      .equals(funcName)));
+                        symbolPredicate
+                                .and(symbol -> symbol.getName().orElse("")
+                                        .equals(funcName)));
             } else {
                 funcName = ((SimpleNameReferenceNode) nameReferenceNode).name().text();
                 List<Symbol> visibleSymbols = context.visibleSymbols(context.getCursorPosition());
@@ -672,6 +673,11 @@ public class SignatureHelpUtil {
         }
         if (rawType.kind() == CLASS && ((ClassSymbol) rawType).initMethod().isPresent()) {
             functionSymbols.add(((ClassSymbol) rawType).initMethod().get());
+        }
+        if (rawType.typeKind() == TypeDescKind.UNION) {
+            ((UnionTypeSymbol) typeDescriptor).memberTypeDescriptors().stream()
+                    .filter(typeSymbol -> CommonUtil.getRawType(typeSymbol).typeKind() == TypeDescKind.OBJECT).findAny()
+                    .ifPresent(objectMember -> functionSymbols.addAll(getFunctionSymbolsForTypeDesc(objectMember)));
         }
         functionSymbols.addAll(typeDescriptor.langLibMethods());
 
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
@@ -498,7 +498,7 @@ public class SignatureHelpUtil {
         }
         NonTerminalNode nodeAtCursor = context.getNodeAtCursor().get();
         if (nodeAtCursor.kind() == SyntaxKind.FUNCTION_CALL) {
-            return getFunctionSymbol(nodeAtCursor, context);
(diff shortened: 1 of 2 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
```diff
@@ -673,6 +674,11 @@ private static boolean withinTextRange(int position, @Nonnull NonTerminalNode no
         if (rawType.kind() == CLASS && ((ClassSymbol) rawType).initMethod().isPresent()) {
             functionSymbols.add(((ClassSymbol) rawType).initMethod().get());
         }
+        if (rawType.typeKind() == TypeDescKind.UNION) {
+            ((UnionTypeSymbol) typeDescriptor).memberTypeDescriptors().stream()
```
nadeeshaan, line 678: Here we get any of the members. Could there be more than one object types? If so wouldn't this introduce inconsistent behavior?
Also, shall we break the `getFunctionSymbol` into two? This method includes two cases and I believe it is better for clarity to break it into two, one for the function call expression and the other for method call, remote method call and new expression. WDYT?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
malinthar: +1. Sure will do

malinthar: > Here we get any of the members. Could there be more than one object types? If so wouldn't this introduce inconsistent behavior? 

It is unlikely there can be more than one object types.. Should we consider the first member  in that case?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
```diff
@@ -497,28 +498,38 @@ private static boolean withinTextRange(int position, @Nonnull NonTerminalNode no
...
-                                                                              .and(symbol -> symbol.getName().orElse("")
-                                                                                      .equals(funcName)));
-            } else {
-                funcName = ((SimpleNameReferenceNode) nameReferenceNode).name().text();
-                List<Symbol> visibleSymbols = context.visibleSymbols(context.getCursorPosition());
-                filteredContent = visibleSymbols.stream()
-                        .filter(symbolPredicate.and(symbol -> symbol.getName().get().equals(funcName)))
-                        .collect(Collectors.toList());
-            }
+            return getFunctionSymbol(((FunctionCallExpressionNode) nodeAtCursor), context);
```
IMS94, line 501: I don't think we need to explicitly cast here. In the runtime, this should get mapped to the matching overloaded method.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
```diff
@@ -497,28 +498,38 @@ private static boolean withinTextRange(int position, @Nonnull NonTerminalNode no
...
+            funcName = (qNameRef).identifier().text();
+            filteredContent = QNameReferenceUtil.getModuleContent(context, qNameRef,
+                    symbolPredicate
+                            .and(symbol -> symbol.getName().orElse("")
+                                    .equals(funcName)));
+        } else {
+            funcName = ((SimpleNameReferenceNode) nameReferenceNode).name().text();
+            List<Symbol> visibleSymbols = context.visibleSymbols(context.getCursorPosition());
+            filteredContent = visibleSymbols.stream()
+                    .filter(symbolPredicate.and(symbol -> symbol.getName().get().equals(funcName)))
```
IMS94, line 525: Shall we check if symbol name is present as well? 
`symbolPredicate.and(symbol -> symbol.getName().isPresent()).and(symbol -> symbol.getName().get().equals(funcName))`

IMS94 (changes requested):
LGTM! Added minor comments related to refactoring code.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
IMS94: If there are multiple object types, we simply can replace `.findFirst().ifPresent((objectMember -> ...)` with `forEach(objectMember -> ....)`

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
malinthar: Done

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
```diff
@@ -497,28 +498,38 @@ private static boolean withinTextRange(int position, @Nonnull NonTerminalNode no
...
+            funcName = (qNameRef).identifier().text();
+            filteredContent = QNameReferenceUtil.getModuleContent(context, qNameRef,
+                    symbolPredicate
+                            .and(symbol -> symbol.getName().orElse("")
+                                    .equals(funcName)));
+        } else {
+            funcName = ((SimpleNameReferenceNode) nameReferenceNode).name().text();
+            List<Symbol> visibleSymbols = context.visibleSymbols(context.getCursorPosition());
+            filteredContent = visibleSymbols.stream()
+                    .filter(symbolPredicate.and(symbol -> symbol.getName().isPresent() && symbol.getName().get().equals(funcName)))
```
TharushiJay, line 525: Checkstyle failure due to number of characters on this line

malinthar: Done

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
nadeeshaan: > > Here we get any of the members. Could there be more than one object types? If so wouldn't this introduce inconsistent behavior?
> 
> It is unlikely there can be more than one object types.. Should we consider the first member in that case?

Then it's fine. For safety, let's add a comment to avoid unnecessary usage of this method.

language-server/modules/langserver-core/src/test/resources/signature/expressions/expressions.bal
```diff
@@ -303,8 +303,8 @@ public client class Stub {
     public Child obj = new Child();
 
     # Create a new Stub
-    public function init(any arg1,string arg2) {
-
+    public function init(any arg1,string arg2) returns error?{
```
IMS94, line 306:
```suggestion
    public function init(any arg1,string arg2) returns error? {
```

IMS94 (changes requested):
LGTM! Added a minor suggestion.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/signature/SignatureHelpUtil.java
malinthar: We need the explicit cast here. Otherwise the `getFunctionSymbol(Node nodeAtCursor, SignatureContext context)` will be implemented.

mohanvive (approved):

nadeeshaan (approved):
LGTM

PR #35160, opened 2022-02-22
Fix map creation for user defined type that refers to a predefined union

### LINKED ISSUE #35105: Runtime exception when assigning a mapping constructor for user-defined type of `json`
**Description:**
$title. This works fine for other basic types which are supported by `json`. 
E.g. 
```ballerina
type T7 json;
// Following scenarios are working
T7 t15 = 21; // int
T7 t16 = 1.2; // float
T7 t17 = "JSON"; // string
T7 t18 = null; // nil
T7 t19 = true; // boolean

T7 t20 = {"a": 1}; // mapping constructor, runtime exception!
```
**Steps to reproduce:**
```ballerina
type T7 json;
T7 t20 = {"a": 1};
```
Output:
```
[2022-02-18 12:13:10,125] SEVERE {b7a.log.crash} - Given type can't be instantiated at runtime : json 
io.ballerina.runtime.internal.util.exceptions.BallerinaException: Given type can't be instantiated at runtime : json
        at io.ballerina.runtime.internal.values.TypedescValueImpl.instantiate(TypedescValueImpl.java:90)
        at test2.init(test2.bal:57)
        at $_init.$gen$$0046$0060init$0062(.:1)
        at $_init.$moduleInit(.)
        at $_init.$lambda$$moduleInit$(.)
        at io.ballerina.runtime.internal.scheduling.SchedulerItem.execute(Scheduler.java:611)
        at io.ballerina.runtime.internal.scheduling.Scheduler.run(Scheduler.java:317)
        at io.ballerina.runtime.internal.scheduling.Scheduler.runSafely(Scheduler.java:284)
```
**Affected Versions:**
SL 2201.0.0

HindujaB:
## Purpose
$subject
Fixes #35105 

## Approach
When creating a map value from the user-defined type which refers to the Ballerina in-built unions (`anydata`, `json`, `any`), it fails at runtime as the union type is not a proper map type. This PR fixes it by providing the correct type when creating such a map.

## Samples
```ballerina
type AnyDataType anydata;
AnyDataType map1 = {"a": 1};
```
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -2051,7 +2051,7 @@ public class TypeChecker extends BLangNodeVisitor {
         if (tag == TypeTags.TYPEREFDESC) {
             BType refType = Types.getReferredType(bType);
             BType compatibleType = checkMappingConstructorCompatibility(refType, mappingConstructor);
-            return (refType.tag != TypeTags.UNION && refType.tag != TypeTags.INTERSECTION) ? bType : compatibleType;
+            return (!isUnionType(refType.tag) && refType.tag != TypeTags.INTERSECTION) ? bType : compatibleType;
         }
 
         if (tag == TypeTags.INTERSECTION) {
@@ -2078,6 +2078,11 @@ public class TypeChecker extends BLangNodeVisitor {
         reportIncompatibleMappingConstructorError(mappingConstructor, bType);
         validateSpecifiedFields(mappingConstructor, symTable.semanticError);
         return symTable.semanticError;
+    }
+
+    private boolean isUnionType(int refTypeTag) {
+        return refTypeTag == TypeTags.UNION || refTypeTag == TypeTags.ANY || refTypeTag == TypeTags.ANYDATA ||
+                refTypeTag == TypeTags.JSON;
     }
 
     private BType checkReadOnlyMappingType(BLangRecordLiteral mappingConstructor) {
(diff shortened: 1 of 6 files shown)


HindujaB:
The code-coverage patch is low due to the BIR emitter changes which do not have any tests.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2051,7 +2051,8 @@ private BType checkMappingConstructorCompatibility(BType bType, BLangRecordLiter
         if (tag == TypeTags.TYPEREFDESC) {
             BType refType = Types.getReferredType(bType);
             BType compatibleType = checkMappingConstructorCompatibility(refType, mappingConstructor);
-            return (refType.tag != TypeTags.UNION && refType.tag != TypeTags.INTERSECTION) ? bType : compatibleType;
+            return (!TypeTags.isUnionTypeTag(refType.tag) && refType.tag != TypeTags.INTERSECTION) ? bType :
```
pcnfernando, line 2054: If the tag is not union/intersection, shouldn't we handle it as if we are handling non-union type without special-casing json etc ?

HindujaB: As we are checking for the mapping constructor here, for the non-type referenced cases they are given as a map of the respective type.
 Eg. : 
 ```ballerina
anydata a = {"a": 1};
 ```
 if we have `anydata` then this method returns `map<anydata>` to construct the map. 

This is because we always expect a map type for constructing the map directly from runtime.

pcnfernando: Yes, I get that. There's anyway a bug in the logic. My question is if we perform the logic below this specific if check without returning; wouldn't it handle the mentioned cases?

pcnfernando: Blocked on https://github.com/ballerina-platform/ballerina-lang/issues/35237

HindujaB:
Changing this to a draft as we need to decide on the target branch. @pcnfernando

HindujaB:
As per the offline discussion, the preferred `TypeChecker` fix can impact some other functions (#35237 ). It is decided that we need to handle that first before adding this fix. 
Therefore, closing this PR. 

The type emitter changes can be found in PR
https://github.com/ballerina-platform/ballerina-lang/pull/35312

PR #35166, opened 2022-02-22
Add support to show error lang-lib methods for all-member error union type

### LINKED ISSUE #34789: Completions not working for unions of errors
Consider the following:
```ballerina
type ErrX distinct error;
type ErrY distinct error;

type Err ErrX|ErrY|error<map<anydata>>;

public function main() {
    Err err = error ErrX("Foo");
    string message = err.<cursor>
}
```
Calling `err.message()`, `err.cause()` is allowed on unions of errors. Therefore, I think we need to provide those suggestions as well.

dulajdilshan:
## Purpose
$title

Fixes #34789

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


pubudu91 (dismissed):
Semantic API changes look ok to me. Tagged Maryam as well, to see if the type checker change is ok.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java
```diff
@@ -55,6 +55,7 @@
     public Boolean isAnyData = null;
     public Boolean isPureType = null;
     public boolean isCyclic = false;
+    public Boolean isAllErrorType = null;
```
MaryamZi, line 58: Do we really need this? I think we can live with the existing check.

dulajdilshan: But without this, things get iterated again, isn't it?

dulajdilshan: https://github.com/ballerina-platform/ballerina-lang/pull/35166/files#diff-51851f1451506fedd7c81b67a92282fada4eb00fe4728c70445d7262aed401f3R2400

MaryamZi: I meant to say that that's OK. 

There are quite a few optimizations we can do by caching type information, which we don't do atm. I'm guessing the approach we will take will mostly likely not depend on variables like this. IMO, we can live with iteration for now.

MaryamZi: This also opens up room for error given that we can add members to a union after creation - https://github.com/ballerina-platform/ballerina-lang/blob/16da1ac806f08c939b27a3b9f73ea4a11ef7e171/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/types/BUnionType.java#L252. In which case `isAllErrorType` has to be recomputed.

dulajdilshan: Ah, that's a good point. will this attribute.

MaryamZi (approved):

PR #35192, opened 2022-02-23
Fix a bug in `is-expr` with readonly arrays

### LINKED ISSUE #34879: `is-expr` gives incorrect results for the closed arrays

Nadeeshan96:
$title
Fixes #34879

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
@@ -2505,7 +2505,12 @@ public class TypeChecker {
             }
         }
 
-        for (int i = 0; i < source.size(); i++) {
+        int targetTypeSize = targetType.getSize();
+        int sourceSize = source.size();
+        if ((targetTypeSize != -1) && (sourceSize != targetTypeSize)) {
+            return false;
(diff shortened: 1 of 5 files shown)


HindujaB (approved):
LGTM

Nadeeshan96:
Closing this because a new PR will be sent to patch branch.

PR #35211, opened 2022-02-23
Improve create function code action to add isolated qualifier

### LINKED ISSUE #34981: Create function code action does not add the isolated qualifier
**Description:**
Consider the following scenario where I try the create function code action within a isolated function. The resulting function  is not accurate as it does not contain the isolated qualifier.

Peek 2022-02-09 10-53

**Steps to reproduce:**
```ballerina
import ballerina/io;
import ballerina/tcp;

const a = 10;
const b = 20;

service  on new tcp:Listener(0) {
    isolated resource function get addition () {
        int addResult = add(a,b);
        io:print(addResult);
    }
}

```
**Affected Versions:**
SL

malinthar:
## Purpose
$subject

Fixes #34981 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/IsolatedBlockFinder.java
```diff
@@ -0,0 +1,56 @@
+package org.ballerinalang.langserver.command.visitors;
```
nadeeshaan, line 1: Missing license header

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/IsolatedBlockFinder.java
```diff
@@ -0,0 +1,56 @@
...
+import io.ballerina.compiler.syntax.tree.Token;
+
+import java.util.Optional;
+
+/**
+ * Finds if a provided {@link FunctionCallExpressionNode} node is contained within an isolated block.
+ *
+ * @since 2.0.0
+ */
+public class IsolatedBlockFinder extends NodeVisitor {
```
nadeeshaan, line 18: Can't we use a node transformer here? Since we only capture the boolean status, it is better use a node transformer.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/FunctionGenerator.java
```diff
@@ -164,10 +166,11 @@ public static String generateFunction(DocumentServiceContext context, boolean ne
      * @param returnsClause  Returns clause
      * @param returnStmt     Return statement
      * @param newLineAtStart Whether to add a new line at the start of the function
+     * @param isolated       Whether the created function should be prefixed with isolated qualifier
      * @return Created function
      */
     private static String generateFunction(String functionName, List<String> args, String returnsClause,
```
nadeeshaan, line 172: If we have a function signature like this, when we expect a function signature without `isolated` qualifier, we have to include `false` explicitly. What if we provide a separate method to generate the isolated function?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/IsolatedBlockFinder.java
IMS94: +1 for that. But, we have to be aware of a possible NPE or CCE due to `ExternalTreeNodeList` returning null in it's `apply()` method. We cannot use primitive `boolean`. Instead, will have to use `Boolean` and handle `null` case as well.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/FunctionGenerator.java
malinthar: Done

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/IsolatedBlockFinder.java
malinthar: Done

nadeeshaan (approved):
LGTM

PR #35266, opened 2022-02-28
Fix issue in enum member value

### LINKED ISSUE #35035: Invalid value set for enum member specified using a quoted identifier
**Description:**
Let's say I have an enum like below.

```ballerina
import ballerina/io;

public enum CallState_State {
    undefined,
    'new
}
public function main() {
    CallState_State s = 'new;
    io:println(s);
}
```

When I print the enum, it prints the value of,
```
'new
```
The correct value should be `new` since I'm escaping a Ballerina keyword here.

**Related Discussions**
https://ballerina-platform.slack.com/archives/C47EAELR1/p1644562370179699

**Affected Versions:**
SL GA

- macOS
- Java 11

rdulmina:
Fix issue in enum member value when member name is a quoted identifier

Fixes #35035

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
@@ -5384,6 +5384,10 @@ public class BLangNodeBuilder extends NodeTransformer<BLangNode> {
                 }
             }
 
+            if (type == SyntaxKind.IDENTIFIER_TOKEN && text.startsWith(IDENTIFIER_LITERAL_PREFIX)) {
+                text = text.substring(1);
+            }
+
             if (type != SyntaxKind.TEMPLATE_STRING && type != SyntaxKind.XML_TEXT_CONTENT) {
                 Location pos = getPosition(literal);
                 validateUnicodePoints(text, pos);
@@ -5543,19 +5547,6 @@ public class BLangNodeBuilder extends NodeTransformer<BLangNode> {
                 builtInValueType.pos = getPosition(type);
                 return builtInValueType;
         }
-    }
-
-    private VariableNode createBasicVarNodeWithoutType(Location location, String identifier,
-                                                       Location identifierLocation, ExpressionNode expr) {
-        BLangSimpleVariable bLSimpleVar = (BLangSimpleVariable) TreeBuilder.createSimpleVariableNode();
-        bLSimpleVar.pos = location;
-        IdentifierNode name = this.createIdentifier(identifierLocation, identifier);
-        ((BLangIdentifier) name).pos = identifierLocation;
-        bLSimpleVar.setName(name);
-        if (expr != null) {
-            bLSimpleVar.setInitialExpression(expr);
-        }
-        return bLSimpleVar;
     }
 
     private BLangInvocation createBLangInvocation(Node nameNode, NodeList<FunctionArgumentNode> arguments,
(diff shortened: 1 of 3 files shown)


lochana-chathura (dismissed):

tests/jballerina-unit-test/src/test/resources/test-src/enums/enums.bal
```diff
@@ -115,13 +115,21 @@ function testEnumInRecursiveArray() {
     Keyword u = "union";
     assert(u, "union");
 
-    ApproxType b = "'boolean";
-    assert(b, 'boolean);
+    ApproxType b = "boolean";
+    assert(b, "boolean");
```
MaryamZi, line 119: Shall we keep the previous version also? 

```ballerina
    assert(b, 'boolean);
```

If there's a bug in the assignability check for `ApproxType b = "boolean";`, we may not be asserting the actual value of `'boolean`.

tests/jballerina-unit-test/src/test/resources/test-src/enums/enums.bal
```diff
@@ -115,13 +115,21 @@ function testEnumInRecursiveArray() {
...
+    assert(i, "int");
 }
 
+public enum CallState {
+    'new\u{5C}
+}
+
+function testEnumMemberContainingNumericEscape() {
+    CallState state = "new\\";
+    assert(state, "new\\");
```
MaryamZi, line 131: IMO we also need to test

```ballerina
    assert('new\u{5C}, "new\\");
```

rdulmina: Yes, My intention was to test it but missed it when writing the test :-D

gimantha (dismissed):

tests/jballerina-unit-test/src/test/resources/test-src/enums/enums.bal
```diff
@@ -115,13 +115,24 @@ function testEnumInRecursiveArray() {
     Keyword u = "union";
     assert(u, "union");
 
-    ApproxType b = "'boolean";
+    ApproxType b = "boolean";
+    assert(b, "boolean");
     assert(b, 'boolean);
```
MaryamZi, line 120: Missed this previously, shall we also add a separate test like `testEnumMemberContainingQuotedIdentifier`? Although this test also tests it, it's actually intended to test a different scenario (enum in recursive array).

MaryamZi (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
```diff
@@ -5384,6 +5383,10 @@ private BLangLiteral createSimpleLiteral(Node literal, SyntaxKind sign, boolean
                 }
             }
 
+            if (type == SyntaxKind.IDENTIFIER_TOKEN && text.startsWith(IDENTIFIER_LITERAL_PREFIX)) {
```
KRVPerera, line 5386: Isnt this a bug when transforming `transformEnumMember`. Shouldnt we do something like "transform(member.identifier())"

rdulmina: No, This is a kind of a general issue since if we r to create a string literal from identifier we should get rid of the initial quote if present. `transformEnumMember` is the only place that calls `createSimpleLiteral` with an identifier at the moment.

lochana-chathura: I also checked this previously.  Basically, we have to handle this in two places. 
1. createIdentifier() 2. createSimpleLiteral()
We do the unescaping also separately in the two cases.

lochana-chathura (approved):

PR #35308, opened 2022-03-03
Support validating test labels using the test runner

### LINKED ISSUE #33691: Validate test labels using the test runner
**Description:**
$title.

We need to
1. validate that the labels are correct, and have no typos
2. identify extra labels we have used but haven't yet added in the label list

lasinicl:
## Purpose
$title

Fixes #33691

## Approach
Validates whether the labels defined under the `Labels` header of the test case are among the predefined set of labels or else such labels are identified as "unknown labels". 
Also, check whether there are duplicate labels. 
We can't distinguish between the typos and the unknown labels.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/ballerina-spec-conformance-tests/src/test/java/org/ballerinalang/test/TestRunner.java
```diff
@@ -112,4 +113,9 @@ public void generateReports(ITestContext context) throws IOException {
             return testCases.iterator();
         }
     }
+
+    private Set<String> getPredefinedLabels() {
```
lochana-chathura, line 117: Every time we call this we dynamically populate the predefined labels. 
Instead, shall we have a static final variable to keep the predefined labels, so we populate it only once.
```java
private static final Set<String> predefinedLabels = TestRunnerUtils.readLabels(TEST_DIR.toString()).keySet();
```

tests/ballerina-spec-conformance-tests/src/test/java/org/ballerinalang/test/TestRunnerUtils.java
```diff
@@ -196,6 +199,36 @@ private static String validateKindOfTest(String kind) {
         }
     }
 
+    public static void validateLabels(String labels, Set<String> selectedLabels) {
```
lochana-chathura, line 202: Any reason to use the term `selectedLabels` instead of `predefinedLabels`(as per the passed argument name) here?

tests/ballerina-spec-conformance-tests/src/test/java/org/ballerinalang/test/TestRunnerUtils.java
```diff
@@ -196,6 +199,36 @@ private static String validateKindOfTest(String kind) {
...
+                continue;
+            }
+            if (!selectedLabels.contains(trimmedLabel)) {
+                unknownLabels.add(trimmedLabel);
+            }
+            labelsList.add(trimmedLabel);
+        }
+
+        boolean hasUnknownLabels = unknownLabels.length() != 0;
+        boolean hasDuplicateLabels = duplicateLabels.length() != 0;
```
lochana-chathura, line 220: Length cannot be negative. Lets use the exact logical expression.
```suggestion
        boolean hasUnknownLabels = unknownLabels.length() > 0;
        boolean hasDuplicateLabels = duplicateLabels.length() > 0;
```

lochana-chathura (approved):
With this, there seem to be failing tests due to the unknown, duplicate labels. Maybe we can fix them separately. Hence approving.

tests/ballerina-spec-conformance-tests/src/test/java/org/ballerinalang/test/TestRunnerUtils.java
```diff
@@ -196,6 +199,36 @@ private static String validateKindOfTest(String kind) {
...
+            if (!predefinedLabels.contains(trimmedLabel)) {
+                unknownLabels.add(trimmedLabel);
+            }
+            labelsList.add(trimmedLabel);
+        }
+
+        boolean hasUnknownLabels = unknownLabels.length() > 0;
+        boolean hasDuplicateLabels = duplicateLabels.length() > 0;
+
+        String diagnosticMsg = hasUnknownLabels ? (hasDuplicateLabels ?
```
pcnfernando, line 222: Shall we propagate the line number here and log it.

lasinicl: Do you mean the line number of the test case or the line number of the label? We don't seem to keep information on the line numbers of the labels ATM though. Will need to add support for this if that's the requirement here.

pcnfernando: Line number in the test case. Specifically .balt file.
Isn't `absLineNum` argument TestRunner.test() represent that?

lochana-chathura: Data provider provides the `absLineNum`. The line number of the last row of the labels will be  `absLineNum - 1` right?

lasinicl: Yeah. But anyway we can see which test has failed right?. Then, we can just check and correct the label there. If we can't pinpoint the row I don't think it makes any difference, even if log the error to the last line.

pcnfernando: Pin-pointing the raw is not needed. Just the line number will be useful as I see it. It's matter of checking the label string in the test. 
Does the failed test pin-point to the specific line on the original balt. No right? 
I think logging the `absLineNum` would be enough for the moment and it will be really useful. Theat's what I felt trying out your changes against .balt files

lasinicl: As discussed offline, added the line number in 7f9e8ab

pcnfernando (approved):
LGTM

MaryamZi:
@lasinicl, is there a way to figure out test failures due to invalid labels from any of the reports? Without going through the logs? If not, IMO, we need to introduce/improve a report to include this.

Right now, the failed test summary is also misleading when tests fail due to unknown labels.

Let's create an issue for this too.

lasinicl:
> test failures due to invalid labels

Created https://github.com/ballerina-platform/ballerina-lang/issues/35454

Issue #29389, opened 2021-03-23
Bad sad error hides the exact compilation error


chamil321:
**Description:**
Following sample results in a bad sad error hiding the exact compilation error
```ballerina
import ballerina/log;

public class Proto3SerDes {


    private final typedesc<anydata> & readonly dataType;

    # Generates a schema for a given data type.
    #
    # + ballerinaDataType - The data type of the value that needs to be serialized
    # + return - `serdes:Error` if the data type is not supported else nil
    public isolated function init(typedesc<anydata> ballerinaDataType) returns error? {
        self.dataType = ballerinaDataType;
        log:printInfo("new");
    }

    # Deserializes a given array of bytes.
    #
    # + encodedMessage - The encoded byte array of the value that is serialized
    # + return - The value represented by the encoded byte array
    public isolated function deserialize(byte[] encodedMessage, typedesc<anydata> T = dataType) returns anydata|error {
        log:printInfo("deserialize");
        return "deserialize(self, encodedMessage, self.dataType)";
    }
}

public function main() returns error? {

    Proto3SerDes|error p = new(json);
}
```

**Affected Versions:**
alpha3

Error
```
[2021-03-23 09:33:15,144] SEVERE {b7a.log.crash} - null 
java.lang.NullPointerException
	at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer$RHSTempVarOptimizer.visit(BIROptimizer.java:159)
	at org.wso2.ballerinalang.compiler.bir.model.BIRNode$BIRBasicBlock.accept(BIRNode.java:421)
	at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer$RHSTempVarOptimizer.lambda$visit$3(BIROptimizer.java:119)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer$RHSTempVarOptimizer.visit(BIROptimizer.java:119)
	at org.wso2.ballerinalang.compiler.bir.model.BIRNode$BIRFunction.accept(BIRNode.java:380)
	at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer$RHSTempVarOptimizer.lambda$visit$2(BIROptimizer.java:103)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
[...]

PR #35326, opened 2022-03-03
Fix compiler crash for defualtable params in methods

KavinduZoysa:
## Purpose

Fixes [#29389](https://github.com/ballerina-platform/ballerina-lang/issues/29389)

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
@@ -2165,8 +2165,9 @@ public class SymbolEnter extends BLangNodeVisitor {
         funcSymbol.source = funcNode.pos.lineRange().filePath();
         funcSymbol.markdownDocumentation = getMarkdownDocAttachment(funcNode.markdownDocumentationAttachment);
         SymbolEnv invokableEnv;
-        if (env.node.getKind() == NodeKind.CLASS_DEFN || env.node.getKind() == NodeKind.OBJECT_TYPE) {
-            invokableEnv = SymbolEnv.createFunctionEnv(funcNode, funcSymbol.scope, env.enclEnv);
+        NodeKind previousNodeKind = env.node.getKind();
+        if (previousNodeKind == NodeKind.CLASS_DEFN || previousNodeKind == NodeKind.OBJECT_TYPE) {
+            invokableEnv = SymbolEnv.createFunctionEnv(funcNode, funcSymbol.scope, fieldsRemovedEnv(env));
         } else {
             invokableEnv = SymbolEnv.createFunctionEnv(funcNode, funcSymbol.scope, env);
         }
@@ -2185,6 +2186,27 @@ public class SymbolEnter extends BLangNodeVisitor {
         if (funcNode.receiver != null) {
             defineAttachedFunctions(funcNode, funcSymbol, invokableEnv, validAttachedFunc);
         }
+    }
+
+    private SymbolEnv fieldsRemovedEnv(SymbolEnv currentEnv) {
+        Scope currentScope = currentEnv.scope;
+        Scope newScope = new Scope(currentScope.owner);
+        Map<Name, ScopeEntry> currentEntries = currentScope.entries;
+        boolean fieldsAvailable = false;
+        for (Name name : currentEntries.keySet()) {
+            ScopeEntry entry = currentEntries.get(name);
+            if (entry.symbol.getFlags().contains(Flag.FIELD)) {
+                    fieldsAvailable = true;
+            } else {
+                newScope.entries.put(name, entry);
+            }
+        }
+        if (!fieldsAvailable) {
+            return currentEnv;
+        }
+        SymbolEnv newEnv = new SymbolEnv(currentEnv.node, newScope);
+        currentEnv.copyTo(newEnv, currentEnv.enclEnv);
+        return newEnv;
     }
 
     private boolean isDeprecated(List<BLangAnnotationAttachment> annAttachments) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
@@ -2164,7 +2164,12 @@ public class SymbolEnter extends BLangNodeVisitor {
                                                                    getOrigin(funcNode.name.value));
         funcSymbol.source = funcNode.pos.lineRange().filePath();
         funcSymbol.markdownDocumentation = getMarkdownDocAttachment(funcNode.markdownDocumentationAttachment);
-        SymbolEnv invokableEnv = SymbolEnv.createFunctionEnv(funcNode, funcSymbol.scope, env);
+        SymbolEnv invokableEnv;
+        if (env.node.getKind() == NodeKind.CLASS_DEFN || env.node.getKind() == NodeKind.OBJECT_TYPE) {
+            invokableEnv = SymbolEnv.createFunctionEnv(funcNode, funcSymbol.scope, env.enclEnv);
+        } else {
+            invokableEnv = SymbolEnv.createFunctionEnv(funcNode, funcSymbol.scope, env);
+        }
         defineInvokableSymbol(funcNode, funcSymbol, invokableEnv);
         funcNode.setBType(funcSymbol.type);
 
@@ -4316,11 +4321,6 @@ public class SymbolEnter extends BLangNodeVisitor {
         invokableNode.symbol = funcSymbol;
         defineSymbol(invokableNode.name.pos, funcSymbol);
         invokableEnv.scope = funcSymbol.scope;
-        SymbolEnv enclEnv = invokableEnv.enclEnv;
-        NodeKind enclEnvNodeKind = enclEnv.node.getKind();
-        if (enclEnvNodeKind == NodeKind.CLASS_DEFN || enclEnvNodeKind == NodeKind.OBJECT_TYPE) {
-            invokableEnv.enclEnv = enclEnv.enclEnv;
-        }
         defineInvokableSymbolParams(invokableNode, funcSymbol, invokableEnv);
 
         if (Symbols.isFlagOn(funcSymbol.type.tsymbol.flags, Flags.ISOLATED)) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
@@ -4316,6 +4316,11 @@ public class SymbolEnter extends BLangNodeVisitor {
         invokableNode.symbol = funcSymbol;
         defineSymbol(invokableNode.name.pos, funcSymbol);
         invokableEnv.scope = funcSymbol.scope;
+        SymbolEnv enclEnv = invokableEnv.enclEnv;
+        NodeKind enclEnvNodeKind = enclEnv.node.getKind();
+        if (enclEnvNodeKind == NodeKind.CLASS_DEFN || enclEnvNodeKind == NodeKind.OBJECT_TYPE) {
+            invokableEnv.enclEnv = enclEnv.enclEnv;
+        }
         defineInvokableSymbolParams(invokableNode, funcSymbol, invokableEnv);
 
         if (Symbols.isFlagOn(funcSymbol.type.tsymbol.flags, Flags.ISOLATED)) {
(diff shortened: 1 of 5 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/object/object_functions_with_default_parameters_negative.bal
```diff
@@ -0,0 +1,124 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+public class Class1 {
+    int i = 0;
+
+    public function foo(int arg1 = i) { // error
```
gimantha, line 20: Shall we add more tests to include `self.i` also??

tests/jballerina-unit-test/src/test/resources/test-src/object/object_functions_with_default_parameters_negative.bal
```diff
@@ -0,0 +1,124 @@
...
+}
+
+public class Class6 {
+    int i = 0;
+
+    public function foo(int arg1 = i) { // error
+        object {} _ = object {
+            int i = 0;
+
+            function foo(int arg2 = i) { // error
```
gimantha, line 75: We should be able to refer to fields from the enclosing class right? Shall add a test case for that also, but with different field name?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -4313,6 +4313,11 @@ private void defineInvokableSymbol(BLangInvokableNode invokableNode, BInvokableS
         invokableNode.symbol = funcSymbol;
         defineSymbol(invokableNode.name.pos, funcSymbol);
         invokableEnv.scope = funcSymbol.scope;
+        SymbolEnv enclEnv = invokableEnv.enclEnv;
+        NodeKind enclEnvNodeKind = enclEnv.node.getKind();
+        if (enclEnvNodeKind == NodeKind.CLASS_DEFN || enclEnvNodeKind == NodeKind.OBJECT_TYPE) {
+            invokableEnv.enclEnv = enclEnv.enclEnv;
+        }
```
chiranSachintha, line 4320: IMO After creating a `SymbolEnv` changing `enclEnv` of that `env` is not correct. 
by doing `invokableEnv.enclEnv = enclEnv.enclEnv;` We set  `packageEnv` as `enclEnv` of `invokableEnv`

As I understand the parameters of methods(in class) cannot see the fields of that class. So when creating `invokableEnv` for methods(in class or object) we need to use class env without fields in the scope.

tests/jballerina-unit-test/src/test/resources/test-src/object/object_functions_with_default_parameters_negative.bal
```diff
@@ -0,0 +1,124 @@
...
+type Object1 object {
+    int i;
+    function foo(int arg1 = i); // error
+};
+
+type Object2 object {
+    int i;
+    function foo(int arg1 = i); // error
+    function bar(int arg1 = i); // error
+};
```
MaryamZi, line 124: Can we also add something like

```ballerina
function fn() {
    int i = 0;

    object {} _ = object {

        function foo(int x = i) {

        }
    };
}
```

and assert we don't give the error here. May not be able to run a positive test due to https://github.com/ballerina-platform/ballerina-lang/issues/35359.

tests/jballerina-unit-test/src/test/resources/test-src/object/object_functions_with_default_parameters_negative.bal
KavinduZoysa: Do you mean an example like this?

```ballerina
public class Class {
    int i = 0;

    public function foo(int arg1) {
        object {} _ = object {
            function foo(int arg2) {
                int _ = self.i;
            }
        };
    }
}
```

tests/jballerina-unit-test/src/test/resources/test-src/object/object_functions_with_default_parameters_negative.bal
KavinduZoysa: Added

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
KavinduZoysa: Changed the login, please have a look

tests/jballerina-unit-test/src/test/resources/test-src/object/object_functions_with_default_parameters_negative.bal
gimantha: No i mean, can we refer to the fields in outer most class as defaultable values to the functions in anonymous objects?

gimantha: I think it is already covered

gimantha:
```ballerina
public class Class {
    int i = 0;

    public function foo(int arg1) {
        object {} _ = object {
            function foo(int arg2 =  self.i) {
                int _ = self.i;
            }
        };
    }
}
```

like this?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2164,7 +2164,12 @@ public void visit(BLangFunction funcNode) {
                                                                    getOrigin(funcNode.name.value));
         funcSymbol.source = funcNode.pos.lineRange().filePath();
         funcSymbol.markdownDocumentation = getMarkdownDocAttachment(funcNode.markdownDocumentationAttachment);
-        SymbolEnv invokableEnv = SymbolEnv.createFunctionEnv(funcNode, funcSymbol.scope, env);
+        SymbolEnv invokableEnv;
+        if (env.node.getKind() == NodeKind.CLASS_DEFN || env.node.getKind() == NodeKind.OBJECT_TYPE) {
```
chiranSachintha, line 2168: We can extract `env.node.getKind()` into separate variable

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2164,7 +2164,12 @@ public void visit(BLangFunction funcNode) {
...
+            invokableEnv = SymbolEnv.createFunctionEnv(funcNode, funcSymbol.scope, env.enclEnv);
+        } else {
+            invokableEnv = SymbolEnv.createFunctionEnv(funcNode, funcSymbol.scope, env);
+        }
```
chiranSachintha, line 2172: Consider the following example
```ballerina
public class Class2 {
    int i = 0;

    public function foo(int i = i) { // error
        
    }
}
```
Here, the `enclEnv` of  `invokableEnv` created for the function `foo` is `pkgEnv`. IMO it should be `objectEnv` without fields.
@MaryamZi WDYT?

MaryamZi: Yeah, using the `pkgEnv` isn't correct.

I think what we discussed was that what we've use as the method's env would be an `env` that sees `self` instead of the fields like now? @KavinduZoysa, did we look into that?

tests/jballerina-unit-test/src/test/resources/test-src/object/object_functions_with_default_parameters_negative.bal
KavinduZoysa: This example does not work as expected. 
The reason is that method `foo` of the object can see the self of class `Class`. Therefore `int arg2 = self.i` is not identified as an error. 
I think this is not related to this PR. @gimantha shall we fix it separately?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
KavinduZoysa: Changed, please check

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolEnv.java
```diff
@@ -95,14 +95,23 @@ public SymbolEnv(BLangNode node, Scope scope) {
...
+        this.withoutEnclEnvCopyTo(target);
+        target.enclEnv = this;
+    }
+
+    public void exactCopyTo(SymbolEnv target) {
+        this.withoutEnclEnvCopyTo(target);
+        target.enclEnv = this.enclEnv;
+    }
+
+    private void withoutEnclEnvCopyTo(SymbolEnv target) {
```
MaryamZi, line 107: Can't we have an overloaded `copyTo` method that accepts an `enclEnv` instead?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2182,6 +2188,21 @@ public void visit(BLangFunction funcNode) {
         }
     }
 
+    private SymbolEnv fieldsRemovedEnv(SymbolEnv currentEnv) {
+        Scope currentScope = currentEnv.scope;
```
MaryamZi, line 2192: We can probably avoid creating a new env if the class/object doesn't have fields.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolEnv.java
KavinduZoysa: Yes, that is good, changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
KavinduZoysa: Changed

tests/jballerina-unit-test/src/test/resources/test-src/object/object_functions_with_default_parameters_negative.bal
gimantha: Lets create a separate issue then?

gimantha (approved):
LGTM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolEnv.java
```diff
@@ -95,14 +95,23 @@ public SymbolEnv(BLangNode node, Scope scope) {
...
+        this.withoutEnclEnvCopyTo(target);
+        target.enclEnv = this;
+    }
+
+    public void copyTo(SymbolEnv target, SymbolEnv enclEnv) {
+        this.withoutEnclEnvCopyTo(target);
+        target.enclEnv = enclEnv;
+    }
+
+    private void withoutEnclEnvCopyTo(SymbolEnv target) {
```
MaryamZi, line 107:
```suggestion
    private void copyToWithoutEnclEnv(SymbolEnv target) {
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2182,6 +2188,27 @@ public void visit(BLangFunction funcNode) {
...
+        Scope newScope = new Scope(currentScope.owner);
+        Map<Name, ScopeEntry> currentEntries = currentScope.entries;
+        boolean fieldsAvailable = false;
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/SymbolEnv.java
KavinduZoysa: Changed

gimantha (approved):

PR #35336, opened 2022-03-04
Update syntax tree visible endpoint node with `packageName` and `version`

kanushka:
## Purpose
Update syntax tree visible endpoint node with `packageName` and `version`.

## Approach
Use module object's package name and version to update custom syntax tree.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/diagram-util/src/main/java/org/ballerinalang/diagramutil/SyntaxTreeMapGenerator.java
+++ b/misc/diagram-util/src/main/java/org/ballerinalang/diagramutil/SyntaxTreeMapGenerator.java
@@ -249,88 +249,74 @@ public class SyntaxTreeMapGenerator extends NodeTransformer<JsonElement> {
 
     private JsonObject visibleEP(Node node, TypeSymbol typeSymbol, boolean isRemoteAction) {
         JsonObject symbolMetaInfo = new JsonObject();
-        ModuleID moduleID = typeSymbol.getModule().isPresent() ? typeSymbol.getModule().get().id() : null;
-        String orgName = moduleID != null ? moduleID.orgName() : "";
-        String packageName = moduleID != null ? moduleID.packageName() : "";
-        String moduleName = moduleID != null ? moduleID.moduleName() : "";
-        String version = moduleID != null ? moduleID.version() : "";
 
-        if (node.kind() == SyntaxKind.REQUIRED_PARAM) {
-            RequiredParameterNode requiredParameterNode = (RequiredParameterNode) node;
-            Optional<Token> paramName = requiredParameterNode.paramName();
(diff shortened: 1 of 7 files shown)


NipunaMarcus:
There are a few places not covered by the test you added ...  can you take a look?

misc/diagram-util/src/main/java/org/ballerinalang/diagramutil/SyntaxTreeMapGenerator.java
```diff
@@ -270,7 +274,9 @@ private JsonObject visibleEP(Node node, TypeSymbol typeSymbol, boolean isRemoteA
             symbolMetaInfo.addProperty("isCaller", "Caller".equals(typeSymbol.getName().orElse(null)));
```
nadeeshaan, line 274: Do not use `orElse(null)` this bypasses the advantage of optional, hence use `orElse("")`

misc/diagram-util/src/main/java/org/ballerinalang/diagramutil/SyntaxTreeMapGenerator.java
```diff
@@ -251,7 +251,9 @@ private JsonObject visibleEP(Node node, TypeSymbol typeSymbol, boolean isRemoteA
         JsonObject symbolMetaInfo = new JsonObject();
         ModuleID moduleID = typeSymbol.getModule().isPresent() ? typeSymbol.getModule().get().id() : null;
         String orgName = moduleID != null ? moduleID.orgName() : "";
+        String packageName = moduleID != null ? moduleID.packageName() : "";
         String moduleName = moduleID != null ? moduleID.moduleName() : "";
+        String version = moduleID != null ? moduleID.version() : "";
 
         if (node.kind() == SyntaxKind.REQUIRED_PARAM) {
```
nadeeshaan, line 258: If we consider the if-else ladder we have used here, there is a considerable amount of duplication for setting the properties such as `typeName`, `orgName` and etc. Let's refactor this one in this PR itself. Check whether if we can come up with a solution based on a design pattern as well (Builder pattern maybe).

nadeeshaan (changes requested):

misc/diagram-util/src/main/java/org/ballerinalang/diagramutil/SyntaxTreeMapGenerator.java
kanushka: sure. will refactor this.

nadeeshaan (approved):
LGTM

PR #35361, opened 2022-03-07
Fix renaming params in undefined function invocations

### LINKED ISSUE #35016: Rename symbol does not rename parameter name when the function is undefined
**Description:**

Peek 2022-02-10 18-04

**Affected Versions:**
Swarn Lake

IMS94:
## Purpose
$subject

Fixes #35016

## Approach
Updated `ReferenceFinder` to visit `argExprs` of invocations if required and rest args are empty while `argExprs` are not empty.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/ReferenceFinder.java
```diff
@@ -816,6 +816,12 @@ public void visit(BLangInvocation invocationExpr) {
         find(invocationExpr.annAttachments);
         find(invocationExpr.restArgs);
 
+        // If the invocation is undefined, required and rest args won't be set. Need to visit argExprs then.
+        if (invocationExpr.requiredArgs.isEmpty() && invocationExpr.restArgs.isEmpty() &&
```
nadeeshaan, line 820: Should we use the ``argExprs` at all? This is because the particular field is coming from an old implementation and it is not recommended to use this field even in the phases such as `typeChecker` and `desgugar`.

IMS94: @nadeeshaan I was thinking the same and wanted to ask @pubudu91 if we can populate requiredArgs or restArgs in this scenario. +1 for the suggestion.

IMS94:
In an offline discussion, it was decided to visit `BLangInvocation.argsExprs` instead of `requiredArgs` and `restArgs`. Updated the PR with that approach. As a result a set of semantic token tests had to be updated which were valid cases.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/ReferenceFinder.java
```diff
@@ -812,9 +812,8 @@ public void visit(BLangInvocation invocationExpr) {
             find(invocationExpr.expr);
         }
 
-        find(invocationExpr.requiredArgs);
         find(invocationExpr.annAttachments);
-        find(invocationExpr.restArgs);
```
pubudu91, line 817: Are rest args also available in argExprs?

IMS94: Just checked. Rest args are there.

pubudu91 (approved):

pubudu91 (approved):

PR #35380, opened 2022-03-08
Fix NPE in `decimal:fromString()`

### LINKED ISSUE #35318: NPE when `decimal:fromString()` is called with an empty string as the argument
**Description:**
$title.

**Steps to reproduce:**
```ballerina
public function main() {
    decimal|error fromString = decimal:fromString("");
}
```

```cmd
error: java.lang.NullPointerException
        at ballerina.lang.decimal.0:fromString(decimal.bal:103)
           test:main(test.bal:2)
```

**Affected Versions:**
2201.0.1

Nadeeshan96:
$title
Fixes #35318

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/DecimalValue.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/DecimalValue.java
@@ -72,7 +72,8 @@ public class DecimalValue implements SimpleValue, BDecimal {
                 this.value = new BigDecimal(value, MathContext.DECIMAL128);
             } catch (NumberFormatException exception) {
                 String message = exception.getMessage();
-                if (message.equals("Too many nonzero exponent digits.") || message.equals("Exponent overflow.")) {
+                if ((message != null) && (message.equals("Too many nonzero exponent digits.") ||
+                        message.equals("Exponent overflow."))) {
                     throw ErrorCreator.createError(BallerinaErrorReasons.LARGE_EXPONENT_ERROR,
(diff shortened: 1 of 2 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/DecimalValue.java
```diff
@@ -72,7 +72,8 @@ public DecimalValue(String value) {
                 this.value = new BigDecimal(value, MathContext.DECIMAL128);
             } catch (NumberFormatException exception) {
                 String message = exception.getMessage();
-                if (message.equals("Too many nonzero exponent digits.") || message.equals("Exponent overflow.")) {
+                if ((message != null) && (message.equals("Too many nonzero exponent digits.") ||
```
warunalakshitha, line 75: Should we handle empty string separately instead of replying on message != null

Nadeeshan96: No, If the received exception message is not equal to those two strings it will anyway `throw exception;`. The problem was `exception.getMessage();` sometimes may be null, and then couldn't use `equals` method on `null`.

warunalakshitha (approved):
LGTM

PR #35408, opened 2022-03-10
Modify treegen to allow "createdYear" in new node license header

### LINKED ISSUE #31047: License header is hard coded to "2020" in treegen
**Description:**
$subject makes the new nodes adding to ST have the 2020 license header.

lochana-chathura:
## Purpose
$subject.

Fixes #31047 

## Approach
- Add `createdYear` property to `syntax_tree_descriptor.json`. In absence of specification, it is defaulted to "2020". 
- `{{createdYear}}` variable was only introduced to `internal_node_template.mustache` & `external_node_template.mustache`
as other template license headers are not impacted on new node creation.

## Samples
To test, add the following to [syntax_tree_descriptor.json](https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json) and run `./gradlew treegen`.
```json
        {
            "name": "SpreadMemberNode",
            "base": "Node",
            "kind": "SPREAD_MEMBER",
            "createdYear": "2022",
            "attributes": [
                {
                    "name": "ellipsis",
                    "type": "Token"
                },
                {
                    "name": "expression",
                    "type": "ExpressionNode"
                }
            ]
        }
```
To test the default behavior you can remove the `"createdYear": "2022"` property.

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-treegen/src/main/java/io/ballerinalang/compiler/internal/treegen/model/json/SyntaxNode.java
+++ b/compiler/ballerina-treegen/src/main/java/io/ballerinalang/compiler/internal/treegen/model/json/SyntaxNode.java
@@ -32,19 +32,22 @@ public class SyntaxNode {
     private String base;
     private String type;
     private boolean isAbstract;
+    private String createdYear;
 
     public SyntaxNode(String name,
                       List<SyntaxNodeAttribute> attributes,
                       String kind,
                       String base,
                       String type,
-                      boolean isAbstract) {
+                      boolean isAbstract,
+                      String createdYear) {
         this.name = name;
         this.attributes = attributes;
         this.kind = kind;
         this.base = base;
         this.type = type;
         this.isAbstract = isAbstract;
+        this.createdYear = createdYear;
     }
 
     public SyntaxNode() {
@@ -73,5 +76,9 @@ public class SyntaxNode {
 
     public boolean isAbstract() {
         return isAbstract;
+    }
+
+    public String getCreatedYear() {
+        return createdYear;
     }
 }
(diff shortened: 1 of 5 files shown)


sameerajayasoma (changes requested):
I am not sure whether mixing year with the syntactic structure of a node. two different concerns IMO.

hasithaa:
I also agree with Sameera's point. Going forward, adding new nodes is rare. So after generating the nodes, We can update Licenses headers, before pushing the code.

lochana-chathura:
Agreed on point on mixing the year.

> So after generating the nodes, We can update Licenses headers, before pushing the code.

Currently, treegen re-writes all the nodes(including existing ones). So if we go forward with manually updating the license header, then the diff after executing `./gradlew treegen` will grow over time, as we add new nodes. Yeah. since adding nodes is rare this is not going to be an immediate problem or may never reach an unbearable level.

However, we can keep the year information separately, so that the `./gradlew treegen` output will always be in line with the actual one?

rdulmina:
> Agreed on point on mixing the year.
> 
> > So after generating the nodes, We can update Licenses headers, before pushing the code.
> 
> Currently, treegen re-writes all the nodes(including existing ones). So if we go forward with manually updating the license header, then the diff after executing `./gradlew treegen` will grow over time, as we add new nodes. Yeah. since adding nodes is rare this is not going to be an immediate problem or may never reach an unbearable level.
> 
> However, we can keep the year information separately, so that the `./gradlew treegen` output will always be in line with the actual one?

I think keeping a separate file to store year info the best and the easiest approach

lochana-chathura:
Closing temporarily as this was not originally planned for the sprint. 
I will revisit considering the suggestions.

Issue #35378, opened 2022-03-08
Toml parser inconsistency between table and inline table nodes


HindujaB:
**Description:**
$subject

**Steps to reproduce:**
When we parse a TOML file with an in-line table that contains in-line tables at different levels, the parsed nodes are not always `TomlInlineTableValueNode`. They provide`TomlTableNode` sometimes.

For the below TOML file ,
```toml
a = [{ address = { number = 20, road = "Palm Grove", city = "Colombo 3" }, accounts = [{ id = "1" }, { id = "2" }] }]
```
This is parsed as below. 
![Screenshot from 2022-03-08 16-03-35](https://user-images.githubusercontent.com/28644893/157219506-2fe101fa-607e-4163-ab3c-af9caac14210.png)

**Affected Versions:**
Swanlake GA

**OS, DB, other environment details and versions:**

**Related Issues (optional):**
#35024

PR #35416, opened 2022-03-10
Add `toTable` util into InlineTableValueNode

### LINKED ISSUE #35304: Need an API to convert TOML in-line table to a TOML table
**Description:**
$subject
As per the TOML spec, an in-line TOML table is equivalent to a TOML table.
https://toml.io/en/v1.0.0#inline-table

**Describe your problem(s)**
The configurable implementations currently do not support the in-line tables. But, it will be required as we try to improve this. 
In order to support this, we will need an API that can convert an in-line TOML table to a TOML table. 

**Describe your solution(s)**
An API in the `TomlInlineTableValueNode` as `toTable()` that can return an equivalent `TomlTableNode` would be better.

xlight05:
## Purpose
Add `toTable` util into InlineTableValueNode which returns a new TomlTableNode. 
Resolves #35304

We need to completely hide InlineTableValueNode from the ast in array values after we fix the hierarchy.  properly. Issue is tracked from #35378

Fixes #<Issue Number>

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlInlineTableValueNode.java
+++ b/misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlInlineTableValueNode.java
@@ -50,7 +50,7 @@ public class TomlInlineTableValueNode extends TomlValueNode {
      * Get a single element of the array based on the Index.
      *
      * @param index index of the element
-     * @param <T> Type of the element
+     * @param <T>   Type of the element
      * @return Element object
      */
     public <T extends TopLevelNode> T get(int index) {
@@ -106,7 +106,7 @@ public class TomlInlineTableValueNode extends TomlValueNode {
         }
         return list;
     }
-    
+
     public TomlTableNode toTable() {
         Map<String, TopLevelNode> table = new HashMap<>();
         for (TopLevelNode node : elements) {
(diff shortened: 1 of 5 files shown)


anuruddhal (approved):

anuruddhal (dismissed):

HindujaB:
When we have a toml value like below 
```toml
a = [{accounts = [{id = "1"}, {id = "2"}]}]
```
The API seems to not convert the in-line tables recursively into table nodes. 

The in-line table node 
Screenshot from 2022-03-10 19-22-43

The `toTable()` result
Screenshot from 2022-03-10 19-21-17

Is this expected? @xlight05

xlight05:
> When we have a toml value like below
> 
> ```toml
> a = [{accounts = [{id = "1"}, {id = "2"}]}]
> ```
> 
> The API seems to not convert the in-line tables recursively into table nodes.
> 
> The in-line table node Screenshot from 2022-03-10 19-22-43
> 
> The `toTable()` result Screenshot from 2022-03-10 19-21-17
> 
> Is this expected? @xlight05

Yes, In this util class we don't edit the tree in any way. Instead we simply return a toml table for the inline value node.  You might have to check recursively on array nodes and handle this when consuming.  We will properly fix this in #35378

HindujaB (approved):
Verified the API. LGTM

Issue #33377, opened 2021-10-25
The inherent type for `any arr = [1];` should be `(any|error)[]`


lochana-chathura:
**Description:**
$subject.

For `any arr = [1];` currently the inherent type is `any[]` But it should be `(any|error)[]`.
i.e. Applicable contextually expected type here should be  `(any|error)[]` not `any[]`, since `(any|error)[]`  belongs to `any`.

**Spec-Issue:**
ballerina-platform/ballerina-spec#969

MaryamZi:
Similarly for `any _ = {}`, the type should be `map<any|error>`,

PR #35437, opened 2022-03-12
Fix array and map inherent types when the contextually-expected type for the constructor is `any`

MaryamZi:
## Purpose
$title.

The types were previously `any[]` and `map<any>`. This PR updates them to be `(any|error)[]` and `map<any|error>` in-line with the spec.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/33377

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/ballerina-shell/modules/shell-core/src/test/resources/testcases/evaluator/query.expr.json
+++ b/ballerina-shell/modules/shell-core/src/test/resources/testcases/evaluator/query.expr.json
@@ -19,7 +19,7 @@
   {
     "description": "Query stream evaluation - get value.",
     "code": "reportListStream",
-    "expr": "stream <stream<map>>"
+    "expr": "stream <stream<map<(any|error)>>>"
   },
   {
     "description": "String from query evaluation.",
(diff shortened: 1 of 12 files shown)


gimantha (approved):

chiranSachintha:
By doing this we disallow following kind of scenario. Is it correct?
```ballerina
type val any[];
public function main() {
    any|val arr = [1]; // error: ambiguous type '(any|val)'
}
```

MaryamZi:
> By doing this we disallow following kind of scenario. Is it correct?
> 
> ```ballerina
> type val any[];
> public function main() {
>     any|val arr = [1]; // error: ambiguous type '(any|val)'
> }
> ```

Any reason why it should be allowed?

This is like saying
```ballerina
any|any[] arr = [1];
```

which is like saying
```ballerina
(any|error)[]|any[] arr = [1];
```

Here `[1]` is a valid constructor with both `(any|error)[]` and `any[]`. So we don't know what type to use.

chiranSachintha:
Yeah. It's fine. Shall we add a test for this?

gimantha (approved):

Issue #35187, opened 2022-02-22
`array.reduce(..)` is not developer friendly


ayeshLK:
**Description:**
Currently when we are going to invoke `reduce` langlib function on an array we have to do as follows:
```ballerina
int[] values = [80, 85, 90];

int total = values.reduce(function (int a, int b) returns int => a + b, 0);
```

But ideally this should work as follows:
```ballerina
int[] values = [80, 85, 90];

int total = values.reduce((a, b) => a + b, 0);
```

Currently Kotlin [1]  has this feature.

[1] - [https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/reduce.html](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/reduce.html)

hasithaa:
Yes, this has to be improved in the Compiler type checker.

KavinduZoysa:
@hasithaa, are we expecting to remove/reduce types for an example like this?

```ballerina
public function main() {
    int[] values = [80, 85, 90];
    string _ = values.reduce(function (string a, int b) returns string => a + b.toString(), "0");
}
```

KavinduZoysa:
This is OKAY, we have to infer type of `a` from lhs type of statement.

hasithaa:
Ideally, we should be able to infer the type of the lambda function by looking at LHS side. So we should support this, 
```
public function main() {
    int[] values = [80, 85, 90];
    string _ = values.reduce((a,b) => a + b.toString(), "0");
}
```

KavinduZoysa:
We have investigated how to solve this issue. In order to support this, we have to backtrack type matching. But our current typeparam implementation does not support this. So this becomes a design change.

Since there are planned design changes in the future, we will differ this issue.

PR #35460, opened 2022-03-14
Handle typeParam when parameter type is not availabe in function

### LINKED ISSUE #34694: Getting `table<$UndefinedType$>` as  the type when executing "change variable code action"
**Description:**
$title

**Steps to reproduce:**
Try executing the *change variable type* code action for the following code (On the line 
```ballerina
type DataRecord record {|
    string date;
    string state;
    int death;
    int positive;
|};

type DataRow record {|
    time:Utc dateTime;
    string state;
    int death;
    int positive;
|};

function sortTable(table<DataRecord> a) returns table<DataRow> {
    table<DataRow> aa = a.map(function(drecord) returns DataRow {
        return {
            dateTime: [0, 0],
            death: 0,
            positive: 0,
            state: ""
        };
    });
}
```
Screenshot from 2022-01-18 12-06-23
Screenshot from 2022-01-18 12-08-30

**Affected Versions:**
Ballerina 2201.0.0 (Swan Lake) Snapshot build 1

KavinduZoysa:
## Purpose

Fixes #34694

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
@@ -19,6 +19,7 @@ package org.wso2.ballerinalang.compiler.semantics.analyzer;
 
 import io.ballerina.tools.diagnostics.Location;
 import org.ballerinalang.model.Name;
+import org.ballerinalang.model.tree.NodeKind;
 import org.ballerinalang.util.diagnostic.DiagnosticErrorCode;
 import org.wso2.ballerinalang.compiler.diagnostic.BLangDiagnosticLog;
 import org.wso2.ballerinalang.compiler.parser.BLangAnonymousModelHelper;
@@ -53,6 +54,8 @@ import org.wso2.ballerinalang.compiler.semantics.model.types.BType;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BTypedescType;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BUnionType;
 import org.wso2.ballerinalang.compiler.semantics.model.types.BXMLType;
+import org.wso2.ballerinalang.compiler.tree.expressions.BLangExpression;
+import org.wso2.ballerinalang.compiler.tree.expressions.BLangLambdaFunction;
 import org.wso2.ballerinalang.compiler.util.CompilerContext;
 import org.wso2.ballerinalang.compiler.util.ImmutableTypeCloner;
 import org.wso2.ballerinalang.compiler.util.Names;
@@ -129,7 +132,7 @@ public class TypeParamAnalyzer {
         return containsTypeParam(type, new HashSet<>());
     }
 
-    void checkForTypeParamsInArg(Location loc, BType actualType, SymbolEnv env, BType expType) {
+    void checkForTypeParamsInArg(BLangExpression arg, Location loc, BType actualType, SymbolEnv env, BType expType) {
 
         // Not a langlib module invocation
         if (notRequireTypeParams(env)) {
@@ -137,7 +140,7 @@ public class TypeParamAnalyzer {
         }
 
         FindTypeParamResult findTypeParamResult = new FindTypeParamResult();
-        findTypeParam(loc, expType, actualType, env, new HashSet<>(), findTypeParamResult);
+        findTypeParam(arg, loc, expType, actualType, env, new HashSet<>(), findTypeParamResult);
     }
 
     boolean notRequireTypeParams(SymbolEnv env) {
@@ -286,6 +289,16 @@ public class TypeParamAnalyzer {
     private void findTypeParam(Location loc, BType expType, BType actualType, SymbolEnv env,
                                HashSet<BType> resolvedTypes, FindTypeParamResult result, boolean checkContravariance) {
 
+        findTypeParam(null, loc, expType, actualType, env, resolvedTypes, result, checkContravariance);
+    }
+
+    private void findTypeParam(BLangExpression expr, Location loc, BType expType, BType actualType, SymbolEnv env,
+                               HashSet<BType> resolvedTypes, FindTypeParamResult result) {
+        findTypeParam(expr, loc, expType, actualType, env, resolvedTypes, result, false);
+    }
+
+    private void findTypeParam(BLangExpression expr, Location loc, BType expType, BType actualType, SymbolEnv env,
+                               HashSet<BType> resolvedTypes, FindTypeParamResult result, boolean checkContravariance) {
         if (resolvedTypes.contains(expType)) {
             return;
         }
@@ -298,17 +311,17 @@ public class TypeParamAnalyzer {
 
             if (checkContravariance) {
                 types.checkType(loc, getMatchingBoundType(expType, env, new HashSet<>()), actualType,
-                                DiagnosticErrorCode.INCOMPATIBLE_TYPES);
+                        DiagnosticErrorCode.INCOMPATIBLE_TYPES);
             } else {
                 types.checkType(loc, actualType, getMatchingBoundType(expType, env, new HashSet<>()),
-                                DiagnosticErrorCode.INCOMPATIBLE_TYPES);
+                        DiagnosticErrorCode.INCOMPATIBLE_TYPES);
             }
             return;
         }
-        visitType(loc, expType, actualType, env, resolvedTypes, result, checkContravariance);
+        visitType(expr, loc, expType, actualType, env, resolvedTypes, result, checkContravariance);
     }
 
-    private void visitType(Location loc, BType expType, BType actualType, SymbolEnv env,
+    private void visitType(BLangExpression expr, Location loc, BType expType, BType actualType, SymbolEnv env,
                       HashSet<BType> resolvedTypes, FindTypeParamResult result, boolean checkContravariance) {
         // Bound type is a structure. Visit recursively to find bound type.
         switch (expType.tag) {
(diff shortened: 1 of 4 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -408,6 +421,12 @@ private void visitType(Location loc, BType expType, BType actualType, SymbolEnv
                     findTypeParamInInvokableType(loc, (BInvokableType) expType, (BInvokableType) actualType, env,
                             resolvedTypes, result);
                 }
+                if (actualType.tag == TypeTags.SEMANTIC_ERROR) {
+                    if (expr != null && expr.getKind() == NodeKind.LAMBDA) {
```
gimantha, line 425: Can we merge these two` if`'s to one?

langlib/langlib-test/src/test/resources/test-src/type-param/type_param_test_negative.bal
```diff
@@ -137,3 +137,35 @@ function testStringIntFloatSimpleAndArrayUnionReturnParameterNarrowing() {
...
+type DataRecord record {|
+    string name;
+|};
+
+type DataRow record {|
+    string name;
+|};
+
+function foo(table<DataRecord> a) returns table<DataRow> {
+    table<DataRow> aa = a.map(function(dRecord) returns DataRow {
```
gimantha, line 150: Shall we add similar tests for langlib functions with one typeparam as well?

KavinduZoysa: Added more tests

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
KavinduZoysa: Yes, we can, but I think, having it like this will be more readable. It is something like,
```ballerina
if (actualType.tag == TypeTags.SEMANTIC_ERROR) { // type-tag check
        // do something
}
```
WDYT?

MaryamZi (changes requested):
With this change, for something like 

```ballerina
type DataRecord record {|
    string state;
    int death;
    int positive;
|};

type DataRow record {|
    string state;
    int positive;
|};

function sortTable(table<DataRecord> a) {
    table<DataRow> aa = a.map(function(drecord) returns DataRow {
        return {
            positive: 0,
            state: ""
        };
    });
}
```

the difference is that we no longer log

```cmd
incompatible types: expected 'table<DataRow>', found 'table<other> key<never>'
```

Is this what we discussed? IIRC, we discussed looking into the possibility of the RHS type being `table<DataRow>` since we can compute that based on the return type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
MaryamZi: They both read pretty much the same though?

KavinduZoysa:
> With this change, for something like
> 
> ```ballerina
> type DataRecord record {|
>     string state;
>     int death;
>     int positive;
> |};
> 
> type DataRow record {|
>     string state;
>     int positive;
> |};
> 
> function sortTable(table<DataRecord> a) {
>     table<DataRow> aa = a.map(function(drecord) returns DataRow {
>         return {
>             positive: 0,
>             state: ""
>         };
>     });
> }
> ```
> 
> the difference is that we no longer log
> 
> ```batchfile
> incompatible types: expected 'table<DataRow>', found 'table<other> key<never>'
> ```
> 
> Is this what we discussed? IIRC, we discussed looking into the possibility of the RHS type being `table<DataRow>` since we can compute that based on the return type.

@MaryamZi, this error is removed as we identify the rhs type correctly. Can you please check.

chiranSachintha:
```ballerina
type DataRecord record {|
    string name;
|};

type DataRow record {|
    string name;
|};

function foo1(table<DataRecord> a) {
    table<DataRow> aa = a.map(function(dRecord) returns dRecord { 
        return {
            name: ""
        };
    });
}
```
If the return type is not defined still we are getting table<$UndefinedType$> as the type when executing code action. Is it correct?

gimantha:
> ```ballerina
> type DataRecord record {|
>     string name;
> |};
> 
> type DataRow record {|
>     string name;
> |};
> 
> function foo1(table<DataRecord> a) {
>     table<DataRow> aa = a.map(function(dRecord) returns dRecord { 
>         return {
>             name: ""
>         };
>     });
> }
> ```
> 
> If the return type is not defined still we are getting table<$UndefinedType$> as the type when executing code action. Is it correct?

Are you talking about the return type of the lambda function as the argument right? if so, it has to be fixed.

chiranSachintha:
> Are you talking about the return type of the lambda function as the argument right? if so, it has to be fixed.

Yes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -408,6 +421,10 @@ private void visitType(Location loc, BType expType, BType actualType, SymbolEnv
                     findTypeParamInInvokableType(loc, (BInvokableType) expType, (BInvokableType) actualType, env,
                             resolvedTypes, result);
                 }
+                if (actualType.tag == TypeTags.SEMANTIC_ERROR && expr != null && expr.getKind() == NodeKind.LAMBDA) {
```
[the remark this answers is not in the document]
KavinduZoysa: This can be null, please check 
https://github.com/ballerina-platform/ballerina-lang/pull/35460/files#diff-de9cff80105102f5b314d27f1c08c3a65f5034cc0b27e92999d9130bacb6d51cR292

KavinduZoysa:
> ```ballerina
> type DataRecord record {|
>     string name;
> |};
> 
> type DataRow record {|
>     string name;
> |};
> 
> function foo1(table<DataRecord> a) {
>     table<DataRow> aa = a.map(function(dRecord) returns dRecord { 
>         return {
>             name: ""
>         };
>     });
> }
> ```
> 
> If the return type is not defined still we are getting table<$UndefinedType$> as the type when executing code action. Is it correct?

@gimantha, @chiranSachintha As you mentioned there is an issue we need to fix here. Since the return type is not available, we have to get the return type from lhs. But I think this cannot be done based on the current type param resolving logic. This may be a design change. Please check https://github.com/ballerina-platform/ballerina-lang/issues/35187

I think we have to ignore this for now. WDYT?

MaryamZi:
> > ```ballerina
> > type DataRecord record {|
> >     string name;
> > |};
> > 
> > type DataRow record {|
> >     string name;
> > |};
> > 
> > function foo1(table<DataRecord> a) {
> >     table<DataRow> aa = a.map(function(dRecord) returns dRecord { 
> >         return {
> >             name: ""
> >         };
> >     });
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If the return type is not defined still we are getting table<$UndefinedType$> as the type when executing code action. Is it correct?
> 
> @gimantha, @chiranSachintha As you mentioned there is an issue we need to fix here. Since the return type is not available, we have to get the return type from lhs. But I think this cannot be done based on the current type param resolving logic. This may be a design change. Please check this similar issue #35187
> 
> I think we have to ignore this for now. WDYT?

What if we set semantic error as the type if the argument cannot be resolved properly? 

This would mean that in the original example, the type of the RHS won't even be resolved to a table, which IMO is OK and we won't need these special cases. The behaviour will be uniform for all error scenarios also. WDYT?

MaryamZi:
Please also sync the PR with master, in case we are going ahead with this.

KavinduZoysa:
> 

Currently, we are setting the semantic error for the places that cannot resolve type.
We are getting `table<$UndefinedType$>` because we cannot set the type. Therefore I think we have to properly set the type. Please check https://github.com/ballerina-platform/ballerina-lang/issues/35187

But as per your explanation, in order to be uniform with all error scenarios, I also feel that existing behavior is OK.

KavinduZoysa:
> Please also sync the PR with master, in case we are going ahead with this.

Done

gimantha (dismissed):

gimantha:
BTW, with this PR, now the following code works

```ballerina
public function main() {
    EvenNumberGenerator evenGen = new();
    stream evenNumberStream = new(evenGen);
   // stream<int,error?> str = <stream<int,error?>>evenNumberStream;
}

class EvenNumberGenerator {
    int i = 0;
    public isolated function next() returns record {| int value; |}|error? {
        self.i += 2;
        return { value: self.i };
    }
}
```
But if I uncomment the commented line, it crashes at runtime

langlib/langlib-test/src/test/resources/test-src/type-param/type_param_test_negative.bal
```diff
@@ -137,3 +137,42 @@ function testStringIntFloatSimpleAndArrayUnionReturnParameterNarrowing() {
...
+
+public function foo2(int[] ia) {
+    int[] _ = ia.filter(function (x) returns boolean { // error
+        return x > 0; // error
+    });
+}
+
+function testAnonymousFunctionsWithMissingParameterType() {
+    _ = foo1(table []);
+    _ = foo2([1, 2, 3]);
```
MaryamZi, line 165: Why do we need these?

KavinduZoysa: Changed

MaryamZi:
This is patch-compatible, right? Shall we send the PR to https://github.com/ballerina-platform/ballerina-lang/tree/2201.1.x?

KavinduZoysa:
> This is patch-compatible, right? Shall we send the PR to https://github.com/ballerina-platform/ballerina-lang/tree/2201.1.x?

changed

MaryamZi (approved):

PR #35484, opened 2022-03-16
Infer stream parameters when not given

### LINKED ISSUE #34963: A value should belongs to type stream (without the type-parameter) if it has basic type stream
**Description:**
Need to support this in compiler. Currently this supports when completion type is not provided.

```ballerina
stream a = new stream<int, error?>();
```

KRVPerera:
Fixes #34963

## Remarks
Just a quick fix only need to fix this properly

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/types/stream/stream-value.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/types/stream/stream-value.bal
@@ -453,7 +453,6 @@ function testInvalidCast() {
 }
 
 function testBasicStreamType() {
-    boolean testPassed = true;
     stream<int> emptyStream1 = new;
     stream<int> emptyStream2 = new stream<int>();
     stream<int, ()> emptyStream3 = new;
@@ -485,6 +484,29 @@ function testBasicStreamType() {
     assertTrue(a9.next() === ());
 }
 
+function getNextErrorValue(stream errorStream) returns error? {
+    record {error|any value;}|error? val = errorStream.next();
+    if (val is ()) {
+        return;
+    }
+    if (val is error) {
+        return val;
+    }
+    return <error>val.value;
+}
+
+function testErrorStreamTypeAssignedToStreamWithoutParams() {
+    error[] errorList = [];
+    foreach int i in 0 ... 2 {
+        errorList.push(error(i.toBalString()));
+    }
+
+    stream errorStream = errorList.toStream();
+    assertTrue(getNextErrorValue(errorStream) === errorList[0]);
+    assertTrue(getNextErrorValue(errorStream) === errorList[1]);
+    assertTrue(getNextErrorValue(errorStream) === errorList[2]);
+    assertTrue(getNextErrorValue(errorStream) == ());
+}
 
 const ASSERTION_ERROR_REASON = "AssertionError";
 
--- a/tests/jballerina-unit-test/src/test/resources/test-src/types/stream/stream-value.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/types/stream/stream-value.bal
@@ -20,29 +20,29 @@ type ResultValue record {|
 
 class NumberGenerator {
     int i = 0;
-    public isolated function next() returns record {| int value; |}? {
+    public isolated function next() returns record {|int value;|}? {
         self.i += 1;
-        return { value: self.i };
+        return {value: self.i};
     }
 }
 
 class EvenNumberGenerator {
     int i = 0;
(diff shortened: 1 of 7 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangStreamType.java
```diff
@@ -34,6 +34,8 @@
     public BLangType constraint;
     public BLangType error;
 
+    public boolean inferType;
```
MaryamZi, line 37: Shall we check if we can adopt an approach similar to how `future` is handled, instead of introducing this?

KRVPerera: Need to discuss with Chiran

KRVPerera: Looks like I have to come up with a default type which can handle any expected type.

KRVPerera: I think this is bit complex
```
 stream a = new stream<int, error?>();
 stream b = new stream<int>();
 ```
 
 We need to take decisions for completion type therefore we need some boolean/flag to tell us that we need to have a completion type based on RHS. I believe you need to avoid the boolan/flag?

Comment (KRVPerera) [reply compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangStreamType.java]: Had a offline chat with Maryam and resolved this. The way I understood is we need to infer but spec suggests `stream` should accept all possible stream types

**Review by gimantha — APPROVED** [review]
LGTM

**Comment by gimantha** [comment]
BTW, with this PR, now the following code works

```ballerina
public function main() {
    EvenNumberGenerator evenGen = new();
    stream evenNumberStream = new(evenGen);
   // stream<int,error?> str = <stream<int,error?>>evenNumberStream;
}

class EvenNumberGenerator {
    int i = 0;
    public isolated function next() returns record {| int value; |}|error? {
        self.i += 2;
        return { value: self.i };
    }
}
```
But if I uncomment the commented line, it crashes at runtime

tests/jballerina-unit-test/src/test/resources/test-src/types/stream/stream-value.bal
```diff
@@ -451,3 +451,55 @@ function testInvalidCast() {
...
+
+    stream a1 = emptyStream1;
+    stream a2 = emptyStream2;
+    stream a3 = emptyStream3;
+    stream a4 = emptyStream4;
+    stream a5 = emptyStream5;
+    stream a6 = emptyStream6;
+    stream a7 = emptyStream7;
+    stream a8 = emptyStream8;
+    stream a9 = emptyStream9;
```
rdulmina, line 475: Let's add some tests with a stream of errors and stream of unions

KRVPerera:
> BTW, with this PR, now the following code works
> 
> ```ballerina
> public function main() {
>     EvenNumberGenerator evenGen = new();
>     stream evenNumberStream = new(evenGen);
>    // stream<int,error?> str = <stream<int,error?>>evenNumberStream;
> }
> 
> class EvenNumberGenerator {
>     int i = 0;
>     public isolated function next() returns record {| int value; |}|error? {
>         self.i += 2;
>         return { value: self.i };
>     }
> }
> ```
> 
> But if I uncomment the commented line, it crashes at runtime

Somehow I cannot reproduce this. What I get is the below error which is correct.
```LogTalk
error: {ballerina}TypeCastError {"message":"incompatible types: 'stream<(any|error),error?>' cannot be cast to 'stream<int,error?>'"}
	at t4:main(t4.bal:5)
```

tests/jballerina-unit-test/src/test/resources/test-src/types/stream/stream-value.bal
KRVPerera: Added by 0ebbae52a7f95f1bacad73c6d05ae6d955efa5e6

gimantha (approved):

gimantha (approved):

PR #35497, opened 2022-03-18
Introduce symbol visitor and transformer APIs

### LINKED ISSUE #30027: [Semantic API] Provide a type visitor/transformer as a part of the API
Having a type visitor/transformer would be useful to have if the users want to perform some sort of validations on the types. Currently we don't have any such structure available in the API, hence the users resort to arbitrary ways of achieving this. 

Currently when users need to do such analysis where they consider many different scenarios with different types, they either have to use a switch statement or an if-elseif ladder. This is not the best way to go on about doing such analysis since you can easily miss addressing some cases and it's not easy to maintain either. The following is one such example where this visitor/transformer can be used: https://github.com/ballerina-platform/module-ballerina-http/blob/5a955edcc4a9d3212e0fe6ea5c5ad4f8277d8983/compiler-plugin/src/main/java/io/ballerina/stdlib/http/compiler/HttpResourceValidator.java#L143 to L272

pubudu91:
## Purpose
This PR introduces a visitor and a transformer to the Semantic API to be used with symbols. This will allow the users of the API to organize the analysis logic more methodically.

Fixes #30027

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/SymbolTransformer.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/SymbolTransformer.java
@@ -81,8 +81,11 @@ import io.ballerina.compiler.api.symbols.XMLProcessingInstructionTypeSymbol;
 import io.ballerina.compiler.api.symbols.XMLTextTypeSymbol;
 import io.ballerina.compiler.api.symbols.XMLTypeSymbol;
 
+
 /**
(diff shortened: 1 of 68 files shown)


hasithaa (approved):
Design looks good.

PR #35524, opened 2022-03-22
Refactor create function code action

### LINKED ISSUE #33138: `Create function` code action adds a new line unnecessarily to the end of the newly created function
**Description:**

1) Consider the scenario below, where there's a new line after the undefined function.
```
int i = getInt();

string s = "hello world";
```
`Create function` code action adds a new line at the end of the created function `getInt()`, which is not really necessary in this case.

Screenshot 2021-10-08 at 15 46 16

Can we add it only if there's no new line after the undefined function?

For example, in this case it would be necessary.
```
int i = getInt();
string s = "hello world";
```
2) There's an existing check in `CreateFunctionExecutor.java` that checks whether the file ends with a new line, but that doesn't seem to be needed now since it was only relevant when the newly created function was added to the end of the file.

```
// If the file ends with a new line
        boolean newLineAtEnd = rootLineRange.endLine().offset() == 0;
```

Suggestion to refactor the code here.

malinthar:
## Purpose
$subject to  action to conditionally  add a new line at the end of the generated function.

Fixes #33138 

## Approach
Check if the immediate line after the enclosing node is occupied by a module member node.

## Samples

Peek 2022-03-22 12-50

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/CreateFunctionExecutor.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/CreateFunctionExecutor.java
@@ -175,10 +175,8 @@ public class CreateFunctionExecutor implements LSCommandExecutor {
         Optional<NonTerminalNode> enclosingNode = findEnclosingModulePartNode(fnCallExprNode.get());
         Range insertRange;
         if (enclosingNode.isPresent()) {
-            LineRange lineRange = enclosingNode.get().lineRange();
             newLineAtEnd = addNewLineAtEnd(enclosingNode.get());
-            insertRange = new Range(new Position(lineRange.endLine().line(), lineRange.endLine().offset()),
-                    new Position(lineRange.endLine().line(), lineRange.endLine().offset()));
+            insertRange = CommonUtil.toRange(enclosingNode.get().lineRange().endLine());
         } else {
             insertRange = new Range(new Position(endLine, endCol), new Position(endLine, endCol));
         }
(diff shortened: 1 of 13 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/CreateFunctionExecutor.java
```diff
@@ -173,11 +175,10 @@ public Object execute(ExecuteCommandContext context) throws LSCommandExecutorExc
...
         if (enclosingNode.isPresent()) {
-            LineRange endLineRange = enclosingNode.get().lineRange();
-            newLineAtEnd = false;
-            insertRange = new Range(new Position(endLineRange.endLine().line(), endLineRange.endLine().offset()),
-                    new Position(endLineRange.endLine().line(), endLineRange.endLine().offset()));
-
+            LineRange lineRange = enclosingNode.get().lineRange();
+            newLineAtEnd = addNewLineAtEnd(enclosingNode.get());
+            insertRange = new Range(new Position(lineRange.endLine().line(), lineRange.endLine().offset()),
+                    new Position(lineRange.endLine().line(), lineRange.endLine().offset()));
```
nadeeshaan, line 181: Shall we use `CommonUtil.toRange()` here?

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/executors/CreateFunctionExecutor.java
malinthar: We can't use that here, as we want the Range to be starting from the endLine and offset.

nadeeshaan: Then shall we add separate util to accept the line?

malinthar: Done

nadeeshaan (approved):
LGTM

PR #35528, opened 2022-03-22
Update workflows for timestamp builds

udda1996:
## Purpose
$title


.github/workflows/publish_timestamped_release_patch.yml
```diff
@@ -0,0 +1,59 @@
+name: Publish Timestamped Snapshot Artifacts
```
keizer619, line 1: @udda1996 Shall we use existing workflow for both branches? Maybe we can update the current workflow name without mentioning branch name

PR #35537, opened 2022-03-23
Fix sorting in call expression context

### LINKED ISSUE #35343: Improve sorting in call expression context
**Description:**
Consider the following scenario
image

The symbols of `int` type should be sorted to the top of the completion list.

```ballerina

public function main() {
    string a;
    int b;
    int[] arr = [];

    arr.push();
}



```

malinthar:
## Purpose
$subject

Fixes #35343 

## Samples
image

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
@@ -1575,7 +1575,8 @@ public class CommonUtil {
         // We can be in required params or rest params
         if (parameterSymbols.get().size() > argIndex) {
             return Optional.of(parameterSymbols.get().get(argIndex));
-        } else if (restParam.isPresent()) {
+        } 
+        if (restParam.isPresent()) {
             return Optional.of(restParam.get());
         }
         return Optional.empty();
@@ -1640,8 +1641,7 @@ public class CommonUtil {
         Optional<TypeSymbol> typeSymbol = context.currentSemanticModel()
                 .flatMap(semanticModel -> semanticModel.typeOf(methodCallExprNode.expression()));
         return typeSymbol.flatMap(value -> value.langLibMethods().stream()
-                .filter(method -> method.getName().isPresent() &&
-                        method.getName().get().equals(typeRefNode.name().text()))
+                .filter(method -> method.getName().orElse("").equals(typeRefNode.name().text()))
                 .findFirst()
                 .map(FunctionSymbol::typeDescriptor));
     }
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
@@ -19,10 +19,12 @@ import io.ballerina.compiler.api.ModuleID;
 import io.ballerina.compiler.api.symbols.ArrayTypeSymbol;
 import io.ballerina.compiler.api.symbols.ClassSymbol;
 import io.ballerina.compiler.api.symbols.ErrorTypeSymbol;
+import io.ballerina.compiler.api.symbols.FunctionSymbol;
 import io.ballerina.compiler.api.symbols.FunctionTypeSymbol;
 import io.ballerina.compiler.api.symbols.IntersectionTypeSymbol;
 import io.ballerina.compiler.api.symbols.ModuleSymbol;
 import io.ballerina.compiler.api.symbols.ObjectTypeSymbol;
+import io.ballerina.compiler.api.symbols.ParameterKind;
 import io.ballerina.compiler.api.symbols.ParameterSymbol;
 import io.ballerina.compiler.api.symbols.RecordFieldSymbol;
 import io.ballerina.compiler.api.symbols.RecordTypeSymbol;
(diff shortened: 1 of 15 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ContextTypeResolver.java
```diff
@@ -293,7 +293,7 @@ public ContextTypeResolver(PositionedOperationContext context) {
         if (typeRefSymbol.isEmpty()) {
             return Optional.empty();
         }
-        
+
```
nadeeshaan, line 296: I think still the sorting order is not finalized. We get the following, where `b` should be at the top. Are we having the context type resolving issue here for the langlib methods?
Screen Shot 2022-03-24 at 8 56 17 AM

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ContextTypeResolver.java
malinthar: For some reason https://github.com/ballerina-platform/ballerina-lang/blob/02411da0cf01ac4f76c768ed911728b0e0ad115b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MethodCallExpressionNodeContext.java#L111 we have decided to rank the methods/functions first. Should we change this behavior. Is there a specific reason behind this @IMS94

IMS94: @malinthar Seems like the logic should be improved. If we are within the parameter context, we should rank `symbolKind.VARIABLE` higher. Adding a new `else if` condition for `Variables` will do the trick IMO.
```java
                    } else if (symbol.stream().anyMatch(sym -> sym.kind() == SymbolKind.VARIABLE)) {
                        rank = withinParameterCtx ? 1 : 3;
                        break;
                    } else {
```

malinthar: Or we can SortingUtil.toRank() method which ranks the completion items in the standard order

nadeeshaan: My concern here is, since the main method returns `()` while b is already an int variable, why we are not showing the variable `b` at the top.

IMS94: @nadeeshaan We haven't considered assignability when sorting here. That's the issue. If we consider assignability, then this issue will be sorted automatically.

malinthar: I have fixed the sorting order as follows. 
image

IMS94: Can we move the `arr` to bottom? What's the context type we get here? `int[]` or `int`?

malinthar: https://github.com/ballerina-platform/ballerina-lang/blob/95f08123f94dafc0e1a7758917f6f6822b6c83c1/langlib/lang.array/src/main/ballerina/array.bal#L245

As depicted above the first parameter of the push method is of type Type[]. Hence, sorting `arr` to the top is correct in this case IMO.

nadeeshaan: @malinthar not really. What we are looking at here is method access where the method is coming from a standard library. For the standard libraries, the first parameter is being ignored if the types of the first parameter and the expression's type are the same. Hence we have to revamp this sorting

IMS94 (dismissed):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
```diff
@@ -1451,83 +1455,118 @@ private static String getQualifiedModuleName(Module module) {
...
+        if (isLangLibFunction) {
+            argIndex = argIndex + 1;
+        }
+
+        // We can be in required params or rest params
+        if (parameterSymbols.get().size() > argIndex) {
+            return Optional.of(parameterSymbols.get().get(argIndex));
+        } else if (restParam.isPresent()) {
+            return Optional.of(restParam.get());
+        }
```
nadeeshaan, line 1503: We can use `if` clause, instead of `else if`

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
```diff
@@ -1451,83 +1455,118 @@ private static String getQualifiedModuleName(Module module) {
...
+                                                                            PositionedOperationContext context) {
+
+        if (methodCallExprNode.methodName().kind() != SyntaxKind.SIMPLE_NAME_REFERENCE) {
+            return Optional.empty();
+        }
+        SimpleNameReferenceNode typeRefNode = (SimpleNameReferenceNode) methodCallExprNode.methodName();
+        Optional<TypeSymbol> typeSymbol = context.currentSemanticModel()
+                .flatMap(semanticModel -> semanticModel.typeOf(methodCallExprNode.expression()));
+        return typeSymbol.flatMap(value -> value.langLibMethods().stream()
+                .filter(method -> method.getName().isPresent() &&
```
nadeeshaan, line 1566: In a simpler manner, we can use `method.getName().orElse("").equals(...)`

nadeeshaan (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
malinthar: Done

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/CommonUtil.java
malinthar: +1

mohanvive (approved):

PR #35598, opened 2022-03-28
Add `padStart`, `padEnd`, `padZero` langlib functions to `string.bal`

### LINKED ISSUE #35546: Add `padStart` langlib function to string.bal
Please refer to : https://github.com/ballerina-platform/ballerina-spec/commit/a411a67b2216422f685d4fff27baedfcacd06a41

### LINKED ISSUE #35548: Add `padEnd` langlib function to string.bal
Please refer to https://github.com/ballerina-platform/ballerina-spec/commit/a411a67b2216422f685d4fff27baedfcacd06a41

SandaruJayawardana:
Fixes #35546
Fixes #35548

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadEnd.java
```diff
@@ -0,0 +1,52 @@
...
+import io.ballerina.runtime.api.values.BString;
+
+import static io.ballerina.runtime.api.constants.RuntimeConstants.STRING_LANG_LIB;
+import static io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons.LENGTH_SHOULD_BE_LESS_THAN_2147483647;
+import static io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons.getModulePrefixedReason;
+
+/**
+ * Extern function ballerina.model.strings:padStart.
+ *
+ * @since 2201.0.0
```
gimantha, line 32:
```suggestion
 * @since 2201.1.0
```

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadEnd.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        if (len > Integer.MAX_VALUE) {
+            throw ErrorCreator.createError(getModulePrefixedReason(STRING_LANG_LIB,
+                            LENGTH_SHOULD_BE_LESS_THAN_2147483647));
+        }
+
+        int targetLen = (int) len;
+        String pad = padChar.toString().repeat(targetLen - strLength);
+        return StringUtils.fromString(str + pad);
```
gimantha, line 50:
```suggestion
        return StringUtils.fromString(str.concat(pad));
```

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadStart.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        if (len > Integer.MAX_VALUE) {
+            throw ErrorCreator.createError(getModulePrefixedReason(STRING_LANG_LIB,
+                    LENGTH_SHOULD_BE_LESS_THAN_2147483647));
+        }
+
+        int targetLen = (int) len;
+        String pad = padChar.toString().repeat(targetLen - strLength);
+        return StringUtils.fromString(pad + str);
```
gimantha, line 50:
```suggestion
        return StringUtils.fromString(str.concat(pad));
```

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadZero.java
```diff
@@ -0,0 +1,58 @@
...
+        if (len > Integer.MAX_VALUE) {
+            throw ErrorCreator.createError(getModulePrefixedReason(STRING_LANG_LIB,
+                    LENGTH_SHOULD_BE_LESS_THAN_2147483647));
+        }
+
+        int targetLen = (int) len;
+        String pad = padChar.toString().repeat(targetLen - strLength);
+        String result;
+        if (str.substring(0, 1).toString().equals("+") || str.substring(0, 1).toString().equals("-")) {
+            result = str.substring(0, 1) + pad + str.substring(1, strLength);
```
gimantha, line 52: Shall we use a joiner instead of + here? Also please check which one ( using +, joiner or stringBuilders) is more efficient here.

langlib/langlib-test/src/test/resources/test-src/stringlib_test.bal
```diff
@@ -219,6 +219,80 @@ function concatNonBMP(string prefix, string expected) {
...
+    assertEquals("-0-123", y7);
+    string y8 = str2.padZero(5, " ");
+    assertEquals("+ 123", y8);
+    string y9 = str3.padZero(6, "+");
+    assertEquals("-+-123", y9);
+
+    string y10 = str.padZero(30, "-");
+    assertEquals("---------------------------123", y10);
+}
+
```
gimantha, line 295: We need more tests for where we send 
1. empty padding, 
2. default padding
3. Unicode characters (the String being padded and also the Character as a Unicode Character)
4. and runtime Error cases

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadZero.java
gimantha: Also without `substring`ing you can check the index

gimantha: Also adding  `0` after `+` or `-` is useful if the string is a number right? what will happen if there are multiple signs together? does the rule still valid? (`The zeros are added at the start of the string, after a `+` or `-` sign if there is one.`)

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/util/exceptions/BallerinaErrorReasons.java
```diff
@@ -74,6 +74,7 @@
     public static final String KEY_NOT_FOUND_ERROR_IDENTIFIER = "KeyNotFound";
     public static final String INVALID_TYPE_TO_SORT = "SortOperationError";
     public static final String UNORDERED_TYPES = "UnorderedTypesError";
+    public static final String LENGTH_SHOULD_BE_LESS_THAN_2147483647 = "length should be less than 2147483647";
```
KavinduZoysa, line 77: Shall we explain here why it is `2147483647`?
Maybe you can find a more informative error message

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadEnd.java
```diff
@@ -0,0 +1,52 @@
...
+    public static BString padEnd(BString str, long len, BString padChar) {
+        int strLength = str.length();
+        if (len <= strLength) {
+            return str;
+        }
+
+        if (len > Integer.MAX_VALUE) {
+            throw ErrorCreator.createError(getModulePrefixedReason(STRING_LANG_LIB,
+                            LENGTH_SHOULD_BE_LESS_THAN_2147483647));
+        }
```
KavinduZoysa, line 46: I think we do not need this check.
Ballerina `int` represents java `long`. So the user should be able to pass values greater than `Integer.MAX_VALUE`. If the user passes a value more than Long:MAX_VALUE it will be a compile error.

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadEnd.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Extern function ballerina.model.strings:padEnd.
+ *
+ * @since 2201.1.0
+ */
+
+public class PadEnd {
+
+    public static BString padEnd(BString str, long len, BString padChar) {
```
KavinduZoysa, line 37: Have you considered what should happen if the `len < 0`

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadZero.java
```diff
@@ -0,0 +1,58 @@
...
+
+        if (len > Integer.MAX_VALUE) {
+            throw ErrorCreator.createError(getModulePrefixedReason(STRING_LANG_LIB,
+                    LENGTH_SHOULD_BE_LESS_THAN_2147483647));
+        }
+
+        int targetLen = (int) len;
+        String pad = padChar.toString().repeat(targetLen - strLength);
+        String result;
+        if (str.substring(0, 1).toString().equals("+") || str.substring(0, 1).toString().equals("-")) {
```
KavinduZoysa, line 51: Let's move `str.substring(0, 1).toString()` into a seperate variable.

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadEnd.java
SandaruJayawardana: Yeah. According to `If the length of 'str' is >= 'len', returns 'str'`, we have to return `str` without any padding.

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadEnd.java
KavinduZoysa: I think you have done this because the length of the java string cannot exceed Integer:MAX_VALUE.

But by looking at ballerina's point of view use can pass a `len` greater than `Integer:MAX_VALUE`. So I suggest, if the `len` is greater than `Interger.MAX_VALUE`, set len = Interger.MAX_VALUE and continue without throwing an error.

@gimantha WDYT?

gimantha: +1

gimantha: on a second thought, if we ever thought of throwing an error for this java implementation in a future version, your suggestion can be a breaking change.

SandaruJayawardana: Otherwise we can change the error into something like `length greater that '2147483647' not yet supported`?. We are already using these kinds of error msgs for implementation dependent errors. 
ex. 
```ballerina
int[int:MAX_VALUE] x = []; // array length greater that '2147483637' not yet supported
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/util/exceptions/BallerinaErrorReasons.java
SandaruJayawardana: Will update the error msg (https://github.com/ballerina-platform/ballerina-lang/pull/35598#discussion_r837624620).

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadZero.java
SandaruJayawardana: According to the [spec discussion](https://github.com/ballerina-platform/ballerina-spec/issues/770#issuecomment-791258822), 
```
Padding would be done by providing three functions in lang.string with equivalent functionality to Python ljust, rjust and zfill.
```

here `zfill` is equivalent to ballerina `padZero()`. Python defines `zfill` as,

```
Return a copy of the string left filled with ASCII '0' digits to make a string of length width. A leading sign prefix ('+'/'-') is handled by inserting the padding after the sign character rather than before. The original string is returned if width is less than or equal to len(s).
```
In Python, they have clearly mentioned that `A leading sign prefix ('+'/'-') is handled by inserting the padding after the sign character rather than before`. 
Since we are implementing the string padding with equivalent functionality to Python, the current behavior is okay compared to Python.
BTW we can ask to modify the definition of padZero into something like `The zeros are added at the start of the string, after the leading + or - sign if there is one.`?

SandaruJayawardana: Example for zfill(),
image

langlib/lang.string/src/main/ballerina/string.bal
```diff
@@ -267,3 +267,43 @@ public isolated function fromCodePointInt(int codePoint) returns Char|error = @j
...
+# + len - the length of the string to be returned
+# + padChar - the character to use for padding `str`; defaults to a space character
+# + return - `str` padded with `padChar`
+public isolated function padEnd(string str, int len, Char padChar = " ") returns string = @java:Method {
+    'class: "org.ballerinalang.langlib.string.PadEnd",
+    name: "padEnd"
+} external;
+
+# Pads a string with zeros.
+# The zeros are added at the start of the string, after a `+` or `-` sign if there is one.
```
gimantha, line 298: Do we consider the length with or without the + or - sign?

language-server/modules/langserver-core/src/test/resources/completion/field_access_expression_context/config/foreach_stmt_ctx_config13.json
```diff
@@ -1,635 +1 @@
...
... defaults to a space character(Defaultable)  \n  \n**Return** `string`   \n- `str` padded with `padChar`  \n  \n"}},"sortText":"CD","filterText":"padEnd","insertText":"padEnd(${1})","insertTextFormat":"Snippet","command":{"title":"editor.action.triggerParameterHints","command":"editor.action.triggerParameterHints"}},{"label":"padZero(int len, string:Char zeroChar)","kind":"Function","detail":"string","documentation":{"right":{"kind":"markdown","value":"**Package:** _ballerina/lang.string:0.0.0_  \n  \nPads a string with zeros.\nThe zeros are added at the start of the string, after a `+` or `-` sign if there is one.\nSufficient zero characters are added to `str` to make its length be `len`.\nIf the length of `str` is >= `len`, returns `str`.\n  \n**Params**  \n- `int` len: the length of the string to be returned  \n- `string:Char` zeroChar: the character to use for the zero; defaults to ASCII zero `0`(Defaultable)  \n  \n**Return** `string`   \n- `str` padded with zeros  \n  \n"}},"sortText":"CD","filterText":"padZero","insertText":"padZero(${1})","insertTextFormat":"Snippet","command":{"title":"editor.action.triggerParameterHints","command":"editor.action.triggerParameterHints"}}]}
```
gimantha, line 1: reformat this file

langlib/lang.string/src/main/ballerina/string.bal
SandaruJayawardana: We have consider the length with the + or -.  `Sufficient zero characters are added to 'str' to make its length be 'len'`.

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadZero.java
SandaruJayawardana: Addressed in https://github.com/ballerina-platform/ballerina-lang/pull/35598/commits/151962b2b5f5049129028d345e1809d6715a76b0.

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadZero.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Extern function ballerina.model.strings:PadZero.
+ *
+ * @since 2201.1.0
+ */
+
+public class PadZero {
+
+    public static BString padZero(BString str, long len, BString padChar) {
+        int strLength = str.length();
```
gimantha, line 38: if padChar is a empty string, should we proceed with all the below logic? can we not simply return here?

gimantha: check also the other method too.

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadStart.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Extern function ballerina.model.strings:padStart.
+ *
+ * @since 2201.1.0
+ */
+
+public class PadStart {
+
+    public static BString padStart(BString str, long len, BString padChar) {
+        int strLength = str.length();
```
gimantha, line 38: Should we not consider returnning immediately if the padChar is an empty string? we don't need to go through all the logic below right? please check for padEnd also

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadZero.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Extern function ballerina.model.strings:PadZero.
+ *
+ * @since 2201.1.0
+ */
+
+public class PadZero {
+
+    public static BString padZero(BString str, long len, BString padChar) {
```
gimantha, line 37: Should we not give an error if padChar is not a '0' Character?

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadStart.java
gimantha: in fact, if the padChar is empty and the len is larger than the str length, then we can never add padding to satisfy the length. How does python handle this?

SandaruJayawardana: Empty string is not belonging to `string:Char`. Therefore, it should be a compile time error.
```ballerina
string:Char x = ""; // incompatible types: expected 'string:Char', found 'string'
```
I have added few tests for such scenarios
L34, L53, L72 in https://github.com/ballerina-platform/ballerina-lang/pull/35598/files#diff-52b72539d21e3d57324fd2d52853587aeb1e2ad70fa93efec05ad83fa177bf54

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadZero.java
SandaruJayawardana: There are several zero symbols are available in Unicode. Refer https://en.wikipedia.org/wiki/Symbols_for_zero

SandaruJayawardana: Since there are lot of them, it is better to skip this check, isn't it?

gimantha (approved):
LGTM

KavinduZoysa (approved):

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/PadEnd.java
```diff
@@ -0,0 +1,52 @@
...
+        int strLength = str.length();
+        if (len <= strLength) {
+            return str;
+        }
+
+        if (len > Integer.MAX_VALUE) {
+            throw ErrorCreator.createError(getModulePrefixedReason(STRING_LANG_LIB,
+                    LENGTH_GREATER_THAT_2147483647_NOT_YET_SUPPORTED));
```
hasithaa, line 45: Please create an issue to support this.

rdulmina (approved):

PR #35667, opened 2022-04-05
Apply formatting for statements within the partial parser extension 

madushajg:
Fixes https://github.com/wso2-enterprise/internal-support-ballerina/issues/64


misc/ls-extensions/modules/partial-parser/build.gradle
```diff
@@ -28,7 +28,9 @@ dependencies {
     compileOnly project(':ballerina-parser')
     compileOnly project(':ballerina-lang')
     compileOnly project(':diagram-util')
+    compileOnly project(':formatter:formatter-core')
     compileOnly project(':language-server:language-server-commons')
+    compileOnly project(':language-server:language-server-core')
```
IMS94, line 33: It's not recommended to use `language-server-core` as a direct dependency. Shall we remove it?

misc/ls-extensions/modules/partial-parser/src/main/java/io.ballerina.parsers/PartialParserService.java
```diff
@@ -39,24 +45,28 @@
 @JavaSPIService("org.ballerinalang.langserver.commons.service.spi.ExtendedLanguageServerService")
 @JsonSegment("partialParser")
 public class PartialParserService implements ExtendedLanguageServerService {
+    private LSClientLogger clientLogger;
```
IMS94, line 48: I think you wanted ls core to get the client logger. Instead, you should override the `connect(ExtendedLanguageClient client)` method and get the `ExtendedLanguageClient`. Then you can use it to log messages when required.

misc/ls-extensions/modules/partial-parser/src/main/java/io.ballerina.parsers/PartialParserService.java
```diff
@@ -92,4 +106,17 @@
...
     }
+
+    private String getFormattedSource(String statement) {
+
+        String formattedSourceCode = statement;
+
+        try {
+            formattedSourceCode = Formatter.format(statement);
+        } catch (FormatterException e) {
+            this.clientLogger.notifyUser("Formatting", e);
```
IMS94, line 117: `notifyUser()` internally uses `ExtendedLanguageClient#showMessage()` which will show an error popup in VSCode. Is that the correct UX here? What if we just log the error message since you are anyway continuing even after the failure.

IMS94 (changes requested):

IMS94 (approved):

misc/ls-extensions/modules/partial-parser/src/main/java/io.ballerina.parsers/PartialParserService.java
madushajg: Tried this approach and encountered [this](https://github.com/wso2-enterprise/internal-support-ballerina/issues/67).
Will address this once that issue is fixed.

Issue #35675, opened 2022-04-05
Package, Org and Module name should not start with a numeric character


pramodya1994:
**Description:**
According to the [spec](https://ballerina.io/spec/lang/2022R1/#RestrictedIdentifier) package, org and module name should not start with a numeric character. We need to add validations to ensure this.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**
- https://github.com/ballerina-platform/ballerina-lang/issues/35550

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

Issue #35550, opened 2022-03-23
Module name is not validated with project build


prabushi:
**Description:**
According to the Ballerina spec, the module-name should start with an ASCII letter (https://ballerina.io/spec/lang/master/#module-name) However, the module name is not validated via the project build. Also, the `bal add` CLI command doesn't validate the module-name too.

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**
https://github.com/wso2-enterprise/choreo/issues/11360

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #35676, opened 2022-04-05
Avoid org, package & module names with initial numeric characters

pramodya1994:
- Fixes https://github.com/ballerina-platform/ballerina-lang/issues/35675
- Fixes https://github.com/ballerina-platform/ballerina-lang/issues/35550

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/InitCommand.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/InitCommand.java
@@ -144,6 +144,16 @@ public class InitCommand implements BLauncherCmd {
             return;
         }
 
+        if (!ProjectUtils.validateInitialNumericsOfName(packageName)) {
+            CommandUtil.printError(errStream,
+                    "invalid package name : '" + packageName + "' :\n" +
+                            "Package name cannot have initial numeric characters.",
+                    null,
+                    false);
+            CommandUtil.exitError(this.exitWhenFinish);
+            return;
+        }
+
         if (!ProjectUtils.validatePackageName(packageName)) {
             errStream.println("unallowed characters in the project name were replaced by " +
                     "underscores when deriving the package name. Edit the Ballerina.toml to change it.");
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/NewCommand.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/NewCommand.java
@@ -165,6 +165,16 @@ public class NewCommand implements BLauncherCmd {
             return;
         }
 
+        if (!ProjectUtils.validateInitialNumericsOfName(packageName)) {
+            CommandUtil.printError(errStream,
+                    "invalid package name : '" + packageName + "' :\n" +
+                            "Package name cannot have initial numeric characters.",
+                    null,
+                    false);
+            CommandUtil.exitError(this.exitWhenFinish);
+            return;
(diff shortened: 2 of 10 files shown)


cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/InitCommand.java
```diff
@@ -144,6 +144,16 @@ public void execute() {
             return;
         }
 
+        if (!ProjectUtils.validateInitialNumericsOfName(packageName)) {
```
azinneera, line 147: Init command only produces an empty Ballerina.toml. So I think we don't need to do any validations here. Instead, we should replace the disallowed characters when we derive the package name in ManifestBuilder

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/NewCommand.java
```diff
@@ -165,6 +165,16 @@ public void execute() {
             return;
         }
 
+        if (!ProjectUtils.validateInitialNumericsOfName(packageName)) {
+            CommandUtil.printError(errStream,
+                    "invalid package name : '" + packageName + "' :\n" +
+                            "Package name cannot have initial numeric characters.",
+                    null,
+                    false);
+            CommandUtil.exitError(this.exitWhenFinish);
```
azinneera, line 174: We should not fail the project creation. We should replace the disallowed characters and continue after printing a message. See L178-L180

azinneera: For package name validation. the only occasion that should fail is when the user has specified the package name in Ballerina.toml. If the Ballerina.toml is empty, then we should properly derive it to match the allowed pattern.

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/InitCommand.java
pramodya1994: What we should replace with numeric characters? 
Here if the user is giving initial `_`s we give it as an error. 
I followed the same approach for initial numeric characters.

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/NewCommand.java
hevayo: Shouldn't that case also should fail ? `a` is not a good default IMO. We can inform the user that directory name is not a valid package name and they can either the change the directory name or add a valid package name in Ballerina.toml.

azinneera: Just tested and realized that we fail the package creation when there is an invalid initial character. We can do the same for digits as well.

azinneera: @hevayo should we do the same for init command as well? Since we produce an empty Ballerina.toml, should we still fail the command saying the package name is invalid?

pramodya1994:
Added an improved implementation to `master` via https://github.com/ballerina-platform/ballerina-lang/pull/35788

Issue #35691, opened 2022-04-06
Duplicate fields allowed when giving included params as both record and named param


niveathika:
**Description:**
$subject

**Steps to reproduce:**
```
public type Condition record {|
    never 'table?;
    never rowType?;
    anydata...;
|};
remote isolated function getRecord(*Condition condition) {};
```

The above function can be called as below even though the id is duplicated,
```
getRecord({"id":3}, id = 1);
```

**Affected Versions:**
2201.0.x

Issue #3569, opened 2017-10-11
Can we let the user send a request without a resource path.


yasassri:
Note the following example scenario,

   ```
 http:ClientConnector httpConnector = create http:ClientConnector("https://localhost:8080/rh01/verify");
    messages:setJsonPayload(m, jsonMsg);
    message response = httpConnector.post("",m);
```
Is it possible to let the user execute the verb like  `httpConnector.post(m);` ?

**Comment by irunika** [comment]
Since we do not have method overloading in ballerina we cannot have two post actions.

In fact IMO the better approach is to separate address from the URI as in the equivalent to this in v0.94 as in the following example
```
http:ClientConnector httpConnector = create http:ClientConnector("https://localhost:8080", {});
http:Request 
request.setJsonPayload(jsonMsg);
http:Response response = httpConnector.post("/rh01/verify", request);
```

shafreenAnfar:
As per above comment I am closing the issue.

PR #35800, opened 2022-04-12
Disallow named-args that specify field of included-record which is an arg

### LINKED ISSUE #34566: Compiler crashes when arguments are provided for both the parameter and the individual fields of an included record parameter
**Description:**
$title. Should result in a compilation error.

**Steps to reproduce:**
```ballerina
type Configuration record {|
    int id?;
    string name?;
|};

function fn(*Configuration config) {

}

public function main() {
    fn(id = 1, config = {name: "default"}); // no compilation error
}
```

**Affected Versions:**
slbeta6, 2201.0.0 snapshot build 1

KavinduZoysa:
## Purpose

Fixes #34566

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
@@ -178,6 +178,7 @@ public enum DiagnosticErrorCode implements DiagnosticCode {
     INVALID_SPREAD_OP_FIXED_MEMBER_EXPECTED("BCE2134", "invalid.spread.operator.fixed.member.expected"),
     CANNOT_INFER_TYPE_FROM_SPREAD_OP("BCE2135", "cannot.infer.type.from.spread.op"),
     TUPLE_AND_EXPRESSION_SIZE_DOES_NOT_MATCH("BCE2136", "tuple.and.expression.size.does.not.match"),
+    NAMED_ARG_NOT_ALLOWED_WITH_INCLUDED_RECORD_PARAM("BCE2137", "named.arg.not.allowed.with.included.record.param"),
 
     //Transaction related error codes
     ROLLBACK_CANNOT_BE_OUTSIDE_TRANSACTION_BLOCK("BCE2300", "rollback.cannot.be.outside.transaction.block"),
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
@@ -178,8 +178,8 @@ public enum DiagnosticErrorCode implements DiagnosticCode {
     INVALID_SPREAD_OP_FIXED_MEMBER_EXPECTED("BCE2134", "invalid.spread.operator.fixed.member.expected"),
     CANNOT_INFER_TYPE_FROM_SPREAD_OP("BCE2135", "cannot.infer.type.from.spread.op"),
     TUPLE_AND_EXPRESSION_SIZE_DOES_NOT_MATCH("BCE2136", "tuple.and.expression.size.does.not.match"),
-    NAMED_ARG_NOT_ALLOWED_INCLUDED_IN_INCLUDED_RECORD_PARAM("BCE2137",
-            "named.arg.not.allowed.included.in.included.record.param"),
+    DUPLICATE_ARG_VIA_NAMED_ARG_AND_ARG_FOR_INCLUDED_RECORD_PARAM("BCE2137",
+            "duplicate.arg.via.named.argument.and.argument.for.included.record.param"),
 
     //Transaction related error codes
     ROLLBACK_CANNOT_BE_OUTSIDE_TRANSACTION_BLOCK("BCE2300", "rollback.cannot.be.outside.transaction.block"),
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
@@ -178,7 +178,8 @@ public enum DiagnosticErrorCode implements DiagnosticCode {
     INVALID_SPREAD_OP_FIXED_MEMBER_EXPECTED("BCE2134", "invalid.spread.operator.fixed.member.expected"),
     CANNOT_INFER_TYPE_FROM_SPREAD_OP("BCE2135", "cannot.infer.type.from.spread.op"),
     TUPLE_AND_EXPRESSION_SIZE_DOES_NOT_MATCH("BCE2136", "tuple.and.expression.size.does.not.match"),
-    NAMED_ARG_NOT_ALLOWED_WITH_INCLUDED_RECORD_PARAM("BCE2137", "named.arg.not.allowed.with.included.record.param"),
+    NAMED_ARG_NOT_ALLOWED_INCLUDED_IN_INCLUDED_RECORD_PARAM("BCE2137",
+            "named.arg.not.allowed.included.in.included.record.param"),
 
     //Transaction related error codes
     ROLLBACK_CANNOT_BE_OUTSIDE_TRANSACTION_BLOCK("BCE2300", "rollback.cannot.be.outside.transaction.block"),
--- a/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
+++ b/compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
@@ -178,8 +178,8 @@ public enum DiagnosticErrorCode implements DiagnosticCode {
     INVALID_SPREAD_OP_FIXED_MEMBER_EXPECTED("BCE2134", "invalid.spread.operator.fixed.member.expected"),
     CANNOT_INFER_TYPE_FROM_SPREAD_OP("BCE2135", "cannot.infer.type.from.spread.op"),
     TUPLE_AND_EXPRESSION_SIZE_DOES_NOT_MATCH("BCE2136", "tuple.and.expression.size.does.not.match"),
-    DUPLICATE_ARG_VIA_NAMED_ARG_AND_ARG_FOR_INCLUDED_RECORD_PARAM("BCE2137",
-            "duplicate.arg.via.named.argument.and.argument.for.included.record.param"),
+    CANNOT_SPECIFY_NAMED_ARG_FOR_FIELD_OF_INCLUDED_RECORD_WHEN_ARG_SPECIFIED_FOR_INCLUDED_RECORD("BCE2137",
+            "cannot.specify.named.argument.for.field.of.included.record.when.arg.specified.for.included.record"),
 
     //Transaction related error codes
     ROLLBACK_CANNOT_BE_OUTSIDE_TRANSACTION_BLOCK("BCE2300", "rollback.cannot.be.outside.transaction.block"),
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -18,7 +18,6 @@
 package org.wso2.ballerinalang.compiler.semantics.analyzer;
 
 import io.ballerina.identifier.Utils;
-import io.ballerina.runtime.api.flags.SymbolFlags;
 import io.ballerina.tools.diagnostics.DiagnosticCode;
 import io.ballerina.tools.diagnostics.Location;
 import org.ballerinalang.model.TreeBuilder;
@@ -6819,7 +6818,7 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
                 // if arg is positional, corresponding parameter in the same position should be of same type.
                 if (i < nonRestParams.size()) {
                     BVarSymbol param = nonRestParams.get(i);
-                    addToIncRecords(param, incRecords);
+                    populateIncludedRecordParams(param, incRecords);
                     checkTypeParamExpr(arg, param.type, iExpr.langLibInvocation, data);
                     valueProvidedParams.add(param);
                     requiredParams.remove(param);
@@ -6839,7 +6838,7 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
                     dlog.error(arg.pos, DiagnosticErrorCode.UNDEFINED_PARAMETER, argName);
                     break;
                 }
-                addToIncRecords(varSym, incRecords);
+                populateIncludedRecordParams(varSym, incRecords);
                 namedArgs.add(arg);
                 requiredParams.remove(varSym);
                 requiredIncRecordParams.remove(varSym);
@@ -7031,8 +7030,8 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
         }
     }
 
-    private void addToIncRecords(BVarSymbol param, List<BRecordType> incRecords) {
-        if (SymbolFlags.isFlagOn(param.flags, Flags.INCLUDED)) {
+    private void populateIncludedRecordParams(BVarSymbol param, List<BRecordType> incRecords) {
+        if (Symbols.isFlagOn(param.flags, Flags.INCLUDED)) {
             incRecords.add((BRecordType) Types.getReferredType(param.type));
         }
     }
@@ -7044,8 +7043,9 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
             String argName = ((NamedArgNode) namedArg).getName().value;
             for (BRecordType record : incRecords) {
                 if (record.fields.containsKey(argName)) {
-                    dlog.error(namedArg.pos, DiagnosticErrorCode.NAMED_ARG_NOT_ALLOWED_WITH_INCLUDED_RECORD_PARAM,
-                            argName, record.tsymbol.name);
+                    dlog.error(namedArg.pos,
+                            DiagnosticErrorCode.NAMED_ARG_NOT_ALLOWED_INCLUDED_IN_INCLUDED_RECORD_PARAM, argName,
+                            record.tsymbol.name);
                     break;
                 }
             }
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -18,6 +18,7 @@
 package org.wso2.ballerinalang.compiler.semantics.analyzer;
 
 import io.ballerina.identifier.Utils;
+import io.ballerina.runtime.api.flags.SymbolFlags;
 import io.ballerina.tools.diagnostics.DiagnosticCode;
 import io.ballerina.tools.diagnostics.Location;
 import org.ballerinalang.model.TreeBuilder;
@@ -6789,6 +6790,8 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
             }
         }
 
+        List<BRecordType> incRecords = new ArrayList<>();
+        List<BLangExpression> namedArgs = new ArrayList<>();
         int i = 0;
         for (; i < nonRestArgCount; i++) {
             BLangExpression arg = nonRestArgs.get(i);
@@ -6816,6 +6819,7 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
                 // if arg is positional, corresponding parameter in the same position should be of same type.
                 if (i < nonRestParams.size()) {
                     BVarSymbol param = nonRestParams.get(i);
+                    addToIncRecords(param, incRecords);
                     checkTypeParamExpr(arg, param.type, iExpr.langLibInvocation, data);
                     valueProvidedParams.add(param);
                     requiredParams.remove(param);
@@ -6835,6 +6839,8 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
                     dlog.error(arg.pos, DiagnosticErrorCode.UNDEFINED_PARAMETER, argName);
                     break;
                 }
+                addToIncRecords(varSym, incRecords);
+                namedArgs.add(arg);
                 requiredParams.remove(varSym);
                 requiredIncRecordParams.remove(varSym);
                 if (valueProvidedParams.contains(varSym)) {
@@ -6846,7 +6852,7 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
                 valueProvidedParams.add(varSym);
             }
         }
-
+        checkSameNamedArgsInIncRecords(namedArgs, incRecords);
         BVarSymbol restParam = invokableTypeSymbol.restParam;
 
         boolean errored = false;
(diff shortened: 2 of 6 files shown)


MaryamZi:
@KavinduZoysa, please resolve the conflict.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -18,6 +18,7 @@
 package org.wso2.ballerinalang.compiler.semantics.analyzer;
 
 import io.ballerina.identifier.Utils;
+import io.ballerina.runtime.api.flags.SymbolFlags;
```
MaryamZi, line 21: Why are we importing a runtime class? Please fix.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7006,6 +7012,27 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
         }
     }
 
+    private void addToIncRecords(BVarSymbol param, List<BRecordType> incRecords) {
```
MaryamZi, line 7015: We can say 
```suggestion
    private void populateIncludedRecordParams(BVarSymbol param, List<BRecordType> incRecords) {
```

I don't see why we need to shorten `Included` to `Inc`.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1902,3 +1902,6 @@ error.cannot.infer.type.from.spread.op=\
 
 error.tuple.and.expression.size.does.not.match=\
   tuple and expression size does not match
+
+error.named.arg.not.allowed.with.included.record.param=\
+  named-arg ''{0}'' not allowed with included-record-param ''{1}''
```
MaryamZi, line 1907: We shouldn't directly use productions from the grammar, these aren't things we expect the user to know.

MaryamZi: This error currently doesn't describe the actual issue anyway. It reads like we can't specify a named argument when there is an included record parameter in the function.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: Changed, 
`Inc` is used in order to align with other usages of this file.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/resources/compiler.properties
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7024,6 +7029,28 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
...
+        }
+    }
+
+    // If there is a named-arg or positional-arg corresponding to an included-record-param,
+    // it is an error for a named-arg to specify a field of that included-record-param.
+    private void checkSameNamedArgsInIncRecords(List<BLangExpression> namedArgs, List<BRecordType> incRecords) {
+        for (BLangExpression namedArg : namedArgs) {
+            String argName = ((NamedArgNode) namedArg).getName().value;
+            for (BRecordType record : incRecords) {
+                if (record.fields.containsKey(argName)) {
```
MaryamZi, line 7044: This check isn't correct. It disallows the following also, but it should work.

```ballerina
type Rec record {|
    never a?;
    int b;
|};

function fn(string a, *Rec rec) {

}

public function main() {
    fn(a = "", rec = {b: 2});
}
```

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1905,3 +1905,6 @@ error.tuple.and.expression.size.does.not.match=\
 
 error.infer.size.only.supported.in.the.first.dimension=\
   inferred array size is only allowed in the first dimension of an array type descriptor
+
+error.named.arg.not.allowed.included.in.included.record.param=\
+  named arg ''{0}'' is not allowed since it is a filed in ''{1}'' which is an arg
```
MaryamZi, line 1910: Typo - filed -> field.

Like I mentioned previously, error messages need to be as complete and clear as possible, since these are for users. We should use the full form `argument`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Please add a test also.

compiler/ballerina-lang/src/main/resources/compiler.properties
MaryamZi: The error is still not clear IMO. It is not that the named argument is not allowed, it's that you can only specify one of the two.

I would use something like

```suggestion
  duplicate argument for ''{0}'' via named argument and argument for included record parameter
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi, line 7044: Alternatively you can collect all the field names in a list first, to avoid retrieving and iterating through all the fields per named argument.

compiler/ballerina-lang/src/main/resources/compiler.properties
MaryamZi: Please update the diagnostic and the key also to better reflect the error.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6788,6 +6788,8 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
             }
         }
 
+        List<BRecordType> incRecords = new ArrayList<>();
```
MaryamZi, line 6791: Let's use `included` in whatever new construct you're adding.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Can probably update `populateIncludedRecordParams` to do this.

MaryamZi (changes requested):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
chiranSachintha: Here we can use `incRecordParams` list instead of creating a new list.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6834,6 +6837,8 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
                     dlog.error(arg.pos, DiagnosticErrorCode.UNDEFINED_PARAMETER, argName);
                     break;
                 }
+                populateIncludedRecordParams(varSym, incRecords);
+                namedArgs.add(arg);
```
chiranSachintha, line 6841: Without creating a separate list for named arguments can't we refactor the overall logic as follows,
- When visiting a named arg we can check whether the symbol of that argument in the `incRecordParams` list or not

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7024,6 +7029,28 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
...
+    // If there is a named-arg or positional-arg corresponding to an included-record-param,
+    // it is an error for a named-arg to specify a field of that included-record-param.
+    private void checkSameNamedArgsInIncRecords(List<BLangExpression> namedArgs, List<BRecordType> incRecords) {
+        for (BLangExpression namedArg : namedArgs) {
+            String argName = ((NamedArgNode) namedArg).getName().value;
+            for (BRecordType record : incRecords) {
+                if (record.fields.containsKey(argName)) {
+                    dlog.error(namedArg.pos,
+                            DiagnosticErrorCode.NAMED_ARG_NOT_ALLOWED_INCLUDED_IN_INCLUDED_RECORD_PARAM, argName,
+                            record.tsymbol.name);
```
chiranSachintha, line 7047: Consider the below example
```ballerina
type Configuration record {|
    int id?;
    string name = "";
|};

function fn(*Configuration config) {

}

public function main() {
    fn(id = 1, config = {name: "default"});
    // fn(id = 1, name = "");
    // fn({name: "default"}, id = 1);
}

```
With the current approach, we give the error to `id=1`. IMO actual error with the ` config = {name: "default"}`.
Any thoughts @MaryamZi

compiler/ballerina-lang/src/main/resources/compiler.properties
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: This won't work, we need to get the included record params from caller, but `incRecordParams` is created using function signature.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: Changed the logic, please check

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: Added

KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7024,6 +7029,28 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
...
 
+    private void populateIncludedRecordParams(BVarSymbol param, List<BRecordType> incRecords) {
+        if (Symbols.isFlagOn(param.flags, Flags.INCLUDED)) {
+            incRecords.add((BRecordType) Types.getReferredType(param.type));
+        }
+    }
+
+    // If there is a named-arg or positional-arg corresponding to an included-record-param,
+    // it is an error for a named-arg to specify a field of that included-record-param.
+    private void checkSameNamedArgsInIncRecords(List<BLangExpression> namedArgs, List<BRecordType> incRecords) {
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

tests/jballerina-unit-test/src/test/resources/test-src/functions/functions_with_included_record_parameters_negative.bal
```diff
@@ -194,3 +194,67 @@ function functionWithIncludedRecordParam14(int a, *Options options, int c = 10)
...
+    int id?;
+    string name?;
+|};
+
+function fn2(*Record2 record2) {
+
+}
+
+function functionWithIncludedRecordParam16() {
+    fn1(id = 2, record1 = {id: 2, name: ""}); // error
```
rdulmina, line 217: Shouldn't we be using `{id: 2, name: ""}` 's pos for the error pos since value for field `id` is actually duplicated there?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: @chiranSachintha, is your concern the same as https://github.com/ballerina-platform/ballerina-lang/pull/35800#discussion_r874452262? That we need to try and fix to give for the second occurrence onward.

I think your particular example is an interesting case, because the argument for `config` doesn't actually specify a value for `id`. But we don't always know this since the argument for `config` may be a variable reference.

```ballerina
    Configuration config = {name: "default"};
    fn(id = 1, config = config);
```

I think it's OK for it to be an error, but may be we should reword the error for optional fields. E.g.,
```cmd
duplicate arguments may have been specified for 'id' via the named argument and the argument for the included record parameter
```

rdulmina: If the argument for `config` doesn't actually specify a value for `id` then the value for `id` will be nil right? so it's anyway going to duplicate `id=1`?

MaryamZi: No, `id` is an optional field (rather than a nilable field with nil as the default value, in which case the value will be nil).

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6782,12 +6782,16 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
             requiredParams.add(nonRestParam);
         }
 
+        List<String> includedRecordParamNames = new ArrayList<>();
```
KRVPerera, line 6785: Shall we try to populate data structures with the initial sizes? This will improve performance. I believe java already has set this to 16 which is not accurate for our code. It is fine if it is not the exact amount but setting it to the original list size will prevent using extra memory and processor usage shift elements in arrays,

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7024,6 +7031,27 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
...
+    }
+
+    // If there is a named-arg or positional-arg corresponding to an included-record-param,
+    // it is an error for a named-arg to specify a field of that included-record-param.
+    private void checkSameNamedArgsInIncludedRecords(List<BLangExpression> namedArgs, List<String> incRecordFields,
+                                                     List<String> includedRecordParamNames) {
+        incRecordFields.removeIf(field -> !includedRecordParamNames.contains(field));
+        for (BLangExpression namedArg : namedArgs) {
+            String argName = ((NamedArgNode) namedArg).getName().value;
+            if (incRecordFields.contains(argName)) {
```
KRVPerera, line 7048: Since we do searching here I believe a hash set is the more accurate data structure here. Searching on array list is O(n) where we are doing lots of searching here inside another loop making it O(n<sup>2</sup>).

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: Changed the error message

tests/jballerina-unit-test/src/test/resources/test-src/functions/functions_with_included_record_parameters_negative.bal
KavinduZoysa: I think it is matched with the error message we use,
`duplicate arguments may have been specified for ''{0}'' via the named argument and the argument for the \
  included record parameter`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: Changed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: Changed, thank you for the comment

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Seems like you have changed it for the non-optional case also.

```ballerina
type Foo record {
    int id;
    string name;
};

function fn(*Foo record1) {

}

public function main() {
    fn(id = 2, record1 = {id: 2, name: ""});
}
```

Here, `id` is definitely duplicated. So saying "duplicate arguments may have been specified" isn't quite right.

KavinduZoysa: Both optional case and non-optional case both fall into the following sentence in the spec right?
```
If there is a named-arg or positional-arg corresponding to an included-record-param, it is an error for a named-arg to specify a field of that included-record-param. 
```
Therefore error messages should be same right?

MaryamZi: Not necessarily. We need to make sure we are properly communicating what the issue is to the user.

KavinduZoysa: Okay, For this case, you mean having the word `may` is not necessary in `duplicate arguments may have been specified` right?

MaryamZi: Yes. Basically have two different errors for the optional and non-optional case.

Or we can actually say something like

```
cannot specify a named argument for a field of an included record parameter when an argument is specified for the included record parameter
```

This is more in-line with the spec. The errors have to be for the arguments for the fields.

KavinduZoysa: Changed

MaryamZi:
As discussed, let's include the fix for https://github.com/ballerina-platform/ballerina-lang/issues/35691 also.

KavinduZoysa:
> As discussed, let's include the fix for #35691 also.

```ballerina
public type Condition record {|
    never 'table?;
    never rowType?;
    anydata...;
|};
isolated function getRecord(*Condition condition) {};

public function main() {
    getRecord({"id":3}, id = 1);
}
```
@MaryamZi, For the above example mentioned in #3569, I think the error message we should get is, `too many arguments in call to 'getRecord()'`.
Because spec says, 
```
The result of evaluating the i-th positional-arg becomes the i-th member of the argument list. 
The static type of the expression for the i-th positional-arg must be a subtype of the type declared for the i-th parameter 
that is not a rest-param, if there is such a parameter; otherwise, there must be a rest-param and 
the static type must be a subtype of the type declared for the rest-param.
```
The 1st arg which is a positional arg `{"id":3}` matches the includes record param `*Condition condition` and there is not a param for 2nd arg. 
WDYT?

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
```diff
@@ -178,6 +178,8 @@ public enum DiagnosticErrorCode implements DiagnosticCode {
     INVALID_SPREAD_OP_FIXED_MEMBER_EXPECTED("BCE2134", "invalid.spread.operator.fixed.member.expected"),
     CANNOT_INFER_TYPE_FROM_SPREAD_OP("BCE2135", "cannot.infer.type.from.spread.op"),
     TUPLE_AND_EXPRESSION_SIZE_DOES_NOT_MATCH("BCE2136", "tuple.and.expression.size.does.not.match"),
+    CANNOT_SPECIFY_NAMED_ARG_FOR_FIELD_OF_INCLUDED_RECORD_PARAM("BCE2137",
+            "cannot.specify.named.argument.for.field.of.included.record.parameter"),
```
MaryamZi, line 182: This isn't comprehensive and kinda misleading, right? Shall we update this to reflect the actual error?

KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6783,12 +6783,16 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
             requiredParams.add(nonRestParam);
         }
 
+        List<String> includedRecordParamNames = new ArrayList<>(incRecordParams.size());
```
MaryamZi, line 6786: Aren't these actually `includedRecordParamFieldNames`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7025,6 +7037,30 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
         }
     }
 
+    private void populateIncludedRecordParams(BVarSymbol param, HashSet<String> includedRecordFields,
+                                              List<String> includedRecordParamNames) {
+        Set<String> fields = ((BRecordType) Types.getReferredType(param.type)).fields.keySet();
+        fields.removeIf(field -> !includedRecordParamNames.contains(field));
```
MaryamZi, line 7043: This will remove the field from the underlying `BRecordType` also since we are operating on the keys returned by `keySet()`.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/LinkedHashMap.html#keySet()

MaryamZi, line 7043: What's an example for when this will be true?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: Changed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
KavinduZoysa: Changed.

KavinduZoysa: Ex:
```ballerina
type Foo record {
    never a?;
    int b;
};

function callFoo(int a, *Foo foo) returns int {
    return a + foo.b;
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7025,6 +7037,30 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
         }
     }
 
+    private void populateIncludedRecordParams(BVarSymbol param, HashSet<String> includedRecordFields,
+                                              List<String> includedRecordParamNames) {
+        Set<String> fields = new HashSet<>(((BRecordType) Types.getReferredType(param.type)).fields.keySet());
+        fields.removeIf(field -> !includedRecordParamNames.contains(field));
+        includedRecordFields.addAll(fields);
```
MaryamZi, line 7044: Why aren't we just iterating over `((BRecordType) Types.getReferredType(param.type)).fields.keySet()` and directly adding to `includedRecordFields`? We don't need the extra set?

PR #35849, opened 2022-04-19
Fix readonly check for in-built union type

### LINKED ISSUE #35813: Issue when creating a readonly value using `fromJsonStringWithType`
**Description:**
Tried creating a read-only JSON value using the  `fromJsonStringWithType` function. (Ballerina and using the native function). When I checked if the data is readonly it returns `false`

**Steps to reproduce:**

```ballerina
import ballerina/io;

    json j = {a: 1};
    string s = j.toJsonString();
    json & readonly x = check s.fromJsonStringWithType();
    io:println(x is readonly);
}
```

output:

```
Compiling source
	encode.bal
HINT [encode.bal:(7:16,7:29)] unnecessary condition: expression will always evaluate to 'true'

Running executable

false
```

HindujaB:
## Purpose
$subject
Fixes #35813 

## Approach
The readonly parameter gets ignored when we create the json type constant. This causes wrong behavior in the runtime type checker when we create values from the type constant. this PR fixes this. 

## Samples
```ballerina
import ballerina/io;

    json j = {a: 1};
    string s = j.toJsonString();
    json & readonly x = check s.fromJsonStringWithType();
    io:println(x is readonly); // should be true
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BAnydataType.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BAnydataType.java
@@ -50,7 +50,7 @@ public class BAnydataType extends BUnionType implements AnydataType {
     }
 
     public BAnydataType(BUnionType unionType, String typeName, boolean readonly) {
-        super(unionType, typeName);
+        super(unionType, typeName, readonly);
         if (!readonly) {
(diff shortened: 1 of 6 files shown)


warunalakshitha (approved):
LGTM

PR #35908, opened 2022-04-25
Fix non-worker thread calling to `FromJsonWithType()`

### LINKED ISSUE #35780: IllegalStateException when binding json to record from non ballerina thread via FromJsonWithType
**Description:**

Getting `java.lang.IllegalStateException: strand is not accessible from non-strand-worker threads`[1] exception when trying to bind a json payload to a record. This is done in the HTTP transport thread and not in the ballerina thread. 

Stdlibs are in the process of introducing data binding for may modules, and this is a major requirement to bind data and start ballerina thread. Earlier we used `CloneWithType` which worked fine. 

What is the solution for this or an alternative?

Update: If you give same record as an array, it works fine. :) Looks like a bug

[1] https://github.com/ballerina-platform/ballerina-lang/blob/master/langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java#L208

**Steps to reproduce:**

**Affected Versions:**
2201.0.1

HindujaB:
## Purpose
$subject

Fixes #35780

## Approach
The issue happens because the implementation calls the `fromJsonWithType()` before starting a scheduler thread while passing a typedesc.
This PR fixes it by checking whether the scheduler thread is started before calling the `typedesc.instantiate()`.

## Remarks
Since this will result in a call to value creator API to create a record value, it is not guaranteed that the record will initialize properly if it contains any default values in the type description. 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
+++ b/langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/FromJsonWithType.java
@@ -42,6 +42,7 @@ import io.ballerina.runtime.internal.TypeConverter;
 import io.ballerina.runtime.internal.XmlFactory;
 import io.ballerina.runtime.internal.commons.TypeValuePair;
 import io.ballerina.runtime.internal.scheduling.Scheduler;
+import io.ballerina.runtime.internal.scheduling.Strand;
 import io.ballerina.runtime.internal.util.exceptions.BLangExceptionHelper;
 import io.ballerina.runtime.internal.util.exceptions.BallerinaException;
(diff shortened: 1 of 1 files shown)


HindujaB:
Closing as we have sent this fix to patch
https://github.com/ballerina-platform/ballerina-lang/pull/35975

PR #35925, opened 2022-04-26
Fix issue with BRecordType fields in BSameTypeVisitor

### LINKED ISSUE #34986: No error for invalid assignment of record with union-typed fields
**Description:**
$title.

Seems to be an issue with the same type check.

**Steps to reproduce:**
```ballerina
import ballerina/io;

public function main() {
    record {int|string|boolean a; int|boolean b;} r = {a: 1, b: true};
    record {int|string|boolean a; int|string b;} _ = r; // NO ERROR!

    io:println(<any> r is record {int|string|boolean a; int|string b;}); // false, as expected
}
```

**Affected Versions:**
2201.0.0

lochana-chathura:
## Purpose
$subject.

Fixes #34986 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -2609,7 +2609,7 @@ public class Types {
             for (BField sourceField : source.fields.values()) {
                 if (t.fields.containsKey(sourceField.name.value)) {
                     BField targetField = t.fields.get(sourceField.name.value);
-                    if (isSameType(sourceField.type, targetField.type, this.unresolvedTypes) &&
+                    if (isSameType(sourceField.type, targetField.type, new HashSet<>(this.unresolvedTypes)) &&
                             hasSameOptionalFlag(sourceField.symbol, targetField.symbol) &&
                             (!Symbols.isFlagOn(targetField.symbol.flags, Flags.READONLY) ||
                                      Symbols.isFlagOn(sourceField.symbol.flags, Flags.READONLY))) {
@@ -2618,7 +2618,7 @@ public class Types {
                 }
                 return false;
             }
-            return isSameType(source.restFieldType, t.restFieldType, this.unresolvedTypes);
+            return isSameType(source.restFieldType, t.restFieldType, new HashSet<>(this.unresolvedTypes));
         }
 
         private boolean hasSameOptionalFlag(BVarSymbol s, BVarSymbol t) {
--- a/tests/jballerina-unit-test/src/test/resources/test-src/record/record_assignability_negative.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/record/record_assignability_negative.bal
@@ -14,8 +14,22 @@
 // specific language governing permissions and limitations
 // under the License.
 
-function testRecordAssignabilityNegative() {
-    record {int|string|boolean a; int|boolean b;} r = {a: 1, b: true};
-    record {int|string|boolean a; int|string b;} r1 = r;
-    record {int|string|boolean a; int|string b;} _ = r;
+function testOpenRecordAssignabilityNegative() {
+    record {int|string|boolean a; int|boolean b;} r1 = {a: 1, b: true};
+    record {int|string|boolean a; int|string b;} r2 = r1;
+    record {int|string|boolean a; int|string b;} _ = r1;
+}
+
+function testClosedRecordAssignabilityNegative() {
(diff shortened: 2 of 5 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/record/record_assignability_negative.bal
```diff
@@ -0,0 +1,21 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+function testRecordAssignabilityNegative() {
+    record {int|string|boolean a; int|boolean b;} r = {a: 1, b: true};
+    record {int|string|boolean a; int|string b;} r1 = r;
+    record {int|string|boolean a; int|string b;} _ = r;
```
lasinicl, line 20: Shall we also test this with rest descriptor and with closed records too?

lasinicl (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -2609,7 +2609,7 @@ public Boolean visit(BRecordType t, BType s) {
             for (BField sourceField : source.fields.values()) {
                 if (t.fields.containsKey(sourceField.name.value)) {
                     BField targetField = t.fields.get(sourceField.name.value);
-                    if (isSameType(sourceField.type, targetField.type, this.unresolvedTypes) &&
+                    if (isSameType(sourceField.type, targetField.type, new HashSet<>(this.unresolvedTypes)) &&
```
KRVPerera, line 2612: I think this algorithm is logically wrong. The unresolved list can have pairs which are totally different but due to `unresolvedTypes`  cache we say they are the same if found again in the cache. int, int|string|boolean is such a pair in an unresolved list during the original issue. This cache ideally should be utilized to work when there is a cyclic type.
What if we remove pairs from cache if they are not matching. This will also be a slightly better improvement not a good solution. 
eg :  We can change `isSameType` to remove bad pairs in the end.
```Java
        TypePair pair = new TypePair(source, target);
        if (unresolvedTypes.contains(pair)) {
            return true;                                        // THIS IS TRUE FOR CYCLIC TYPES, NOT FOR OTHER TYPES. 
        }
        unresolvedTypes.add(pair);
        BTypeVisitor<BType, Boolean> sameTypeVisitor = new BSameTypeVisitor(unresolvedTypes);
        boolean result = target.accept(sameTypeVisitor, source);
        if (result) {
            return true;
        }
        unresolvedTypes.remove(pair);
        return false;
```

KRVPerera: I agree to look at this complexity later. I will create an issue to track this.

KRVPerera (approved):
Creating a separate issue to look at the existing logic. The change introduced by this PR is valid and correct with the existing logic. 
The main argument is that we are not visiting duplicate pairs unless it is actually a cycle. So making sure this assumption is valid is fix done here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
lochana-chathura: thanks!.

KRVPerera (commented):
Please check this too.

```Ballerina
public function main() {
    boolean|string|int|record {boolean|string b;} r = {b : true};
    boolean|string|int|record {boolean|int b;} _ = r;
}
```

May be we can solve it here https://github.com/ballerina-platform/ballerina-lang/issues/36069

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: New issue https://github.com/ballerina-platform/ballerina-lang/issues/36069

KRVPerera (approved):

pcnfernando (approved):

PR #36006, opened 2022-05-03
Fix throwing ambiguity error for convertible numeric unions

### LINKED ISSUE #35855: Getting an ambiguous error when creating a field with a number union type
**Description:**
Even if the number union field has an array with a specific number type, the ambiguous error returns by this function.

**Steps to reproduce:**
``` ballerina
import ballerina/io;

type Movie record {
    string title;
    int[]|decimal[] year;
    string released;
    Person writer;
};

type Person record {
    string fname;
    string lname;
    int age;
};

    json inceptionJ = {
        title: "Inception",
        year: [2010.0d, 2011.0d],
        released: "16 Jul 2010",
        writer: {
            fname: "Christopher",
            lname: "Nolan",
            age: 30
        }
    };

    Movie cloneWithType = check inceptionJ.cloneWithType(Movie);
    io:println(cloneWithType);
}
```

**Affected Versions:**
Ballerina 2201.0.3

HindujaB:
## Purpose
$subject

Fixes #35855 

## Approach
The issue happens when the type of a numeric array value is a union array/tuple and it needed to be converted into another union. If the target union type contains the exact type of the numeric array it should provide that same typed value even if there are other numeric types in the union members. 
This PR fixes it by keeping track of the union members that do not need numeric conversion through a `Set` and validating it for ambiguity. 

## Samples
```ballerina
    json[] jsonArr = [23.1d];
    string[]|int[]|decimal[] arrayUnionVal = check  jsonArr.cloneWithType(); // result should be decimal[]

    jsonArr = [23.0d, 24];
    int[]|[decimal, int] arrTupleUnionVal = check jsonArr.cloneWithType(); // result should be [decimal, int] 

    jsonArr = [23.0, 24];
    int[]|decimal[]|float[]|error result2 = jsonArr.cloneWithType(); // should throw ambiguity error
}

```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
@@ -263,8 +263,7 @@ public class TypeConverter {
                                                 List<TypeValuePair> unresolvedValues, List<String> errors,
                                                 boolean allowAmbiguity, Set<Type> numericTypeSet, Type parentType) {
         Set<Type> convertibleTypes = new LinkedHashSet<>();
-//        TODO: Add tests for ambiguity
-//        TODO: Check whether this can happen for other union types - record, map etc.
+
         Type inputValueType;
         int targetTypeTag = targetType.getTag();
 
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
@@ -261,7 +261,7 @@ public class TypeConverter {
     // TODO: return only the first matching type
     public static Set<Type> getConvertibleTypes(Object inputValue, Type targetType, String varName, boolean isFromJson,
                                                 List<TypeValuePair> unresolvedValues, List<String> errors,
-                                                boolean allowAmbiguity) {
+                                                boolean allowAmbiguity, List<Type> numericTypeList, Type parentType) {
         Set<Type> convertibleTypes = new LinkedHashSet<>();
 
         Type inputValueType;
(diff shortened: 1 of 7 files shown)


langlib/langlib-test/src/test/resources/test-src/valuelib_test.bal
```diff
@@ -1764,6 +1764,242 @@ function testCloneWithTypeJsonToRecordRestField() {
...
+    assertEquality("'json[]' value cannot be converted to '(int[]|decimal[]|float[])': \n\t	value '[23.0,24]' cannot be converted to '(int[]|decimal[]|float[])': ambiguous target type", <string>checkpanic err.detail()["message"]);
+
+    jVal = {a: [1d, 2.03], b: [2, 3d, 4]};
+    map<int[]|decimal[]>|error result3 = jVal.cloneWithType(); 
+    assertTrue(result3 is error);
+    err = <error>result3;
+    assertEquality("{ballerina/lang.value}ConversionError", err.message());
+    assertEquality("'map<json>' value cannot be converted to 'map<(int[]|decimal[])>': \n\t	value '[1,2.03]' cannot be converted to '(int[]|decimal[])': ambiguous target type\n\t	value '[2,3,4]' cannot be converted to '(int[]|decimal[])': ambiguous target type", <string>checkpanic err.detail()["message"]);
+}
+
```
Nadeeshan96, line 2002: Shall we also add tests with finite types included in the unions

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -275,32 +275,37 @@ public static Set<Type> getConvertibleTypes(Object inputValue, Type targetType,
...
+                    convertibleTypes.addAll(getConvertibleTypes(inputValue, memType, varName, isFromJson,
+                            unresolvedValues, errors, allowAmbiguity, numericTypeSet, targetType));
                 }
-                if (!allowAmbiguity && (convertibleTypes.size() > 1) && !convertibleTypes.contains(inputValueType) &&
-                        !hasIntegerSubTypes(convertibleTypes)) {
-                    errors.subList(initialErrorCount, errors.size()).clear();
-                    addErrorMessage(0, errors, "value '" + getShortSourceValue(inputValue)
-                            + "' cannot be converted to '" + targetType + "': ambiguous target type");
-                    return new LinkedHashSet<>();
+                if (convertibleTypes.size() > 1) {
```
Nadeeshan96, line 281: Shall we check whether we can write like https://github.com/HindujaB/ballerina-lang/pull/1? Shall we also add tests with tuple type rest types, and 2D arrays to check whether recursive calls are working?

Nadeeshan96, line 281: Shall we check whether we can write like https://github.com/HindujaB/ballerina-lang/pull/1? Shall we also add tests with tuple type rest types, and 2D arrays to check whether recursive calls are working?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -275,32 +275,37 @@ public static Set<Type> getConvertibleTypes(Object inputValue, Type targetType,
                     if ((inputValueType == memType) || isIntegerSubtypeAndConvertible(inputValue, memType)) {
```
Nadeeshan96, line 275: I also think rather than doing what is done here and in https://github.com/HindujaB/ballerina-lang/pull/1, because `getConvertibleTypes()` will return all convertible types, and `isConvertible.. ()` methods will return whether they are convertible, rather than tampering with them, need to introduce a proper method here in this if condition which will check whether the given inputValue belongs to the the `memType`

HindujaB: If we are to do that, then that implementation needs to check all the internal elements of the respected type (Eg. array elements, record fields, etc.). As we already have those implementations in `getConvertibleTypes()`,  IMO, it is better to add these validations inside those methods.

HindujaB:
An improved fix is added in https://github.com/ballerina-platform/ballerina-lang/pull/36199

PR #36042, opened 2022-05-05
Validate invalid `UTF-8` values for `fromBytes()` langlib method

### LINKED ISSUE #35931: string:fromBytes doesn't return an error for invalid utf-8 encodings
**Description:**
```ballerina
import ballerina/io;
    // utf-8 sequence in valid range
    int code = 0x1f600; // "😀"
    io:println(code);
    string su = check string:fromCodePointInt(code);
    io:println(su);
    byte[] sub = su.toBytes();
    io:println(<string[]>from var b in sub select toBinary(b)); // codepoint 0x1f600 turned into 4-bytes in utf-8 encoding.
    io:println(check string:fromBytes(sub));
    sub[3] = <byte>(sub[3] + 5); // "😅" = 0x1f605
    io:println(<string[]>from var b in sub select toBinary(b));
    io:println(check string:fromBytes(sub));

    // utf-8 sequence in invalid range
    sub = (check string:fromCodePointInt(0xd7ff)).toBytes(); // one before surrogate range
    io:println(check string:fromBytes(sub));
    io:println(<string[]>from var b in sub select toBinary(b));
    sub[1] = <byte>(sub[1] + 1); // bring into surrogate range (0xd83f)
    io:println(<string[]>from var b in sub select toBinary(b));
    io:println(check string:fromBytes(sub)); // this should result in an error
}

function toBinary(byte b) returns string {
    int shift = 1;
    string result = "";
    foreach int i in 0 ..< 8 {
        result = (((shift & b) == shift) ? "1" : "0") + result;
        shift = shift << 1;
    }
    return result;
}
```

What we seem to be getting is the behavior from Java's utf-8 decoding, which is different from what Ballerina describes in the docs for the langlib string functions.

**Steps to reproduce:**

**Affected Versions:**
Ballerina 2201.0.2 (Swan Lake)
Very likely all the previous versions are also exhibiting the same behavior.

HindujaB:
## Purpose
$subject
Fixes #35931 

## Approach
As the earlier implementation uses the default charset decoder from Java, it replaces the invalid chars with `�` instead of throwing an error. 
This PR fixes it by using a new `CharsetDecoder` where we can catch the exception thrown for invalid chars and provide a proper ballerina error.

## Samples
```ballerina
    bytes = [237, 160, 191];
    string result = check strings:fromBytes(bytes); // should throw error
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/FromBytes.java
```diff
@@ -37,12 +40,19 @@
...
         try {
-            return StringUtils.fromString(new String(bytes.getBytes(), StandardCharsets.UTF_8));
-        } catch (Exception e) {
+            String str = charsetDecoder.decode(ByteBuffer.wrap(bytes.getBytes())).toString();
+            charsetDecoder.reset();
+            return StringUtils.fromString(str);
+        } catch (CharacterCodingException e) {
             return ErrorCreator.createError(StringUtils.fromString("FailedToDecodeBytes"),
-                                            StringUtils.fromString(e.getMessage()));
+                    StringUtils.fromString("byte array contains invalid value"));
```
warunalakshitha, line 55: array contains invalid UTF-8 byte value

gabilang, line 55: Shall we add the error reason and details to `BallerinaErrorReasons` and `RuntimeErrors` respectively.

warunalakshitha (approved):
LGTM

PR #36120, opened 2022-05-12
Fix pruning issue in `block-stmt` parsing

### LINKED ISSUE #34568: Token missing in the parsed ST for the invalid code
**Description:**
Consider the following invalid code.
```bal
public function main() {
    {
        io:println("a");
        io:println("b");
    }
}
```
It is parsed as,
```bal
public function main() {
    {
println("a");
        io:println("b");
    }
}
```
Notice that the first `io` is missing in the generated ST.

lochana-chathura:
## Purpose
$subject.

Fixes #34568 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
+++ b/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
@@ -32,6 +32,7 @@ import io.ballerina.compiler.internal.parser.tree.STDefaultableParameterNode;
 import io.ballerina.compiler.internal.parser.tree.STErrorConstructorExpressionNode;
 import io.ballerina.compiler.internal.parser.tree.STFieldAccessExpressionNode;
 import io.ballerina.compiler.internal.parser.tree.STFunctionArgumentNode;
+import io.ballerina.compiler.internal.parser.tree.STFunctionCallExpressionNode;
 import io.ballerina.compiler.internal.parser.tree.STFunctionSignatureNode;
 import io.ballerina.compiler.internal.parser.tree.STFunctionTypeDescriptorNode;
 import io.ballerina.compiler.internal.parser.tree.STIndexedExpressionNode;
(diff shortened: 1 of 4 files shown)


KRVPerera (commented):
I agree with the parser change. But if the original ballerina code is invalid we should also provide information for the user to actually fix the issue. I think this is not properly defined in spec but looks like the parser may define the inner block as a statement. Then there is also another approach to fixing the semantic analyzer to properly resolve the environment to find `io:println` properly. WDYT?

KRVPerera (approved):

rdulmina:
> I agree with the parser change. But if the original ballerina code is invalid we should also provide information for the user to actually fix the issue. I think this is not properly defined in spec but looks like the parser may define the inner block as a statement. Then there is also another approach to fixing the semantic analyzer to properly resolve the environment to find `io:println` properly. WDYT?

Can u explain with a sample code?

lochana-chathura:
Looks like `block-stmt` was just another statement in our old spec (2020 june - by the time we wrote the new parser) https://ballerina.io/spec/lang/draft/v2020-06-18/#statement and later it has been removed from the spec. According to the current spec, it is invalid and we seem to still support it in our implementation.

lochana-chathura:
>Then there is also another approach to fixing the semantic analyzer to properly resolve the environment to find `io:println` properly. WDYT?

The parsing issue fixed here was only caused when the first statement in the `block-stmt` is `io:println()`.
i.e. The following sample works in GA,
```bal
public function main() {
    {
        int x = 3;
        x += 2;
        io:println(x);
    }
}
```

lochana-chathura:
> Looks like `block-stmt` was just another statement in our old spec (2020 june - by the time we wrote the new parser) https://ballerina.io/spec/lang/draft/v2020-06-18/#statement and later it has been removed from the spec. According to the current spec, it is invalid and we seem to still support it in our implementation.

Currently `block-stmt` is equivalent to the `do-stmt` with `no on-fail-caluse`. That could be the reason for the spec not having a `block-stmt`. Will create an issue for this. We can give `missing 'do' keyword` diagnostic.

pcnfernando (approved):

pcnfernando:
> #34568

Point the issue as a reply

> > Looks like `block-stmt` was just another statement in our old spec (2020 june - by the time we wrote the new parser) https://ballerina.io/spec/lang/draft/v2020-06-18/#statement and later it has been removed from the spec. According to the current spec, it is invalid and we seem to still support it in our implementation.
> 
> Currently `block-stmt` is equivalent to the `do-stmt` with no `on-fail-caluse`. That could be the reason for the spec not having a `block-stmt`. Will create an issue for this. We can give `missing 'do' keyword` diagnostic.

Please point the issue as a reply

lochana-chathura:
> Please point the issue as a reply @lochana-chathura

Created #36187

PR #36291, opened 2022-05-24
Remove the note on identifying asyncapi command as an experimental feature

HasithaAthukorala:


anupama-pathirage:
@HasithaAthukorala Do we need this PR? Looks like it got automatically closed without any attention. If you really don't need this PR anymore, please remove the `stale` label from the PR. Then it won't get added to the `[Stale PR] Ballerina Stale PR Details` mail in the future.

PR #36296, opened 2022-05-24
Remove GraphQL tool experimental feature note

RolandHewage:
The GraphQL tool supports queries, mutations and the 80% use cases. But currently we do not support subscriptions which we do have some plans to bring this support in Update 2.
Based on this reason we thought of adding it as an experimental feature. But since the existing tool works well for the above features, we decided to remove this experimental note as discussed and agreed in the offline discussion.


RolandHewage:
This PR is not necessary since it's already available in the master branch.

PR #36335, opened 2022-05-26
Fix incorrect sorting in statement node contexts

### LINKED ISSUE #36146: Improve completion item sorting in already completed statements
**Description:**
Right now, LS sorts the completion items correctly for partially written statements. Ex:
```ballerina
    int val = get<cursor>
}

function getInt() returns int {
    return 1;
}

function getStr() returns string {
    return "";
}
```
At `<cursor>`, functions that return an int and int type variables will be sorted first. 

But if the user has already completed the statement similar to the below case, the sorting of completion items are not correct.
```ballerina
    int val = get<cursor>Str();
}

function getInt() returns int {
    return 1;
}

function getStr() returns string {
    return "";
}
```
In the above example, `getStr()` will be the 1st while `getInt()` comes 2nd. As per the variable declaration statement, user expect a function/var of type `int` to come first.

**Describe your problem(s)**
See description. 

**Describe your solution(s)**
Need to improve the LS `ContextTypeResolver` to derive the contextually expected type for already completed statements.

malinthar:
## Purpose
$subject by fixing incorrect context type resolution. 

Fixes #36146

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ConstantDeclarationNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ConstantDeclarationNodeContext.java
@@ -67,8 +67,7 @@ public class ConstantDeclarationNodeContext extends NodeWithRHSInitializerProvid
             }
             resolvedContext = ResolvedContext.TYPEDESC;
         } else if (this.onExpressionContext(context, node)) {
-            completionItems.addAll(this.initializerContextCompletions(context,
-                    node.typeDescriptor().orElse(null), node.initializer()));
+            completionItems.addAll(this.initializerContextCompletions(context, node.initializer()));
             resolvedContext = ResolvedContext.EXPRESSION;
         }
         this.sort(context, node, completionItems, resolvedContext);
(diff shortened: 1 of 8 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ConstantDeclarationNodeContext.java
```diff
@@ -105,8 +104,8 @@ public boolean onPreValidation(BallerinaCompletionContext context, ConstantDecla
     }
 
     @Override
-    protected List<LSCompletionItem> initializerContextCompletions(BallerinaCompletionContext context,
-                                                                   Node typeDesc, Node initializer) {
+    protected List<LSCompletionItem> initializerContextCompletions(BallerinaCompletionContext context,
```
mindula, line 107: Shall we remove the extra space after the comma?

mindula, line 107: Shall we remove the extra space after the comma?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ImplicitAnonymousFunctionExpressionNodeContext.java
```diff
@@ -0,0 +1,58 @@
...
+import org.ballerinalang.langserver.commons.completion.LSCompletionException;
+import org.ballerinalang.langserver.commons.completion.LSCompletionItem;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Completion provider for {@link io.ballerina.compiler.syntax.tree.ImplicitAnonymousFunctionExpressionNode} context.
+ *
+ * @since 2.0.0
```
TharushiJay, line 31: Shall we update the version?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ConstantDeclarationNodeContext.java
malinthar: If we do so the length of the line will be greater than 120 chars.

IMS94 (approved):

PR #36341, opened 2022-05-27
Fix spec deviations of `on-conflict-clause`

### LINKED ISSUE #36281: `on-conflict-clause` should support nil type expressions
**Description:**
[Spec says ](https://ballerina.io/spec/lang/2022R1/#on-conflict-clause)
```
on-conflict-clause := on conflict expression
The static type of the expression must be a subtype of error?.
```
But currently on-conflict-clause doesn't support nil type as shown in below.

```ballerina
type Employee record {|
    readonly int id;
    string firstName;
    string lastName;
    int salary;
|};

type ErrorOrNil error|();

ErrorOrNil x = error("123");

public function main() {
    Employee[] employees = [
        {id: 1, firstName: "John", lastName: "Smith", salary: 1000}
    ];

    var highPaidEmployees2 = table key(id) from var e in employees
                            where e.salary >= 1000
                            select e on conflict x; // incompatible types: expected 'error', found 'ErrorOrNil'
                                                    // type 'error' not allowed here; expected an 'error' or a subtype of 'error'.
}
```
**Affected Versions:**
2201.1.0 (Swan Lake)

### LINKED ISSUE #36340: Query expressions don't properly handle duplicate key values when constructing tables
**Description:**
According to the [spec](https://ballerina.io/spec/lang/2022R1/#on-conflict-clause), when the `on-conflict-clause` is missing (or expr of `on-conflict-clause` is `nil`), old value of a duplicate key should be replace with new value. But currently compiler gives an error as shown in below.

```
If the result of evaluating the expression is an error e, then the result of evaluating the query-expr is e. Otherwise, the result must be nil, and the earlier new value replaces the earlier conflicting value, in the same way as if there not been an on conflict clause.
```

```ballerina
import ballerina/io;
type Customer record {|
    readonly int id;
    readonly string name;
    int noOfItems;
|};

type CustomerTable table<Customer> key(id, name);

public function main() {
    Customer c1 = {id: 1, name: "Melina", noOfItems: 12};
    Customer c2 = {id: 1, name: "James", noOfItems: 5};
    Customer c3 = {id: 3, name: "Anne", noOfItems: 20};

    Customer[] customerList = [c1, c2, c3];

    var customerTable1 = table key(id) from var customer in customerList
        select {
            id: customer.id,
            name: customer.name,
            noOfItems: customer.noOfItems
        };

    io:println(customerTable1); // output - error("{ballerina/lang.table}KeyConstraintViolation",message="a value found for key '1'")
}
```

**Affected Versions:**
Current master

SandaruJayawardana:
Fixes #36340
Fixes #36281 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/query/query-expr-with-query-construct-type.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/query/query-expr-with-query-construct-type.bal
@@ -593,6 +593,26 @@ function testTableOnConflict() {
         on conflict onConflictError5;
 
     assertEqual(customerTable5, onConflictError5);
+
+    var customerTable6 = table key(id) from var customer in customerList
+        select {
+            id: customer.id,
+            name: customer.name,
+            noOfItems: customer.noOfItems
+        }
+        on conflict null;
+
+    assertEqual(customerTable6, table key(id) [{"id":1,"name":"James","noOfItems":5},{"id":3,"name":"Anne","noOfItems":20}]);
+
+    var customerTable7 = table key(id) from var customer in customerList
+        select {
+            id: customer.id,
+            name: customer.name,
+            noOfItems: customer.noOfItems
+        }
+        on conflict ();
+
+    assertEqual(customerTable7, table key(id) [{"id":1,"name":"James","noOfItems":5},{"id":3,"name":"Anne","noOfItems":20}]);
 }
 
 function assertEqual(anydata|error actual, anydata|error expected) {
--- a/tests/jballerina-unit-test/src/test/resources/test-src/query/query-expr-with-query-construct-type.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/query/query-expr-with-query-construct-type.bal
@@ -275,7 +275,7 @@ function testSimpleQueryExprReturnTable() returns boolean {
     return testPassed;
 }
 
-function testTableWithDuplicateKeys() returns CustomerTable|error {
+function testTableWithDuplicateKeys() {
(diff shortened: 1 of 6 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/query/query-expr-with-query-construct-type.bal
```diff
@@ -522,3 +522,84 @@ function testQueryConstructingTableUpdateKeyPanic2() returns error? {
...
+
+    assertEqual(customerTable4, onConflictError4);
+
+    var customerTable5 = table key(id) from var customer in customerList
+        select {
+            id: customer.id,
+            name: customer.name,
+            noOfItems: customer.noOfItems
+        }
+        on conflict onConflictError5;
```
KavinduZoysa, line 593: Shall we add a test case with nil literal as on conflict epression?

SandaruJayawardana: Added in https://github.com/ballerina-platform/ballerina-lang/pull/36341/commits/efd65f8590074d58c24bb72ab56de4b8014c1288

KavinduZoysa (approved):

lasinicl (approved):

PR #36367, opened 2022-05-31
Identify distinct error when error is an union of errors

### LINKED ISSUE #36142: Union of distinct error types is working differently
**Description:**

When we have a union of two distinct error types, the resulting error type is working differently for the given two scenarios.

**Steps to reproduce:**

**Scenario 1 :** 

```ballerina
import ballerina/io;

type Error distinct error;

type ClientError distinct Error;
type ListenerError distinct Error;

type GenericClientError distinct ClientError;
type GenericListenerError distinct ListenerError;

type PayloadBindingClientError distinct ClientError;
type PayloadBindingListenerError distinct ListenerError;
type PayloadBindingError PayloadBindingClientError|PayloadBindingListenerError;

function getGenericClientError() returns ClientError {
    return error GenericClientError("Whoops!");
}

function getPayloadBindingError() returns ClientError {
    return error PayloadBindingClientError("Whoops!");
}

    var r = getGenericClientError();
    if r is PayloadBindingError {
        io:println("true");
    } else {
        io:println("false");
    }

    r = getPayloadBindingError();
    if r is PayloadBindingError {
        io:println("true");
    } else {
        io:println("false");
    }
}
```

When I run this code, I get the `false` and then `true`, which is expected.

**Scenario 2 :**
```ballerina
import ballerina/io;

type Error distinct error;

type ClientError distinct Error;
type ListenerError distinct Error;

type GenericClientError distinct ClientError;
type GenericListenerError distinct ListenerError;

type PayloadBindingError distinct ClientError | distinct ListenerError;

function getGenericClientError() returns ClientError {
    return error GenericClientError("Whoops!");
}

function getPayloadBindingError() returns ClientError {
    PayloadBindingError err = error ClientError("Whoops!");
    return <ClientError> err;
}

    var r = getGenericClientError();
    if r is PayloadBindingError {
        io:println("true");
    } else {
        io:println("false");
    }

    r = getPayloadBindingError();
    if r is PayloadBindingError {
        io:println("true");
    } else {
        io:println("false");
    }
}
```

When I run this I am getting the following compiler errors saying that `r` is always a payload binding error, so the condition is unnecessary.
```
HINT [sample.bal:(24:8,24:32)] unnecessary condition: expression will always evaluate to 'true'
ERROR [sample.bal:(27:9,27:29)] unreachable code
HINT [sample.bal:(31:8,31:32)] unnecessary condition: expression will always evaluate to 'true'
ERROR [sample.bal:(34:9,34:29)] unreachable code
```

**Affected Versions:**
Ballerina 2201.1.0 (SwanLake) - RC1

KavinduZoysa:
## Purpose

Related to #36142

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -3672,13 +3672,16 @@ public class Types {
         List<BType> errorTypes = new ArrayList<>(errorTypesSet);
         int size = errorTypes.size();
         Set<BTypeIdSet.BTypeId> t1TypeIds = ((BErrorType) errorTypes.get(0)).typeIdSet.getAll();
+        if (t1TypeIds.size() == 0) {
+            return false;
+        }
         for (int i = 1; i < size; i++) {
             Set<BTypeIdSet.BTypeId> t2TypeIds = ((BErrorType) errorTypes.get(i)).typeIdSet.getAll();
             if (!t1TypeIds.equals(t2TypeIds)) {
                 return false;
             }
         }
-        return t1TypeIds.size() != 0;
+        return true;
     }
 
     public boolean isSelfReferencedStructuredType(BType source, BType s) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -2545,7 +2545,11 @@ public class Types {
     }
 
     public boolean isAllErrorMembers(BUnionType actualType) {
-        return actualType.getMemberTypes().stream().allMatch(t -> isAssignable(t, symTable.errorType));
+        return isAllErrorMembers(actualType.getMemberTypes());
+    }
+
+    public boolean isAllErrorMembers(Set<BType> types) {
+        return !types.isEmpty() && types.stream().allMatch(t -> isAssignable(t, symTable.errorType));
     }
 
     public void setImplicitCastExpr(BLangExpression expr, BType actualType, BType targetType) {
@@ -3546,7 +3550,7 @@ public class Types {
         if (target.tag == TypeTags.UNION) {
             targetIsAUnion = true;
             targetTypes.addAll(getEffectiveMemberTypes((BUnionType) target));
-            allTargetTypesErrors = allTypesErrors(targetTypes);
+            allTargetTypesErrors = isAllErrorMembers(targetTypes);
             if (allTargetTypesErrors) {
                 targetUnionIsDistinctError = containsDistinctError(targetTypes);
             }
@@ -3638,7 +3642,7 @@ public class Types {
                     if (targetUnionIsDistinctError) {
                         sourceTypeIsNotAssignableToAnyTargetType = !isErrorTypeAssignable((BErrorType) sourceMember,
                                 (BErrorType) targetMember, unresolvedTypes);
-                    } else if (targetMember.tag == TypeTags.ERROR && allTargetTypesErrors) {
+                    } else if (allTargetTypesErrors && targetMember.tag == TypeTags.ERROR) {
                         sourceTypeIsNotAssignableToAnyTargetType =
                                 !isErrorTypeAssignableWithNoTypeIds((BErrorType) sourceMember,
                                         (BErrorType) targetMember, unresolvedTypes);
@@ -3674,16 +3678,7 @@ public class Types {
                 return false;
             }
         }
-        return true;
-    }
-
-    private boolean allTypesErrors(Set<BType> memberTypes) {
-        for (BType type : memberTypes) {
-            if (type.tag != TypeTags.ERROR) {
-                return false;
-            }
-        }
-        return !memberTypes.isEmpty();
+        return t1TypeIds.size() != 0;
     }
 
     public boolean isSelfReferencedStructuredType(BType source, BType s) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -1180,10 +1180,9 @@ public class Types {
             return true;
         }
         TypePair pair = new TypePair(source, target);
-        if (unresolvedTypes.contains(pair)) {
+        if (!unresolvedTypes.add(pair)) {
             return true;
         }
-        unresolvedTypes.add(pair);
         return isAssignable(source.detailType, target.detailType, unresolvedTypes);
     }
 
@@ -3549,7 +3548,7 @@ public class Types {
             targetTypes.addAll(getEffectiveMemberTypes((BUnionType) target));
             allTargetTypesErrors = allTypesErrors(targetTypes);
             if (allTargetTypesErrors) {
-                targetUnionIsDistinctError = isUnionDistinctError(targetTypes);
+                targetUnionIsDistinctError = containsDistinctError(targetTypes);
             }
         } else {
             targetTypes.add(target);
@@ -3639,15 +3638,13 @@ public class Types {
                     if (targetUnionIsDistinctError) {
                         sourceTypeIsNotAssignableToAnyTargetType = !isErrorTypeAssignable((BErrorType) sourceMember,
                                 (BErrorType) targetMember, unresolvedTypes);
+                    } else if (targetMember.tag == TypeTags.ERROR && allTargetTypesErrors) {
+                        sourceTypeIsNotAssignableToAnyTargetType =
+                                !isErrorTypeAssignableWithNoTypeIds((BErrorType) sourceMember,
+                                        (BErrorType) targetMember, unresolvedTypes);
                     } else {
-                        if (targetMember.tag == TypeTags.ERROR && allTargetTypesErrors) {
-                            sourceTypeIsNotAssignableToAnyTargetType =
-                                    !isErrorTypeAssignableWithNoTypeIds((BErrorType) sourceMember,
-                                            (BErrorType) targetMember, unresolvedTypes);
-                        } else {
-                            sourceTypeIsNotAssignableToAnyTargetType = !isAssignable(sourceMember, targetMember,
-                                    unresolvedTypes);
-                        }
+                        sourceTypeIsNotAssignableToAnyTargetType = !isAssignable(sourceMember, targetMember,
+                                unresolvedTypes);
                     }
                 } else {
                     sourceTypeIsNotAssignableToAnyTargetType = !isAssignable(sourceMember, targetMember,
@@ -3667,15 +3664,13 @@ public class Types {
         return true;
     }
 
-    private boolean isUnionDistinctError(Set<BType> memberTypesSet) {
-        List<BType> memberTypes = new ArrayList<>(memberTypesSet);
-        int size = memberTypes.size();
-        for (int i = 0; i < size - 1; i++) {
-            BErrorType t1 = (BErrorType) memberTypes.get(i);
-            BErrorType t2 = (BErrorType) memberTypes.get(i + 1);
-            Set<BTypeIdSet.BTypeId> t1TypeIds = t1.typeIdSet.getAll();
-            Set<BTypeIdSet.BTypeId> t2TypeIds = t2.typeIdSet.getAll();
-            if (!t1TypeIds.containsAll(t2TypeIds) || !t2TypeIds.containsAll(t1TypeIds)) {
+    private boolean containsDistinctError(Set<BType> errorTypesSet) {
+        List<BType> errorTypes = new ArrayList<>(errorTypesSet);
+        int size = errorTypes.size();
+        Set<BTypeIdSet.BTypeId> t1TypeIds = ((BErrorType) errorTypes.get(0)).typeIdSet.getAll();
+        for (int i = 1; i < size; i++) {
+            Set<BTypeIdSet.BTypeId> t2TypeIds = ((BErrorType) errorTypes.get(i)).typeIdSet.getAll();
+            if (!t1TypeIds.equals(t2TypeIds)) {
                 return false;
             }
         }
@@ -3688,7 +3683,7 @@ public class Types {
                 return false;
             }
         }
-        return true;
+        return !memberTypes.isEmpty();
     }
 
     public boolean isSelfReferencedStructuredType(BType source, BType s) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -1161,7 +1161,7 @@ public class Types {
         return Symbols.isFlagOn(targetFlags, Flags.READONLY) && !Symbols.isFlagOn(sourceFlags, Flags.READONLY);
     }
 
-    private boolean     isErrorTypeAssignable(BErrorType source, BErrorType target, Set<TypePair> unresolvedTypes) {
+    private boolean isErrorTypeAssignable(BErrorType source, BErrorType target, Set<TypePair> unresolvedTypes) {
         if (target == symTable.errorType) {
             return true;
         }
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
@@ -1161,7 +1161,7 @@ public class Types {
         return Symbols.isFlagOn(targetFlags, Flags.READONLY) && !Symbols.isFlagOn(sourceFlags, Flags.READONLY);
     }
 
-    private boolean isErrorTypeAssignable(BErrorType source, BErrorType target, Set<TypePair> unresolvedTypes) {
+    private boolean     isErrorTypeAssignable(BErrorType source, BErrorType target, Set<TypePair> unresolvedTypes) {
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3733,8 +3752,26 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
                 }
 
-                if (isAssignable(sourceMember, targetMember, unresolvedTypes)) {
-                    sourceTypeIsNotAssignableToAnyTargetType = false;
+                if (sourceMember.tag == TypeTags.ERROR) {
+                    if (targetUnionIsDistinctError) {
+                        sourceTypeIsNotAssignableToAnyTargetType = !isErrorTypeAssignable((BErrorType) sourceMember,
+                                (BErrorType) targetMember, unresolvedTypes);
+                    } else {
+                        if (targetMember.tag == TypeTags.ERROR && allTargetTypesErrors) {
```
KRVPerera, line 3760: Do we need this extra level here. Can be just else if I think.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3747,6 +3784,30 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
         return true;
     }
 
+    private boolean isUnionDistinctError(Set<BType> memberTypesSet) {
+        List<BType> memberTypes = new ArrayList<>(memberTypesSet);
+        int size = memberTypes.size();
+        for (int i = 0; i < size - 1; i++) {
+            BErrorType t1 = (BErrorType) memberTypes.get(i);
+            BErrorType t2 = (BErrorType) memberTypes.get(i + 1);
+            Set<BTypeIdSet.BTypeId> t1TypeIds = t1.typeIdSet.getAll();
```
KRVPerera, line 3793: If we cache the first member, we can get his type id set as a final set and cache it outside for loop also and compare again others. We don't need to call `getAll` many times and also create many temporary objects.
As an example for a 4 element set, this logic will call `getAll` 6 times with duplicate calls. But if we cache first one we will call it only 4 times.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3747,6 +3784,30 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
+            Set<BTypeIdSet.BTypeId> t2TypeIds = t2.typeIdSet.getAll();
+            if (!t1TypeIds.containsAll(t2TypeIds) || !t2TypeIds.containsAll(t1TypeIds)) {
```
KRVPerera, line 3795: I believe we can just use equals
`if (!t1TypeIds.equals(t2TypeIds)) {
 return false;
}
`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1065,6 +1065,19 @@ private boolean isErrorTypeAssignable(BErrorType source, BErrorType target, Set<
                 && target.typeIdSet.isAssignableFrom(source.typeIdSet);
     }
 
+    private boolean isErrorTypeAssignableWithNoTypeIds(BErrorType source, BErrorType target,
+                                                       Set<TypePair> unresolvedTypes) {
+        if (target == symTable.errorType) {
+            return true;
+        }
+        TypePair pair = new TypePair(source, target);
+        if (unresolvedTypes.contains(pair)) {
```
KRVPerera, line 1074:
```suggestion
        if (!unresolvedTypes.add(pair)) {
```
And we can remove `add` command below

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: (e1, e2, e3, e4)
iteration 1 : `e1.getAll, e2.getAll`
iteration 2 : `e2.getAll, e3.getAll`
iteration 3 : `e3.getAll, e4.getAll`

As you can see ` e2.getAll` and `e3.getAll` is repeated.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3747,6 +3784,30 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
         return true;
     }
 
+    private boolean isUnionDistinctError(Set<BType> memberTypesSet) {
```
KRVPerera, line 3787: Shall we rename this method? We are not passing an error to this but a set of type ids. So this is doing some uniqueness check on a type id set right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KRVPerera: I believe this should prevent extra lookups or extra hash calculations. First to check if contains and second to add it again to the set.

KRVPerera (commented):
Added some comments about code quality

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KavinduZoysa: Changed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KavinduZoysa: Changed

KavinduZoysa:
> Added some comments about code quality

Thank you. Addressed all.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3749,6 +3783,29 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
+    private boolean containsDistinctError(Set<BType> errorTypesSet) {
+        List<BType> errorTypes = new ArrayList<>(errorTypesSet);
+        int size = errorTypes.size();
+        Set<BTypeIdSet.BTypeId> t1TypeIds = ((BErrorType) errorTypes.get(0)).typeIdSet.getAll();
+        for (int i = 1; i < size; i++) {
+            Set<BTypeIdSet.BTypeId> t2TypeIds = ((BErrorType) errorTypes.get(i)).typeIdSet.getAll();
+            if (!t1TypeIds.equals(t2TypeIds)) {
+                return false;
+            }
+            t1TypeIds = t2TypeIds;
```
KRVPerera, line 3795: Do we need this line? `t1TypeIds = t2TypeIds`

KRVPerera: I am assuming that we have 
a == b, a == c we can also say b == c kind of logic here. So only comparing to `t1TypeIds` should suffice.

KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3749,6 +3783,28 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
+        List<BType> errorTypes = new ArrayList<>(errorTypesSet);
+        int size = errorTypes.size();
+        Set<BTypeIdSet.BTypeId> t1TypeIds = ((BErrorType) errorTypes.get(0)).typeIdSet.getAll();
+        for (int i = 1; i < size; i++) {
+            Set<BTypeIdSet.BTypeId> t2TypeIds = ((BErrorType) errorTypes.get(i)).typeIdSet.getAll();
+            if (!t1TypeIds.equals(t2TypeIds)) {
+                return false;
+            }
+        }
+        return true;
```
pcnfernando, line 3796: for empty type ids also this will be true; isn't it?

```
type Foo error;
type Bar error;

Foo|Bar e = error Foo("");
```

pcnfernando:
@KavinduZoysa build seems to be failing

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3749,6 +3783,28 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
+    }
+
+    private boolean allTypesErrors(Set<BType> memberTypes) {
+        for (BType type : memberTypes) {
+            if (type.tag != TypeTags.ERROR) {
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3540,8 +3562,24 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
                     break;
                 }
 
-                if (isAssignable(sourceMember, targetMember, unresolvedTypes)) {
-                    sourceTypeIsNotAssignableToAnyTargetType = false;
+                if (sourceMember.tag == TypeTags.ERROR) {
+                    if (targetUnionIsDistinctError) {
+                        sourceTypeIsNotAssignableToAnyTargetType = !isErrorTypeAssignable((BErrorType) sourceMember,
+                                (BErrorType) targetMember, unresolvedTypes);
+                    } else if (targetMember.tag == TypeTags.ERROR && allTargetTypesErrors) {
```
SasinduDilshara, line 3569: its better to use
`} else if (allTargetTypesErrors && targetMember.tag == TypeTags.ERROR) {`

KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -1174,6 +1174,18 @@ private boolean isErrorTypeAssignable(BErrorType source, BErrorType target, Set<
...
     }
 
+    private boolean isErrorTypeAssignableWithNoTypeIds(BErrorType source, BErrorType target,
+                                                       Set<TypePair> unresolvedTypes) {
+        if (target == symTable.errorType) {
+            return true;
+        }
+        TypePair pair = new TypePair(source, target);
+        if (!unresolvedTypes.add(pair)) {
+            return true;
```
pcnfernando, line 1184: Can we try to add a test for this

tests/jballerina-unit-test/src/test/resources/test-src/error/distinct_error_check_negative.bal
```diff
@@ -0,0 +1,40 @@
...
+
+type ClientError distinct Error;
+type ListenerError distinct Error;
+
+type GenericClientError distinct ClientError;
+
+type PayloadBindingClientError distinct ClientError;
+type PayloadBindingListenerError distinct ListenerError;
+type PayloadBindingError PayloadBindingClientError | PayloadBindingListenerError;
+type PayloadBindingError1 distinct ClientError | distinct ListenerError;
```
prakanth97, line 12: Shall we add tests for union of error and distinct error `type PayloadBindingError2 ClientError | distinct ListenerError;` ?

KavinduZoysa: Added

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KavinduZoysa: I have tried a couple of scenarios but still, I was unable to find a test case for this. So I will remove this, If we get an issue later we can consider adding this. WDYT?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3616,8 +3638,24 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
+                    } else {
+                        sourceTypeIsNotAssignableToAnyTargetType = !isAssignable(sourceMember, targetMember,
+                                unresolvedTypes);
+                    }
+                } else {
+                    sourceTypeIsNotAssignableToAnyTargetType = !isAssignable(sourceMember, targetMember,
+                            unresolvedTypes);
+                }
+
+                if (!sourceTypeIsNotAssignableToAnyTargetType) {
```
KRVPerera, line 3658: We are inside a while loop for `targetIterator`. Is it correct to break here if this variable is false? If that is the case IMO it is better to rename the variable. If we break here it is because a source type is not assignable to one of the target types right?

Variable is intended for line 3662 which is after we check a particular source against all the targets.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3616,8 +3638,24 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
                     break;
```
KRVPerera, line 3659: If we break here we are not agreeing with original logic. Here we are inside a while loop which goes through targets.
Then we are breaking when a source type is not assignable to one of the targets.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3630,6 +3668,19 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
+        List<BType> errorTypes = new ArrayList<>(errorTypesSet);
+        int size = errorTypes.size();
+        Set<BTypeIdSet.BTypeId> t1TypeIds = ((BErrorType) errorTypes.get(0)).typeIdSet.getAll();
+        for (int i = 1; i < size; i++) {
+            Set<BTypeIdSet.BTypeId> t2TypeIds = ((BErrorType) errorTypes.get(i)).typeIdSet.getAll();
+            if (!t1TypeIds.equals(t2TypeIds)) {
+                return false;
+            }
+        }
+        return t1TypeIds.size() != 0;
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
KavinduZoysa: Actually, we are breaking here for the negation of source type is not assignable to one of the targets

KRVPerera: But we are still going through a loop to check remaining sources against remaining targets, right? Shouldn't we check this after the loop?

KavinduZoysa: But we are returning early after `!sourceTypeIsNotAssignableToAnyTargetType` right?

KRVPerera: Check the inner loop in line 3610. I believe this logic should be around line 3580

KavinduZoysa:
Closing this PR until runtime changes are done.

KavinduZoysa:
As per this discussion https://github.com/ballerina-platform/ballerina-spec/issues/1177, this fix is not correct. According to the original issue, scenario 2 should be changes according to scenario 1. Therefore this change will be closed.

PR #36368, opened 2022-05-31
Fix create var code action in async send action

### LINKED ISSUE #36078: Add code action for sync send variable assignment
**Description:**

It is better when code action for sync send variable assignment is added.

**Describe your problem(s)**
Given the incomplete code,
```bal
import ballerina/io;
int x = 5;
public function main() {
    func();
    
}

function func() {
    @strand {thread: "any"}
    worker A {
        io:println("worker A start");
        1 ->> B;
        io:println("worker A end");
    }

    @strand {thread: "any"}
    worker B returns error?{
        x-=1;
        if (x>1) {
            return error("err");
        }
        int x1 = <- A;
        io:println(x1);
    }

}
```
Here variable assignment is required to give the line
```bal
error? syncResult = 1 ->> B;
```
image

But no code actions are available.

**Describe your solution(s)**
It is better if this is also available.

IMS94:
## Purpose
$subject

Fixes #36078

## Approach
The diagnostic property includes the `WorkerSymbol` instead of the `FutureTypeSymbol`. This cause a class cast exception. Fixed it.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
@@ -141,7 +141,8 @@ public class CreateVariableCodeAction extends AbstractCodeActionProvider {
      */
     protected Optional<TypeSymbol> getExpectedTypeSymbol(DiagBasedPositionDetails positionDetails) {
         Optional<Symbol> symbol = positionDetails.diagnosticProperty(
-                DiagBasedPositionDetails.DIAG_PROP_VAR_ASSIGN_SYMBOL_INDEX);
+                CodeActionUtil.getDiagPropertyFilterFunction(
+                        DiagBasedPositionDetails.DIAG_PROP_VAR_ASSIGN_SYMBOL_INDEX));
         if (symbol.isEmpty()) {
             return Optional.empty();
         }
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
@@ -76,7 +76,7 @@ public class CreateVariableCodeAction extends AbstractCodeActionProvider {
                                                     DiagBasedPositionDetails positionDetails,
                                                     CodeActionContext context) {
         Optional<TypeSymbol> typeSymbol = getExpectedTypeSymbol(positionDetails);
-        if (typeSymbol.isEmpty() || typeSymbol.get().typeKind() == TypeDescKind.NONE){
+        if (typeSymbol.isEmpty() || typeSymbol.get().typeKind() == TypeDescKind.NONE) {
             return Collections.emptyList();
         }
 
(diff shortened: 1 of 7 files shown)


IMS94:
@dulajdilshan At [`org.wso2.ballerinalang.compiler.semantics.analyzer.CodeAnalyzer#dlog`](https://github.com/ballerina-platform/ballerina-lang/blob/2201.1.x/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java#L4373), we are setting the `WorkerSymbol` instead of `FutureTypeSymbol` as the diagnostic property. In other cases (of variable assignment required), it is setting the type symbol. But only in this case, it's setting the symbol itself. Was this intentional?

At LS, this causes a CCE when we try to get the diagnostic property assuming it's of type `TypeSymbol`.

dulajdilshan:
> @dulajdilshan At [`org.wso2.ballerinalang.compiler.semantics.analyzer.CodeAnalyzer#dlog`](https://github.com/ballerina-platform/ballerina-lang/blob/2201.1.x/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/CodeAnalyzer.java#L4373), we are setting the `WorkerSymbol` instead of `FutureTypeSymbol` as the diagnostic property. In other cases (of variable assignment required), it is setting the type symbol. But only in this case, it's setting the symbol itself. Was this intentional?
> 
> At LS, this causes a CCE when we try to get the diagnostic property assuming it's of type `TypeSymbol`.

LS is set-up to check for `TypeSymbol` when it comes to ASSIGNMENT_REQUIRED?

IMS94:
@dulajdilshan as discussed offline, let's use the existing diagnostic property and proceed.

dulajdilshan (dismissed):
LGTM

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
```diff
@@ -129,6 +131,33 @@ CreateVariableOut getCreateVariableTextEdits(Range range, DiagBasedPositionDetai
...
     }
 
+    /**
+     * Given the position details, this method will determine the expected type symbol for the required variable
+     * assignment from diagnostic properties.
+     *
+     * @param positionDetails Position details
+     * @return Optional expected type symbol
+     */
+    protected Optional<TypeSymbol> getExpectedTypeSymbol(DiagBasedPositionDetails positionDetails) {
```
malinthar, line 141: Alternatively we can extract this function to a `Function<>` and call positionDetails.diagnosticProperty(function). Refer https://github.com/ballerina-platform/ballerina-lang/blob/0e66fbf41e11c71f0d2b0211697f3d43867eb506/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/changetype/ChangeVariableTypeCodeAction.java#L82

IMS94: Done!

Issue #36069, opened 2022-05-09
`isSameType` check blindly returns true for pairs found in `unresolvedTypes` set


KRVPerera:
**Description:**

Check out the below examples. When the algorithm compares fields there are already some pairs added to `unresolvedTypes` set which blindly returns true when we compare field types. Ideally, we only need this set to handle stack overflow in cyclic types. At least we can remove invalid pairs or at least cache them with results of the `isSameType` check.

***Case 1***

Example invalid pair P(x,y) here is when x = `(boolean|string)`, y = `(boolean|int)`

```Ballerina
public function main() {
    boolean|string|int|record {boolean|string b;} r = {b : true};
    boolean|string|int|record {boolean|int b;} _ = r; // NO ERROR!
}
```

![some_pairs](https://user-images.githubusercontent.com/2173530/167360016-32148349-d57d-41ed-ab13-aea493ec2a6f.png)

***Case 2*** 
This is fixed by https://github.com/ballerina-platform/ballerina-lang/pull/35925
Example invalid pair P(x,y) here is when x = `string`, y = `int`

```Ballerina
public function main() {
    record {int|string|boolean a; int|boolean b;} r = {a: 1, b: true};
    record {int|string|boolean a; int|string b;} _ = r; // NO ERROR!
}
```
![first](https://user-images.githubusercontent.com/2173530/167360987-e3e1bb90-f3dd-4e65-8933-77caafae06ac.png)

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**
https://github.com/ballerina-platform/ballerina-lang/issues/34986

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

PR #36374, opened 2022-05-31
Fix erroneous readonly assignability

### LINKED ISSUE #35847: Erroneous readonly assignability
**Description:**
```ballerina
    readonly readonlyVal = 1;
    error? errOrNil = readonlyVal; //should be a compile time assignability error
    io:println(errOrNil);//prints 1 atm
```

**Steps to reproduce:**

**Affected Versions:**
2202.0.2

ushirask:
Fixes #35847

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3689,7 +3689,7 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
                     }
                 }
                 // readonly can match to a union similar to any|error
-                if (sMember.tag == TypeTags.READONLY) {
+                if (sMember.tag == TypeTags.READONLY && isAssignable(symTable.anyOrErrorType, targetUnion)) {
```
MaryamZi, line 3692: While this will work, I'm not sure if this is the best fix. The original issue is because true is returned incorrectly because there is an entry in unresolvedTypes, right? May be related to https://github.com/ballerina-platform/ballerina-lang/issues/36069 in a way too.

@KRVPerera, thoughts please? Since you're familiar with the union type checking logic also.

MaryamZi: Also, `any|error` isn't what is equivalent to `readonly`. It is `(any & readonly)|error`. Any specific reason to check with `symTable.anyOrErrorType`?

ushirask: Assigning to `(any & readonly)|error` is checked prior to this, this code snippet is only used for assigning `readonly` to `any|error` unions right?

MaryamZi: I don't think this check is correct. If `any|error` is a subtype of `targetUnion`, any type is a subtype of `targetUnion`. But, since the value space represented by `readonly` is a proper subset of that of `any|error`, just because `symTable.anyOrErrorType` is not a subtype of `targetUnion`, doesn't mean `readonly` isn't a subtype of `targetUnion`.

Why do we need this block anyway? What was the reason to check specifically with `symTable.anyOrErrorType`?

ushirask: Would it make sense to directly return `true` here if the `targetUnion` is `(any|error)` (regardless of the source type) ?

MaryamZi: That's a check we do in `isAssignable` anyway. I would have thought the check we need to do here is if `sMember.tag == TypeTags.READONLY` check if `(any & readonly)|error` is assignable to `targetUnion`.

@KRVPerera, would you happen to recall what we were trying to address here?

ushirask: We only check in `isAssignable` if `sourceType` doesn't contain an error type right? Without this check 
```
readonly x = 2;
any|error y = x;
```
will fail.

KRVPerera: Since readonly is also a complex union it may not match individual members of the target union but may be assignable to the whole target union as a whole. This is similar to the previous lines checking JSON.

MaryamZi: Didn't we discuss to use `(any & readonly)|error` instead of using `readonly` in that case? The newly added check wouldn't completely fix the original issue, right? Since we'll only be allowing `readonly` with target unions that are `any|error`?

KRVPerera: Yes, Ushira said the original problem is that code is adding readonly, target union pair. Yes, I remember but I think we had a discussion not for this code but for some change after this may be runtime. Need to fix the logic too.

KRVPerera: Yes looks like I have added it in runtime type checker. @ushirask Check runtime type checker too. There is `ANY_AND_READONLY_OR_ERROR_TYPE`

KRVPerera: I think chiran has adde `anyAndReadonlyOrError` to typeChecker too

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3689,7 +3689,7 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
                     }
                 }
                 // readonly can match to a union similar to any|error
-                if (sMember.tag == TypeTags.READONLY) {
+                if (sMember.tag == TypeTags.READONLY && isAssignable(symTable.anyAndReadonlyOrError, targetUnion)) {
                     unresolvedTypes.add(new TypePair(sMember, targetUnion));
                     if (isAssignable(sMember, targetUnion, unresolvedTypes)) {
```
KRVPerera, line 3694: We can remove this second `isAssignable ` check if condition. Keep the logic inside the if block.

ushirask: Fixed.

KRVPerera (commented):
LGTM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3689,12 +3689,10 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
                     }
                 }
                 // readonly can match to a union similar to any|error
-                if (sMember.tag == TypeTags.READONLY) {
+                if (sMember.tag == TypeTags.READONLY && isAssignable(symTable.anyAndReadonlyOrError, targetUnion)) {
```
chiranSachintha, line 3692: The following program should be a compile-time error. right?
```
public function main() {
    readonly readonlyVal = 1;
    error? errOrNil = <readonly>readonlyVal; 
}

```

ushirask: I believe so. It's currently giving a runtime error, I'll check.

ushirask: In a case like
```
readonly x = 2;
int? y = <readonly>x;
```
```
incompatible types: 'readonly' cannot be cast to '(int? & readonly)'
```
, the error is in casting and not assignment, since the casting target type is changed to `int? & readonly`
In the `error` case this cast is currently considered castable, therefore since the type is taken as `error? & readonly`, the assignment is also passing. Should this be a cast error or a assignment error in compiler-time?

MaryamZi: This seems to be a bug. Shall we create a separate issue? The assignment is what should result in an error.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3689,12 +3689,10 @@ private boolean isAssignableToUnionType(BType source, BType target, Set<TypePair
...
                     unresolvedTypes.add(new TypePair(sMember, targetUnion));
```
MaryamZi, line 3693: Do we still need to add this to unresolvedTypes?

ushirask: @KRVPerera mentioned that if we just remove that line, it may cause problems with cyclic types. I'm not sure if that's still the case, I'll check.

MaryamZi (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
MaryamZi: Please mention the issue here.

ushirask: Created in #36580

PR #36388, opened 2022-06-01
Fix Error Messages for Bala Path Not Existing and Invalid Bala path

### LINKED ISSUE #36359: Wrong error message when unable to find bala in the expected path
**Description:**

When publishing a module to central there is a validation of the ball file. However, when the ball file does not exist in the expected location, the error message thrown says ` ballerina: error while validating the Bala file.`

This error message is misleading since the ball file does not exist in the expected path. It should be corrected as `ballerina: unable to find Bala file <EXPECTED_BALA_FILE_PATH>.bala`.

**Affected Versions:**
latest master

MacOS

gayaldassanayake:
## Purpose
`bal push <bala_path>` command takes `bala_path` as the relative or absolute path for bala. However if,
- the given path is incorrect _or_ ,
- the provided file extension is not `bala`,

the error thrown is `ballerina: error while validating the bala file.` which is not matching with the real issues. This PR provides the user with proper error messages.

Fixes #36359 

## Approach

The initial code block led to this situation is,
``` java
try {
    validatePackageMdAndBalToml(balaPath);
    if (!balaPath.toFile().exists()) {
        throw new ProjectException("provided path does not exist.");
    }
    if (!FileUtils.getExtension(balaPath).equals("bala")) {
        throw new ProjectException("file provided is not a bala file.");
    }
} catch (IOException e) {
    throw new ProjectException("error while validating the bala file.", e);
}
```
The aforementioned exceptions to be caught and a generic exception is thrown.

The if clauses are taken out of the `try/catch` block so the user recieves the correct error message.

``` java
if (!balaPath.toFile().exists()) {
    throw new ProjectException("provided path does not exist.");
}
if (!FileUtils.getExtension(balaPath).equals("bala")) {
    throw new ProjectException("file provided is not a bala file.");
}
try {
    validatePackageMdAndBalToml(balaPath);
} catch (IOException e) {
    throw new ProjectException("error while validating the bala file.", e);
}
```
## Samples
Below are the current command line outputs and those of the fix.

- the given path is incorrect
  - Current
    ``` zsh
    $ bal push /no/such/path.bala
    ballerina: error while validating the bala file.
    ```
  - Fix
    ``` zsh
    $ bal push /no/such/path.bala
    ballerina: path provided for the bala file does not exist: /no/such/path.bala.
    ```

- the provided file extension is not `bala`.
  - Current
    ``` zsh
    $ bal push /incorrect/extension.txt
    ballerina: Oh no, something really went wrong. Bad. Sad. (omitted_for_brevity)
    ```
  - Fix
    ``` zsh
    $ bal push /incorrect/extension.txt
    ballerina: file provided is not a bala file: /incorrect/extension.txt.
    ```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/PushCommandTest.java
```diff
@@ -71,12 +71,28 @@ public void setup() throws IOException {
...
         PushCommand pushCommand = new PushCommand(validBalProject, printStream, printStream, false);
-        new CommandLine(pushCommand).parse("tests");
+        String invalidPath = "tests";
+        new CommandLine(pushCommand).parse(invalidPath);
         pushCommand.execute();
 
         String buildLog = readOutput(true);
         String actual = buildLog.replaceAll("\r", "");
-        String expected = "file provided is not a bala file.";
+        String expected = "path provided for the bala file does not exist: " + invalidPath + ".";
```
pramodya1994, line 80: If we think of this from a more user perspective,
1. User will run `bal pack` command
2. After completion user will run `bal push`
3. Meantime if somehow bala is not in the `target/bala` folder, This error happens.
`path provided for the bala file does not exist: target/bala/winery-lib-1.0.0.bala`
Since the user is not giving any path, is this the best error message we can show?
@azinneera WDYT?

pramodya1994: Seems this error only shows when the user has given the bala path.
Hense resolving the above comment.

pramodya1994 (approved):

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/PushCommand.java
```diff
@@ -325,6 +306,8 @@ private static void validatePackageMdAndBalToml(Path balaPath) throws IOExceptio
                     return;
                 }
             }
+        } catch (IOException e) {
+            throw new ProjectException("error while validating the bala file.", e);
```
azinneera, line 310: This will again hide the original cause of the exception. Let's append the message of the orginal IOExceptioh here.

```suggestion
            throw new ProjectException("error while validating the bala file: " + e.getMessage(), e);
```

azinneera: Please check for other places that this happens.

azinneera (dismissed):

azinneera (approved):

PR #36473, opened 2022-06-08
[Debugger] Fix additional fields not retrieved for open records

### LINKED ISSUE #36444: [Debugger] Additional fields of open records not shown in variable window
**Description:**
$subject. Refer screenshot below

Screenshot 2022-06-06 at 14 41 52

**Steps to reproduce:**

- Create an open record with additional fields
- Debug the ballerina program and track the variables on the variable window

SumayyaZiyad:
Fixes #36444 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/variables/VariableVisibilityTest.java
+++ b/tests/jballerina-debugger-integration-test/src/test/java/org/ballerinalang/debugger/test/adapter/variables/VariableVisibilityTest.java
@@ -69,8 +69,8 @@ public class VariableVisibilityTest extends BaseTestCase {
 
     @Test(description = "Variable visibility test in the middle of the main() method for a new variable")
     public void newVariableVisibilityTest() throws BallerinaTestException {
-        debugTestRunner.addBreakPoint(new BallerinaTestDebugPoint(debugTestRunner.testEntryFilePath, 243));
-        debugTestRunner.addBreakPoint(new BallerinaTestDebugPoint(debugTestRunner.testEntryFilePath, 313));
+        debugTestRunner.addBreakPoint(new BallerinaTestDebugPoint(debugTestRunner.testEntryFilePath, 245));
+        debugTestRunner.addBreakPoint(new BallerinaTestDebugPoint(debugTestRunner.testEntryFilePath, 315));
         debugTestRunner.initDebugSession(DebugUtils.DebuggeeExecutionKind.RUN);
         debugHitInfo = debugTestRunner.waitForDebugHit(25000);
         localVariables = debugTestRunner.fetchVariables(debugHitInfo.getRight(), DebugTestRunner.VariableScope.LOCAL);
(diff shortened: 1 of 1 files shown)


misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/variable/types/BRecord.java
```diff
@@ -55,34 +64,67 @@ public Map<String, Value> computeChildVariables() {
...
-                    .filter(field -> field.toString().contains(RECORD_FIELD_PATTERN_IDENTIFIER))
-                    .count();
-
-            return Long.valueOf(recordFieldCount).intValue();
+            Optional<Method> getKeysMethod = VariableUtils.getMethod(jvmValue,
+                                                                    METHOD_GET_KEYS,
+                                                                    GETKEYS_METHOD_SIGNATURE_PATTERN);
+            Value keyArray = ((ObjectReference) jvmValue).invokeMethod(context.getOwningThread().getThreadReference(),
+                    getKeysMethod.get(), Collections.emptyList(), ObjectReference.INVOKE_SINGLE_THREADED);
+            return ((ArrayReference) keyArray).length();
```
NipunaRanasinghe, line 127: Can we store this `keyArray` in the `loadedKeys` field, so that we can load the keys only if the `loadedKeys` is null inside the `getRecordFields()` method? We can avoid retrieving the same keyset in multiple times in that way.

NipunaRanasinghe: or we can write a `loadAllKeys()` method and re-use it in both places

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/variable/types/BRecord.java
```diff
@@ -55,34 +64,67 @@ public Map<String, Value> computeChildVariables() {
...
         } catch (Exception ignored) {
             return new LinkedHashMap<>();
         }
     }
 
+    private Map<Value, Value> getRecordFields() {
+        try {
+            Optional<Method> getKeysMethod = VariableUtils.getMethod(jvmValue,
+                                                                    METHOD_GET_KEYS,
+                                                                    GETKEYS_METHOD_SIGNATURE_PATTERN);
```
NipunaRanasinghe, line 84: Do we need to keep separate lines for each argument as we only have 3 args in here? :)

NipunaRanasinghe:
@SumayyaZiyad Shall we also add a test case to cover the child variable visibility for open records with additional fields? 

IMO we can update the existing `org.ballerinalang.debugger.test.adapter.variables.VariableVisibilityTest#localVariableChildrenVisibilityTest()` test case

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/variable/types/BRecord.java
```diff
@@ -55,34 +64,72 @@ public Map<String, Value> computeChildVariables() {
...
         } catch (Exception ignored) {
             return new LinkedHashMap<>();
         }
     }
 
+    private Map<Value, Value> getRecordFields() {
+        try {
+            if (loadedKeys == null) {
+                loadAllKeys();
+            }
```
NipunaRanasinghe, line 84: Shall we move this check inside to `loadAllKeys()` method instead? Otherwise we'll have to use the check whenever we call this method

misc/debug-adapter/modules/debug-adapter-core/src/main/java/org/ballerinalang/debugadapter/variable/types/BRecord.java
```diff
@@ -55,34 +64,72 @@ public Map<String, Value> computeChildVariables() {
...
-            long recordFieldCount = fieldValueMap.keySet()
-                    .stream()
-                    .filter(field -> field.toString().contains(RECORD_FIELD_PATTERN_IDENTIFIER))
-                    .count();
-
-            return Long.valueOf(recordFieldCount).intValue();
+            if (loadedKeys == null) {
+                loadAllKeys();
+            }
+            return loadedKeys.length();
```
NipunaRanasinghe, line 132: Here we may have to use an null check for `loadedKeys` because  the variable can be null if any exception is thrown while loading the keys (L106)

NipunaRanasinghe (approved):
LGTM

PR #36491, opened 2022-06-09
Sort documents of a module

### LINKED ISSUE #31082: Dependency order between packages are not consistent.
**Description:**
We found out that dependency resolution between packages is not consistent between compilations while fixing https://github.com/ballerina-platform/ballerina-lang/issues/31009.

The dependency order is correct but different. 

**Steps to reproduce:**

**Affected Versions:**

nipunayf:
## Purpose
The order of the documents is not consistent at project creation or when adding a new module. This list should retain the sort at each build.

Fixes #31082 

## Approach
Sort the `DocumentId`s based on the document path when extracting the keyset from the context map.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/ModuleContext.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/ModuleContext.java
@@ -44,8 +44,10 @@ import org.wso2.ballerinalang.programfile.PackageFileWriter;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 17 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/projects/ModuleContext.java
```diff
@@ -556,6 +558,15 @@ ModuleContext duplicate(Project project) {
                 this.moduleDescDependencies, this.resourceContextMap, this.testResourceContextMap);
     }
 
+    private Collection<DocumentId> sortDocuments(Map<DocumentId, DocumentContext> docContextMap) {
```
azinneera, line 561: We create a new ModuleContext for every document edit. I think this would affect the LS since it will sort for every edit request. IMO, we can specifically sort the map when we create a package and when a document is added. @sameerajayasoma WDYT?

nipunayf: Addressed this in db53c289c2f4ec33ec8d539e58b4877ffdf0a0f7

azinneera (approved):
LGTM

sameerajayasoma (approved):

PR #36492, opened 2022-06-09
Invalidate the environment in project when the compiler crashes

### LINKED ISSUE #36231: Redesign the compiler context to handle compiler crashes
**Description:**
$subject

Related Issue: https://github.com/ballerina-platform/ballerina-lang/issues/35158

azinneera:
Fixes #36231

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/Project.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/Project.java
@@ -18,7 +18,6 @@
 package io.ballerina.projects;
 
 import io.ballerina.projects.environment.ProjectEnvironment;
-import org.wso2.ballerinalang.compiler.PackageCache;
 import org.wso2.ballerinalang.compiler.util.CompilerContext;
 import org.wso2.ballerinalang.compiler.util.CompilerOptions;
 
@@ -36,7 +35,7 @@ public abstract class Project {
     protected final Path sourceRoot;
     private Package currentPackage;
     private BuildOptions buildOptions;
-    private final ProjectEnvironment projectEnvironment;
+    protected ProjectEnvironment projectEnvironment;
     private final ProjectKind projectKind;
 
     protected Project(ProjectKind projectKind,
@@ -110,13 +109,7 @@ public abstract class Project {
      * (i.e. package resolution caches, compilation caches)
      * generated during project compilation will be discarded.
      */
-    public void clearCaches() {
-        cloneProject(this);
-        CompilerContext compilerContext = this.projectEnvironmentContext()
-                .getService(CompilerContext.class);
-        PackageCache packageCache = PackageCache.getInstance(compilerContext);
-        packageCache.flush();
-    }
+    public abstract void clearCaches();
 
     /**
      * Creates a new Project instance which has the same structure as this Project.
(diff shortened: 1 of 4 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageContext.java
```diff
@@ -212,6 +214,19 @@ ModuleCompilation getModuleCompilation(ModuleContext moduleContext) {
     PackageCompilation getPackageCompilation() {
         if (packageCompilation == null) {
             packageCompilation = PackageCompilation.from(this, this.compilationOptions());
+            
+            // Check if the compiler crashed
+            Collection<Diagnostic> errors = packageCompilation.diagnosticResult().errors();
+            if (errors != null) {
```
sameerajayasoma, line 220: Is this null check required?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/Project.java
```diff
@@ -87,6 +92,11 @@ protected void setCurrentPackage(Package currentPackage) {
     }
 
     public ProjectEnvironment projectEnvironmentContext() {
```
sameerajayasoma, line 94: How do we ensure that this method is invoked right after a compiler crash? 

There can be places where we have cached compilerContext instances. If that's the case, then we will end up having multiple stale instances of compilerContext objects.

sameerajayasoma (changes requested):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageContext.java
azinneera: Yes, the diagnostics in `PackageManifest` are added without a code. 
https://github.com/ballerina-platform/ballerina-lang/blob/4175b5491ca992402a9ff93a47db3899ad7acb31/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/ManifestBuilder.java#L438

Will create a separate issue to address that.

azinneera: Yes, the diagnostics in `PackageManifest` are added without a code. 
https://github.com/ballerina-platform/ballerina-lang/blob/4175b5491ca992402a9ff93a47db3899ad7acb31/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/ManifestBuilder.java#L438

Will create a separate issue to address that.

azinneera: We can remove it.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/Project.java
azinneera: Refactored the implementation as follows:
* Removed the crash recovery part from the Project API (which was added in the initial commit)
* Improved `Project.clearCaches` to reset the entire environment

The API user can call the `Project.clearCaches` method when there is a crash.

Issue #35389, opened 2022-03-09
Get rid of manually updating `api_gen_syntax_tree_descriptor.json`


lochana-chathura:
**Description:**
Can we $subject? 

- When changing the [syntax_tree_descriptor.json](https://github.com/ballerina-platform/ballerina-lang/blob/master/compiler%2Fballerina-treegen%2Fsrc%2Fmain%2Fresources%2Fsyntax_tree_descriptor.json), we have to manually change the [api_gen_syntax_tree_descriptor.json](https://github.com/ballerina-platform/ballerina-lang/blob/master/misc%2Fsyntax-api-calls-gen%2Fsrc%2Fmain%2Fresources%2Fapi_gen_syntax_tree_descriptor.json) in parallel. 
- We may miss doing the parallel update if there were no tests with the syntax of the affected node in the syntaxapicallsgen.
e.g. no XMLCDATA bbe.

Related commit: https://github.com/ballerina-platform/ballerina-lang/pull/35386/commits/68d183e61d9eb83efa8589a293a9df4603a6ff62

anupama-pathirage:
We are closing this issue due to inactivity. If you need further assistance or have additional input, feel free to reopen it. Thank you for helping us maintain a relevant and focused issue list!

Issue #35240, opened 2022-02-25
Formatter should not remove all whitespaces between invalid tokens


lochana-chathura:
**Description:**
$subject as the formatting would make the code different than it was before. 

Consider the following samples. 
Original Code:
![Screenshot from 2022-02-25 13-43-42](https://user-images.githubusercontent.com/39232462/155680585-45de1396-eab1-4fd1-978c-2dd4f47873ae.png)

After Formatting:
![Screenshot from 2022-02-25 13-44-03](https://user-images.githubusercontent.com/39232462/155680606-ef9ec2c3-ef2f-4263-8b72-fde150f086ec.png)

As you can see after formatting instead of `where` keyword and `i` identifier now we have  `wherei` identifier.
Basically removing all whitespaces makes the code tokenized differently than it was before. 

<details>
  <summary><b>Another Sample</b>  (<i>click to expand</i>)</summary>

Original Code:
![Screenshot from 2022-02-25 14-12-08](https://user-images.githubusercontent.com/39232462/155683610-5108aa37-f433-4b92-b5a5-307efd409b87.png)

After Formatting:
![Screenshot from 2022-02-25 14-12-32](https://user-images.githubusercontent.com/39232462/155683668-c2a1e71c-13e8-4c41-a8f8-5189a38ae1a6.png)

</details>

<details>
  <summary><b>Code Snippets</b>  (<i>click to expand</i>)</summary>

```bal
public function main() {
    var c = from var i in 0 ... 5
        select i
        where i % 2 != 0;
}
```

```bal
public function main() {
    map<string> v = {};
    match v {
        {...var c, var c} => {
            io:println(c);
        }
        _ => {

        }
    }
}
```
</details>

**Affected Versions:**
2201.0.0

PR #30227, opened 2021-04-28
Fix issue with the order of parameter list

chiranSachintha:
According to the spec param list should be in the following order.
```
param-list :=
   required-params [, defaultable-params] [, included-record-params] [, rest-param]
   | defaultable-params [, included-record-params] [, rest-param]
   | included-record-params [, rest-param]
   | [rest-param]
```

Fixes #27818 

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #36495, opened 2022-06-10
Add parser changes to allow annotations for each member in a tuple

### LINKED ISSUE #36390: Allow annotations for each type in a tuple
Please refer to : https://github.com/ballerina-platform/ballerina-spec/issues/1104

ushirask:
Fixes #36390

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -4371,6 +4371,22 @@
                     "type": "ExpressionNode"
                 }
             ]
+        },
+        {
+            "name": "TupleMemberDescriptorNode",
+            "base": "TypeDescriptorNode",
```
lochana-chathura, line 4377: We extend a node with `TypeDescriptorNode` only if the node itself is a type descriptor. We cannot use this in other places we have type-descriptor.

```suggestion
            "base": "Node",
```

Also, when updating this file, we need to update `api_gen_syntax_tree_descriptor.json` in parallel. Ballerina shell uses that file. We already have issue #35389 to avoid duplication.

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -4371,6 +4371,22 @@
...
+            "kind": "TYPE_DESC",
+            "attributes": [
+                {
+                    "name": "annotation",
```
lochana-chathura, line 4381: Similar to we have defined in other places
```suggestion
                    "name": "annotations",
```

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -4371,6 +4371,22 @@
...
+                    "type": "AnnotationNode",
+                    "occurrences": "MULTIPLE"
+                },
+                {
+                    "name": "typeDesc",
+                    "type": "Node"
```
lochana-chathura, line 4387: Here we can use the `TypeDescriptorNode` abstraction as whatever the node comes here is a `TypeDescriptorNode`.
```suggestion
                    "name": "typeDescriptor",
                    "type": "TypeDescriptorNode"
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/tree/STTupleMemberDescriptorNode.java
```diff
@@ -0,0 +1,94 @@
...
+import io.ballerina.compiler.syntax.tree.SyntaxKind;
+import io.ballerina.compiler.syntax.tree.TupleMemberDescriptorNode;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * This is a generated internal syntax tree node.
+ *
+ * @since 2202.0
```
lochana-chathura, line 31:
```suggestion
 * @since 2201.2.0
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/TupleMemberDescriptorNode.java
```diff
@@ -0,0 +1,114 @@
+/*
+ *  Copyright (c) 2020, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
lochana-chathura, line 2:
```suggestion
 *  Copyright (c) 2022, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/TupleMemberDescriptorNode.java
```diff
@@ -0,0 +1,114 @@
...
+package io.ballerina.compiler.syntax.tree;
+
+import io.ballerina.compiler.internal.parser.tree.STNode;
+
+import java.util.Objects;
+
+/**
+ * This is a generated syntax tree node.
+ *
+ * @since 2.0.0
```
lochana-chathura, line 27:
```suggestion
 * @since 2201.2.0
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/TupleMemberDescriptorNode.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public TupleMemberDescriptorNodeModifier modify() {
+        return new TupleMemberDescriptorNodeModifier(this);
+    }
+
+    /**
+     * This is a generated tree node modifier utility.
+     *
+     * @since 2.0.0
```
lochana-chathura, line 81:
```suggestion
     * @since 2201.2.0
```

compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/syntax/misc/AnnotationsTest.java
```diff
@@ -82,4 +82,9 @@ public void testAnnotAttachmentAtEOF() {
     public void testAnnotAtExprLevelRecovery() {
         testFile("annotations/annotations_source_11.bal", "annotations/annotations_assert_11.json");
     }
+
+    @Test
+    public void testAnnotTupleMembers() {
```
lochana-chathura, line 87: Since the new change is not for the [annotation](https://ballerina.io/spec/lang/master/#annotation) grammar, shall we move this to `TupleTypeTest` class?

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
```diff
@@ -4371,6 +4371,22 @@
                     "type": "ExpressionNode"
                 }
             ]
+        },
+        {
+            "name": "TupleMemberDescriptorNode",
```
lochana-chathura, line 4376:
```suggestion
            "name": "MemberTypeDescriptorNode",
```
@rdulmina WDYT? to be aligned with the spec terms.

MaryamZi:
@ushirask, please fix the PR title to be similar to the commit messages. See https://github.com/ballerina-platform/ballerina-lang/pull/30227#issuecomment-829996231.

compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json
rdulmina: +1

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -11358,17 +11358,25 @@ private STNode parseTupleMemberTypeDescList() {
...
 
+    private STNode parseTupleTypeMembers(STNode annot, STNode typeDesc, List<STNode> typeDescList) {
+        typeDesc = parseComplexTypeDescriptor(typeDesc, ParserRuleContext.TYPE_DESC_IN_TUPLE, false);
         STNode tupleMemberRhs = parseTypeDescInTupleRhs();
         if (tupleMemberRhs != null) {
+            if (annot != null) {
+                reportInvalidMetaData(annot, "tuple rest descriptor");
+            }
             typeDesc = STNodeFactory.createRestDescriptorNode(typeDesc, tupleMemberRhs);
+        } else if (typeDesc.kind != SyntaxKind.REST_TYPE) {
```
lochana-chathura, line 11378: `REST_TYPE` is not a type descriptor kind. So the `typeDesc` will never be `REST_TYPE`.
```suggestion
        } else {
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -11358,17 +11358,25 @@ private STNode parseTupleMemberTypeDescList() {
...
     private STNode parseTupleTypeMembers(STNode typeDesc, List<STNode> typeDescList) {
-        typeDesc = parseComplexTypeDescriptor(typeDesc, ParserRuleContext.TYPE_DESC_IN_TUPLE, false);
+        return parseTupleTypeMembers(STNodeFactory.createEmptyNode(), typeDesc, typeDescList);
+    }
 
+    private STNode parseTupleTypeMembers(STNode annot, STNode typeDesc, List<STNode> typeDescList) {
+        typeDesc = parseComplexTypeDescriptor(typeDesc, ParserRuleContext.TYPE_DESC_IN_TUPLE, false);
         STNode tupleMemberRhs = parseTypeDescInTupleRhs();
         if (tupleMemberRhs != null) {
+            if (annot != null) {
```
lochana-chathura, line 11374: Once `createEmptyNodeList()` is used as suggest above.
```suggestion
            if (!((STNodeList) annot).isEmpty()) {
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -11358,17 +11358,25 @@ private STNode parseTupleMemberTypeDescList() {
...
+                reportInvalidMetaData(annot, "tuple rest descriptor");
```
lochana-chathura, line 11375: underline `reportInvalidMetaData()`, we call `addInvalidNodeToNextToken()`. What it does is the invalid token is attached as the leading minutiae for the immediate token we `consume()`. Here invalid annotation has to be the leading minutiea of the `typeDesc` node which has been already consumed. 

Therefore, correct thing to do here is (otherwise invalid token will be on a wrong place in the tree messing up the node position calculation),
`typeDesc = SyntaxErrors.cloneWithLeadingInvalidNodeMinutiae(typeDesc, annot, DiagnosticErrorCode.ERROR_INVALID_TOKEN);`

Maybe we can use define a new diagnostic code as `metadata` is referred to both annotations + documentation.
We can specifically say "Annotations are not allowed for tuple rest descriptor".

lochana-chathura:
We also need to add a path for the optional annotation in tuple members in the BallerinaParserErrorhandler

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/diagnostics/DiagnosticErrorCode.java
```diff
@@ -345,7 +345,8 @@ public enum DiagnosticErrorCode implements DiagnosticCode {
     ERROR_INVALID_BINDING_PATTERN("BCE0677", "error.invalid.binding.pattern"),
     ERROR_RESOURCE_PATH_CANNOT_BEGIN_WITH_SLASH("BCE0678", "error.resource.path.cannot.begin.with.slash"),
     REST_PARAMETER_CANNOT_BE_INCLUDED_RECORD_PARAMETER("BCE0679",
-            "error.rest.parameter.cannot.be.included.record.parameter")
+            "error.rest.parameter.cannot.be.included.record.parameter"),
+    ERROR_INVALID_ANNOTATIONS_REST_MEMBER("BCE0680", "error.annotations.not.allowed.for.rest.member"),
```
lochana-chathura, line 349: We usually use the same enum member name in the dotted format as the `messageKey`.  Also, "not allowed for rest member" is generic. Shall we say,  "tuple rest descriptor"?
```suggestion
    ERROR_ANNOTATIONS_NOT_ALLOWED_FOR_TUPLE_REST_DESCRIPTOR("BCE0680", "error.annotations.not.allowed.for.tuple.rest.descriptor"),
```

When adding an entry here, you also need to add an entry at `syntax_diagnostic_message.properties` for the `messageKey`.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/SpreadMemberNode.java
```diff
@@ -78,7 +78,7 @@ public SpreadMemberNodeModifier modify() {
     /**
      * This is a generated tree node modifier utility.
      *
-     * @since 2201.1.0
+     * @since 2.0.0
```
lochana-chathura, line 81: Do we need to commit to this change?

ushirask: Seems to have accidentally added in treegen, I'll revert.

compiler/ballerina-parser/src/test/resources/types/tuple-type/tuple_type_source_10.bal
```diff
@@ -0,0 +1,9 @@
+type T1 [@annot {} int, @annot {} string];
+
+type T2 [int, @annot {} string];
+
+type T1 [int, @annot {} string...];
```
lochana-chathura, line 5: Since this is a negative test with a syntax error, shall we add it separately?
(Under `// Valid source test` we only keep the correct syntax. Any test with syntax errors goes under `// Recovery test` in `TupleTypeTest`.)

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -11332,7 +11332,7 @@ private boolean isEndOfAnonFuncParametersList(SyntaxKind tokenKind) {
      */
     private STNode parseTupleTypeDesc() {
         STNode openBracket = parseOpenBracket();
-        startContext(ParserRuleContext.TYPE_DESC_IN_TUPLE);
+        startContext(ParserRuleContext.TUPLE_MEMBER_OPTIONAL_ANNOTATIONS);
```
lochana-chathura, line 11335: Well, we are using this context to parse a list of tuple members. So the context name is misleading. 
Shall we have something like `TUPLE_MEMBERS`.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -808,6 +808,9 @@ public class BallerinaParserErrorHandler extends AbstractParserErrorHandler {
     private static final ParserRuleContext[] ANNOTATION_DECL_START =
             { ParserRuleContext.ANNOTATION_KEYWORD, ParserRuleContext.CONST_KEYWORD };
 
+    private static final ParserRuleContext[] TUPLE_MEMBER_OPTIONAL_ANNOTATIONS =
```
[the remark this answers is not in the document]
lochana-chathura: The next rule for `TUPLE_MEMBERS` we can define something like `TUPLE_MEMBER` which is this alternative path 
`{ ParserRuleContext.ANNOTATIONS, ParserRuleContext.TYPE_DESC_IN_TUPLE };`

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -4134,9 +4143,9 @@ private void startContextIfRequired(ParserRuleContext currentCtx) {
             case TYPE_DESC_IN_EXPRESSION:
             case TYPE_DESC_IN_STREAM_TYPE_DESC:
             case TYPE_DESC_IN_PARENTHESIS:
-            case TYPE_DESC_IN_TUPLE:
```
lochana-chathura, line 4137: In the parser when we call, ` parseTypeDescriptor(ParserRuleContext.TYPE_DESC_IN_TUPLE);` we start a context for `TYPE_DESC_IN_TUPLE` inside. Therefore we need it in the error handler as well.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -4134,9 +4143,9 @@ private void startContextIfRequired(ParserRuleContext currentCtx) {
...
             case TYPE_DESC_IN_SERVICE:
             case TYPE_DESC_IN_PATH_PARAM:
+            case TUPLE_MEMBERS:
```
lochana-chathura, line 4148: Shall we move this out of the `// Contexts that expect a type` section as this is not relevant? 
you can add this after the `case BRACED_EXPRESSION` above.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -4510,7 +4519,8 @@ private ParserRuleContext getNextRuleForTypeDescriptor() {
                 return ParserRuleContext.CLOSE_PARENTHESIS;
             case TYPE_DESC_IN_TUPLE:
             case STMT_START_BRACKETED_LIST:
-                return ParserRuleContext.TYPE_DESC_IN_TUPLE_RHS;
+            case TUPLE_MEMBERS:
```
lochana-chathura, line 4522: We should not reach here for the `TUPLE_MEMBERS`. We need to end the context for `TYPE_DESC_IN_TUPLE`, before we go into `TYPE_DESC_IN_TUPLE_RHS`. 
```java
            case TYPE_DESC_IN_TUPLE:
                endContext();
                return ParserRuleContext.TYPE_DESC_IN_TUPLE_RHS;
            case STMT_START_BRACKETED_LIST:
                return ParserRuleContext.TYPE_DESC_IN_TUPLE_RHS;
```

compiler/ballerina-parser/src/test/resources/types/tuple-type/tuple_type_source_10.bal
```diff
@@ -0,0 +1,7 @@
+type T1 [@annot {} int, @annot {} string];
```
lochana-chathura, line 1: To cover the optional mapping-constructor expr in annotation and a type ref scenario
```suggestion
type T1 [@annot {} int, @annot {} string, @annot T];
```

lochana-chathura:
Sorry I missed this before.
1. We also need to change,
https://github.com/ballerina-platform/ballerina-lang/blob/56e03dd7b2938ef06557f46baf457c965f83ed13/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java#L5037-L5038 to

```java
if (isInTypeDescContext()) {
    startContext(ParserRuleContext.TUPLE_MEMBERS);
    return ParserRuleContext.TUPLE_MEMBER;
}
```

2. Because we added `ANNOTATIONS` under `TUPLE_MEMEBERS` ctx we need to add an entry in `getNextRuleForAnnotationEnd()` for that ctx.

lochana-chathura:
Shall we also add a recovery test for tuple annotations?

ushirask:
@lochana-chathura the alterenatives for `TUPLE_MEMBER` should be
```
private static final ParserRuleContext[] TUPLE_MEMBER =
            { ParserRuleContext.ANNOTATIONS, ParserRuleContext.TYPE_DESCRIPTOR };
```
right? not `ANNOTATIONS` & `TYPE_DESC_IN_TUPLE`

lochana-chathura:
> @lochana-chathura the alterenatives for `TUPLE_MEMBER` should be
> 
> ```
> private static final ParserRuleContext[] TUPLE_MEMBER =
>             { ParserRuleContext.ANNOTATIONS, ParserRuleContext.TYPE_DESCRIPTOR };
> ```
> 
> right? not `ANNOTATIONS` & `TYPE_DESC_IN_TUPLE`

Nope. it has to be `TYPE_DESC_IN_TUPLE`. will explain you offline.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -16977,6 +16976,9 @@ private STNode parseStatementStartsWithOpenBracket(STNode annots, boolean isRoot
                 case TUPLE_TYPE_DESC:
                     // If the member type was figured out as a tuple-type-desc member, then parse the
                     // remaining members as tuple type members and be done with it.
+                    member = createMemberOrRestNode(STNodeFactory.createEmptyNodeList(), member);
+                    return parseAsTupleTypeDesc(annots, openBracket, memberList, member, isRoot);
```
rdulmina, line 16980: For samples like bellow
```bal
public function main() {
    [T...] a;
}
``` 
This will create a member descriptor containing rest descriptor inside

rdulmina: I think we should return the member descriptor from `parseStatementStartBracketedListMember` itself instead of returning just the type desc and creating the member descriptor here

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -11358,49 +11359,49 @@ private STNode parseTupleMemberTypeDescList() {
...
+            membertypeDesc = parseMemberDescriptor();
         }
 
-        typeDescList.add(typeDesc);
-        return STNodeFactory.createNodeList(typeDescList);
+        memberTypeDescList.add(membertypeDesc);
+        return STNodeFactory.createNodeList(memberTypeDescList);
     }
 
     private STNode parseMemberDescriptor() {
```
lochana-chathura, line 11388: Not from your change. Let's rename this to some like `parseMemberDescOrRestDescriptor()`or simply `parseTupleMember`.  Otherwise according to the spec, member descriptor means only `[annots] type-descriptor`.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -11358,49 +11359,49 @@ private STNode parseTupleMemberTypeDescList() {
...
 
-    private STNode parseTupleTypeMembers(STNode typeDesc, List<STNode> typeDescList) {
-        typeDesc = parseComplexTypeDescriptor(typeDesc, ParserRuleContext.TYPE_DESC_IN_TUPLE, false);
-
-        STNode tupleMemberRhs = parseTypeDescInTupleRhs();
-        if (tupleMemberRhs != null) {
-            typeDesc = STNodeFactory.createRestDescriptorNode(typeDesc, tupleMemberRhs);
-        }
-
+    private STNode parseTupleTypeMembers(STNode membertypeDesc, List<STNode> memberTypeDescList) {
```
lochana-chathura, line 11366: to about confusion with "member type desc" definition
```suggestion
    private STNode parseTupleTypeMembers(STNode firstMember, List<STNode> memberList) {
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -16978,6 +16979,8 @@ private STNode parseStatementStartsWithOpenBracket(STNode annots, boolean isRoot
                     // If the member type was figured out as a tuple-type-desc member, then parse the
                     // remaining members as tuple type members and be done with it.
                     return parseAsTupleTypeDesc(annots, openBracket, memberList, member, isRoot);
+                case MEMBER_TYPE_DESC:
```
lochana-chathura, line 16982: Logic is same as `TUPLE_TYPE_DESC` case, shall we merge?

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -17158,6 +17168,8 @@ private STNode parseTupleTypeDescOrListConstructor(STNode annots, STNode openBra
...
+                    return parseAsTupleTypeDesc(annots, openBracket, memberList, member, isRoot);
```
lochana-chathura, line 17172: Here also.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -17300,9 +17318,11 @@ private STNode parseStmtStartsWithTupleTypeOrExprRhs(STNode annots, STNode tuple
 
     private STNode parseAsTupleTypeDesc(STNode annots, STNode openBracket, List<STNode> memberList, STNode member,
                                         boolean isRoot) {
-        memberList = getTypeDescList(memberList);
+        memberList = getTupleMemberList(memberList);
+        startContext(ParserRuleContext.TUPLE_MEMBERS);
         STNode tupleTypeMembers = parseTupleTypeMembers(member, memberList);
         STNode closeBracket = parseCloseBracket();
+        endContext();
```
lochana-chathura, line 17325: At this point `stmt-start-bracketed-list` ctx due to ambiguity. When we reach here we know definite it is tuple. So we can switch it to the tuple.
```suggestion
        switchContext(ParserRuleContext.TUPLE_MEMBERS);
        STNode tupleTypeMembers = parseTupleTypeMembers(member, memberList);
        STNode closeBracket = parseCloseBracket();
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -1209,7 +1212,7 @@ protected Result seekMatch(ParserRuleContext currentCtx, int lookahead, int curr
                 lookahead++;
                 isEntryPoint = false;
             }
-
+
```
lochana-chathura, line 1215:
```suggestion

```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -3582,7 +3593,7 @@ private ParserRuleContext getNextRuleInternal(ParserRuleContext currentCtx, int
             case SIGNED_INT_OR_FLOAT_RHS:
                 return getNextRuleForExpr();
             case TUPLE_TYPE_DESC_START:
-                return ParserRuleContext.TYPE_DESC_IN_TUPLE;
+                return ParserRuleContext.TUPLE_MEMBER;
```
lochana-chathura, line 3596: Here after the `[` we can't directly call `TUPLE_MEMBER`. We have to start a context for the `TUPLE_MEMBERS`.
```suggestion
                return ParserRuleContext.TUPLE_MEMBERS;
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
lochana-chathura: My bad. you are right here.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
ushirask: When we create the member descriptor here since both tuple type desc and typed binding patterns are parsed in `parseStatementStartBracketedListMember` in a case like `[[int, int], int] [[a,b], c]`, `[a,b]` will be parsed into a `STMemberTypeDescriptorNode`. 
We can either  create member descriptor outside the function (as earlier), use separate functions or change STNode type later in the parser.

lochana-chathura: I see. We rearrange the node in the binding pattern scenario anyway.
e.g. a `SimpleNameReferenceNode` will be converted to a `CaptureBindingPatternNode`. 

Therefore shall add the `MemberTypeDescriptorNode` conversion at the same place?

https://github.com/ballerina-platform/ballerina-lang/blob/31e1d2c7b78e23ab880f16650ac74a11dcc0c89b/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java#L18650

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
lochana-chathura: Here we need the newline after the `}`. It is good to have a newline after an if-else block etc.
What I suggested was to remove the extra whitespaces at the beginning of the newline. :)

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
ushirask: Is there a problem with creating the member descriptor here and having a separate match case for `REST_TYPE`, otherwise it will add a lot of somewhat duplicate code?

ushirask: Reverted as discussed.

compiler/ballerina-parser/src/test/resources/types/tuple-type/tuple_type_assert_09.json
```diff
@@ -1,5 +1,6 @@
 {
   "kind": "FUNCTION_DEFINITION",
+  "hasDiagnostics": true,
```
lochana-chathura, line 3: This is a valid syntax test. Now it has diagnostics. Can you check?

compiler/ballerina-parser/src/test/resources/types/tuple-type/tuple_type_assert_04.json
```diff
@@ -1222,12 +1343,51 @@
...
                       ]
                     }
                   ]
                 },
+                {
+                  "kind": "EQUAL_TOKEN",
+                  "isMissing": true,
+                  "hasDiagnostics": true,
+                  "diagnostics": [
+                    "ERROR_MISSING_EQUAL_TOKEN"
```
lochana-chathura, line 1357: In here,
Input: ` [string % int a] j;`
Previous Recovery: `[string  MISSING[,] INVALID[%] int  MISSING[,]a] j;`
New Recovery:`[string  MISSING[,] INVALID[%] int  MISSING[]]a MISSING[ INVALID[]] =]j;`

Seems the parser can't detect the missing comma between `int` and `a` now.

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/ParserTestFormatter.java
```diff
@@ -73,6 +73,15 @@ public List<String> skipList() {
...
                 "resiliency_source_04.bal",
+                "tuple_type_source_09.bal",
+                "tuple_type_source_05.bal",
+                "tuple_type_source_08.bal",
+                "tuple_type_source_12.bal",
+                "tuple_type_source_11.bal",
+                "tuple_type_source_02.bal",
+                "tuple_type_source_10.bal",
+                "ambiguity_source_20.bal",
+                "ambiguity_source_01.bal",
```
lochana-chathura, line 84: Any reason we are skipping these tests in the Formatter?

compiler/ballerina-parser/src/main/resources/syntax_diagnostic_message.properties
```diff
@@ -329,3 +329,4 @@ error.resource.path.cannot.begin.with.slash=resource path cannot begin with slas
 error.missing.field.match.pattern.member=missing field match pattern member
 error.missing.object.constructor.expression=missing object constructor expression
 error.rest.parameter.cannot.be.included.record.parameter=rest parameter cannot be an included record parameter
+error.annotations.not.allowed.for.tuple.rest.descriptor =annotations not allowed for tuple rest descriptor
```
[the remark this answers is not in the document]
lochana-chathura: @ushirask shall we fix this also?

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/ParserTestFormatter.java
ushirask: they were skipped because `type T [@a x, @annot{} y]` was formatted as `type T[@ax , @annot {}y]`, I'll look at fixing the formatter

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
```diff
@@ -808,6 +808,9 @@ public class BallerinaParserErrorHandler extends AbstractParserErrorHandler {
     private static final ParserRuleContext[] ANNOTATION_DECL_START =
             { ParserRuleContext.ANNOTATION_KEYWORD, ParserRuleContext.CONST_KEYWORD };
 
+    private static final ParserRuleContext[] TUPLE_MEMBER =
+            { ParserRuleContext.ANNOTATIONS, ParserRuleContext.TYPE_DESC_IN_TUPLE };
```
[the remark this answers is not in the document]
lochana-chathura: In the parser, we allow annotations followed by the rest descriptor and then do the invalidation for the annotations. So this is consistence?

compiler/ballerina-parser/src/test/resources/types/tuple-type/tuple_type_source_10.bal
```diff
@@ -0,0 +1,11 @@
+type T1 [@annot {} int, @annot {} string, @annot T];
+
+type T2 [int, @annot {} string];
+
+function foo (string inMsg) returns [@annot int, @annot string] {
+    return [1, "s"];
+}
+
+function bar() returns [@annot{} int, [@annot{} int, string] {
```
rdulmina, line 9:
```suggestion
function bar() returns [@annot{} int, [@annot{} int, string]] {
```

compiler/ballerina-parser/src/test/resources/types/tuple-type/tuple_type_assert_10.json
```diff
@@ -0,0 +1,1170 @@
+{
+  "kind": "MODULE_PART",
+  "hasDiagnostics": true,
```
rdulmina, line 3: Valid syntax tests cannot contain `"hasDiagnostics": true` Please check

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParserErrorHandler.java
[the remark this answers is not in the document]
rdulmina: Resolving this as per the offline discussion

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/ParserTestFormatter.java
ushirask: Fixed.

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
```diff
@@ -575,6 +576,17 @@ public RecordFieldNode transform(RecordFieldNode recordField) {
                 .apply();
     }
 
+    @Override
+    public MemberTypeDescriptorNode transform(MemberTypeDescriptorNode member) {
+        NodeList<AnnotationNode> annotations = formatNodeList(member.annotations(),
+                1, 0, 1, 0);
```
lochana-chathura, line 582: Looks like we can write this inline without exceeding the 120 limit?

misc/formatter/modules/formatter-core/src/main/java/org/ballerinalang/formatter/core/FormattingTreeModifier.java
```diff
@@ -575,6 +576,17 @@ public RecordFieldNode transform(RecordFieldNode recordField) {
...
+        TypeDescriptorNode type = formatNode(member.typeDescriptor(), 0, 0);
```
lochana-chathura, line 583: We use the parent node's `trailingWS` and `trailingNL` for the last node

```suggestion
        TypeDescriptorNode type = formatNode(member.typeDescriptor(), env.trailingWS, env.trailingNL);
```

misc/formatter/modules/formatter-core/src/test/java/org/ballerinalang/formatter/core/ParserTestFormatter.java
```diff
@@ -172,7 +172,8 @@ public List<String> skipList() {
                 "predeclared-module-prefix_02.bal", "object_type_def_source_44.bal", "record_type_def_source_27.bal",
                 "func_type_source_09.bal", "func_type_source_13.bal", "func_type_source_14.bal",
                 "func_type_source_15.bal", "func_type_source_16.bal", "import_decl_source_24.bal",
-                "member_access_expr_source_11.bal", "float_literal_source_08.bal", "resiliency_source_05.bal");
+                "member_access_expr_source_11.bal", "float_literal_source_08.bal", "resiliency_source_05.bal",
+                "tuple_type_source_12.bal");
```
lochana-chathura, line 176: Formatter should be able to run on top of syntax errors as well.
Any reason we are skipping `tuple_type_source_12.bal`?

ushirask: `type T1 [int % @annot string]` is formatted to `type T1 [int% @annot string]` , should this be handled in the formatter?

lochana-chathura: I see. We have #35240. Since `int%` get tokenized the same way as `int %` we can keep this test as `int%` as well while enabling via the formatter.

ushirask: This had to be changed to `type T1 [int%  @annot string]` (extra space between % and @annot). Does this need to be changed?

lochana-chathura: that's okay for now.

lochana-chathura (approved):

compiler/ballerina-parser/src/main/resources/syntax_diagnostic_message.properties
```diff
@@ -329,3 +329,4 @@ error.resource.path.cannot.begin.with.slash=resource path cannot begin with slas
 error.missing.field.match.pattern.member=missing field match pattern member
 error.missing.object.constructor.expression=missing object constructor expression
 error.rest.parameter.cannot.be.included.record.parameter=rest parameter cannot be an included record parameter
+error.annotations.not.allowed.for.tuple.rest.descriptor=annotations not allowed for tuple rest descriptor
```
rdulmina, line 332:
```suggestion
error.annotations.not.allowed.for.tuple.rest.descriptor=annotations are not allowed for tuple rest descriptor
```

rdulmina (approved):
Let's update the error message

compiler/ballerina-parser/src/main/resources/syntax_diagnostic_message.properties
lochana-chathura: We had a chat ont this. I'll mention you there.

PR #36510, opened 2022-06-12
Fix xml template interpolated expr type checking

### LINKED ISSUE #36479: Align the xml template interpolation contextually type with the spec
**Description:**
$subject as per the clarification in https://github.com/ballerina-platform/ballerina-spec/issues/708#issuecomment-1149680048. 

We need to do type-check for interpolated expr first with `xml` (slient type check) and then with noType if there were errors in first round.

lochana-chathura:
## Purpose
$subject.

Fixes #36479 

## Approach
N/A

## Samples
N/A

## Remarks
The preferred CET is xml. Hence, first silent type check with xml (to guide the type that is chosen for expressions that can have multiple types). If it fails type check with noType.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -7564,12 +7564,28 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
         List<BLangExpression> tempConcatExpressions = new ArrayList<>();
 
         for (BLangExpression expr : exprs) {
-            BType exprType;
-            if (expr.getKind() == NodeKind.QUERY_EXPR) {
+            boolean prevNonErrorLoggingCheck = data.nonErrorLoggingCheck;
+            data.nonErrorLoggingCheck = true;
+            int prevErrorCount = this.dlog.errorCount();
+            this.dlog.resetErrorCount();
+            this.dlog.mute();
+
+            BType exprType = checkExpr(nodeCloner.cloneNode(expr), xmlElementEnv, symTable.xmlType, data);
+
+            data.nonErrorLoggingCheck = prevNonErrorLoggingCheck;
+            int errorCount = this.dlog.errorCount();
+            this.dlog.setErrorCount(prevErrorCount);
+
+            if (!prevNonErrorLoggingCheck) {
+                this.dlog.unmute();
+            }
+
+            if (errorCount == 0 && exprType != symTable.semanticError) {
                 exprType = checkExpr(expr, xmlElementEnv, symTable.xmlType, data);
             } else {
                 exprType = checkExpr(expr, xmlElementEnv, data);
             }
+
             if (TypeTags.isXMLTypeTag(Types.getReferredType(exprType).tag)) {
                 if (!tempConcatExpressions.isEmpty()) {
                     newChildren.add(getXMLTextLiteral(tempConcatExpressions));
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7564,12 +7564,28 @@ private List<BLangExpression> concatSimilarKindXMLNodes(List<BLangExpression> ex
...
         for (BLangExpression expr : exprs) {
-            BType exprType;
-            if (expr.getKind() == NodeKind.QUERY_EXPR) {
+            boolean prevNonErrorLoggingCheck = data.nonErrorLoggingCheck;
+            data.nonErrorLoggingCheck = true;
+            int prevErrorCount = this.dlog.errorCount();
+            this.dlog.resetErrorCount();
+            this.dlog.mute();
+
+            BType exprType = checkExpr(nodeCloner.cloneNode(expr), xmlElementEnv, symTable.xmlType, data);
```
MaryamZi, line 7573: I believe you need to increment the clone attempt on `expr`, or the node may not be cloned if has already been cloned.

```suggestion
            expr.cloneAttempt++;
            BType exprType = checkExpr(nodeCloner.cloneNode(expr), xmlElementEnv, symTable.xmlType, data);
```

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-literals.bal
```diff
@@ -395,4 +395,7 @@ function testQueryInXMLTemplateExpr() {
 
     string str3 = "<doc><doc><num>1</num><num>2</num><num>3</num></doc></doc>";
     test:assertEquals(x6.toString(), str3);
+
+    var x7 = xml `<doc>${(from string s in ["a", "b", "c"] select xml `${s}z`)}</doc>`;
+    test:assertEquals(x7.toString(), "<doc>azbzcz</doc>");
```
MaryamZi, line 400: Shall we add some tests for when the query expression is a nested expression in another kind of expression, like we discussed during the call?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
lochana-chathura: aren't we incrementing the `cloneAttempt` here for every clone?
https://github.com/ballerina-platform/ballerina-lang/blob/4175b5491ca992402a9ff93a47db3899ad7acb31/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/NodeCloner.java#L267

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-literals.bal
lochana-chathura: [1].
```bal
    xml x = xml `foo`;
    xml y = x + from string s in ["a", "b"] select xml `${s}`; // operator '+' not defined for 'xml' and 'xml:Text[]'
```

[2].
```bal
var y = xml `foo` + <xml>from string s in ["a", "b"] select xml `${s}`;
```

Well, I believe we discussed something like above.
[1] does not work, as we infer the rhs-expr type and it will be `xml:Text[]`.
[2] works even without this fix. (if we add a test it only tests the type cast type checking logic)

MaryamZi: Could even be error scenarios.

What I meant was something like

```ballerina
var x7 = xml `<doc>${xml `foo` + (from string s in ["a", "b", "c"] select xml `${s}z`)}</doc>`;
```

But yeah, this doesn't work atm either. Shall we create an issue for this?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: My bad, `clone` is what doesn't increment.

MaryamZi (approved):

PR #36532, opened 2022-06-13
Disallow field initializer expr that is available within class scope

### LINKED ISSUE #22192: Referring to other object fields in a default value of a field allowed
Consider the following code:
```ballerina
import ballerina/io;

public function main() {
    Person p = new;
    io:println(p.lname);
    io:println(typeof p.lname);
}

public type Person object {
    string name = "John Doe";
    int age = 25;
    string lname = name;
};
```
The above compiles and runs and prints the following:
```
age
typedesc string
```

### LINKED ISSUE #34532: Runtime error when casting a 'any' to 'string' in class scope
**Description:**
$title

**Steps to reproduce:**

```ballerina
public class Counter {

    private int n;
    any x = 12; 
    int k = <int> x; // error: casting is not supported from 'string' to 'int'
    public function init(int n = 0) {
        self.n = n;
    }
}

public function main() {
    Counter count = new(10);
}
```
error: `error: casting is not supported from 'string' to 'int'`

Following program compiles without an error

```ballerina
public class Counter {

    private int n;
    any x = function() {}; 
    string k = <string> x;
    public function init(int n = 0) {
        self.n = n;
    }
}

public function main() {
    Counter count = new(10);
}
```

**Affected Versions:**
Ballerina Swan Lake Beta 6

KavinduZoysa:
## Purpose

Fixes #34532
Fixes #22192


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -630,13 +630,9 @@ public void visit(BLangClassDefinition classDefinition, AnalyzerData data) {
     }
 
     private void analyzeClassDefinition(BLangClassDefinition classDefinition, AnalyzerData data) {
-        SymbolEnv currentEnv = data.env;
-        SymbolEnv classEnv = SymbolEnv.createClassEnv(classDefinition, classDefinition.symbol.scope, currentEnv);
         for (BLangSimpleVariable field : classDefinition.fields) {
-            data.env = classEnv;
             analyzeDef(field, data);
```
chiranSachintha, line 634: By doing this, when analyzing fields of objects we set the scope as package scope. 
But in symbol enter when we define these fields we create new scope(for classes) and define these fields within that scope. So IMO here also we need to create that scope.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -975,20 +971,21 @@ public void visit(BLangSimpleVariable varNode, AnalyzerData data) {
...
         SymbolEnv currentEnv = data.env;
         int ownerSymTag = currentEnv.scope.owner.tag;
-        boolean isListenerDecl = varNode.flagSet.contains(Flag.LISTENER);
-        if ((ownerSymTag & SymTag.INVOKABLE) == SymTag.INVOKABLE || (ownerSymTag & SymTag.LET) == SymTag.LET
-                || currentEnv.node.getKind() == NodeKind.LET_CLAUSE) {
+        Set<Flag> flagSet = varNode.flagSet;
+        boolean isListenerDecl = flagSet.contains(Flag.LISTENER);
+        if (flagSet.contains(Flag.OBJECT_FIELD)) {
+            analyzeVarNode(varNode, data, AttachPoint.Point.OBJECT_FIELD, AttachPoint.Point.FIELD);
+        } else if (flagSet.contains(Flag.RECORD_FIELD)) {
```
chiranSachintha, line 978: Can't we use the flag `FIELD` instead of defining new flags. If we set env correctly then we can use `ownerSymTag`  to separate the record field and class field. right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
KavinduZoysa: Simply, what I want to do is that when analyzing a field, that field should not be able to see previous fields. I don't think we need to create a new scope even if we did in SymbolEnter.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
KavinduZoysa: What do you mean by setting env correctly? 
I think that it is clean to having to type for field.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
chiranSachintha: We need. if not we are going to define fields of class within the package scope.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
chiranSachintha: With this change when visiting a field of a class, the env is set to packageEnv. IMO it should be classEnv.
If we are going to introduce two new flags for each type then we can remove the `FIELD` flag. right?

KavinduZoysa: I don't think we need to change the env when visiting fields. Fields without self should not be seen within the class.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
KavinduZoysa: Can we find any failing scenario?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
KavinduZoysa: There are few usages of `FIELD`, therefore we cannot use that

tests/jballerina-unit-test/src/test/resources/test-src/klass/class-def-field-negative.bal
```diff
@@ -0,0 +1,24 @@
+class ClassA {
```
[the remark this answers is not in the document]
KavinduZoysa: Added

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/klass/ClassTest.java
```diff
@@ -106,6 +106,15 @@ public void classDefNegativeSemanticErrors() {
         Assert.assertEquals(negative.getErrorCount(), i);
     }
 
+    @Test(description = "Negative tests to check invalid fields")
+    public void classDefInvalidField() {
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/klass/ClassTest.java
```diff
@@ -106,6 +106,15 @@ public void classDefNegativeSemanticErrors() {
...
+        CompileResult negative = BCompileUtil.compile("test-src/klass/class-def-field-negative.bal");
+        int i = 0;
+        BAssertUtil.validateError(negative, i++, "undefined symbol 'a'", 3, 13);
+        BAssertUtil.validateError(negative, i++, "undefined symbol 'a'", 11, 19);
+        BAssertUtil.validateError(negative, i, "undefined symbol 'name'", 20, 20);
+    }
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

KavinduZoysa:
@MaryamZi, @chiranSachintha, Maybe I misunderstood what you said. At function `analyzeClassDefinition` in SemanticAnalyser, field are already defined in class scope.
```ballerina
class ClassA {
    int a = 12;
    int b = a;

    function init() {
    }
}
```
If we set the class scope, `int b = a` statement can see the variable reference `a` which is not correct. 
Therefore what I have done is, set the scope to package scope, when looking up `a`. At this point, the statement `int b = a` cannot see variable reference `a`. Is this approach wrong? 
Within `analyzeClassDefinition`, I am not defining fields in package scope, those are already defined in the class scope.

chiranSachintha:
> @MaryamZi, @chiranSachintha, Maybe I misunderstood what you said. At function `analyzeClassDefinition` in SemanticAnalyser, field are already defined in class scope.
> 
> ```ballerina
> class ClassA {
>     int a = 12;
>     int b = a;
> 
>     function init() {
>     }
> }
> ```
> 
> If we set the class scope, `int b = a` statement can see the variable reference `a` which is not correct. Therefore what I have done is, set the scope to package scope, when looking up `a`. At this point, the statement `int b = a` cannot see variable reference `a`. Is this approach wrong? Within `analyzeClassDefinition`, I am not defining fields in package scope, those are already defined in the class scope.
With this change, we provide an error as `undefined symbol a` for the above example. IMO that is incorrect. 
Because within the class scope, we have defined `a`. With your approach, you are not looking at the class scope. That is the reason to give such kind of error. I think we need to give a separate error for this. (Because `a` is defined within the scope and we restrict assigning it as expr of another field.)

chiranSachintha:
> @MaryamZi, @chiranSachintha, Maybe I misunderstood what you said. At function `analyzeClassDefinition` in SemanticAnalyser, field are already defined in class scope.
> 
> ```ballerina
> class ClassA {
>     int a = 12;
>     int b = a;
> 
>     function init() {
>     }
> }
> ```
> 
> If we set the class scope, `int b = a` statement can see the variable reference `a` which is not correct. Therefore what I have done is, set the scope to package scope, when looking up `a`. At this point, the statement `int b = a` cannot see variable reference `a`. Is this approach wrong? Within `analyzeClassDefinition`, I am not defining fields in package scope, those are already defined in the class scope.

With this change, we provide an error as an undefined symbol for the above example. IMO that is incorrect.
Because within the class scope, we have defined a. With your approach, you are not looking at the class scope. That is the reason to give such kind of error. I think we need to give a separate error for this. (Because a is defined within the scope and we restrict assigning it as expr of another field.)

MaryamZi:
> @MaryamZi, @chiranSachintha, Maybe I misunderstood what you said. At function `analyzeClassDefinition` in SemanticAnalyser, field are already defined in class scope.
> 
> ```ballerina
> class ClassA {
>     int a = 12;
>     int b = a;
> 
>     function init() {
>     }
> }
> ```
> 
> If we set the class scope, `int b = a` statement can see the variable reference `a` which is not correct. Therefore what I have done is, set the scope to package scope, when looking up `a`. At this point, the statement `int b = a` cannot see variable reference `a`. Is this approach wrong? Within `analyzeClassDefinition`, I am not defining fields in package scope, those are already defined in the class scope.

Does this work for classes defined for object constructors within a function? Defaults of such fields should be able to see local variables.

```ballerina
public function main() {
    int a = 1;

    var _ = object {
        int x = a; // OK
        int y = x; // error

        function init() {

        }
    };
}
```

> With this change, we provide an error as an undefined symbol for the above example. IMO that is incorrect.
> Because within the class scope, we have defined a. With your approach, you are not looking at the class scope. That is the reason to give such kind of error. I think we need to give a separate error for this. (Because a is defined within the scope and we restrict assigning it as expr of another field.)

I think we can consider this as an improvement. At the end of the day, it is just not possible to refer to another field, as opposed to accessing it incorrectly.

KavinduZoysa:
> Does this work for classes defined for object constructors within a function? Defaults of such fields should be able to see local variables.
> 
> ```ballerina
> public function main() {
>     int a = 1;
> 
>     var _ = object {
>         int x = a; // OK
>         int y = x; // error
> 
>         function init() {
> 
>         }
>     };
> }
> ```
> 

@MaryamZi. This example works as expected. Added related tests.

MaryamZi:
> > Does this work for classes defined for object constructors within a function? Defaults of such fields should be able to see local variables.
> > ```ballerina
> > public function main() {
> >     int a = 1;
> > 
> >     var _ = object {
> >         int x = a; // OK
> >         int y = x; // error
> > 
> >         function init() {
> > 
> >         }
> >     };
> > }
> > ```
> 
> @MaryamZi. This example works as expected with this change. Added related tests.

How does it work though?

> Therefore what I have done is, set the scope to package scope, when looking up `a`

If we are doing this for all classes?

KavinduZoysa:
@MaryamZi I am sorry that my explanation is not a general one for all cases. I explained it based on the very first example we considered. 
In general, when visiting the fields in the class, the scope is set to its previous one (This scope is already set), In your example when visiting `int x = a` and `int y = x`, the scope is set to `main` functions scope. That is why this is working.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: The field should still be defined in the scope of the relevant symbol. So why can't we rely on FIELD + owner symbol? We shouldn't have to define new flags.

KavinduZoysa: Since we are using an already set scope as mentioned in https://github.com/ballerina-platform/ballerina-lang/pull/36532#issuecomment-1254848680, the owner is not the object anymore, it is the owner of the object. That is why I followed this approach.

MaryamZi: That is wrong, right? The field's owner should be the object. We just need to make sure the fields aren't visible when visiting the default value expression.

KavinduZoysa: I think it is not okay to say that the `currentEnv.scope.owner` gives the owner of field. It is what the scopes owner. 
Similar to fields that I have done this PR, for methods also we are not using a new scope, it uses the scope what is already set. Please check https://github.com/ballerina-platform/ballerina-lang/blob/26592f29cd6d0889f4ad47f85df695341257c532/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java#L700

MaryamZi: I'm not sure why that was done for methods, I think it could be a different env but still a class env, we can check maybe. But, since we've been defining fields in the proper (or most relevant scope) why would be go back and define it in a less accurate (wider) scope?

The issue you are trying to solve only applies to the default value expression. I would rather we use a correct env (where the other fields are not visible, maybe somewhat similar to what was done in https://github.com/ballerina-platform/ballerina-lang/pull/35326) than change how the field itself is defined.

KavinduZoysa: @MaryamZi, the approach is changed, please have a look.

tests/jballerina-unit-test/src/test/resources/test-src/object/anon_object.bal
```diff
@@ -292,3 +292,23 @@ function testCodeAnalyzerRunningOnAnonymousObjectsForDeprecatedFunctionAnnotatio
...
 }
+
+function testLocalVariablesFromFields() {
+    int a = 10;
+
+    var _ = object {
+        int x = a;
+
+        function init() {
+            assertEquality(self.x, 10);
```
MaryamZi, line 303: This test relies on the init being called. IMO, it would be better to do the assertion outside: field access on the object constructed.

tests/jballerina-unit-test/src/test/resources/test-src/object/anon_object.bal
```diff
@@ -292,3 +292,23 @@ function testCodeAnalyzerRunningOnAnonymousObjectsForDeprecatedFunctionAnnotatio
 isolated function Test() returns int {
```
MaryamZi, line 292: This isn't the correct file for this test, right?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/object/AnonymousObjectTest.java
```diff
@@ -169,6 +169,11 @@ public void testCodeAnalyzerRunningOnAnonymousObjectsForDeprecatedFunctionAnnota
         BAssertUtil.validateWarning(compileResult, 1, "usage of construct 'Test' is deprecated", 287, 25);
     }
 
+    @Test
+    public void testLocalVariablesFromFields() {
```
MaryamZi, line 173: Check the comment in the test bal file. This is more an object constructor test.

tests/jballerina-unit-test/src/test/resources/test-src/klass/class-def-field-negative.bal
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2022 WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
MaryamZi, line 1: We use underscores in bal file names.

MaryamZi (commented):
1. Let's add tests like

```ballerina
int a = 1;

class Class {
    int a = 0;
    int b = a; // OK

    function init() {
    }
}
```

2. Please make sure the tests are in the appropriate files. E.g., object constructor and class tests

tests/jballerina-unit-test/src/test/resources/test-src/object/anon_object.bal
KavinduZoysa: Changed the test

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/object/AnonymousObjectTest.java
KavinduZoysa: Changed

tests/jballerina-unit-test/src/test/resources/test-src/object/anon_object.bal
KavinduZoysa: Changed

tests/jballerina-unit-test/src/test/resources/test-src/klass/class_def_test.bal
```diff
@@ -31,6 +31,22 @@ function testFooClass() {
...
+public class Bar {
+    int a = 4;
+    int b = a;
+
+    function init() {
+        assertEquality(self.b, 3);
+    }
+}
+
+function testGlobalVariablesAssignmentToFields() {
```
MaryamZi, line 45: This is more 
```suggestion
function testModuleLevelVariableAsFieldDefault() {
```

Let's make sure we're using correct terms.

tests/jballerina-unit-test/src/test/resources/test-src/klass/class_def_test.bal
```diff
@@ -31,6 +31,22 @@ function testFooClass() {
...
 }
 
+int a = 3;
+
+public class Bar {
+    int a = 4;
+    int b = a;
+
+    function init() {
+        assertEquality(self.b, 3);
```
MaryamZi, line 41: Why are we doing the assertion here rather than in the actual test function?

tests/jballerina-unit-test/src/test/resources/test-src/klass/class_def_test.bal
```diff
@@ -31,6 +31,22 @@ function testFooClass() {
...
+    }
+}
+
+function testGlobalVariablesAssignmentToFields() {
+    Bar b = new();
+    assertEquality(b.b, 3);
```
MaryamZi, line 47: Shall we also update the module-level variable and test?

```ballerina
    a = 6;
    assertEquality(b.b, 3);
    Bar b2 = new;
    assertEquality(b2.b, 6);
```

tests/jballerina-unit-test/src/test/resources/test-src/object/object_constructor.bal
```diff
@@ -12,3 +12,24 @@ public function testObjectInit() returns int {
     p = new;
     return a;
 }
+
+function testLocalVariablesAssignmentToFields() {
```
MaryamZi, line 16: Same comments as that for the class.

tests/jballerina-unit-test/src/test/resources/test-src/object/object_constructor.bal
```diff
@@ -12,3 +12,24 @@ public function testObjectInit() returns int {
...
+    int a = 10;
+
+    var obj = object {
```
MaryamZi, line 19: Let's add tests for module-level object constructors also.

tests/jballerina-unit-test/src/test/resources/test-src/object/object_field_negative.bal
```diff
@@ -24,3 +24,15 @@ class Person {
         self.'error = er;
     }
 }
+
+function testFieldInitializationUsingAnotherField() {
```
MaryamZi, line 28: Is this the correct file for this test? We can add a new file if there's no appropriate object constructor file.

MaryamZi: Let's add negative tests at module-level also.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/object/ObjectTest.java
```diff
@@ -46,11 +46,14 @@ public class ObjectTest {
...
     private CompileResult checkInInitializerResult;
     private CompileResult checkFunctionReferencesResult;
+    private CompileResult objectCtrResult;
 
     @BeforeClass
     public void setUp() {
         checkInInitializerResult = BCompileUtil.compile("test-src/object/object_field_initializer_with_check.bal");
         checkFunctionReferencesResult = BCompileUtil.compile("test-src/object/object_function_pointer.bal");
+        objectCtrResult = BCompileUtil.compile("test-src/object/object_constructor.bal");
+
```
MaryamZi, line 56: Extra new line.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/object/ObjectTest.java
```diff
@@ -46,11 +46,14 @@ public class ObjectTest {
 
     private CompileResult checkInInitializerResult;
     private CompileResult checkFunctionReferencesResult;
+    private CompileResult objectCtrResult;
 
     @BeforeClass
     public void setUp() {
         checkInInitializerResult = BCompileUtil.compile("test-src/object/object_field_initializer_with_check.bal");
         checkFunctionReferencesResult = BCompileUtil.compile("test-src/object/object_function_pointer.bal");
+        objectCtrResult = BCompileUtil.compile("test-src/object/object_constructor.bal");
```
MaryamZi, line 55: Since this is only used in one test, it can be a local variable in that method.

tests/jballerina-unit-test/src/test/resources/test-src/object/object_field_negative.bal
KavinduZoysa: Removed

tests/jballerina-unit-test/src/test/resources/test-src/object/object_constructor.bal
KavinduZoysa: Changed

tests/jballerina-unit-test/src/test/resources/test-src/klass/class_def_test.bal
KavinduZoysa: Added

tests/jballerina-unit-test/src/test/resources/test-src/klass/class_def_test.bal
KavinduZoysa: Removed

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/klass/ClassTest.java
```diff
@@ -131,6 +141,7 @@ public void classDefTest() {
         CompileResult result = BCompileUtil.compile("test-src/klass/class_def_test.bal");
         Assert.assertEquals(result.getErrorCount(), 0);
         BRunUtil.invoke(result, "testFooClass");
+        BRunUtil.invoke(result, "testGlobalVariablesAssignmentToFields");
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

tests/jballerina-unit-test/src/test/resources/test-src/object/object_constructor.bal
KavinduZoysa: Added

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/object/ObjectTest.java
KavinduZoysa: Added few more usages to this variable

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/object/ObjectTest.java
```diff
@@ -46,11 +46,13 @@ public class ObjectTest {
...
```
MaryamZi, line 55: Why are we adding this to `ObjectTest` and not `ObjectConstructorTest`?

tests/jballerina-unit-test/src/test/resources/test-src/object/object_constructor.bal
```diff
@@ -12,3 +12,38 @@ public function testObjectInit() returns int {
...
+}
+
+int b = 11;
+
+var obj = object {
+        int x = b;
+
+        function init() {
+        }
+    };
```
MaryamZi, line 35: Formatting issue.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/klass/ClassTest.java
```diff
@@ -126,11 +139,17 @@ public void classDefDataflowNegative() {
...
-        BRunUtil.invoke(result, "testFooClass");
+    @Test(dataProvider = "classDefTestFunctions")
+    public void testClassDef(String functionName) {
+        BRunUtil.invoke(classDefCompileResult, functionName);
+    }
+
+    @DataProvider(name = "classDefTestFunctions")
+    public Object[] classDefTestFunctions() {
+        return new String[]{
+                "testFooClass",
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -770,6 +772,18 @@ private void analyzeClassDefinition(BLangClassDefinition classDefinition, Analyz
         analyzerClassInitMethod(classDefinition, data);
     }
 
+    private Scope createFieldsRemovedScope(Scope classScope) {
+        Scope scope = new Scope(classScope.owner);
+        for (Name key : classScope.entries.keySet()) {
+            Scope.ScopeEntry entry = classScope.entries.get(key);
+             BSymbol fieldSymbol = entry.symbol;
```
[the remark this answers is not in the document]
KavinduZoysa: Fixed all places

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/object/ObjectTest.java
KavinduZoysa: Changed

tests/jballerina-unit-test/src/test/resources/test-src/klass/class_def_field_negative.bal
```diff
@@ -0,0 +1,40 @@
...
+    int b = <int> a; // error
+
+    function init() {
+    }
+}
+
+class ClassC {
+    string name = "John Doe";
+    int age = 25;
+    string lname = name; // error
```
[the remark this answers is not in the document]
KavinduZoysa: Added a test

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -770,6 +772,18 @@ private void analyzeClassDefinition(BLangClassDefinition classDefinition, Analyz
...
+             if (!Symbols.isFlagOn(fieldSymbol.flags, Flags.FIELD)) {
+                 scope.entries.put(key, entry);
```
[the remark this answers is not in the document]
KavinduZoysa: Changed

language-server/modules/langserver-core/src/test/resources/codeaction/type-cast/source/type_cast_in_obj_fields.bal
```diff
@@ -1,10 +1,8 @@
+float f1 = 10.2;
```
MaryamZi, line 1: Please get these changes reviewed.

tests/jballerina-unit-test/src/test/resources/test-src/object/object_constructor.bal
```diff
@@ -12,3 +12,4 @@ public function testObjectInit() returns int {
     p = new;
     return a;
 }
+
```
MaryamZi, line 15: Unnecessary change.

MaryamZi (commented):
Compiler changes and tests LGTM.

language-server/modules/langserver-core/src/test/resources/codeaction/type-cast/source/type_cast_in_obj_fields.bal
KavinduZoysa: @TharushiJay, can you review LS changes.

TharushiJay (approved):
LS changes LGTM

MaryamZi (approved):

PR #36557, opened 2022-06-16
Fix isSameType wrong skipping for pairs

### LINKED ISSUE #36069: `isSameType` check blindly returns true for pairs found in `unresolvedTypes` set

lochana-chathura:
## Purpose
$subject.

Fixes #36069 

## Approach
N/A

## Samples
N/A

## Remarks
N/A

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


KRVPerera:
Can you change the order of expressions in if condition to have `isSameType` call as the last call
https://github.com/ballerina-platform/ballerina-lang/blob/03275ec1c2e80c558adeb78165eaf534519b40c2/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java#L2643

KRVPerera:
Shall we also create a new set here 
https://github.com/ballerina-platform/ballerina-lang/blob/03275ec1c2e80c558adeb78165eaf534519b40c2/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java#L2741

KRVPerera:
We may have to create another issue for `isSameOrderedType` no test case yet.

KRVPerera (commented):
Shall we change the order of expressions in if condition to have `isSameType` call as the last call
https://github.com/ballerina-platform/ballerina-lang/blob/03275ec1c2e80c558adeb78165eaf534519b40c2/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java#L2643


Shall we also create a new set here 
https://github.com/ballerina-platform/ballerina-lang/blob/03275ec1c2e80c558adeb78165eaf534519b40c2/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java#L2741

https://github.com/ballerina-platform/ballerina-lang/blob/03275ec1c2e80c558adeb78165eaf534519b40c2/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java#L2765


We may have to create another issue for `isSameOrderedType` no test case yet.

KRVPerera (commented):
Would you please check why we need to add `sUnionType` and `tUnionType` here? 
https://github.com/ballerina-platform/ballerina-lang/blob/03275ec1c2e80c558adeb78165eaf534519b40c2/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java#L2732

I believe this is a bug. This may be needed for cyclic unions only in that case we should be able to do.

```
if (sUnionType.isCyclic) {
    sourceTypes.add(sUnionType);
}
if (tUnionType.isCyclic) {
    targetTypes.add(tUnionType);
}
```


This may be needed because when we take apart individual members of a cyclic union the corresponding members cannot properly represent the cyclic nature. But assuming cyclic union should be already added to the unresolved list we may not need to do this at all.

lochana-chathura:
Closing temporary.

KRVPerera (approved):

PR #36586, opened 2022-06-17
Fix source pruning when there is a keyword in resource path

### LINKED ISSUE #36128: Source pruning when there is a keyword in resource path
**Description:**

Consider the below sample

```bal
    resource function get limit/[int id] () {
        
    }
}
```

Parser generated source file
```bal
    resource function get  MISSING[]/[int id] () {
        
    }
}
```
Node: the `limit` keyword is removed from the source file instead of reporting it as invalid token

lasinicl:
## Purpose
$title

Fixes #36128

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
+++ b/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
@@ -6915,7 +6915,8 @@ public class BallerinaParser extends AbstractParser {
         STToken nextToken = peek();
         switch (nextToken.kind) {
             case IDENTIFIER_TOKEN:
-                if (isFirstSegment && nextToken.isMissing() && getNextNextToken().kind == SyntaxKind.SLASH_TOKEN) {
+                if (isFirstSegment && nextToken.isMissing() && peek(0).kind == SyntaxKind.IDENTIFIER_TOKEN &&
+                        getNextNextToken().kind == SyntaxKind.SLASH_TOKEN) {
                     // special case `[MISSING]/` to improve the error message for `/hello`
                     consume(); // to ignore current missing identifier diagnostic
                     return SyntaxErrors.createMissingTokenWithDiagnostics(SyntaxKind.IDENTIFIER_TOKEN, 
--- a/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
+++ b/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
@@ -6918,7 +6918,7 @@ public class BallerinaParser extends AbstractParser {
                 if (isFirstSegment && nextToken.isMissing() && peek(0).kind == SyntaxKind.IDENTIFIER_TOKEN &&
                         getNextNextToken().kind == SyntaxKind.SLASH_TOKEN) {
                     // special case `[MISSING]/` to improve the error message for `/hello`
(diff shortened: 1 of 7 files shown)


compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -6915,7 +6915,8 @@ private STNode parseResourcePathSegment(boolean isFirstSegment) {
         STToken nextToken = peek();
         switch (nextToken.kind) {
             case IDENTIFIER_TOKEN:
-                if (isFirstSegment && nextToken.isMissing() && getNextNextToken().kind == SyntaxKind.SLASH_TOKEN) {
+                if (isFirstSegment && nextToken.isMissing() && peek(0).kind == SyntaxKind.IDENTIFIER_TOKEN &&
+                        getNextNextToken().kind == SyntaxKind.SLASH_TOKEN) {
                     // special case `[MISSING]/` to improve the error message for `/hello`
                     consume(); // to ignore current missing identifier diagnostic
```
lochana-chathura, line 6921: The fix `peek(0).kind == SyntaxKind.IDENTIFIER_TOKEN` is correct. But, we rely on the previously consumed token which is vulnerable to error if the grammar before `resource-path-segment` changes. 

The core problem here is using `consume();` to ignore the missing identifier. When we call `consume()` if there are any invalid tokens in the invalid token stack they are also attached to the leading minutiae of the `consume()` token.

Therefore, I suggest following to ignore the missing token(virtually inserted token). 
```suggestion
                    this.insertedToken = null; // to ignore current missing identifier diagnostic
```

rdulmina: Now we don't need `peek(0).kind == SyntaxKind.IDENTIFIER_TOKEN` check right?. Also it's more clean to introduce `removeInsertedToken()` method in `AbstractParser` and do this.insertedToken = null there WDYT?

lasinicl: If we remove `peek(0).kind == SyntaxKind.IDENTIFIER_TOKEN` check, `resource path cannot begin with slash` error is also logged with the `invalid token 'limit'` error. Wouldn't that be confusing for the user?

rdulmina: +1 let's keep the check

rdulmina (approved):

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
lochana-chathura: Our assumption here is when the `peek(0).kind == SyntaxKind.IDENTIFIER_TOKEN`, the invalid node stack is empty.
Therefore, I prefer `this.invalidNodeInfoStack.isEmpty()` check instead.  `invalidNodeInfoStack` is private atm.  We can have something like `isInvalidNodeStackEmpty()`. WDYT?

rdulmina: +1

PR #36668, opened 2022-06-22
Detect cyclic module dependencies

### LINKED ISSUE #27371: Detect dependency cycles in package dependencies and modules within a package
**Description:**
$subject.

nipunayf:
## Purpose
The topological sort cannot have an output if the dependency graph is not acyclic. Hence, all the cyclic dependencies are diagnosed at the package resolution.

Fixes #27371 

## Approach
The topological sort algorithm is extended to detect cycles through back edges in DFS.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/DependencyGraph.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/DependencyGraph.java
@@ -71,7 +71,7 @@ public class DependencyGraph<T> {
      *
      * @param other other dependency graph to compare with
      * @return the set of nodes that needs to be deleted from this graph to
-     * to get the other graph
+     * get the other graph
      */
     Set<T> difference(DependencyGraph<T> other) {
         // If the diff is empty, either this graph is equivalent to or a subset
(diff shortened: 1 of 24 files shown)


azinneera:
Having an empty topologically sorted list can be an issue. For the CLI commands, it would not have an impact but the LS will crash as I understand. Empty sorted list means we won't compile any modules. The LS suggestions rely on the semantic model of a Module which can only be accessed after compiling the module

compiler/ballerina-lang/src/test/java/io/ballerina/projects/CyclicDependenciesTest.java
```diff
@@ -0,0 +1,100 @@
+package io.ballerina.projects;
```
azinneera, line 1: Missing license header

compiler/ballerina-lang/src/main/java/io/ballerina/projects/DependencyGraph.java
```diff
@@ -139,6 +141,27 @@ public List<T> toTopologicallySortedList() {
         return topologicallySortedNodes;
     }
 
+    public List<T> toTopologicallySortedListWithCycles() {
```
azinneera, line 144: Shall we move the logic to ` toTopologicallySortedList` and remove this?

nipunayf: I did this initially before I split them into two. Since there are usages that do not consider the cycles, I kept them as two different functions. However, since the performance dip is negligible, I can merge them again.

azinneera: Since we can't get rid of the cycle detection in package resolution and also as we are returning the cached list after the first call to resolution, it won't make a difference AFAIU.

azinneera (approved):
LGTM

azinneera (approved):
LGTM

Issue #34756, opened 2022-01-20
Invalid subtyping error when record type inclusion overrides a field with a record and readonly intersection type


shafreenAnfar:
Consider the below program.

```ballerina
import ballerina/http;

service /covid/status on new http:Listener(9000) {

    resource function get countries() returns CovidEntry[] {
        return covidTable.toArray();
    }

    resource function post countries(@http:Payload CovidEntry covidEntry) 
                                    returns CreatedCovidEntry|ConflictingIsoCodeError {
        if covidTable.hasKey(covidEntry.iso_code) {
            return <ConflictingIsoCodeError>{ 
                body : { 
                    errmsg : "bar"
                } 
            };
        }
        covidTable.add(covidEntry); 
        return <CreatedCovidEntry>{ body: covidEntry }; 
    }
    resource function get countries/[string iso_code]() returns CovidEntry|InvalidIsoCodeError {
        CovidEntry? covidEntry = covidTable[iso_code];
        if covidEntry is () {
            return { 
                body: { 
                    errmsg: "foo"
                }
            };
        }
        return covidEntry;
    }
}
public type CovidEntry record {|
    readonly string iso_code;
    string country;
    decimal cases;
    decimal deaths;
    decimal recovered;
    decimal active;
|};
public final table<CovidEntry> key(iso_code) covidTable = table [
    { iso_code: "AFG", country: "Afghanistan", cases: 159303, deaths: 7386, recovered: 146084, active: 5833 },
    { iso_code: "SL", country: "Sri Lanka", cases: 598536, deaths: 15243, recovered: 568637, active: 14656 },
    { iso_code: "US", country: "USA", cases: 69808350, deaths: 880976, recovered: 43892277, active: 25035097 }
];
public type CreatedCovidEntry record {|
    *http:Created;
    CovidEntry body;
|};
public type ConflictingIsoCodeError record {|
    *http:Conflict;
    ErrorMsg body;
|};
public type InvalidIsoCodeError record {|
    *http:NotFound;
    ErrorMsg body;
|};
public type ErrorMsg record {|
    string errmsg;
|};
```
Now if you make the below change, it starts complaining saying the below. 

```ballerina
public type CovidEntry readonly & record {|
    readonly string iso_code;
    string country;
    decimal cases;
    decimal deaths;
    decimal recovered;
    decimal active;
|};
```
```
[...]

PR #36707, opened 2022-06-24
[Master] Fix overriding included fields with readonly intersection types

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/34756

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


gimantha:
@MaryamZi Shall we resolve the conflicts?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -4684,6 +4691,68 @@ private List<BAnnotationAttachmentSymbol> getAnnotationAttachmentSymbols(
...
+
+                BType fieldTypeInIncludingType = fieldsOfIncludingType.get(fieldName).type;
+                if (fieldTypeInIncludingType == symTable.semanticError) {
+                    continue;
+                }
+
+                BType fieldTypeInIncludedType = includedFieldEntry.getValue().type;
+                if (fieldTypeInIncludedType == symTable.semanticError) {
+                    continue;
+                }
```
chiranSachintha, line 4735: Scenario 1:
```ballerina
type Bar record {|
    record {string x;} body;
|};


public type Baz record {|
    *Bar;
    Mat body;
|};
```
Scenario 2:
```ballerina
type Bar record {|
    record {string x;} body;
|};


public type Baz record {|
    *Bar;
    record {Mat x;} body;
|};

```
for the first scenario, we are getting an error as
```
ERROR [test3.bal:(13:5,13:8)] unknown type 'Mat'

```
For the second scenario, we are getting an error as
```
ERROR [test3.bal:(13:5,13:26)] included field 'body' of type 'record {| string x; anydata...; |}' cannot be overridden by a field of type 'record {| anydata...; |}': expected a subtype of 'record {| string x; anydata...; |}'
ERROR [test3.bal:(13:13,13:16)] unknown type 'Val'

```
this is something like two behaviours for the same scenario. Isn't it?
anyway for the second scenario error message is incorrect.

scenario 3
```
type Bar record {|
    record {} body;
|};


public type Baz record {|
    *Bar;
    record {Mat x;} body;
|};

```
When considering scenario 3 we get only the 
```
ERROR [test3.bal:(13:13,13:16)] unknown type 'Mat'

```

MaryamZi: The behaviour is due to how we resolve `record {Mat x;}` when `Mat` is undefined, right? We seem to be resolving it as `record {}` since we don't know the type of `x` to define it as a field, which IMO is acceptable.

This isn't specific to the changes introduced in this PR either. The errors are similar for the following, which does pretty much the same checks.

```ballerina
function fn(record {Mat x;} a) { // unknown type 'Mat'
    record {string x;} _ = a; // error: incompatible types: expected 'record {| string x; anydata...; |}', found 'record {| anydata...; |}'
    record {} _ = a; // no error
}
```

gimantha (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
chiranSachintha: okay. 
But here we check whether the type is defined or not. As an example
```ballerina
type Bar record {|
    record {string x;} body;
|};


public type Baz record {|
    *Bar;
    Mat body; // we check whether `Mat` defined or not.
|};
```
if it is an anon record type we didn't check types of fields of the anon type are defined or not.
So do we need these checks here?
```
if (fieldTypeInIncludingType == symTable.semanticError) {
    continue;
}

if (fieldTypeInIncludedType == symTable.semanticError) {
    continue;
}
```

MaryamZi: The type being undefined is different from a field of a type being undefined. Without these checks, I believe there'll be errors like

```cmd
included field 'body' of type 'anydata' cannot be overridden by a field of type 'other': expected a subtype of 'anydata'
``` 

which don't add any value.

PR #36768, opened 2022-06-29
Add create function support for lhs in conditional expression

### LINKED ISSUE #36321: Create function code action doesn't work in lhs expression in ternary conditional expression
Consider the following case
```
function testConditional3(int arg) {
    string res = isGreaterThanZero(arg) ? "a" : "b";
}
```
currently create function code action creates following method while we expect a `boolean` return
```
function isGreaterThanZero(int i) returns string {
    return "";
}
```

_Originally posted by @gayalkuruppu in https://github.com/ballerina-platform/ballerina-lang/pull/35815#pullrequestreview-981097638_

bimsarapathiraja:
## Purpose
Create function code action doesn't work in lhs expression in ternary conditional expression

Fixes #36321

## Approach
Check if the current cursor position is within the range of the functionalCallExpressionNode and the lhsExpression of the conditionalExpressionNode.
If True, type Boolean is returned. Otherwise the existing procedure follows.

## Check List 

- [ ] Updated Change Log
- [ ] Checked Tooling Support (#)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ]   Spec Conformance Tests
  - [ ]   Integration Tests
  - [ ]   Ballerina By Example Tests
- [x]   Increased Test Coverage
  - [ ]   API documentation
  - [ ]   Module documentation in Module.md files
  - [ ]   Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
@@ -429,8 +429,7 @@ public class FunctionCallExpressionTypeFinder extends NodeVisitor {
     @Override
     public void visit(ConditionalExpressionNode conditionalExpressionNode) {
 
-        if (this.functionCallExpressionNode != null &&
-                PositionUtil.isWithinLineRange(this.functionCallExpressionNode.lineRange(),
+        if (PositionUtil.isWithinLineRange(this.functionCallExpressionNode.lineRange(),
                 conditionalExpressionNode.lhsExpression().lineRange())) {
             checkAndSetTypeDescResult(TypeDescKind.BOOLEAN);
             return;
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
@@ -429,7 +429,8 @@ public class FunctionCallExpressionTypeFinder extends NodeVisitor {
     @Override
     public void visit(ConditionalExpressionNode conditionalExpressionNode) {
 
-        if (PositionUtil.isWithinLineRange(this.functionCallExpressionNode.lineRange(),
+        if (this.functionCallExpressionNode != null &&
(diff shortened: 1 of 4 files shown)


gayalkuruppu (commented):
Shall we fix the code action for the following case as well?
```
function testConditional4() {
    string res = getFunc() ?: "b";
}
```
Following is the expected function
```
function getFunc() returns string? {
    return "";
}
```

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
```diff
@@ -424,6 +428,13 @@ public void visit(WhileStatementNode whileStatementNode) {
 
     @Override
     public void visit(ConditionalExpressionNode conditionalExpressionNode) {
+
+        if (this.functionCallExpressionNode != null &&
```
IMS94, line 432: We don't need the null check here. Because due to the approach we are using this visitor, `functionCallExpressionNode` cannot be `null`.

IMS94 (changes requested):
Overall the implementation is good. Added a minor comment.

mohanvive:
@bimsarapathiraja Can you address the review comments please.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
bimsarapathiraja: Removing `null` condition results spot bugs

bimsarapathiraja:
> @bimsarapathiraja Can you address the review comments please.

done

gayalkuruppu:
> Shall we fix the code action for the following case as well?
> 
> ```
> function testConditional4() {
>     string res = getFunc() ?: "b";
> }
> ```
> 
> Following is the expected function
> 
> ```
> function getFunc() returns string? {
>     return "";
> }
> ```

@bimsarapathiraja

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/command/visitors/FunctionCallExpressionTypeFinder.java
IMS94: No problem then

IMS94 (approved):
LGTM!

gayalkuruppu (approved):
LGTM (other test case for `nil-conditional-expr` was added as a new issue)

PR #36775, opened 2022-06-29
Improve AfterBefore functionality for DataDriven Tests

### LINKED ISSUE #36529: Shoudn't `after` function be triggered after each execution of a test method
**Description:**
Given the test report consider each run of a test method as an individual test, $title?

**Steps to reproduce:**

**Affected Versions:**

Ibaqu:
## Purpose
After and Before functions for Data-driven tests do not execute after each value is being run. 

Fixes #36529

## Approach
- Split the execution of the 'normal test function' and the 'data-driven test' to new functions `executeTestFunction()` and `executeDataDrivenTestFunction()` 
- Move the 'after' and 'before' functions to the respective places inside the new functions

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/testerina/modules/testerina-runtime/src/main/java/org/ballerinalang/test/runtime/BTestRunner.java
+++ b/misc/testerina/modules/testerina-runtime/src/main/java/org/ballerinalang/test/runtime/BTestRunner.java
@@ -437,6 +437,7 @@ public class BTestRunner {
 
     private void executeBeforeFunction(Test test, TestSuite suite, ClassLoader classLoader, Scheduler scheduler,
                                        AtomicBoolean shouldSkip, AtomicBoolean shouldSkipTest)  {
+        if (!shouldSkip.get()) {
             // run before tests
             String errorMsg;
             try {
@@ -453,6 +454,7 @@ public class BTestRunner {
                         test, formatErrorMessage(e));
                 errStream.println(errorMsg);
             }
+        }
     }
 
     /**
--- a/misc/testerina/modules/testerina-runtime/src/main/java/org/ballerinalang/test/runtime/BTestRunner.java
+++ b/misc/testerina/modules/testerina-runtime/src/main/java/org/ballerinalang/test/runtime/BTestRunner.java
@@ -589,6 +589,12 @@ public class BTestRunner {
                             // run the after tests
                             executeAfterFunction(test, suite, classLoader, scheduler, shouldSkip, shouldSkipTest,
                                     failedAfterFuncTests);
+                        } else {
+                            // If the test function is skipped lets add it to the failed test list
+                            failedOrSkippedTests.add(test.getTestName());
+                            // report the test result
+                            functionResult = new TesterinaResult(test.getTestName(), false, true, null);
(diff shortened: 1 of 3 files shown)


misc/testerina/modules/testerina-runtime/src/main/java/org/ballerinalang/test/runtime/BTestRunner.java
```diff
@@ -435,7 +436,6 @@ private void executeBeforeEachFunction(Test test, TestSuite suite, ClassLoader c
 
     private void executeBeforeFunction(Test test, TestSuite suite, ClassLoader classLoader, Scheduler scheduler,
                                        AtomicBoolean shouldSkip, AtomicBoolean shouldSkipTest)  {
-        if (!shouldSkip.get() && !shouldSkipTest.get()) {
```
Dilhasha, line 438: If we are to skip the test, we don't need to run the `before` test for that too right?

misc/testerina/modules/testerina-runtime/src/main/java/org/ballerinalang/test/runtime/BTestRunner.java
```diff
@@ -516,61 +515,110 @@ private void invokeDataDrivenTest(TestSuite suite, String testName, String key,
...
+                    computeFunctionResult(test.getTestName(), packageName, shouldSkip, failedOrSkippedTests,
+                            new Error("The provided data set is empty."));
+                } else {
+                    List<String> keyValues = getKeyValues((BMap) valueSets);
+                    Class<?>[] argTypes = extractArgumentTypes((BMap) valueSets);
+                    List<Object[]> argList = extractArguments((BMap) valueSets);
+                    for (int i = 0, argListSize = argList.size(); i < argListSize; i++) {
+                        // Reset shouldSkipTest for each iteration, since Before function may not fail for other
+                        // iterations
+                        shouldSkipTest.set(false);
```
Dilhasha, line 578: Before function has the same logic right(It's not specific to the data that gets passed to the data-driven test) ? If it fails once, it's likely to fail in each iteration too.
I don't think we should reset.

misc/testerina/modules/testerina-runtime/src/main/java/org/ballerinalang/test/runtime/BTestRunner.java
Ibaqu: I have added back the condition incase the beforeEach function fails.

misc/testerina/modules/testerina-runtime/src/main/java/org/ballerinalang/test/runtime/BTestRunner.java
Ibaqu: As discussed in the review we are special casing the data driven testcases so we need this particular reset to be implemented

azinneera (approved):

PR #36670, opened 2022-06-22
Introduce includes field to Ballerina.toml

gayaldassanayake:
## Purpose
Introduce a new field `includes` to `Ballerina.toml`. The paths of the files that needs to be included in the BALA file when using `bal pack` are specified here. The files/directories can be specified using patterns.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/36429, https://github.com/ballerina-platform/ballerina-spec/issues/988

## Approach
The files that match the patterns given in `includes` and packing them into the BALA.

### Project Structure
```
foo/
. icon.png
. Ballerina.toml
. main.bal
. modules/
. . sub-mod-1/
. . . icon.jpg
. . . docs/
. . . . email.txt
. . . . header.txt
. . . sub_mod1.bal
```

### Ballerina.toml
``` toml
includes = ["icon.*", "docs/"]
```

This will include
- `icon.jpg`, `icon.png`, `docs/email.txt`, `docs/header.txt`

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #36776, opened 2022-06-29
Copy include field files when creating from a template

nipunayf:
Extended #36670


cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/NewCommandTest.java
```diff
@@ -464,6 +464,30 @@ public void testNewCommandCentralPullWithCentralDependency() throws IOException
...
+        // Create a new package with the foo/winery:0.1.0 template
+        String packageName = "sample_project";
+        String[] args = {packageName, "-t", "foo/winery:0.1.0"};
+        NewCommand newCommand = new NewCommand(tmpDir, printStream, false, homeCache);
+        new CommandLine(newCommand).parseArgs(args);
+        newCommand.execute();
+
+        // Check the include files
+        Path packageDir = tmpDir.resolve(packageName);
+        Assert.assertTrue(Files.exists(packageDir));
```
azinneera, line 482: Shall we also check the addition of a file that is in the package root?

compiler/ballerina-lang/src/main/resources/ballerina-toml-schema.json
```diff
@@ -121,6 +121,12 @@
             "type": "string"
           }
         },
+        "includes": {
```
azinneera, line 124: Shall we make this `include` instead of `includes`? This is to be consistent with the imperative mood we have used for `export`

PR #36715, opened 2022-06-24
Fix `toJsonString()` API providing wrong values

HindujaB:
## Purpose
$subject

Fixes #36332

## Approach
Converted the value into Json before getting the string value.

## Samples
Use this API with unescaped string values. 
https://github.com/ballerina-platform/ballerina-lang/blob/86afb2400cbba87ed72b7c2cd56b0e0193e15a17/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/StringUtils.java#L351

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [x] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #36844, opened 2022-07-04
Fix `toJsonString()` API providing wrong values - master

### LINKED ISSUE #36332: Output of `getJsonString` fails JSON-parse
Consider the following `getJsonString` function in `StringUtils` which supposed to returns the json string value of Ballerina values.
https://github.com/ballerina-platform/ballerina-lang/blob/f8f34ab671e36f3b6f7c912ee915967fbabb700d/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/StringUtils.java#L351-L377

In that function we are calling `stringValue` which is overridden in implementations of `BValue`.

For an example, if we consider the overrriden implementation of `stringValue` in `MapValueImpl`,
https://github.com/ballerina-platform/ballerina-lang/blob/f8f34ab671e36f3b6f7c912ee915967fbabb700d/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/MapValueImpl.java#L466-L494

It depends on the `informalStringValue` for a given value.
https://github.com/ballerina-platform/ballerina-lang/blob/f8f34ab671e36f3b6f7c912ee915967fbabb700d/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/StringValue.java#L73-L75
Since informal string value does not escape characters like double quotes `"`, `\n`, `\t`, the final result will not be JSON parse-able, which makes the output string of `getJsonString` for a given Ballerina value is not JSON parse-able.

HindujaB:
## Purpose
$subject

Fixes #36332

## Approach
Converted the value into Json before getting the string value.

## Samples
Use this API with unescaped string values. 
https://github.com/ballerina-platform/ballerina-lang/blob/86afb2400cbba87ed72b7c2cd56b0e0193e15a17/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/StringUtils.java#L351

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


HindujaB:
Duplicate of https://github.com/ballerina-platform/ballerina-lang/pull/36715

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/JsonUtils.java
```diff
@@ -258,6 +277,114 @@ public static void serialize(Object json, Writer writer) throws BError {
...
+                newValue = convertArrayToJson((BArray) value, unresolvedValues);
+                break;
+            case TypeTags.TABLE_TAG:
+                BTable bTable = (BTable) value;
+                Type constrainedType = ((TableType) bTable.getType()).getConstrainedType();
+                if (constrainedType.getTag() == TypeTags.MAP_TAG) {
+                    newValue = convertMapConstrainedTableToJson((BTable) value, unresolvedValues);
+                } else {
+                    try {
+                        newValue = io.ballerina.runtime.internal.JsonUtils.toJSON(bTable);
```
Nadeeshan96, line 321: Shall we import `io.ballerina.runtime.internal.JsonUtils` and call as `JsonUtils.toJSON` here for readability? There are more places like this below.

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -56,114 +28,7 @@
 public class ToJson {
```
Nadeeshan96, line 28: Shall we add a private constructor to this class which has only static methods?

langlib/lang.value/src/main/java/org/ballerinalang/langlib/value/ToJson.java
```diff
@@ -56,114 +28,7 @@
...
-                newValue = convertMapToJson((BMap<?, ?>) value, unresolvedValues);
-                break;
-            case TypeTags.ERROR_TAG:
-            default:
-                throw createConversionError(value, jsonType);
-        }
-
-        unresolvedValues.remove(typeValuePair);
-        return newValue;
+        return io.ballerina.runtime.api.utils.JsonUtils.convertToJson(value, new ArrayList<>());
```
Nadeeshan96, line 31: Another place where we can replace the fully qualified name if required.

misc/ls-extensions/modules/bal-shell-service/src/main/java/io/ballerina/shell/service/ShellValue.java
```diff
@@ -29,10 +30,11 @@ public class ShellValue {
     private final String mimeType;
     private final String type;
 
-    public ShellValue(String value, String type, int tag) {
-        this.value = value;
-        this.type = type;
-        this.mimeType = TypeUtils.getMimeTypeFromName(tag);
+    public ShellValue(Object value) {
+        Type type = io.ballerina.runtime.api.utils.TypeUtils.getType(value);
```
Nadeeshan96, line 34: Another place where we can replace the fully qualified name if required.

misc/ls-extensions/modules/bal-shell-service/src/main/java/io/ballerina/shell/service/util/TypeUtils.java
```diff
@@ -51,4 +53,33 @@ public static String getMimeTypeFromName(int typeTag) {
...
+
+    /**
+     * Returns json string for a given object if that object is accepted as json convertible type
+     * otherwise returns the string representation for the object.
+     *
+     * @param value object needs to be converted
+     * @return converted string
+     */
+    public static String convertToJsonIfAcceptable(Object value) {
+        Type type = io.ballerina.runtime.api.utils.TypeUtils.getType(value);
```
Nadeeshan96, line 65: Another place where we can replace the fully qualified name if required.

HindujaB: This is because the class name `TypeUtils` is common here. So, on the other way we can import the function `io.ballerina.runtime.api.utils.TypeUtils.getType` and use. 
The same happens for the `JsonUtils` as well. 
I'll change the other places.

misc/ls-extensions/modules/bal-shell-service/src/test/resources/testcases/getResult/mime_types.json
```diff
@@ -80,7 +80,7 @@
     "source": "t",
     "result": {
       "shellValue": {
-        "value": "[{\"username\":\"John\",\"salary\":100,\"fullname\":{\"firstname\":\"John\",\"lastname\":\"Doe\"}},{\"username\":\"Adam\",\"salary\":300,\"fullname\":{\"firstname\":\"Adam\",\"lastname\":\"Smith\"}},{\"username\":\"Jake\",\"salary\":100,\"fullname\":{\"firstname\":\"jake\",\"lastname\":\"Peralta\"}}]",
+        "value": "[{\"username\":\"John\", \"salary\":100, \"fullname\":{\"firstname\":\"John\", \"lastname\":\"Doe\"}}, {\"username\":\"Adam\", \"salary\":300, \"fullname\":{\"firstname\":\"Adam\", \"lastname\":\"Smith\"}}, {\"username\":\"Jake\", \"salary\":100, \"fullname\":{\"firstname\":\"jake\", \"lastname\":\"Peralta\"}}]",
```
warunalakshitha, line 83: Why we need to change these?

HirumalPriyashan: The string value of `value` is generated by using this `getJsonString()` method,  with the previous implementation it does not add a space to separate values, but the new one does.
Since the generated value is still JSON parse-able it still satisfies the use case in `bal-shell-service`

HindujaB: This is the expected behavior. Because with this fix we convert the record value to json before getting the json string representation. So the record value will be converted to a `map<json>` value. The json-string of a map value contains spaces between fields. 

The same behavior can be seen in the langlib method `toJsonString()` as well.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/JsonUtils.java
```diff
@@ -258,6 +281,114 @@ public static void serialize(Object json, Writer writer) throws BError {
         }
     }
 
+    public static Object convertToJson(Object value, List<TypeValuePair> unresolvedValues) {
+        Type jsonType = PredefinedTypes.TYPE_JSON;
+
```
warunalakshitha, line 286: Shall we remove the unwanted spaces if we copy the same code?

warunalakshitha: There are some unwanted spaces in the convertToJson method.

warunalakshitha (approved):
LGTM

Issue #36429, opened 2022-06-03
Allow a Package to Specify Which Files to Include When Creating a BALA File


gayaldassanayake:
Description: https://github.com/ballerina-platform/ballerina-spec/issues/988

pramodya1994:
There's a requirement in the graphql package of resources API.
As per the internal discussion had with @DimuthuMadushan, He confirmed it is used in runtime, So we can replace that with golang `@embed` like design.
This is the HTML file they are using as response dispatch to the browser. ATM It's only 46 lines long.
https://github.com/ballerina-platform/module-ballerina-graphql/blob/master/native/src/main/resources/graphiql.html

gayaldassanayake:
Reopening since the usage of patterns will be completed only after the merging of https://github.com/ballerina-platform/ballerina-lang/pull/36877

Ibaqu:
Add both PRs to 2.x branch

PR #36877, opened 2022-07-06
Allow use of patterns for `include` field in Ballerina.toml

gayaldassanayake:
## Purpose
Pull request https://github.com/ballerina-platform/ballerina-lang/pull/36670 introduces the `include` field that can be used to pack additional files into the BALA. However, the user has to specify the complete path relative to the package root in the include array.

This PR enables the user to specify patterns similar to .gitignore in the `include` array.

Updates the `include` field to match the requirements of https://github.com/ballerina-platform/ballerina-lang/issues/36429, https://github.com/ballerina-platform/ballerina-spec/issues/988

## Approach
Introduce `getPathsMatchingIncludePattern` that finds all the file paths matching to a given pattern. 

## Samples
Current implementation
``` toml
include = ["documents/template.txt", "modules/submodule/images/"]
```

New implementation
```toml
include = ["documents/*.txt", "images/"]
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


gayaldassanayake:
**Need to be discussed**

1. 
In the current implementation of BALA creation, we are adding the includes in the default module to the root. But the includes in the non-default modules are added to `modules/<package_name>.<module_name>`. 

However the bala contains a separate directory inside the `modules` directory for the default module (`modules/<package_name>`). 

Wouldn't it be consistent if we have the default module includes to `modules/<package_name>`?

2. 
In the creation of new packages using existing templates, the non-default modules are copied directly with all their content. Therefore the include copying logic there is redundant for non-default modules. 

But include copying logic is needed for the default module, since that content is in the root of the BALA, not in `modules/<package_name>`.

If we are to do the suggested change in [1], we can safely eliminate the include copying logic.

azinneera (approved):

PR #36963, opened 2022-07-12
Send diagnostics information with JsonToRecordConverter response

AzeemMuzammil:
Fix https://github.com/ballerina-platform/ballerina-lang/issues/36992

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [x] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/ls-extensions/modules/json-to-record-converter/src/main/java/io/ballerina/converters/JsonToRecordConverterService.java
```diff
@@ -49,26 +54,30 @@ public Class<?> getRemoteInterface() {
...
-                JsonNode inputJson = objectMapper.readTree(jsonString);
-                if (inputJson.has("$schema")) {
-                    response = JsonToRecordConverter.convert(jsonString, recordName, isRecordTypeDesc, isClosed);
+            try {
+                JsonElement parsedJson = JsonParser.parseString(jsonString);
+                if (parsedJson.isJsonObject() && parsedJson.getAsJsonObject().has("$schema")) {
+                    try {
+                        response.setCodeBlock(JsonToRecordConverter.convert(jsonString, recordName, isRecordTypeDesc,
+                                isClosed).getCodeBlock());
+                    } catch (IOException | JsonToRecordConverterException | FormatterException |
```
lnash94, line 70: Is it possible to use a generic `Exception`  instead of using a separate exceptions?

AzeemMuzammil: Using `Exception` instead of its subclasses cause build failure. The performance check something.

NipunaRanasinghe: @AzeemMuzammil it should be spotbugs. because Catching `Exception`s are generally not recommended, since it will catch all the runtime exceptions. But in case you really need to do that, we can ignore the spotbug warning and catch `Exception`

lnash94 (approved):
LGTM

NipunaRanasinghe (dismissed):
LGTM

NipunaRanasinghe (approved):
LGTM

PR #36983, opened 2022-07-14
Improve completions after configurable keyword in ModulePart cxt

### LINKED ISSUE #36323: Type suggestions are not provided after `configurable` keyword when its followed by a function
```
configurable <cursor>

public function main() {
    
}
``` 
Here the completion items include module level completions items (like main function, type defs, service declarations, etc). This needs to be fixed to suggest types instead.

_Originally posted by @mindula in https://github.com/ballerina-platform/ballerina-lang/issues/36189#issuecomment-1131190584_

LakshanWeerasinghe:
## Purpose
$subject

Fixes #36323

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ModulePartNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ModulePartNodeContext.java
@@ -143,6 +143,7 @@ public class ModulePartNodeContext extends AbstractCompletionProvider<ModulePart
                 break;
             case CONFIGURABLE_KEYWORD:
                 completionItems.addAll(this.getTypeDescContextItems(context));
+                break;
             default:
                 break;
         }
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ModulePartNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ModulePartNodeContext.java
@@ -33,6 +33,7 @@ import org.ballerinalang.langserver.completions.providers.context.util.ModulePar
 import org.ballerinalang.langserver.completions.providers.context.util.ServiceTemplateGenerator;
 import org.ballerinalang.langserver.completions.util.QNameRefCompletionUtil;
 import org.ballerinalang.langserver.completions.util.Snippet;
+import org.ballerinalang.langserver.completions.util.SortingUtil;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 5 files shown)


language-server/modules/langserver-core/src/test/java/org/ballerinalang/langserver/completion/ModulePartNodeContextTest.java
```diff
@@ -34,6 +38,12 @@ public void init() throws Exception {
         preLoadAndInit();
     }
 
+    @Test(dataProvider = "completion-data-provider")
+    @Override
+    public void test(String config, String configPath) throws WorkspaceDocumentException, IOException {
+        super.test(config, configPath);
+    }
+
```
malinthar, line 46: Let's remove this unwanted changes

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ModulePartNodeContext.java
```diff
@@ -56,7 +56,6 @@ public ModulePartNodeContext() {
     @Override
     public List<LSCompletionItem> getCompletions(BallerinaCompletionContext context, ModulePartNode node) {
         List<LSCompletionItem> completionItems = new ArrayList<>();
-
```
malinthar, line 59: Let's remove this unwanted formatting changes

malinthar (changes requested):
1. We have to figure out the `type descriptors` allowed to be configurable.
Thereafter, we have to decide if we are going to sort them to the top or provide completions only for allowed `type descriptors` + modules.

2. Shall we add test cases covering following cases as well.

case 1

``` ballerina
import ballerina/module1;

configurable s<cursor>

function getInt() returns int {
    return 12;
}
```
case 2

```ballerina
import ballerina/module1;

configurable module1:<cursor>

function getInt() returns int {
    return 12;
}
```
case 3

``` ballerina
import ballerina/module1;

configurable <cursor>
int a = 10;

function getInt() returns int {
    return 12;
}
```

language-server/modules/langserver-core/src/test/java/org/ballerinalang/langserver/completion/ModulePartNodeContextTest.java
LakshanWeerasinghe: @malinthar If we remove this the test case won't execute during the build. Am I correct?

malinthar: We already have a data provider with the same name in the class. So we don't need this ryt?

Issue #36861, opened 2022-07-05
Code coverage drops all lines of a bal file when a function in that file is being mocked


gabilang:
**Description:**

Code coverage report shows all of actually covered lines by tests are not covered when that particular `bal` file has a function mocked by another function. If the mocked function is moved to a separate `bal` file, the issue is gone away and code-coverage backs to normal. 

**Steps to reproduce:**
case 1: For mock function calls:
```ballerina
function stringAdd(string s) returns string {
    return "Hello " + s;
}

public function floatAdd(float a, float b) returns float {
    return (a + b);
}
```
in tests test.bal
```ballerina
import ballerina/test;

public function mockFloatAdd(float a, float b) returns (float) {
    return 5;
}

@test:Mock {
    functionName: "floatAdd",
    moduleName: "function_mocking"
}
test:MockFunction mock_floatAdd = new();

@test:Config {}
function testStringAdd() {
    test:assertEquals(stringAdd("Ballerina"), "Hello Ballerina");
}

@test:Config {}
function testFloatAdd() {
    test:when(mock_floatAdd).call("mockFloatAdd");
    test:assertEquals(floatAdd(2, 5), 5.0);
}
```
Case 2: For legacy (old version of) mocking cases:
main.bal
```ballerina
function stringAdd(string s) returns string {
    return "Hello " + s;
}

public function intAdd(int a, int b) returns (int) {
    return (a + b);
}
```
test.bal
```ballerina
import ballerina/test;

@test:Mock {
    functionName: "intAdd",
    moduleName: "function_mocking_legacy"
}
function mockIntAdd(int a, int b) returns int {
    return 7;
}

@test:Config {}
public function testIntAdd() {
    test:assertEquals(intAdd(3, 2), 7);
}

@test:Config {}
function testStringAdd() {
    test:assertEquals(stringAdd("World"), "Hello World");
}
```
If we move the `intAdd` function to a separate bal file, code-coverage back to normal.  

**Affected Versions:**
2201.1.x

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

https://github.com/wso2-enterprise/internal-support-ballerina/issues/116

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

gabilang:
[...]

PR #37019, opened 2022-07-15
Fix code coverage drops all lines of a bal file when a function in that file is being mocked 

warunalakshitha:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/36861

## Approach
We need to use offline instrumentation for classes which needs to be transform for mocking during runtime.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/testerina-integration-test/src/test/resources/project-based-tests/mocking-coverage-tests/mock.bal
```diff
@@ -0,0 +1,23 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+public function intAdd(int a, int b) returns int {
+    return a + b;
+}
+
+public function byteAdd(byte a, byte b) returns int {
+    return a + b;
+}
```
gabilang, line 23: Shall we add a new line here?

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
```diff
@@ -531,6 +555,7 @@ private int runTestSuit(Path testCachePath, Target target, Package currentPackag
         // Adds arguments to be read at the Test Runner
         // Index [0 - 3...]
         cmdArgs.add(target.path().toString());
+        cmdArgs.add(jacocoAgentJarPath);
```
Ibaqu, line 558: Here we are adding the `jacocoAgentJarPath` regardless of wether the code coverage or report flag is provided. 

Would it make sense to have the jacocoAgentJarPath be set only when the coverage is requested like how it used to be? Or would that change break the code coverage?

Ibaqu: The condition used to be `if (coverage is enabled) then set jacoco jar path`.  But now we set the jar path to jacoco everytime. Would declaring the string variable outside the condition and initializign it inside work?

warunalakshitha: Fixed the code as if codecoverage is false it will be empty string.

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
```diff
@@ -193,13 +196,15 @@ public void execute(Project project) {
...
         // in packages.
+        List<String> mockFunctionClassNames = new ArrayList<>();
         for (ModuleDescriptor moduleDescriptor :
                 project.currentPackage().moduleDependencyGraph().toTopologicallySortedList()) {
             Module module = project.currentPackage().module(moduleDescriptor.name());
             ModuleName moduleName = module.moduleName();
 
             TestSuite suite = testProcessor.testSuite(module).orElse(null);
 
+
```
HindujaB, line 207: Shall we remove this unnecessary new line?

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
```diff
@@ -228,14 +233,25 @@ public void execute(Project project) {
                     module.isDefaultModule() ? moduleName.toString() : module.moduleName().moduleNamePart();
             testSuiteMap.put(resolvedModuleName, suite);
             moduleNamesList.add(resolvedModuleName);
+            Map<String, String> mockFunctionMap = suite.getMockFunctionNamesMap();
+            for (Map.Entry<String, String> entry : mockFunctionMap.entrySet()) {
+                String key = entry.getKey();
+                String functionToMockClassName;
+                if (key.contains(MOCK_ANNOTATION_DELIMITER)) {
```
HindujaB, line 240: Shall we check whether this logic fails if the function name contains `#` or `~`?

azinneera (dismissed):
LGTM

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
warunalakshitha: I have used the same logic in other usages as well. Seems not working. I have created a issue since same issue can be observed update 1 as well.

warunalakshitha: Created https://github.com/ballerina-platform/ballerina-lang/issues/37027

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
```diff
@@ -254,7 +271,7 @@ public void execute(Project project) {
                         throw createLauncherException("error occurred while generating test report :", e);
                     }
                 }
-            } catch (IOException | InterruptedException e) {
+            } catch (IOException | InterruptedException | ClassNotFoundException e) {
```
warunalakshitha, line 274: CNF issue will never happen unless there is a bug.

gabilang (approved):
LGTM

PR #36638, opened 2022-06-21
 Add type checking changes to allow annotations for each member in a tuple

ushirask:
Fixes #36390

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #37041, opened 2022-07-18
Change hash for tuple in TypeHashVisitor

### LINKED ISSUE #37038: Type loosing type name when there is annonymous type of same shape
**Description:**
The type name of a tuple typedesc is removed at runtime if there is an anonymous type of the same shape

**Steps to reproduce:**
```
import ballerina/io;

function foo() {
    var [a, ...b] = getData();
    io:println(a, b);
}

function getData() returns [int, int, string...] {
    return [1, 2, "hello", "world"];
}

type T [int, string...];

public function main() {
    T x = [1, "a", "b"];
    io:println(x);
}
```
Here type name of `x` is `""` and not `"T"` as expected.
 
**Affected Versions:**

ushirask:
Fixes #37038 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeHashVisitor.java
```diff
@@ -414,7 +414,7 @@ public Integer visit(BTupleType type) {
             return 0;
         }
         List<Integer> tupleTypesHashes = getOrderedTypesHashes(type.getTupleTypes());
-        Integer hash = hash(baseHash(type), tupleTypesHashes, visit(type.restType), type.flags);
+        Integer hash = hash(baseHash(type), tupleTypesHashes, visit(type.restType), type.flags, type.tsymbol);
```
MaryamZi, line 417: We don't quite use the symbol in other places, right? How come we don't have the same issue with records, for example? Just wondering if this is the correct place to fix.

pcnfernando: My thinking is; that it should be updated for other cases as well.  @HindujaB @warunalakshitha 

btw @ushirask whether this is not reproducible for records and other cases?

ushirask: The flags are different for the 2 types in the record case, I'll check and see why it's not the same when in comes to tuples.

KRVPerera: I am continuing the discussion here which started the below comment.
https://github.com/ballerina-platform/ballerina-lang/pull/36638/files#r940983777

`tsymbol` does not override the hash function. Therefore if we add it like this hash will be the hash of the address value which is going to change from compilation to compilation. We need an invariant value.

KRVPerera: Check the code in `BIRPackageSymbolEnter` https://github.com/ballerina-platform/ballerina-lang/blob/bbd084cd679015c8aff2ef56d126ded38030a5cd/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java#L1516

This code is run when reading from BIR instead of Ballerina code. We are not adding a name here and creating a new symbol here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
```diff
@@ -1045,6 +1045,7 @@ public BLangNode transform(TypeParameterNode typeParameterNode) {
     public BLangNode transform(TupleTypeDescriptorNode tupleTypeDescriptorNode) {
         BLangTupleTypeNode tupleTypeNode = (BLangTupleTypeNode) TreeBuilder.createTupleTypeNode();
         SeparatedNodeList<Node> types = tupleTypeDescriptorNode.memberTypeDesc();
+        boolean isAnonymous = checkIfAnonymous(tupleTypeDescriptorNode);
```
KRVPerera, line 1048: Is this Boolean used?

ushirask: It's not used, will remove.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangTupleTypeNode.java
```diff
@@ -40,6 +40,8 @@ public class BLangTupleTypeNode extends BLangType implements TupleTypeNode {
     public List<BLangType> memberTypeNodes = new ArrayList<>();
     public BLangType restParamType;
 
+    public boolean isAnonymous;
```
MaryamZi, line 43: Why do we need a separate field for this? We can check via flags when necessary, right?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/tuples/BasicTupleTest.java
```diff
@@ -325,6 +327,13 @@ public void testTupleAsTupleFirstMember() {
         BRunUtil.invoke(result, "testTupleAsTupleFirstMember");
     }
 
+    @Test
+    public void testTupleName() {
+        Object returns = BRunUtil.invoke(result, "testTupleName", new Object[]{});
+        Type t = TypeUtils.getType(returns);
+        Assert.assertEquals(t.getName(), "T");
+    }
```
MaryamZi, line 335: Don't we need a runtime test for arrays also?

ushirask: Is there a way to test this since `ArrayValueImpl` doesn't have a type name?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangTupleTypeNode.java
ushirask: Fixed.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/tuples/BasicTupleTest.java
MaryamZi: Please share the issue here too.

ushirask: #37470

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -1443,13 +1443,15 @@ public BType transform(BLangTupleTypeNode tupleTypeNode, AnalyzerData data) {
             memberTypes.add(type);
         }
 
-        BTypeSymbol tupleTypeSymbol = Symbols.createTypeSymbol(SymTag.TUPLE_TYPE, Flags.asMask(EnumSet.of(Flag.PUBLIC)),
+        EnumSet<Flag> flags = tupleTypeNode.flagSet.contains(Flag.ANONYMOUS) ?
```
pcnfernando, line 1446: With this change, the actual flag set will be a subset of the original flags, isn't it? Any reason for this?

ushirask: Only `ANONYMOUS` flag will be added before here right, so no flags will be lost?

pcnfernando: We anyway hadn't used any other flags as well. I wondered whether propagating all the flags to the type symbol would be feasible?

ushirask: Updated for the tuple transform. Should we change for other such cases here as well?

pcnfernando:
@ushirask `testAssignmentNegativeCases` seems to be failing

pcnfernando:
@ushirask can you check on the build

pcnfernando (approved):

tests/jballerina-unit-test/src/test/resources/test-src/types/tuples/tuple_basic_test.bal
```diff
@@ -487,6 +487,16 @@ function testTupleToJSONAssignment() {
      assertEquality(false, K is json);
 }
 
+type TC [int, string...];
+TC tc = [1, "2"];
```
MaryamZi, line 491: Why aren't we testing the name of this also?

tests/jballerina-unit-test/src/test/resources/test-src/types/tuples/tuple_basic_test.bal
```diff
@@ -487,6 +487,16 @@ function testTupleToJSONAssignment() {
...
+
+type T [int, string...];
```
MaryamZi, line 493: Please add similar tests for tuples without a rest descriptor also.

MaryamZi: And combinations of named and anonymous tuples.

E.g.,

```ballerina
type TC [int, string...];

function testFn() returns [TC, TC] {
    TC tc = [1, "2"];
    [int, string...] x = [1, "a", "b"];
    return [tc, x];
}
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/tuples/BasicTupleTest.java
```diff
@@ -325,6 +328,22 @@ public void testTupleAsTupleFirstMember() {
...
 
+    @Test
+    public void testTupleName() {
+        Object returns = BRunUtil.invoke(result, "testTupleName", new Object[]{});
+        TupleValueImpl returnTuple = (TupleValueImpl) returns;
+
+        Type t1 = TypeUtils.getType(returnTuple.get(0));
+        Type t2 = TypeUtils.getType(returnTuple.get(1));
+        Type t3 = TypeUtils.getType(returnTuple.get(2));
+        Type t4 = TypeUtils.getType(returnTuple.get(3));
```
MaryamZi, line 339: Do we need these separate variables? Even if we are continuing with them can we move them closer to usage?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/tuples/BasicTupleTest.java
```diff
@@ -325,6 +328,22 @@ public void testTupleAsTupleFirstMember() {
...
+
+        Assert.assertEquals(t1.getName(), "T");
+        Assert.assertEquals(t2.getName(), "");
+        Assert.assertEquals(t3.getName(), "T2");
+        Assert.assertEquals(t4.getName(), "TC");
```
MaryamZi, line 344: Shall we also assert the string representation? We'd then need the separate variable (re: previous comment).

tests/jballerina-unit-test/src/test/resources/test-src/types/tuples/tuple_basic_test.bal
```diff
@@ -487,6 +487,20 @@ function testTupleToJSONAssignment() {
...
+
+type T2 [int, string, int];
+
+function testTupleName() returns [T, T, T2, TC] {
+    T x1 = [1, "a", "b"];
+    [int, string...] x2 = [2, "b", "c"];
+    T2 x3 = [1, "a", 2];
+    return [x1, x2, x3, tc];
```
MaryamZi, line 501: Shall we include a module-level anonymous tuple also?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/types/tuples/BasicTupleTest.java
```diff
@@ -325,6 +328,32 @@ public void testTupleAsTupleFirstMember() {
...
     }
 
+    @Test
+    public void testTupleName() {
+        Object returns = BRunUtil.invoke(result, "testTupleName", new Object[]{});
+        TupleValueImpl returnTuple = (TupleValueImpl) returns;
+
+        Type t1 = TypeUtils.getType(returnTuple.get(0));
+        Assert.assertEquals(t1.getName(), "T");
+        Assert.assertEquals(t1.toString(), "[int,string...]");
```
MaryamZi, line 338: If there is a name, why isn't `toString` returning that? Isn't that what we do for records?

ushirask: We're currently computing the string representation and returning, should we change that similar to records for named tuples?

Issue #36104, opened 2022-05-11
Relational operator doesn't work in same way for `null` and `()`


SandaruJayawardana:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type x boolean|null; // works for `()`
type y boolean|();

x a1 = ();
y b1 = true;

boolean x2 = a1 < b1; // operator '<' not defined for 'x' and 'y'
```
**Affected Versions:**
2201.1.0 (Swan Lake)

Issue #36103, opened 2022-05-11
Relational operator doesn't support tuples when one is readonly


SandaruJayawardana:
**Description:**
$title.

**Steps to reproduce:**
```ballerina
type x [boolean?, int] & readonly;
type y [boolean?, int]; 

x a1 = [(), 1];
y b1 = [()];

boolean x2 = a1 < b1; // operator '<' not defined for '[boolean?,int] & readonly' and 'y
```
**Affected Versions:**
2201.1.0 (Swan Lake)

Issue #36102, opened 2022-05-11
Relational operator doesn't work properly for tuples and arrays


SandaruJayawardana:
**Description:**
$title. Consider following scenarios.

**Steps to reproduce:**
```ballerina
type x [boolean?, int]|[boolean, int, decimal?, string]; // ordered type -> [boolean?, int, decimal?, string]
type y [boolean?, int]|[boolean, int?, decimal]; // ordered type -> [boolean?, int, decimal]
// for x and y, ordered type can be taken as [boolean?, int, decimal?, string]

x a1 = [(), 1];
y b1 = [()];

boolean x2 = a1 < b1; // error: operator '<' not defined for 'x' and 'y'
```

```ballerina
type x int?[2]|int[5]; // ordered type -> [int?, int?, int, int, int]
type y int?[2]; // ordered type -> [int?, int?]
// for x and y, ordered type can be taken as [int?, int?, int, int, int]

x a1 = [(), 1];
y b1 = [()];

boolean x2 = a1 < b1; // operator '<' not defined for 'x' and 'y'
```
**Affected Versions:**
2201.1.0 (Swan Lake)

MaryamZi:
The requirement is that there is an ordered type to which both operands belong to.

> ```ballerina
> type x [boolean?, int]|[boolean, int, decimal?, string]; // ordered type -> [boolean?, int, decimal?, string]
> type y [boolean?, int]|[boolean, int?, decimal]; // ordered type -> [boolean?, int, decimal]
> // for x and y, ordered type can be taken as [boolean?, int, decimal?, string]
> 
> x a1 = [(), 1];
> y b1 = [()];
> 
> boolean x2 = a1 < b1; // error: operator '<' not defined for 'x' and 'y'
> ```

`[boolean?, int]` doesn't belong to `[boolean?, int, decimal?, string]`. Neither do `[boolean?, int]` nor `[boolean, int?, decimal]`. There is no ordered type to which both operands (and therefore members of the union) belong to.

> ```ballerina
> type x int?[2]|int[5]; // ordered type -> [int?, int?, int, int, int]
> type y int?[2]; // ordered type -> [int?, int?]
> // for x and y, ordered type can be taken as [int?, int?, int, int, int]
> 
> x a1 = [(), 1];
> y b1 = [()];
> 
> boolean x2 = a1 < b1; // operator '<' not defined for 'x' and 'y'
> ```

Similarly there's no ordered type here either.

[...]

Issue #35195, opened 2022-02-23
Operator not defined error when one operand is optional and other one is non optional in relational expressions


rpjayasekara:
**Description:**
Gets operator not defined error when one operand is optional and other one is non optional in relational expressions.

```
function init() {
    decimal? a = -32.45;
    decimal b = -32;
    io:println(a >= b); // operator '>=' not defined for 'decimal?' and 'decimal'
    io:println(b >= a); // operator '>=' not defined for 'decimal?' and 'decimal'
}
```

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

SandaruJayawardana:
Related issue
```ballerina
type x false|();
type y true|();

x a1 = ();
y b1 = true;

boolean x2 = a1 < b1; // operator '<' not defined for 'x' and 'y'
````

PR #37047, opened 2022-07-18
Fix relational operator related issues

### LINKED ISSUE #37301: Relational expressions give incorrect results for `byte`
**Description:**
$title. Seems this happens when the `byte` type is used with list constructors. In following example, both `byte` and `int:Unsigned8` scenarios should give same results.

```ballerina
import ballerina/io;

type Type17 [int, int]; 
type Type18 [int, byte]; 
type Type19 [int, int:Unsigned8]; 

public function main() {
    Type17 a = [1, 1];
    Type18 b = [1, 1];
    Type19 c = [1, 1];

    int d = 1;
    byte e = 1;

    boolean x0 = a > b;
    boolean x1 = b > a;
    io:println(x0); // false [correct]
    io:println(x1); // false [correct]

    boolean x2 = a >= b;
    boolean x3 = b >= a;
    io:println(x2); // false [incorrect!]
    io:println(x3); // false [incorrect!]

    boolean x4 = a >= c;
    boolean x5 = c >= a;
    io:println(x4); // true [correct]
    io:println(x5); // true [correct]

    boolean x6 = d >= e;
    boolean x7 = e >= d;
    io:println(x6); // true [correct]
    io:println(x7); // true [correct]
}
```

**Steps to reproduce:**
Shown above.

**Affected Versions:**
Current master.

SandaruJayawardana:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/36104
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/36103
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/36102
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/35195
Fixes #37301

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3266,7 +3271,9 @@ public Boolean visit(BType target, BType source) {
                 targetTag = getEffectiveTypeForIntersection(targetType).tag;
             }
             if (isSimpleBasicType(sourceTag) && isSimpleBasicType(targetTag)) {
-                return (source == target) || isIntOrStringType(sourceTag, targetTag);
+                // If type T is ordered, then type T? Is also ordered.
+                return (source == target) || sourceTag == TypeTags.NIL || targetTag == TypeTags.NIL ||
```
KavinduZoysa, line 3275: Do we need nil check here, because we have check `isNil` in `isSameOrderedType` method?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3276,36 +3283,82 @@ public Boolean visit(BType target, BType source) {
...
+                //     [string, (), float, int...] can be taken as the ordered type which the static type of both
+                //     operands belong.
+                BType baseType = sourceTupleTypes.get(targetTupleCount);
+                if (!hasCommonOrderedTypeForTuples(baseType, sourceTupleTypes, targetTupleCount + 1)) {
+                    return false;
+                }
+            } else if (sourceTupleCount < targetTupleCount) {
+                // Target tuple has higher number of member types.
+                BType baseType = targetTupleTypes.get(sourceTupleCount);
+                if (!hasCommonOrderedTypeForTuples(baseType, targetTupleTypes, sourceTupleCount + 1)) {
```
KavinduZoysa, line 3361: Can we rearrange this logic in order to match the following pattern?
```ballerina
if (sourceTupleCount == targetTupleCount) {
} else if (sourceTupleCount < targetTupleCount) {
} else {
}
``` 
The else if case and else case are the same, so you can create a separate function for that.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3324,6 +3377,19 @@ public Boolean visit(BTupleType target, BType source) {
                     sourceRestType);
         }
 
+        private boolean hasCommonOrderedTypeForTuples(BType baseType, List<BType> typeList, int startIndex) {
```
KavinduZoysa, line 3380: Since the `baseType = typeList.get(startIndex - 1)`, can we create it here, without passing as a param?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3482,27 +3548,55 @@ public Boolean visit(BParameterizedType t, BType s) {
...
+        } else if (referredTypeKind == TypeKind.FINITE) {
+            Set<BLangExpression> valueSpace = ((BFiniteType) referredType).getValueSpace();
+            return valueSpace.size() == 1 && valueSpace.iterator().next().getBType().getKind() == TypeKind.NIL;
+        }
+        return false;
+    }
+
+    private boolean checkUnionHasSameType(BUnionType unionType, BType baseType) {
+        LinkedHashSet<BType> memberTypes = unionType.getMemberTypes();
+        boolean hasSameOrderedType = false;
```
KavinduZoysa, line 3567: Do we need this variable? 
By looking at whole logic I think we do not need this variable, returning true is enough at the end.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
SandaruJayawardana: But that method will not be called when the both `target`, `source` are simple basic types.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
SandaruJayawardana: I changed the logic into the above pattern. Since each if block contains only 3 method calls, I think we can keep them as it is to improve the readability. WDYT @KavinduZoysa?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
SandaruJayawardana: +1. We can remove that variable.

KavinduZoysa:
@SandaruJayawardana, can you check why Codecov warnings have appeared?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3293,27 +3345,49 @@ public Boolean visit(BParameterizedType t, BType s) {
...
             type = getReferredType(type);
             if (type.tag == TypeTags.FINITE) {
                 for (BLangExpression expr : ((BFiniteType) type).getValueSpace()) {
-                    isSameType = isSameOrderedType(expr.getBType(), baseType);
-                    if (!isSameType) {
+                    if (!isSameOrderedType(expr.getBType(), baseType)) {
                         return false;
                     }
                 }
             } else if (type.tag == TypeTags.UNION) {
```
SandaruJayawardana, line 3372: This scenario is not covered with unit tests (locally test with commenting out this). I guess, unions cannot be a member type of another union. Therefore, we cannot cover this part with tests. IMO we can remove this condition. But this condition is already available in the master branch. WDYT @KavinduZoysa?

KavinduZoysa: Please check whether tests are passing after removing this test.

SandaruJayawardana: Tests are passing even without this. (https://github.com/ballerina-platform/ballerina-lang/pull/37047/commits/85848bce6e929fd06d4002aeec1f9351f6ec1634)

KavinduZoysa: We can go head.

KavinduZoysa (approved):

Issue #34910, opened 2022-02-01
`NPE` in `BIROptimizer` when using query action


rdulmina:
**Description:**

Consider the below sample containing query action with where clause. It causes NPE when we return `item` inside do block.

```bal
function testReachabilityWithQueryAction() returns string {
    string[] stringArray = ["Hello", " ", "World"];

    error? unionResult = from var item in stringArray
        where item is string
        do {
            if item == "Hello" {
                return item;
            }
        };

    if unionResult is error {
    }
    return "c#";
}
```
<details>
  <summary>Stack trace.(Click to expand)</summary>

```
[2022-02-01 09:47:02,178] SEVERE {b7a.log.crash} - null 
java.lang.NullPointerException
        at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer$RHSTempVarOptimizer.visit(BIROptimizer.java:164)
        at org.wso2.ballerinalang.compiler.bir.model.BIRNode$BIRBasicBlock.accept(BIRNode.java:445)
        at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer$RHSTempVarOptimizer.lambda$visit$4(BIROptimizer.java:126)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1540)
        at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer$RHSTempVarOptimizer.visit(BIROptimizer.java:126)
        at org.wso2.ballerinalang.compiler.bir.model.BIRNode$BIRFunction.accept(BIRNode.java:403)
        at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer$RHSTempVarOptimizer.lambda$visit$1(BIROptimizer.java:103)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1540)
        at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer$RHSTempVarOptimizer.visit(BIROptimizer.java:103)
        at org.wso2.ballerinalang.compiler.bir.model.BIRNode$BIRPackage.accept(BIRNode.java:83)
        at org.wso2.ballerinalang.compiler.bir.optimizer.BIROptimizer.optimizePackage(BIROptimizer.java:82)
        at org.wso2.ballerinalang.compiler.bir.BIRGen.genBIR(BIRGen.java:271)
[...]

PR #37054, opened 2022-07-19
Set properly the symbols of var ref inside do clause

KavinduZoysa:
## Purpose

Fixes part of #34910

This PR fixes the https://github.com/ballerina-platform/ballerina-lang/issues/34910#issue-1120229801

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/query/query-action.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/query/query-action.bal
@@ -675,6 +675,48 @@ function testQueryActionWithDoClauseContainsCheck() {
     assertTrue(res is string && res == "str2");
 }
 
+function foo8() returns string{
+    string[] stringArray = ["Hello", " ", "World"];
+    error? unionResult = from var item in stringArray
+                            where item is string
+                            do {
+                                if item == "Hello" {
+                                    return item;
+                                }
+                            };
+    return "no match";
+}
+
+function foo9() returns string{
+    (string?)[] stringArray = ["Hello", " ", "World", ()];
+    error? unionResult = from var item in stringArray
+                            where item is string
+                            do {
+                                if item == "Hello" {
+                                    return item;
+                                }
+                            };
+    return "no match";
+}
+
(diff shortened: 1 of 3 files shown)


pcnfernando:
Shall we resolve the conflict and add a test for https://github.com/ballerina-platform/ballerina-lang/issues/37292

KavinduZoysa:
> Shall we resolve the conflict and add a test for #37292

Tests for this issue are already added in https://github.com/ballerina-platform/ballerina-lang/blob/master/tests/jballerina-unit-test/src/test/resources/test-src/query/query-action.bal#L661

tests/jballerina-unit-test/src/test/resources/test-src/query/query-action.bal
```diff
@@ -675,6 +675,48 @@ function testQueryActionWithDoClauseContainsCheck() {
...
+    return "no match";
+}
+
+function foo10() returns string{
+    (string|int)[] stringArray = ["Hello", 2, "World", 4];
+    error? unionResult = from var item in stringArray
+                            where item is string
+                            do {
+                                if item == "Hello" {
+                                    return item;
```
pcnfernando, line 708: It seems this narrowed symbol is not considered. 
Can you please check below is related to the fix
```ballerina
function foo10() returns "Hello" {
    (string|int)[] stringArray = ["Hello", 2, "World", 4];
    error? unionResult = from var item in stringArray
                            where item is string
                            do {
                                if item == "Hello" {
                                    return item;
                                }
                            };
    return "Hello";
}
```

pcnfernando: I assume this is because we override the `item` symbol within the if check with the original symbol; which is the narrowed symbol at `where`

KavinduZoysa: The given example is working as expected with this change

pcnfernando (approved):

pcnfernando:
@KavinduZoysa Shall we back-port to the applicable branches?

PR #37064, opened 2022-07-19
Fix issue in accessing a resource method with a included record param

### LINKED ISSUE #37060: Client resource method does not support included record param
**Description:**

To support the new `client with resource methods` in HTTP, we have a requirement to have the query parameters as an included record parameter as suggested in the [docs](https://docs.google.com/document/d/1u4eT3LRRIlAKRVriA8bKnJAmCgkLR6jvTFMap-sUuzs/edit).

The following is the definition of the resource of an untyped client :
```ballerina
public type SimpleParamType boolean|int|float|decimal|string;
public type ParamType SimpleParamType[]|SimpleParamType;

public type Params record {|
    never headers?;
    never targetType?;
    ParamType...;
|};

resource function get [string ...path](map<string|string[]>? headers = (), 
       TargetType targetType = <>, *Params params) returns targetType|ClientError = external;
```

And the following client call gives an error :

```ballerina
string response = check clientEP->/foo/bar(id=1234);
```

Error : `undefined defaultable parameter 'id'`

rdulmina:
## Purpose
$title

Fixes #37060

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/resources/compiler.properties
+++ b/compiler/ballerina-lang/src/main/resources/compiler.properties
@@ -981,8 +981,8 @@ error.cannot.attach.functions.to.abstract.object=\
 error.abstract.object.function.cannot.have.body=\
   function ''{0}'' in abstract object ''{1}'' cannot have a body
 
-error.resource.function.cannot.be.extern=\
-  external resource functions are not supported by the implementation
+error.service.resource.method.cannot.be.extern=\
+  external resource methods of service objects are not supported by the implementation
 
 error.object.init.function.cannot.be.extern=\
   object ''init'' method cannot have an ''external'' implementation
--- a/compiler/ballerina-lang/src/main/resources/compiler.properties
+++ b/compiler/ballerina-lang/src/main/resources/compiler.properties
@@ -982,7 +982,7 @@ error.abstract.object.function.cannot.have.body=\
   function ''{0}'' in abstract object ''{1}'' cannot have a body
 
 error.service.resource.method.cannot.be.extern=\
-  external resource methods of service objects are not supported by the implementation
+  external resource methods are not yet supported in service objects
 
 error.object.init.function.cannot.be.extern=\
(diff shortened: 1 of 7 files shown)


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -981,8 +981,8 @@ error.cannot.attach.functions.to.abstract.object=\
 error.abstract.object.function.cannot.have.body=\
   function ''{0}'' in abstract object ''{1}'' cannot have a body
 
-error.resource.function.cannot.be.extern=\
-  external resource functions are not supported by the implementation
+error.service.resource.method.cannot.be.extern=\
+  external resource methods of service objects are not supported by the implementation
```
MaryamZi, line 985:
```suggestion
  external resource methods are not yet supported in service objects
```

tests/jballerina-unit-test/src/test/resources/test-src/action/client_resource_access_action.bal
```diff
@@ -612,6 +612,30 @@ function testClosuresFromPathParams() {
...
+}
+
+public function testAccessingResourceWithIncludedRecordParam() {
+    MyClient10 cl = new;
+    int a = cl->/foo/bar(id = 1);
+    assertEquality(a, 1);
+
+    int b = cl->/foo/[4]/bar.post(id = 1);
+    assertEquality(b, 5);
+}
```
MaryamZi, line 637: We can add more tests, right?

e.g., 
1. calling the first method with an argument for `headers` 
2. passing a record value for `params` instead of passing named args for the fields both as a positional arg and as a named arg

MaryamZi (approved):

PR #37078, opened 2022-07-20
Add validation to check if versions are in SemVer format in getPackages

### LINKED ISSUE #36470: LS features fail due to unexpected character in package version
**Description:**
Screenshot 2022-06-05 at 21 20 34

**Steps to reproduce:**

**Affected Versions:**
2201.1.0

MacOS Monterey

gayaldassanayake:
## Purpose
`FileSystemRepository->getPackages()` returns the packages that are stored in the file system. The format is, `<org_name>/<package_name>/<version>`. 

During `bal pull` the incompletely downloaded packages are contained inside a temporary dir `<org_name>/<package_name>/<version_temp>` which is renamed correctly later. If the download is incomplete, the `_temp` prefix stays hence causing `UnexpectedCharacterException` when using `getPackages()` API.

This PR validates if the package version is a valid SemVer and ignores the package if not.

Fixes #36470

## Approach
`PackageVersion.from` is used to validate if the given version is in accordance with SemVer.

### Rational behind getPackages() changes
1.  (line 142) Ignoring organisations if the org name starts with a "." since an organization name cannot start with "."
2. (line 148) If within an organization `filesList==null` due to a path issue, should continue to add packages from other organizations.
3. (line 152) A package can start with the character ".". Hence pkgDir starting with "." should not be ignored.
4. (lines 167-174) Validation of the version to check if it matches SemVer format.

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/repositories/FileSystemRepository.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/repositories/FileSystemRepository.java
@@ -139,30 +139,38 @@ public class FileSystemRepository extends AbstractPackageRepository {
             return packagesMap;
         }
         for (File file : orgDirs) {
-            if (!file.isDirectory()) {
+            if (!file.isDirectory() || file.isHidden()) {
                 continue;
             }
             String orgName = file.getName();
             File[] filesList = this.bala.resolve(orgName).toFile().listFiles();
             if (filesList == null) {
-                return packagesMap;
+                continue;
             }
             List<String> pkgList = new ArrayList<>();
             for (File pkgDir : filesList) {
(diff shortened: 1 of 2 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/repositories/FileSystemRepository.java
```diff
@@ -139,31 +139,39 @@ public Map<String, List<String>> getPackages() {
             return packagesMap;
         }
         for (File file : orgDirs) {
-            if (!file.isDirectory()) {
+            if (!file.isDirectory() || file.isHidden()) {
```
gayaldassanayake, line 142: Since an organization name cannot start with ".", directories starting with "." (hidden dirs)are ignored

azinneera (approved):

Issue #33773, opened 2021-11-11
Mapping constructor with a `readonly` field for which the applicable contextually-expected type has a mutable default value panics at runtime


MaryamZi:
**Description:**
$title, once https://github.com/ballerina-platform/ballerina-lang/pull/33767 is merged. Without https://github.com/ballerina-platform/ballerina-lang/pull/33767 the default isn't used at all.

Once https://github.com/ballerina-platform/ballerina-lang/issues/33771 is fixed, we can call the default value closures of only the fields that are not provided, which will stop using the default values for `readonly` fields.

**Steps to reproduce:**
```ballerina
type Foo record {
    string[] x = ["foo", "bar"];
};

// Panics now with `invalid value for record field 'x': expected value of type 'string[] & readonly', found 'string[]'`,
// since `x` is initialized using the mutable value from `Foo` before the mapping constructor's value.
Foo _ = {readonly x: [""]};
```

**Affected Versions:**
master, slbeta5-SNAPSHOT once https://github.com/ballerina-platform/ballerina-lang/pull/33767 is merged.

MaryamZi:
May depend on https://github.com/ballerina-platform/ballerina-lang/issues/24077.

MaryamZi:
https://github.com/ballerina-platform/ballerina-spec/issues/544#issuecomment-1059640197.

MaryamZi:
Should be fixed with record default value changes. cc @chiranSachintha

PR #37184, opened 2022-08-01
 Introduce closures for record default values

### LINKED ISSUE #32012: Overriding a field included with record-type-inclusion doesn't remove the default value closure for that particular field being overriden

### LINKED ISSUE #33771: Introduce closures for record default values
**Description:**
$title, in line with the [spec](https://ballerina.io/ballerina-spec/spec.html#record-type-descriptor).

> A default-expression is an expression that specifies a default value for the field, which is used when the record type descriptor is used to construct a mapping value but no value is specified explicitly for the field. The type descriptor contains a 0-argument function closure for each default value. The closure is created from the expression when the type descriptor is resolved. The expression must meet the requirements for an isolated function. The closure is evaluated to create a field value each time the default is used in the construction of a mapping value. The default value does not affect the type described by the type descriptor.

Once this is fixed, we can rewrite inclusion so that it fixes https://github.com/ballerina-platform/ballerina-lang/issues/32012.

### LINKED ISSUE #38558: Intersection of readonly and record type with defaults
https://github.com/ballerina-platform/ballerina-spec/issues/544

chiranSachintha:
Fixes #33771
Fixes #32012
Fixes #38558

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


MaryamZi:
Please add tests including
- tests for https://github.com/ballerina-platform/ballerina-lang/issues/32012 
- tests to make sure the default value closure isn't called when the user specifies a value for the field
e.g.,
```ballerina
import ballerina/io;

type Foo record {|
    int x = fn();
|};

isolated function fn() returns int {
    io:println("called!");
    return 10;
}

public function main() {
    Foo _ = {x: 1};
}
```

Without this change this prints "called", but it shouldn't with the change.

MaryamZi:
Please enable the build for this branch too.

.github/workflows/pull_request_ubuntu_build.yml
```diff
@@ -11,6 +11,7 @@ on:
       - stage-swan-lake
       - 2201.[0-9]+.x
       - 2201.[0-9]+.[0-9]+-stage
+      - closure-changes-for-function-defult-values
```
MaryamZi, line 14: These can be removed?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -1292,28 +1294,15 @@ public BType readType(int cpI) throws IOException {
...
-                        recordSymbol.initializerFunc = new BAttachedFunction(initFuncName, recordInitFuncSymbol,
-                                                                             recordInitFuncType, symTable.builtinPos);
-                        recordSymbol.scope.define(initFuncName, recordInitFuncSymbol);
-                    }
-
                     recordType.typeInclusions = readTypeInclusions();
 
+                    int defaultValues = inputStream.readInt();
+                    for (int i = 0; i < defaultValues; i++) {
+                        String paramName = getStringCPEntryValue(inputStream);
```
MaryamZi, line 1301: `fieldName`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1564,8 +1563,7 @@ public void visit(BLangTypeConversionExpr astTypeConversionExpr) {
 
     @Override
     public void visit(BLangStructLiteral astStructLiteralExpr) {
-        List<BIROperand> varDcls = mapToVarDcls(astStructLiteralExpr.enclMapSymbols);
```
MaryamZi, line 1567: What was `enclMapSymbols` being used for?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -2194,16 +2179,16 @@ public void visit(BLangTableConstructorExpr tableConstructorExpr) {
...
 
-    private void visitTypedesc(Location pos, BType type, List<BIROperand> varDcls) {
+    private void visitTypedesc(Location pos, BType type) {
         BIRVariableDcl tempVarDcl =
                 new BIRVariableDcl(symTable.typeDesc, this.env.nextLocalVarId(names), VarScope.FUNCTION, VarKind
                         .TEMP);
         this.env.enclFunc.localVars.add(tempVarDcl);
         BIROperand toVarRef = new BIROperand(tempVarDcl);
-        setScopeAndEmit(new BIRNonTerminator.NewTypeDesc(pos, toVarRef, type, varDcls));
+        setScopeAndEmit(new BIRNonTerminator.NewTypeDesc(pos, toVarRef, type, Collections.emptyList()));
```
MaryamZi, line 2191: With this change, aren't we passing an empty list to both calls of `org.wso2.ballerinalang.compiler.bir.model.BIRNonTerminator.NewTypeDesc#NewTypeDesc`? May be able to refactor, please check.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -1187,42 +1148,11 @@ public void visit(BLangRecordTypeNode recordTypeNode) {
 
         for (BLangSimpleVariable bLangSimpleVariable : recordTypeNode.fields) {
             bLangSimpleVariable.typeNode = rewrite(bLangSimpleVariable.typeNode, env);
+//            bLangSimpleVariable.expr = rewrite(bLangSimpleVariable.expr, env);
```
MaryamZi, line 1151: Commented code? Is this visited elsewhere?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5838,10 +5740,72 @@ public void visit(BLangGroupExpr groupExpr) {
     @Override
     public void visit(BLangRecordLiteral recordLiteral) {
         List<RecordLiteralNode.RecordField> fields =  recordLiteral.fields;
+        updateFieldsOfRecordLiteral(recordLiteral, fields);
         fields.sort((v1, v2) -> Boolean.compare(isComputedKey(v1), isComputedKey(v2)));
         result = rewriteExpr(rewriteMappingConstructor(recordLiteral));
     }
 
+    public void visit(BFunctionPointerInvocation functionPointerInvocation) {
```
MaryamZi, line 5748:
```suggestion
    @Override
    public void visit(BFunctionPointerInvocation functionPointerInvocation) {
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -10114,10 +10049,8 @@ private BLangRecordLiteral rewriteMappingConstructor(BLangRecordLiteral mappingC
         }
 
         fields.clear();
-        BType refType = Types.getReferredType(type);
-        return refType.tag == TypeTags.RECORD ?
-                new BLangStructLiteral(pos, type, refType.tsymbol, rewrittenFields) :
-                new BLangMapLiteral(pos, type, rewrittenFields);
+        // no need of map literal
```
MaryamZi, line 10052: ?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5838,10 +5740,72 @@ public void visit(BLangGroupExpr groupExpr) {
...
+            member.key = new BLangRecordLiteral.BLangRecordKey(ASTBuilderUtil.createLiteral(pos, symTable.stringType,
+                                                                                                            fieldName));
+            member.valueExpr = addConversionExprIfRequired(closureInvocation, closureInvocation.getBType());
+
+            fields.add(member);
+        }
+    }
+
+    private void updateFieldsOfRecordLiteral(BLangRecordLiteral recordLiteral,
+                                             List<RecordLiteralNode.RecordField> fields) {
```
MaryamZi, line 5787: Aren't these the user-specified fields? I would use

```suggestion
                                             List<RecordLiteralNode.RecordField> userSpecifiedFields) {
```

MaryamZi: Let's update wherever applicable.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5838,10 +5740,72 @@ public void visit(BLangGroupExpr groupExpr) {
...
+        BType type = Types.getReferredType(recordLiteral.getBType());
+        if (type.getKind() != TypeKind.RECORD) {
```
MaryamZi, line 5789: Is this ever not a record type? Even via the lang server path?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5838,10 +5740,72 @@ public void visit(BLangGroupExpr groupExpr) {
...
+        List<String> fieldNames = new ArrayList<>();
+        for (RecordLiteralNode.RecordField field : fields) {
+            if (field.getKind() == NodeKind.RECORD_LITERAL_KEY_VALUE) {
+                BLangExpression key = ((BLangRecordLiteral.BLangRecordKeyValueField) field).key.expr;
+                if (key.getKind() == NodeKind.LITERAL) {
+                    fieldNames.add(((BLangLiteral) key).originalValue);
+                } else if (key.getKind() == NodeKind.SIMPLE_VARIABLE_REF) {
+                    fieldNames.add(((BLangSimpleVarRef) key).variableName.value);
+                }
+            }
```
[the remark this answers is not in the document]
MaryamZi: Same with spread fields.

```ballerina
import ballerina/io;

type Foo record {|
    int x = xFn();
    int y = yFn();
|};

isolated function xFn() returns int {
    io:println("xFn!");
    return 10;
}

isolated function yFn() returns int {
    io:println("yFn!");
    return 20;
}

public function main() {
    record {| int x; int y?; |} r = {x: 123};
    Foo f = {...r};
    io:println(f);
}
```

Prints both `xFn` and `yFn`, but we shouldn't call the closure for fields we know will definitely be available.

Please add tests also for different combinations.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5838,10 +5740,72 @@ public void visit(BLangGroupExpr groupExpr) {
...
+        }
+        return fieldNames;
+    }
+
+    private void updateFieldsOfRecordLiteral(List<RecordLiteralNode.RecordField> fields, List<String> fieldNames,
```
MaryamZi, line 5767: I would rename this method to reflect that we are adding the defaults for fields the user hasn't specified.

tests/jballerina-unit-test/src/test/resources/test-src/record/closed_record_type_inclusion.bal
```diff
@@ -286,9 +286,21 @@ type SimpleConstNegateExpr record {|
...
+    string f;
+};
+
+
 function testTypeInclusionWithFiniteField() {
     SimpleConstNegateExpr expr = {};
+    R1 r1 = { f : "hello"};
     assertEquality(true, expr is UnaryExpr);
+    assertEquality("hello", r1.f);
 }
```
MaryamZi, line 305: Please add more tests.

E.g.,

```
type Foo record {|
    int a = 1250;
|};

type Bar record {|
    *Foo;
    byte a = 100;
|};

function testFn() {
    Bar b = {}; // Should use the default from `Bar`, not `Foo`.
}
```

tests/jballerina-unit-test/src/test/resources/test-src/record/open_record.bal
```diff
@@ -99,9 +99,19 @@ function testStructExpressionAsIndex () returns string {
     return dpt.employees[0].family.children[dpt.employees[0].family.noOfChildren - 1];
 }
 
-function testDefaultVal () returns [string, string, int] {
+type Mat record {|
```
MaryamZi, line 102: Is this the correct file for this test? This is a closed record.

tests/jballerina-unit-test/src/test/resources/test-src/record/open_record.bal
```diff
@@ -99,9 +99,19 @@ function testStructExpressionAsIndex () returns string {
...
+    int x = fn();
+|};
+
+isolated function fn() returns int {
+    return 10;
+}
+
+function testDefaultVal () returns [string, string, int, int, int] {
```
MaryamZi, line 110:
```suggestion
function testDefaultValue() returns [string, string, int, int, int] {
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/record/OpenRecordTest.java
```diff
@@ -89,6 +89,12 @@ public void testDefaultValue() {
 
         Assert.assertTrue(returns.get(2) instanceof Long);
         Assert.assertEquals(returns.get(2), 999L);
+
+        Assert.assertTrue(returns.get(3) instanceof Long);
+        Assert.assertEquals(returns.get(3), 10L);
+
+        Assert.assertTrue(returns.get(4) instanceof Long);
+        Assert.assertEquals(returns.get(4), 1L);
```
MaryamZi, line 97: If we are continuing with this test, let's move the assertion for just this test to Ballerina.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5838,10 +5740,72 @@ public void visit(BLangGroupExpr groupExpr) {
...
+    }
+
+    private void updateFieldsOfRecordLiteral(BRecordType recordType, List<RecordLiteralNode.RecordField> fields,
+                                             List<String> fieldNames, Location pos) {
+        Map<String, BInvokableSymbol> defaultValues = ((BRecordTypeSymbol) recordType.tsymbol).defaultValues;
+        updateFieldsOfRecordLiteral(fields, fieldNames, defaultValues, pos);
+
+        List<BType> typeInclusions = recordType.typeInclusions;
+
+        for (BType typeInclusion : typeInclusions) {
```
MaryamZi, line 5804: The current tests don't test that the default is used from the including record rather that the included record, do they? See my comment on tests.

If someone changes the order by mistake it may not be captured via tests.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
```diff
@@ -187,29 +189,31 @@
...
 
 /**
- * Parameter desugar for create closures for default values.
+ * GenerateClosuresForDefaultValues for create closures for default values.
  *
  * @since 2201.3.0
  */
-public class ParameterDesugar extends BLangNodeVisitor {
-    private static final CompilerContext.Key<ParameterDesugar> PARAMETER_DESUGAR_KEY = new CompilerContext.Key<>();
+public class GenerateClosuresForDefaultValues extends BLangNodeVisitor {
```
MaryamZi, line 196: Can't we implement this as a `org.wso2.ballerinalang.compiler.tree.SimpleBLangNodeAnalyzer` instead? Can pass temporary fields via data. We can do this separately though.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
```diff
@@ -377,8 +381,16 @@ public void visit(BLangObjectConstructorExpression objectConstructorExpression)
 
     @Override
     public void visit(BLangRecordTypeNode recordTypeNode) {
+        if (((BRecordType) recordTypeNode.getBType()).mutableType != null) {
+            BRecordTypeSymbol typeSymbol =
+                    (BRecordTypeSymbol) ((BRecordType) recordTypeNode.getBType()).mutableType.tsymbol;
+            ((BRecordTypeSymbol) recordTypeNode.getBType().tsymbol).defaultValues = typeSymbol.defaultValues;
+            recordTypeNode.restFieldType = rewrite(recordTypeNode.restFieldType, env);
```
MaryamZi, line 388: Why does only the rest field type need to be visited?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
```diff
@@ -483,13 +495,22 @@ public void visit(BLangFunctionTypeNode functionTypeNode) {
 
     @Override
     public void visit(BLangSimpleVariable varNode) {
+        BLangExpression bLangExpression;
```
MaryamZi, line 498: Why do we need this here if we are returning in the if block?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
```diff
@@ -483,13 +495,22 @@ public void visit(BLangFunctionTypeNode functionTypeNode) {
...
+        if (Symbols.isFlagOn(varNode.symbol.flags, Flags.FIELD) && varNode.symbol.isDefaultable) {
+            typeNodeOfRecordField = varNode.typeNode;
+            String closureName = generateName(varNode.symbol.name.value, env.node);
+            bLangExpression = createClosureForDefaultValueOfRecordField(closureName, varNode.name.value, varNode);
+            varNode.expr = bLangExpression;
```
MaryamZi, line 503:
```suggestion
            varNode.expr = createClosureForDefaultValueOfRecordField(closureName, varNode.name.value, varNode);;
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
```diff
@@ -524,6 +545,25 @@ private BLangExpression createClosureForDefaultValue(String closureName, String
         env.enclPkg.functions.add(function);
         env.enclPkg.topLevelNodes.add(function);
         symbol.defaultValues.put(paramName, varSymbol);
+        rewrite(lambdaFunction, lambdaFunction.capturedClosureEnv);
+        return returnStmt.expr;
+    }
+
+    private BLangExpression createClosureForDefaultValueOfRecordField(String closureName, String paramName,
```
MaryamZi, line 552: This is pretty much a duplicate of the preceding method, right? Why can't we extract the common logic out to (a) separate method(s)?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
```diff
@@ -524,6 +545,25 @@ private BLangExpression createClosureForDefaultValue(String closureName, String
...
+        returnStmt.expr = varNode.expr;
+        BLangLambdaFunction lambdaFunction = createLambdaFunction(function);
+        lambdaFunction.capturedClosureEnv = env.createClone();
+        BInvokableSymbol varSymbol = createSimpleVariable(function, lambdaFunction);
+        env.enclPkg.symbol.scope.define(function.symbol.name, function.symbol);
+        env.enclPkg.functions.add(function);
+        env.enclPkg.topLevelNodes.add(function);
+        symbol.defaultValues.put(paramName, varSymbol);
+        rewrite(lambdaFunction, lambdaFunction.capturedClosureEnv);
         return returnStmt.expr;
```
MaryamZi, line 567: Isn't this basically the original expression? Why do we need to return it?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
```diff
@@ -1526,14 +1569,16 @@ private <E extends BLangStatement> List<E> rewriteStmt(List<E> nodeList, SymbolE
         Queue<BLangSimpleVariableDef> previousQueue = this.queue;
         this.queue = new LinkedList<>();
         int size = nodeList.size();
+        rewrite(typeNodeOfRecordField, env);
```
MaryamZi, line 1572: Why do we need to add `typeNodeOfRecordField` as an instance variable and visit it like this instead of doing it in-line?

MaryamZi (changes requested):
Please add more tests including those for anonymous records, both at module-level and within functions.

MaryamZi:
> Please add tests including
> 
> * tests for [Overriding a field included with record-type-inclusion doesn't remove the default value closure for that particular field being overriden #32012](https://github.com/ballerina-platform/ballerina-lang/issues/32012)
> * tests to make sure the default value closure isn't called when the user specifies a value for the field
>   e.g.,
> 
> ```ballerina
> import ballerina/io;
> 
> type Foo record {|
>     int x = fn();
> |};
> 
> isolated function fn() returns int {
>     io:println("called!");
>     return 10;
> }
> 
> public function main() {
>     Foo _ = {x: 1};
> }
> ```
> 
> Without this change this prints "called", but it shouldn't with the change.

How are you testing things like these? There don't seem to be tests at the moment.

You'll probably have to depend on the updating of a module-level variable.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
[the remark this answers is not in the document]
chiranSachintha: Changed as `ClosureGenerator`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: to keep closure variables(variable references used as default values).

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Yes. when visiting the generated closures(lambda functions).

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5838,10 +5740,72 @@ public void visit(BLangGroupExpr groupExpr) {
     @Override
     public void visit(BLangRecordLiteral recordLiteral) {
         List<RecordLiteralNode.RecordField> fields =  recordLiteral.fields;
+        updateFieldsOfRecordLiteral(recordLiteral, fields);
         fields.sort((v1, v2) -> Boolean.compare(isComputedKey(v1), isComputedKey(v2)));
         result = rewriteExpr(rewriteMappingConstructor(recordLiteral));
     }
 
+    public void visit(BFunctionPointerInvocation functionPointerInvocation) {
+        result = functionPointerInvocation;
```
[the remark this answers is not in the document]
chiranSachintha: args are desugared when visiting the `blanginvocation`. After that, we create the `functionPointerInvocation`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Yes. This can be `map type` also.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
chiranSachintha: Here, no need to generate closures for default values of fields. Those are already generated for to mutable type and we set default values from the mutable type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
chiranSachintha: removed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
chiranSachintha: Removed

tests/jballerina-unit-test/src/test/resources/test-src/record/open_record.bal
chiranSachintha: Update the testcase

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Rename as `generateFieldsForUserUnspecified`

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: I will send a separate PR for this. Need to do some changes in `codegen` also.

MaryamZi:
The behaviour changes for something like

```ballerina
import ballerina/io;

type Foo record {
    anydata i = 100;
};

type Bar record {
    anydata i;
};

public function main() {
    record {} r = {"i": "hello"};
    Foo f = {...r};
    // Previously `{"i":"hello"}`, now `{"i":100}`.
    io:println(f);

    // Compilation error for `r`: 
    // invalid usage of mapping constructor expression: spread field 'r' may have already specified keys
    // Bar _ = {i: 1, ...r}; 
}
```

`Foo f = {...r}` should probably result in an error or defaults should be used first before spreading out the spread fields. Can you recheck in the spec and raise a spec issue if necessary?

MaryamZi:
Also need tests like

```ballerina
type Foo record {
    anydata i = fn();
};

isolated function fn() returns int {
    // test that this is called
    return 100;
}

public function main() {
    string s = "i";
    // test that the value of field `i` is `"hello"`
    Foo f = {[s]: "hello"};
}
```

MaryamZi:
Please fix commit messages that start with a lower-case letter to start with an upper-case letter instead.

tests/jballerina-unit-test/src/test/resources/test-src/record/closed_record_type_inclusion.bal
```diff
@@ -286,9 +286,21 @@ type SimpleConstNegateExpr record {|
...
+
+type R1 record {
+    *R0;
+    string f;
+};
+
+
 function testTypeInclusionWithFiniteField() {
     SimpleConstNegateExpr expr = {};
+    R1 r1 = { f : "hello"};
```
MaryamZi, line 301:
```suggestion
    R1 r1 = {f : "hello"};
```

MaryamZi, line 301: Why are we adding this to `testTypeInclusionWithFiniteField` instead of introducing a new test?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5855,10 +5756,80 @@ public void visit(BLangGroupExpr groupExpr) {
...
+        for (Map.Entry<String, BInvokableSymbol> entry : defaultValues.entrySet()) {
+            String fieldName = entry.getKey();
+            if (fieldNames.contains(fieldName)) {
+                continue;
+            }
+            fieldNames.add(fieldName);
+            BInvokableSymbol invokableSymbol = entry.getValue();
+            BLangExpression expression = getInvocation(invokableSymbol);
+            if (isReadonly && !Symbols.isFlagOn(invokableSymbol.retType.flags, Flags.READONLY)) {
+                expression = visitCloneReadonly(expression, invokableSymbol.retType);
```
MaryamZi, line 5797: Consider

```ballerina
type Foo record {|
    any[] x = [1, new stream<int>()];
|};

public function main() {
    Foo & readonly _ = {};
}
```

This results in a panic at runtime atm.

```cmd
error: java.lang.UnsupportedOperationException
        at ballerina.lang.value.0:cloneReadOnly(value.bal:58)
           y:main(y.bal:6)
```

This isn't correct?

MaryamZi: In https://github.com/ballerina-platform/ballerina-spec/issues/544#issuecomment-1064777468, James has said

> Then I guess we say that in the readonly case if the type of the default expression is not Cloneable then there is no default. 

Please follow up on the issue if required.

Btw, if we are to consider this scenario as one with no default, we need to have this default value information at semantic analysis.

tests/jballerina-unit-test/src/test/resources/test-src/record/closed_record_type_inclusion.bal
chiranSachintha: Move to separate test

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/record/OpenRecordTest.java
chiranSachintha: Fixed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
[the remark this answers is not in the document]
chiranSachintha: fixed the first scenario. Second scenario still needs confirmation. I will create a specification issue for it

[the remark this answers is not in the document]
chiranSachintha: We need to allow the second scenario also. https://github.com/ballerina-platform/ballerina-spec/issues/1240

chiranSachintha:
can review the changes related to this PR. According to the spec clarification, we need to allow [this](https://github.com/ballerina-platform/ballerina-spec/issues/1240) kind of scenario. Currently, working on it

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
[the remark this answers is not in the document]
MaryamZi: Isn't `xFn` still called? We can start by avoiding that because we know for sure it will be available?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
MaryamZi: How is it different for the rest type vs other field types then? Wouldn't those of the rest type also be generated?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
MaryamZi: ?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: What is the plan for this?

MaryamZi:
@chiranSachintha, can you please reply to the individual comments on tests pointing to where each test has been added?

tests/jballerina-unit-test/src/test/resources/test-src/record/closed_record_type_inclusion.bal
chiranSachintha: https://github.com/ballerina-platform/ballerina-lang/blob/bd04cbf4262fc1698cc8af2b248711df6c44ead6/tests/jballerina-unit-test/src/test/resources/test-src/record/open_record.bal#L688

chiranSachintha:
> Please add tests including
> 
>     * tests for [Overriding a field included with record-type-inclusion doesn't remove the default value closure for that particular field being overriden #32012](https://github.com/ballerina-platform/ballerina-lang/issues/32012)
> 
>     * tests to make sure the default value closure isn't called when the user specifies a value for the field
>       e.g.,
> 
> 
> ```ballerina
> import ballerina/io;
> 
> type Foo record {|
>     int x = fn();
> |};
> 
> isolated function fn() returns int {
>     io:println("called!");
>     return 10;
> }
> 
> public function main() {
>     Foo _ = {x: 1};
> }
> ```
> 
> Without this change this prints "called", but it shouldn't with the change.

Added test https://github.com/ballerina-platform/ballerina-lang/blob/bd04cbf4262fc1698cc8af2b248711df6c44ead6/tests/jballerina-unit-test/src/test/resources/test-src/record/open_record.bal#L669

```ballerina
import ballerina/io;

type Foo record {|
    int x = fn();
|};

isolated function fn() returns int {
    io:println("called!");
    return 10;
}

public function main() {
    Foo _ = {x: 1};
}
```
With the current changes, the above program does not print "called!".

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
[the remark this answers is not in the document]
chiranSachintha: Fixed. Added tests with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/d7beeef38dc7477c3a1460538abf44aaf07ccba8

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
chiranSachintha: Updated the logic.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
chiranSachintha:
```ballerina
function testScopingRules() {

    record {
        int a;
        record {
            int p = 10;
        } c = {};
    } f2 = {a: 20};

}
```
For the above scenario, the closure for P should be generated within the closure of C.

```ballerina
function desugaredTestScopingRules() {
    function () returns  record {int p = 10;} c = function() returns  record {int p = 10;} {
        function () returns int p = function () returns int {
            return 10;
        };
        return {p: p()};
    };

    record {
        int a;
        record {
            int p = 10;
        } c = {};
    } f2 = {a: 20};
}

```
To achieve this, we need to generate the closure for P while visiting the closure of C. To implement this logic, I have visit  the type node of record literal separately.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Looking into this

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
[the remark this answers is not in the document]
chiranSachintha:
```ballerina
import ballerina/io;

type Foo record {|
    int x = xFn();
    int y = yFn();
|};

// generate closures

function () returns int foo_x = function () returns int {
    return xFn();
};

function () returns int foo_y = function () returns int {
    return yFn();
};


isolated function xFn() returns int {
    io:println("xFn!");
    return 10;
}

isolated function yFn() returns int {
    io:println("yFn!");
    return 20;
}

public function main() {
    record {| int x; int y?; |} r = {x: 123};
    //desugared code
    int field_y;
    if (r.y is ()) {
        field_y = foo_y();
    } else {
        field_y = 0;
    }
    Foo f = {...r, y: field_y};
    io:println(f);
}
```
As a solution for the second example, I attempted to desugar it as follows: 
I created an if statement to check whether a value for the "y" field is given or not. If a value is given, I use that value for the "y" field. The issue arises when we provide a value for "y" in the record literal, for instance:
```
record {| int x; int y?; |} r = {x: 123, y: 10};
```
In this case, the desugaring would be as follows:
```
Foo f = {...r, y: field_y};
```
In this case, the garbage value we set for "field_y" will be used as the value for "y".

MaryamZi:
@chiranSachintha, there seem to be issues once synced with master. Can you please update the PR?

chiranSachintha:
> @chiranSachintha, there seem to be issues once synced with master. Can you please update the PR?

Updated.

tests/jballerina-unit-test/src/test/resources/test-src/record/closed_record_type_inclusion.bal
MaryamZi: Was this added? Please reply to these comments with relevant commits and links to tests.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
MaryamZi: Please create an issue and mention the issue here.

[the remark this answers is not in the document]
MaryamZi: Not changed?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
MaryamZi: Not sure how this is updated?

Also why do we need to do this again? We wrap this instead at value creation, right?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
MaryamZi: Are your changes pushed?

MaryamZi (commented):
The following doesn't work as expected. We may need some runtime logic similar to desugaring with `cloneReadOnly`. Is this part of the change the runtime team is working on?

```ballerina
import ballerina/io;

type Foo record {|
    int[] x = [1, 2];
|};

    map<json> m = {};
    Foo & readonly n = check m.cloneWithType();
    io:println(n.x.toBalString()); // ()
    io:println(n.x is int[] & readonly); // false
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: Link for tests?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
MaryamZi: No, right? How do we call p's closure only, if required?

MaryamZi: The following crashes at the moment. 

```ballerina
public function main() {
    record {
        record {
            int c = 10;
        } b = {};
    } _ = {b: {}};
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: https://github.com/ballerina-platform/ballerina-spec/issues/544#issuecomment-1598057258

dulajdilshan:
```ballerina
public function main() {
    int a;
    int b;
    [{a, b}] = [{a: 10, b: 11}];
}
```

Need to test this scenario too (currently this is not working till https://github.com/ballerina-platform/ballerina-lang/issues/39324 is merged )

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
chiranSachintha: Yes, you mean the `bLangExpression variable`, right? I have updated the logic for that.
https://github.com/ballerina-platform/ballerina-lang/blob/3abc989bdd82a9a7d91a41d4a446c5efa1e9ed5c/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java#L525

tests/jballerina-unit-test/src/test/resources/test-src/record/closed_record_type_inclusion.bal
chiranSachintha: Yes.
https://github.com/ballerina-platform/ballerina-lang/blame/3abc989bdd82a9a7d91a41d4a446c5efa1e9ed5c/tests/jballerina-unit-test/src/test/resources/test-src/record/open_record.bal#L696

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
chiranSachintha: https://github.com/ballerina-platform/ballerina-lang/issues/40884

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
chiranSachintha: Yes. I have removed that logic.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
```diff
@@ -541,22 +554,71 @@ private BSymbol getOwner(SymbolEnv symbolEnv) {
         return symbolEnv.enclPkg.symbol;
     }
 
-    private BLangExpression createClosureForDefaultValue(String closureName, String paramName,
-                                                         BLangSimpleVariable varNode) {
+    BLangExpression addConversionExprIfRequired(BLangExpression expr, BType lhsType) {
```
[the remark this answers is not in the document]
chiranSachintha: Yes, but that method is in the desugar phase. The reason I should not use it here is that we are planning to move this logic before the desugaring phase. Once we make that change, we can remove this logic.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Added a test with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/fb9f549c2b690169aca53dd30eef490c30658c6e

What do you mean by `someone changes the order`?

MaryamZi: Can't quite recall now, but I think it was to do with if someone populated values from the included records first and then checked and used values from the including record.

MaryamZi:
Have we checked https://github.com/ballerina-platform/ballerina-lang/issues/33773 with this?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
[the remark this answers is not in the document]
chiranSachintha: Fixed this issue with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/91bdac558dd0440e4b548f5b6216bdbf4913260a

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/GenerateClosuresForDefaultValues.java
chiranSachintha: Fixed. Added tests also

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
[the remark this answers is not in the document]
MaryamZi: Yeah, I meant we can move this to where we have common methods.

MaryamZi:
> The following doesn't work as expected. We may need some runtime logic similar to desugaring with `cloneReadOnly`. Is this part of the change the runtime team is working on?
> 
> ```ballerina
> import ballerina/io;
> 
> type Foo record {|
>     int[] x = [1, 2];
> |};
> 
> public function main() returns error? {
>     map<json> m = {};
>     Foo & readonly n = check m.cloneWithType();
>     io:println(n.x.toBalString()); // ()
>     io:println(n.x is int[] & readonly); // false
> }
> ```

Still doesn't work? Checked with https://github.com/ballerina-platform/ballerina-lang/pull/40579.

MaryamZi:
> Have we checked #33773 with this?

Tests for this?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/MapValueImpl.java
```diff
@@ -281,17 +281,22 @@ public V put(K key, V value) {
...
+                    continue;
+                }
                 populateInitialValue((K) keyValueEntry.key, (V) keyValueEntry.value);
                 continue;
             }
 
             MapValueImpl<K, V> values =
                     (MapValueImpl<K, V>) ((MappingInitialValueEntry.SpreadFieldEntry) initialValue).values;
             for (Map.Entry<K, V> entry : values.entrySet()) {
+                definedFields.add(entry.getKey().toString());
```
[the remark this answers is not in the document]
chiranSachintha: I have fixed this in this PR https://github.com/ballerina-platform/ballerina-lang/pull/40579/commits/edf5438a9e486413b5972b9a25824113c3462c93 as we need runtime changes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/bddee51f61fd88f2147c7c29a072ad5bcd98bc3c

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -2970,7 +2906,7 @@ private void createVarRefAssignmentStmts(BLangErrorVarRef parentErrorVarRef, BLa
             cause.varRef = parentErrorVarRef.cause;
         }
 
-        // When no detail nor rest detail are to be destructured, we don't need to generate the detail invocation.
+        // When no detail nor rest detail are to be destructured, we don't need to generate the detail getInvocation.
```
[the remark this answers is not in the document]
chiranSachintha: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/60ef6112f7522849eed7b373a3ac84c7524abd7b

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
```diff
@@ -541,22 +555,75 @@ private BSymbol getOwner(SymbolEnv symbolEnv) {
...
+    private void generateClosureForDefaultValues(String closureName, String paramName, BLangSimpleVariable varNode) {
         BSymbol owner = getOwner(env);
-        BInvokableTypeSymbol symbol = (BInvokableTypeSymbol) env.node.getBType().tsymbol;
         BLangFunction function = createFunction(closureName, varNode.pos, owner.pkgID, owner, varNode.getBType());
-        updateFunctionParams(function, symbol.params, paramName);
         BLangReturn returnStmt = ASTBuilderUtil.createReturnStmt(function.pos, (BLangBlockFunctionBody) function.body);
-        returnStmt.expr = varNode.expr;
+        if (varNode.expr.getKind() == NodeKind.RECORD_LITERAL_EXPR) {
+            BLangRecordLiteral recordLiteral = (BLangRecordLiteral) varNode.expr;
+            desugar.generateFieldsForUserUnspecifiedRecordFields(recordLiteral, recordLiteral.fields);
```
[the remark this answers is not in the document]
chiranSachintha: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/24933d2a562f4f107c96f8c9570fd2832eb8e755

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
```diff
@@ -200,10 +203,13 @@ public class ClosureGenerator extends BLangNodeVisitor {
     private Queue<BLangSimpleVariableDef> queue;
     private Queue<BLangSimpleVariableDef> annotationClosureReferences;
     private SymbolTable symTable;
+    private BLangType typeNodeOfRecordField;
```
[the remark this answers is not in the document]
chiranSachintha: Fixed https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/60ef6112f7522849eed7b373a3ac84c7524abd7b

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
```diff
@@ -191,7 +194,7 @@
 import static org.ballerinalang.model.symbols.SymbolOrigin.VIRTUAL;
 
 /**
- * Parameter desugar for create closures for default values.
+ * ClosureGenerator for create closures for default values.
```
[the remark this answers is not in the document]
chiranSachintha: Fixed https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/60ef6112f7522849eed7b373a3ac84c7524abd7b

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -484,7 +484,9 @@ private void defineFunction(DataInputStream dataInStream) throws IOException {
                     BStructureTypeSymbol structureTypeSymbol = (BStructureTypeSymbol) attachedType.tsymbol;
                     if (Names.USER_DEFINED_INIT_SUFFIX.value.equals(funcName)
                             || funcName.equals(Names.INIT_FUNCTION_SUFFIX.value)) {
-                        structureTypeSymbol.initializerFunc = attachedFunc;
+                        if (structureTypeSymbol.getKind() == SymbolKind.OBJECT) {
+                        ((BObjectTypeSymbol) structureTypeSymbol).initializerFunc = attachedFunc;
+                    }
```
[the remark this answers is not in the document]
chiranSachintha: Fixed https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/60ef6112f7522849eed7b373a3ac84c7524abd7b

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Removed the comment

chiranSachintha:
> > Have we checked #33773 with this?
> 
> Tests for this?

Added with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/9aa5efe75205dba37fe5952cd9ae63a9584bc60b

chiranSachintha:
> Have we checked #33773 with this?

Fixed with this PR and add tests with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/9aa5efe75205dba37fe5952cd9ae63a9584bc60b

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/MapValueImpl.java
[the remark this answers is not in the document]
MaryamZi: Which PR?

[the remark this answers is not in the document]
chiranSachintha: https://github.com/ballerina-platform/ballerina-lang/pull/40579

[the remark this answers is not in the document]
MaryamZi: Does it work yet? Still seeing "getB".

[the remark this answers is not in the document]
chiranSachintha: The issue occurred when syncing the PRs. I have resolved it.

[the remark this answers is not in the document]
MaryamZi: Are there test for this?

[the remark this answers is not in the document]
MaryamZi: Please check the order for missing fields. Seems like the value changes for

```ballerina
import ballerina/io;

type Foo record {|
    int a;
    int b?;
|};

type Bar record {|
    int a;
    int b = getValue();
    int c = getValue();
|};

isolated int v = 100;

isolated function getValue() returns int {
    lock {
        v += 1;
        return v;
    }
}

public function main() {
    Foo f = {a: 1};
    Bar b = {...f};
    io:println(b);

    Bar bd = {a: 1};
    io:println(bd);
}
```

2201.7.0

```bash
{"a":1,"b":101,"c":102}
{"a":1,"b":103,"c":104}
```

With this change

```bash
{"a":1,"b":102,"c":101}
{"a":1,"b":103,"c":104}
```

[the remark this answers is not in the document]
chiranSachintha: Add tests with https://github.com/ballerina-platform/ballerina-lang/pull/40579/commits/198e9eb9decde554cbdf8fa0c2e8bd7520bedaf5

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -482,9 +482,11 @@ private void defineFunction(DataInputStream dataInStream) throws IOException {
                             new BAttachedFunction(names.fromString(funcName), invokableSymbol, funcType,
                                     symTable.builtinPos);
                     BStructureTypeSymbol structureTypeSymbol = (BStructureTypeSymbol) attachedType.tsymbol;
-                    if (Names.USER_DEFINED_INIT_SUFFIX.value.equals(funcName)
-                            || funcName.equals(Names.INIT_FUNCTION_SUFFIX.value)) {
-                        structureTypeSymbol.initializerFunc = attachedFunc;
+                    if (Names.USER_DEFINED_INIT_SUFFIX.value.equals(funcName) ||
+                            funcName.equals(Names.INIT_FUNCTION_SUFFIX.value)) {
+                        if (structureTypeSymbol.getKind() == SymbolKind.OBJECT) {
```
MaryamZi, line 487: Why do we need to check this? Won't L485-L486 be true only for objects now?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
```diff
@@ -542,22 +554,25 @@ private BSymbol getOwner(SymbolEnv symbolEnv) {
...
-    private BLangExpression createClosureForDefaultValue(String closureName, String paramName,
-                                                         BLangSimpleVariable varNode) {
+    private void generateClosureForDefaultValues(String closureName, String paramName, BLangSimpleVariable varNode) {
         BSymbol owner = getOwner(env);
-        BInvokableTypeSymbol symbol = (BInvokableTypeSymbol) env.node.getBType().tsymbol;
         BLangFunction function = createFunction(closureName, varNode.pos, owner.pkgID, owner, varNode.getBType());
-        updateFunctionParams(function, symbol.params, paramName);
         BLangReturn returnStmt = ASTBuilderUtil.createReturnStmt(function.pos, (BLangBlockFunctionBody) function.body);
-        returnStmt.expr = varNode.expr;
+        returnStmt.expr = desugar.addConversionExprIfRequired(varNode.expr, function.returnTypeNode.getBType());
```
MaryamZi, line 561: Move `addConversionExprIfRequired` to Types instead and avoid the cycle between desugar and this.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
```diff
@@ -542,22 +554,25 @@ private BSymbol getOwner(SymbolEnv symbolEnv) {
...
         BLangLambdaFunction lambdaFunction = createLambdaFunction(function);
         lambdaFunction.capturedClosureEnv = env.createClone();
         BInvokableSymbol varSymbol = createSimpleVariable(function, lambdaFunction, false);
+        BTypeSymbol symbol = env.node.getBType().tsymbol;
+        if (symbol.getKind() == SymbolKind.INVOKABLE_TYPE) {
+            updateFunctionParams(function, ((BInvokableTypeSymbol) symbol).params, paramName);
+            ((BInvokableTypeSymbol) symbol).defaultValues.put(Utils.unescapeBallerina(paramName), varSymbol);
```
MaryamZi, line 568: Can extract `(BInvokableTypeSymbol) symbol` out to a variable.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
```diff
@@ -688,6 +702,9 @@ private String generateName(String name, BLangNode parent) {
             case SERVICE:
                 name = ((BLangService) parent).name.getValue() + UNDERSCORE + name;
                 return generateName(name, parent.parent);
+            case RECORD_TYPE:
+                name = ((BLangRecordTypeNode) parent).symbol.name.getValue() + UNDERSCORE + name;
```
MaryamZi, line 706: Can pass directly?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
```diff
@@ -1639,7 +1658,10 @@ private <E extends BLangStatement> List<E> rewriteStmt(List<E> nodeList, SymbolE
             E node = rewrite(nodeList.remove(0), env);
             Iterator<BLangSimpleVariableDef> iterator = queue.iterator();
             while (iterator.hasNext()) {
-                nodeList.add(rewrite((E) queue.poll(), env));
+                BLangSimpleVariableDef bLangSimpleVariableDef = queue.poll();
+                nodeList.add(rewrite((E) bLangSimpleVariableDef, env));
+                BSymbol varSymbol = bLangSimpleVariableDef.var.symbol;
+                env.scope.define(varSymbol.name, varSymbol);
```
MaryamZi, line 1664: Why do we need to do this now? Due to the cloned env?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5927,10 +5837,149 @@ public void visit(BLangGroupExpr groupExpr) {
...
+        if (env.enclEnv.node != null && env.enclEnv.node.getKind() == NodeKind.ARROW_EXPR) {
+            return env.enclEnv;
+        }
+
+        if (env.enclEnv.node != null && (env.enclEnv.node.getKind() == NodeKind.ON_FAIL)) {
+            return env.enclEnv;
+        }
+
+        if (env.enclInvokable != null && env.enclInvokable == encInvokable) {
+            return findEnclosingInvokableEnv(env.enclEnv, encInvokable);
```
MaryamZi, line 5855: Why do we need these special cases?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5927,10 +5837,149 @@ public void visit(BLangGroupExpr groupExpr) {
...
+        }
+    }
+
+    private void generateFieldsForUserUnspecifiedRecordFields(List<RecordLiteralNode.RecordField> fields,
+                                                              List<String> fieldNames,
+                                                              Map<String, BInvokableSymbol> defaultValues,
+                                                              Location pos, boolean isReadonly) {
+        for (Map.Entry<String, BInvokableSymbol> entry : defaultValues.entrySet()) {
+            String fieldName = entry.getKey();
+            if (!fieldNames.contains(fieldName)) {
```
MaryamZi, line 5929: Check inverse and continue early.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2668,24 +2665,80 @@ private boolean validateSpecifiedFields(BLangRecordLiteral mappingConstructor, B
         return isFieldsValid;
     }
 
+    private BRecordType getRecordTypeInIntersection(BIntersectionType intersectionType) {
+        for (BType constituentType : intersectionType.getConstituentTypes()) {
+            BType referredType = Types.getReferredType(constituentType);
+            if (referredType.tag == TypeTags.RECORD) {
+                return (BRecordType) referredType;
```
MaryamZi, line 2672: This will not work with generic intersections btw, please make a note.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
chiranSachintha: We are defining variable definitions with function pointers and adding them to the block scope. Within that scope, we need to define those variable definitions.
consider the below example
```ballerina
function value() {
    record {|int a = 10;|} k = {};
}
```
we desugar it in the following manner
```ballerina
function value() {
    function() returns int k_a = function () returns int {
        return 10;
    };
    record {|int a = 10;|} k = {};
}
```
So we need to define variable `k_a` within the body scope of the `value` function.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: Why specifically arrow and on-fail checks?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
chiranSachintha: Yes. We can remove it. Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/f846176e6ff38c3e1366f21da4cfa05ea8a4fbdc

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
chiranSachintha: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/01c17a5215a1a8d4c7a242762dfc26294ebb7dec

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Fixed with https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/f846176e6ff38c3e1366f21da4cfa05ea8a4fbdc

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureGenerator.java
chiranSachintha: Updated https://github.com/ballerina-platform/ballerina-lang/pull/37184/commits/f846176e6ff38c3e1366f21da4cfa05ea8a4fbdc

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: Doesn't the closure desugar already handle these?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -5929,10 +5839,149 @@ public void visit(BLangGroupExpr groupExpr) {
     @Override
     public void visit(BLangRecordLiteral recordLiteral) {
         List<RecordLiteralNode.RecordField> fields =  recordLiteral.fields;
+        generateFieldsForUserUnspecifiedRecordFields(recordLiteral, fields);
         fields.sort((v1, v2) -> Boolean.compare(isComputedKey(v1), isComputedKey(v2)));
         result = rewriteExpr(rewriteMappingConstructor(recordLiteral));
     }
 
+    private SymbolEnv findEnclosingInvokableEnv(SymbolEnv env, BLangInvokableNode encInvokable) {
```
MaryamZi, line 5847: Isn't this a duplicate of what's in the closure generator?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: No, When comes to the closure desugar phase, need to identify and set them as closure variables. (Defining a variable as a closure variable is not done during the closure desugar phase.)

Since we are generating the record fields during the desugar phase, we need to indicate whether the function pointers associated with those invocations are closure variables or not.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: Yes. Here we are generating the record fields during the desugar phase(invocation of function pointers), we need to indicate whether the function pointers associated with those invocations are closure variables or not.

chiranSachintha:
These changes are added to the https://github.com/ballerina-platform/ballerina-lang/pull/40579 PR

PR #37207, opened 2022-08-02
[2201.2.0-stage] Add isBlank check to ls simulator skipped types

IMS94:
## Purpose
$subject

### DIFF
--- a/tests/language-server-simulator/src/main/java/org/ballerinalang/langserver/simulator/generators/Generators.java
+++ b/tests/language-server-simulator/src/main/java/org/ballerinalang/langserver/simulator/generators/Generators.java
@@ -42,7 +42,7 @@ public class Generators {
         // Get skipped generators
         String property = System.getProperty(PROP_SKIPPED_GENERATORS, "");
         Set<Type> skippedGenerators = Stream.of(property.split(","))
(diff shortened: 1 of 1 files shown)


tests/language-server-simulator/src/main/java/org/ballerinalang/langserver/simulator/generators/Generators.java
```diff
@@ -42,7 +42,7 @@ private Generators() {
         // Get skipped generators
         String property = System.getProperty(PROP_SKIPPED_GENERATORS, "");
         Set<Type> skippedGenerators = Stream.of(property.split(","))
-                .map(String::trim)
+                .filter(String::isBlank)
```
malinthar, line 45: We should add the negation of this condition ryt?

malinthar (changes requested):

tests/language-server-simulator/src/main/java/org/ballerinalang/langserver/simulator/generators/Generators.java
IMS94: Fixed

malinthar (approved):
LGTM!

PR #37211, opened 2022-08-02
Fix updating return type for func with check expr

### LINKED ISSUE #37153: Not providing union type when there is a check expression when function has an error return type
**Description:**
In below scenario `testReturnType` function has a `error?` return type when we apply change return type CA it gives `string` type as the return type. But it should provide the `string|error?` union type.

https://user-images.githubusercontent.com/46857198/181437521-29ff433c-0ee1-49aa-92ce-83e643254e1f.mov

Area/CodeAction, Team/LanguageServer

LakshanWeerasinghe:
## Purpose
$subject

Fixes #37153

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/changetype/FixReturnTypeCodeAction.java
```diff
@@ -216,4 +225,24 @@ public static List<Set<String>> getPossibleCombinations(List<List<String>> combi
         }
         return typeList;
     }
+
+    /* A visitor to find check expression exist inside a Node */
```
IMS94, line 229: We should use a javadoc comment here:
```
/**
* <comment>
*/
```

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/changetype/FixReturnTypeCodeAction.java
```diff
@@ -216,4 +225,24 @@ public static List<Set<String>> getPossibleCombinations(List<List<String>> combi
...
+    static class CheckExprNodeFinder extends NodeVisitor {
+
+        private CheckExpressionNode checkExpressionNode = null;
+
+        @Override
+        public void visit(FunctionDefinitionNode functionDefinitionNode) {
```
IMS94, line 235: There's no need to override this method

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/changetype/FixReturnTypeCodeAction.java
```diff
@@ -142,6 +147,10 @@ public List<CodeAction> getCodeActions(Diagnostic diagnostic,
                 }
             }
 
+            if (checkExprNodeFinder.containCheckExprNode()) {
```
IMS94, line 150: Why do we need to check if the function body has `check` expressions? Instead, we should simply check if the function return type has an `error` member. If yes, we have to union the new return type with the error. WDYT?

IMS94 (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/changetype/FixReturnTypeCodeAction.java
LakshanWeerasinghe: The issue here is we need to add `error` return type for possible return types if there is a `check` expression node in the function body. There can be cases where user specified `error` return and there are no any `error` returns in the function body.

IMS94:
Closing in favor of #37308

PR #37227, opened 2022-08-02
Fix loop running only once despite `trap` expression

### LINKED ISSUE #37174: Loop runs only once despite `trap` expression
**Description:**

$title

**Steps to reproduce:**
```bal
import ballerina/io;

public function main() {
    foo();
}

function foo() {
    foreach int i in 0...2 {
        error? unionResult = trap func();
        // io:println(i,"\n");
    }
}

function func() {
    io:println("func called");
    panic error("err");
}
```
will give
```
Running executable

func called
```

When I uncomment the `io:println(i,"\n");` it works fine giving the output.
```
Running executable

func called
0

func called
1

func called
2
```

**Affected Versions:**
Ballerina 2201.1.0 (Swan Lake)

Nadeeshan96:
$title
Fixes #37174

The problem occurred due to removing the targetBB of a trap error entry by the BIR basic block optimiser. So changed it to not remove the targetBBs.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/optimizer/BIRBasicBlockOptimizer.java
```diff
@@ -116,13 +116,18 @@ private void updatePredecessorMap(BIRBasicBlock removable) {
...
+        }
+
         Set<BIRBasicBlock> removableBasicBlocks = new HashSet<>();
         for (int i = birFunction.basicBlocks.size() - 1; i >= 0; --i) {
             BIRBasicBlock basicBlock = birFunction.basicBlocks.get(i);
 
             // Remove basic blocks with unnecessary jump statement
             if (basicBlock.terminator instanceof BIRTerminator.GOTO && basicBlock.instructions.isEmpty()
-                    && basicBlock.terminator.pos == null) {
+                    && basicBlock.terminator.pos == null && !errorTableTargetBBs.contains(basicBlock)) {
```
gabilang, line 130: Since we are not going to remove `trap` and `end` BBs, can't we simply check whether the `errorTable` contains the basicblock which we are consider for removing?

Nadeeshan96: Do you mean there is a possibility where the previous conditions will allow the removal of `trap` and `end` BBs, and we need to also include them in the `errorTableTargetBBs` set?
If not, we can insert only the `target` BBs like this, because it will be efficient.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/trap/TrapOperatorTest.java
```diff
@@ -32,25 +34,42 @@
  */
 public class TrapOperatorTest {
 
+    CompileResult result;
+    CompileResult resultNegative;
+
+    @BeforeClass
+    public void setup() {
+        result = BCompileUtil.compile("test-src/expressions/trap/trap-expr.bal");
+        resultNegative = BCompileUtil.compile("test-src/expressions/trap/trap_negative.bal");
```
HindujaB, line 43: Since the `trap_negative.bal` file is only used in one test case, shall we move it to the test itself? 
So that we can identify if there is an error occured at compilation separately.

Nadeeshan96:
Closing this because two PRs will be sent to `2201.2.x` branch and `master` branch.

Issue #36527, opened 2022-06-13
LS simulator workflow fails with OOM errors on 2201.1.x branch


IMS94:
**Description:**
Recently the LS simulator workflow failed frequently with OOM errors on the 2201.1.x branch. Workflow runs can be found at:
https://github.com/ballerina-platform/ballerina-lang/actions/workflows/language_server_simulator.yml

Some of the failed runs:
1. https://github.com/ballerina-platform/ballerina-lang/actions/runs/2488244655
2. https://github.com/ballerina-platform/ballerina-lang/actions/runs/2479894651
3. https://github.com/ballerina-platform/ballerina-lang/actions/runs/2468242304

**Steps to reproduce:**
Since the LS simulator adds random stuff to the ballerina documents, exact source code cannot be determined. The workflow runs contain the heapdump files generated. We can analyze them.

**Affected Versions:**
2201.1.1-SNAPSHOT

gimantha:
I tried to reproduce this locally as we cannot find the root of the OOM from the dumps. I increased the running time of the Simulator tests to 6 hours and monitored with VisualVM. I am not seeing any memory leaks for that run. Below are the results from VisualVM.

<img width="590" alt="image" src="https://user-images.githubusercontent.com/9362271/173983927-91668877-13bd-46a7-ad47-e3f255c919d5.png">

<img width="1175" alt="image" src="https://user-images.githubusercontent.com/9362271/173983981-d66442d9-70ff-483b-b55e-7dcbace7fb86.png">

My suspect is that, this is happening only for a certain path in simulation and we need to find that specific case. This can take time

IMS94:
Me and @malinthar tested a ballerina distribution built from 2201.1.x with nBallerina repo and couldn't observe any unusual behavior. Let's see if this gets reproduced to actual users.

malinthar:
I was able to reproduce the OOM locally in the nballerina project. Following is the recording of the steps to reproduce.
https://wso2.zoom.us/rec/share/eGbrlwWbUmyawrb4HbqUwYELBo9--96no8rmhreDesZT_T8bitQ1Gmht7t-sfvD7.qCYITMonVxtY1yUw

PR #37240, opened 2022-08-03
Fix OOM in LS

KavinduZoysa:
## Purpose

Fixes #https://github.com/ballerina-platform/ballerina-lang/issues/36527

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageCompilation.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageCompilation.java
@@ -198,7 +198,8 @@ public class PackageCompilation {
             for (ModuleContext moduleContext : packageResolution.topologicallySortedModuleList()) {
                 moduleContext.compile(compilerContext);
                 for (Diagnostic diagnostic : moduleContext.diagnostics()) {
-                    diagnostics.add(new PackageDiagnostic(diagnostic, moduleContext.descriptor(), moduleContext.project()));
+                    diagnostics.add(new PackageDiagnostic(diagnostic, moduleContext.descriptor(),
+                            moduleContext.project()));
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


gimantha (approved):

KavinduZoysa:
@azinneera, Until we come up with a proper solution as mention in https://github.com/ballerina-platform/ballerina-lang/issues/37242, we have to change several project API tests. Can you please check those?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageCompilation.java
```diff
@@ -194,10 +194,13 @@ private void compileModulesInternal() {
         // add dependency manifest diagnostics
         diagnostics.addAll(packageContext().dependencyManifest().diagnostics().allDiagnostics);
         // add compilation diagnostics
-        for (ModuleContext moduleContext : packageResolution.topologicallySortedModuleList()) {
-            moduleContext.compile(compilerContext);
-            for (Diagnostic diagnostic : moduleContext.diagnostics()) {
-                diagnostics.add(new PackageDiagnostic(diagnostic, moduleContext.descriptor(), moduleContext.project()));
+        if (!packageResolution.diagnosticResult().hasErrors()) {
```
azinneera, line 197: There can be different types of errors added in the package resolution phase. E.g. cyclic dependencies, incompatible dependency version conflicts, and with IDL imports, there can be errors related to the client generation as well. Are we going to stop compilation for all these scenarios? This will stop giving suggestions in LS

azinneera: if it is only the cyclic dependency case that we need to handle, then we can use `packageResolution.dependencyGraph().findCycles()` which will return a Set

KavinduZoysa: We can't do this because of https://github.com/ballerina-platform/ballerina-lang/issues/37257

gimantha (approved):
LGTM

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageCompilation.java
azinneera: Ack

azinneera (approved):

PR #37297, opened 2022-08-08
Add attachment-value symbol for annotations

### LINKED ISSUE #28856: [Semantic API] Add support for getting the annotation attachment values at source-only attach points
Currently when you get the annotations of a particular symbol, it only gives the symbol(s) of the annotation(s) attached to that construct. Need to improve this to be able to provide the value of the annotation at the attach points.

### LINKED ISSUE #37194: Accessing the annotation field values via AnnotationSymbol
**Description:**
```ballerina
import ballerina/http;
import ballerina/constraint;

@constraint:String {minLength: 5}
public type Address string;

public type Person record {
    @constraint:String {maxLength: 14}
    string name?;
    @constraint:Array {maxLength: 5, minLength: 2}
    string[] hobby?;
    @constraint:Int {maxValue: 5}
    int id;
    Address address?;
    @constraint:Float {maxValue: 100000}
    float salary?;
    @constraint:Number {minValue: 500000}
    decimal net?;
};

service /payloadV on new http:Listener(9090) {
    resource function post pet(@http:Payload Person body) returns error? {
        return;
    }
}
```
If there is a annotation in record field,  In openAPI we reuired to access, given annotation values for generating the openAPI specification. 
The current implementation is driven by accessing RecordFieldTypeSymbol in the given record field. It returns the details about recordfield and we pick annotation details via calling .getAnnotations() but it didn't content annotation current available field values.

In above example, we need to get the details in `name` record field's annotation field **maxLength**: **14**

**Describe your problem(s)**

1. For current openAPI tool, we tried to back track to get syntax node from typeSymbol location from calling `.findNode()` . But this approach I met with limitation when we are going to access type that defined in the separate model. Because our util we use only `sematicModel` and `SyntaxTree` that are not enough for getting separate module syntax node details.
2. Due to the limitation we have to pass the full project to the util instead of the sematicModel and Syntax tree, then it cause to some API changes to whole Ballerina to OpenAPI tool. Although we passed it there still we have do some kind of process finding dependency graph for tracking dependencies to access type define in the separate `package`.  It will cause to some kind of conflict to the current util since it may additional process to mapping. Complex can be delaying to generate openapi spec, resolving dependency issue, interminante issue etc. (This util has used in several modules http, choreo, vsCode. Therefore we required to safe method to solve this.) 

**Describe your solution(s)**
Considering the above issue and 1, 2 points , it would be great if we can have given `AnnotationSymbol` with extrafield that value with Optional regarding available record fields with their values. 

Related issue : https://github.com/ballerina-platform/openapi-tools/issues/1040

dulajdilshan:
## Purpose
$title

Fixes #37194
Fixes #28856

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


mohanvive:
@dulajdilshan can you check the test failures in the PR.

dulajdilshan:
Here, we have to do some refactoring of the approach. Had a chat with @mohanvive  and @pubudu91 separately. We have 2 approaches to do

Approach 1:
Screenshot from 2022-08-23 14-20-53
Here we provide the annotation-value information through the annotation symbol. Here, we don't alter any of the APIs results nor introduce new APIs to attached symbols.

Approach 2:
Screenshot from 2022-08-23 14-27-06
Here we create a new symbol as annotaion-attachment-symbol. We will be implementing a new API as `annotAttachments()` for attached symbols which returns the list of annotation attachments. In annotation-attachment, the `typeDescriptor()` will return the annotation-symbol.

We decided to go with Approach 2 but we are not going to change the behaviour for `symbol()` API.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaAnnotationAttachmentSymbol.java
```diff
@@ -0,0 +1,85 @@
...
+import io.ballerina.compiler.api.values.ConstantValue;
+import org.wso2.ballerinalang.compiler.semantics.model.symbols.BAnnotationAttachmentSymbol;
+import org.wso2.ballerinalang.compiler.util.CompilerContext;
+
+import java.util.Optional;
+
+/**
+ * Represents an Annotation attachment.
+ *
+ * @since 2201.2.0
```
DamithaSenevirathne, line 33: Shall we update the version?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/values/BallerinaConstantValue.java
```diff
@@ -0,0 +1,79 @@
...
+
+import io.ballerina.compiler.api.symbols.TypeSymbol;
+import io.ballerina.compiler.api.values.ConstantValue;
+
+import java.util.Objects;
+
+/**
+ *  Value of the constant.
+ *
+ *  @since 2201.2.0
```
DamithaSenevirathne, line 28: Shall we update the version?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/symbols/AnnotationAttachmentSymbol.java
```diff
@@ -0,0 +1,50 @@
...
+package io.ballerina.compiler.api.symbols;
+
+import io.ballerina.compiler.api.values.ConstantValue;
+
+import java.util.Optional;
+
+/**
+ * Represents an Annotation attachment.
+ *
+ * @since 2201.2.0
```
DamithaSenevirathne, line 26: Update the version here as well

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/values/ConstantValue.java
```diff
@@ -0,0 +1,31 @@
...
+ * under the License.
+ */
+package io.ballerina.compiler.api.values;
+
+import io.ballerina.compiler.api.symbols.TypeSymbol;
+
+/**
+ *  Represent the constant value.
+ *
+ *  @since 2201.2.0
```
DamithaSenevirathne, line 25: Update the verison

tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/symbols/ConstAnnotationAttachmentSymbolTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+import static io.ballerina.semantic.api.test.util.SemanticAPITestUtils.getDefaultModulesSemanticModel;
+import static io.ballerina.semantic.api.test.util.SemanticAPITestUtils.getDocumentForSingleSource;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+/**
+ * Test cases for constant annotation attachments.
+ *
+ * @since 2201.2.0
```
DamithaSenevirathne, line 49: Update the version here as well

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaObjectFieldSymbol.java
```diff
@@ -86,14 +87,30 @@ public List<AnnotationSymbol> annotations() {
 
         List<AnnotationSymbol> annots = new ArrayList<>();
         SymbolFactory symbolFactory = SymbolFactory.getInstance(this.context);
-        for (AnnotationAttachmentSymbol annot : bField.symbol.getAnnotations()) {
+        for (org.ballerinalang.model.symbols.AnnotationAttachmentSymbol annot : bField.symbol.getAnnotations()) {
```
DamithaSenevirathne, line 90: Any possibility of getting `CCE` here ?

dulajdilshan: No, here `getAnnotations()` returns symbols of `org.ballerinalang.model.symbols.AnnotationAttachmentSymbol`

DamithaSenevirathne (approved):

mohanvive (approved):

PR #37308, opened 2022-08-09
Fix updating return type for func with check expr

### LINKED ISSUE #37153: Not providing union type when there is a check expression when function has an error return type

LakshanWeerasinghe:
## Purpose
$subject

Fixes #37153

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/changetype/FixReturnTypeCodeAction.java
```diff
@@ -142,6 +147,10 @@ public List<CodeAction> getCodeActions(Diagnostic diagnostic,
                 }
             }
 
+            if (checkExprNodeFinder.containCheckExprNode()) {
```
IMS94, line 150: We can move the execution of the node visitor to the line before this line. No need to run it everytime (except in this else case)

IMS94 (changes requested):

IMS94 (approved):

PR #37312, opened 2022-08-09
Provide missing completions in mapping constructor context

### LINKED ISSUE #36339: Field completions are missing in mapping constructor expression in named arg and new expression
**Description:**
Some completions are missing in the master branch for the following scenarios.

**Steps to reproduce:**
> mapping_expr_ctx_source29.bal
```
type Application record {|
    string id;
    string name;
    int 'version;
|};

type ErrorData record {
    string errorCode;
    Application application;
};

type CustomError error<ErrorData>;

public function checkError() returns CustomError? {
    return error CustomError("Custom error", errorCode = "123", application = {<cur>});
}

```
> mapping_expr_ctx_source63.bal
```
type ClientConfig record {|
    string host;
    int port;
    string scheme;

    int timeout;
|};

public type Error distinct error;

client class Client {

    private ClientConfig config;

    public function init(ClientConfig config) returns error? {
        self.config = config;
    }

    remote function findByName(string name) returns json|Error {
        return {
            id: "123",
            name: "Test"
        };
    }
}

    Client cl = check new ({<cur>});
}
```

**Affected Versions:**
master branch

https://github.com/ballerina-platform/ballerina-lang/pull/35210

malinthar:
## Purpose
$subject by fixing new expression visiting logic in ContextTypeResolver class.

Fixes #36339 

## Samples
image

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ContextTypeResolver.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ContextTypeResolver.java
@@ -811,6 +811,18 @@ public class ContextTypeResolver extends NodeTransformer<Optional<TypeSymbol>> {
         Optional<MethodSymbol> methodSymbol = context.currentSemanticModel()
                 .flatMap(semanticModel -> semanticModel.typeOf(newExpressionNode))
                 .flatMap(typeSymbol -> Optional.of(CommonUtil.getRawType(typeSymbol)))
+                .map(typeSymbol -> {
+                    if (typeSymbol.typeKind() == TypeDescKind.UNION) {
+                        Optional<TypeSymbol> classType =
+                                ((UnionTypeSymbol) typeSymbol).memberTypeDescriptors().stream()
+                                        .map(CommonUtil::getRawType)
+                                        .filter(member -> member instanceof ClassSymbol).findFirst();
+                        if (classType.isPresent()) {
+                            return classType.get();
+                        }
+                    }
+                    return typeSymbol;
+                })
                 .filter(typeSymbol -> typeSymbol instanceof ClassSymbol)
                 .flatMap(typeSymbol -> (((ClassSymbol) typeSymbol).initMethod()));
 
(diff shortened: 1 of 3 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ContextTypeResolver.java
```diff
@@ -811,6 +811,18 @@ private Optional<TypeSymbol> getPositionalArgumentTypeForNewExpr(NodeList<Functi
         Optional<MethodSymbol> methodSymbol = context.currentSemanticModel()
                 .flatMap(semanticModel -> semanticModel.typeOf(newExpressionNode))
                 .flatMap(typeSymbol -> Optional.of(CommonUtil.getRawType(typeSymbol)))
+                .map(typeSymbol -> {
+                    if (typeSymbol.typeKind() == TypeDescKind.UNION) {
+                        Optional<TypeSymbol> classType =
+                                ((UnionTypeSymbol) typeSymbol).memberTypeDescriptors().stream()
+                                        .map(CommonUtil::getRawType)
+                                        .filter(member -> member instanceof ClassSymbol).findFirst();
```
IMS94, line 819: Should we handle this here? What if we return the union from here and then ask the corresponding context (ExplicitNewExpressionNodeContext) to do the narrowing down?

IMS94 (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/ContextTypeResolver.java
malinthar: As per the offline discussion we had here the context type resolver should return the expected type of the positional argument node. Therefore we should not return the union type. We need to narrow it down here and the resolve a posible type symbol for the positional argument.

IMS94 (approved):

PR #37313, opened 2022-08-09
Adding an API to get the shape of the Singleton Type

### LINKED ISSUE #36995: [Semantic API] Need an API to get the shape of the SingletonType
**Description:**
Task of https://github.com/ballerina-platform/ballerina-lang/issues/35783#issuecomment-1181626283

**Describe your problem(s)**
There is no way to get the shape of the internal type of singleton-type symbols.
Ex: get the shape as Integer, Float, String

https://github.com/ballerina-platform/ballerina-lang/issues/35783

sanjana:
Fixes #36995

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaSingletonTypeSymbol.java
```diff
@@ -53,6 +55,10 @@ public BallerinaSingletonTypeSymbol(CompilerContext context, BLangLiteral shape,
         } else {
             this.typeName = shape.toString();
         }
+
+        // Getting the original type of the shape
+        TypesFactory typesFactory = TypesFactory.getInstance(this.context);
+        originalType = typesFactory.getTypeDescriptor(shape.getBType());
```
dulajdilshan, line 61: It is not good to create a type-factory instance inside the constructor. Why don't we pass the original type from the outside?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaConstantSymbol.java
```diff
@@ -162,6 +162,11 @@ private String toStringVal(Object obj, BType valType) {
         return String.valueOf(obj);
     }
 
+    @Override
+    public TypeSymbol originalType() {
```
dulajdilshan, line 166: We don't have to add `originalType()` API in constant symbols right?

sanjana: @dulajdilshan, `BallerinaConstantSymbol` class implements the `ConstantSymbol` interface which extends the `SingletonTypeSymbol` interface. Therefore, when a new API method is exposed via the `SingletonTypeSymbol` interface, its implementation should also be made available in the `BallerinaConstantSymbol` class. 

As per a discussion held with @mohanvive, it was decided to proceed with providing the `broaderTypeDescriptor` for constants which aligns with singleton's original type.

dulajdilshan: Ah okay. I noticed `ConstantSymbol` extends `SingletonType`.

dulajdilshan (approved):
LGTM

PR #37329, opened 2022-08-10
Fix array size mismatch in runtime array value and type

### LINKED ISSUE #35522: `ArrayValue` creation does not evaluate array size mismatch
**Description:**
$subject
The following public `ArrayValue` constructor is used in the value creator APIs which accepts parameters for array type and the array size.
```java
    public ArrayValueImpl(ArrayType type, long size, BListInitialValueEntry[] initialValues) {
    }
```
As we keep the size information in the array type as well, there can be two different sizes given from these parameters. 
Currently, this size difference is not validated when using this constructor. But it can cause erroneous behavior.

### LINKED ISSUE #35523: Support fill values for configurable variables
**Description:**
The following configurable variable is not supported currently. 
```ballerina
import ballerina/io;

configurable int[5] arr1 = ?;

public function main() {
    int[5] arr = [1, 2]; 
    io:println(arr); // [1,2,0,0,0]
    io:println(arr1); // [1,2,3]  - will fail after merging #35468 
}
```
This has to be supported by the configurable variables. 

can be related to #35522

### LINKED ISSUE #37332: Filler value check returns compilation error for readonly types
**Description:**
$subject

As per the spec,

> For type `readonly & T`,  It will be the filler value for `T` constructed as read-only if it belongs to the type `T`

https://ballerina.io/spec/lang/2022R2/#FillMember

The following code fails at compilation. 

```ballerina
public function main() {
    map<int>[5] & readonly arr = [];
}
```
error 
```
invalid usage of list constructor: type '(map<int> & readonly)[5] & readonly' does not have a filler value
```
**Affected Versions:**
latest master

This can affect #35523

HindujaB:
## Purpose
$subject

Fixes #35522
Fixes #35523
Fixes #37332 

## Approach
- Removed usages of creating array values with the size parameter and modified them to use the size from the array type.
- Deprecated such `ValueCreator` APIs.
- Fixed compiler bug for `readonly` filler values.
- Added configurable support for array fill values. 
```ballerina
configurable int[5] arr = ?; // will create [1,2,0,0,0]
```
in `Config.toml`
```toml
arr = [1, 2]
```

## Remarks 
We can't remove the following constructors as they are used to create array values from code generation.
- create a simple array 
https://github.com/ballerina-platform/ballerina-lang/blob/bbd084cd679015c8aff2ef56d126ded38030a5cd/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ArrayValueImpl.java#L261
- create a record array 
https://github.com/ballerina-platform/ballerina-lang/blob/bbd084cd679015c8aff2ef56d126ded38030a5cd/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/ArrayValueImpl.java#L265

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/runtime/api/values/modules/arrays/arrays.bal
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) (2020-2022), WSO2 Inc. (http://www.wso2.com).
```
warunalakshitha, line 1: Should this be only 2022

warunalakshitha (approved):
LGTM

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -6006,6 +6006,8 @@ public boolean hasFillerValue(BType type) {
                 return tupleType.getTupleTypes().stream().allMatch(eleType -> hasFillerValue(eleType));
             case TypeTags.TYPEREFDESC:
                 return hasFillerValue(getReferredType(type));
+            case TypeTags.INTERSECTION:
+                return hasFillerValue(((BIntersectionType) type).effectiveType);
```
Nadeeshan96, line 6010: Do we have to add this case to the `io.ballerina.runtime.internal.TypeChecker.hasFillerValue(Type, List<Type>)` method also?

HindujaB: Reproducible in the following code.
```ballerina
anydata arr = [];
(map<int> & readonly)[2] arr4 = checkpanic arr.cloneWithType();
```
Fixed in the latest commit.

Nadeeshan96:
LGTM. Btw there is a merge conflict.

warunalakshitha (approved):
LGTM

PR #37431, opened 2022-08-17
Fix reference equals for  function pointer values

### LINKED ISSUE #37379: Function pointers are not equal when the same pointer is passed as two different parameters
**Description:**
If the same function pointer is passed as two different parameters to a function, then there are not equal under the `===` operator.

**Steps to reproduce:**
The output of the following program returns false where it should be true.
```ballerina
function func() {}
function twoFunctionsEqual(function f, function g) returns boolean => f === g;

public function main() {
    io:println(twoFunctionsEqual(func, func));
}
```

**Affected Versions:**
2201.2.0-RC1

HindujaB:
## Purpose
$subject

Fixes #37379

## Approach
- Removed creating multiple lambda functions when creating multiple `FPValue` values that point to the same function. 
- Modified `refEquals()` logic to support function pointers pointing same function.

## Samples
```ballerina
import ballerina/io;
function func() {}

public function main() {
    io:println(func === func);
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
```diff
@@ -1763,6 +1772,7 @@ void generateFPLoadIns(BIRNonTerminator.FPLoad inst) {
 
         this.storeToVar(inst.lhsOp.variableDcl);
         asyncDataCollector.add(lambdaName, inst);
+        functions.put(funcKey, lambdaName);
```
warunalakshitha, line 1775: Should we only add if funcKey does not exist

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
```diff
@@ -1715,8 +1711,17 @@ void generateFPLoadIns(BIRNonTerminator.FPLoad inst) {
...
+
+        String funcKey = inst.pkgId.toString() + ":" + name;
+        String lambdaName;
+        if (functions.containsKey(funcKey)) {
+            lambdaName = functions.get(funcKey);
+        } else {
+            lambdaName = Utils.encodeFunctionIdentifier(inst.funcName.value) + "$lambda" +
+                    asyncDataCollector.getLambdaIndex() + "$";
+            functions.put(funcKey, lambdaName);
+        }
```
warunalakshitha, line 1723:
```suggestion
        String lambdaName = functions.get(funcKey);
        if (lambdaName == null) {
            lambdaName = Utils.encodeFunctionIdentifier(inst.funcName.value) + "$lambda" +
                    asyncDataCollector.getLambdaIndex() + "$";
            functions.put(funcKey, lambdaName);
        }
```

warunalakshitha (approved):

PR #37485, opened 2022-08-23
Add `from` keyword to common query completions

### LINKED ISSUE #37245: `from` keyword not suggesting after a `from-clause` in query expressions
**Description:**
In query expressions  multiple `from-clause` expressions can be followed by first `from-clause`. So `from` keyword should be suggested in the completion item list.

Screenshot 2022-08-03 at 22 32 22

LakshanWeerasinghe:
Fixes #37245

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ListConstructorExpressionNodeContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ListConstructorExpressionNodeContext.java
@@ -62,6 +62,12 @@ public class ListConstructorExpressionNodeContext extends AbstractCompletionProv
     }
 
     @Override
+    public boolean onPreValidation(BallerinaCompletionContext context, ListConstructorExpressionNode node) {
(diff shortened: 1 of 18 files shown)


malinthar (changes requested):
Changes look good!. However, one thing that we can do is adding few tests to cover the cases where from clause used as a intermediate-clause. 
Note: In the existing test cases we have not covered such cases. Following is one example to start with. Let's try out few scenarios and check if the LS support is desirable in those cases.

Eg:
```Ballerina
import ballerina/io;
public function main() {
    string[] result = from int a in [1,2,3] let int b = a * a 
    from string d in ["one","two","three"] let string power = string `${d}:${b}` select power;
    io:print(result);
}
```

mohanvive (approved):

mohanvive:
@malinthar Need to review on this.

malinthar (approved):
LGTM!

PR #37494, opened 2022-08-23
[2201.2.0-stage] Fix type checking issue for objects with resource methods

### LINKED ISSUE #37490: No compilation error when implementation not provided for a resource function
**Description:**
In the below code, Teacher class can be instantiated even without providing the implementation for `get name()` resource function. Instantiating Teacher class should result in a compilation error  but this code complies just fine

**Steps to reproduce:**
```ballerina
import ballerina/io;

type Person distinct service object {
    resource function get name() returns string;
};

distinct service class Teacher {
    *Person;

    resource function get salary() returns decimal {
        return 1000.00;
    }
}

    Person p = getPerson();
    if p is Teacher {
        io:println("Teacher");
    }
}

function getPerson() returns Person {
    Teacher t = new;
    return t;
}
```

**Affected Versions:**

rdulmina:
## Purpose
$title

Fixes #37490

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -4546,14 +4546,11 @@ private void validateReferencedFunction(Location pos, BAttachedFunction func, Sy
             return;
         }
 
-        // Service typing does not consider resource methods when type checking.
-        if (Symbols.isResource(func.symbol)) {
-            return;
-        }
-
         // There must be an implementation at the outer level, if the function is an interface.
```
MaryamZi, line 4549: Not related to your change, but this comment isn't quite correct. Let's remove/update separately.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -4546,14 +4546,11 @@ private void validateReferencedFunction(Location pos, BAttachedFunction func, Sy
...
         if (!env.enclPkg.objAttachedFunctions.contains(func.symbol)) {
-            dlog.error(pos, code, func.funcName, func.symbol.receiverSymbol.type);
+            // Here we use the whole function in the error msg since resource function name will contain `$`s if 
+            // we use the func.funcName
```
MaryamZi, line 4552: Shall we do this only for resource methods? It complicates/lengthens error messages for normal methods (the common case) and remote methods.

MaryamZi (commented):
I think we need to add a significant number of tests since typing is a bit different from other methods.

e.g.,
```ballerina
type Foo client object {
    resource function get foo() returns string;
};

client class Bar {
    *Foo;
    resource function post foo() returns string => "";
}
```

And resource methods aren't the same as what they used to be in early days. 

Given the time constraints, let's try to cover as many different scenarios as possible in this PR, and create an issue if necessary to track adding more tests.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
rdulmina: But that will give inconsistent error messages for resource functions and other functions?

tests/jballerina-unit-test/src/test/resources/test-src/klass/class-def-negative.bal
```diff
@@ -13,3 +13,26 @@ type C object {
     *A;
 };
 
+type ServiceObject service object {
+     resource function get name() returns string;
```
SasinduDilshara, line 17: indentation is wrong.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -5055,9 +5055,8 @@ private void defineReferencedFunction(Location location, Set<Flag> flagSet, Symb
             return;
         }
 
-        if (Symbols.isPrivate(referencedFuncSymbol) || Symbols.isResource(referencedFuncSymbol)) {
-            // we should not copy private functions. And we ignore the resource functions as they are not part of the
-            // type.
+        if (Symbols.isPrivate(referencedFuncSymbol)) {
+            // we should not copy private functions
```
SasinduDilshara, line 5059: Do we have tests case for private functions in this PR

rdulmina: No that is existing code

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: That is better than complicating error messages for the common case IMO.

MaryamZi: And they are different from other methods anyway. Even for resource methods, we can see if we can improve the error to not include stuff that doesn't actually help to identify the method specifically (like params), and make them more uniform with other methods.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -4546,14 +4546,14 @@ private void validateReferencedFunction(Location pos, BAttachedFunction func, Sy
...
-
-        // There must be an implementation at the outer level, if the function is an interface.
         if (!env.enclPkg.objAttachedFunctions.contains(func.symbol)) {
-            dlog.error(pos, code, func.funcName, func.symbol.receiverSymbol.type);
+            if (Symbols.isResource(func.symbol)) {
+                // Use the function signature in the error msg since resource function name will contain `$`s if
+                // we use the func.funcName
+                dlog.error(pos, code, func, func.symbol.receiverSymbol.type);
+            } else {
+                dlog.error(pos, code, func.funcName, func.symbol.receiverSymbol.type);
```
MaryamZi, line 4555: Let's extract `func.symbol` out to a variable.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/klass/ServiceClassTest.java
```diff
@@ -43,8 +43,12 @@ public void testBasicStructAsObject() {
 
     @Test
     public void testResourceMethodsDoesNotAffectAssignability() {
-        CompileResult compileResult = BCompileUtil.compile("test-src/klass/resource-method-assignability-test.bal");
-        BRunUtil.invoke(compileResult, "testServiceObjectValue");
+        CompileResult result = BCompileUtil.compile("test-src/klass/resource-method-assignability-negative.bal");
```
MaryamZi, line 46: We usually include `test` in the file.
```suggestion
        CompileResult result = BCompileUtil.compile("test-src/klass/resource-method-assignability-negative-test.bal");
```

Not introduced by this PR, but we use underscores in bal file names.

tests/jballerina-unit-test/src/test/resources/test-src/klass/class-def-negative.bal
```diff
@@ -13,3 +13,26 @@ type C object {
...
+    *ClientObject;
+}
+
+type ClientObject2 client object {
+    *ClientObject;
+};
+
+client class ClientClass2 {
+    *ClientObject2;
+}
```
MaryamZi, line 38: We can also include tests where what's included is a class, rather than just an object typedesc.

MaryamZi (approved):

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/klass/ServiceClassTest.java
rdulmina: Yes, will update

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3608,11 +3609,34 @@ private boolean isInvalidNeverField(BField lhsField, BRecordType rhsType) {
...
                                                        Set<TypePair> unresolvedTypes) {
-        return rhsFuncList.stream()
+        BAttachedFunction matchingFunction = rhsFuncList.stream()
                 .filter(rhsFunc -> lhsFunc.funcName.equals(rhsFunc.funcName))
                 .filter(rhsFunc -> isFunctionTypeAssignable(rhsFunc.type, lhsFunc.type, unresolvedTypes))
                 .findFirst()
                 .orElse(null);
+        
+        // Todo: We could include this logic in `isFunctionTypeAssignable` if we have `resourcePathType` information in
+        // `BInvokableType`
```
MaryamZi, line 3619: Let's create an issue.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -3608,11 +3609,34 @@ private boolean isInvalidNeverField(BField lhsField, BRecordType rhsType) {
...
+        if (matchingFunction == null || 
+                !Symbols.isResource(matchingFunction.symbol) || !Symbols.isResource(lhsFunc.symbol)) {
+            return matchingFunction;
+        }
+
+        List<BType> lhsFuncResourcePathTypes = ((BResourceFunction) lhsFunc).resourcePathType.tupleTypes;
```
MaryamZi, line 3625: Don't we need these changes at runtime too?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
rdulmina: #37502

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
rdulmina: added in [64953a6](https://github.com/ballerina-platform/ballerina-lang/pull/37494/commits/64953a6b9690cf3ea747fc453cb01bc1b176428c)

rdulmina:
Will add is exper tests

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/ClientType.java
```diff
@@ -0,0 +1,26 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package io.ballerina.runtime.api.types;
+
+/**
+ * {@code BServiceType} represents a client object in Ballerina.
```
warunalakshitha, line 20: Should be Client Type?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/NetworkObjectType.java
```diff
@@ -0,0 +1,29 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package io.ballerina.runtime.api.types;
+
+/**
+ * {@code BServiceType} represents a network object in Ballerina.
```
warunalakshitha, line 20: java doc needs to be corrected

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -1902,6 +1950,15 @@ private static boolean checkIsServiceType(Type sourceType, Type targetType, List
         return false;
     }
 
+    private static boolean checkIsClientType(Type sourceType, Type targetType, List<TypePair> unresolvedTypes) {
+        if (sourceType.getTag() == TypeTags.CLIENT_TAG) {
+            return checkObjectEquivalency(sourceType, (BObjectType) targetType, unresolvedTypes);
+        }
+
+        // TODO: Check for client in the flags
```
warunalakshitha, line 1958: Do we need this? Shall we directly  return sourceType.getTag() == TypeTags.CLIENT_TAG && checkObjectEquivalency(sourceType, (BObjectType) targetType, unresolvedTypes)

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/TypeTags.java
```diff
@@ -61,7 +61,8 @@ public class TypeTags {
     public static final int ANY_TAG = INVOKABLE_TAG + 1;
     public static final int ENDPOINT_TAG = ANY_TAG + 1;
     public static final int SERVICE_TAG = ENDPOINT_TAG + 1;
-    public static final int ARRAY_TAG = SERVICE_TAG + 1;
+    public static final int CLIENT_TAG = SERVICE_TAG + 1;
```
MaryamZi, line 64: Since this is part of the public API, shouldn't we avoid changes like this? The value of the subsequent tags changes with this, so we should probably add CLIENT_TAG at the end.

MaryamZi: Do we need CLIENT_TAG anyway? This will also break existing code if someone has checked for the object tag for clients, right?

MaryamZi: It is cleaner to have CLIENT_TAG, but I'm not sure if we should do it in an update release. cc @warunalakshitha

MaryamZi: On further thought, we probably should have only OBJECT_TAG, and use flags to differentiate between non-networks objects vs service objects vs client objects. We can't remove `SERVICE_TAG` now either though.

rdulmina: We dont hardcode the values right, so changing the value of subsequent tags is ok WDYT?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
rdulmina: `checkIsServiceType` uses this approach. Not sure we need this for `checkIsClientType`

https://github.com/ballerina-platform/ballerina-lang/blob/a1af3b4c1c72ca713a827fb2716346fff08b3116/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java#L1939-L1950

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/TypeTags.java
MaryamZi: Irrespective of how the value is assigned, what type tags are is an integer value associated with each tag. While it is not the ideal way to do it, a user may have checked for equality by the value of the tag.

For example, to check for an array they could have done
```java
type.getTag() == 32 
```

So this can be a breaking change.

rdulmina: Updated with [fb79ee8](https://github.com/ballerina-platform/ballerina-lang/pull/37494/commits/fb79ee83ea126476bbc3ec0fe32f3fe952e6b081)

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/TypeTags.java
```diff
@@ -85,6 +85,7 @@ public class TypeTags {
 
     public static final int PARAMETERIZED_TYPE_TAG = READONLY_TAG + 1;
     public static final int TYPE_REFERENCED_TYPE_TAG = PARAMETERIZED_TYPE_TAG + 1;
+    public static final int CLIENT_TAG = TYPE_REFERENCED_TYPE_TAG + 1;
```
MaryamZi, line 88: Do we really need a separate tag? Related to https://github.com/ballerina-platform/ballerina-lang/pull/37494#discussion_r955892840.

rdulmina: Yes, for cases like https://github.com/ballerina-platform/ballerina-lang/blob/a326d859d7b74f3143307a476b234ee175712ca7/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java#L1697-L1700

We cannot use `OBJECT_TYPE_TAG` for clients.

MaryamZi: Why can't we check for object tag and the client flag?

rdulmina: We don't have a client flag in `TypeFlags.java`

MaryamZi: We have it at and use it at compile-time, right? We can add use it at runtime and add it to the flags. Should be `io.ballerina.runtime.api.flags.SymbolFlags` btw.

rdulmina: Updated with [1f3ba1f](https://github.com/ballerina-platform/ballerina-lang/pull/37494/commits/1f3ba1fa2403bfd99c0dd830d9003032a86ba40b)

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -1685,17 +1687,29 @@ private static boolean checkObjectEquivalency(Object sourceVal, Type sourceType,
...
+                (targetType.flags & SymbolFlags.CLIENT) == SymbolFlags.CLIENT) {
+            BNetworkObjectType targetNetworkObj = (BNetworkObjectType) targetType;
+            Collections.addAll(targetFuncs, targetNetworkObj.getResourceMethods());
+        }
+
+        if (sourceType.getTag() == TypeTags.SERVICE_TAG || 
+                (sourceObjectType.flags & SymbolFlags.CLIENT) == SymbolFlags.CLIENT) {
+            BNetworkObjectType sourceNetworkObj = (BNetworkObjectType) sourceObjectType;
+            Collections.addAll(sourceFuncs, sourceNetworkObj.getResourceMethods());
+        }
```
KavinduZoysa, line 1703: I think that it will be clearer if we can write a small method and call two times with passing arguments appropriately.

rdulmina: Updated in [c7f4e52](https://github.com/ballerina-platform/ballerina-lang/pull/37494/commits/c7f4e52017b2fe3dbcede63d7fbe387cbeca5ea8)

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeChecker.java
```diff
@@ -1816,15 +1828,46 @@ private static boolean isInSameVisibilityRegion(String lhsTypePkg, String rhsTyp
...
                                                                           unresolvedTypes))
-                .findFirst()
-                .orElse(null);
+                .findFirst();
+
+        if (matchingFunction.isEmpty()) {
+            return matchingFunction;
+        }
+        
+        MethodType matchingFunc = matchingFunction.get();
```
warunalakshitha, line 1844: shall we adda a comment here for finding matching function.

warunalakshitha (approved):
LGTM

PR #37539, opened 2022-08-26
Fix escaping special characters method omitting `\` char

### LINKED ISSUE #36912: Special Character escape does not escapes backshash
**Description:**
If an identifier contains special characters, those characters are escaped except for the backslash. See, the below example.

If the input JSON is something like this,
```ballerina
json jsonTest = {
    "firstName": "Joe",
    "last\\Na+me?": "Root"
};
```
Then the expected output should be,
```ballerina
type NewRecord record {
    string firstName;
    string last\\Na\+me\?;
};
```
But the actual output is,
```ballerina
type NewRecord record {
    string firstName;
    string last\Na\+me\?;
};
```
Which is an error where backslash is not escaped.

**Steps to reproduce:**
```java
import static io.ballerina.identifier.Utils.escapeSpecialCharacters;

String identifier = "last\\Na+me?";
identifier = escapeSpecialCharacters(identifier);
```
The output supposed to be `last\\Na\+me\?` but the erroneous output is `last\Na\+me\?`

### LINKED ISSUE #37514: Identifier java module should only be exposed compiler and runtime
**Description:**
This module should not exposed to outside.

HindujaB:
## Purpose
$subject 

Fixes #36912 
Fixes #37514 

## Approach
Added `\` to escaping characters set after verifying the current usages.

## Samples
```ballerina
import ballerina/io;
type Person record {
    string 'name\\_\:\@2;
};

public function main() {
    Person person = {'name\\_\:\@2: "Hinduja"};
    string? result = person["name\\_:@2"]; // should pass
    io:println(result.toString());
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-cli/build.gradle
+++ b/cli/ballerina-cli/build.gradle
@@ -44,7 +44,6 @@ dependencies {
     implementation project(':maven-resolver')
     implementation project(':ballerina-shell:shell-cli')
     implementation project(':toml-parser')
-    implementation project(':identifier-util')
     testCompile project(':ballerina-test-utils')
 
     testCompile 'org.testng:testng'
--- a/cli/ballerina-cli/src/main/java/module-info.java
+++ b/cli/ballerina-cli/src/main/java/module-info.java
@@ -18,5 +18,4 @@ module io.ballerina.cli {
     requires slf4j.api;
     requires io.ballerina.shell.cli;
     requires io.ballerina.toml;
-    requires io.ballerina.identifier;
 }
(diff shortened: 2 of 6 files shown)


warunalakshitha (approved):
LGTM

PR #37546, opened 2022-08-26
Fix create variable and check error code action to not return $CompilationError$

### LINKED ISSUE #37252: `$CompilationError$` is returned with create variable and check error code action
**Description:**
Consider the following scenario where we only have `returns` without the return type.

https://user-images.githubusercontent.com/61020198/182784542-9263a5c0-40f9-49db-8d4a-f47bcdc002a3.mov

In this scenario when we apply the Create variable and check error code action, `$CompilationError$|error` is returned with the variable creation.

**Steps to reproduce:**
```
function createIntWithError() returns int|error {
    return 0;
}

class MyClass {
    function testFunction() returns {
       createIntWithError();
    }
}
```

mindula:
## Purpose
This PR fixes the create variable and check error code action to not return `$CompilationError$` when we only have `returns` without the return type.

Fixes #37252

## Samples

https://user-images.githubusercontent.com/61020198/188059583-845829de-6c26-49db-8c38-70ee10797ff1.mov

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionNodeValidator.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionNodeValidator.java
@@ -292,8 +292,7 @@ public class CodeActionNodeValidator extends NodeTransformer<Boolean> {
                 && !node.closeParenToken().isMissing()
                 && node.leadingInvalidTokens().isEmpty()
                 && node.trailingInvalidTokens().isEmpty()
-                && (node.returnTypeDesc().isEmpty()
-                || node.returnTypeDesc().get().apply(this))
+                && (node.returnTypeDesc().isEmpty() || node.returnTypeDesc().get().apply(this))
                 && node.parameters().stream().allMatch(parameterNode -> parameterNode.apply(this));
     }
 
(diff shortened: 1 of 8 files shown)


gayalkuruppu (commented):
The change fixes the intended problem in the issue (#37252), but for this scenario we should not provide code action since the return-type-descriptor is invalid. Hence we need to improve the CodeActionNodeValidator to validate this scenario (and other scenarios where we have incorrect return-type-descriptors) and not provide any code action.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/CodeActionNodeValidator.java
```diff
@@ -291,6 +292,8 @@ public Boolean transform(FunctionSignatureNode node) {
                 && !node.closeParenToken().isMissing()
                 && node.leadingInvalidTokens().isEmpty()
                 && node.trailingInvalidTokens().isEmpty()
+                && (node.returnTypeDesc().isEmpty()
+                || node.returnTypeDesc().get().apply(this))
```
gayalkuruppu, line 296: Keeping these two in a single line is more readable IMO. WDYT?

mindula: +1

gayalkuruppu (approved):
LGTM

PR #37547, opened 2022-08-26
[Master] Fix the incorrect diagnostic in query expression

### LINKED ISSUE #37450: Incorrect diagnostic property in query expression
**Description:**
Consider the diagnostic `variable assignment is required` at cursor position.

Screenshot 2022-08-19 at 09 29 30

In the diagnostic properties which is used to derive the expected type, we get the `bType` incorrectly as `xml<string>` where it should be `string[]`.

Screenshot 2022-08-19 at 09 28 14

For a different case like `x1.toBalString()`, we get the type correctly as `string` in the diagnostic properties.

**Steps to reproduce:**
```
public function main() {
    xml x1 = xml `<book>The Lost World</book>`;

    from xml element in x1 select element.toBalString();
}
```

**Affected Versions:**
2201.2.0 (Swan Lake Update 2)

### LINKED ISSUE #37670: [Bug]: When inferring type for query-exp, collections of types other than query construct types should throw an error
### Description

From the spec https://ballerina.io/spec/lang/master/#section_6.35:
> If there is no contextually expected type, then the basic type of the value constructed is the basic type of expression following in; it is a compile-time error if the static type of this expression is not a subtype of one of the basic types that a query expression can construct.

```ballerina
import ballerina/io;
public function main() {
    xml x = xml ``;
    var res = from xml i in x // we should get an error here incompatible types: 'xml' cannot be constrained with 'int'
    select 1;

    io:println(res is int[]); //error incompatible types: 'xml<int>' will not be matched to 'int[]'
}
```

```ballerina
public function main() {
    var res = from string i in "x"
    select 1; // atm, no errors. should be an error

    string x = res;
}
```

### Steps to Reproduce

_No response_

### Affected Version(s)

2201.2.0

_No response_

### Labels

CompilerFE

_No response_

_No response_

_No response_

SasinduDilshara:
Fixes #37450
Fixes #37670

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
```diff
@@ -116,6 +116,16 @@ function assignAnyToUnionWithErrorAndAny() {
...
 
+function testAssignVarInQueryExpression() {
+    xml x1 = xml `<book>The Lost World</book>`;
+
+    var x2 = from xml element in x1 select element.toBalString();
+    assertTrue(x2 is string[]);
+
+    var x3 = from xml element in x1 select element;
+    assertTrue(x3 is xml[]);
+}
```
KavinduZoysa, line 127: Shall we add a test case to assert the output of this query expression?

tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
```diff
@@ -116,6 +116,16 @@ function assignAnyToUnionWithErrorAndAny() {
     assertEquality(4, y);
 }
 
+function testAssignVarInQueryExpression() {
+    xml x1 = xml `<book>The Lost World</book>`;
+
+    var x2 = from xml element in x1 select element.toBalString();
+    assertTrue(x2 is string[]);
```
rdulmina, line 123: let's assign x2 to a `string[]` variable (`string[] x21 = x2;`)

tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
SasinduDilshara: I will do it.

tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
SasinduDilshara: I will do it

SasinduDilshara: Updated. 8f225565d0cb19cd953a142da1c38498beefe3a7

tests/jballerina-unit-test/src/test/resources/test-src/statements/assign/assign-stmt.bal
SasinduDilshara: Updated.8f225565d0cb19cd953a142da1c38498beefe3a7

prakanth97:
Added line #L6318 was not covered by tests, shall we add tests to cover that as well?

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/statements/assign/AssignStmtTest.java
```diff
@@ -270,6 +270,11 @@ public void assignAnyToUnionWithErrorAndAny() {
         BRunUtil.invoke(result, "assignAnyToUnionWithErrorAndAny");
     }
 
+    @Test()
+    public void assignVarInQueryExpression() {
```
prakanth97, line 274: Shall we change to 
```suggestion
    public void testAssignQueryExpressionToVar() {
```

SasinduDilshara: Will update.

SasinduDilshara: Added. da824d8c601d69e9fa8df16a5167e6e999de48c3

SasinduDilshara:
> Added line #L6318 in TypeChecker.java was not covered by tests, shall we add tests to cover that as well?

Added. da824d8c601d69e9fa8df16a5167e6e999de48c3

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6299,25 +6299,25 @@ private BType getCompletionType(BType collectionType, BLangQueryExpr queryExpr,
     }
 
     private BType getNonContextualQueryType(BType staticType, BType basicType) {
```
pcnfernando, line 6301: The problem here is we are not getting the actual basic type
From the spec https://ballerina.io/spec/lang/2022R1/#query-expr:
> If there is no contextually expected type, then the basic type of the value constructed is the basic type of expression following in; it is a compile-time error if the static type of this expression is not a subtype of one of the basic types that a query expression can construct.

Also, returning array type as the result type for all other cases also incorrect which would be a breaking change. Hence, please create a separate issue for that and may be handle it in this same PR

pcnfernando: Created https://github.com/ballerina-platform/ballerina-lang/issues/37670. @SasinduDilshara can you please have a look

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6291,26 +6291,37 @@ private BType getCompletionType(BType collectionType, BLangQueryExpr queryExpr,
         return completionType;
     }
 
-    private BType getNonContextualQueryType(BType staticType, BType basicType) {
-        BType resultType;
+    private BType getNonContextualQueryType(BType staticType, BType basicType, Location pos) {
         switch (basicType.tag) {
             case TypeTags.TABLE:
-                resultType = symTable.tableType;
+                if (types.isAssignable(staticType, symTable.mapAllType)) {
```
SandaruJayawardana, line 6297: Why do we need this check?

SasinduDilshara: Table types only allow map types as the static type in query If there is no contextually expected type.
Types other than map will give an error.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6291,26 +6293,37 @@ private BType getCompletionType(BType collectionType, BLangQueryExpr queryExpr,
...
+            case TypeTags.TUPLE:
+            case TypeTags.OBJECT:
+                return new BArrayType(staticType);
             default:
-                resultType = new BArrayType(staticType);
-                break;
+                return symTable.semanticError;
         }
-        return resultType;
+        dlog.error(pos, INCOMPATIBLE_QUERY_CONSTRUCT_TYPE, staticType, basicType);
```
lasinicl, line 6325: For something like this
```
    table<Employee> key(name) t = table [
        { name: "John", salary: 100 },
        { name: "Jane", salary: 200 }
    ];

    var res = from Employee i in t
    select 1;
```
the error would be `incompatible type 'int', expected a construct type of 'table<Employee> key(name)'`. Shouldn't the error be something like
`incompatible types: 'table' cannot be constrained with 'int'`, since the basic type that the query can construct is a 'table' here.

lasinicl: We might need to consider the referred type here to handle the type reference types.
Now the following code crashes at runtime
```ballerina
type T1 table<Employee> key(name);

public function main() {
    T1 t = table [
        { name: "John", salary: 100 },
        { name: "Jane", salary: 200 }
    ];

    var _ = from var i in t
    select 1;
}
```

SasinduDilshara: Updated.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6356,26 +6358,38 @@ private BType getCompletionType(BType collectionType, BLangQueryExpr queryExpr,
...
                 break;
             case TypeTags.STRING:
-                resultType = symTable.stringType;
+                if (types.isSubTypeOfBaseType(staticType, TypeTags.STRING)) {
+                    return symTable.stringType;
+                }
                 break;
+            case TypeTags.ARRAY:
+            case TypeTags.TUPLE:
+            case TypeTags.OBJECT:
```
pcnfernando, line 6386: Can you please check how the below is validate
```ballerina
class Iterable {
    *object:Iterable;
    public function iterator() returns object {
        public isolated function next();
    } {
        return object {    
            public isolated function next()  {
            }
        };
    }
}

public function main() {
    Iterable ite = new();
    var listResult = from var item in ite
    select item;
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -6356,26 +6358,38 @@ private BType getCompletionType(BType collectionType, BLangQueryExpr queryExpr,
...
             case TypeTags.TABLE:
-                resultType = symTable.tableType;
+                if (types.isAssignable(staticType, symTable.mapAllType)) {
+                    return symTable.tableType;
+                }
                 break;
             case TypeTags.STREAM:
-                resultType = symTable.streamType;
-                break;
+                return symTable.streamType;
```
pcnfernando, line 6370: Add a test for stream constrained with literal

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
SasinduDilshara: In here, the static type is become `other`.
It is happen because https://github.com/ballerina-platform/ballerina-lang/issues/38534#issuecomment-1301631231.
Created a issue for this https://github.com/ballerina-platform/ballerina-lang/issues/38534

SasinduDilshara: IMO, this should resolve in separate PR. @pcnfernando wdyt?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
SasinduDilshara: Updated. 933052368982fe42d170de9491275b990a8ce1ed

tests/jballerina-unit-test/src/test/resources/test-src/query/simple-query-with-var-type.bal
```diff
@@ -195,14 +195,6 @@ function testFilteringNullElements() returns Person[] {
     return outputPersonList;
 }
 
-function testMapWithArity() returns boolean {
-    map<any> m = {a: "1A", b: "2B", c: "3C", d: "4D"};
-    var val = from var v in m
-                   where <string> v == "1A"
-                   select <string>v;
-    return val == ["1A"];
-}
```
KavinduZoysa, line 204: Can't we turn this into a positive test case if not available?
Also, did you add a negative test for this case?

SasinduDilshara: There are negative tests for map types in [query-expr-query-construct-type-negative.bal](https://github.com/ballerina-platform/ballerina-lang/pull/37547/files#diff-d872f8d1b5d70e08e4a1e2c3d798402ed49735fd594cf13144d2e0fa6fb4f8f3)

SasinduDilshara: updated

IMS94:
@gimantha @SasinduDilshara shall we get this merged?

gimantha:
@SasinduDilshara Can you please resolve the conflicts?

SasinduDilshara:
> @SasinduDilshara Can you please resolve the conflicts?

Resolved now

pcnfernando:
@SasinduDilshara please resolve the conflict

pcnfernando:
@SasinduDilshara please resolve the conflict

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/QueryTypeChecker.java
```diff
@@ -529,26 +532,38 @@ private BType getResolvedType(BType initType, BType expType, boolean isReadonly,
         return initType;
     }
 
-    private BType getNonContextualQueryType(BType staticType, BType basicType) {
-        BType resultType;
-        switch (Types.getReferredType(basicType).tag) {
+    private BType getNonContextualQueryType(BType staticType, BType basicType, Location pos) {
+        basicType = Types.getReferredType(basicType);
```
pcnfernando, line 536: If we have a type reference as the basic type, the error will have the referred type if we change it here

pcnfernando (approved):
LGTM

sanjana:
@SasinduDilshara Could you please re-sync this with the master?

PR #37622, opened 2022-09-05
Allow ambiguous target types with `cloneWithType` and `fromJsonWithType`

### LINKED ISSUE #36986: Handle ambiguities with fromJsonWithType
Related to spec issue https://github.com/ballerina-platform/ballerina-spec/issues/1088

Nadeeshan96:
$title
Same behaviour is also applied for `convertJSON` method which is used in `io.ballerina.runtime.api.utils.JsonUtils`
Fixes #36986

Also fixed float to decimal conversion. Java API suggests to use `valueOf` method to create a java `BigDecimal` using java `Double`.

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -260,151 +259,116 @@ static boolean isConvertibleToFloatingPointTypes(Object value) {
...
+            if (inputValue == valueSpaceItem) {
+                return Set.of(inputValueType);
+            }
+            if (TypeChecker.isFiniteTypeValue(inputValue, inputValueType, valueSpaceItem, false)) {
+                return Set.of(TypeChecker.getType(valueSpaceItem));
             }
-            if (inputValueType == memType || isIntegerSubtypeAndConvertible(inputValue, memType)) {
-                return Set.of(memType);
+        }
+        for (Object valueSpaceItem : finiteTypeValueSpace) {
```
gabilang, line 357: Can we reuse the above `for` loop by keeping the first convertible type (if there is any) inside a set and return the set outside the `for` loop?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/JsonInternalUtils.java
```diff
@@ -372,22 +363,23 @@ public static Object convertJSON(Object jsonValue, Type targetType) {
...
+                Type inputValueType = TypeChecker.getType(jsonValue);
+                Set<Object> finiteTypeValueSpace = finiteType.valueSpace;
+                for (Object valueSpaceItem : finiteTypeValueSpace) {
                     if (jsonValue == valueSpaceItem) {
                         return convertJSON(jsonValue, inputValueType);
                     }
-                    if (TypeChecker.isFiniteTypeValue(jsonValue, inputValueType, valueSpaceItem, true)) {
-                        matchedValues.add(valueSpaceItem);
+                    if (TypeChecker.isFiniteTypeValue(jsonValue, inputValueType, valueSpaceItem, false)) {
+                        return convertJSON(jsonValue, TypeChecker.getType(valueSpaceItem));
```
gabilang, line 373: Can we add more test cases to cover these missing lines?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
Nadeeshan96: We can do, but I think `isFiniteTypeValue` and `getConvertibleTypes` methods are costly than the for loop. If the source value is of the target type, calls to those methods will be in vain. Hence I think it is better to have the expected sequential order.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TypeConverter.java
```diff
@@ -481,17 +456,17 @@ private static boolean isConvertibleToRecordType(Object sourceValue, BRecordType
...
                                 targetFieldTypes.get(fieldName) + "', found '" +
                                 getShortSourceValue(valueEntry.getValue()) + "'");
                         returnVal = false;
                     }
                 } else if (!targetType.sealed) {
-                    if (getConvertibleTypesFromJson(valueEntry.getValue(), restFieldType, fieldNameLong,
-                            unresolvedValues, errors).size() != 1) {
+                    if (getConvertibleTypeFromJson(valueEntry.getValue(), restFieldType, fieldNameLong,
+                            unresolvedValues, errors, allowNumericConversion) ==  null) {
                         addErrorMessage(errors.size() - initialErrorCount, errors, "value of field '" +
```
warunalakshitha, line 470: Should we add this error messages to Messagebundle.properties. May be we can do as separate task

Nadeeshan96: This is for creating a String which will only be a part of the error message. That will be then used with an error message already in  Messagesbundle.properties. So, we can also keep this like this.

Nadeeshan96: Created https://github.com/ballerina-platform/ballerina-lang/issues/37822 to do this task.

warunalakshitha (approved):
LGTM

PR #37636, opened 2022-09-05
Introduce several validations for client declarations

MaryamZi:
## Purpose
Introduce the following validations.
- Disallow publicly exposing constucts from modules generated for clients at least until we use a type instead of the typedesc with inferring. If not may cause issues with the BIR pointing to generated modules by name.
- Validate that the module generated for a client declaration contains either a class or an object type with the name `client`.
- Disallow mutable module-level variables in a module generated for a client declaration.

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -5198,7 +5198,7 @@ public class SemanticAnalyzer extends SimpleBLangNodeAnalyzer<SemanticAnalyzer.A
 
         @Override
         public void visit(BUnionType bUnionType) {
-            for (BType memType : bUnionType.getMemberTypes()) {
+            for (BType memType : bUnionType.getOriginalMemberTypes()) {
                 visitType(memType);
             }
         }
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -5061,14 +5061,14 @@ public class SemanticAnalyzer extends SimpleBLangNodeAnalyzer<SemanticAnalyzer.A
         List<BType> exposedTypes;
         PackageID currentPkgId;
 
-        public BClientDeclConstructExposureDisallowingTypeVisitor(SymbolResolver symResolver, SymbolTable symTable) {
+        private BClientDeclConstructExposureDisallowingTypeVisitor(SymbolResolver symResolver, SymbolTable symTable) {
             this.symResolver = symResolver;
             this.symTable = symTable;
             this.unresolvedTypes = Collections.emptySet();
             this.exposedTypes = Collections.emptyList();
         }
 
-        public void visitType(BType type, List<BType> exposedTypes, PackageID currentPkgId) {
+        private void visitType(BType type, List<BType> exposedTypes, PackageID currentPkgId) {
             this.unresolvedTypes = new HashSet<>();
             this.exposedTypes = exposedTypes;
             this.currentPkgId = currentPkgId;
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
@@ -38,6 +38,7 @@ import org.wso2.ballerinalang.compiler.diagnostic.BLangDiagnosticLog;
 import org.wso2.ballerinalang.compiler.parser.BLangAnonymousModelHelper;
 import org.wso2.ballerinalang.compiler.semantics.model.SymbolEnv;
 import org.wso2.ballerinalang.compiler.semantics.model.SymbolTable;
+import org.wso2.ballerinalang.compiler.semantics.model.TypeVisitor;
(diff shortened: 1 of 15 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -4923,6 +4943,321 @@ private Map<String, Location> getFieldLocations(List<BLangSimpleVariable> fields
...
+        List<BType> exposedTypes = new ArrayList<>(1);
+        collector.visitType(type, exposedTypes, currentPkgId);
+        if (!exposedTypes.isEmpty()) {
+            dlog.error(location,
+                       DiagnosticErrorCode.UNSUPPORTED_EXPOSURE_OF_CONSTRUCT_FROM_MODULE_GENERATED_FOR_CLIENT_DECL);
+        }
+    }
+
+    private static class BClientDeclConstructExposureDisallowingTypeVisitor implements TypeVisitor {
+
```
SasinduDilshara, line 5057: We can remove this line

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -4923,6 +4943,321 @@ private Map<String, Location> getFieldLocations(List<BLangSimpleVariable> fields
...
+            if (!unresolvedTypes.add(type)) {
+                return;
+            }
+
+            BTypeSymbol tsymbol = type.tsymbol;
+            if (tsymbol != null &&
+                    !this.currentPkgId.equals(tsymbol.pkgID) &&
+                    this.symTable.clientDeclarations.containsValue(tsymbol.pkgID)) {
+                this.exposedTypes.add(type);
+                return;
```
SasinduDilshara, line 5092: Don't we need to remove `type` from `unresolvedTypes` before return in here.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -355,6 +374,7 @@ public void visit(BLangPackage pkgNode, AnalyzerData data) {
             data.env = lambdaFunction.capturedClosureEnv;
             analyzeDef(lambdaFunction.function, data);
         }
+        validateClientDeclarations(pkgNode, data);
```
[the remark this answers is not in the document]
MaryamZi: Intentionally left this new line as is, since the preceding logic (including this addition) applies to analyzing a single module, whereas the next line calls analyze for the testable modules.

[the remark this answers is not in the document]
SasinduDilshara: +1

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: Any specific reason? I prefer having it as it since it's clearer at first glance. We've done the same even for the `SemanticAnalyzer` class.

SasinduDilshara: I suggest it because, the other methods in this class haven't new line at the top.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: Why do we have to? `unresolvedTypes` is always a new set when the analysis is called.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
MaryamZi: But this is a class?

SasinduDilshara: yes. Got it now. Then agree with you

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
SasinduDilshara: yes. agree with you now.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -4923,6 +4943,321 @@ private Map<String, Location> getFieldLocations(List<BLangSimpleVariable> fields
...
+        PackageID currentPkgId;
+
+        public BClientDeclConstructExposureDisallowingTypeVisitor(SymbolResolver symResolver, SymbolTable symTable) {
+            this.symResolver = symResolver;
+            this.symTable = symTable;
+            this.unresolvedTypes = Collections.emptySet();
+            this.exposedTypes = Collections.emptyList();
+        }
+
+        public void visitType(BType type, List<BType> exposedTypes, PackageID currentPkgId) {
```
gimantha, line 5071: We dont need to make this `public` right?

gimantha (approved):
LGTM

PR #37658, opened 2022-09-07
[2201.2.x] Fix module level completions before and after module level decls and defns

### LINKED ISSUE #37444: Suggesting `main()` in incorrect contexts
**Description:**
Screenshot 2022-08-18 at 15 14 32

Screenshot 2022-08-18 at 15 01 19
 

Main method should not be suggested in these scenarios.

**Steps to reproduce:**
```
int a = 10;
<cur>

public function main() {

}
```

**Affected Versions:**
2201.2.0

gayalkuruppu:
## Purpose

1. **Fix module level completions before and after module level declarations and definitions**. (_Cause - We get incorrect completions or no completions after some module level declarations, due to parser identifying those cursor positions not as modulePartNode but as the nearest module level node._)

2. **Deprioritise completion item `main()` function call in the completion list**. ( _Cause - In completions, we prioritize function calls and we do not need the `main()` in the top._ )

3. **Fix completions for qualified name references inside function call expression**. ( _Cause - Implementation needed an improvement_)

Fixes #37444

## Approach

1. Route completions to parent when the cursor position is incorrectly identified as a module level decl/ defn.
2. Make the sort text of function call completion item of `main()` in default sorting
4. Sort qualified name reference completions in function call expression

## Samples

https://user-images.githubusercontent.com/46120162/188779570-add8b571-10b9-4e48-9069-c20b159b2127.mov

## Remarks
1. Future - fix the order to get the service templates after the snippet completion items and give priority to the basic types. Also do not provide `import` keyword after some other module declaration. Created this to track this https://github.com/ballerina-platform/ballerina-lang/issues/37661

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


hasithaa:
What is the new order? 

image

I would suggest sorting it in the following order. Here I have prioritized end users over library developers. 

- main function (hide this if the main function already exits)
- service
- function 
- variable
- configurable
- both open and close record type definition template
- array type definition template. 
- class 
- enum 
- Other top-level constructs.

hasithaa:
> > What is the new order?
> > image
> > I would suggest sorting it in the following order. Here I have prioritized end users over library developers.
> > 
> > * main function (hide this if the main function already exits)
> > * service
> > * function
> > * variable
> > * configurable
> > * both open and close record type definition template
> > * array type definition template.
> > * class
> > * enum
> > * Other top-level constructs.
> 
> @hasithaa I have fixed 3 problems mentioned in the Purpose/Approach and created this new issue to fix the order of the completions #37661. Created this issue separately due to the refactoring needed. (BTW, the deprioritization has been done for the `main()` function call but not the `public main function` snippet)

Thanks, will point out my comment in the new issue.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ServiceDeclarationNodeContext.java
```diff
@@ -186,7 +186,8 @@ public boolean onPreValidation(BallerinaCompletionContext context, ServiceDeclar
         int cursor = context.getCursorPositionInTree();
         Token serviceKeyword = node.serviceKeyword();
 
-        return serviceKeyword.textRange().endOffset() < cursor;
+        return serviceKeyword.textRange().endOffset() < cursor
+                && cursor <= node.closeBraceToken().textRange().endOffset();
```
malinthar, line 190: Let's check if the closeBraceToken is not missing

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/TypeDefinitionNodeContext.java
```diff
@@ -114,10 +113,10 @@ private boolean onTypeNameContext(BallerinaCompletionContext context, TypeDefini
 
     @Override
     public boolean onPreValidation(BallerinaCompletionContext context, TypeDefinitionNode node) {
-        TextRange typeKWRange = node.typeKeyword().textRange();
-        int cursorPosition = context.getCursorPositionInTree();
+        int cursor = context.getCursorPositionInTree();
 
-        return typeKWRange.endOffset() < cursorPosition;
+        return node.typeKeyword().textRange().endOffset() < cursor
+                && cursor <= node.semicolonToken().textRange().endOffset();
```
malinthar, line 119: ditto

malinthar (changes requested):
Changes looks good. However, we need to address the following as well. Here, the initializer of the  testClient is not complete. Hence, should we consider this cursor position a part of the initializer. If so, we should not provide completions items like "annotation" , etc snippets. I think it resolves to the module context completion  provider. WDYT @IMS94 @mohanvive @TharushiJay 

image

gayalkuruppu:
> Changes looks good. However, we need to address the following as well. Here, the initializer of the testClient is not complete. Hence, should we consider this cursor position a part of the initializer. If so, we should not provide completions items like "annotation" , etc snippets. I think it resolves to the module context completion provider. WDYT @gayalkuruppu @IMS94 @mohanvive
> 
> image

@malinthar  The scope of the issue was to fix completions when a module part declarations are complete. check https://github.com/ballerina-platform/ballerina-lang/issues/37444#issuecomment-1228201961 If we are considering all these scenarios, there are many more cases to fix than this.

gayalkuruppu:
> Changes looks good. However, we need to address the following as well. Here, the initializer of the testClient is not complete. Hence, should we consider this cursor position a part of the initializer. If so, we should not provide completions items like "annotation" , etc snippets. I think it resolves to the module context completion provider. WDYT @gayalkuruppu @IMS94 @mohanvive
> 
> image

Are you suggesting to route to a child at the completion router? (here this case is identified as ModulePartNode and we only route to the parent in the current completion router implmentation)

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ConstantDeclarationNodeContext.java
```diff
@@ -100,7 +100,8 @@ public boolean onPreValidation(BallerinaCompletionContext context, ConstantDecla
         int cursor = context.getCursorPositionInTree();
         Token constKeyword = node.constKeyword();
 
-        return !constKeyword.isMissing() && cursor > constKeyword.textRange().endOffset();
+        return !constKeyword.isMissing() && cursor > constKeyword.textRange().endOffset()
+                && cursor <= node.semicolonToken().textRange().endOffset();
```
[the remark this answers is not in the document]
gayalkuruppu: In the ``semicolonToken().isMissing` case, `cursor <= node.semicolonToken().textRange().endOffset()` is always true.

malinthar (approved):
LGTM!

LakshanWeerasinghe (commented):
We also can add declaration of a `configurable` and  a `template` to the test case.

IMS94:
Closing this since there are separate PRs #38241 and #38045 send to master and 2201.3.x

PR #37677, opened 2022-09-07
Fix rename popup in create variable and check error code action

### LINKED ISSUE #37609: [Bug]: Rename popup is not shown with the `Create variable and check error' code action
### Description

Consider the following scenario where we have `string` as the return type. 

https://user-images.githubusercontent.com/61020198/188063400-9997f923-49b1-47cc-be09-810f3c3642dc.mov

When we apply the `Create variable and check error` code action, the return type descriptor is being replaced with `returns string|error` instead of adding `|error` at the end of the return statement. Therefore, the start position calculated by `addRenamePopup()` is not accurate because the length of `returns string` is considered twice.

### Steps to Reproduce

```
function createIntWithError() returns int|error {
    return 0;
}

class MyClass {
    function testFunction() returns string {
        createIntWithError();
    }
}
```

mindula:
## Purpose
This PR adds the missing rename popup to create variable and check error code action when there's a return type.

Fixes #37609

## Samples

https://user-images.githubusercontent.com/61020198/188876325-616593ff-7afb-4fa5-bfab-298a690669f8.mov

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
@@ -206,15 +206,21 @@ public class CreateVariableCodeAction implements DiagnosticBasedCodeActionProvid
         2. If any text edits are applied before the variable creation edit, the length of those edits is added to the
            "sum". In the above example, the length of "error?" will be added.
         3. renameOffset gives the length of the variable type and the white space between the variable type and the
-           variable. In the example the renameOffset will be the length of "int ".
+           variable. In the example, the renameOffset will be the length of "int ".
         */
 
         int startPos = CommonUtil.getTextEdit(syntaxTree.get(), variableEdit).range().startOffset();
         int sum = 0;
         for (TextEdit textEdit : textEdits) {
             io.ballerina.tools.text.TextEdit edits = CommonUtil.getTextEdit(syntaxTree.get(), textEdit);
-            if (edits.range().startOffset() < startPos) {
+            int startOffset = edits.range().startOffset();
+            int endOffset = edits.range().endOffset();
+            if (startOffset < startPos) {
                 sum = sum + edits.text().length();
+                if (startOffset < endOffset) {
+                    int returnTypeLength = endOffset - startOffset;
+                    sum = sum - returnTypeLength;
+                }
             }
         }
 


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
```diff
@@ -206,15 +206,21 @@ function testFunc() returns error? {
...
             io.ballerina.tools.text.TextEdit edits = CommonUtil.getTextEdit(syntaxTree.get(), textEdit);
-            if (edits.range().startOffset() < startPos) {
+            int startOffset = edits.range().startOffset();
+            int endOffset = edits.range().endOffset();
+            if (startOffset < startPos) {
                 sum = sum + edits.text().length();
+                if (startOffset < endOffset) {
+                    int returnTypeLength = endOffset - startOffset;
+                    sum = sum - returnTypeLength;
+                }
```
malinthar, line 223:
```suggestion
                sum = sum + startOffset < endOffset ? edits.text().length() : edits.text().length() - endOffset - startOffset;
```

malinthar (changes requested):
Change look good! Let's add a test case covering this

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
```diff
@@ -206,15 +206,21 @@ function testFunc() returns error? {
         2. If any text edits are applied before the variable creation edit, the length of those edits is added to the
            "sum". In the above example, the length of "error?" will be added.
         3. renameOffset gives the length of the variable type and the white space between the variable type and the
-           variable. In the example the renameOffset will be the length of "int ".
+           variable. In the example, the renameOffset will be the length of "int ".
         */
```
LakshanWeerasinghe, line 211: You change the access modifier of this method to `protected`.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/codeaction/providers/createvar/CreateVariableCodeAction.java
gayalkuruppu: We can also extract `edits.text().length()` to a variable, so we don't repeat the same line again(performance wise no impact, but will improve readability)

gayalkuruppu (commented):
The change looks good. Better if we can limit the commit message length to 50 characters.

malinthar (approved):
LGTM!

PR #37790, opened 2022-09-19
Add new field to display annotation

### LINKED ISSUE #37209: Identify special field types in the ballerina records for UI generation
**Description:**
`ballerina/http` `ballerinax/mysql` clients initialization may require some sensitive data such as `clientId`, `clientSecret`, `password`, `path`, etc. The type of these fields defined as strings. 
```
public type RefreshTokenGrantConfig record {|
    string refreshUrl;
    string refreshToken;
    string clientId;
    string clientSecret;
```

```
public type ClientSecureSocket record {|
    boolean enable = true;
    crypto:TrustStore|string cert?;
    crypto:KeyStore|CertKey key?;
```
```
public type KeyStore record {|
    string path;
    string password;
|};
```
 
**Describe your problem(s)**
When we want to generate UI controls for entering values for these fields, the possible control is a text box. Is it possible to introduce an annotation pattern to describe the data? For example, clientSecret should mask value, and the path should be a file upload, etc. @shafreenAnfar @daneshk thoughts?

**Describe your solution(s)**

KavinduZoysa:
## Purpose

Fixes #37209

### DIFF
--- a/langlib/lang.annotations/src/main/ballerina/annotations.bal
+++ b/langlib/lang.annotations/src/main/ballerina/annotations.bal
@@ -99,6 +99,8 @@ public const annotation record {
     string label;
     # icon path relative to module's resource directory
     string iconPath?;
+    # kind of the data
+    "text"|"password"|"file" kind?;
 } display on source type, source class,
       source function, source return, source parameter, source field, source listener,
       source var, source const, source annotation, source service;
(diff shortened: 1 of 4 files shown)


langlib/lang.annotations/src/main/ballerina/annotations.bal
```diff
@@ -99,6 +99,8 @@ public const annotation record {
     string label;
     # icon path relative to module's resource directory
     string iconPath?;
+    # kind of the data
+    "text"|"password"|"file" kind?;
```
gimantha, line 103: From original issue,


> 
> The default would be `"text"` when there is no `@display` annotation for the given field.
> 

@hasithaa @indikasampath2000 @KavinduZoysa Does this mean, the default value `text` will be handled from UI side? or should it be a default value of the `display` record itself?

indikasampath2000: We can handle this from UI as well. But I think better to have a default value to `text` as you have mentioned.

```
 "text"|"password"|"file" kind = "text";
```

@hevayo what do you think?

hasithaa: No, we shouldn't have a default value here. For example, It doesn't make sense when a @display annotation on a service declaration with `kind = "text"`.

gimantha (approved):
LGTM

PR #37811, opened 2022-09-20
Fix FP value `===` expression for functions with parameters

### LINKED ISSUE #37379: Function pointers are not equal when the same pointer is passed as two different parameters

HindujaB:
## Purpose
$subject

Fixes #37379 

## Approach
The issue happens due to an internal call to the `equals` method for `Parameter` instances which returns false.
Overriding the `equals()` resolves the error

## Samples
```ballerina
import ballerina/io;

function func_with_param(string param) {
}

function funcParam = func_with_param;

public function main() {
    io:println(funcParam === funcParam); // returns false - should be true
}
```

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/Parameter.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/Parameter.java
@@ -18,8 +18,6 @@
 
 package io.ballerina.runtime.api.types;
 
-import java.util.Objects;
-
 /**
  * {@code {@link Parameter } represents the parameter of a function in ballerina.
  *
@@ -48,11 +46,6 @@ public class Parameter {
 
         Parameter that = (Parameter) o;
         return this.name.equals(that.name) && this.type.equals(that.type) && this.isDefault == that.isDefault;
-    }
-
-    @Override
-    public int hashCode() {
-        return Objects.hash(name, type, isDefault);
     }
 
 }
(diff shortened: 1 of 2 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/types/Parameter.java
```diff
@@ -33,4 +35,24 @@ public Parameter(String name, Boolean isDefault, Type type) {
...
+            return false;
+        }
+
+        Parameter that = (Parameter) o;
+        return this.name.equals(that.name) && this.type.equals(that.type) && this.isDefault == that.isDefault;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, type, isDefault);
```
warunalakshitha, line 55: Shall we move to constructor

Nadeeshan96:
LGTM

Nadeeshan96 (approved):

warunalakshitha (approved):
LGTM

PR #37852, opened 2022-09-22
Add --native flag for Build Command

### LINKED ISSUE #37608: GraalVM Native Image Compilation for Build Command
### Description

Improve the build command with a new flag `--native` to the `build` command that takes the built jar file and uses the GraalVM's Native Image command to generate a lightweight counterpart.

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Labels

DevTools

_No response_

_No response_

_No response_

Ibaqu:
Fixes #37608

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
@@ -190,6 +190,11 @@ public class BuildCommand implements BLauncherCmd {
             sticky = false;
         }
 
+        if (enableNativeImage) {
+            this.outStream.println(
+                    "WARNING : This is an experimental feature, which supports only a limited set of functionality.");
+        }
+
         // load project
         Project project;
         BuildOptions buildOptions = constructBuildOptions();
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
@@ -22,6 +22,7 @@ import io.ballerina.cli.TaskExecutor;
 import io.ballerina.cli.task.CleanTargetDirTask;
 import io.ballerina.cli.task.CompileTask;
 import io.ballerina.cli.task.CreateExecutableTask;
+import io.ballerina.cli.task.CreateNativeImageTask;
 import io.ballerina.cli.task.DumpBuildTimeTask;
 import io.ballerina.cli.task.ResolveMavenDependenciesTask;
 import io.ballerina.cli.utils.BuildTime;
@@ -263,7 +264,8 @@ public class BuildCommand implements BLauncherCmd {
                 .addTask(new ResolveMavenDependenciesTask(outStream))
                 // compile the modules
                 .addTask(new CompileTask(outStream, errStream, false, isPackageModified, buildOptions.enableCache()))
-                .addTask(new CreateExecutableTask(outStream, this.output, enableNativeImage))
+                .addTask(new CreateExecutableTask(outStream, this.output))
+                .addTask(new CreateNativeImageTask(outStream, this.output), !enableNativeImage)
                 .addTask(new DumpBuildTimeTask(outStream), !project.buildOptions().dumpBuildTime())
                 .build();
 
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
@@ -192,7 +192,7 @@ public class BuildCommand implements BLauncherCmd {
 
         if (enableNativeImage) {
             this.outStream.println(
-                    "WARNING : This is an experimental feature, which supports only a limited set of functionality.");
+                    "WARNING : Native image generation is an experimental feature, which supports only a limited set of functionality.");
         }
 
         // load project
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
@@ -163,8 +163,8 @@ public class BuildCommand implements BLauncherCmd {
     @CommandLine.Option(names = "--enable-cache", description = "enable caches for the compilation", hidden = true)
     private Boolean enableCache;
 
-    @CommandLine.Option(names = "--native-image")
-    private Boolean enableNativeImage;
+    @CommandLine.Option(names = "--native")
+    private boolean enableNativeImage;
 
     public void execute() {
         long start = 0;
@@ -176,10 +176,6 @@ public class BuildCommand implements BLauncherCmd {
 
         if (sticky == null) {
             sticky = false;
-        }
-
-        if (enableNativeImage == null) {
-            enableNativeImage = false;
         }
 
         // load project
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
@@ -102,6 +102,17 @@ public class BuildCommand implements BLauncherCmd {
         this.offline = true;
     }
 
+    BuildCommand(Path projectPath, PrintStream outStream, PrintStream errStream, boolean exitWhenFinish,
+                 boolean dumpBuildTime, boolean enableNativeImage) {
+        this.projectPath = projectPath;
+        this.outStream = outStream;
+        this.errStream = errStream;
+        this.exitWhenFinish = exitWhenFinish;
+        this.dumpBuildTime = dumpBuildTime;
+        this.offline = true;
+        this.enableNativeImage = enableNativeImage;
+    }
+
     @CommandLine.Option(names = {"--output", "-o"}, description = "Write the output to the given file. The provided " +
                                                                   "output file name may or may not contain the " +
                                                                   "'.jar' extension.")
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
@@ -163,6 +163,9 @@ public class BuildCommand implements BLauncherCmd {
     @CommandLine.Option(names = "--enable-cache", description = "enable caches for the compilation", hidden = true)
     private Boolean enableCache;
 
+    @CommandLine.Option(names = "--native-image")
+    private Boolean enableNativeImage;
+
     public void execute() {
         long start = 0;
         if (this.helpFlag) {
@@ -173,6 +176,10 @@ public class BuildCommand implements BLauncherCmd {
 
         if (sticky == null) {
             sticky = false;
+        }
+
+        if (enableNativeImage == null) {
+            enableNativeImage = false;
         }
 
         // load project
@@ -249,7 +256,7 @@ public class BuildCommand implements BLauncherCmd {
                 .addTask(new ResolveMavenDependenciesTask(outStream))
                 // compile the modules
                 .addTask(new CompileTask(outStream, errStream, false, isPackageModified, buildOptions.enableCache()))
-                .addTask(new CreateExecutableTask(outStream, this.output))
+                .addTask(new CreateExecutableTask(outStream, this.output, enableNativeImage))
                 .addTask(new DumpBuildTimeTask(outStream), !project.buildOptions().dumpBuildTime())
                 .build();
 
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
@@ -191,8 +191,8 @@ public class BuildCommand implements BLauncherCmd {
         }
 
         if (enableNativeImage) {
-            this.outStream.println(
-                    "WARNING : Native image generation is an experimental feature, which supports only a limited set of functionality.");
+            this.outStream.println("WARNING : Native image generation is an experimental feature, " +
+                            "which supports only a limited set of functionality.");
         }
 
         // load project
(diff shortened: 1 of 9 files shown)


cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
```diff
@@ -163,6 +163,9 @@ public BuildCommand() {
     @CommandLine.Option(names = "--enable-cache", description = "enable caches for the compilation", hidden = true)
     private Boolean enableCache;
 
+    @CommandLine.Option(names = "--native-image")
```
azinneera, line 166: We agreed to make it `--native`

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
```diff
@@ -163,6 +163,9 @@ public BuildCommand() {
...
+    private Boolean enableNativeImage;
```
azinneera, line 167: We can make this simple `boolean`

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateExecutableTask.java
```diff
@@ -95,6 +97,19 @@ public void execute(Project project) {
...
+
+            if (enableNativeImage) {
+                String[] command = {"native-image", "-jar", executablePath.toString()};
+                ProcessBuilder builder = new ProcessBuilder();
+                builder.command(command);
+                builder.inheritIO();
+                Process process = builder.start();
+
+                if (process.waitFor() != 0) {
+                    out.println("\t Native-image generation failed");
```
azinneera, line 109: Shall we make this more explanatory? e.g. `GraalVM native image generation failed.`

azinneera: Can we print the output of the native-image command?

azinneera:
Let's add a test case to check the functionality of the native image.

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
Ibaqu: Changed to `boolean`

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
Ibaqu: Changed to `--native`

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateExecutableTask.java
Ibaqu: The output of the graalvm native image command shows the progress and in the end I have added a section that says  `GraalVM image generated`

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateExecutableTask.java
```diff
@@ -95,6 +101,31 @@ public void execute(Project project) {
                 start = System.currentTimeMillis();
             }
             jBallerinaBackend.emit(JBallerinaBackend.OutputType.EXEC, executablePath);
+
+            if (enableNativeImage) {
+                // Create nativeConfig directory
+                if (!Files.exists(target.path().resolve("native"))) {
```
azinneera, line 107: We can set this to a variable and reuse  in L108, L111
`target.path().resolve("native")`

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateExecutableTask.java
```diff
@@ -95,6 +101,31 @@ public void execute(Project project) {
                 start = System.currentTimeMillis();
             }
             jBallerinaBackend.emit(JBallerinaBackend.OutputType.EXEC, executablePath);
+
+            if (enableNativeImage) {
```
azinneera, line 105: We should also check if c2c is enabled here as well. Since the c2c plugin will generate it, we do not have to bother about it. @xlight05 please confirm.

xlight05: Depends, According to the original approach of the spec issue you created, We need to check the --skip-executables flag. We need to finalize the spec issue to be precise :/

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateExecutableTask.java
```diff
@@ -95,6 +101,31 @@ public void execute(Project project) {
...
+                // Create nativeConfig directory
+                if (!Files.exists(target.path().resolve("native"))) {
+                    Files.createDirectory(target.path().resolve("native"));
+                }
+
+                String nativeImageDir = "target" + File.separator + "native" + File.separator;
+                String nativeImageName = project.currentPackage().packageName().toString();
+                String[] command = {"native-image", "-jar", executablePath.toString(),
+                        "." + File.separator + nativeImageDir + nativeImageName};
```
xlight05, line 114: Don't we need to add more flags to support ballerina in all the cases? @TharmiganK  @warunalakshitha

TharmiganK: Can we add `--no-fallback` as the default option? In that case if the native image tool cannot create a standalone, then it will give an error
But [native-image tool has not done that due to some legal reasons](https://github.com/oracle/graal/issues/2648), not sure what are those.

In terms of other flags, they will be automatically added through ballerina runtime and standard library modules

azinneera:
> If a user is using `bal build --native` command, is he allowed to pass any additional native-image specific flags?

@TharmiganK no, but is there a requirement as such?

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateExecutableTask.java
Ibaqu: Added `--no-fallback` as a default option

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateNativeImageTask.java
```diff
@@ -0,0 +1,131 @@
...
+            if (!Files.exists(nativeDirectoryPath)) {
+                Files.createDirectory(nativeDirectoryPath);
+            }
+
+            String[] command = {
+                    "native-image",
+                    "-jar",
+                    executablePath.toString(),
+                    nativeDirectoryPath + File.separator + nativeImageName,
+                    "-H:MaxDuplicationFactor=4.0",
```
TharmiganK, line 82: We don't need to add this flag, since this will be added by ballerina runtime

TharmiganK:
> > If a user is using `bal build --native` command, is he allowed to pass any additional native-image specific flags?
> 
> @TharmiganK no, but is there a requirement as such?

@azinneera I have the following concerns :
1. Lets say if the user wants to add an external java dependency then he cannot pass the additional configurations needed to build the native image. But this is not a blocker since they could embed those configuration inside the external jar and run `bal build --native`
2. This will disable users to trace class initializations when there are any issues when building the image
3. Sometimes depends on the machine native image build may fails with out of memory, to resolve this native image tool supports to pass JVM options as well as some special memory related flags in the command. So those will not be supported.
4. We are configuring `MaxDuplicationFactor` flag for native image in the ballerina runtime, this value is set to 40. But depends on the sample this value can be smaller or larger. When the value is small the build time will be reduced significantly. So if we don't allow any configs then user cannot overwrite this

All of this is resolved by building a jar file and use the native-image command, but the purpose of supporting `bal build --native` is to make this process simple to the user right?

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateNativeImageTask.java
```diff
@@ -0,0 +1,138 @@
+package io.ballerina.cli.task;
```
azinneera, line 1: Missing license header

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
```diff
@@ -175,6 +190,11 @@ public void execute() {
             sticky = false;
         }
 
+        if (enableNativeImage) {
+            this.outStream.println(
+                    "WARNING : This is an experimental feature, which supports only a limited set of functionality.");
```
azinneera, line 195:
```suggestion
                    "WARNING : Native image generation is an experimental feature, which supports only a limited set of functionality.");
```

cli/ballerina-cli/src/main/java/io/ballerina/cli/cmd/BuildCommand.java
```diff
@@ -175,6 +190,11 @@ public void execute() {
             sticky = false;
         }
 
+        if (enableNativeImage) {
+            this.outStream.println(
+                    "WARNING : Native image generation is an experimental feature, which supports only a limited set of functionality.");
```
xlight05, line 195: @azinneera  should we mention the fact that the functionality could change in further releases?

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateNativeImageTask.java
```diff
@@ -0,0 +1,156 @@
...
+            // Create native directory
+            if (!Files.exists(nativeDirectoryPath)) {
+                Files.createDirectory(nativeDirectoryPath);
+            }
+
+            String[] command = {
+                    "native-image",
+                    "-jar",
+                    executablePath.toString(),
+                    nativeDirectoryPath + File.separator + nativeImageName,
```
TharmiganK, line 106: Without specifying the output path we can use the following two flags :
`-H:Name=<FileName>`: the name of the executable file that is generated(Default is the name of the jar file).
`-H:Path=<FileSystemPath>`: the directory where the generated executable is placed(If the directory does not exit then this will create the directory).

So we can only add `-H:Path=nativeDirectoryPath` and skip creating the directory. (Assuming the jar file name is same as the package name)

Ibaqu: Changed as suggested

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateNativeImageTask.java
Ibaqu: Added missing license header

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateNativeImageTask.java
Ibaqu: Removed max duplication factor

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateNativeImageTask.java
```diff
@@ -0,0 +1,151 @@
...
+
+import static io.ballerina.cli.launcher.LauncherUtils.createLauncherException;
+import static io.ballerina.cli.utils.FileUtils.getFileNameWithoutExtension;
+import static io.ballerina.projects.util.ProjectConstants.BLANG_COMPILED_JAR_EXT;
+import static io.ballerina.projects.util.ProjectConstants.USER_DIR;
+
+/**
+ * Task for creating the native image file.
+ *
+ * @since 2.0.0
```
azinneera, line 43:
```suggestion
 * @since 2201.3.0
```

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateNativeImageTask.java
```diff
@@ -0,0 +1,151 @@
...
+        this.out = out;
+        if (output != null) {
+            this.output = Paths.get(output);
+        }
+    }
+
+    @Override
+    public void execute(Project project) {
+        this.out.println();
+        this.out.println("Generating executable with native-image");
```
azinneera, line 60:
```suggestion
        this.out.println("Generating GraalVM native image");
```

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/BuildNativeImageCommandTest.java
```diff
@@ -0,0 +1,84 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Objects;
+
+import static io.ballerina.projects.util.ProjectConstants.DIST_CACHE_DIRECTORY;
+
+/**
+ * Build native command tests.
+ *
+ * @since 2.0.0
```
azinneera, line 25:
```suggestion
 * @since 2201.3.0
```

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/BuildNativeImageCommandTest.java
```diff
@@ -0,0 +1,84 @@
+package io.ballerina.cli.cmd;
```
azinneera, line 1: Missing license header

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/CreateNativeImageTask.java
Ibaqu: Changed as suggested

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/BuildNativeImageCommandTest.java
Ibaqu: Added license header

azinneera (approved):

PR #37923, opened 2022-09-28
[Master] Resolve union types with `return wait`

### LINKED ISSUE #37797: [Bug]: Union types are not resolved properly with `return wait`
### Description

When we have multiple wait and return a union type using `return wait`, compiler is giving an error saying that the incompatible types.

### Steps to Reproduce

Run the following ballerina code : 

```ballerina
import ballerina/http;
import ballerina/io;

type Result record {
    string|error a;
    string|error b;
};

function multiFetch(string url, string pathA, string pathB) returns Result|error {
    final http:Client cl = check new (url);
    worker WA returns string|error {
        return fetch(cl, pathA);
    }
    worker WB returns string|error {
        return fetch(cl, pathB);
    }

    return wait {a: WA, b: WB};
}

    Result|error res = multiFetch("https://postman-echo.com", "/get?lang=ballerina",
                                  "/get?greeting=hello");
    io:println(res);
    return;
}

function fetch(http:Client cl, string path) returns string|error {
    map<json> payload = check cl->get(path);
    return payload["args"].toString();
}
```

Compiler error :
```console
ERROR [main.bal:(18:17,18:31)] incompatible types: expected '(balHack/problem_3_4:1.0.0:Result|error)', 
found 'record {| (string|error) a; (string|error) b; |}'
```

### Affected Version(s)

Ballerina SwanLake 2201.2.0

_No response_

### Related area

Compilation

_No response_

_No response_

_No response_

ushirask:
Fixes #37797

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/workers/worker-in-function-test.bal
```diff
@@ -26,3 +26,28 @@ function testSimpleWorkerVM(string msg) returns string {
...
+            return fetch("pathB");
+    }
+
+    return wait {a: WA, b:WB};
+}
+
+public function testReturnWaitForAll() returns string|error {
+    Result res = check multiFetch("pathA", "pathB");
+    return res.a;
+}
```
SasinduDilshara, line 53: In the related issue, the sample code contain 
```
type Result record {
    string|error a;
    string|error b;
};
```
and return type of fetch function is `string|error`.
Workers also return `string|error`.
Shall we add a similar case as a test.

KavinduZoysa (dismissed):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -4544,6 +4544,28 @@ private void setResultTypeForWaitForAllExpr(BLangWaitForAllExpr waitForAllExpr,
...
+                    data.expType = memberType;
+                    this.dlog.resetErrorCount();
+                    setResultTypeForWaitForAllExpr(waitForAllExpr, memberType, data);
+                    if (this.dlog.errorCount() == 0) {
+                        this.dlog.setErrorCount(prevErrorCount);
+                        this.dlog.unmute();
+                        data.expType = expType;
+                        return;
+                    }
+                }
```
MaryamZi, line 4561: This doesn't handle ambiguities, does it?

E.g.,

```ballerina
type Result record {
    string|error a;
    string|error b;
};

type Result2 record {
    any|error a;
    any|error b;
};

public function main() {
    
    worker WA returns string|error {
        return "";
    }

    worker WB returns string|error {
        return "";
    }

    Result2|Result x = wait {a: WA, b: WB};
}
```

Static type used for the created record value depends on what is matched first, which is not how we construct values. If there's more than one type that can be used, we log an error.

MaryamZi: Did we look at what the spec says for wait with unions?

https://ballerina.io/spec/lang/master/#multiple-wait-action says

> The static type of the multiple-wait-action is a record where the static type of each field is the eventual type of the corresponding wait-future-expr.

It doesn't seem to talk about using the contextually-expected type. So we probably should be creating a new record type for the wait-action, even for the non-union case, and then just type check it against the expected type. This will fix the original issue too, since the record type we create will be a subtype of the expected union type.

We can maybe raise this in the spec and get it clarified, because I believe the following will no longer work.

```ballerina
type Result record {
    any|error x;
};

public function main() {
    worker W returns string|error {
        return "hello";
    }

    Result res = wait {x: W};

    // Doesn't panic at runtime in the current impl., since we use `Result` as the type of the constructed mapping value.
    // But if the type is `record {| string|error x; |}`, this will panic.
    res.x = 1;
}
```

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/worker/WorkerInFunctionTest.java
```diff
@@ -18,33 +18,60 @@
...
+        Object returns = BRunUtil.invoke(result, "testReturnWaitForAll1");
+        Assert.assertTrue(returns instanceof BmpStringValue);
+        Assert.assertEquals(returns.toString(), "pathA");
+    }
+
+    @Test()
+    public void testReturnUnionWithWaitForAll2() {
+        Object returns = BRunUtil.invoke(result, "testReturnWaitForAll2");
+        Assert.assertTrue(returns instanceof BmpStringValue);
+        Assert.assertEquals(returns.toString(), "pathA");
```
MaryamZi, line 70: Let's do these assertions in Ballerina itself.

tests/jballerina-unit-test/src/test/resources/test-src/workers/worker-in-function-test.bal
```diff
@@ -26,3 +26,54 @@ function testSimpleWorkerVM(string msg) returns string {
...
+    return path;
+}
+
+function multiFetch1(string pathA, string pathB) returns Result|error {
+    worker WA returns string {
+        return fetch(pathA);
+    }
+
+    worker WB returns string {
+            return fetch("pathB");
```
MaryamZi, line 53: Incorrect indentation.

MaryamZi (changes requested):

KavinduZoysa:
@ushirask, please check the test failure.

PR #37648, opened 2022-09-06
[2201.2.x] Fix suggesting completions with invalid prefix for already imported packages

mindula:
## Purpose
This PR fixes suggesting completions with invalid module prefix for already imported packages.

Fixes #37536

## Samples

https://user-images.githubusercontent.com/61020198/188563623-29624f5f-f868-472c-9a3a-f86b84d6cdc0.mov

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #37967, opened 2022-09-29
[Master] Fix suggesting completions with invalid prefix for already imported packages

mindula:
PR for Master - #37648


dulajdilshan (approved):
LGTM

PR #38050, opened 2022-10-04
[2201.2.x] Update pos of resource path segments

### LINKED ISSUE #37977: [Bug]: Variable and resource-access-action location are same in `simpleVar->/foo/bar[2](10, 11);`
### Description

```ballerina
type Person record {|
   int val1;
   int val2;
|};

client class Foo {
   resource function post foo/bar/[string id](int q1, int q2) returns Person|error {
      if q1 > q2 {
         return error("Q1 > Q2");
      }
      return {val1: q1, val2: q2};
   }

   resource function get foo/[string id](int q1, int q2) returns int {
     return q1 + q2;
   }

   resource function post bar/[string... ids](string fullname) returns string {
     return ids[0] + fullname;
   }
}

public function main() {
   Foo fooClient = new();
   Person|error wVal = <1>fooClient->/foo/bar/["3"].post(1, 2);
}
```

In Semantic API, when we try to resolve the cursor position for `fooClient` and resource-access-action node, we get the same location (`<1>`). Because of this, we cannot identify whether we should get the resource-function or the `fooClient`

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

-> Compilation

_No response_

_No response_

_No response_

rdulmina:
## Purpose
Update the starting pos of resource path segments to resource access action forward slash token.
This adds pos information for SemanticAPI to fix #37977

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/parser/BLangNodeBuilder.java
@@ -4243,7 +4243,8 @@ public class BLangNodeBuilder extends NodeTransformer<BLangNode> {
             listConstructorExpr.pos = getPosition(clientResourceAccessActionNode.slashToken());
         } else {
             listConstructorExpr.pos = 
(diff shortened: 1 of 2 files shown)


gimantha (approved):

dulajdilshan (approved):
LGTM

PR #38079, opened 2022-10-05
Minimize the usage of capturedClosureEnv in lamda function

### LINKED ISSUE #38071: [Improvement]: Minimize the usage of `capturedClosureEnv` in lamda function
### Description

In `BLangLambdaFunction` there is an env called `capturedClosureEnv` which holds the cloned env at that point.

### Describe your problem(s)

When assigning the env to `capturedClosureEnv`, we create a new hashmap to store scope entries. For a larger project like http_tests, there are a lot of lambda functions (generated for annotations) and large scope entries. Since this takes a considerable amount of memory this may cause OOM issues.

### Describe your solution(s)

As the solution, we try to minimize the usage of `capturedClosureEnv`. We may not able to totally remove this field, but we need to minimize the usage.

### Related area

-> Compilation

_No response_

_No response_

_No response_

KavinduZoysa:
## Purpose

Fixes #38071


KavinduZoysa:
This is not valid anymore, hence closing.

Issue #36731, opened 2022-06-27
Fix AST node positions to include metada and add another property to keep position without medatadata


lochana-chathura:
**Description:**
$subject. 

Please refer to the discussion at [https://ballerina-platform.slack.com/archives/C9WMA6QDC/p1655971048193769](https://ballerina-platform.slack.com/archives/C9WMA6QDC/p1655971048193769).

KavinduZoysa:
As per the offline disussion had with @hasithaa, we decided that the suggested approach will become an extra overhead to the compiler. In order to fix the original issue mentioned in https://github.com/ballerina-platform/ballerina-lang/issues/36351, a different approach was suggested in https://github.com/ballerina-platform/ballerina-lang/issues/37404.

Hence closing this issue. We will https://github.com/ballerina-platform/ballerina-lang/issues/36351 under #37404

PR #38134, opened 2022-10-10
Call SymbolFinder for annot-attachments in methods

### LINKED ISSUE #38053: [Bug]: BLangNodeBuilder doesn't include the annotation attachment positions into FunctionDefinitionNode position
### Description

This surfaced while using the `SymbolFinder.lookupNodes(List<? extends BLangNode> nodes)` method to lookup a symbol for an annotation attachment.

The `BLangNodeBuilder.transform(FunctionDefinitionNode funcDefNode)` method includes the position without metadata as the position of the function def node. This cause the `withinBlock()` check to fail inside the `SymbolFinder` method mentioned above.

### Steps to Reproduce

```
import ballerina/http;

service / on new http:Listener(9090) {

    @http:ResourceConfig{
        
    }
    resource function get greeting() {

    }
}
```
Try to doto definition of the `ResourceConfig` annotation above.

### Affected Version(s)

2201.3.0-SNAPSHOT
2201.2.0

_No response_

### Related area

-> Other Area

#36351
#37404

_No response_

_No response_

KavinduZoysa:
## Purpose

Fixes #38053

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
@@ -1265,7 +1265,14 @@ class SymbolFinder extends BaseVisitor {
         lookupNodes(classDefinition.referencedFields);
         lookupNode(classDefinition.initFunction);
         lookupNodes(classDefinition.functions);
+        lookupAnnAttachmentsAttachedToMethods(classDefinition.functions);
         lookupNodes(classDefinition.typeRefs);
+    }
+
+    private void lookupAnnAttachmentsAttachedToMethods(List<BLangFunction> functions) {
+        for (BLangFunction func : functions) {
+            lookupNodes(func.annAttachments);
+        }
     }
 
     @Override
--- a/language-server/modules/langserver-core/src/test/java/org/ballerinalang/langserver/definition/DefinitionTest.java
+++ b/language-server/modules/langserver-core/src/test/java/org/ballerinalang/langserver/definition/DefinitionTest.java
@@ -149,8 +149,7 @@ public class DefinitionTest {
                 {"def_record_config1.json", "project"},
                 // TODO Blocked by #30688 causing module of user defined errors to become lang.annotations
                 // {"def_error_config1.json", "project"},
-                // TODO Blocked due to BLangResourceFunction position being set incorrectly
-                // {"def_annotation_on_obj_func_config1.json", "project"}   
+                 {"def_annotation_on_obj_func_config1.json", "project"}
         };
     }
 


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
```diff
@@ -1265,9 +1265,16 @@ public void visit(BLangClassDefinition classDefinition) {
         lookupNodes(classDefinition.referencedFields);
         lookupNode(classDefinition.initFunction);
         lookupNodes(classDefinition.functions);
+        lookupAnnAttachmentsAttachedToMethods(classDefinition.functions);
         lookupNodes(classDefinition.typeRefs);
     }
 
+    private void lookupAnnAttachmentsAttachedToMethods(List<BLangFunction> functions) {
```
IMS94, line 1272: What if we add a `lookupNodes(List<BLangFunction> functions)` instead (which overloads the `lookupNodes()` method)? @KavinduZoysa @dulajdilshan

KavinduZoysa: If we use `lookupNodes(List<BLangFunction> functions)`, it is looking up each function, but what we need to do here is, look up function annotations separately. Therefore we need a separate method like this.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
```diff
@@ -1265,9 +1265,16 @@ public void visit(BLangClassDefinition classDefinition) {
         lookupNodes(classDefinition.referencedFields);
         lookupNode(classDefinition.initFunction);
         lookupNodes(classDefinition.functions);
+        lookupAnnAttachmentsAttachedToMethods(classDefinition.functions);
```
MaryamZi, line 1268: Is this a temporary workaround for the position issue?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
IMS94: Yes, inside `lookupNodes(List<BLangFunction>)` we can lookup only the attachments:
```
for (BLangFunction f : functions) {
    lookupNodes(f.annotationAttachments);
    lookupNode(f);
}
```
WDYT?

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
IMS94: Seems like that

KavinduZoysa: This is a different approach, since extending the function position to have annot-attachments is only required here, I think it is better to handle it here.

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
KavinduZoysa: Yes this logic should work, but I think we cannot have a function like 
```ballerina
    private void lookupNodes(List<BLangFunction> functions) {

    }
```
since it matches the generic function. We may need to change the function name

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
MaryamZi: So this doesn't quite fix https://github.com/ballerina-platform/ballerina-lang/issues/38053, right?

KavinduZoysa: I think this is the correct fix. We faced the same kind of issue in NodeFinder.java. At that time we decided to solve that in NodeFinder.java
Please check https://github.com/ballerina-platform/ballerina-lang/issues/36731#issuecomment-1216448160

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/SymbolFinder.java
IMS94: Got it. No problem then

dulajdilshan:
Closing as we are tracking this under Semantic API team via https://github.com/ballerina-platform/ballerina-lang/pull/38541

PR #38172, opened 2022-10-12
Handle decimal subnormal and underflow cases

### LINKED ISSUE #38082: [Bug]: Subnormals resulted from decimal operations should be treated as ZERO
### Description

The decimal operations result in subnormal values should return the decimal value zero according to the IEEE 754-2008 ANSI X3.274.

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

-> Runtime

_No response_

_No response_

_No response_

gabilang:
Fixes #38082

## Approach
The absolute decimal values smaller than the minimum positive normal decimal value are treated as decimal value zero. 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/multiply-operation.bal
```diff
@@ -1189,6 +1189,20 @@ function testNoShortCircuitingInMultiplicationWithNonNullable() {
...
+    decimal d2 = 1E-143;
+    assertEqual(d1 * d2, 0.0d);
+
+    d1 = -9.999999999999999999999999999999999E-6001;
+    d2 = 1E-143;
+    assertEqual(d1 * d2, 0.0d);
+
+    d1 = -9.999999999999999999999999999999999E-6101;
+    d2 = 1E-76;
+    assertEqual(d1 * d2, 0.0d);
```
Nadeeshan96, line 1203: What should be the precision of the resulting decimal value here? Shall we use `assertEqual` here and use assertTrue with `===` operator here to check the decimal value (Can also add statement with `==` if needed) https://ballerina.io/spec/lang/master/#section_6.29

gabilang: The above operations are result in values which are not belonged to the normal decimal values set (they are decimal subnormals and underflow according to the IEEE-754 Decimal128 spec) where the precision is lost. According to https://github.com/ballerina-platform/ballerina-spec/issues/995 subnormal values also need to be treated as decimal value Zero.

HindujaB (approved):
LGTM

KavinduZoysa:
`decimal d1 = 1.01e-6143d;` Does this case a subnormal? 
With this change, the resut is `1.01E-6143`, but for `decimal d1 = 0.01e-6143d;` it gives zero.

gabilang:
> `decimal d1 = 1.01e-6143d;` Does this case a subnormal? With this change, the resut is `1.01E-6143`, but for `decimal d1 = 0.01e-6143d;` it gives zero.

Since the exponent of minimum representable magnitude of normal decimal value according to the [spec](https://speleotrove.com/decimal/dbover.html) is `-6143`, the value `0.01e-6143` is going to fall in subnormal range. So the value of `0.01e-6143d` need to be treat as zero and `1.01e-6143d` is a normal decimal value.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/DecimalValue.java
```diff
@@ -90,6 +100,17 @@ public DecimalValue(String value, DecimalValueKind valueKind) {
         this.valueKind = valueKind;
     }
 
+    private static BigDecimal getValidDecimalValue(BigDecimal bd) {
+        if (bd.compareTo(DECIMAL_MAX) > 0 || bd.compareTo(DECIMAL_MIN) < 0) {
+            throw ErrorCreator.createError(BallerinaErrorReasons.NUMBER_OVERFLOW,
+                    BLangExceptionHelper.getErrorDetails(RuntimeErrors.DECIMAL_VALUE_OUT_OF_RANGE));
+        } else if (bd.abs(MathContext.DECIMAL128).compareTo(MIN_DECIMAL_MAGNITUDE) < 0 &&
+                bd.abs(MathContext.DECIMAL128).compareTo(BigDecimal.ZERO) > 0) {
```
Nadeeshan96, line 108:
```suggestion
        } else if (bd.abs(MathContext.DECIMAL128).compareTo(MIN_DECIMAL_MAGNITUDE) < 0) {
```
This is redundant right? The method will return the same for any `bd` or `BigDecimal.ZERO`, won't it?

gabilang: This check excludes the decimal value zero (whole set of decimal zero values with different precision) from the set of values covered by the check `bd.abs(MathContext.DECIMAL128).compareTo(MIN_DECIMAL_MAGNITUDE) < 0`.

tests/jballerina-unit-test/src/test/resources/test-src/expressions/binaryoperations/multiply-operation.bal
Nadeeshan96: Agreed to use decimal `0d` as the subnormal decimal zero. Hence marking this conversation as resolved.

Nadeeshan96 (approved):
LGTM

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/main/function/ArgumentParserPositiveTest.java
```diff
@@ -158,10 +158,7 @@ public Object[][] decimalValues() {
         return new Object[][]{
                 {"10", "10"},
 //                {"-10.123", "-10.123"},
```
HindujaB, line 160: We currently don't support negative decimal values as args as they can clash with the short command options. Eg. - `-Ckey=value` 
Shall we move this to negative tests and add a comment as well?
They might provide an `undefined CLI argument` error.

gabilang: Created an issue https://github.com/ballerina-platform/ballerina-lang/issues/38311 earlier for this since there are no opened issues to track the `TODO` item added to this test. So we can add the negative tests with that issue.

PR #38183, opened 2022-10-12
Preserve the determined type of constant value expressions

### LINKED ISSUE #36831: Inconsistency in the behaviour of `typeOf` API
**Description:**

Refer to the screen recording below where I need to get the broad type of the following constant declaration.

https://user-images.githubusercontent.com/27485094/177035208-ec0149a5-3b60-4ef3-891d-bceffc259d9b.mov

When I get the `typeSymbol` of a  `BasicLiteralNode` of value `10`, using `context.currentSemanticModel().get().typeOf(basicLiteralNode)`, I get a `BallerinaIntTypeSymbol`.

Screenshot 2022-06-30 at 14 47 01

However, when I get the `typeSymbol` of a  `BinaryExpressionNode` of value `10+20`, using `context.currentSemanticModel().get().typeOf(binaryExpressionNode)`, I get a `BallerinaSingletonTypeSymbol` of value `30`.

Screenshot 2022-06-30 at 14 19 14

**Affected Versions:**
2201.1.1 (Swan Lake Update 1)

sanjana:
Fixes #36831

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/ballerina-compiler-api-test/src/test/resources/test-src/expressions_test.bal
+++ b/tests/ballerina-compiler-api-test/src/test/resources/test-src/expressions_test.bal
@@ -358,3 +358,6 @@ function fooFn() returns future<int> {
 function barFn() returns int {
     return 10;
 }
+
+const int A = 10;
+const int B = 20 + 30;
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
@@ -649,7 +649,15 @@ public class ConstantValueResolver extends BLangNodeVisitor {
             return;
         }
 
+        // When setting the BType of the expression, the determined type as well would be set to the same BType.
+        // Therefore, if the determined type was already not null, it shall be preserved and overridden with its
+        // previous value.
+        BType determinedType = expr.getDeterminedType();
         expr.setBType(resolvedType);
+        if (determinedType != null) {
+            expr.setDeterminedType(determinedType);
+        }
+
         symbol.type = resolvedType;
     }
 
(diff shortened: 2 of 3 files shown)


tests/ballerina-compiler-api-test/src/test/resources/test-src/expressions_test.bal
```diff
@@ -358,3 +358,6 @@ function fooFn() returns future<int> {
 function barFn() returns int {
     return 10;
 }
+
+const int A = 10;
+const int B = 20 + 30;
```
gimantha, line 363: @sanjana I have a doubt on the original issue (https://github.com/ballerina-platform/ballerina-lang/issues/36831) though.
Say I want to extract the constant. The extracted constant can be either the exact singleton value `30` or it can have the wider basic type node `int`. As of the current implementation, it only allows `int`. But @SandaruJayawardana is doing a complete revamp for constant expressions and both have to be allowed. What should be the user experience for this one?

```bal
const 30 A = 10 + 20; // valid
const int A = 10 + 20; // also valid
```

MaryamZi: The issue is with the expression, right?

I think the behaviour should be consistent for both RHS expressions in 

```ballerina
const int A = 10;
const int B = 20 + 30;
```

If the type of `10` is `int`, it should be the same for `20 + 30` also. But if the type for `20 + 30` is singleton `50`, the type for `10` should also be singleton `10`. Feels like the singleton approach is more accurate, but since this applies to all literals, may result in a lot of new types being created.

For example, even with `int i = 10;`, the type of `10` will then be the singleton. 

I think we need to discuss this and come to a conclusion and then fix across the compiler for all literals.

sanjana: Will be holding this PR until the discussion initiated at #38250 is finalized.

PR #38295, opened 2022-10-19
[master] Refactor DocumentServiceContext visibleSymbols() method and completions in import declaration node

### LINKED ISSUE #37234: VisibleSymbols evaluates only once and then used for any given position in AbstractDocumentServiceContext
**Description:**
Due to $title, for any given position, we get the same set of visibleSymbols.

```
    public List<Symbol> visibleSymbols(Position position) {
        if (this.visibleSymbols == null) { // <--- here
            Optional<SemanticModel> semanticModel;
            if (this.cancelChecker == null) {
                semanticModel = this.workspaceManager.semanticModel(this.filePath);
            } else {
                semanticModel = this.workspaceManager.semanticModel(this.filePath, this.cancelChecker);
            }
            Optional<Document> srcFile = this.workspaceManager.document(filePath);

            if (semanticModel.isEmpty() || srcFile.isEmpty()) {
                return Collections.emptyList();
            }

            this.checkCancelled();
            visibleSymbols = semanticModel.get().visibleSymbols(srcFile.get(),
                    LinePosition.from(position.getLine(),
                            position.getCharacter()), DiagnosticState.VALID, DiagnosticState.REDECLARED);
        }

        return visibleSymbols;
    }
```

**Steps to reproduce:**
#37073

**Affected Versions:**
2201.2.0

### LINKED ISSUE #38085: [Bug]: Importing langlib duplicates the already-typed part of the module name
### Description

https://user-images.githubusercontent.com/25485997/194212644-d8d232c9-1e10-41fb-a68a-4a8ed627b9e8.mov

### Steps to Reproduce

Try completions in the cursor location

```ballerina
import ballerina/lang.<cursor>
```

### Affected Version(s)

2201.2.0 atleast

_No response_

### Related area

-> Editor

_No response_

_No response_

_No response_

IMS94:
## Purpose
$subject

Only the contexts which are subclasses of `PositionedOperationContext` are now caching the visible symbols for the cursor position.

Fixes #37234
Fixes #38085

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/AbstractDocumentServiceContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/AbstractDocumentServiceContext.java
@@ -60,9 +60,7 @@ public class AbstractDocumentServiceContext implements DocumentServiceContext {
     private final String fileUri;
 
     private final WorkspaceManager workspaceManager;
-
-    private List<Symbol> visibleSymbols;
-
+    
     private List<ImportDeclarationNode> currentDocImports;
 
     private Map<ImportDeclarationNode, ModuleSymbol> currentDocImportsMap;
@@ -132,26 +130,23 @@ public class AbstractDocumentServiceContext implements DocumentServiceContext {
 
     @Override
     public List<Symbol> visibleSymbols(Position position) {
-        if (this.visibleSymbols == null) {
-            Optional<SemanticModel> semanticModel;
-            if (this.cancelChecker == null) {
-                semanticModel = this.workspaceManager.semanticModel(this.filePath);
-            } else {
-                semanticModel = this.workspaceManager.semanticModel(this.filePath, this.cancelChecker);
-            }
-            Optional<Document> srcFile = this.workspaceManager.document(filePath);
+        Optional<SemanticModel> semanticModel;
+        if (this.cancelChecker == null) {
+            semanticModel = this.workspaceManager.semanticModel(this.filePath);
+        } else {
+            semanticModel = this.workspaceManager.semanticModel(this.filePath, this.cancelChecker);
+        }
+        Optional<Document> srcFile = this.workspaceManager.document(filePath);
 
-            if (semanticModel.isEmpty() || srcFile.isEmpty()) {
-                return Collections.emptyList();
-            }
-
-            this.checkCancelled();
-            visibleSymbols = semanticModel.get().visibleSymbols(srcFile.get(),
-                    LinePosition.from(position.getLine(),
-                            position.getCharacter()), DiagnosticState.VALID, DiagnosticState.REDECLARED);
+        if (semanticModel.isEmpty() || srcFile.isEmpty()) {
+            return Collections.emptyList();
         }
 
-        return visibleSymbols;
+        this.checkCancelled();
+
+        return semanticModel.get().visibleSymbols(srcFile.get(),
+                LinePosition.from(position.getLine(),
+                        position.getCharacter()), DiagnosticState.VALID, DiagnosticState.REDECLARED);
     }
 
     @Override
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/AbstractDocumentServiceContext.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/AbstractDocumentServiceContext.java
@@ -42,6 +42,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+
 import javax.annotation.Nonnull;
 
 /**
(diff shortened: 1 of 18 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/AbstractDocumentServiceContext.java
```diff
@@ -132,47 +128,30 @@ public LSOperation operation() {
 
     @Override
     public List<Symbol> visibleSymbols(Position position) {
```
malinthar, line 130: Is avoiding repetition the reason to keep this without deprecating?  If so shall we make this method protected?

malinthar (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/AbstractDocumentServiceContext.java
IMS94: Didn't get you?

malinthar: We discussed that we don't want this method to act as an interface for the semantic model. So I was wondering why we are not removing this method? (Is it to avoid repetition of logic to access and retrieve the visible symbols in child classes? ).
Each time this method is invoked, we would be calling the visibleSymbols API of the semantic model. So, we should either cache the visible symbols in the class or make this method protected so that this is not publically available. WDYT?

IMS94: Yes, that's correct. We can/should remove this method. But, since this is an interface, we cannot remove a method. What we can do is first deprecate this method and then remove it in a future release. If we are to deprecate, first we have to deprecate the `visibleSymbols()` method in the topmost interface and then add that to the `PositionedOperationContext`. That's why I decided not to remove it and keep it instead.

malinthar: +1. Let's add the deprecated annotation

malinthar: +1. Let's keep it

malinthar (dismissed):
LGTM!

language-server/modules/langserver-core/src/test/resources/completion/import_decl/source/import_decl_with_no_module_name1.bal
```diff
@@ -0,0 +1 @@
+import test/
```
gayalkuruppu, line 1: We can add a new line here and in the other test files

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/AbstractDocumentServiceContext.java
```diff
@@ -132,47 +127,30 @@ public LSOperation operation() {
...
-
-            if (semanticModel.isEmpty() || srcFile.isEmpty()) {
-                return Collections.emptyList();
-            }
+        Optional<SemanticModel> semanticModel;
+        if (this.cancelChecker == null) {
+            semanticModel = this.workspaceManager.semanticModel(this.filePath);
+        } else {
+            semanticModel = this.workspaceManager.semanticModel(this.filePath, this.cancelChecker);
+        }
```
gayalkuruppu, line 135: We can write a one liner here if suitable

gayalkuruppu (commented):
Fixes look good to me. Have some minor, non critical suggestions.

language-server/modules/langserver-core/src/test/resources/completion/import_decl/source/import_decl_with_no_module_name1.bal
IMS94: Here, not adding a new line is intentional because this will be the state of the source code file when a user is adding an import to an empty bal file.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/contexts/AbstractDocumentServiceContext.java
IMS94: with a ternary condition? `condition ? val1 : val2`

IMS94: IMO this way is better than that

gayalkuruppu (approved):
LGTM!

PR #38318, opened 2022-10-20
Add runtime APIs to escape/unescape characters in identifiers

### LINKED ISSUE #37046: Need an API to Unescape Characters in Identifiers
**Description:**
When the Semantic API returns values of identifiers, the escape characters are not unescaped. For example:

```ballerina
import ballerina/graphql;

service on new graphql:Listener(4000) {
    resource function get name(string 'version) returns string {
        return 'version;
    }
}
```

In the above code, the parameter name of the resource method is returned as `'version` from the `getName()` API. There should be a way to remove the leading single quote and also unescape the escaped characters. 

There is an existing API for this, but it is not exposed to external users. It is better to expose this, as some compiler plugins need this functionality.

gabilang:
Fixes #37046

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/runtime/api/identifier_utils/main.bal
```diff
@@ -0,0 +1,55 @@
...
+    testUnescapeSpecialCharacters();
+}
+
+function testIdentifierDecoding() {
+    test:assertEquals(decodeIdentifier("üňĩćőđę_ƈȏɳʂʈ_IL"), "üňĩćőđę_ƈȏɳʂʈ_IL");
+    test:assertEquals(decodeIdentifier("const_IL_123"), "const_IL_123");
+    test:assertEquals(decodeIdentifier(" $0047$0058@$0091`{~_IL"), " /:@[`{~_IL");
+}
+
+function testEscapeSpecialCharacters() {
```
HindujaB, line 32: Shall we add tests with Unicode codepoints? Eg: `\u{03C0}` etc.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/IdentifierUtils.java
```diff
@@ -40,4 +42,23 @@ public static String decodeIdentifier(String encodedIdentifier) {
...
+    }
+
+    /**
+     * Unescapes a ballerina string.
+     *
+     * @param text ballerina string to unescape
+     * @return unescaped ballerina string
+     */
+    public static String unescapeBallerina(String text) {
+        return unescapeJava(Utils.unescapeUnicodeCodepoints(text));
```
Nadeeshan96, line 62:
```suggestion
        return Utils.unescapeBallerina(text);
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/IdentifierUtils.java
```diff
@@ -40,4 +42,23 @@ public static String decodeIdentifier(String encodedIdentifier) {
...
+     */
+    public static String escapeSpecialCharacters(String identifier) {
+        return Utils.escapeSpecialCharacters(identifier);
+    }
+
+    /**
+     * Unescapes a ballerina string.
+     *
+     * @param text ballerina string to unescape
+     * @return unescaped ballerina string
```
Nadeeshan96, line 59:
```suggestion
     * @return ballerina string with any escaped characters converted to their unescaped form
```
Just wanted to improve the understandability. May change as you wish by adding something like this to the description instead of here.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/IdentifierUtils.java
Nadeeshan96: Then can remove the above static import also.

tests/jballerina-unit-test/src/test/resources/test-src/runtime/api/identifier_utils/main.bal
Nadeeshan96: Also have a small look at https://github.com/ballerina-platform/ballerina-lang/issues/38334 sample 1. May need to fix and add a negative test.

Nadeeshan96 (approved):
LGTM

tests/jballerina-unit-test/src/test/resources/test-src/runtime/api/identifier_utils/main.bal
```diff
@@ -0,0 +1,60 @@
...
+    test:assertEquals(escapeSpecialCharacters("\u{1f479}\u{1F47A}\u{1f47B}\u{1F63A}\u{1F408}\u{1F981}\u{1F600}"), "👹👺👻😺🐈🦁😀");
+}
+
+function testUnescapeSpecialCharacters() {
+    test:assertEquals(unescapeSpecialCharacters("\\'version"), "'version");
+    test:assertEquals(unescapeSpecialCharacters("\\ \\'\\ \\$0047\\$0058\\@\\$0091\\`\\{\\~_IL"), " ' $0047$0058@$0091`{~_IL");
+    test:assertEquals(unescapeSpecialCharacters("\\'ü\\?ňĩ\\$ć\\%őđ\\&ę_ƈ\\-ȏɳ\\+ʂʈ_\\=L\\#"), "'ü?ňĩ$ć%őđ&ę_ƈ-ȏɳ+ʂʈ_=L#");
+    test:assertEquals(unescapeSpecialCharacters("πA9őđ\\&ę©®™\\@3\\)\\^\\~\\`ΣΨ"), "πA9őđ&ę©®™@3)^~`ΣΨ");
+    test:assertEquals(unescapeSpecialCharacters("πA9őđ\\&ę©®™\\@3\\)\\^\\~\\`ΣΨ"), "\u{03C0}A9őđ&ę\u{00A9}\u{00AE}\u{2122}@3)^~`\u{03A3}\u{03A8}");
+    test:assertEquals(unescapeSpecialCharacters("👹👺👻😺🐈🦁😀\u{1f920}👐"), "\u{1f479}\u{1F47A}\u{1f47B}\u{1F63A}\u{1F408}🦁😀🤠\u{1F450}");
```
HindujaB, line 47: Shall we check why `🦁😀` characters didn't get unescaped to `\u{1F981}\u{1F600}` as the other characters? It could be a bug. 
We can create an issue and fix it later.

gabilang: Assertion is succeed we have either `\u{1F981}\u{1F600}` or 🦁😀. Is it a wrong behavior?

gabilang: The above case is not due to a bug. Actually I missed to include some test cases with escape chars. Added some tests with https://github.com/ballerina-platform/ballerina-lang/pull/38318/commits/73efab5391b5670ed6b03b31040c7042599f9e0b.

HindujaB (approved):
LGTM

Issue #37994, opened 2022-09-29
[Bug]: Cannot use list/mapping binding pattern with readonly record/list


lochana-chathura:
### Description

$subject.

### Steps to Reproduce

Consider the following examples.
```bal
type Foo readonly & record {|
    int x;
    string y;
|};

type Bar readonly & [int, string];

public function main() {
    Foo foo = {x: 1, y: "s"};
    Foo {x, y} = foo; // Gives error saying invalid record binding pattern 

    Bar bar = [1, "s"];
    Bar [x, y] = bar; // Gives error saying invalid list binding pattern
}
```

However, when `Foo` is declared with readonly fields instead,
e.g.
```bal
type Foo record {|
    readonly int x;
    readonly string y;
|};
```
the mapping binding pattern works as expected. 

### Affected Version(s)

2201.2.0

### Related area

-> Compilation

lochana-chathura:
The above seems to work with `var`.
```bal
public function main() {
    Foo foo = {x: 1, y: "s"};
    var {x, y} = foo; // works.

    Bar bar = [1, "s"];
    var [x, y] = bar; // works.
}
```

PR #38388, opened 2022-10-25
Fix list and mapping binding patterns with immutable types

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/37994

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


KavinduZoysa:
@MaryamZi, shall we add a test case for error destructuring if it is useful.
ex:
```ballerina
type SampleErrorData record {
    string message?;
    error cause?;
    string info;
    boolean fatal;
};

type SampleError readonly & error<SampleErrorData>;

function getSampleError() returns SampleError {
    SampleError e = error SampleError("Sample Error", info = "Detail Info", fatal = true);
    var message = "";
    var cause = ();
    string info;
    error SampleError (message, cause, info = info) = e;
    return e;
}
```

MaryamZi:
> @MaryamZi, shall we add a test case for error destructuring if it is useful. ex:
> 
> ```ballerina
> type SampleErrorData record {
>     string message?;
>     error cause?;
>     string info;
>     boolean fatal;
> };
> 
> type SampleError readonly & error<SampleErrorData>;
> 
> function getSampleError() returns SampleError {
>     SampleError e = error SampleError("Sample Error", info = "Detail Info", fatal = true);
>     var message = "";
>     var cause = ();
>     string info;
>     error SampleError (message, cause, info = info) = e;
>     return e;
> }
> ```


This PR fixes issues with immutable mapping and lists, added two tests anyway and came across an issue: https://github.com/ballerina-platform/ballerina-lang/issues/38726.

KavinduZoysa (dismissed):

tests/jballerina-unit-test/src/test/resources/test-src/statements/variabledef/tuple-variable-definition.bal
```diff
@@ -261,6 +261,35 @@ function testTupleVarDefWithRestBPContainsErrorBPWithRestBP() {
     assertEquality(true, b[0]);
 }
 
+type ReadOnlyTuple readonly & [int[], string];
```
suleka96, line 264: Shall we also add these kind of tests:
```
type IntArray int[];
type ReadOnlyTuple readonly & [IntArray, string];
```

tests/jballerina-unit-test/src/test/resources/test-src/expressions/varref/record-variable-reference-semantics-negative.bal
```diff
@@ -242,3 +242,25 @@ function testMappingBindingPatternWithMap() {
...
 }
+
+type ReadOnlyRecord readonly & record {|
+    int[] x;
+    string y;
+|};
+
+function testReadOnlyRecordWithMappingBindingPatternInDestructuringAssignmentNegative1() {
+    ReadOnlyRecord f1 = {x: [1, 2], y: "s1"};
+    string[] & readonly x;
```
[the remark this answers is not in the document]
MaryamZi: Added a few positive tests in [75ac0d0](https://github.com/ballerina-platform/ballerina-lang/pull/38388/commits/75ac0d0a627b5fb01a3d26bd3ddf7a7d270eec97).

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java
```diff
@@ -2704,7 +2704,7 @@ boolean symbolEnterAndValidateRecordVariable(BLangRecordVariable var, SymbolEnv
     }
 
     boolean validateRecordVariable(BLangRecordVariable recordVar, SymbolEnv env) {
-        BType recordType = Types.getReferredType(recordVar.getBType());
+        BType recordType = Types.getEffectiveType(Types.getReferredType(recordVar.getBType()));
```
chiranSachintha, line 2707: Why do we perform a check for the effective type without including it as a case for intersection within the switch block?

MaryamZi: Wouldn't that require changes to the method to be able to call it recursively?

chiranSachintha: Yes, we need to call it recursively, but only for intersection types. Here, We perform a check on every type to determine if it is an intersection.
(We also perform this check in other parts of the code. My concern is why we treat intersections as a special case. Is this not a problem for code maintainability?)

MaryamZi: It isn't just intersections, we've also checked for type reference types here. Going forward, with the changes to intersection types, we will get intersection/reference types more often. And the effective type retrieval method will handle intersections, type reference types, and eventually union types in one go. 

Given that it is a check by tag, I don't think it makes much of a difference, especially because we'll have to continue to do this in some form and the union type has specific logic.

chiranSachintha (dismissed):

chiranSachintha (approved):

Issue #38333, opened 2022-10-21
[New Feature]: Need a code snippet suggestion to create a transformer function 


madushajg:
### Description

A new LS suggestion should be provided to create a base expression bodied function.

### Usage
The users will be able start writing data mapping functions by using this suggestion and they can either continue complete the data mapping using the code editor or use the visual data mapper to create the transformation.

### Code snippet to be generated
```ballerina
function transform() => ();
```

IMS94:
We can provide a snippet completion item like below
```
function <name>(<>) => (<>);
```
The `<>` areas will be placeholders (with tab completions).

PR #38394, opened 2022-10-25
[2201.2.x] Add expression bodied function snippet completion item

IMS94:
## Purpose
$subject

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/38333

## Samples
Peek 2022-10-25 23-45

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SnippetGenerator.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SnippetGenerator.java
@@ -465,7 +465,7 @@ public class SnippetGenerator {
      * @return {@link SnippetBlock} Generated snippet
      */
     public static SnippetBlock getExpressionBodiedFunctionDefSnippet() {
-        String snippet = "function ${1:name}(${2})${3} => ${4}";
+        String snippet = "function ${1:name}(${2})${3} => (${4});";
         return new SnippetBlock(ItemResolverConstants.EXPRESSION_BODIED_FUNCTION, ItemResolverConstants.FUNCTION,
                 snippet, ItemResolverConstants.SNIPPET_TYPE, Kind.SNIPPET);
     }
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SnippetGenerator.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SnippetGenerator.java
@@ -460,6 +460,17 @@ public class SnippetGenerator {
     }
 
     /**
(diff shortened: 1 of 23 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/util/SnippetGenerator.java
```diff
@@ -459,6 +459,17 @@ public static SnippetBlock getFunctionDefSnippet() {
                 ItemResolverConstants.SNIPPET_TYPE, Kind.SNIPPET);
     }
 
+    /**
+     * Get expression bodied function snippet block.
+     *
+     * @return {@link SnippetBlock} Generated snippet
+     */
+    public static SnippetBlock getExpressionBodiedFunctionDefSnippet() {
+        String snippet = "function ${1:name}(${2})${3} => ${4}";
```
TharushiJay, line 468: Is there any reason why we're not including `();` in the generated snippet? 
What if we suggest `function name() => ();` in the snippet?

IMS94: As per the spec, there can be any expression after the `=>`. Having `()` is not a must.
```
function-defn := 
   metadata
   [public] function-quals
   function identifier function-signature function-defn-body

function-defn-body :=
   block-function-body
   | expr-function-body ;
   | external-function-body ;

expr-function-body := => expression
```

madushajg: Do we have some other incomplete snippet suggestion too? Can't we think `()` as the default expression since the return type of this function is also `()`. In snippet suggestions like `if true {}` also uses default values right?

madushajg: Also given that the 'design' code lens is provided for these expression bodied functions, we are expecting some novice Ballerina users to use this snippet completion to get started with the visual data mapper.
In that case we thought to have the `();` part since that shows the entire base structure.
@hevayo @kaviththiranga WDYT?

IMS94: @madushajg Had a discussion with @TharushiJay as well. Agreed with both of you. I will update the snippet to include the `();` part.

mohanvive (approved):

Issue #30577, opened 2021-05-14
Compiler doesn't complain about assigning value to module level final constructs


lasinicl:
**Description:**
$title

**Steps to reproduce:**
The following compiles. 
```ballerina
function foo() {

}

public function main() {
    foo = foo;
}
```

**Affected Versions:**
slalpha5

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

MaryamZi:
Duplicate of https://github.com/ballerina-platform/ballerina-lang/issues/24315.

PR #38430, opened 2022-10-27
[Master] Log error for assign values to function identifier

### LINKED ISSUE #24315: No error given for assigning new values to a function's identifier
Consider the following:
```ballerina
import ballerina/io;

public function main() {
    foo = function(string s, int i) returns string => "Bar"; // Not complaining about this
    io:println(foo("", 10));
}

function foo(string s, int i) returns string => "Foo";
```
The above compiles and runs fine. It doesn't print `Bar`, which means the function `foo` points to remains unchanged. But that assignment should've caused a compile error since IINM, we consider function names as final.

prakanth97:
## Purpose

Fixes #24315

## Approach
Disallow assigning values to function construct

## Samples
```
function foo() returns int => 1;

public function main() {
      foo = foo;
}
```
This will error `invalid assignment: 'function' declaration is final`

## Remarks
https://github.com/ballerina-platform/ballerina-lang/issues/30577

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -2876,12 +2876,6 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
                 varRefExpr.symbol = varSym;
                 actualType = varSym.type;
                 markAndRegisterClosureVariable(symbol, varRefExpr.pos, data.env, data);
-
-                if (varRefExpr.isLValue && (varSym.tag & SymTag.FUNCTION) == SymTag.FUNCTION) {
-                    actualType = symTable.semanticError;
-                    dlog.error(varRefExpr.pos, DiagnosticErrorCode.INVALID_ASSIGNMENT_DECLARATION_FINAL,
-                            Names.FUNCTION);
-                }
             } else if ((symbol.tag & SymTag.TYPE_DEF) == SymTag.TYPE_DEF) {
                 actualType = symbol.type.tag == TypeTags.TYPEDESC ? symbol.type : new BTypedescType(symbol.type, null);
                 varRefExpr.symbol = symbol;
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -2877,8 +2877,7 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
                 actualType = varSym.type;
                 markAndRegisterClosureVariable(symbol, varRefExpr.pos, data.env, data);
 
-                if ((varRefExpr.isLValue || varRefExpr.isCompoundAssignmentLValue)
-                        && (varSym.tag & SymTag.FUNCTION) == SymTag.FUNCTION) {
+                if (varRefExpr.isLValue && (varSym.tag & SymTag.FUNCTION) == SymTag.FUNCTION) {
                     actualType = symTable.semanticError;
                     dlog.error(varRefExpr.pos, DiagnosticErrorCode.INVALID_ASSIGNMENT_DECLARATION_FINAL,
                             Names.FUNCTION);
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -3021,6 +3021,12 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
             refItem.isLValue = true;
             checkExpr(refItem, data);
 
+            if (types.isFunctionVarRef(refItem)) {
+                dlog.error(refItem.pos, DiagnosticErrorCode.INVALID_ASSIGNMENT_DECLARATION_FINAL,
+                        Names.FUNCTION);
+                unresolvedReference = true;
+            }
+
             if (!isValidVariableReference(refItem)) {
                 unresolvedReference = true;
                 continue;
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -2876,6 +2876,13 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
                 varRefExpr.symbol = varSym;
                 actualType = varSym.type;
                 markAndRegisterClosureVariable(symbol, varRefExpr.pos, data.env, data);
+
+                if ((varRefExpr.isLValue || varRefExpr.isCompoundAssignmentLValue)
+                        && (varSym.tag & SymTag.FUNCTION) == SymTag.FUNCTION) {
+                    actualType = symTable.semanticError;
+                    dlog.error(varRefExpr.pos, DiagnosticErrorCode.INVALID_ASSIGNMENT_DECLARATION_FINAL,
+                            Names.FUNCTION);
+                }
             } else if ((symbol.tag & SymTag.TYPE_DEF) == SymTag.TYPE_DEF) {
                 actualType = symbol.type.tag == TypeTags.TYPEDESC ? symbol.type : new BTypedescType(symbol.type, null);
                 varRefExpr.symbol = symbol;
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/functions/CustomFunctionTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/functions/CustomFunctionTest.java
@@ -56,9 +56,8 @@ public class CustomFunctionTest {
     public void testAssignValueToObjectMethodNegative() {
         CompileResult compile = BCompileUtil.compile("test-src/functions/assign_value_to_object_method.bal");
         int i = 0;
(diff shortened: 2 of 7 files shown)


tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/functions/CustomFunctionTest.java
```diff
@@ -42,4 +42,23 @@ public void testDuplicateParameters() {
...
+        BAssertUtil.validateError(compile, i++, "invalid assignment: 'function' declaration is final", 20, 5);
+        BAssertUtil.validateError(compile, i++, "invalid assignment: 'function' declaration is final", 21, 5);
+        Assert.assertEquals(compile.getErrorCount(), i);
+    }
+
+    @Test
+    public void testAssignValueToObjectMethodNegative() {
+        CompileResult compile = BCompileUtil.compile("test-src/functions/assign_value_to_object_method.bal");
+        int i = 0;
+        BAssertUtil.validateError(compile, i++, "cannot update 'final' object field '$anonType$$myObj$$_0.getVal'",
```
pcnfernando, line 59: Make use of a type reference and give a name to the anon type

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2876,6 +2876,13 @@ public void visit(BLangSimpleVarRef varRefExpr, AnalyzerData data) {
                 varRefExpr.symbol = varSym;
                 actualType = varSym.type;
                 markAndRegisterClosureVariable(symbol, varRefExpr.pos, data.env, data);
+
+                if ((varRefExpr.isLValue || varRefExpr.isCompoundAssignmentLValue)
```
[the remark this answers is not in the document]
prakanth97: `varRefExpr.isCompoundAssignmentLValue` true case is not possible here, because the issue occur only for `functions`. I will remove the flag check

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/functions/CustomFunctionTest.java
prakanth97: Updated error msgs 2071b739b73103f9352bc9c60afc5320df799740

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -2876,6 +2876,12 @@ public void visit(BLangSimpleVarRef varRefExpr, AnalyzerData data) {
                 varRefExpr.symbol = varSym;
                 actualType = varSym.type;
                 markAndRegisterClosureVariable(symbol, varRefExpr.pos, data.env, data);
+
+                if (varRefExpr.isLValue && (varSym.tag & SymTag.FUNCTION) == SymTag.FUNCTION) {
+                    actualType = symTable.semanticError;
+                    dlog.error(varRefExpr.pos, DiagnosticErrorCode.INVALID_ASSIGNMENT_DECLARATION_FINAL,
+                            Names.FUNCTION);
+                }
```
KavinduZoysa, line 2884: I think it is better to do this check in https://github.com/ballerina-platform/ballerina-lang/blob/acd3357a4fe56a4f01fb85d942f499795cadf172/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java#L2256,
because we don't want to check for all the simple variable references.

prakanth97: Here we first check for `varRefExpr.isLValue` and then we do the check, So we are not checking for all the simple variable references. IMO, both are same.

prakanth97: Updated 648b6f5c2286d7e8281a917dfa543eb2e24448e1

KavinduZoysa:
@pcnfernando, shall we add a test like this
```ballerina
function foo(string s, int i) returns string => "Foo";

function init() {
    [foo] = [function(string s, int i) returns string => "Bar"];
}
```

KavinduZoysa:
> @prakanth97, shall we add a test like this?
> 
> ```ballerina
> function foo(string s, int i) returns string => "Foo";
> 
> function init() {
>     [foo] = [function(string s, int i) returns string => "Bar"];
> }
> ```

@prakanth97, we can extend this for other destructuring statements, right?

prakanth97:
> > @prakanth97, shall we add a test like this?
> > ```ballerina
> > function foo(string s, int i) returns string => "Foo";
> > 
> > function init() {
> >     [foo] = [function(string s, int i) returns string => "Bar"];
> > }
> > ```
> 
> @prakanth97, we can extend this for other destructuring statements, right?

Yes, I have updated f5d3e1a895e63072bfb813cdc3a32862588fd60f

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -3021,6 +3021,12 @@ public void visit(BLangErrorVarRef varRefExpr, AnalyzerData data) {
             refItem.isLValue = true;
             checkExpr(refItem, data);
 
+            if (types.isFunctionVarRef(refItem)) {
+                dlog.error(refItem.pos, DiagnosticErrorCode.INVALID_ASSIGNMENT_DECLARATION_FINAL,
+                        Names.FUNCTION);
+                unresolvedReference = true;
```
prakanth97, line 3027: This test (https://github.com/ballerina-platform/ballerina-lang/pull/38430/commits/f5d3e1a895e63072bfb813cdc3a32862588fd60f#diff-fe936e5cf31d19393d76ddb6a810f5b5fdd587ebd7d59e8523f71c76dfd2d814R39) covers this

tests/jballerina-unit-test/src/test/resources/test-src/functions/assign_value_to_function_negative.bal
```diff
@@ -0,0 +1,44 @@
...
+
+type BarRec record {|
+    function() returns string func;
+    int id;
+|};
+
+function testAssignValueToFunctionIdentifierInMapBindingPatternNegative() {
+    int n;
+    BarRec rec = {func : function() returns string => "Bar", id : 1};
+    {func : foo, id : n} = rec;
```
KavinduZoysa, line 34: We can add a similar case like this right?
`{func : foo, id : n} = {func : function() returns string => "Bar", id : 1};`
Applies to error-destructuring as well.

tests/jballerina-unit-test/src/test/resources/test-src/functions/assign_value_to_function_negative.bal
```diff
@@ -0,0 +1,44 @@
...
+}
+
+type TestError error<BarRec>;
+
+function testAssignValueToFunctionIdentifierInErrorBindingPatternNegative() {
+    string reason;
+    int n;
+    TestError testError = error("Error", func = function() returns string => "Bar", id = 1);
+    error(reason, func = foo, id = n) = testError;
+}
```
KavinduZoysa, line 44: Shall we add similar cases for methods as well?

tests/jballerina-unit-test/src/test/resources/test-src/functions/assign_value_to_function_negative.bal
prakanth97: `record literal is not supported for record binding pattern` for this we are getting this error

prakanth97: `error(reason, func = foo) = error("Error", func = function() returns string => "Bar");` For this got `cannot bind undefined error detail field 'func'` error.

Should we add these cases since they are restricted already?

tests/jballerina-unit-test/src/test/resources/test-src/functions/assign_value_to_function_negative.bal
prakanth97: According to the spec https://ballerina.io/spec/lang/master/#binding-pattern, 
this syntax is not allowed `[a.x] = [2];`, where `a` is the object and `x` is the identifier of the method. So, do we need to add tests for that?

KavinduZoysa: Not needed

tests/jballerina-unit-test/src/test/resources/test-src/functions/assign_value_to_function_negative.bal
KavinduZoysa: Not needed

KavinduZoysa (approved):

KavinduZoysa:
@prakanth97, please check this too https://github.com/ballerina-platform/ballerina-lang/issues/24315#issuecomment-1055001491

prakanth97:
> @prakanth97, please check this too [#24315 (comment)](https://github.com/ballerina-platform/ballerina-lang/issues/24315#issuecomment-1055001491)

Checked that. This is not reproducible in other constructs

KavinduZoysa:
> > @prakanth97, please check this too [#24315 (comment)](https://github.com/ballerina-platform/ballerina-lang/issues/24315#issuecomment-1055001491)
> 
> Checked that. This is not reproducible in other constructs

@prakanth97, can you check whether relevant tests are there?

prakanth97:
> > > @prakanth97, please check this too [#24315 (comment)](https://github.com/ballerina-platform/ballerina-lang/issues/24315#issuecomment-1055001491)
> > 
> > 
> > Checked that. This is not reproducible in other constructs
> 
> @prakanth97, can you check whether relevant tests are there?

Checked that, it seems not all the construct have the tests. Created this issue to https://github.com/ballerina-platform/ballerina-lang/issues/38841 add tests

gimantha (approved):

PR #38447, opened 2022-10-28
[master] Improve sorting in module part context

### LINKED ISSUE #37661: [Improvement]: Improve default sorting of module level completions
### Description

Currently the order of default sorting is,

1. Snippets
2. Keywords (`SnippetBlock.Kind.KEYWORD`, does not include basic types)
3. Module completion items
4. Type completion items (Defined types, constants, enums, records, class. Does not include basic types)
5. Rest of the completions

Here we get the basic types and service templates in the 5th category because of its symbol kind. We need to prioritise them.  

More info:

- Completion items for basic types are built by getting the hardcoded set of lang lib module names and removing the initial `lang.` part of it. 
- Service templated are created as static completion items and the kind, `StaticCompletionItem.Kind.OTHER`(not the completionItemKind)

### Describe your problem(s)

_No response_

### Describe your solution(s)

1. Prioritise basic types and service templates
2. Remove providing the completion item `import` after other declarations which is invalid

### Labels

LanguageServer

https://github.com/ballerina-platform/ballerina-lang/issues/37444

_No response_

_No response_

malinthar:
## Purpose
$subject

Fixes #37661

With this PR we have done multiple refactoring in the completion providers. 
1. Add type symbols to completion items like `string` , `int`, etc which were previously considered modules (pre-declared) instead of types. So the completion item type has changed with this addition. So this affects the test cases in many contexts.

3. Improve the sorting order of completion items in module level as follows.

- main snippet
- service snippet
- service template snippets
- service keyword
- function snippet
- open record snippet
- close record snippet
- other snippets
- types
- keywords
- modules
- Other

## Samples
capture

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -94,11 +95,11 @@
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
-
 import javax.annotation.Nonnull;
```
TharushiJay, line 103: Checkstyle failure `'javax.annotation.Nonnull' should be separated from previous imports.`

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -596,6 +613,61 @@ private List<LSCompletionItem> getBasicAndOtherTypeCompletions(BallerinaCompleti
...
+                    }
+                    CompletionItem cItem = TypeCompletionItemBuilder.build(
+                            typeSymbol, moduleSymbol.id().modulePrefix());
+                    completionItems.add(new SymbolCompletionItem(context, 
+                            Objects.requireNonNullElse(typeSymbol, moduleSymbol), cItem));
+                });
+        return completionItems;
+    }
+
+    protected List<LSCompletionItem> getInBuildTypes(BallerinaCompletionContext context) {
```
IMS94, line 670: Method name should be `getInBuiltTypes()`

IMS94 (changes requested):

gayalkuruppu (commented):
1. Is there a specific reason to include `service` keyword after the service snippets but not `function` keyword after function snippet?
2. Is there a way that we can improve following scenarios(only those of which are within the scope of this PR/issue)
 a. getting `main function` snippet in `test.bal`
 b. not getting relevant completions after `public` (here we get `error` after `public`)
 c. getting `main function` snippet after `public`

https://user-images.githubusercontent.com/46120162/203648932-ad9e423d-6888-4702-96c3-4026080ee858.mov

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/util/ModulePartNodeContextUtil.java
```diff
@@ -62,22 +69,63 @@ private ModulePartNodeContextUtil() {
...
 
         return completionItems;
     }
 
+    private static boolean isMainFunctionAvailable(BallerinaCompletionContext context) {
+        Optional<ModulePartNode> modulePartNode = getModulePartNode(context);
+        if (modulePartNode.isEmpty()) {
+            return false;
+        }
+        return modulePartNode.get().members().stream().anyMatch(moduleMemberDeclarationNode ->
```
IMS94, line 98: I think we should be able to use `visibleSymbols` to determine if we have a `main()` function already. Shall we use that approach?

IMS94 (changes requested):

gayalkuruppu:
> 

Created https://github.com/ballerina-platform/ballerina-lang/issues/38817 to track 2.a and 2.b

gayalkuruppu:
Previously I had a hard time getting the `string` completion item which is used more often than `stream`, and it was annoying as a user(hope this is relatable to others as well).

Screenshot 2022-11-24 at 15 38 17

But with this new refactoring we get the `string` completion item before `stream` and it so satisfying. 

Screenshot 2022-11-24 at 15 38 31

Maybe we can use this as an example to improve UX in future. It can be a whole new different scenario, but can make a huge difference for a regular user.

malinthar:
> 1. Is there a specific reason to include `service` keyword after the service snippets but not `function` keyword after function snippet?
> 
>     2. Is there a way that we can improve following scenarios(only those of which are within the scope of this PR/issue)
>        a. getting `main function` snippet in `test.bal`
>        b. not getting relevant completions after `public` (here we get `error` after `public`)
>        c. getting `main function` snippet after `public`
> 
> 
> Screen.Recording.2022-11-24.at.02.51.53.mov

service is a module level member in ballerina. That's why we have special cased it. We can special case the function keyword as well. +1

IMS94:
Shall we send this to 2201.3.x as well?

IMS94 (approved):

PR #38563, opened 2022-11-03
Fix completions for quote identifier given without the quote

### LINKED ISSUE #38485: [Bug]: Completions for quote identifier given without the quote
### Description

Consider the following:
image

In the above, when I click on `type`, it gets completed as `type`, whereas it should've been completed as `'type`. Same when i type `f.` and get the completion. IIRC, this used to work properly, where we used to give quoted identifiers in completions if the user had written it as a quoted identifier in the code. 

```ballerina
public function main() {
    Foo f = {<cursor1>};
    f.<cursor2>
}

type Foo record {|
    int 'type;
|};
```

### Steps to Reproduce

_No response_

### Affected Version(s)

2201.2.1

_No response_

### Related area

-> Editor

_No response_

_No response_

_No response_

mindula:
## Purpose
$subject

Fixes #38485

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/RecordUtil.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/RecordUtil.java
@@ -73,7 +73,7 @@ public class RecordUtil {
             CompletionItem fieldItem = new CompletionItem();
             fieldItem.setInsertText(insertText);
             fieldItem.setInsertTextFormat(InsertTextFormat.Snippet);
-            fieldItem.setLabel(name);
+            fieldItem.setLabel(CommonUtil.escapeReservedKeyword(name));
             fieldItem.setKind(CompletionItemKind.Field);
             fieldItem.setSortText(Priority.PRIORITY120.toString());
             completionItems.add(new RecordFieldCompletionItem(context, field, fieldItem, detail));
--- a/language-server/modules/langserver-core/src/test/resources/completion/expression_context/config/mapping_expr_ctx_config29.json
+++ b/language-server/modules/langserver-core/src/test/resources/completion/expression_context/config/mapping_expr_ctx_config29.json
@@ -31,7 +31,7 @@
       "insertTextFormat": "Snippet"
     },
     {
-      "label": "version",
+      "label": "'version",
       "kind": "Field",
       "detail": "Application.version",
       "sortText": "K",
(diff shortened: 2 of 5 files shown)


IMS94 (approved):

language-server/modules/langserver-core/src/test/resources/completion/expression_context/config/mapping_expr_ctx_config29.json
```diff
@@ -31,7 +31,7 @@
       "insertTextFormat": "Snippet"
     },
     {
-      "label": "version",
+      "label": "'version",
```
IMS94, line 34: Seems like the original issue complains about the `insertText`, not the label. And I think the insertText has been fixed by another PR. Which means we don't have to do anything new to fix this.

IMS94 (changes requested):

language-server/modules/langserver-core/src/test/resources/completion/expression_context/config/mapping_expr_ctx_config29.json
IMS94: Just saw your comment on the issue regarding https://github.com/ballerina-platform/ballerina-lang/pull/38003. This is fine then.

IMS94 (approved):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/common/utils/RecordUtil.java
```diff
@@ -73,7 +73,7 @@ public static List<LSCompletionItem> getRecordFieldCompletionItems(BallerinaComp
             CompletionItem fieldItem = new CompletionItem();
             fieldItem.setInsertText(insertText);
             fieldItem.setInsertTextFormat(InsertTextFormat.Snippet);
-            fieldItem.setLabel(name);
+            fieldItem.setLabel(CommonUtil.escapeReservedKeyword(name));
```
pubudu91, line 76: From where are you getting this `name` var from? If it's from the symbol for the field, then there shouldn't be any need for you to manually escape it like this. When getting the name from the symbol it should give the correctly escaped name :thinking:

IMS94: We use `RecordTypeSymbol.fieldDescriptors()` map to determine the field names and their symbols. We use the keys of that map as names.

PR #38578, opened 2022-11-08
[master] Improve sorting in error constructor expression 

### LINKED ISSUE #38291: [Bug]: Invalid completions for arguments in error constructor
### Description

Refer to the following screenshot. The second arg is expected to be a subtype of `error?`. But here, it's listing non `error?` typed constructs as well.

Screenshot from 2022-10-19 19-44-23

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

-> Editor

_No response_

_No response_

_No response_

gayalkuruppu:
Fixes #38291

https://user-images.githubusercontent.com/46120162/203263254-56e2b560-c7c8-4648-9182-b23121dea557.mov

Screenshot 2022-11-08 at 09 29 40

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


malinthar (changes requested):
Changes look good. Let's get the following fixed as well. 
According to the spec we only allow named arguments after the first positional argument in error constructor context. (https://ballerina.io/spec/lang/2021R1/#error-arg-list)

We can improve the sorting further to list the named arg completion items `code =`  and `detail=` to the top in the following scenario.
image

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ErrorConstructorExpressionNodeContext.java
```diff
@@ -201,8 +219,22 @@ private boolean isInErrorMessageArgContext(BallerinaCompletionContext context,
...
+                                             ErrorConstructorExpressionNode node) {
+
+        if (!withinArgs(context, node) || node.arguments().isEmpty() || node.arguments().size() == 1) {
+            return false;
+        }
+        FunctionArgumentNode secondArgNode = node.arguments().get(1);
+        TextRange textRange = secondArgNode.textRange();
+        int cursor = context.getCursorPositionInTree();
+        return cursor < textRange.startOffset() || (secondArgNode.kind() == SyntaxKind.POSITIONAL_ARG
+                && textRange.startOffset() <= cursor && cursor <= textRange.endOffset());
```
malinthar, line 232: Shall we remove this condition and instead check if the cursor is before the `close paran`. 

```
                cursor <= textRange.endOffset());
```

The reason is, when there is more than two params this would not work.

Eg: 
error("message", namedArg1 = val, <cursor>)

malinthar (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ErrorConstructorExpressionNodeContext.java
gayalkuruppu: This is written only aiming the second pos arg which is optional and is the error cause with the type `error?`. Correct me if I'm wrong.

gayalkuruppu: In error constructor expressions we can expect an optional second positional argument. Hence in that case, I have provided both `error?` typed completions and the expected named args in the case for the second argument. After the second argument, I only provide named args completions and as you suggested I have prioritised the snippets. As per the offline discussion, we agreed on the approach. Please unresolve and reply if there is any deviation.

IMS94 (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/ErrorConstructorExpressionNodeContext.java
```diff
@@ -138,45 +145,55 @@ private List<LSCompletionItem> getErrorTypeRefCompletions(BallerinaCompletionCon
...
-            if (semanticModel.isEmpty()) {
-                super.sort(context, node, completionItems);
-                return;
-            }
-            TypeSymbol typeSymbol = semanticModel.get().types().STRING;
+            TypeSymbol stringTypeSymbol = types.STRING;
             for (LSCompletionItem completionItem : completionItems) {
                 completionItem.getCompletionItem()
-                        .setSortText(SortingUtil.genSortTextByAssignability(context, completionItem, typeSymbol));
+                        .setSortText(SortingUtil.genSortTextByAssignability(context, completionItem, stringTypeSymbol));
```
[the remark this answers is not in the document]
gayalkuruppu: Done in [ceb2a93](https://github.com/ballerina-platform/ballerina-lang/pull/38578/commits/ceb2a93c753ebcba7e5eaa2c36c5b91b7050c350)

IMS94 (approved):

malinthar (approved):
LGTM!

IMS94 (approved):

PR #38637, opened 2022-11-11
Update JSON to Record convert tool to generate non conflicting record names with already existing types

### LINKED ISSUE #38441: [Improvement]: Generate non conflicting record names in json to record implementation
### Description

When we are converting json to Record we have to look at the conflicting record names and generate non conflicting record names.

### Steps to Reproduce
Please try `Paste as JSON` option in VSCode with a simple JSON. Then try to paste the same JSON again. It will create the record names.
As a solution we can create record names appending some postfix if a record exists with the same name in the current ballerina module.

### Affected Version(s)

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

-> Other Area

_No response_

_No response_

_No response_

AzeemMuzammil:
Fixes #38441

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


misc/json-to-record-converter/src/main/java/io/ballerina/jsonmapper/JsonToRecordMapper.java
```diff
@@ -117,7 +116,7 @@ public static JsonToRecordResponse convert(String jsonString, String recordName,
      * @return {@link JsonToRecordResponse} Ballerina code block or the Diagnostics
      */
     public static JsonToRecordResponse convert(String jsonString, String recordName, boolean isRecordTypeDesc,
-                                               boolean isClosed, boolean forceFormatRecordFields, Path filePath) {
+                                               boolean isClosed, boolean forceFormatRecordFields, String filePath) {
```
NipunaRanasinghe, line 119: Any specific reason to change the parameter type from `Path` to `String` in here?

misc/ls-extensions/modules/json-to-record-converter/src/main/java/io/ballerina/converters/JsonToRecordRequest.java
```diff
@@ -78,4 +80,12 @@ public boolean getForceFormatRecordFields() {
     public void setForceFormatRecordFields(boolean forceFormatRecordFields) {
         this.forceFormatRecordFields = forceFormatRecordFields;
     }
+
+    public String getFilePath() {
+        return filePath;
+    }
```
NipunaRanasinghe, line 86: Since the file uri is optional in the request, shall we return an `Optional<String>` in here instead, so that the getter will be self explanatory?

misc/json-to-record-converter/src/main/java/io/ballerina/jsonmapper/JsonToRecordMapper.java
AzeemMuzammil: Underlying serializer of LSP(gson) cannot serialize String/URI to `java.nio.file.Path` since it's an interface. And even now I realized to use URI instead of String as no intermediate conversion required then (we discussed that the request will contain filePath URI)

misc/ls-extensions/modules/json-to-record-converter/src/main/java/io/ballerina/converters/JsonToRecordRequest.java
AzeemMuzammil: Yeah, that would make sense 👍

AzeemMuzammil: Serialization of `string` to `Optional<String>` fails as it is not supported. so we have to handle `null` manually.

misc/json-to-record-converter/src/main/java/io/ballerina/jsonmapper/util/ConverterUtils.java
```diff
@@ -74,6 +84,77 @@ public static String escapeIdentifier(String identifier) {
...
+                try {
+                    project = BuildProject.load(projectRoot);
+                    List<Symbol> moduleSymbols = project.currentPackage()
+                            .module(project.documentId(filePathResolved).moduleId())
+                            .getCompilation().getSemanticModel().moduleSymbols();
+                    moduleSymbols.forEach(symbol -> {
+                        if (symbol.getName().isPresent()) {
+                            existingTypeNames.add(symbol.getName().get());
+                        }
+                    });
```
NipunaRanasinghe, line 130: Shall we improve the logic to access the semantic model directly from the LS workspace manager, to avoid redundant project loadings in the tool side?

NipunaRanasinghe:
@AzeemMuzammil Also let's create a separate issue to add integration test case(s) for the record converter command, from the VSCode side.

misc/json-to-record-converter/src/main/java/io/ballerina/jsonmapper/diagnostic/DiagnosticMessage.java
```diff
@@ -96,4 +96,28 @@ public static DiagnosticMessage jsonToRecordConverter104(Object[] args) {
...
+
+    public static DiagnosticMessage jsonToRecordConverter106(Object[] args) {
+        if (args != null && args.length >= 2) {
+            return new DiagnosticMessage("JSON_TO_RECORD_CONVERTER_106",
+                    String.format("The record name ''%s'' is renamed as ''%s''. " +
+                            "Consider rename it back to a meaningful name.", args[0], args[1]), DiagnosticSeverity.INFO,
+                    Arrays.copyOfRange(args, 2, args.length));
+        }
+        return new DiagnosticMessage("JSON_TO_RECORD_CONVERTER_106",
+                "Few of the record names are renamed. " +
```
NipunaRanasinghe, line 120:
```suggestion
                "Few of the records are renamed. " +
```

misc/json-to-record-converter/src/main/java/io/ballerina/jsonmapper/util/ConverterUtils.java
```diff
@@ -74,6 +87,89 @@ public static String escapeIdentifier(String identifier) {
...
+                } catch (ProjectException pe1) {
+                    return existingTypeNames;
+                }
+            }
```
NipunaRanasinghe, line 149: Instead of handling build projects inside the catch block, cant we first check for the project root and, branch out into the two separate logics(single file and package) based on the output?

AzeemMuzammil: Updated as per the suggestion. Please review and resolve the conversation. @NipunaRanasinghe

NipunaRanasinghe (approved):
LGTM

NipunaRanasinghe (approved):
LGTM

Issue #38483, opened 2022-10-31
[Bug]: Named worker strands do not run on separate threads even if the function is isolated


Nadeeshan96:
### Description

$title

The spec says in https://ballerina.io/spec/lang/master/#section_7.3.2
> Each named worker runs on its own new strand. The strand can run on a separate thread from the default worker if the function is isolated.

We see the `BIRFunction.annotAttachments` is empty in the functions created for the named workers even if they are inside the isolated functions.

### Steps to Reproduce

```bal
import ballerina/io;

type Person record {|
    string name;
    boolean employed;
|};

isolated function process(Person[] & readonly members, int[] & readonly quantities) {
    worker w1 {
        Person[] employedMembers = from Person p in members
            where p.employed
            select p;
        int count = employedMembers.length();
        count -> function;
    }

    int totalMemberCount = members.length();

    int memberCount = <- w1;
    io:println(string `Employed Members: ${memberCount}`);

    future<int> avgFuture = start get_average(quantities, memberCount);

    int unemployedCount = totalMemberCount - memberCount;
    io:println(string `Unemployed Members: ${unemployedCount}`);

    int avg = checkpanic wait avgFuture;
    io:println(string `Average: ${avg}`);
}

isolated function get_average(int[] & readonly quantities, int employedCount) returns int {
    int total = int:sum(...quantities);
    int avg = employedCount == 0 ? 0 : total / employedCount;
    return avg;
}

public function main() {
    Person[] & readonly members = [{name: "aaa", employed: false}, {name: "aaa", employed: true}, {name: "aaa", employed: true}];
    int[] & readonly quantities = [10, 20, 30];
    process(members, quantities);
}
```

We see though the `get_average` function contains the annotation, the `$lambda$0` function for worker `w1` does not have the annotation.

### Affected Version(s)

Ballerina 2201.2.2 (Swan Lake Update 2)

### Related area

-> Compilation

Nadeeshan96:
Here need to add the `strand thread : any` annotation to the `BLangFunction` of the workers which are going to run on separate threads. There it is needed to:
add the annotation to `function.annAttachments` and `function.symbol.annotationAttachments` and change `function.symbol.schedulerPolicy` to `any`.

lasinicl:
Fixed with https://github.com/ballerina-platform/ballerina-lang/issues/38483

PR #38652, opened 2022-11-14
Fix adding strand annotation to isolated workers

lasinicl:
## Purpose
$title

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/38483

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


Nadeeshan96:
@lasinicl Since the issue is not a blocker, shall we send this fix only to the `master` branch? We may also need to change the 3 strand dump tests there because now some worker strands will be on separate strand groups.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -6453,10 +6454,14 @@ public void visit(BLangInvocation.BLangActionInvocation actionInvocation) {
...
         }
 
-        // Add `@strand {thread: "any"}` annotation to an isolated start-action or
-        // isolated named worker declaration.
-        if (actionInvocation.async && Symbols.isFlagOn(actionInvocation.symbol.type.flags, Flags.ISOLATED)) {
-            addStrandAnnotationWithThreadAny(actionInvocation);
+        // Add `@strand {thread: "any"}` annotation to an isolated start-action in an isolated function.
+        if (!actionInvocation.functionPointerInvocation && actionInvocation.async &&
+                Symbols.isFlagOn(actionInvocation.symbol.type.flags, Flags.ISOLATED) &&
+                isInIsolatedFunction(env.enclInvokable)) {
```
Nadeeshan96, line 6460: Here I think we don't need to check whether the start action is inside an isolated function. I think what is mentioned in the spec is that isolated functions can only contain isolated start actions.
Strands created by the isolated start actions can run on a separate thread even when it is not inside an isolated function. So I think we can omit the `&& isInIsolatedFunction(env.enclInvokable)` part.

Nadeeshan96: Btw previously even though we added `@strand {thread: "any"}` annotation to the `functionPointerInvocation start actions` also, it did not cause a problem because it got omitted when creating a FPcall. Either way if we have `!actionInvocation.functionPointerInvocation &&` part or not, I think it is fine, but need to do the most suitable thing. I think we can remove that check also.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -7707,9 +7707,35 @@ public void visit(BLangTypeConversionExpr conversionExpr) {
     @Override
     public void visit(BLangLambdaFunction bLangLambdaFunction) {
         bLangLambdaFunction.function = rewrite(bLangLambdaFunction.function, bLangLambdaFunction.capturedClosureEnv);
+        BLangFunction function = bLangLambdaFunction.function;
+        // Add `@strand {thread: "any"}` annotation to an isolated named worker declaration in an isolated function.
+        if (function.flagSet.contains(Flag.WORKER) && Symbols.isFlagOn(function.symbol.type.flags, Flags.ISOLATED) &&
+                isInIsolatedFunction(env.enclInvokable)) {
```
Nadeeshan96, line 7713: Does `isInIsolatedFunction(env.enclInvokable)` returns `true` if it is now in an inferred isolated function also? If it is so the check is okay, otherwise I think we should call a method which does that.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
lasinicl: > Here I think we don't need to check whether the start action is inside an isolated function. I think what is mentioned in the spec is that isolated functions can only contain isolated start actions. Strands created by the isolated start actions can run on a separate thread even when it is not inside an isolated function. So I think we can omit the `&& isInIsolatedFunction(env.enclInvokable)` part.
 
Fixed this. 


> Btw previously even though we added `@strand {thread: "any"}` annotation to the `functionPointerInvocation start actions` also, it did not cause a problem because it got omitted when creating a FPcall. Either way if we have `!actionInvocation.functionPointerInvocation &&` part or not, I think it is fine. I think we can remove that check also.

Unnecessarily adding an annotation would cause an overhead, i think, because we do some desugaring related to the record init function introduced with the annotation in the closure desugar. Hence keeping this as it is.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
lasinicl: Yes. The function symbol contains the `ISOLATED` flag even when it's inferred isolated since we add it in the IsolationAnayzer.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -7710,9 +7709,35 @@ public void visit(BLangTypeConversionExpr conversionExpr) {
...
         result = bLangLambdaFunction;
     }
 
+    private boolean isInIsolatedFunction(BLangInvokableNode enclInvokable) {
+        if (enclInvokable == null) {
+            if (env.node.getKind() != NodeKind.EXPR_FUNCTION_BODY ||
+                    env.enclEnv.node.getKind() != NodeKind.ARROW_EXPR) {
+                return false;
+            }
+            return isIsolated(((BLangArrowFunction) env.enclEnv.node).funcType.flags);
```
KavinduZoysa, line 7731: Can't this cast be cause CCE?

lasinicl: used the same function in isolationAnalyzer. Willl check and update if necessary.

tests/jballerina-integration-test/src/test/resources/troubleshoot/strandDump/testOutputs/testPackageWithModulesStrandDumpRegEx.txt
```diff
@@ -1,12 +1,12 @@
...
 ===========================================
 
-Total strand group count \t:\t8
+Total strand group count \t:\t14
 Total strand count       \t:\t19
-Active strand group count\t:\t5
+Active strand group count\t:\t11
 Active strand count      \t:\t17
 
-group \d* \[QUEUED\]: \[5\]
```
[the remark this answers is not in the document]
Nadeeshan96: Btw, in the `master` branch there is one more test, so may need that also like this when sending the PR to that.

Nadeeshan96 (approved):
LGTM

HindujaB (approved):
LGTM

PR #38892, opened 2022-11-30
Add examples for the array langlib

MaryamZi:
## Purpose
$title.

Closes https://github.com/ballerina-platform/ballerina-lang/issues/38907

### DIFF
--- a/langlib/lang.array/src/main/ballerina/array.bal
+++ b/langlib/lang.array/src/main/ballerina/array.bal
@@ -37,6 +37,11 @@ type AnydataType anydata;
 
 # Returns the number of members of an array.
 #
+# ```ballerina
+# string[] greetings = ["Hello", "Bonjour", "Hola", "Ciao"];
+# int length = greetings.length();
+# ```
+#
 # + arr - the array
 # + return - number of members in parameter `arr`
 public isolated function length((any|error)[] arr) returns int = @java:Method {
@@ -46,8 +51,16 @@ public isolated function length((any|error)[] arr) returns int = @java:Method {
 
 # Returns an iterator over an array.
 #
+# ```ballerina
+# int[] evenNumbers = [2, 4, 6, 8];
+# object {
+#     public isolated function next() returns record {|int value;|}?;
+# } iterator = evenNumbers.iterator();
+# record {|int value;|}? next = iterator.next();
+# ```
+#
 # + arr - the array
-# + return - a new iterator object that will iterate over the members of parameter `arr`.
+# + return - a new iterator object that will iterate over the members of parameter `arr`
 public isolated function iterator(Type[] arr) returns object {
     public isolated function next() returns record {|
         Type value;
@@ -59,6 +72,11 @@ public isolated function iterator(Type[] arr) returns object {
 
 # Returns a new array consisting of index and member pairs.
 #
+# ```ballerina
+# string[] greetings = ["Hello", "Bonjour", "Hola", "Ciao"];
+# [int, string][] enumeration = greetings.enumerate();
+# ```
+#
 # + arr - the array
 # + return - array of index, member pairs
 public isolated function enumerate(Type[] arr) returns [int, Type][] = @java:Method {
@@ -69,6 +87,33 @@ public isolated function enumerate(Type[] arr) returns [int, Type][] = @java:Met
 // Functional iteration
 
 # Applies a function to each member of an array and returns an array of the results.
+#
+# ```ballerina
+# string[] greetings = ["Hello", "Bonjour", "Hola", "Ciao"];
+# int[] lengths = greetings.'map((greeting) => greeting.length());
+#
+# Employee[] employees = [
+#     {firstName: "Jo", lastName: "David", id: 2121},
+#     {firstName: "Emma", id: 2122},
+#     {firstName: "John", lastName: "Doe", id: 2123}
+# ];
+# string[] fullNames =
+#     employees.'map(isolated function (Employee employee) returns string {
+#         string? lastName = employee.lastName;
+#
+#         if lastName is () {
+#             return employee.firstName;
+#         }
+#
+#         return string `${employee.firstName} ${lastName}`;
+#     });
+#
+# type Employee record {
+#     string firstName;
+#     string lastName?;
+#     int id;
+# };
+# ```
 #
 # + arr - the array
 # + func - a function to apply to each member
@@ -82,6 +127,26 @@ public isolated function 'map(Type[] arr, @isolatedParam function(Type val) retu
 #
 # The parameter `func` is applied to each member of parameter `arr` in order.
 #
+# ```ballerina
+# Employee[] employees = [
+#     {name: "Jo", salary: 1200, id: 2121},
+#     {name: "Emma", id: 2122, salary: ()},
+#     {name: "John", salary: 1500, id: 2123}
+# ];
+#
+# employees.forEach(isolated function (Employee employee) {
+#     if employee.salary == () {
+#         employee.salary = 1000;
+#     }
+# });
+#
+# type Employee record {|
+#     string name;
+#     int id;
+#     decimal? salary;
+# |};
+# ```
+#
 # + arr - the array
 # + func - a function to apply to each member
 public isolated function forEach(Type[] arr, @isolatedParam function(Type val) returns () func) returns () = @java:Method {
@@ -90,6 +155,27 @@ public isolated function forEach(Type[] arr, @isolatedParam function(Type val) r
 } external;
 
 # Selects the members from an array for which a function returns true.
+#
+# ```ballerina
+# string[] greetings = ["Hello", "Bonjour", "Hola", "Ciao"];
+# string[] filteredByLength = greetings.filter((greeting) => greeting.length() > 4);
+#
+# Employee[] employees = [
+#     {name: "Jo", salary: 1200, id: 2121},
+#     {name: "Emma", id: 2122, salary: 900},
+#     {name: "John", salary: 1500, id: 2123}
+# ];
+# Employee[] filteredBySalary = employees.filter(isolated function (Employee employee) returns boolean {
+#     decimal salary = employee.salary;
+#     return salary > 1000d && salary < 1500d;
+# });
+#
+# type Employee record {|
+#     string name;
+#     int id;
+#     decimal salary;
+# |};
+# ```
 #
 # + arr - the array
 # + func - a predicate to apply to each member to test whether it should be selected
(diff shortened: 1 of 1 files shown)


langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -46,6 +51,14 @@ public isolated function length((any|error)[] arr) returns int = @java:Method {
...
+# ```ballerina
+# int[] evenNumbers = [2, 4, 6, 8];
+# object {
+#     public isolated function next() returns record {|int value;|}?;
+# } iterator = evenNumbers.iterator();
+# record {|int value;|}? next = iterator.next();
+# ```
+#
 # + arr - the array
 # + return - a new iterator object that will iterate over the members of parameter `arr`.
```
SasinduDilshara, line 63: I think we need to remove this `.`I

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -70,6 +88,33 @@ public isolated function enumerate(Type[] arr) returns [int, Type][] = @java:Met
...
+#
+#         if lastName is () {
+#             return employee.firstName;
+#         }
+#
+#         return string `${employee.firstName} ${lastName}`;
+#     });
+#
+# type Employee record {
+#     string firstName;
```
SasinduDilshara, line 112: Imo, this should be define before the usage. it will improve user experience. wdyt?

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: Can you add it as a suggestion?

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: This was done intentionally since the idea is to be able to quickly understand how to use the function. By adding the type at the end, the user has to focus on the type only if they really want to. WDYT?

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -70,6 +88,33 @@ public isolated function enumerate(Type[] arr) returns [int, Type][] = @java:Met
 
 # Applies a function to each member of an array and returns an array of the results.
 #
+# ```ballerina
+# string[] greetings = ["Hello", "Bonjour", "Hola", "Ciao"];
+# int[] lengths = greetings.'map((greeting) => greeting.length());
+#
```
sanjana, line 94: Would it be better if a comment is added here to distinguish between the two examples?

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -91,6 +156,27 @@ public isolated function forEach(Type[] arr, @isolatedParam function(Type val) r
 
 # Selects the members from an array for which a function returns true.
 #
+# ```ballerina
+# string[] greetings = ["Hello", "Bonjour", "Hola", "Ciao"];
+# string[] filteredByLength = greetings.filter((greeting) => greeting.length() > 4);
+#
```
sanjana, line 162: Ditto.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -104,11 +190,30 @@ public isolated function filter(Type[] arr, @isolatedParam function(Type val) re
 # The combining function takes the combined value so far and a member of the array,
 # and returns a new combined value.
 #
-# For example
-# ```
-# reduce([1, 2, 3], function (int total, int n) returns int { return total + n; }, 0)
+# ```ballerina
+# int[] integers = [1, 2, 3];
+# int sum = integers.reduce(isolated function (int total, int next) returns int => total + next, 0);
+#
```
sanjana, line 196: Ditto.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -104,11 +190,30 @@ public isolated function filter(Type[] arr, @isolatedParam function(Type val) re
...
+#
```
sanjana, line 197: Better if this superfluous newline is removed.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -124,6 +229,27 @@ public isolated function reduce(Type[] arr, @isolatedParam function(Type1 accum,
 # The parameter `func` is called for each member of parameter `arr` in order unless and until a call returns true.
 # When the array is empty, returns false.
 #
+# ```ballerina
+# int[] numbers = [1, 2, 3, 5];
+# boolean hasEvenNumber = numbers.some((number) => number % 2 == 0);
+#
```
sanjana, line 235: Again, would it be better if a comment is added here to distinguish between the two examples?

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -152,6 +288,27 @@ public isolated function slice(Type[] arr, int startIndex, int endIndex = arr.le
 # The parameter `func` is called for each member of `arr` in order unless and until a call returns false.
 # When the array is empty, returns true.
 #
+# ```ballerina
+# int[] numbers = [1, 2, 3, 5];
+# boolean allEvenNumbers = numbers.every((number) => number % 2 == 0);
+#
```
sanjana, line 294: Ditto.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -152,6 +288,27 @@ public isolated function slice(Type[] arr, int startIndex, int endIndex = arr.le
 # The parameter `func` is called for each member of `arr` in order unless and until a call returns false.
 # When the array is empty, returns true.
 #
+# ```ballerina
+# int[] numbers = [1, 2, 3, 5];
+# boolean allEvenNumbers = numbers.every((number) => number % 2 == 0);
```
sanjana, line 293: Wouldn't it be better to use the variable name as something like `hasOnlyEvenNumbers` or, `isAllEvenNumbers`?

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: IMO, it is self-explanatory, since both calls use the same number of arguments unlike with something like `array:sort`.

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: Please add this as a suggestion.

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: We use `is`/`has` as a prefix generally with the function/method as opposed to the variable. Is the current form not clear?

langlib/lang.array/src/main/ballerina/array.bal
sanjana, line 197:
```suggestion
```

Remove the superfluous newline.

langlib/lang.array/src/main/ballerina/array.bal
sanjana: Since the type of the variable is `boolean`, it would be better to indicate it as a flag variable.
Something like: `evenNumbersFlag` or `allEvenNumbersFlag`. 
IMO, the current form would convey the idea of **an array consisting of all the even numbers**.

sanjana (changes requested):
Added suggestions.

langlib/lang.array/src/main/ballerina/array.bal
SasinduDilshara: +1

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: Given that this is the variable declaration and that it is a boolean variable, I don't think it will be misinterpreted to be an array.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -82,6 +127,26 @@ public isolated function 'map(Type[] arr, @isolatedParam function(Type val) retu
 #
 # The parameter `func` is applied to each member of parameter `arr` in order.
 #
+# ```ballerina
+# Employee[] employees = [
+#     {name: "Jo", salary: 1200, id: 2121},
+#     {name: "Emma", id: 2122, salary: ()},
+#     {name: "John", salary: 1500, id: 2123}
```
dulajdilshan, line 134:
```suggestion
#     {name: "John", id: 2123, salary: 1500}
```

Or is this intentional?

MaryamZi:
Lang server tests haven't been fixed yet.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -82,6 +127,26 @@ public isolated function 'map(Type[] arr, @isolatedParam function(Type val) retu
...
+# ];
```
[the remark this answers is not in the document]
SasinduDilshara: Yes. But it will improve the readability, because of the value order is equal to the field order in type definition.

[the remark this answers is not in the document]
MaryamZi: For consistency between the members in the list constructor we can change to the proposed order. But I don't see how readability improves when the order is the same in the value as in the type-descriptor. I don't think there'll generally be a 1:1 comparison between the two.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -70,6 +88,33 @@ public isolated function enumerate(Type[] arr) returns [int, Type][] = @java:Met
...
+#
+# Employee[] employees = [
+#     {firstName: "Jo", lastName: "David", id: 2121},
+#     {firstName: "Emma", id: 2122},
+#     {firstName: "John", lastName: "Doe", id: 2123}
+# ];
+# string[] fullNames =
+#     employees.'map(isolated function (Employee employee) returns string {
+#         string? lastName = employee.lastName;
+#
```
SasinduDilshara, line 103:
```suggestion
```
We can remove this line?

SasinduDilshara (changes requested):

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -104,11 +190,29 @@ public isolated function filter(Type[] arr, @isolatedParam function(Type val) re
 # The combining function takes the combined value so far and a member of the array,
 # and returns a new combined value.
 #
-# For example
-# ```
-# reduce([1, 2, 3], function (int total, int n) returns int { return total + n; }, 0)
```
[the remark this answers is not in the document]
MaryamZi: We can do this when we are adding the examples to the spec.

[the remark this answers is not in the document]
SasinduDilshara: ok, Is it same for L50 and L111 ?.

MaryamZi:
New PR - https://github.com/ballerina-platform/ballerina-lang/pull/38989

PR #38905, opened 2022-12-01
[Master] Add lang library examples for float langlib apidocs

### LINKED ISSUE #38904: [Improvement]: Add lang library examples for float langlib apidocs
### Description

Currently lang library API docs does not contains usage examples.
So we need to add examples for float lang library API docs.

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

-> Other Area

_No response_

_No response_

_No response_

SasinduDilshara:
Fixes #38904

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/langlib/lang.float/src/main/ballerina/float.bal
+++ b/langlib/lang.float/src/main/ballerina/float.bal
@@ -33,6 +33,13 @@ public const float Infinity = 1.0/0.0;
 #
 # Exactly one of isFinite, isInfinite and IsNaN will be true for any float value
 #
+# ```ballerina
+# float f = 1.2;
+# f.isFinite() ⇒ true
+# 
+# float:Infinity.isFinite() ⇒ false
+# ```
+#
 # + x - the float to be tested
 # + return - true if parameter `x` is finite, i.e., neither NaN nor +∞ nor -∞
 public isolated function isFinite(float x) returns boolean = @java:Method {
@@ -43,6 +50,13 @@ public isolated function isFinite(float x) returns boolean = @java:Method {
 # Tests whether a float is infinite.
 #
 # Exactly one of isFinite, isInfinite and IsNaN will be true for any float value
+#
+# ```ballerina
+# float f = 3.21;
+# f.isInfinite() ⇒ false
+# 
+# float:Infinity.isInfinite() ⇒ true
+# ```
 #
 # + x - the float to be tested
 # + return - true if parameter `x` is either +∞ or -∞
@@ -55,6 +69,13 @@ public isolated function isInfinite(float x) returns boolean = @java:Method {
 #
 # Exactly one of isFinite, isInfinite and IsNaN will be true for any float value.
 #
+# ```ballerina
+# float f = 0.23;
+# f.isNaN() ⇒ false
+# 
+# float:NaN.isNaN() ⇒ true
+# ```
+#
 # + x - the float to be tested
 # + return - true if parameter `x` is NaN
 public isolated function isNaN(float x) returns boolean = @java:Method {
@@ -65,6 +86,18 @@ public isolated function isNaN(float x) returns boolean = @java:Method {
 # Returns the sum of zero or more float values.
 #
 # Result is NaN if any arg is NaN
+#
+# ```ballerina
+# float:sum(1.2, 2.3, 3.4) ⇒ 6.9
+#
+# float[] scores = [11.1, 22.2, 33.3];
+# float:sum(...scores) ⇒ 66.6
+#
+# float f = 21.2;
+# f.sum(10.5, 21, 32.4) ⇒ 85.1
+# 
+# float:sum(float:NaN, 2.3, 3.4) ⇒ NaN
+# ```
 #
 # + xs - float values to sum
 # + return - sum of all of parameter `xs`, +0.0 if parameter `xs` is empty
@@ -78,6 +111,20 @@ public isolated function sum(float... xs) returns float = @java:Method {
 # Result is -∞ if no args
 # NaN if any arg is NaN
 #
+# ```ballerina
+# float:max(1.2, 12.3, 3.4) ⇒ 12.3
+#
+# float[] marks = [70.3, 80.5, 98.1, 92.3];
+# float:max(...marks) ⇒ 98.1
+#
+# float f = 21.2;
+# f.max(40.5, 21, 32.4) ⇒ 40.5
+#
+# float:max() ⇒ -Infinity
+# 
+# float:max(1.2, float:NaN, 3.4) ⇒ NaN
+# ```
+#
 # + xs - float values to operate on
 # + return - maximum value of all of parameter `xs`
 public isolated function max(float... xs) returns float = @java:Method {
@@ -90,6 +137,20 @@ public isolated function max(float... xs) returns float = @java:Method {
 # Result is +∞ if no args
 # Result is NaN if any arg is NaN
 #
+# ```ballerina
+# float:min(5.2, 2.3, 3.4) ⇒ 2.3
+#
+# float[] marks = [90.3, 80.5, 98, 92.3];
+# float:min(...marks) ⇒ 80.5
+#
+# float f = 1.2;
+# f.min(10.5, 21, 32.4) ⇒ 1.2
+#
+# float:min() ⇒ Infinity
+# 
+# float:min(5.2, float:NaN, 3.4) ⇒ NaN
+# ```
+#
 # + xs - float values to operate on
 # + return - minimum value of all of parameter `xs`
 public isolated function min(float... xs) returns float = @java:Method {
@@ -98,6 +159,11 @@ public isolated function min(float... xs) returns float = @java:Method {
 } external;
 
 # Returns the IEEE absolute value of a float value.
+#
+# ```ballerina
+# float f = -3.21;
+# f.abs() ⇒ 3.21
+# ```
 #
 # + x - float value to operate on
 # + return - absolute value of parameter `x`
@@ -113,11 +179,25 @@ public isolated function abs(float x) returns float = @java:Method {
 # (this is the round-to-nearest rounding mode, which is the default for IEEE and for Ballerina).
 # A value of `fractionDigits` greater than 0 thus corresponds to the number of digits after the decimal
 # point being `fractionDigits`; a value of 0 for `fractionDigits` rounds to an integer.
-# If `x` is NaN, +0, -0, +∞ or -∞, then the result is `x``.
+# If `x` is NaN, +0, -0, +∞ or -∞, then the result is `x`.
 # When `fractionDigits` is 0, this is
 # the same as Java Math.rint method, .NET Math.Round method and
 # IEEE roundToIntegralTiesToEven operation
 # Note that `<int>x` is the same as `<int>x.round(0)`.
+#
+# ```ballerina
+# float f = 3.55;
+# f.round() ⇒ 4.0
+# 
+# float g = 4.55555;
+# g.round(3) ⇒ 4.556
+# 
+# float h = 2.5;
+# h.round(0) ⇒ 2.0
+# 
+# float i = 3.5;
+# i.round(0) ⇒ 4.0
+# ```
 #
 # + x - float value to operate on
 # + fractionDigits - the number of digits after the decimal point
@@ -129,6 +209,11 @@ public isolated function round(float x, int fractionDigits = 0) returns float = 
 
 # Rounds a float down to the closest integral value.
 #
+# ```ballerina
+# float f = 3.51;
+# f.floor() ⇒ 3.0
+# ```
+#
 # + x - float value to operate on
 # + return - largest (closest to +∞) float value not greater than parameter `x` that is a mathematical integer
 public isolated function floor(float x) returns float = @java:Method {
@@ -137,6 +222,11 @@ public isolated function floor(float x) returns float = @java:Method {
 } external;
 
 # Rounds a float up to the closest integral value.
+#
+# ```ballerina
+# float f = 3.51;
+# f.ceiling() ⇒ 4.0
+# ```
 #
 # + x - float value to operate on
 # + return - smallest (closest to -∞) decimal value not less than parameter `x` that is a mathematical integer
@@ -149,6 +239,11 @@ public isolated function ceiling(float x) returns float = @java:Method {
 #
 # Corresponds to IEEE squareRoot operation.
 #
+# ```ballerina
+# float f = 1.96;
+# f.sqrt() ⇒ 1.4
+# ```
+#
 # + x - float value to operate on
 # + return - square root of parameter `x`
 public isolated function sqrt(float x) returns float = @java:Method {
@@ -160,6 +255,11 @@ public isolated function sqrt(float x) returns float = @java:Method {
 #
 # Corresponds to IEEE rootn(x, 3) operation.
 #
+# ```ballerina
+# float f = 0.125;
+# f.cbrt() ⇒ 0.5
+# ```
+#
 # + x - float value to operate on
 # + return - cube root of parameter `x`
 public isolated function cbrt(float x) returns float = @java:Method {
@@ -170,6 +270,11 @@ public isolated function cbrt(float x) returns float = @java:Method {
 # Raises one float value to the power of another float values.
 #
 # Corresponds to IEEE pow(x, y) operation.
+#
+# ```ballerina
+# float f = 2.1;
+# f.pow(2) ⇒ 4.41
+# ```
 #
 # + x - base value
 # + y - the exponent
@@ -183,6 +288,11 @@ public isolated function pow(float x, float y) returns float = @java:Method {
 #
 # Corresponds to IEEE log operation.
 #
+# ```ballerina
+# float f = 234.56;
+# f.log() ⇒ 5.4577114186982865
+# ```
+#
 # + x - float value to operate on
 # + return - natural logarithm of parameter `x`
 public isolated function log(float x) returns float = @java:Method {
@@ -193,6 +303,11 @@ public isolated function log(float x) returns float = @java:Method {
 # Returns the base 10 logarithm of a float value.
 #
 # Corresponds to IEEE log10 operation.
+#
+# ```ballerina
+# float f = 0.1;
+# f.log10() ⇒ -1.0
+# ```
 #
 # + x - float value to operate on
 # + return - base 10 logarithm of parameter `x`
@@ -205,6 +320,11 @@ public isolated function log10(float x) returns float = @java:Method {
 #
 # Corresponds to IEEE exp operation.
 #
+# ```ballerina
+# float f = 2.3;
+# f.exp() ⇒ 9.974182454814718
+# ```
+#
 # + x - float value to operate on
 # + return - Euler's number raised to the power parameter `x`
 public isolated function exp(float x) returns float = @java:Method {
@@ -215,6 +335,11 @@ public isolated function exp(float x) returns float = @java:Method {
 # Returns the sine of a float value.
 #
 # Corresponds to IEEE sin operation.
+#
+# ```ballerina
+# float f = 2.3;
+# f.sin() ⇒ 0.7457052121767203
+# ```
 #
 # + x - float value, specifying an angle in radians
 # + return - the sine of parameter `x`
@@ -227,6 +352,11 @@ public isolated function sin(float x) returns float = @java:Method {
 #
 # Corresponds to IEEE cos operation.
 #
+# ```ballerina
+# float f = 0.7;
+# f.cos() ⇒ 0.7648421872844885
+# ```
+#
 # + x - float value, specifying an angle in radians
 # + return - the cosine of parameter `x`
 public isolated function cos(float x) returns float = @java:Method {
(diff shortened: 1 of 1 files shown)


langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -55,6 +65,11 @@ public isolated function isInfinite(float x) returns boolean = @java:Method {
 #
 # Exactly one of isFinite, isInfinite and IsNaN will be true for any float value.
 #
+# ```ballerina
+# float nanValue = 0.0/0.0;
```
SasinduDilshara, line 69: We can wrap the `0.0/0.0` to another variable, or we can use ballerina `float:NaN`
what will be the best choice?

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -64,10 +79,21 @@ public isolated function isNaN(float x) returns boolean = @java:Method {
...
+# // Take the sum of multiple float numbers.
+# float total = float:sum(1.2, 2.3, 3.4);
+#
+# // Take the sum of existing float number list.
+# float[] scores = [11.1, 22.2, 33.3];
+# float totalScore = float:sum(...scores);
+#
+# // Take the sum of multiple float numbers with an existing float number.
+# float value = 21.2;
+# float summation = value.sum(10.5, 21, 32.4);
```
SasinduDilshara, line 92: This example is added because Ballerina has this functionality. But this may not a good usage.
Shall I remove this example?.
applicable for `min, max` as well.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -33,6 +33,11 @@ public const float Infinity = 1.0/0.0;
 #
 # Exactly one of isFinite, isInfinite and IsNaN will be true for any float value
 #
+# ```ballerina
+# float finiteValue = 1.2;
+# boolean isFinite = finiteValue.isFinite();
```
MaryamZi, line 38: The example has to be generic, i.e., show how it works with an arbitrary float value. So I'd name the variable something like `floatValue`. If the value is known to be finite, there's no point actually calling `isFinite`, right?

MaryamZi, line 38: We generally use the `is` prefix with the function/method and not the variable. Let's discuss this and finalize a convention. This came up with the array langlib too.

MaryamZi: Applies to the other functions also.

langlib/lang.float/src/main/ballerina/float.bal
MaryamZi: This isn't just with an existing float variable. Even `21.2.sum(10.5, 21, 32.4)` works. I think it's OK to add an example, but let's change the description to say you can also use a method call expression.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -64,10 +79,21 @@ public isolated function isNaN(float x) returns boolean = @java:Method {
...
+# float totalScore = float:sum(...scores);
+#
+# // Take the sum of multiple float numbers with an existing float number.
+# float value = 21.2;
+# float summation = value.sum(10.5, 21, 32.4);
+#```
 #
 # + xs - float values to sum
-# + return - sum of all of parameter `xs`, +0.0 if parameter `xs` is empty
+# + return - sum of all of parameter `xs`, +0.0 if parameter `xs` is empty, NaN if any member in xs is NaN
```
MaryamZi, line 96: Please don't change the documentation, here and in L67. Documentation changes need to be added in/come from the spec.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -113,12 +170,22 @@ public isolated function abs(float x) returns float = @java:Method {
...
 # point being `fractionDigits`; a value of 0 for `fractionDigits` rounds to an integer.
-# If `x` is NaN, +0, -0, +∞ or -∞, then the result is `x``.
+# If `x` is NaN, +0, -0, +∞ or -∞, then the result is `x`.
 # When `fractionDigits` is 0, this is
 # the same as Java Math.rint method, .NET Math.Round method and
 # IEEE roundToIntegralTiesToEven operation
 # Note that `<int>x` is the same as `<int>x.round(0)`.
 #
+# ```ballerina
+# // Round float number to nearest integer
```
MaryamZi, line 180:
```suggestion
# // Round a float number to the nearest integer.
```

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -332,7 +490,12 @@ public isolated function tanh(float x) returns float = @java:Method {
 # - `NaN` is allowed
 # - `Infinity` is allowed with an optional leading `+` or `-` sign
 # - a FloatingPointTypeSuffix is not allowed
-# This is the inverse of function ``value:toString`` applied to an `float`.
+# This is the inverse of function `value:toString` applied to an `float`.
+#
+# ```ballerina
+# string strMarks = "0.2453";
+# float floatMarks = check float:fromString(strMarks);
```
MaryamZi, line 497:
```suggestion
# string scoreStr = "0.2453";
# float score = check float:fromString(scoreStr);
```

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -344,13 +507,18 @@ public isolated function fromString(string s) returns float|error = @java:Method
 # Returns a string that represents a float value as a hexadecimal floating point number.
 #
 # The returned string will comply to the grammar of HexFloatingPointLiteral
-# in the Ballerina spec with the following modifications:
+# in the Ballerina specification with the following modifications:
```
MaryamZi, line 510: Let's not change these here.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -425,6 +616,14 @@ public isolated function toFixedString(float x, int? fractionDigits) returns str
 # to represent the result. If `x` is zero, the exponent is zero. A zero exponent
 # is represented with a `+` sign.
 #
+# ```ballerina
+# // Take the string representation of scientific notation without decimal points.
+# string strScientificIntValue = float:toExpString(12.456, 0);
```
MaryamZi, line 621: Is saying int value here (and in the previous example) correct?

langlib/lang.float/src/main/ballerina/float.bal
MaryamZi: I prefer `float:NaN` in order to keep the example simple. See https://github.com/ballerina-platform/ballerina-lang/pull/38905#discussion_r1037775654 for variable naming though.

dulajdilshan: +1 to use `float:NaN`

langlib/lang.float/src/main/ballerina/float.bal
sameerajayasoma: I prefer `f` in this context.  see https://go.dev/tour/moretypes/19. They are concise and easier the explain the point.

langlib/lang.float/src/main/ballerina/float.bal
mohanvive: I am also +1 to float:NaN .

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: I agree with `min, max, sum`. But it is not helpful with `atan2` ryt?

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: Shall I go with `finite`, `infinite` and `nan` as variable names for this scenarios instead of `isFinite`, `isInfinite`, `isnan`?
wdyt?

MaryamZi: +1

langlib/lang.float/src/main/ballerina/float.bal
MaryamZi: Okay, let's have one example here then.

langlib/lang.float/src/main/ballerina/float.bal
MaryamZi: +1 to have for just min, max, and sum.

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: Updated

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: Updated

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: Updated

SasinduDilshara: Updated

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: Removed

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: updated

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: updated

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: yes, but here our target is to expose how the function can use. So because of there is no optional arguments, I guess it is better to have one example. wdyt?

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -216,6 +325,11 @@ public isolated function exp(float x) returns float = @java:Method {
 #
 # Corresponds to IEEE sin operation.
 #
+# ```ballerina
+# float radians = 2.3;
+# float sinValue = radians.sin();
```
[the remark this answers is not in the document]
SasinduDilshara: `sine` is use within the descriptions. Is it ok to use `sin` in variable names, since it is the mathematical notation?.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -66,6 +81,19 @@ public isolated function isNaN(float x) returns boolean = @java:Method {
 #
 # Result is NaN if any arg is NaN
 #
+# ```ballerina
+# // Compute the sum of multiple float numbers.
+# float total = float:sum(1.2, 2.3, 3.4);
+#
+# // Compute the sum of existing float number list.
```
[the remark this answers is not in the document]
SasinduDilshara: Updated.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -66,6 +81,19 @@ public isolated function isNaN(float x) returns boolean = @java:Method {
...
+# float[] scores = [11.1, 22.2, 33.3];
+# float totalScore = float:sum(...scores);
+#
+# // you can also use a method call expression to calculate the summation.
```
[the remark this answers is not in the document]
SasinduDilshara: Updated

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -113,12 +172,22 @@ public isolated function abs(float x) returns float = @java:Method {
...
 # point being `fractionDigits`; a value of 0 for `fractionDigits` rounds to an integer.
-# If `x` is NaN, +0, -0, +∞ or -∞, then the result is `x``.
+# If `x` is NaN, +0, -0, +∞ or -∞, then the result is `x`.
 # When `fractionDigits` is 0, this is
 # the same as Java Math.rint method, .NET Math.Round method and
 # IEEE roundToIntegralTiesToEven operation
 # Note that `<int>x` is the same as `<int>x.round(0)`.
 #
+# ```ballerina
+# // Round a float number to the nearest integer.
```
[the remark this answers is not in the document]
SasinduDilshara: Updated. Please check it is ok or not.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -401,6 +586,11 @@ public isolated function fromBitsInt(int x) returns float = @java:Method {
 # If `fractionDigits` is zero, there will be no decimal point.
 # Any necessary rounding will use the roundTiesToEven rounding direction.
 #
+# ```ballerina
+# float f = 12.456;
+# string strFixedFloatValue = f.toFixedString(2);
```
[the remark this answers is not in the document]
SasinduDilshara: updated

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -113,12 +172,22 @@ public isolated function abs(float x) returns float = @java:Method {
 # (this is the round-to-nearest rounding mode, which is the default for IEEE and for Ballerina).
 # A value of `fractionDigits` greater than 0 thus corresponds to the number of digits after the decimal
 # point being `fractionDigits`; a value of 0 for `fractionDigits` rounds to an integer.
-# If `x` is NaN, +0, -0, +∞ or -∞, then the result is `x``.
+# If `x` is NaN, +0, -0, +∞ or -∞, then the result is `x`.
```
MaryamZi, line 175: Please send a PR fixing this to the spec repo also.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -113,12 +172,22 @@ public isolated function abs(float x) returns float = @java:Method {
...
 # When `fractionDigits` is 0, this is
 # the same as Java Math.rint method, .NET Math.Round method and
 # IEEE roundToIntegralTiesToEven operation
 # Note that `<int>x` is the same as `<int>x.round(0)`.
 #
+# ```ballerina
+# float score = 3.55;
+# // Round a float number to the nearest integer and return as a float.
```
MaryamZi, line 183: I think we can just say

```suggestion
# // Round a float number to zero fractional digits.
```

Since the description explains it a bit.

langlib/lang.float/src/main/ballerina/float.bal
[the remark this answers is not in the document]
SasinduDilshara: Update the PR with `sine` instead of `sin`.

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: Will update

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: Updated.

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: https://github.com/ballerina-platform/ballerina-spec/pull/1201

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -417,14 +607,19 @@ public isolated function toFixedString(float x, int? fractionDigits) returns str
 # But if `x` is NaN or infinite, the result will be the same as `value:toString`.
 # The digit before the decimal point will be zero only if all other digits
 # are zero.
-# This will panic if fractionDigits is less than 0.
+# This will panic if `fractionDigits` is less than 0.
```
SasinduDilshara, line 610: Did this change to make this consistent with spec.
Please refer this PR https://github.com/ballerina-platform/ballerina-spec/pull/1201

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -33,6 +33,11 @@ public const float Infinity = 1.0/0.0;
 #
 # Exactly one of isFinite, isInfinite and IsNaN will be true for any float value
 #
+# ```ballerina
+# float f = 1.2;
+# boolean finite = f.isFinite();
```
SasinduDilshara, line 38: What will be the best option for update this example
```
1. 1.2.isFinite() 
2. (1.2).isFinite()
```

MaryamZi: I think we can use the variable here. This is somewhat similar to using a variable in `str.length()` vs `"abc".length()`. 

cc @mohanvive

mohanvive: Hmm, yes. IMO, in this case introducing a variable gives more clarity.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -78,6 +103,16 @@ public isolated function sum(float... xs) returns float = @java:Method {
 # Result is -∞ if no args
 # NaN if any arg is NaN
```
SasinduDilshara, line 104: Do we need to add examples for these scenarios?
Imo, no need to add.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -171,6 +249,11 @@ public isolated function cbrt(float x) returns float = @java:Method {
 #
 # Corresponds to IEEE pow(x, y) operation.
 #
+# ```
+# float width = 2.1;
+# width.pow(2) ⇒ 4.41
```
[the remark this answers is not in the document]
MaryamZi: Same applies to some others too.

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -334,6 +481,14 @@ public isolated function tanh(float x) returns float = @java:Method {
 # - a FloatingPointTypeSuffix is not allowed
 # This is the inverse of function ``value:toString`` applied to an `float`.
 #
+# ```
+# float:fromString("0.2453") ⇒ 0.2453
+# 
+# float:fromString("-10") ⇒ -10.0
+# 
+# float:fromString("Hello") ⇒ error
```
MaryamZi, line 489:
```suggestion
# float:fromString("") ⇒ error
```

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -366,6 +526,12 @@ public isolated function toHexString(float x) returns string = @java:Method {
 # - `NaN` is allowed
 # - `Infinity` is allowed with an optional leading `+` or `-` sign
 #
+# ```
+# float:fromHexString("0x1.0a3d70a3d70a4p4") ⇒ 16.64
+# 
+# float:fromHexString("John") ⇒ error
```
MaryamZi, line 532: Let's use a different example, where the string is more closer to being a valid one.

langlib/lang.float/src/main/ballerina/float.bal
MaryamZi: We can demonstrate them with simple (one-line) examples, right? So +1 to add.

langlib/lang.float/src/main/ballerina/float.bal
SasinduDilshara: I think we need an example that clearly show the difference,
When user see the example they should get the idea that, other strings will not work in here.

MaryamZi: The fact that "John" is not a hex string and therefore cannot be used to create a float is obvious, right? Why not something like 

```suggestion
# float:fromHexString("0x1J") ⇒ error
```

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -33,6 +33,14 @@ public const float Infinity = 1.0/0.0;
 #
 # Exactly one of isFinite, isInfinite and IsNaN will be true for any float value
 #
+# ```ballerina
+# float f = 1.2;
+# f.isFinite() ⇒ true
+# 
+# float f2 = float:Infinity;
+# f2.isFinite() ⇒ false
```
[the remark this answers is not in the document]
MaryamZi: Actually with the references, can even do
```suggestion
# float:Infinity.isFinite() ⇒ false
```

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -417,14 +621,25 @@ public isolated function toFixedString(float x, int? fractionDigits) returns str
...
 #
+# ```ballerina
+# float f = 12.456;
+# f.toExpString(2) ⇒ 1.25e+1
+# 
+# float f2 = 12.456;
+# f2.toExpString(()) ⇒ 1.2456e+1
+# 
+# float f3 = 12.456;
+# f3.toExpString(-2) ⇒ panic
```
sanjana, line 640: Wouldn't it better to add a trap expression here?

sanjana: Will be resolving the conversation since decided to use the current evaluation.

langlib/lang.float/src/main/ballerina/float.bal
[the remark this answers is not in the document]
SasinduDilshara: Updated in this float PRs

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -44,6 +51,13 @@ public isolated function isFinite(float x) returns boolean = @java:Method {
 #
 # Exactly one of isFinite, isInfinite and IsNaN will be true for any float value
 #
+# ```ballerina
+# float:Infinity.isInfinite() ⇒ true
```
MaryamZi, line 55: Shall we make this the second example? Same for isNan. Basically have the generic case as the first example.

MaryamZi (approved):

Issue #38929, opened 2022-12-02
[Task]: Add examples for function and future langlib functions


MaryamZi:
### Description

$title.

### Related area

-> Other Area

MaryamZi:
Done via https://github.com/ballerina-platform/ballerina-lang/pull/39110 and https://github.com/ballerina-platform/ballerina-lang/pull/39075.

PR #38935, opened 2022-12-02
Add examples for the future and function langlibs

MaryamZi:
## Purpose
$title.

Completes https://github.com/ballerina-platform/ballerina-lang/issues/38929

### DIFF
--- a/langlib/lang.function/src/main/ballerina/function.bal
+++ b/langlib/lang.function/src/main/ballerina/function.bal
@@ -21,21 +21,14 @@ import ballerina/jballerina.java;
 # If the arguments specified in `args` are not of the type required by `func`,
 # then this will panic.
 #
-# ```ballerina
-# function fn = getGreeting;
-# // Call `getGreeting` without arguments.
-# any|error greeting = function:call(fn);
+# ```
+# function getGreeting(string? name = ()) returns string => name is () ? "Hello" : string `Hello ${name}!`;
 #
-# // Call `getGreeting` with an argument.
-# any|error greetingWithName = function:call(fn, "David");
+# function:call(getGreeting) ⇒ Hello
 #
-# function getGreeting(string? name = ()) returns string {
-#     if name is () {
-#         return "Hello!";
-#     }
+# function:call(getGreeting, "David") ⇒ Hello David!
 #
-#     return string `Hello ${name}!`;
-# }
+# function:call(getGreeting, 1) ⇒ panic
 # ```
 #
 # + func - the function to be called
--- a/langlib/lang.function/src/main/ballerina/function.bal
+++ b/langlib/lang.function/src/main/ballerina/function.bal
@@ -21,6 +21,23 @@ import ballerina/jballerina.java;
 # If the arguments specified in `args` are not of the type required by `func`,
 # then this will panic.
 #
+# ```ballerina
+# function fn = getGreeting;
+# // Call `getGreeting` without arguments.
+# any|error greeting = function:call(fn);
+#
+# // Call `getGreeting` with an argument.
+# any|error greetingWithName = function:call(fn, "David");
+#
+# function getGreeting(string? name = ()) returns string {
+#     if name is () {
+#         return "Hello!";
+#     }
+#
+#     return string `Hello ${name}!`;
+# }
+# ```
+#
 # + func - the function to be called
 # + args - the arguments to be passed to the function
 # + return - the return value of the call to the function
--- a/langlib/lang.future/src/main/ballerina/future.bal
+++ b/langlib/lang.future/src/main/ballerina/future.bal
@@ -21,6 +21,25 @@ import ballerina/jballerina.java;
 # This sets the cancellation flag in the strand corresponding to `f`.
 # Each time that a strand yields, it will check the cancellation flag
 # and terminate abnormally if the flag is set.
+#
+# ```ballerina
+# int[] sales = [10, 13, 54, 245, 24, 29, 343, 34, 23, 45, 67, 93, 846];
+# future<int?> computationFuture = start computeEntryCountForTarget(sales, 500);
+# computationFuture.cancel();
+#
+# function computeEntryCountForTarget(int[] sales, int target) returns int? {
+#     int sum = 0;
+#     foreach [int, int] [index, item] in sales.enumerate() {
+#         sum += item;
+#
+#         if sum >= target {
+#             return index + 1;
+#         }
+#     }
+#     return ();
+# }
+# ```
+#
 # + f - the future to be cancelled
 public isolated function cancel(future<any|error> f) returns () = @java:Method {
     'class: "org.ballerinalang.langlib.future.Cancel",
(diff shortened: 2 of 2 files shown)


langlib/lang.function/src/main/ballerina/function.bal
```diff
@@ -21,6 +21,23 @@ import ballerina/jballerina.java;
 # If the arguments specified in `args` are not of the type required by `func`,
 # then this will panic.
 #
+# ```ballerina
+# function fn = getGreeting;
+# // Call `getGreeting` without arguments.
+# any|error greeting = function:call(fn);
+#
+# // Call `getGreeting` with an argument.
```
SasinduDilshara, line 29: Shall we change this description to something like
`// Call `getGreeting` with name parameter.` wdyt?

langlib/lang.function/src/main/ballerina/function.bal
```diff
@@ -21,6 +21,23 @@ import ballerina/jballerina.java;
 # If the arguments specified in `args` are not of the type required by `func`,
 # then this will panic.
 #
+# ```ballerina
+# function fn = getGreeting;
+# // Call `getGreeting` without arguments.
+# any|error greeting = function:call(fn);
```
SasinduDilshara, line 27: We can change the return type to `string` in here.

langlib/lang.function/src/main/ballerina/function.bal
```diff
@@ -21,6 +21,23 @@ import ballerina/jballerina.java;
...
+#
+# // Call `getGreeting` with an argument.
+# any|error greetingWithName = function:call(fn, "David");
```
SasinduDilshara, line 30: We can change the return type to `string` in here.

langlib/lang.future/src/main/ballerina/future.bal
```diff
@@ -21,6 +21,25 @@ import ballerina/jballerina.java;
...
+# function computeEntryCountForTarget(int[] sales, int target) returns int? {
+#     int sum = 0;
+#     foreach [int, int] [index, item] in sales.enumerate() {
+#         sum += item;
+#
+#         if sum >= target {
+#             return index + 1;
+#         }
+#     }
+#     return ();
```
SasinduDilshara, line 39: We don't need this line.

SasinduDilshara (changes requested):

langlib/lang.function/src/main/ballerina/function.bal
MaryamZi: Doesn't compile, does it?

langlib/lang.function/src/main/ballerina/function.bal
MaryamZi: Did not go into detail here intentionally since at this point the parameter name(s) could be anything. As opposed to directly calling `getGreeting`.

langlib/lang.function/src/main/ballerina/function.bal
MaryamZi: Won't compile.

langlib/lang.future/src/main/ballerina/future.bal
MaryamZi: Without this, a `this function should explicitly return a value` warning will be logged. In the future, it will result in a compilation error.

MaryamZi: This is by design to minimize room for mistakes: https://github.com/ballerina-platform/ballerina-spec/issues/973.

langlib/lang.future/src/main/ballerina/future.bal
```diff
@@ -21,6 +21,25 @@ import ballerina/jballerina.java;
...
+#
+# function computeEntryCountForTarget(int[] sales, int target) returns int? {
+#     int sum = 0;
+#     foreach [int, int] [index, item] in sales.enumerate() {
+#         sum += item;
+#
+#         if sum >= target {
+#             return index + 1;
+#         }
+#     }
```
SasinduDilshara, line 38: Shall we put a new line after this, just like L36 in `function.bal`.

langlib/lang.function/src/main/ballerina/function.bal
```diff
@@ -21,6 +21,23 @@ import ballerina/jballerina.java;
...
+#         return "Hello!";
+#     }
+#
+#     return string `Hello ${name}!`;
+# }
+# ```
+#
 # + func - the function to be called
 # + args - the arguments to be passed to the function
 # + return - the return value of the call to the function
```
SasinduDilshara, line 43: Isn't the return type `any|error` always?

SasinduDilshara:
LGTM

SasinduDilshara (approved):

langlib/lang.function/src/main/ballerina/function.bal
MaryamZi: The description is about the value, not the type.

langlib/lang.future/src/main/ballerina/future.bal
MaryamZi: Added a new line there because it was two returns. I'd rather remove the new line there if required than add a new line here.

langlib/lang.future/src/main/ballerina/future.bal
```diff
@@ -21,6 +21,25 @@ import ballerina/jballerina.java;
...
+# ```ballerina
+# int[] sales = [10, 13, 54, 245, 24, 29, 343, 34, 23, 45, 67, 93, 846];
+# future<int?> computationFuture = start computeEntryCountForTarget(sales, 500);
+# computationFuture.cancel();
+#
+# function computeEntryCountForTarget(int[] sales, int target) returns int? {
+#     int sum = 0;
+#     foreach [int, int] [index, item] in sales.enumerate() {
+#         sum += item;
+#
```
SasinduDilshara, line 34: Do we need this line?

langlib/lang.function/src/main/ballerina/function.bal
```diff
@@ -21,6 +21,14 @@ import ballerina/jballerina.java;
 # If the arguments specified in `args` are not of the type required by `func`,
 # then this will panic.
 #
+# ```
+# function:call(getGreeting) ⇒ Hello
+#
+# function:call(getGreeting, "David") ⇒ Hello David!
+#
+# function getGreeting(string? name = ()) returns string => name is () ? "Hello" : string `Hello ${name}!`;
```
LakshanWeerasinghe, line 29: We need to define this function at the top if we are running this examples on shell. WDYT?

SasinduDilshara: I am also think that, defining in the top will be better in this situation.

SasinduDilshara (changes requested):

langlib/lang.function/src/main/ballerina/function.bal
MaryamZi: Given that the user will often be looking for a way to use this with a function they've written I still think it's OK to have it at the end. However, I've moved this to the top since we added a panic example also.

MaryamZi:
New PR to 2201.3.x - https://github.com/ballerina-platform/ballerina-lang/pull/39075

PR #38949, opened 2022-12-05
Fix issues in bal init command

### LINKED ISSUE #38473: [Bug]: "bal init ." can result in invalid package name
### Description

When I run `bal init .` it will create a package with "." as the package name. This will result in a package where most of the tooling will not work. We should probably validate this and not allow "." as the package name for bal init. 

Also suggest to change the help text to "Initialize a Ballerina package inside the current directory" instead of  "Create a new Ballerina package inside the current  directory" which gives a different meaning.
```
NAME
       ballerina-init - Initialize a Ballerina package inside the current directory

``` 

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

-> Other Area

_No response_

_No response_

_No response_

ShammiL:
Fixes #38473

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/util/ProjectUtils.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/util/ProjectUtils.java
@@ -124,6 +124,7 @@ import static io.ballerina.projects.util.ProjectConstants.USER_NAME;
 public class ProjectUtils {
     private static final String USER_HOME = "user.home";
     private static final Pattern separatedIdentifierPattern = Pattern.compile("^[a-zA-Z0-9_.]*$");
+    private static final Pattern onlyDotsPattern = Pattern.compile("^[.]+$");
     private static final Pattern orgNamePattern = Pattern.compile("^[a-zA-Z0-9_]*$");
 
     /**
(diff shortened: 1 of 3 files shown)


compiler/ballerina-lang/src/main/java/io/ballerina/projects/util/ProjectUtils.java
```diff
@@ -721,7 +722,9 @@ public static void checkExecutePermission(Path path) {
 
     private static boolean validateDotSeparatedIdentifiers(String identifiers) {
         Matcher m = separatedIdentifierPattern.matcher(identifiers);
-        return m.matches();
+        Matcher mm = onlyDotsPattern.matcher(identifiers);
```
Dilhasha, line 725: Have we verified whether this change has no impact on other places where this util is being used?

ShammiL: Other places where this is used validates either the package names or the module names. So this change will prevent `.` from being used as a package name or a module name.

Dilhasha (approved):

PR #38989, opened 2022-12-08
Add examples for the array langlib - Take II

MaryamZi:
## Purpose
$title. In an expression-oriented style.

Original PR - https://github.com/ballerina-platform/ballerina-lang/pull/38892

### DIFF
--- a/langlib/lang.array/src/main/ballerina/array.bal
+++ b/langlib/lang.array/src/main/ballerina/array.bal
@@ -105,9 +105,9 @@ public isolated function 'map(Type[] arr, @isolatedParam function(Type val) retu
 # ```
 # int total = 0;
 # [1, 2, 3].forEach(function (int i) {
-#     total += i * 2;
+#     total += i;
 # });
-# total ⇒ 12
+# total ⇒ 6
 # ```
 #
 # + arr - the array
@@ -180,6 +180,8 @@ public isolated function some(Type[] arr, @isolatedParam function(Type val) retu
 # numbers.slice(3) ⇒ [8,10,12]
 #
 # numbers.slice(0, 4) ⇒ [2,4,6,8]
+#
+# numbers.slice(0, 10) ⇒ panic
 # ```
 #
 # + arr - the array
@@ -220,7 +222,10 @@ public isolated function every(Type[] arr, @isolatedParam function(Type val) ret
 #
 # ```
 # int[] numbers = [2, 4, 6, 8];
+#
 # numbers.remove(1) ⇒ 4
+#
+# numbers.remove(7) ⇒ panic
 # ```
 #
 # + arr - the array
@@ -239,6 +244,9 @@ public isolated function remove(Type[] arr, int index) returns Type = @java:Meth
 # int[] numbers = [2, 4, 6, 8];
 # numbers.removeAll();
 # numbers ⇒ []
+#
+# int[2] values = [1, 2];
+# values.removeAll() ⇒ panic
 # ```
 #
 # + arr - the array
@@ -277,6 +285,8 @@ public isolated function setLength((any|error)[] arr, int length) returns () = @
 #
 # strings.indexOf("e") is () ⇒ true
 #
+# strings.indexOf("b") ⇒ 1
+#
 # strings.indexOf("b", 2) ⇒ 3
 # ```
 #
@@ -297,6 +307,8 @@ public isolated function indexOf(AnydataType[] arr, AnydataType val, int startIn
 # string[] strings = ["a", "b", "d", "b", "d", "b"];
 #
 # strings.lastIndexOf("e") is () ⇒ true
+#
+# strings.lastIndexOf("b") ⇒ 5
 #
 # strings.lastIndexOf("b", strings.length() - 2) ⇒ 3
 # ```
@@ -371,6 +383,9 @@ public isolated function sort(Type[] arr, SortDirection direction = ASCENDING,
 # ```
 # int[] numbers = [2, 4, 6, 8, 10];
 # numbers.pop() ⇒ 10
+#
+# int[] values = [];
+# values.pop() ⇒ panic
 # ```
 #
 # + arr - the array
@@ -411,6 +426,9 @@ public isolated function push(Type[] arr, Type... vals) returns () = @java:Metho
 # ```
 # int[] numbers = [2, 4, 6, 8, 10];
 # numbers.shift() ⇒ 2
+#
+# int[] values = [];
+# values.shift() ⇒ panic
 # ```
 #
 # + arr - the array
@@ -452,8 +470,8 @@ public isolated function unshift(Type[] arr, Type... vals) returns () = @java:Me
 # There will be no whitespace in the returned string.
 #
 # ```
-# byte[] byteArray = [104, 101, 108, 108, 111, 32, 98, 97, 108, 108, 101, 114, 105, 110, 97, 32, 33, 33, 33];
-# byteArray.toBase64() ⇒ aGVsbG8gYmFsbGVyaW5hICEhIQ==
+# byte[] byteArray = [104, 101, 108, 108, 111, 32, 98, 97];
+# byteArray.toBase64() ⇒ aGVsbG8gYmE=
 # ```
 #
 # + arr - the array
@@ -469,9 +487,9 @@ public isolated function toBase64(byte[] arr) returns string = @java:Method {
 # and whitespace as allowed by a Ballerina Base64Literal.
 #
 # ```
-# check array:fromBase64("aGVsbG8gYmFsbGVyaW5hICEhIQ==") ⇒ [104,101,108,108,111,32,98,97,108,108,101,114,105,110,97,32,33,33,33]
+# check array:fromBase64("aGVsbG8gYmE=") ⇒ [104,101,108,108,111,32,98,97]
 #
-# array:fromBase64("aGVsbG8gYmFsbGVyaW5hICEhIQ--") ⇒ error
+# array:fromBase64("aGVsbG8gYmE--") ⇒ error
 # ```
 #
 # + str - Base64 string representation
@@ -488,8 +506,8 @@ public isolated function fromBase64(string str) returns byte[]|error = @java:Met
 # There will be no whitespace in the returned string.
 #
 # ```
-# byte[] byteArray = [170, 171, 207, 204, 173, 175, 205, 52, 26, 75, 223, 171, 205, 137, 18, 223];
-# byteArray.toBase16() ⇒ aaabcfccadafcd341a4bdfabcd8912df
+# byte[] byteArray = [170, 171, 207, 204, 173, 175, 205];
+# byteArray.toBase16() ⇒ aaabcfccadafcd
 # ```
 #
 # + arr - the array
@@ -505,9 +523,9 @@ public isolated function toBase16(byte[] arr) returns string = @java:Method {
 # and whitespace as allowed by a Ballerina Base16Literal.
 #
 # ```
-# check array:fromBase16("aaabcfccadafcd341a4bdfabcd8912df") ⇒ [170,171,207,204,173,175,205,52,26,75,223,171,205,137,18,223]
+# check array:fromBase16("aaabcfccadafcd") ⇒ [170,171,207,204,173,175,205]
 #
-# array:fromBase16("aGVsbG8gYmFsbGVyaW5hICEhIQ==") ⇒ error
+# array:fromBase16("aaabcfccadafcd==") ⇒ error
 # ```
 #
 # + str - Base16 string representation
--- a/langlib/lang.array/src/main/ballerina/array.bal
+++ b/langlib/lang.array/src/main/ballerina/array.bal
@@ -38,7 +38,7 @@ type AnydataType anydata;
 # Returns the number of members of an array.
 #
 # ```
-# [1, 2, 3, 4].length() => 4
+# ["a", "b", "c", "d"].length() ⇒ 4
 # ```
 #
 # + arr - the array
@@ -54,7 +54,7 @@ public isolated function length((any|error)[] arr) returns int = @java:Method {
 # object {
 #     public isolated function next() returns record {|int value;|}?;
 # } iterator = [2, 4, 6, 8].iterator();
-# iterator.next() => {"value":2}
+# iterator.next() ⇒ {"value":2}
 # ```
 #
 # + arr - the array
@@ -71,7 +71,7 @@ public isolated function iterator(Type[] arr) returns object {
 # Returns a new array consisting of index and member pairs.
 #
 # ```
-# [1, 2, 3, 4].enumerate() => [[0,1],[1,2],[2,3],[3,4]]
+# [1, 2, 3, 4].enumerate() ⇒ [[0,1],[1,2],[2,3],[3,4]]
 # ```
 #
 # + arr - the array
@@ -86,7 +86,8 @@ public isolated function enumerate(Type[] arr) returns [int, Type][] = @java:Met
 # Applies a function to each member of an array and returns an array of the results.
 #
 # ```
-# [0, 1, 2].map(n => n * 2) => [0,2,4]
+# int[] numbers = [0, 1, 2];
+# numbers.map(n => n * 2) ⇒ [0,2,4]
 # ```
 #
 # + arr - the array
@@ -102,11 +103,11 @@ public isolated function 'map(Type[] arr, @isolatedParam function(Type val) retu
 # The parameter `func` is applied to each member of parameter `arr` in order.
 #
 # ```
-# string str = "";
-# [0, 1, 2].forEach(function (int i) {
-#     str += i.toString();
+# int total = 0;
+# [1, 2, 3].forEach(function (int i) {
+#     total += i * 2;
 # });
-# str => 012
+# total ⇒ 12
 # ```
 #
 # + arr - the array
@@ -119,7 +120,8 @@ public isolated function forEach(Type[] arr, @isolatedParam function(Type val) r
 # Selects the members from an array for which a function returns true.
 #
 # ```
-# [12, 43, 60, 75, 10].filter(n => n > 50) => [60,75]
+# int[] numbers = [12, 43, 60, 75, 10];
+# numbers.filter(n => n > 50) ⇒ [60,75]
 # ```
 #
 # + arr - the array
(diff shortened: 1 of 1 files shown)


langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -37,6 +37,10 @@ type AnydataType anydata;
 
 # Returns the number of members of an array.
 #
+# ```
+# [1, 2, 3, 4].length() => 4
```
MaryamZi, line 41: Options being considered atm (including some proposed by others too)

1. `# [1, 2, 3, 4].length() => 4`
2. `# [1, 2, 3, 4].length() >> 4`
3. `# [1, 2, 3, 4].length() == 4`
4. `# [1, 2, 3, 4].length() // 4`

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -201,6 +264,16 @@ public isolated function setLength((any|error)[] arr, int length) returns () = @
 # Returns `()` if not found.
 # Equality is tested using `==`.
 #
+# ```
+# string[] greetings = ["Hello", "Hola", "Bonjour", "Hola", "Ciao"];
+#
+# // First index of "guten tag" if it exists in the list.
+# greetings.indexOf("guten tag") => ()
```
MaryamZi, line 271: Output is currently the `toString` representation except where the value is nil since for nil toString is the empty string and we can't represent it here. Alternatively, we can probably do something like
```suggestion
# greetings.indexOf("guten tag") is () => true
```

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -70,6 +85,10 @@ public isolated function enumerate(Type[] arr) returns [int, Type][] = @java:Met
 
 # Applies a function to each member of an array and returns an array of the results.
 #
+# ```
+# [0, 1, 2].map(n => n * 2) => [0,2,4]
```
MaryamZi, line 89: In the original PR [we had another example](https://github.com/ballerina-platform/ballerina-lang/pull/38892/files#diff-2c948d23bf7d2faa264a43f44ca32e7f64f06b4dfb220764247218037b2d10f4R91) with an explicit anonymous function expression as the argument.

```
# Employee[] employees = [
#     {firstName: "Jo", lastName: "David", id: 2121},
#     {firstName: "Emma", id: 2122},
#     {firstName: "John", lastName: "Doe", id: 2123}
# ];
# employees.'map(isolated function (Employee employee) returns string {
#     string? lastName = employee.lastName;
# 
#     if lastName is () {
#         return employee.firstName;
#     }
# 
#     return string `${employee.firstName} ${lastName}`;
# }) => ["Jo David","Emma","John Doe"]
# 
# type Employee record {
#     string firstName;
#     string lastName?;
#     int id;
# };
```

Should we continue to have this example too (after the simple one)? IMO, it will be useful to the user.

Applies to some other functions also.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -169,6 +214,11 @@ public isolated function every(Type[] arr, @isolatedParam function(Type val) ret
 # This removes the member of parameter `arr` with index parameter `index` and returns it.
 # It panics if there is no such member.
 #
+# ```
+# int[] evenNumbers = [2, 4, 6, 8];
+# evenNumbers.remove(1) => 4
```
MaryamZi, line 219: Can't do `[2, 4, 6, 8].remove(1) => 4` since a tuple type is inferred for the list and the removal fails at runtime. Same for some other functions too.

Related to https://github.com/ballerina-platform/ballerina-spec/issues/1050

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -341,6 +496,11 @@ public isolated function fromBase64(string str) returns byte[]|error = @java:Met
 # The result will contain only characters  `0..9`, `a..f`.
 # There will be no whitespace in the returned string.
 #
+# ```
+# byte[] byteArray = [170, 171, 207, 204, 173, 175, 205, 52, 26, 75, 223, 171, 205, 137, 18, 223];
+# byteArray.toBase16() => aaabcfccadafcd341a4bdfabcd8912df
```
MaryamZi, line 501: Can't do `[170, 171, 207, 204, 173, 175, 205, 52, 26, 75, 223, 171, 205, 137, 18, 223].toBase16()` due to https://github.com/ballerina-platform/ballerina-lang/issues/38997

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -46,8 +50,15 @@ public isolated function length((any|error)[] arr) returns int = @java:Method {
 
 # Returns an iterator over an array.
 #
+# ```
+# object {
+#     public isolated function next() returns record {|int value;|}?;
+# } iterator = [2, 4, 6, 8].iterator();
+# iterator.next() => {"value":2}
```
MaryamZi, line 57: Had to call `next` to get the string representation of something meaningful. Same for `toStream`.

langlib/lang.array/src/main/ballerina/array.bal
hasithaa: +1

langlib/lang.array/src/main/ballerina/array.bal
hasithaa: I think this is too much to include in a lang doc. A better place could be a BBE.

langlib/lang.array/src/main/ballerina/array.bal
hasithaa: Options,

1. break this into two. Variable definition and expression.
2. Use a type cast with an array literal. 
3. Use let expression

langlib/lang.array/src/main/ballerina/array.bal
mohanvive: I prefer the option 4 because I kind of have doubt whether we could use options for all the available cases.

langlib/lang.array/src/main/ballerina/array.bal
mohanvive: +1

langlib/lang.array/src/main/ballerina/array.bal
mohanvive: Hmm, I feel we should have examples with above option as well. That really more understandable for a new user.

langlib/lang.array/src/main/ballerina/array.bal
mohanvive: I prefer option #2 and I feel we should do that for all the examples not only above.

mohanvive:
I think we have to make a decision on what is our goal and whom we are targeting with our examples. I still believe,  examples in API docs should be simple and easy for any kind of users to understand. It is OK to write the examples by showing the best practices of the language. But I feel we shouldn't consider that as the first priority. Our first priority should be helping users with easily understandable code snippets.

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: Updated both occurrences in [037bdb0](https://github.com/ballerina-platform/ballerina-lang/pull/38989/commits/037bdb0467633b46a5fb089b26ac4dd84cdf4bf6).

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi:
```suggestion
# [1, 2, 3, 4].length() ⇒ 4
```

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -201,6 +264,16 @@ public isolated function setLength((any|error)[] arr, int length) returns () = @
 # Returns `()` if not found.
 # Equality is tested using `==`.
 #
+# ```
+# string[] greetings = ["Hello", "Hola", "Bonjour", "Hola", "Ciao"];
+#
+# // First index of "guten tag" if it exists in the list.
+# greetings.indexOf("guten tag") is () => true
```
manuranga, line 271: this could be simplified  to
```suggestion
# greetings.indexOf("guten tag") => ()
```

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -301,6 +434,19 @@ public isolated function shift(Type[] arr) returns Type = @java:Method {
 # The values newly added to the array will be in the same order
 # as they are in parameter `vals`.
 #
+# ```
+# int[] evenNumbers = [14];
+#
+# // Add multiple members to the start of the list.
+# evenNumbers.unshift(10, 12);
+# evenNumbers => [10,12,14]
```
manuranga, line 442: need spaces

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: This was to allow for the toString representation, differentiating between nil and empty string. https://github.com/ballerina-platform/ballerina-lang/pull/38989#discussion_r1044052894

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: This is the toString representation, which is also what the shell produces.

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi, line 41: Use different values since 4 is also the last member.

MaryamZi:
Cover error/panic scenarios where applicable/useful too.

Use `⇒ error` or `⇒ panic`. Can't include the toString representation of the error since it depends on the implementation.

langlib/lang.array/src/main/ballerina/array.bal
MaryamZi: Conclusion was to not add. We're trying to explain the function rather than show the user how to write code.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -247,6 +335,28 @@ public type OrderedType ()|boolean|int|float|decimal|string|OrderedType[];
 # must be specified.
 # Sorting works the same as with the parameter `sort` clause of query expressions.
 #
+# ```
```
MaryamZi, line 338: Suggestion from James

```ballerina
["c", "a", "B"].sort(key = string:toLowerAscii)
```

langlib/lang.array/src/main/ballerina/array.bal
SasinduDilshara: If we use option 4,
user can copy paste the code and use in both shell and in vscode(after wrap the code in main method).

langlib/lang.array/src/main/ballerina/array.bal
manuranga: Is using `toBalString` any better?

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -82,6 +102,14 @@ public isolated function 'map(Type[] arr, @isolatedParam function(Type val) retu
 #
 # The parameter `func` is applied to each member of parameter `arr` in order.
 #
+# ```
+# string str = "0";
+# [1, 2, 3].forEach(function (int i) {
+#     str += string ` ${i}`;
+# });
+# str ⇒ 0 1 2 3
```
SasinduDilshara, line 110:
```suggestion
# str ⇒ "0 1 2 3"
```
wdyt?

SasinduDilshara (changes requested):

langlib/lang.array/src/main/ballerina/array.bal
manuranga: No, the idea is to minimally represent what is done by the given langlib function. This require users to understand string interpolation as well. Maybe better to even remove the string concat and just + the ints.
eg:
`sum = sum + i`

langlib/lang.array/src/main/ballerina/array.bal
dulajdilshan: How about this option? This is will not break the code snippet ( we can copy and paste the code as it is)
```ballerina
[1, 2, 3, 4].length() // => 4
```

mohanvive: I think, better we use a different example for this. Either some strings in a list, just to avoid the confusion.

MaryamZi: This was already updated - https://github.com/ballerina-platform/ballerina-lang/pull/38989/commits/c88f6f680f4652a909d513b21e57e9688b4a1ce2#diff-2c948d23bf7d2faa264a43f44ca32e7f64f06b4dfb220764247218037b2d10f4R41

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -82,6 +102,14 @@ public isolated function 'map(Type[] arr, @isolatedParam function(Type val) retu
 #
 # The parameter `func` is applied to each member of parameter `arr` in order.
 #
+# ```
+# int total = 0;
+# [1, 2, 3].forEach(function (int i) {
+#     total += i * 2;
```
SasinduDilshara, line 108: I think it is more simple if we use `total += i`. It will also match with the variable name `total`

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -70,6 +85,11 @@ public isolated function enumerate(Type[] arr) returns [int, Type][] = @java:Met
 
 # Applies a function to each member of an array and returns an array of the results.
 #
+# ```
+# int[] numbers = [0, 1, 2];
+# numbers.map(n => n * 2) ⇒ [0,2,4]
```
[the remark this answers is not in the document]
MaryamZi: This is because we decided to use a separate variable when the value is used with an arrow function (since the parameter type is inferred.)

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -169,6 +218,11 @@ public isolated function every(Type[] arr, @isolatedParam function(Type val) ret
 # This removes the member of parameter `arr` with index parameter `index` and returns it.
 # It panics if there is no such member.
 #
+# ```
+# int[] numbers = [2, 4, 6, 8];
+# numbers.remove(1) ⇒ 4
```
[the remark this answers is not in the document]
MaryamZi: This doesn't work as expected, does it? A tuple is inferred here and therefore this will panic.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -180,6 +234,13 @@ public isolated function remove(Type[] arr, int index) returns Type = @java:Meth
 # Removes all members of an array.
 #
 # Panics if any member cannot be removed.
+#
+# ```
+# int[] numbers = [2, 4, 6, 8];
+# numbers.removeAll();
```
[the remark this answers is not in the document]
MaryamZi: Same as https://github.com/ballerina-platform/ballerina-lang/pull/38989#discussion_r1048317138.

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -274,6 +382,17 @@ public isolated function pop(Type[] arr) returns Type = @java:Method {
...
 # Adds values to the end of an array.
 #
+# ```
+# int[] numbers = [2];
+#
+# numbers.push(4, 6);
+# numbers ⇒ [2,4,6]
+#
+# int[] moreNumbers = [8, 10, 12, 14];
+# numbers.push(...moreNumbers);
```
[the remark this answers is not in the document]
MaryamZi: Since we have the example in L388, IMO it makes sense to have a separate variable here.

SasinduDilshara (changes requested):

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -316,6 +451,11 @@ public isolated function unshift(Type[] arr, Type... vals) returns () = @java:Me
 # The result will contain only characters  `A..Z`, `a..z`, `0..9`, `+`, `/` and `=`.
 # There will be no whitespace in the returned string.
 #
+# ```
+# byte[] byteArray = [104, 101, 108, 108, 111, 32, 98, 97, 108, 108, 101, 114, 105, 110, 97, 32, 33, 33, 33];
```
SasinduDilshara, line 455: Isn't it better if we make this array short?

SasinduDilshara (changes requested):

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -169,6 +218,11 @@ public isolated function every(Type[] arr, @isolatedParam function(Type val) ret
 # This removes the member of parameter `arr` with index parameter `index` and returns it.
 # It panics if there is no such member.
```
SasinduDilshara, line 219: Need and example for panic situation

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -180,6 +234,13 @@ public isolated function remove(Type[] arr, int index) returns Type = @java:Meth
 # Removes all members of an array.
 #
 # Panics if any member cannot be removed.
```
SasinduDilshara, line 236: Need an example for panic situations

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -265,6 +368,11 @@ public isolated function sort(Type[] arr, SortDirection direction = ASCENDING,
 #
 # The array must not be empty.
```
SasinduDilshara, line 369: Need an example for panic situation

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -289,6 +408,11 @@ public isolated function push(Type[] arr, Type... vals) returns () = @java:Metho
...
```
SasinduDilshara, line 409: Need an example for panic situation

SasinduDilshara (changes requested):

langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -104,11 +137,9 @@ public isolated function filter(Type[] arr, @isolatedParam function(Type val) re
 # The combining function takes the combined value so far and a member of the array,
 # and returns a new combined value.
 #
-# For example
 # ```
-# reduce([1, 2, 3], function (int total, int n) returns int { return total + n; }, 0)
+# [1, 2, 3].reduce(isolated function (int total, int next) returns int => total + next, 0) ⇒ 6
```
dulajdilshan, line 141:
```suggestion
# [1, 2, 3].reduce(isolated function (int accTot, int next) returns int => accTot + next, 0) ⇒ 6
```
This is the accumulated total. So how about using `accTot` instead of `total`?

MaryamZi: How is accTotal different from total?

MaryamZi:
New PR to 2201.3.x - https://github.com/ballerina-platform/ballerina-lang/pull/39073

Issue #38939, opened 2022-12-05
[Task]: Add examples to streams langlib operations


mohanvive:
### Description

$subject

### Related area

-> Compilation

Issue #34934, opened 2022-02-03
`StreamImplementor` has a default behaviour for calling `next` after `close` method


KRVPerera:
**Description:**
> A missing close method behaves like a close method that puts the object into a closed state,in which calls to 'next' will result in a panic, and then immediately returns ().

But current behaviour is a silent ignore and subsequent next operations does not panic. It is currently a responsibility of the stream implementor to decide the behaviour. 

#### 5.7.3 StreamImplementor
An object belongs to the object type StreamImplementor<T,C> if it belongs to Iterator<T,C> and also optionally has a method close() with return value C?. This is equivalent to belonging to the following type.
```Ballerina
    object {
       public isolated function next() returns record {| T value; |}|C;
    }
    | object {
       public isolated function next() returns record {| T value; |}|C;
       public isolated function close() returns C?;
    }
```
The close method says that there will be no more calls to the next method. Any call to the next method after the close method has been called must result in a panic. `A missing close method behaves like a close method that puts the object into a closed state, in which calls to 'next' will result in a panic, and then immediately returns ().`

#### lang.stream
```ballerina
# Closes a stream.
# This releases any system resources being used by the stream.
# Closing a stream that has already been closed has no effect and returns `()`.
```
**Related Issues (optional):**
- https://github.com/ballerina-platform/ballerina-lang/issues/25450
- https://github.com/ballerina-platform/ballerina-spec/issues/592
- https://github.com/ballerina-platform/ballerina-spec/issues/1036

KRVPerera:
With James's clarification we need to implement a closed state for `StreamImplementor`. Check the related spec issues linked here.

PR #39052, opened 2022-12-14
Add examples related to stream operations

mohanvive:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/38939


langlib/lang.stream/src/main/ballerina/stream.bal
```diff
@@ -43,6 +43,12 @@ type Type1 any|error;
 
 # Selects the members from a stream for which a function returns true.
 #
+# ```ballerina
+# stream<int> scoreStream = [45, 60, 75, 30, 90].toStream();
+# stream<int> filtered = scoreStream.filter(score => score > 50);
+# filtered.next() ⇒ {"value":60}
```
MaryamZi, line 49: Going with the suggestion for array functions, shouldn't this be

```suggestion
# scoreStream.filter(score => score > 50).next() ⇒ {"value":60}
```

MaryamZi: Can't we also just use something like `strm` as the name?

langlib/lang.stream/src/main/ballerina/stream.bal
```diff
@@ -55,6 +61,11 @@ public isolated function filter(stream<Type,CompletionType> stm, @isolatedParam
 
 # Returns the next element in the stream wrapped in a record or () if the stream ends.
 #
+# ```ballerina
+# stream<string> greetingStream = ["Hello", "Bonjour", "Hola", "Ciao"].toStream();
+# greetingStream.next() ⇒ {"value":"Hello"}
```
MaryamZi, line 66: Shall we change this also to use a stream of numbers?

langlib/lang.stream/src/main/ballerina/stream.bal
```diff
@@ -148,8 +187,14 @@ public isolated function iterator(stream<Type,CompletionType> stm) returns objec
 
 # Closes a stream.
 #
+# ```ballerina
+# stream<int> scoreStream = [45, 60, 75, 30, 90].toStream();
+# _ = scoreStream.close();
+# scoreStream.next() ⇒ {"value":45}
```
MaryamZi, line 193: Do we need this?

MaryamZi: Is the current implementation correct btw?

The spec says

> The close method says that there will be no more calls to the next method. Any call to a next method after the close method has been called must result in a panic. A missing close method behaves like a close method that puts the object into a closed state, in which calls to next will result in a panic, and then immediately returns ().

MaryamZi: May be related to https://github.com/ballerina-platform/ballerina-lang/issues/34934.

mohanvive: Yes, I wanted to have a chat on this. It seems like an issue.

langlib/lang.stream/src/main/ballerina/stream.bal
mohanvive: yes, will do.

langlib/lang.stream/src/main/ballerina/stream.bal
mohanvive: hmm. I kind of prefer having proper names. :)

mohanvive: BTW, what if we have the earlier version?

```
stream<int> filtered = scoreStream.filter(score => score > 50);
filtered.next() ⇒ {"value":60}
```

I feel like it is better, something similar to below array case,

```
object {
    public isolated function next() returns record {|int value;|}?;
} iterator = [2, 4, 6, 8].iterator();
iterator.next() ⇒ {"value":2}
```

WDYT?

MaryamZi: > hmm. I kind of prefer having proper names. :)

Can we have something like `scores` instead of `scoreStream`?



> BTW, what if we have the earlier version?
> 
> ```
> stream<int> filtered = scoreStream.filter(score => score > 50);
> filtered.next() ⇒ {"value":60}
> ```
> 
> I feel like it is better, something similar to below array case,
> 
> ```
> object {
>     public isolated function next() returns record {|int value;|}?;
> } iterator = [2, 4, 6, 8].iterator();
> iterator.next() ⇒ {"value":2}
> ```
> 
> WDYT?

The issue with including `stream<int> filtered = scoreStream.filter(score => score > 50);` though is that we kinda expose the type param behaviour, right? Which we haven't done with others.

langlib/lang.stream/src/main/ballerina/stream.bal
```diff
@@ -76,6 +86,12 @@ public isolated function next(stream<Type, CompletionType> stm) returns record {
 
 # Applies a function to each member of a stream and returns a stream of the results.
 #
+# ```ballerina
+# stream<float> ms = [14.5f, 45.5f, 6.8f, 4f].toStream();
+# stream<float> cms = ms.'map(m => m * 100.0);
```
MaryamZi, line 91:
```suggestion
# stream<float> cms = ms.map(m => m * 100.0);
```

langlib/lang.stream/src/main/ballerina/stream.bal
```diff
@@ -148,8 +186,13 @@ public isolated function iterator(stream<Type,CompletionType> stm) returns objec
 
 # Closes a stream.
 #
+# ```ballerina
+# stream<int> scoreStream = [45, 60, 75, 30, 90].toStream();
+# _ = scoreStream.close();
+# ```
+#
```
MaryamZi, line 193: Example needs to be after the function description, before the param descriptions.

mohanvive: My bad. Fixed.

langlib/lang.stream/src/main/ballerina/stream.bal
mohanvive: > > hmm. I kind of prefer having proper names. :)
> 
> Can we have something like `scores` instead of `scoreStream`?

Hmm, OK. 
> 
> > BTW, what if we have the earlier version?
> > ```
> > stream<int> filtered = scoreStream.filter(score => score > 50);
> > filtered.next() ⇒ {"value":60}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I feel like it is better, something similar to below array case,
> > ```
> > object {
> >     public isolated function next() returns record {|int value;|}?;
> > } iterator = [2, 4, 6, 8].iterator();
> > iterator.next() ⇒ {"value":2}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > WDYT?
> 
> The issue with including `stream<int> filtered = scoreStream.filter(score => score > 50);` though is that we kinda expose the type param behaviour, right? Which we haven't done with others.

Hmm. I feel it is fine.

langlib/lang.stream/src/main/ballerina/stream.bal
```diff
@@ -149,7 +187,12 @@ public isolated function iterator(stream<Type,CompletionType> stm) returns objec
 # Closes a stream.
 #
 # This releases any system resources being used by the stream.
-# Closing a stream that has already been closed has no efffect and returns `()`.
+# Closing a stream that has already been closed has no effect and returns `()`.
+#
+# ```ballerina
+# stream<int> scores = [45, 60, 75, 30, 90].toStream();
+# _ = scores.close();
```
pcnfernando, line 194: Shall we add another sample with a completion other than (). wdyt?
```ballerina
function name(stream<int, error?> strm) {
    error? r = strm.close();
}
```

mohanvive: @pcnfernando do you have a compact example to show this (instead of above)?

pcnfernando: No, if we are to create a steeam with other completions it will be much lengthier. Hence, this would be the most compact version as I see it

mohanvive: ya, guessed it. I don't prefer above example though 😄

MaryamZi: How about

```
stream<int, error?> strm = new;
error? res = strm.close();
```

although new doesn't really help.

Generally, the completion type will include error right? So may be helpful to not emphasize (just) ignoring with `_`.

mohanvive: IMO, I don't think this really helps. But I am OK to add this since we wanted to show the error completion type. Will change it then.

mohanvive: I'll go with this,

```
    stream<int, error?> strm = new;
    _ = check strm.close();
```

pcnfernando: We don't need to ignore (). Hence, shall we do either
```ballerina
    stream<int, error?> strm = new;
    check strm.close();
```
or
```ballerina
    stream<int, error?> strm = new;
    error? res = strm.close();
```

mohanvive: Let's go with 1st option then.

pcnfernando (approved):

PR #38926, opened 2022-12-02
Add initial examples related to string operations

mohanvive:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/38915

PR #39068, opened 2022-12-15
[2201.3.x] Add examples related to string operations

mohanvive:
Fixes https://github.com/ballerina-platform/ballerina-lang/pull/38926

### DIFF
--- a/langlib/lang.string/src/main/ballerina/string.bal
+++ b/langlib/lang.string/src/main/ballerina/string.bal
@@ -310,7 +310,7 @@ public isolated function toBytes(string str) returns byte[] = @java:Method {
 #
 # ```ballerina
 # string:fromBytes([72, 101, 108, 108, 111, 32, 66, 97, 108, 108, 101, 114, 105, 110, 97, 33]) ⇒ Hello, World!
-# string:fromBytes([12, 45, 45]) ⇒ error
+# string:fromBytes(base64 `lang`) ⇒ error
 # ```
 #
 # + bytes - UTF-8 byte array
--- a/langlib/lang.string/src/main/ballerina/string.bal
+++ b/langlib/lang.string/src/main/ballerina/string.bal
@@ -25,8 +25,9 @@ public type RegExp regexp:RegExp;
 
 # Returns the length of the string.
 #
-# ```ballerina
-# "Hello, World!".length() ⇒ 13;
+# ```
+# string greeting = "Hello, World!";
+# greeting.length() ⇒ 13;
 # ```
 #
 # + str - the string
@@ -89,11 +90,8 @@ public isolated function getCodePoint(string str, int index) returns int = @java
 # Returns a substring of a string.
 #
 # ```ballerina
-# "Hello, my name is John".substring(18, 22) ⇒ "John"
-#
-# // Consider only the start index. In this situation,
-# // input string length would be considered as end index.
 # "Hello, my name is John".substring(7) ⇒ "my name is John"
+# "Hello, my name is John".substring(18, 22) ⇒ "John"
 # ```
 #
 # + str - source string.
@@ -128,10 +126,8 @@ public isolated function codePointCompare(string str1, string str2) returns int 
 # Joins zero or more strings together with a separator.
 #
 # ```ballerina
-# // Join multiple strings with space as the separator.
 # string:'join(" ", "Ballerina", "is", "a", "programming", "language") ⇒ "Ballerina is a programming language"
 #
-# // Join multiple strings with comma as the separator.
 # string[] array = ["John", "23", "USA", "Computer Science", "Swimmer"];
 # string:'join(",", ...array) ⇒ "John,23,USA,Computer Science,Swimmer"
 # ```
--- a/langlib/lang.string/src/main/ballerina/string.bal
+++ b/langlib/lang.string/src/main/ballerina/string.bal
@@ -444,8 +444,8 @@ public isolated function padZero(string str, int len, Char zeroChar = "0") retur
 # + str - the string
 # + re - the regular expression
 # + return - true if there is full match of `re` with `str`, and false otherwise
-public isolated function matches(string str, RegExp 're) returns boolean {
-   return 're.isFullMatch(str);
+public function matches(string str, RegExp re) returns boolean {
+   return re.isFullMatch(str);
 }
 
 # Tests whether there is a match of a regular expression somewhere within a string.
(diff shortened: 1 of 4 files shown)


langlib/lang.string/src/main/ballerina/string.bal
```diff
@@ -309,27 +432,32 @@ public isolated function padZero(string str, int len, Char zeroChar = "0") retur
...
+# "New Zealand".includesMatch(re `lan?`) ⇒ true
+# ```
+#
 # + str - the string to be matched
 # + re - the regular expression
 # + return - true if the is a match of `re` somewhere within `str`, otherwise false
-public function includesMatch(string str, RegExp re, int startIndex = 0) returns boolean {
-   return re.find(str, startIndex) != ();
+public isolated function includesMatch(string str, RegExp 're, int startIndex = 0) returns boolean {
+   return 're.find(str, startIndex) != ();
```
mohanvive, line 462: @pcnfernando Are there any difference between the these regex functions in master and 3.x ?

mohanvive: Updated as per the offline chat.

langlib/lang.string/src/main/ballerina/string.bal
```diff
@@ -54,6 +75,10 @@ public isolated function concat(string... strs) returns string = @java:Method {
 
 # Returns the code point of a character in a string.
 #
+# ```ballerina
+# "Hello, World!".getCodePoint(3) ⇒ 108
+# ```
```
pcnfernando, line 80:
```suggestion
# ```ballerina
# "Hello, World!".getCodePoint(3) ⇒ 108
# string:getCodePoint("Hello, World!", 3) ⇒ 108
# ```
```

langlib/lang.string/src/main/ballerina/string.bal
```diff
@@ -54,6 +75,10 @@ public isolated function concat(string... strs) returns string = @java:Method {
 
 # Returns the code point of a character in a string.
 #
+# ```ballerina
+# "Hello, World!".getCodePoint(3) ⇒ 108
```
pcnfernando, line 79: Should we have both func call and method call in the examples?

langlib/lang.string/src/main/ballerina/string.bal
```diff
@@ -114,6 +162,13 @@ public isolated function indexOf(string str, string substr, int startIndex = 0)
 
 # Finds the last occurrence of one string in another string.
 #
+# ```ballerina
+# "Ballerinalang is a unique programming language".lastIndexOf("lang") ⇒ 38
+#
+# // Search for the last occurrence of a string from a specific index onwards.
```
pcnfernando, line 168: Should we need to have an explanation in the example as we have it in the parameter description?
```suggestion
# // Search for the last occurrence of a string from a specific index backwards.
```

langlib/lang.string/src/main/ballerina/string.bal
mohanvive: Decided to follow above approach.

langlib/lang.string/src/main/ballerina/string.bal
```diff
@@ -64,6 +89,11 @@ public isolated function getCodePoint(string str, int index) returns int = @java
 
 # Returns a substring of a string.
 #
+# ```ballerina
+# "Hello, my name is John".substring(7) ⇒ my name is John
+# "Hello, my name is John Anderson".substring(18, 22) ⇒ John
```
pcnfernando, line 94:
```suggestion

// Retrieve the substring between 18(inclusive) and 22 indexes
# "Hello, my name is John Anderson".substring(18, 22) ⇒ John
```

langlib/lang.string/src/main/ballerina/string.bal
```diff
@@ -200,6 +282,10 @@ public isolated function equalsIgnoreCaseAscii(string str1, string str2) returns
 #
 # The ASCII white space characters are 0x9...0xD, 0x20.
 #
+# ```ballerina
+# " BALLERINA  ".trim() ⇒ BALLERINA
```
pcnfernando, line 286: I think the output doesn't convey that trailing space was removed. Because of cases like these, should we consider enclosing the output within ""

mohanvive: hmm, that's a valid point.

langlib/lang.string/src/main/ballerina/string.bal
mohanvive: @pcnfernando Isn't the details given in the comment covered by the description of the `return`.

pcnfernando: Yes, true. But I thought explaining it in the example itself will be useful. wdyt?

mohanvive: Actually, we eliminated such comments which are already covered by the description and etc. I believe let's avoid this just to keep it synced with other such example.

langlib/lang.string/src/main/ballerina/string.bal
mohanvive: Will change this.

mohanvive: BTW, there are a few cases like this which I have to show that spaces are available at the end or start of the string. 
There are two options to show this.
1. Include within double quotes
2. Append with another string and show the output.

What is your preference? Thoughts?

pcnfernando: Since other output doesn't have ""; shall we then append a string? maybe "" start and end? wdyt?

mohanvive: hmm, I don't have a strong opinion though. But slightly prefer option #1 :). @MaryamZi your thoughts?

mohanvive: It is not only about the `trim` function. We need to consider some other string functions as well, such as `padZero`, `padStart` and `padEnd` cases as well.  Do you guys think appending looks nice for all of them?

MaryamZi: But in those examples, unless we pad with whitespace, it is comparatively clear, right? IMO, those are OK as is.

mohanvive: OK, fine then.

mohanvive: Thanks, will update accordingly.

MaryamZi (dismissed):

pcnfernando (approved):

PR #39101, opened 2022-12-17
 Generate a typedesc instruction once for record and tuple types

### LINKED ISSUE #38844: [Task]: Generate `new typedesc` instruction for record and tuple when `type descriptor` resolving
### Description

In the current implementation, new typedesc instruction generates when value is created.
When considering the following example. we create two new typedesc instructions for `record{int k = 10;}` type.
```ballerina
function foo() {
    record{int k = 10;} m;
    m = {};
    m = {k: 15};
}
```

### Describe your task(s)

_No response_

### Related area

-> Compilation

_No response_

_No response_

_No response_

chiranSachintha:
Fixes #38844 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1101,6 +1123,10 @@ public void visit(BLangSimpleVariableDef astVarDefStmt) {
             this.currentScope = newScope;
         }
 
+        if (astVarDefStmt.var.typeNode != null) {
+            astVarDefStmt.var.typeNode.accept(this);
```
MaryamZi, line 1127: Extract `astVarDefStmt.var.typeNode` out to a variable.

MaryamZi: Please update wherever applicable.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1134,6 +1160,106 @@ public void visit(BLangSimpleVariable varNode) {
...
+
+    @Override
+    public void visit(BLangErrorType errorType) {
+        if (errorType.detailType != null) {
+            errorType.detailType.accept(this);
+        }
+    }
+
+    @Override
+    public void visit(BLangFunctionTypeNode functionTypeNode) {
```
MaryamZi, line 1209: Don't we have to visit parameter and return types?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1134,6 +1160,106 @@ public void visit(BLangSimpleVariable varNode) {
...
+        keyTypeConstraint.keyType.accept(this);
+    }
+
+    @Override
+    public void visit(BLangObjectTypeNode objectTypeNode) {
+        for (BLangSimpleVariable field : objectTypeNode.fields) {
+            if (field.typeNode != null) {
+                field.typeNode.accept(this);
+            }
+        }
```
MaryamZi, line 1252: How about the param/return types in methods?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1134,6 +1160,106 @@ public void visit(BLangSimpleVariable varNode) {
...
+    }
+
+    @Override
+    public void visit(BLangFiniteTypeNode finiteTypeNode) {
+    }
+
+    @Override
+    public void visit(BLangIntersectionTypeNode intersectionTypeNode) {
```
MaryamZi, line 1260: Why don't we have to visit the constituent types?

MaryamZi: Where is the effective type visited? Is it via the type definitions?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
```diff
@@ -1963,8 +1965,10 @@ public void visit(BLangRecordLiteral.BLangMapLiteral mapLiteral) {
     public void visit(BLangRecordLiteral.BLangStructLiteral structLiteral) {
         SymbolEnv symbolEnv = env.createClone();
         BLangFunction enclInvokable = (BLangFunction) symbolEnv.enclInvokable;
-        structLiteral.enclMapSymbols = collectClosureMapSymbols(symbolEnv, enclInvokable, false);
-
+        if (structLiteral.getBType().getKind() == TypeKind.RECORD) {
+            ((BRecordType) structLiteral.getBType()).enclMapSymbols =
+                                                     collectClosureMapSymbols(symbolEnv, enclInvokable, false);
+        }
```
MaryamZi, line 1971: Why can't we do this when visiting the type?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/model/BIRNonTerminator.java
```diff
@@ -192,20 +192,23 @@ public BIROperand[] getRhsOperands() {
     public static class NewStructure extends BIRNonTerminator {
         public BIROperand rhsOp;
         public List<BIRMappingConstructorEntry> initialValues;
+        public BType type;
```
MaryamZi, line 195: @warunalakshitha, is it OK to have AST/FE types in the BIR? There are some other places where this is already done, but just wondering if it is ideal.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
```diff
@@ -2099,12 +2105,9 @@ void generateNewTypedescIns(BIRNonTerminator.NewTypeDesc newTypeDesc) {
     }
 
     private boolean isNonReferredRecord(BType type) {
```
MaryamZi, line 2107: Isn't this now a check for a module-level record? Please update the name appropriately.

chiranSachintha:
**Updated the logic as follows:**

-     Create type instances
-     Create type constants
-     Create typedesc instances for package level type definitions
Then,
-     Use the generated typedesc to create a typedescvalue.
-     If there is a typedesc defined within the function, generate a typedesc instruction when visiting the expr for the first time and keep a reference to the generated typedesc. Use the reference if necessary."

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Changed the logic https://github.com/ballerina-platform/ballerina-lang/pull/39101/commits/ed9d56f60b01a2744651c7e805361dd574b8357d

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
chiranSachintha: Updated

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Updated https://github.com/ballerina-platform/ballerina-lang/pull/39101/commits/ed9d56f60b01a2744651c7e805361dd574b8357d

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Removed this logic.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Removed

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
chiranSachintha: Updated

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
MaryamZi: What about

```ballerina
public function main() {
    function (record {| float f; |}) returns record {| string b = "hello"; |} y;
}
```

Where are the new typedesc instructions generated for these records?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
MaryamZi: Still the same name?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
MaryamZi: Has the PR been updated?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/TypeUtils.java
```diff
@@ -159,4 +160,10 @@ public static Type getReferredType(Type type) {
         return constraint;
     }
 
+    public static Type getEffectiveType(Type type) {
```
MaryamZi, line 163: Does this have to be added to the public API?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -737,6 +758,20 @@ private BIRVariableDcl getSelf(BSymbol receiver) {
         return self;
     }
 
+    private void analyzeParametersAndReturnType(BLangFunction astFunc) {
+        if (astFunc.returnTypeNode != null) {
```
MaryamZi, line 762: Extract repeated access out to variables.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/JvmInstructionGen.java
chiranSachintha: Introduce a separate function called `isUserDefinedTypeDefinition` in the `Types` class. Updated

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BTypeReferenceType.java
```diff
@@ -47,6 +47,13 @@ public BTypeReferenceType(String typeName, Module pkg, int typeFlags, boolean re
         this.readOnly = readOnly;
     }
 
+    public BTypeReferenceType(Type referredType) {
+        super(referredType.getName(), referredType.getPkg(), Object.class);
+        this.typeFlags = (int) referredType.getFlags();
+        this.referredType = referredType;
+        this.readOnly = true;
+    }
```
HindujaB, line 55:
```suggestion
    public BTypeReferenceType(Type referredType) {
        this(referredType.getName(), referredType.getPkg(), (int) referredType.getFlags(), true);
        setReferredType(referredType);
  ```

File: compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java (hunk 5 above)
Comment (chiranSachintha) [reply compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java:1971]: I forgot to update the comment. The type definitions are resolving before visiting functions, so all the enclosing map symbols cannot be captured at that stage.

File: bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/TypedescValueImpl.java [hunk 11]
```diff
@@ -91,6 +93,9 @@ public Object instantiate(Strand s, BInitialValueEntry[] initialValues) {
             return new MapValueImpl(this.describingType, (BMapInitialValueEntry[]) initialValues);
         } else if (referredType.getTag() == TypeTags.TUPLE_TAG) {
             return new TupleValueImpl(this.describingType, (BListInitialValueEntry[]) initialValues);
+        } else if (getEffectiveType(referredType).getTag() == TypeTags.TUPLE_TAG) {
```
HindujaB, line 96: Shall we store `getEffectiveType(referredType)` result in a local variable and reuse it instead of calling the method twice?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Didn't generate `typedesc` for this because there is no usage of defining typedescs for these records.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
MaryamZi: That should be the case only for module-level definitions and anon records, right? For local anon records, they should be visited when the function is being visited?

MaryamZi: Doesn't this also mean `enclMapSymbols` isn't properly set for local anon records if a value isn't created using that record typedesc?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
MaryamZi: What do you mean no usage?

```ballerina
import ballerina/io;

annotation record {| float f = foo(); |} annot on record field;

public function main() {
    function (record {| @annot float f; |}) r;
}

isolated function foo() returns float {
    io:println("foo");
    return 1.0;
}
```

This should print `"foo"`, but it doesn't atm.

Either way, we shouldn't make assumptions about usage. The spec defines when the typedesc is resolved, and we have to resolve it at that point.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/TypedescValueImpl.java
```diff
@@ -91,6 +93,9 @@ public Object instantiate(Strand s, BInitialValueEntry[] initialValues) {
...
+            return new TupleValueImpl(new BTypeReferenceType(getEffectiveType(referredType)),
```
HindujaB, line 97: Why do we create a new `TypeReferenceType` and pass it? 
IIRC, as the `getEffectiveType(referredType)` is expected to be a TupleType here, we can pass the `TupleType` itself.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/utils/TypeUtils.java
warunalakshitha: We can move this to internal package since we do not have requirement to expose.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/model/BIRNonTerminator.java
warunalakshitha: We already have BType in Bir java model and also type attributes are added to bir binary. IMO it should be ok

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/TypedescValueImpl.java
chiranSachintha:
```ballerina
type A [int, string|xml] & readonly;
public function main() {
    A tupleValue9 = [1, ""];

}
```
Previously, we did not generate a type reference for the intersection of a tuple with a readonly type. With these changes, for the above example, I have generated a type reference for type 'A'. Due to this, the referred type of type reference  is an intersection type, so I have created a new type reference with a tuple as the referred type

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Updated.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/TypedescValueImpl.java
chiranSachintha: Is it correct to generate two type descriptors for the following example?
```ballerina
type A [int, string|xml] & readonly;
public function main() {
    A tupleValue1 = [1, ""];
    A tupleValue2 = [1, ""];
}
```

MaryamZi: OK, if it for `A`, we'll need the type reference type. I misunderstood it as for the effective type (that we don't need to be the same).

chiranSachintha: > Is it correct to generate two typedescs for the following example?
> 
> ```ballerina
> type A [int, string|xml] & readonly;
> public function main() {
>     A tupleValue1 = [1, ""];
>     A tupleValue2 = [1, ""];
> }
> ```

HindujaB: Ack. 
In that case, shall we modify the `BTypeReferenceType` constructor to not to hardcode the readonly part?
```java
    public BTypeReferenceType(Type referredType, boolean readOnly)) {
    }
```

MaryamZi:
@chiranSachintha, please change the base to https://github.com/ballerina-platform/ballerina-lang/tree/annot-tuple-members.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
MaryamZi: I assume we need changes in https://github.com/ballerina-platform/ballerina-lang/pull/38488 for the sample I shared above to work as expected? We need to make sure we have tests for every possible type node scenario.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
MaryamZi: Ping.

chiranSachintha: Yes, When traversing a function, local record-type nodes are visited. But the problem is when visiting a record-type node, it is not known which enclosure map symbols are necessary for a specific record literal.
```ballerina
function value() {
    record{int x;} m;
    int z = 10;
    var x = function() {
            int k = z;
            var y = function() {
                m = {x:k+z};
            };
    };
}
```
Consider the above example, When visiting m, cannot identify what enclosure map symbols are needed for the record literal.

MaryamZi: Do you mean mapping constructor? Is `enclMapSymbols` relevant to the type or the value? If it is the latter, moving it to the type is wrong.

chiranSachintha: Updated with https://github.com/ballerina-platform/ballerina-lang/pull/39101/commits/c0ef0a550116ffe5b0a0eaf59a6177e39afbbaaf

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/util/RuntimeUtils.java
```diff
@@ -247,4 +248,11 @@ private static boolean isInvalidBallerinaValue(Object value) {
 
     private RuntimeUtils() {
     }
+
+    public static Type getEffectiveType(Type type) {
+        if (type.getTag() == TypeTags.INTERSECTION_TAG) {
+            return ((BIntersectionType) type).getEffectiveType();
+        }
+        return type;
```
MaryamZi, line 256: Codecov warnings?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
```diff
@@ -963,6 +965,19 @@ public void visit(BLangArrayType arrayType) {
 
     @Override
     public void visit(BLangUserDefinedType userDefinedType) {
+        BType type = userDefinedType.getBType();
+        if (type == null) {
```
MaryamZi, line 969: When can the type be null?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
```diff
@@ -963,6 +965,19 @@ public void visit(BLangArrayType arrayType) {
...
+            result = userDefinedType;
+            return;
+        }
+        if (type.tag == TypeTags.RECORD) {
+            ((BRecordType) type).enclMapSymbols = collectClosureMapSymbols(env, false);
```
MaryamZi, line 974: So `enclMapSymbols` are relevant to the type?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
```diff
@@ -963,6 +965,19 @@ public void visit(BLangArrayType arrayType) {
...
+        } else if (type.tag == TypeTags.ERROR) {
+            BType detailType = ((BErrorType) type).detailType;
+            if (detailType.tag == TypeTags.RECORD) {
+                ((BRecordType) detailType).enclMapSymbols = collectClosureMapSymbols(env, false);
+            }
+        }
```
MaryamZi, line 980: What about other user-defined types that may have the record? Don't we get here for them?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -5415,6 +5416,14 @@ private BType getIntersectionForErrorTypes(IntersectionContext intersectionConte
         return intersectionErrorType;
     }
 
+    // This function checks whether the specified type definition is a user-defined type or not.
```
MaryamZi, line 5419: Doesn't add value? Unnecessary?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1074,12 +1116,17 @@ public void visit(BLangSimpleVariableDef astVarDefStmt) {
             this.currentScope = newScope;
         }
 
-        if (astVarDefStmt.var.expr == null) {
+        BLangSimpleVariable simpleVariable = astVarDefStmt.var;
+        if (simpleVariable.typeNode != null) {
+            simpleVariable.typeNode.accept(this);
```
MaryamZi, line 1121: typeNode can be extracted out to a variable.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1107,6 +1154,135 @@ public void visit(BLangSimpleVariable varNode) {
         env.enclPkg.isListenerAvailable |= Symbols.isFlagOn(varNode.symbol.flags, Flags.LISTENER);
     }
 
+    @Override
+    public void visit(BLangUserDefinedType userDefinedType) {
+        if (Types.getReferredType(userDefinedType.getBType()).getKind() == TypeKind.RECORD) {
+            createNewTypeDescInst(userDefinedType.getBType());
```
MaryamZi, line 1160: Why do we need to do this? When do we get here?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
MaryamZi: Applicable in other places too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1107,6 +1154,135 @@ public void visit(BLangSimpleVariable varNode) {
...
+    public void visit(BLangUnionTypeNode unionTypeNode) {
+        unionTypeNode.memberTypeNodes.forEach(typeNode -> typeNode.accept(this));
+    }
+
+    @Override
+    public void visit(BLangRecordTypeNode recordTypeNode) {
+        createNewTypeDescInst(recordTypeNode.getBType());
+        for (BLangSimpleVariable field : recordTypeNode.fields) {
+            if (field.typeNode != null) {
+                field.typeNode.accept(this);
```
MaryamZi, line 1194: Codecov warnings. Fields of record types are a common scenario too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1107,6 +1154,135 @@ public void visit(BLangSimpleVariable varNode) {
...
+        }
+    }
+
+    @Override
+    public void visit(BLangTableKeyTypeConstraint keyTypeConstraint) {
+        keyTypeConstraint.keyType.accept(this);
+    }
+
+    @Override
+    public void visit(BLangObjectTypeNode objectTypeNode) {
```
MaryamZi, line 1267: Fields?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
chiranSachintha: Yes. This is used to maintain the references of variables within the 'record' type node.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Updated.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: In the desugaring phase, we created a user-defined type for the local record. I have now changed it to retain it as a record-type node without changing it to a user-defined type node

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
chiranSachintha: Updated. Earlier, In the desugaring phase, we created a user-defined type for the local record.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/util/RuntimeUtils.java
chiranSachintha: Fixed with current changes

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Updated.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Earlier, in the desugaring phase, we created a user-defined type for the local record. I have now changed it to retain it as a record type without changing it to a user-defined type.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/util/RuntimeUtils.java
chiranSachintha: With the current changes, the issue has been resolved.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
chiranSachintha: Updated

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
MaryamZi: Any update on this?

MaryamZi:
The build is failing?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
chiranSachintha: Move this logic to record type node visitor.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/ClosureDesugar.java
chiranSachintha: This logic moves to record type node visitor. https://github.com/ballerina-platform/ballerina-lang/pull/39101/commits/0b954eeccda6e6f7a813c79b9edab058f4c3d159

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Removed this logic as it is no longer needed with recent changes in the implementation(earlier this was needed because In the desugaring phase, we created a user-defined type for the local record).

chiranSachintha:
> The build is failing?
The build failure has been resolved.

tests/jballerina-unit-test/src/test/resources/test-src/bir/bir-dump/failLockWithinLock
```diff
@@ -116,4 +115,4 @@ failLockWithinLock function() -> (int, string) {
     | bb2          | bb11         | bb12         | %4           |
     | bb3          | bb3          | bb4          | %7           |
     -------------------------------------------------------------
-}
+}
```
gabilang, line 118: Shall we add the missing new line here?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -710,6 +741,26 @@ private BIRVariableDcl getSelf(BSymbol receiver) {
...
+                typeNode.accept(this);
+            }
+        }
+        BLangSimpleVariable restParam = astFunc.restParam;
+        if (restParam != null) {
+            typeNode = restParam.typeNode;
+            if (typeNode != null) {
+                typeNode.accept(this);
+            }
+        }
```
MaryamZi, line 761: I would visit these in order. Required params, rest param, return type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
MaryamZi: Did we add this as a test in the other PR?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1107,6 +1168,145 @@ public void visit(BLangSimpleVariable varNode) {
...
+        BLangVariable restParam = functionTypeNode.restParam;
+        if (restParam != null) {
+            typeNode = restParam.typeNode;
+            if (typeNode != null) {
+                typeNode.accept(this);
+            }
```
MaryamZi, line 1252: Same comment re: order. Also, isn't this duplicated logic?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1107,6 +1168,145 @@ public void visit(BLangSimpleVariable varNode) {
...
+    }
+
+    @Override
+    public void visit(BLangFiniteTypeNode finiteTypeNode) {
+    }
+
+    @Override
+    public void visit(BLangIntersectionTypeNode intersectionTypeNode) {
+        BType type = intersectionTypeNode.getBType();
+        if (type.tag != TypeTags.INTERSECTION) {
```
MaryamZi, line 1301: When is this false?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1107,6 +1168,145 @@ public void visit(BLangSimpleVariable varNode) {
...
+            return;
+        }
+        BType effectiveType = ((BIntersectionType) intersectionTypeNode.getBType()).effectiveType;
+        if (effectiveType.tag == TypeTags.RECORD || effectiveType.tag == TypeTags.TUPLE) {
+            createNewTypedescInst(effectiveType, intersectionTypeNode.pos);
```
MaryamZi, line 1306: Why aren't we visiting them instead of special casing just the instruction creation?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -1564,15 +1764,16 @@ public void visit(BLangLiteral astLiteralExpr) {
 
     @Override
     public void visit(BLangMapLiteral astMapLiteralExpr) {
-        visitTypedesc(astMapLiteralExpr.pos, astMapLiteralExpr.getBType(), Collections.emptyList());
+        BType type = astMapLiteralExpr.getBType();
+        createNewTypedescInst(type, astMapLiteralExpr.pos);
```
MaryamZi, line 1768: Why is this here? Wasn't the whole point moving typedesc construction to when resolving the type(desc)?

MaryamZi: If this is for the readonly field scenario, do a specific check here, rather than within `createNewTypedescInst`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
```diff
@@ -1243,22 +1243,16 @@ public void visit(BLangRecordTypeNode recordTypeNode) {
         // referenced types are invoked on the current record type.
 
         if (recordTypeNode.isAnonymous && recordTypeNode.isLocal) {
-            BLangUserDefinedType userDefinedType = desugarLocalAnonRecordTypeNode(recordTypeNode);
             TypeDefBuilderHelper.createTypeDefinitionForTSymbol(recordTypeNode.getBType(),
                     recordTypeNode.getBType().tsymbol, recordTypeNode, env);
             recordTypeNode.desugared = true;
-            result = userDefinedType;
+            result = recordTypeNode;
```
MaryamZi, line 1249: Why are we changing this? Didn't we discuss that this was correct?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Updated

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: Yeah, we can do that. Update logic to iterate through constituent types of intersection type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: This happens when the type becomes a type reference. Remove this logic as part of the new changes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: The reason for creating a 'typedesc' instruction is to differentiate a map from a record. We need that information when generating typedesc instructions. (For maps we need to create `TypedescValueImpl`).

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
chiranSachintha: We have previously discussed this also. But, If we have created a user-defined type, details related to fields will miss.  In my opinion, adding a new type definition, in this case, may not be necessary. Instead, the record type should be visited within the function.
```ballerina
function foo() {
    record {[int, int] l;} k;
}
```
In the above scenario typedesc of tuple should be created within the function.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
MaryamZi: My question is about why we are doing that when visiting the mapping constructor. Wasn't the conclusion that the typedesc has to be created when visiting the typedesc (except with readonly fields)? 

> We need that information when generating typedesc instructions. (For maps we need to create `TypedescValueImpl`).

Why can't we do this when visiting the type/typdesc?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: How do we lose field details? What we do is add a module-level record type definition and point to that, right?

```
type $anonRecord$0 record {[int, int] l;};

function foo() {
    $anonRecord$0 k;
}
```

MaryamZi: Right now, what you seem to have done is desugar to

```ballerina
type $anonRecord$0 record {[int, int] l;};

function foo() {
    record {[int, int] l;} k;
}
```

which introduces a disconnect?

What doesn't work without this change?

chiranSachintha: When visiting the type definition, do we visit the details of the record? If so, how can we create a typedesc for a tuple within a function?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha:
```ballerina
function value() {
    map<record{}> k;
}
```

You mean to generate the type descriptor(typedesc) when visiting the type node, right? In the above scenario, we need to create the type descriptor(typedesc) with a  `map<record{}>` type. In the generated constrained type node, the type is a map and constraint is record. Is there a way to obtain that map type(map<record{}>) from the constraint type node?

MaryamZi: In the generated constrained type node, the type is a map and constraint is record

Isn't the map type here the type you are looking for?

chiranSachintha:
```ballerina
function value() {
    map<record{int a = 10;}> k = {};
}
```
No, In this example type of constraint type node is set to builtin ref type(builtin Map type) but the type of the expression is `map<record {| int a; anydata...; |}>`. We need to create typedesc for `map<record {| int a; anydata...; |}>` type.

MaryamZi: What about `org.wso2.ballerinalang.compiler.tree.BLangNode#determinedType`? Please create an issue for the type being the builtin type.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: As discussed on the call, the issue is the disconnect, given that we are still adding the module-level definition.

chiranSachintha: With these changes, we only require a type definition to create the type. The current implementation of the BIR did not have a method for representing this and required the creation of a separate type definition. In the future, we can add these types to a separate list and store that information in the BIR in a different manner.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
chiranSachintha: https://github.com/ballerina-platform/ballerina-lang/issues/39288

MaryamZi: I assume we can use `determinedType` for now.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/Desugar.java
MaryamZi: We just need the type node when visiting the `BLangUserDefinedType` in `BIRGen`, right? Since this is always in the current module, I think we can look up the `BPackageSymbol` from `org.wso2.ballerinalang.compiler.semantics.model.SymbolTable#pkgEnvMap`, and get the type definition (and thereby type node) from `enclPkg`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
MaryamZi: @chiranSachintha, any update on this? Doesn't seem to work for me with the other PR also, will recheck.

MaryamZi: Can you send a draft PR with changes from both PRs, resolving conflicts?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -710,6 +742,26 @@ private BIRVariableDcl getSelf(BSymbol receiver) {
...
     }
 
+    private void analyzeParametersAndReturnType(List<BLangSimpleVariable> requiredParams, BLangVariable restParam,
+                                                BLangType returnTypeNode) {
+        BLangType typeNode;
+        for (BLangSimpleVariable parameter : requiredParams) {
+            typeNode = parameter.typeNode;
+            if (typeNode != null) {
+                typeNode.accept(this);
+            }
```
dulajdilshan, line 752: It seems like this `null` check and visiting happens frequently. How about extracting this logic to a new function and call it?
```java
analyze(paramter.typenode)

private void analyze(BLangType node) {
    if (node != null) {
        node.accept(this);
    }
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/split/JvmCreateTypeGen.java
```diff
@@ -247,6 +269,86 @@ private void createTypesInstance(ClassWriter cw, List<BIRTypeDefinition> typeDef
...
 
+    private void createTypedescInstances(ClassWriter cw, List<BIRTypeDefinition> typeDefs, String moduleInitClass,
+                                         List<BIRNonTerminator.NewTypeDesc> typesecs) {
+        MethodVisitor mv = cw.visitMethod(ACC_PUBLIC + ACC_STATIC, CREATE_TYPEDESC_INSTANCES_METHOD,
+                                          "()V", null, null);
+        mv.visitCode();
+        for (BIRTypeDefinition typeDefinition : typeDefs) {
+            BType bType = JvmCodeGenUtil.getReferredType(typeDefinition.type);
+            BSymbol owner = bType.tsymbol.owner;
+           if (bType.tag == TypeTags.RECORD && (owner != null && owner.getKind() != SymbolKind.PACKAGE)) {
```
warunalakshitha, line 280: Why we need to separately handle record?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/split/JvmCreateTypeGen.java
```diff
@@ -247,6 +269,86 @@ private void createTypesInstance(ClassWriter cw, List<BIRTypeDefinition> typeDef
...
+        mv.visitInsn(RETURN);
+        mv.visitMaxs(0, 0);
+        mv.visitEnd();
+    }
+
+    private void createTypedescInstances(MethodVisitor mv, BIRNonTerminator.NewTypeDesc typeDesc) {
+        BType type = JvmCodeGenUtil.getReferredType(typeDesc.type);
+        String className = TYPEDESC_VALUE_IMPL;
+        if (type.tag == TypeTags.RECORD) {
+            className = getTypeDescClassName(JvmCodeGenUtil.getPackageName(type.tsymbol.pkgID), toNameString(type));
```
warunalakshitha, line 298: Can we add a test to cover this?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/codegen/split/JvmCreateTypeGen.java
```diff
@@ -247,6 +269,86 @@ private void createTypesInstance(ClassWriter cw, List<BIRTypeDefinition> typeDef
...
+    }
+
+    private void createTypedescInstance(MethodVisitor mv, BType bType, BIRTypeDefinition typeDefinition,
+                                        String moduleInitClass) {
+        String className;
+        if (bType.tag == TypeTags.RECORD) {
+            PackageID pkgID = bType.tsymbol.pkgID;
+            String packageName = JvmCodeGenUtil.getPackageName(pkgID);
+            className = getTypeDescClassName(packageName, toNameString(bType));
+        } else if (getEffectiveType(bType).tag == TypeTags.TUPLE) {
```
warunalakshitha, line 326: Shall we assign getEffectiveType(bType) to variable and use

warunalakshitha:
Shall we add test cases as well

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/BIRGen.java
```diff
@@ -2232,17 +2415,36 @@ public void visit(BLangTableConstructorExpr tableConstructorExpr) {
...
+        List<BIROperand> varDcls = mapToVarDcls(type);
+        BTypeSymbol typeSymbol = type.tsymbol;
+        BIRVariableDcl tempVarDcl;
+        BIROperand toVarRef;
+        BIRNonTerminator.NewTypeDesc newTypeDesc;
+        if (this.env.enclBB == null) {
+            if (type.tag == TypeTags.RECORD || typeSymbol.name != Names.EMPTY) {
+                return;
+            }
+            tempVarDcl = new BIRGlobalVariableDcl(symTable.typeDesc, this.env.nextTypedescId(names), VarScope.GLOBAL,
```
KavinduZoysa, line 2431: Let's define this a local variable inside if block, then the casting is not needed.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/Types.java
```diff
@@ -5415,6 +5417,19 @@ private BType getIntersectionForErrorTypes(IntersectionContext intersectionConte
         return intersectionErrorType;
     }
 
+    public static boolean isUserDefinedTypeDefinition(BType type) {
+        type = Types.getReferredType(type);
```
KavinduZoysa, line 5421: Let's create a new variable here.

chiranSachintha:
In the current implementation, we create the type descriptor for the type definition prior to evaluating the initialization function. However, at that time we do not have the required annotations for the type descriptor. To resolve this issue, we need to move the generation of the type descriptor logic to the init function. This will be addressed after merging the tuple annotation PR.

chiranSachintha:
In this pull request (PR), the following changes have been made:
1. The typedesc instruction is now created when visiting the type node of the tuple and record type nodes.
2. The typedesc instruction has been moved to the "init" function. Previously, the typedesc instruction was generated before the "init" function.
3. To achieve this, the top-level nodes in the package have been reordered. Previously, only the global variables were reordered, but now the global variables, type definitions, and constants are reordered together. This change has been implemented in this PR.
4. The annotation details have been added to the "typedesc" instruction.

some tests are failing with the change of reordering of global variables and adding the annotation details to the typedesc of the tuple type node. I am currently working on resolving this issue.

chiranSachintha:
I have opened a new PR for this change as we have modified the logic(https://github.com/ballerina-platform/ballerina-lang/pull/41945)

Issue #34378, opened 2021-12-17
Changing org name in Ballerina.toml does not show any diagnostics


malinthar:
**Description:**
Consider the following Ballerina.toml config and the default module where I have added an import statement for importing the sub-module `module1`.

```toml
[package]
distribution = "slbeta6"
name="project2"
org="malintha"
version="1.0.0"
export=["project2","project2.module1"]


[build-options]
observabilityIncluded = true
```

```ballerina
import malintha/project2.module1;
import ballerina/io;


public function main() {
    io:print(module1:hello("Ballerina"));
}
```

Now, When I change the org name in ballerina toml as follows I do not get any diagnostics for the import statement. 

![baltoml](https://user-images.githubusercontent.com/35211477/146489242-5c5ad4fa-34c4-4984-aa00-d358e706bc57.png)

![main](https://user-images.githubusercontent.com/35211477/146489277-1a88af4d-f8a4-4551-a6a3-7c252ca623e1.png)

**Steps to reproduce:**
This is reproducible only when there is the `exports` array in the `package` table. 

**Affected Versions:**
slbeta6, slbeta7-snapshot

pramodya1994:
Here you can use the following import declaration without the org name.
```
import project2.module1;
```
Since we are using the same package modules org name is optional.

Dilhasha:
We can use the `clearCaches` API after updating the Ballerina.toml this will fix the diagnostics issue mentioned. I've opened a PR  https://github.com/ballerina-platform/ballerina-lang/pull/39138 
Appreciate your feedback @malinthar @IMS94

malinthar:
> We can use the `clearCaches` API after updating the Ballerina.toml this will fix the diagnostics issue mentioned. I've opened a PR #39138 Appreciate your feedback @malinthar @IMS94

Added https://github.com/ballerina-platform/ballerina-lang/pull/39138#pullrequestreview-1225617287

PR #39138, opened 2022-12-21
Clear caches after the Ballerina toml update to get diagnostics

Dilhasha:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/34378

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/io/ballerina/projects/Package.java
+++ b/compiler/ballerina-lang/src/main/java/io/ballerina/projects/Package.java
@@ -697,7 +697,7 @@ public class Package {
                         oldModuleContext.isDefaultModule(), srcDocContextMap, testDocContextMap,
                         oldModuleContext.moduleMdContext().orElse(null),
                         oldModuleContext.moduleDescDependencies(), resourceMap, testResourceMap));
-                // Clear compilation package cache
+                // Remove the module with old PackageID from the compilation cache
                 PackageCache.getInstance(project.projectEnvironmentContext().getService(CompilerContext.class)).
                         remove(oldModuleContext.descriptor().moduleCompilationId());
             }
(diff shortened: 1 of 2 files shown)


malinthar (commented):
IMO, Clearing the cache every time we update the Ballerina.toml could slow things down. Shall we try to identify if the package name or org name has changed, and clear the caches only in those cases?

IMS94:
> IMO, Clearing the cache every time we update the Ballerina.toml could slow things down. Shall we try to identify if the package name or org name has changed, and clear the caches only in those cases?

+1

gayaldassanayake (approved):

Dilhasha:
When there's a toml update, we are updating all the modules of the package and create a new package. This replaces the package cache with the newly created package. When there is a  `package` or `org` change, this does not replace the existing package in the cache due to the name change. Instead, it creates a new package. So, since the package with the old name is available in the cache, it doesn't produce a diagnostic. 

With a8730a406e961bb981d11f9eb86fc52e8b790446 , we clear the old package from cache before creating the new package.

malinthar (dismissed):
LGTM!

compiler/ballerina-lang/src/main/java/io/ballerina/projects/Package.java
```diff
@@ -697,6 +697,9 @@ private void updateModules() {
                         oldModuleContext.isDefaultModule(), srcDocContextMap, testDocContextMap,
                         oldModuleContext.moduleMdContext().orElse(null),
                         oldModuleContext.moduleDescDependencies(), resourceMap, testResourceMap));
+                // Clear compilation package cache
```
azinneera, line 700:
```suggestion
                // Remove the module with old PackageID from the compilation cache
```

azinneera (approved):

compiler/ballerina-lang/src/main/java/io/ballerina/projects/Package.java
Dilhasha: Fixed with f4ad8c6fe9bdff0b316bd3cef1b17b8804dd0a05

Issue #39169, opened 2023-01-04
[Bug]: Get Invalid mutation error in table


KavinduZoysa:
### Description

https://stackoverflow.com/questions/74913033/error-table-was-mutated-after-iterator-created-although-iterator-and-mutation-a

### Steps to Reproduce

Check above link

### Affected Version(s)

2201.3.0

### Related area

-> Compilation

PR #39170, opened 2023-01-04
Fix issue in table iteration

KavinduZoysa:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/39169

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TableUtils.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TableUtils.java
@@ -96,6 +96,8 @@ public class TableUtils {
             } else {
                 return (long) obj.hashCode();
             }
+        } else if (obj instanceof Long) {
+            return (long) obj;
         } else {
             return (long) obj.hashCode();
         }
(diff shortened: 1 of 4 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/TableUtils.java
```diff
@@ -96,6 +96,8 @@ public static Long hash(Object obj, Node parent) {
             } else {
                 return (long) obj.hashCode();
             }
+        } else if (obj instanceof Long) {
```
gimantha, line 99: I suspect that the same thing can happen for the floats also. Shall we add test cases for all basic data types? internally it seems the same algo is used for java double values to calculate the hashcode

hasithaa (approved):

hasithaa:
Do we have test cases covering each type hashes?

hasithaa:
@KavinduZoysa Please send a PR to Update 3.2 branch.

KavinduZoysa:
> #39169

This is the new PR https://github.com/ballerina-platform/ballerina-lang/pull/39184

PR #39186, opened 2023-01-05
[2201.3.x] Fix annotation passing for immutable inline records

### LINKED ISSUE #39156: [Bug]: Annotations are not passed in the type if the type is readonly
### Description

The field level annotations are not passed with the type when the type is readonly. Due to this `constraint:validate()` method is not able to validate the constraints. 

Related to : https://github.com/ballerina-platform/ballerina-standard-library/issues/3742

### Steps to Reproduce

See the following code :

```ballerina
import ballerina/io;
import ballerina/constraint;

type Album readonly & record {|
    @constraint:String {
        maxLength: 5,
        minLength: 1
    }
    string title;
    string artist;
|};

    Album a = {
        artist: "test",
        title: "testss"
    };
    Album validAlbum = check constraint:validate(a);
    io:println(validAlbum);
}
```

The `<AnnotatableType>.getAnnotations()` is `null` when the record is `readonly`. If we remove `readonly`, we can get the annotations.

### Affected Version(s)

Ballerina SwanLake 2201.2.x
Ballerina SwanLake 2201.3.x

_No response_

### Related area

-> Runtime

_No response_

_No response_

_No response_

HindujaB:
## Purpose
$subject
Fixes #39156

## Approach
As for inline records, the global annotation map is passed with `anonaType` key, it is not populated to the runtime record type.  So, we don't pass the `anonType` internal name to the runtime for retrieving the annotations which causes the issue. 
This PR passes the internal name to runtime to be used when populating annotations. 

## Samples
```ballerina
import ballerina/io;
import ballerina/constraint;

type Album readonly & record {|
    @constraint:String {
        maxLength: 5,
        minLength: 1
    }
    string title;
    string artist;
|};

    Album a = {
        artist: "test",
        title: "testss"
    };
    Album validAlbum = check constraint:validate(a); // should fail
    io:println(validAlbum);
}
```
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/creators/TypeCreator.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/api/creators/TypeCreator.java
@@ -224,7 +224,7 @@ public class TypeCreator {
      */
     public static RecordType createRecordType(String typeName, Module module, long flags, boolean sealed,
                                               int typeFlags) {
-        return new BRecordType(typeName, module, flags, sealed, typeFlags);
+        return new BRecordType(typeName, typeName, module, flags, sealed, typeFlags);
     }
 
     /**
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
@@ -44,7 +44,7 @@ import java.util.Optional;
  * @since 0.995.0
  */
 public class BRecordType extends BStructureType implements RecordType {
-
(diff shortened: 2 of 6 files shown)


Nadeeshan96 (approved):
LGTM

HindujaB:
As per the offline discussion, if this fix is added to `2201.3.x` it might break the existing `constraint` package usage with `2201.3.x` older versions. Therefore, adding the fix in the master branch will be enough. Hence, closing the PR.

PR #39187, opened 2023-01-05
[master] Fix annotation passing for immutable inline records

### LINKED ISSUE #39156: [Bug]: Annotations are not passed in the type if the type is readonly

HindujaB:
## Purpose
$subject
Fixes #39156

## Approach
As for inline records, the global annotation map is passed with `anonaType` key, it is not populated to the runtime record type.  So, we don't pass the `anonType` internal name to the runtime for retrieving the annotations which causes the issue. 
This PR passes the internal name to runtime to be used when populating annotations. 

## Samples
```ballerina
import ballerina/io;
import ballerina/constraint;

type Album readonly & record {|
    @constraint:String {
        maxLength: 5,
        minLength: 1
    }
    string title;
    string artist;
|};

    Album a = {
        artist: "test",
        title: "testss"
    };
    Album validAlbum = check constraint:validate(a); // should fail
    io:println(validAlbum);
}
```
## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
@@ -44,7 +44,7 @@ import java.util.Optional;
  * @since 0.995.0
  */
 public class BRecordType extends BStructureType implements RecordType {
-    private String internalName = null;
+    private final String internalName;
     public boolean sealed;
     public Type restFieldType;
     public int typeFlags;
@@ -95,6 +95,7 @@ public class BRecordType extends BStructureType implements RecordType {
             this.restFieldType = restFieldType;
             this.fields = fields;
         }
+        this.internalName = typeName;
     }
 
     private Map<String, Field> getReadOnlyFields(Map<String, Field> fields) {
@@ -142,10 +143,7 @@ public class BRecordType extends BStructureType implements RecordType {
 
     @Override
     public String getAnnotationKey() {
-        if (this.internalName != null) {
-            return Utils.decodeIdentifier(this.internalName);
-        }
-        return Utils.decodeIdentifier(this.typeName);
+        return Utils.decodeIdentifier(this.internalName);
     }
 
     @Override
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
@@ -44,7 +44,7 @@ import java.util.Optional;
  * @since 0.995.0
  */
 public class BRecordType extends BStructureType implements RecordType {
-
+    private String internalName = null;
     public boolean sealed;
     public Type restFieldType;
     public int typeFlags;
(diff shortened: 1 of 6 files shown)


tests/jballerina-unit-test/src/test/java/org/ballerinalang/nativeimpl/jvm/runtime/api/tests/Values.java
```diff
@@ -443,4 +444,21 @@ public static BArray getTupleWithInitialValues(BTypedesc typedesc, BArray array)
...
 
+    public static Boolean checkInlineRecordAnnotations(BTypedesc typedesc, BTypedesc constraint) {
+        Type describingType = typedesc.getDescribingType();
+        if (describingType.getTag() == TypeTags.TYPE_REFERENCED_TYPE_TAG &&
+                ((ReferenceType) describingType).getReferredType().getTag() == TypeTags.INTERSECTION_TAG) {
+            describingType = ((IntersectionType)
+                    ((ReferenceType) describingType).getReferredType()).getConstituentTypes().get(0);
+        }
+        if (!(describingType instanceof AnnotatableType)) {
+            throw ErrorCreator.createError(StringUtils.fromString("Invalid type found."));
```
warunalakshitha, line 455: Should we print type name here in the error?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/types/BRecordType.java
```diff
@@ -61,8 +61,9 @@ public class BRecordType extends BStructureType implements RecordType {
      * @param sealed flag indicating the sealed status
      * @param typeFlags flags associated with the type
      */
-    public BRecordType(String typeName, Module pkg, long flags, boolean sealed, int typeFlags) {
+    public BRecordType(String typeName, String internalName, Module pkg, long flags, boolean sealed, int typeFlags) {
```
warunalakshitha, line 64: should we remove other constructor where internal name can be null? or we can set internal name = typeName and add only declaration in the class private String internalName ;

HindujaB: Updated the constructor to use the `typeName`

warunalakshitha (approved):
LGTM

PR #39199, opened 2023-01-09
Fix `fromBytes` and `equalsIgnoreCaseAscii` langlib methods to support async calls

### LINKED ISSUE #39157: [Bug]: `fromBytes()` and `equalsIgnoreCaseAscii()` not safe for concurrent use
### Description

$subject due to how `CharsetDecoder` has been used in the native implementations of the above functions. In the API docs, it mentions that this class is not safe for concurrent access (https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/charset/CharsetDecoder.html#decode(java.nio.ByteBuffer)). But in the implementations a single instance of the decoder is used. This leads to weird intermittent issues when these functions are used in concurrent access contexts (e.g., resources of a service).

These aren't handling the `IllegalStateException` which can be thrown by the `decode()` method. This seems to lead to panics with the exceptions string representation as the error message as well. 

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

-> Runtime

_No response_

_No response_

_No response_

HindujaB:
## Purpose
$subject

Fixes #39157 
Fixes https://github.com/wso2-enterprise/internal-support-ballerina/issues/237

## Approach
The issue happens due to the static `CharSetDecoder` instances used in these methods are not thread-safe. Therefore `IllegalStateException` is thrown when these methods are called asynchronously.
The PR fixes it by creating decoder instances only when the method is called.  

## Samples
```ballerina

public function main() {
    foreach int i in 0 ... 4 {
        callFromBytesAsync();
    }
}

function callFromBytesAsync() {
    string append = "";
    worker w1 {
        future<string|error> bytesFuture = start callFromBytes1();
        runtime:sleep(2);
        string|error str = wait bytesFuture;
        assertTrue(str is string);
        if (str is string) {
            append += str + " ";
        }
    }
    worker w2 {
        future<string|error> bytesFuture = start callFromBytes2();
        string|error str = wait bytesFuture;
        assertTrue(str is string);
        if (str is string) {
            append += str + " ";
        }
    }
    _ = wait {w1, w2};
}

function callFromBytes1() returns string|error {
    byte[] bytes = [
        72,
        101,
        108,
        108,
        111,
        33,
        126,
        63,
        194,
        163,
        195,
        159,
        208,
        175,
        206,
        187,
        226,
        152,
        131,
        226,
        156,
        136,
        224,
        175,
        184,
        240,
        159,
        152,
        128,
        240,
        159,
        132,
        176,
        240,
        159,
        141,
        186
    ];
    return check string:fromBytes(bytes);
}

function callFromBytes2() returns string|error {
    byte[] bytes = [
        72,
        101,
        108,
        108,
        111,
        32,
        66,
        97,
        108,
        108,
        101,
        114,
        105,
        110,
        97,
        33
    ];
    return checkpanic string:fromBytes(bytes);
}

```

## CheckList 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/EqualsIgnoreCaseAscii.java
+++ b/langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/EqualsIgnoreCaseAscii.java
@@ -22,8 +22,8 @@ import io.ballerina.runtime.api.values.BString;
 
 import java.nio.ByteBuffer;
 import java.nio.charset.CharacterCodingException;
-import java.nio.charset.Charset;
 import java.nio.charset.CharsetDecoder;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Extern function lang.string:equalsIgnoreCase(string, string).
@@ -31,12 +31,9 @@ import java.nio.charset.CharsetDecoder;
  * @since 1.2
  */
 public class EqualsIgnoreCaseAscii {
-    private static CharsetDecoder decoder;
 
-    static {
-        decoder = Charset.forName("US-ASCII").newDecoder();
+    private EqualsIgnoreCaseAscii() {
     }
-
     public static boolean equalsIgnoreCaseAscii(BString s1, BString s2) {
         if (s1.length() != s2.length()) {
             return false;
@@ -59,13 +56,14 @@ public class EqualsIgnoreCaseAscii {
     }
 
     private static boolean isPureAscii(String  str) {
-        byte byteArray[] = str.getBytes();
+        byte[] byteArray = str.getBytes();
+        CharsetDecoder decoder = StandardCharsets.US_ASCII.newDecoder();
         try {
-            decoder.decode(ByteBuffer.wrap(byteArray)).toString();
+            decoder.decode(ByteBuffer.wrap(byteArray));
+            decoder.reset();
         } catch (CharacterCodingException e) {
             return false;
         }
-
         return true;
     }
 }
--- a/langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/EqualsIgnoreCaseAscii.java
+++ b/langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/EqualsIgnoreCaseAscii.java
@@ -20,9 +20,6 @@ package org.ballerinalang.langlib.string;
 
 import io.ballerina.runtime.api.values.BString;
 
-import java.nio.ByteBuffer;
-import java.nio.charset.CharacterCodingException;
-import java.nio.charset.CharsetDecoder;
 import java.nio.charset.StandardCharsets;
 
 /**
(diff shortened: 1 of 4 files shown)


langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/EqualsIgnoreCaseAscii.java
```diff
@@ -22,21 +22,18 @@
...
  *
  * @since 1.2
  */
 public class EqualsIgnoreCaseAscii {
-    private static CharsetDecoder decoder;
 
-    static {
-        decoder = Charset.forName("US-ASCII").newDecoder();
+    private EqualsIgnoreCaseAscii() {
     }
```
Nadeeshan96, line 36:
```suggestion
    }
    
```

langlib/lang.string/src/main/java/org/ballerinalang/langlib/string/EqualsIgnoreCaseAscii.java
```diff
@@ -59,13 +56,14 @@ public static boolean equalsIgnoreCaseAscii(BString s1, BString s2) {
...
 
     private static boolean isPureAscii(String  str) {
```
Nadeeshan96, line 58: I think we can replace this method's entire body with the following line
`return StandardCharsets.US_ASCII.newEncoder().canEncode(str);` Shall we please check this?

Nadeeshan96, line 58: Shall we check whether we can replace this method's entire method by the following line?
`return StandardCharsets.US_ASCII.newEncoder().canEncode(str);`

langlib/langlib-test/src/test/resources/test-src/stringlib_test.bal
```diff
@@ -687,6 +688,138 @@ function testIncludesMatch() {
     assertTrue(result6);
 }
 
+function testFromBytesAsync() {
+    foreach int i in 0 ... 4 {
```
Nadeeshan96, line 692: Shall we increase the iterations to 10 and reduce the sleep to value like 0.002?

langlib/langlib-test/src/test/resources/test-src/stringlib_test.bal
```diff
@@ -687,6 +688,138 @@ function testIncludesMatch() {
...
+        callFromBytesAsync();
+    }
+}
+
+function callFromBytesAsync() {
```
Nadeeshan96, line 697: Shall we make these functions containing workers `isolated`?

warunalakshitha (commented):
LGTM

Nadeeshan96 (approved):
LGTM

PR #39206, opened 2023-01-10
Fix array langlib doc deviations

MaryamZi:
## Purpose
$title.


langlib/lang.array/src/main/ballerina/array.bal
```diff
@@ -544,6 +550,9 @@ public isolated function fromBase16(string str) returns byte[]|error = @java:Met
 #
 # + arr - The array from which the stream is created
 # + return - The stream representation of the array `arr`
+# The returned stream will use an iterator over `arr` and
+# will therefore handle mutation of `arr` in the same way
+# as an iterator does.
```
dulajdilshan, line 555: Is this intentional to add this description after the params?

MaryamZi: This is for consistency with the spec for now.

dulajdilshan (approved):

PR #39207, opened 2023-01-10
Add a test option to engage native-image agent

### LINKED ISSUE #39198: [New Feature]: Support for engaging GraalVM tracing-agent with `bal test`
### Description

Add support for engaging GraalVM tracing agent with `bal test` command like this:
```
bal test --engage-native-image-agent
```
This can be used to generate the native-image configurations.

### Describe your problem(s)

Currently, Ballerina Native Build may not work when we have external jar dependencies (for example, incase of adding database drivers). This can be potentially resolved by finding the additional configurations by engaging the [tracing-agent](https://www.graalvm.org/latest/reference-manual/native-image/metadata/AutomaticMetadataCollection/). In order to get all the required configurations, we have to test all the possible scenarios. Since we have `tests` for the same reason, it would be great to have a feature which will run the ballerina tests with the tracing-agent to generate this configs.

This would be also helpful to debug and fix native-image test failures in standard libraries.

### Describe your solution(s)

In order to engage tracing agent we have to add the following argument to the java test command :
```
-agentlib:native-image-agent=config-output-dir=target/native-configs
```
This will generate the configuration files to this directory: `target/native-configs`

> **Note :**
> - Since we are not packing the GraalVM JDK with Ballerina, we have to use the local GraalVM JDK to run the tests with the agent. Need to check the `GRAALVM_HOME` environment variable.
> - Also we may have to remove the configurations related to testerina since those configs will report unwanted warnings during the native image build

### Related area

-> Test Framework

_No response_

_No response_

_No response_

TharmiganK:
Fixes #39198 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
@@ -272,7 +272,7 @@ public class RunTestsTask implements Task {
                 String javaCmd = graalvmHome + File.separator + BIN_DIR_NAME + File.separator
                         + (OS.contains("win") ? "java.exe" : "java");
                 cmdArgs.add(javaCmd);
-                cmdArgs.add("-agentlib:native-image-agent=config-output-dir=target/native-image-configs");
+                cmdArgs.add("-agentlib:native-image-agent=config-output-dir=" + target.getTracedNativeConfigsPath());
             }
         }
         if (!nativeImageAgentEnabled) {
--- a/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
+++ b/cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
@@ -53,6 +53,7 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.StringJoiner;
 import java.util.stream.Collectors;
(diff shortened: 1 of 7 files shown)


TharmiganK:
The above fix will also generate the configs related to testerina which can produce unwanted warnings if the configs are used without filtering. IMO this is fine as I don't see a way to remove these testerina related configs and GraalVM also suggests to do filtering after obtaining the configs.

cli/ballerina-cli/src/main/java/io/ballerina/cli/task/RunTestsTask.java
```diff
@@ -258,7 +264,23 @@ private int runTestSuite(Target target, Package currentPackage, JBallerinaBacken
...
-        cmdArgs.add(System.getProperty("java.command"));
+        boolean nativeImageAgentEnabled = false;
+        if (enableNativeImageAgent) {
+            String graalvmHome = System.getenv("GRAALVM_HOME");
+            if (graalvmHome != null) {
+                nativeImageAgentEnabled = true;
+                String javaCmd = graalvmHome + File.separator + BIN_DIR_NAME + File.separator
+                        + (OS.contains("win") ? "java.exe" : "java");
+                cmdArgs.add(javaCmd);
+                cmdArgs.add("-agentlib:native-image-agent=config-output-dir=target/native-image-configs");
```
azinneera, line 275: Users can pass a custom directory for the target. So we should always access the target from Project.targetDir()

TharmiganK: Fixed via [b274375](https://github.com/ballerina-platform/ballerina-lang/pull/39207/commits/b27437522da4d101d4a4e2c31b711fec2a49d5a1)

azinneera:
Since the option is exclusive to the graalvm native image generation, we can live with the workaround of running the tracing agent manually using graalvm command.

PR #39229, opened 2023-01-11
[2201.3.x] Recompile crashed source only when the source is updated

### LINKED ISSUE #39014: [Bug]: LS does not recover from bad sad error
### Description

Consider https://github.com/ballerina-platform/ballerina-lang/issues/39012.

In the above scenario, although the code snippet that generates the bad, sad error is removed, the subsequent compilation requests keep showing the bad sad error. 

### Steps to Reproduce

_No response_

### Affected Version(s)

Update 3

_No response_

### Related area

-> Editor

_No response_

_No response_

_No response_

pcnfernando:
Fixes #39014 

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/diagnostic/DiagnosticsHelper.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/diagnostic/DiagnosticsHelper.java
@@ -26,6 +26,7 @@ import org.ballerinalang.langserver.commons.LanguageServerContext;
 import org.ballerinalang.langserver.commons.WorkspaceServiceContext;
 import org.ballerinalang.langserver.commons.client.ExtendedLanguageClient;
 import org.ballerinalang.langserver.commons.workspace.WorkspaceManager;
+import org.ballerinalang.langserver.workspace.BallerinaWorkspaceManager;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.DiagnosticSeverity;
 import org.eclipse.lsp4j.Position;
@@ -178,7 +179,7 @@ public class DiagnosticsHelper {
     }
 
     public Map<String, List<Diagnostic>> getLatestDiagnostics(DocumentServiceContext context) {
-        WorkspaceManager workspace = context.workspace();
+        BallerinaWorkspaceManager workspace = (BallerinaWorkspaceManager) context.workspace();
         Map<String, List<Diagnostic>> diagnosticMap = new HashMap<>();
 
         Optional<Project> project = workspace.project(context.filePath());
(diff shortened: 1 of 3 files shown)


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/diagnostic/DiagnosticsHelper.java
```diff
@@ -187,13 +189,13 @@ public Map<String, List<Diagnostic>> getLatestDiagnostics(DocumentServiceContext
...
         // IDE requires the original path.
         Path projectRoot = workspace.projectRoot(context.filePath());
-        if (project.get().kind() == ProjectKind.SINGLE_FILE_PROJECT) {
-            projectRoot = projectRoot.getParent();
-        }
-        PackageCompilation compilation = workspace.waitAndGetPackageCompilation(context.filePath()).orElseThrow();
+        Path originalPath = project.get().kind() == ProjectKind.SINGLE_FILE_PROJECT
+                ? projectRoot.getParent() : projectRoot;
+        Optional<PackageCompilation> compilationResult = workspace.waitAndGetPackageCompilation(context.filePath(),
+                context.operation() == LSContextOperation.TXT_DID_CHANGE);
```
IMS94, line 195: Shall we add `DID_OPEN` as well to here?

IMS94 (changes requested):

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/diagnostic/DiagnosticsHelper.java
pcnfernando: I don't get the requirement though. I think it's not needed since we check this boolean value, iff the previous compilation has crashed (https://github.com/ballerina-platform/ballerina-lang/pull/39229/files#diff-6d4a15825d022264795ed5bed5a6474047b13e15b8329dbc9fce4d5cfa5ca5ebR305).

Assuming DID_OPEN happens for the below scenarios
1. Opening it on a new tab
2. Move back to the tab

The project can be single or multi-filed.
If it's a single file; the previous compilation crashed means that the file has already been opened. Moving back to the tab does not actually need a compilation because nothing changed.
If it's a multi-filed project, let's say the project crashed due to an issue in a separate file and we are opening another file of the project; even in that case, there's no advantage of recompiling. 
If it's a multi-filed project and we are opening the culprit file itself, even then, without a source change being done, there is no need for recompilation.

wdyt?

IMS94: Yes, agreed. There can be an edge case with generated sources though. We can neglect that at this stage.

IMS94 (approved):

PR #39234, opened 2023-01-11
Support annotations for tuple members

ushirask:


compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/tree/STMemberTypeDescriptorNode.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * This is a generated internal syntax tree node.
+ *
+ * @since 2201.4.0
+ */
+public class STMemberTypeDescriptorNode extends STNode {
```
MaryamZi, line 33: Shouldn't this be STMemberNode? Similar concern for things like `SyntaxKind.MEMBER_TYPE_DESC`.

cc @rdulmina

MaryamZi: Another concern is `io.ballerina.compiler.syntax.tree.TupleTypeDescriptorNode#memberTypeDesc`. This doesn't quite return member typedescs now, right?

MaryamZi:
The following sample results in the compiler crashing.

```ballerina
annotation record {| int val; |} annot on field;

public function main() {
    [string, @annot {val: undefined} string] t;
}
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/tree/types/BLangTupleTypeNode.java
```diff
@@ -37,14 +38,21 @@
 public class BLangTupleTypeNode extends BLangType implements TupleTypeNode {
 
     // BLangNodes
-    public List<BLangType> memberTypeNodes = new ArrayList<>();
+    public List<BLangSimpleVariable> memberTypeNodes = new ArrayList<>();
```
MaryamZi, line 41:
```suggestion
    public List<BLangSimpleVariable> members = new ArrayList<>();
```

Please check all fields and update appropriately.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/tree/STMemberTypeDescriptorNode.java
rdulmina: @MaryamZi r u concerned about the name? If so `STMemberNode` is too high-level right we need to be more specific. right now `io.ballerina.compiler.syntax.tree.TupleTypeDescriptorNode#memberTypeDesc` returns `Node` since we have the rest type descriptor node to that we couldn't use `TypeDescriptorNode` as the type

MaryamZi: My concern is that it no longer represents just typedescs. The comparable construct for records is `STRecordFieldNode`, so IMO, would be better to have something like `STTupleMemberNode`.

I rechecked the spec though, and the spec uses member-type-descriptor, so this may be OK (https://ballerina.io/spec/lang/master/#member-type-descriptor). Just that it may be a bit confusing. WDYT?

rdulmina: We usually follow the names in the spec. But I agree with u `STTupleMemberNode` is better. But it is better to align with the spec names.

rdulmina: May be we can create an issue to update the spec name :smiley: WDYT?

.github/workflows/pull_request_ubuntu_build.yml
```diff
@@ -14,6 +14,7 @@ on:
       - 2201.[0-9]+.[0-9]+-stage
       - native-build
       - revert-client-decl-master
+      - annot-tuple-members
```
MaryamZi, line 17: Let's remove this in a separate PR.

MaryamZi (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -1430,23 +1431,34 @@ public BType transform(BLangFiniteTypeNode finiteTypeNode, AnalyzerData data) {
...
+            }
+            if (member.getBType() == symTable.noType) {
+                hasUndefinedMember = true;
+            }
+            members.add(member);
+        }
+        // If at least one member is undefined, return noType as the type.
+        if (hasUndefinedMember) {
+            return symTable.noType;
+        }
```
chiranSachintha, line 1456:
```ballerina
type Foo [int, int, Bar];
type Bar int;
```
With this approach, if we consider the above example, we first resolve the first two members and identify that the "Bar" type is not resolved. Then, we return it and identify that the "Foo" type is not resolved. However, the actual not resolved type is "Bar." To correct this issue for the "function type," we resolve the member types after resolving all the type definitions.

https://github.com/ballerina-platform/ballerina-lang/blob/f715e82171cfac4a4d8010e27ff29177a66d532f/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java#L440

MaryamZi: `Bar` not being resolved means `Foo` isn't resolved either. IMO, that is OK. IIRC, that's what's done for records too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -1430,23 +1431,34 @@ public BType transform(BLangFiniteTypeNode finiteTypeNode, AnalyzerData data) {
...
-
-        BTupleType tupleType = new BTupleType(tupleTypeSymbol, memberTypes);
+        SymbolEnv tupleEnv = SymbolEnv.createTypeEnv(tupleTypeNode, new Scope(tupleTypeSymbol), data.env);
+        boolean hasUndefinedMember = false;
+        for (BLangSimpleVariable member : tupleTypeNode.members) {
+            BType bType = member.getBType();
+            if (bType == null) {
+                symbolEnter.defineNode(member, tupleEnv);
+            } else if (bType == symTable.noType) {
+                member.setBType(null);
```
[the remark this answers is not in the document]
ushirask: if `BType` is already set it type won't be resolved at SymbolEnter

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -993,9 +997,21 @@ public void visit(BLangIntersectionTypeNode intersectionTypeNode, AnalyzerData d
 
     @Override
     public void visit(BLangTupleTypeNode tupleTypeNode, AnalyzerData data) {
-        List<BLangType> memberTypeNodes = tupleTypeNode.memberTypeNodes;
-        for (BLangType memType : memberTypeNodes) {
-            analyzeNode(memType, data);
+        List<BLangSimpleVariable> memberNodes = tupleTypeNode.members;
+        BType bType = tupleTypeNode.getBType();
+        data.env = bType.tag == TypeTags.NONE ? data.env :
+                SymbolEnv.createTypeEnv(tupleTypeNode, new Scope(bType.tsymbol), data.env);
```
chiranSachintha, line 1003: Why do we perform this check? Here, How does the type tag become "None"?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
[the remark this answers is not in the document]
MaryamZi: Can you check how it's handled for say records? Setting null doesn't look correct to me either.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
chiranSachintha: Yes, it is fine. However, we need to resolve member types of type definitions after resolving all the type definitions. If not, we will need to perform additional work. As previously mentioned, when we identify that 'Foo' is not defined, we will need to attempt to define member types again. However, during the previous iteration, the types of some members may have already been defined. We need to check that.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
ushirask: if bType is `noType` we won't to avoid updating the environment. Should be changed to `bType == symTable.noType`.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -1430,23 +1431,34 @@ public BType transform(BLangFiniteTypeNode finiteTypeNode, AnalyzerData data) {
...
-            if (type == symTable.noType) {
-                return symTable.noType;
-            }
-
-            memberTypes.add(type);
-        }
-
-        BTypeSymbol tupleTypeSymbol = Symbols.createTypeSymbol(SymTag.TUPLE_TYPE, Flags.asMask(EnumSet.of(Flag.PUBLIC)),
-                Names.EMPTY, data.env.enclPkg.symbol.pkgID, null,
+        List<BLangSimpleVariable> members = new ArrayList<>(tupleTypeNode.members.size());
```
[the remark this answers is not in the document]
ushirask: Can you elaborate?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -993,9 +997,21 @@ public void visit(BLangIntersectionTypeNode intersectionTypeNode, AnalyzerData d
...
+        boolean isTuple = bType.tag == TypeTags.TUPLE;
+        List<BTupleMember> members = isTuple ? ((BTupleType) bType).getMembers() : null;
+
+        for (int i = 0; i < memberNodes.size(); i++) {
+            BLangSimpleVariable member = memberNodes.get(i);
+            analyzeNode(member, data);
+            if (isTuple) {
+                for (BLangAnnotationAttachment ann : member.annAttachments) {
+                    members.get(i).symbol.addAnnotation(ann.annotationAttachmentSymbol);
```
chiranSachintha, line 1012: Why do we create a separate member list? Is it not possible to use the member list in the "BlangTupleTypeNode"?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -993,9 +997,21 @@ public void visit(BLangIntersectionTypeNode intersectionTypeNode, AnalyzerData d
...
     public void visit(BLangTupleTypeNode tupleTypeNode, AnalyzerData data) {
-        List<BLangType> memberTypeNodes = tupleTypeNode.memberTypeNodes;
-        for (BLangType memType : memberTypeNodes) {
-            analyzeNode(memType, data);
+        List<BLangSimpleVariable> memberNodes = tupleTypeNode.members;
+        BType bType = tupleTypeNode.getBType();
+        data.env = bType.tag == TypeTags.NONE ? data.env :
+                SymbolEnv.createTypeEnv(tupleTypeNode, new Scope(bType.tsymbol), data.env);
+        boolean isTuple = bType.tag == TypeTags.TUPLE;
+        List<BTupleMember> members = isTuple ? ((BTupleType) bType).getMembers() : null;
```
chiranSachintha, line 1005: Why do we need to do this?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
MaryamZi: These should ideally be resolved with the symbol enter revamp, since we decided to define the "shell" of the type first and then resolve members.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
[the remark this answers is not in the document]
chiranSachintha: My concern is why do we create a new list? Is it not possible to use the member list in the tuple typenode instead?
In line 1458, can't we refactor the code in the following way?
```
 tupleTypeNode.members.forEach(member -> tupleMembers.add(new BTupleMember(member.getBType(),

```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
chiranSachintha: Can you provide an example for this? In my opinion, when it comes to the semantic analyzer, the type cannot be `notype`

MaryamZi:
Merging with the sonarcloud scan failure since it is due to a known issue and not specific to this PR.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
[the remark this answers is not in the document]
ushirask: I don't see an immediate issue with updating the tupleTypeNode members itself, will check and update.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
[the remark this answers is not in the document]
ushirask: This is related to https://github.com/ballerina-platform/ballerina-lang/pull/39234#discussion_r1086230782 right? Since there is no flow in SymbolResolver for records to return `noType` if a field is undefined, it's fields can be defined later in https://github.com/ballerina-platform/ballerina-lang/blob/f715e82171cfac4a4d8010e27ff29177a66d532f/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolEnter.java#L440 . How can we handle this in tuples?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
ushirask:
```
type T [B, C];

type B int;
```
if C is undefined bType of T would be `noType` here right?

chiranSachintha: Currently, we set the type of T as `noType`. But, the type of T should be `[int, noType]`.

Issue #39059, opened 2022-12-15
[Bindgen Tool] Generated Ballerina binding for `java.lang.String` parameters doesn't allow passing nil values


NipunaRanasinghe:
### Description

Consider the following Java method and the counterpart Ballerina function generated using the bindgen tool

- Java method
  ```java
  public void testFunction(String param) {
  }
  ```

- Generated ballerina function
  ```ballerina
  public function testFunction(string param)  {
  }
  ```

As per the generated function signature, it doesn't allow passing nil values to the function. This could be problematic if the underlying Java function has logic depending on the parameter being null, as the that scenario won't be supported through Ballerina. 

### Related area

-> Bindgen Tool

### Related issue(s) (optional)

Refer to https://discord.com/channels/957996897782616114/1052287921253798009/1052287921253798009 for the related discord discussion.

sameerajayasoma:
It looks like the Bingen tool generates `string?` as the return type if the Java method returns a `java.lang.String`. 

```java
public String testFunction(String testParam) {
   // Do something.
}
```
```ballerina
public function testFunction(string arg0) returns string?{
  // Do something.
}
```

I agree that this is inconsistent.

NipunaRanasinghe:
@sameerajayasoma Thanks for your input. So based on our discord discussion as well, it seems we have the following 2 options to rectify the current behaviour.

**Option 1:**  Map java String to Ballerina `string?` type(which will be consistent with how generate bindings for the return types)

**Option 2:**  Map java String to Ballerina `handle` type and provide the related string conversion utils via `ballerina/java` library (as proposed in the initial design doc https://docs.google.com/document/d/1hlKSd-tQQfr8tJVvHUor7zOIqHbztqu2jqihnK7IUhw/edit?usp=sharing)


IMO option 1 would make more sense, as it will be aligned with existing the return type generation and, it will also eliminate the requirement of having separate util functions to be used for the conversion between `string` and `handle` datatypes. @sameerajayasoma @hevayo Appreciate if you could confirm.

sameerajayasoma:
I prefer option 1

NipunaRanasinghe:
@sameerajayasoma thanks for your input. We'll proceed accordingly

NipunaRanasinghe:
Closing as this will be addressed with https://github.com/ballerina-platform/ballerina-lang/issues/39434

PR #39306, opened 2023-01-19
[Bindgen Tool] Allow passing null values into Ballerina bindings generated on java string objects

NipunaRanasinghe:
## Purpose
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/39059

## Approach
Refer to the table below to see the list of impacted changes, related to Ballerina bindings generation for java `String` objects

| Context   | Existing | New |
| -------- | ----------- | ----------- |
| function parameter types    | string       | string? |
|    | string[]       | string?[] |
| function return types   |  string?       | string? |        
|   | string[]?       | string?[]? |

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [x] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/model/JField.java
+++ b/misc/ballerina-bindgen/src/main/java/org/ballerinalang/bindgen/model/JField.java
@@ -23,6 +23,8 @@ import org.ballerinalang.bindgen.utils.BindgenEnv;
 import java.lang.reflect.Field;
 import java.util.Collections;
 
+import static org.ballerinalang.bindgen.utils.BindgenConstants.BALLERINA_NILLABLE_STRING;
+import static org.ballerinalang.bindgen.utils.BindgenConstants.BALLERINA_NILLABLE_STRING_ARRAY;
 import static org.ballerinalang.bindgen.utils.BindgenConstants.BALLERINA_STRING;
 import static org.ballerinalang.bindgen.utils.BindgenConstants.BALLERINA_STRING_ARRAY;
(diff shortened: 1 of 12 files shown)


sameerajayasoma:
The design level changes looks good to me.

sameerajayasoma:
Should we make the function parameter type to this `string?[]?`

NipunaRanasinghe:
> Should we make the function parameter type to this `string?[]?`

@sameerajayasoma I also had the same concern, but, if we are to support passing `null` values for java array objects we'd have to rethink of all the other array types (including the primitive arrays, i.e. `int[]` -> `int[]?`) as a whole IMO. 

And if thats also a valid requirement probably we can look into that improvement separately. WDYT?

sameerajayasoma:
> > Should we make the function parameter type to this `string?[]?`
> 
> @sameerajayasoma I also had the same concern, but, if we are to support passing `null` values for java array objects we'd have to rethink of all the other array types (including the primitive arrays, i.e. `int[]` -> `int[]?`) as a whole IMO.
> 
> And if that's also a valid requirement probably we can look into that improvement separately. WDYT?

Shall we have a chat about this? I think we need to look at the overall design as you said.

NipunaRanasinghe:
> Shall we have a chat about this? I think we need to look at the overall design as you said.

@sameerajayasoma yeah that'll be better. I'll schedule one in early next week

PR #39367, opened 2023-01-25
Fix `createVarSymbolForTupleMember` in `Symbols`

ushirask:

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/Symbols.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/Symbols.java
@@ -223,7 +223,7 @@ public class Symbols {
     }
 
     public static BVarSymbol createVarSymbolForTupleMember(BType type) {
-        return new BVarSymbol(null, null, null, type, null, null, null);
+        return new BVarSymbol(0, null, null, type, null, null, null);
     }
 
     public static String getAttachedFuncSymbolName(String typeName, String funcName) {
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/Symbols.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/Symbols.java
@@ -223,9 +223,7 @@ public class Symbols {
     }
 
     public static BVarSymbol createVarSymbolForTupleMember(BType type) {
-        BTypeSymbol tsymbol = type.tsymbol;
-        return new BVarSymbol(type.flags, tsymbol.name, tsymbol.pkgID, type, tsymbol.owner,
-                tsymbol.pos, tsymbol.origin);
+        return new BVarSymbol(type.flags, null, null, type, null, null, null);
     }
 
     public static String getAttachedFuncSymbolName(String typeName, String funcName) {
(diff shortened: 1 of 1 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/Symbols.java
```diff
@@ -223,9 +223,7 @@ public static BXMLNSSymbol createXMLNSSymbol(Name name,
     }
 
     public static BVarSymbol createVarSymbolForTupleMember(BType type) {
-        BTypeSymbol tsymbol = type.tsymbol;
-        return new BVarSymbol(type.flags, tsymbol.name, tsymbol.pkgID, type, tsymbol.owner,
-                tsymbol.pos, tsymbol.origin);
+        return new BVarSymbol(type.flags, null, null, type, null, null, null);
```
MaryamZi, line 226: Flags also are technically not those of the type.

Can you point to where the name is set for the tuple members?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/Symbols.java
```diff
@@ -223,9 +223,7 @@ public static BXMLNSSymbol createXMLNSSymbol(Name name,
     }
 
     public static BVarSymbol createVarSymbolForTupleMember(BType type) {
-        BTypeSymbol tsymbol = type.tsymbol;
-        return new BVarSymbol(type.flags, tsymbol.name, tsymbol.pkgID, type, tsymbol.owner,
-                tsymbol.pos, tsymbol.origin);
+        return new BVarSymbol(null, null, null, type, null, null, null);
```
MaryamZi, line 226: Does this compile? :)

```suggestion
        return new BVarSymbol(0, null, null, type, null, null, null);
```

ushirask: My bad, fixed it.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/Symbols.java
MaryamZi: @ushirask

ushirask: It's done with `visit(BLangSimpleVariable varNode)` in SymbolEnter when going through `transform(BLangTupleTypeNode tupleTypeNode, AnalyzerData data)` in SymbolAnalyzer. We might be missing the name otherwise.

MaryamZi: As discussed, please fix in a separate PR.

MaryamZi (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/Symbols.java
MaryamZi: We need to properly set symbol origin also.

PR #39504, opened 2023-02-06
Add `int:range` Langlib function

### LINKED ISSUE #38608: [Improvement]: Use `panic` in langlib
### Description

The `int:range` langlib function introduced in https://github.com/ballerina-platform/ballerina-lang/pull/38279 uses an external function to panic since the built in panic is not working properly due to bootstrap order. 
This need to be fixed an the the panic should be moved to ballerina code.

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

-> Compilation

#38348

_No response_

_No response_

ushirask:
Fixes #38608


langlib/lang.int/src/main/ballerina/int.bal
```diff
@@ -215,3 +217,39 @@ public isolated function fromHexString(string s) returns int|error = @java:Metho
...
+    };
+} {
+    int direction = getStepDirection(step);
+    internal:__IntRange intRange = new (rangeStart, rangeEnd + direction, step);
+    return intRange;
+}
+
+isolated function getStepDirection(int step) returns int {
+    if (step == 0) {
+        panic error("step cannot be 0");
```
hasithaa, line 248: Move this validation to the caller. Having this validation here, causes it to print an additional stack trace when it panics. 

```
error: step cannot be 0
        at ballerina.lang.int.0:getStepDirection(int.bal:248)
           ballerina.lang.int.0:range(int.bal:241)
           test:main(test.bal:21)

``

File: langlib/lang.int/src/main/ballerina/int.bal [hunk 2]
```diff
@@ -215,3 +217,32 @@ public isolated function fromHexString(string s) returns int|error = @java:Metho
...
+public isolated function range(int rangeStart, int rangeEnd, int step) returns object {
+    *object:Iterable;
+    public isolated function iterator() returns object {
+        public isolated function next() returns record {|
+            int value;
+        |}?;
+    };
+} {
+    if (step == 0) {
+        panic error("step cannot be 0");
```
MaryamZi, line 242: This is inconsistent with other langlib errors. 

The message is usually something like `{ballerina/lang.int}InvalidStepError` and the message field in the detail mapping will have this detailed version.

PR #39523, opened 2023-02-07
Fix resolving types in rest args in langlib functions

### LINKED ISSUE #31847: `jVM generation is not supported for type other` with rest arg for non-rest param
**Description:**
$title.

**Steps to reproduce:**
```ballerina
public function main() {    
    map<int> v7 = {a: 100, b: 200};

    [function (int i, int j) returns int, int] x = [function (int i, int j) returns int {return i + j;}, 1];
    var y = 'map:reduce(v7, ...x);
    var z = 'map:reduce(v7, ...[function (int i, int j) returns int {return i + j;}, 1]);
}
```

**Affected Versions:**
slbeta2

lasinicl:
## Purpose
$title

Fixes #31847

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -7109,14 +7109,9 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
             restType = data.resultType;
         } else if (vararg != null) {
             iExpr.restArgs.add(vararg);
+            checkTypeParamExpr(vararg, listTypeRestArg, iExpr.langLibInvocation, data);
             if (mappingTypeRestArg != null) {
-                LinkedHashSet<BType> restTypes = new LinkedHashSet<>();
-                restTypes.add(listTypeRestArg);
-                restTypes.add(mappingTypeRestArg);
-                BType actualType = BUnionType.create(null, restTypes);
-                checkTypeParamExpr(vararg, actualType, iExpr.langLibInvocation, data);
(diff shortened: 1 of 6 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -7684,14 +7684,9 @@ private BType checkInvocationArgs(BLangInvocation iExpr, List<BType> paramTypes,
...
+            checkTypeParamExpr(vararg, listTypeRestArg, iExpr.langLibInvocation, data);
             if (mappingTypeRestArg != null) {
-                LinkedHashSet<BType> restTypes = new LinkedHashSet<>();
-                restTypes.add(listTypeRestArg);
-                restTypes.add(mappingTypeRestArg);
-                BType actualType = BUnionType.create(null, restTypes);
-                checkTypeParamExpr(vararg, actualType, iExpr.langLibInvocation, data);
-            } else {
-                checkTypeParamExpr(vararg, listTypeRestArg, iExpr.langLibInvocation, data);
+                checkTypeParamExpr(vararg, mappingTypeRestArg, iExpr.langLibInvocation, data);
```
KavinduZoysa, line 7689: As I understand, this change is to avoid creating the union type because, in the type param, the union case is not handled when the actual type is union. I think it is better to fix the issue in type param phase (the expected type is union and actual type is not union). Let's discuss with others too.

KavinduZoysa (approved):

Issue #39447, opened 2023-01-31
[Bug]: Invalid Error in regex


SasinduDilshara:
### Description

Following code

```
public function main() {
  string:RegExp x7 = re `\\-`;
}
```
Gives

```
Compiling source
        test.bal
ERROR [test.bal:(4:27,4:27)] missing syntax char after backslash
```

But according to the spec this should work fine.

`//-` should work in following way according to the spec

Regexp -> ReTerm ReTerm
             -> ReEscape ReAtom
             -> ReQuoteEscape ReLiteralChar
             -> / ReSyntaxChar ` -`
             -> `//-`

### Related area

-> Compilation

mohanvive:
@SasinduDilshara I have checked this issues, isn't above should fail according to the spec? Isn't it resolve something like the below as per the spec,

```
-> ReEscape ReAtom
-> ReQuoteEscape ReLiteralChar
-> \ ReSyntaxChar  -
-> \\^-
```

mohanvive:
> @SasinduDilshara I have checked this issue, isn't above should fail according to the spec? Isn't it resolve something like the below as per the spec,
> 
> ```
> -> ReEscape ReAtom
> -> ReQuoteEscape ReLiteralChar
> -> \ ReSyntaxChar ReLiteralChar
> -> \\^-
> ```

My interpretation is incorrect. Hence, I'll check this further

PR #39577, opened 2023-02-13
Fix regex handling for cases with slash symbol

mohanvive:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/39447

### DIFF
--- a/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/RegExpParser.java
+++ b/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/RegExpParser.java
@@ -190,6 +190,13 @@ public class RegExpParser extends AbstractParser {
             case PIPE_TOKEN:
                 return parseReQuoteEscape(backSlash);
             default:
+                if (nextToken.kind == SyntaxKind.RE_LITERAL_CHAR &&
+                        nextToken.text().equals(Character.toString(LexerTerminals.MINUS))) {
+                    STNode minusToken = consume();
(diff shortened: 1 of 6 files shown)


compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/RegExpParser.java
```diff
@@ -190,6 +190,13 @@ private STNode parseEscapeChar(STNode backSlash) {
             case PIPE_TOKEN:
                 return parseReQuoteEscape(backSlash);
             default:
+                if (nextToken.kind == SyntaxKind.RE_LITERAL_CHAR &&
```
lasinicl, line 193: I think this case should be handled in `parseCharSetAtom()` method just before handling the `ReEscapes` in line no 431. If we handle this here it will affect the `ReEscapes`.
The following should result in a syntax error as `-` is not a `ReSyntaxChar`. but with this change it won't.
```ballerina
var t = re `\-`;
```

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/RegExpParser.java
```diff
@@ -421,6 +421,14 @@ private STNode parseCharSetAtom(STToken nextToken, STNode prevNode) {
             case ESCAPED_MINUS_TOKEN:
                 return consume();
```
lasinicl, line 422: Now we can get rid of this case too.

lasinicl (approved):

Issue #39431, opened 2023-01-30
[Bug]: Can't use nil-conditional-expr (elvis expr) with a variable of a type reference type


MaryamZi:
### Description

$title.

### Steps to Reproduce

```ballerina
type OptionalInt int?;

// function getInt(int? i) returns int { // OK
//     return i ?: 0;
// }

function getInt(OptionalInt i) returns int { // error
    return i ?: 0; // error: operator '?:' cannot be applied to type 'OptionalInt'
}
```

### Affected Version(s)

2201.3.2

### Related area

-> Compilation

Issue #35025, opened 2022-02-11
Invalid error in `nil-conditional-expr`


KavinduZoysa:
**Description:**
$subject

**Steps to reproduce:**
```
() i = ();
io:println(i?:"NIL"); // error: operator '?:' cannot be applied to type '()'
```
Even if this error is correct, according to spec we do not need to give an error

The static type of the result of L ?: R is determined from the static types of L and R as follows:

- if the static type of L is nil, then the static type of the result is the static type of R;
- if the static type of L does not include nil, then the static type of the result is the static type of L;
- otherwise, the static type of the result is the union of the difference of the static type of L and (), and the static type of R.

**Affected Versions:**
Ballerina 2201.0.0 (Swan Lake)

PR #39580, opened 2023-02-13
Fix elvis expression for type reference types

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/39431

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -5139,8 +5139,9 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
         BType lhsType = checkExpr(elvisExpr.lhsExpr, data);
         BType actualType = symTable.semanticError;
         if (lhsType != symTable.semanticError) {
-            if (lhsType.tag == TypeTags.UNION && lhsType.isNullable()) {
-                BUnionType unionType = (BUnionType) lhsType;
+            BType referencedType = Types.getReferredType(lhsType);
+            if (referencedType.tag == TypeTags.UNION && referencedType.isNullable()) {
+                BUnionType unionType = (BUnionType) referencedType;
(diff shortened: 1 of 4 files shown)


pcnfernando (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5139,8 +5139,9 @@ public void visit(BLangElvisExpr elvisExpr, AnalyzerData data) {
         BType lhsType = checkExpr(elvisExpr.lhsExpr, data);
         BType actualType = symTable.semanticError;
         if (lhsType != symTable.semanticError) {
-            if (lhsType.tag == TypeTags.UNION && lhsType.isNullable()) {
-                BUnionType unionType = (BUnionType) lhsType;
+            BType referencedType = Types.getReferredType(lhsType);
+            if (referencedType.tag == TypeTags.UNION && referencedType.isNullable()) {
```
chiranSachintha, line 5143: For clarification, is this check correct? According to the specification, in the second step, it is mentioned that if the static type of L does not include nil, then the static type of the result is the static type of L. Why do we perform a nullable check and throw an error as `operator.not.supported`?
In spec 
```
 The static type of the result of L ?: R is determined from the static types of L and R as follows:

    if the static type of L is nil, then the static type of the result is the static type of R;
    if the static type of L does not include nil, then the static type of the result is the static type of L;
    otherwise, the static type of the result is the union of the difference of the static type of L and (), and the static type of R.


```

MaryamZi: This is being tracked by https://github.com/ballerina-platform/ballerina-lang/issues/35025.

chiranSachintha (approved):

PR #39568, opened 2023-02-10
Allow access field annotations in tuple type descriptors without a type definition at the package level

chiranSachintha:
Fixes #39552

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #39601, opened 2023-02-14
Revert "Allow access field annotations in tuple type descriptors without a type definition at the package level"

chiranSachintha:
Reverts ballerina-platform/ballerina-lang#39568


chiranSachintha:
We created this PR thinking that it caused the code coverage issue(https://github.com/ballerina-platform/ballerina-lang/issues/39609). However, upon further analysis, we have determined that this is not the root cause of the issue. Therefore, we are closing this PR.

Issue #39646, opened 2023-02-17
[Improvement]: Introduce langlib function `avg`


KavinduZoysa:
### Description

Please check https://github.com/ballerina-platform/ballerina-spec/issues/1135#issuecomment-1200665146

### Related area

-> Compilation

PR #39656, opened 2023-02-20
Introduce `avg` langlib functions

KavinduZoysa:
## Purpose

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/39646


langlib/lang.decimal/src/main/java/org/ballerinalang/langlib/decimal/Avg.java
```diff
@@ -0,0 +1,35 @@
...
+ * under the License.
+ */
+package org.ballerinalang.langlib.decimal;
+
+import io.ballerina.runtime.api.values.BDecimal;
+
+/**
+ * Native implementation of lang.decimal:avg(decimal, decimal...).
+ *
+ * @since 2201.5.0
```
gimantha, line 25:
```suggestion
 * @since 2201.6.0
```

KavinduZoysa: Changed

langlib/langlib-test/src/test/resources/test-src/decimallib_test.bal
```diff
@@ -580,6 +580,40 @@ function testQuantizeFunctionWithInvalidOutput() {
...
+
+    arr = [];
+    foreach int i in 0...1000 {
+        arr.push(<decimal>int:MAX_VALUE);
+    }
+    test:assertValueEqual(<decimal>int:MAX_VALUE, decimal:avg(<decimal>int:MAX_VALUE, ...arr));
+
+    decimal|error d = trap decimal:avg(9.999999999999999999999999999999999e6144d, 9.999999999999999999999999999999999e6144d);
+    test:assertValueEqual(true, d is error);
+    test:assertValueEqual("{ballerina}NumberOverflow", (<error>d).message());
```
KavinduZoysa, line 614: @hasithaa, when calculating the avg, first we have to get some. Therefore for long:avg function, there is a possibility to have an overflow error. This is expected, right?

hasithaa: Can we see how other implementations handle this?

KavinduZoysa: @hasithaa, I check this is Java and SQL in the MySql server and python.

1. In Java there is not an inbuilt avg function for BigDecimal (equivalent to ballerina decimal). But there is a function called `IntStream.average()` which returns `OptionalDouble`. There is no possibility to have a number overflow in this function.
2. In SQL, If we get the average of Double max, using its' average function the answer is `0`.
3. In python, it gives an overflow error for the following example
```python
import numpy
def Average(lst):
  avg = numpy.average(lst)
  return(avg)
lst = [9.999999999999999999999999999999999e307, 9.999999999999999999999999999999999e307]
print("Average of the list =", Average(last))
```

hasithaa: Ok, Can we implement something like IntStream? overflow on avg function doesn't seems right.

KavinduZoysa: `int:avg` does not overflow because it returns `decimal`. This is the same as IntStream which returns Double. 
But in our case, overflow happens to `decimal:avg`, it returns a decimal.

langlib/lang.decimal/src/main/ballerina/decimal.bal
```diff
@@ -207,3 +207,17 @@ public isolated function fromString(string s) returns decimal|error = @java:Meth
...
+
+# Returns the average of its arguments.
+#
+# ```ballerina
+# decimal:avg(10, 20, 30, 40) ⇒ 25.0
+# ```
+#
+# + n - first int value
+# + ns - other int values
+# + return - average of parameter `n` and all of parameter `ns`
```
LakshanWeerasinghe, line 219:
```suggestion
# + return - average of parameter `x` and all of parameter `xs`
```

langlib/lang.float/src/main/ballerina/float.bal
```diff
@@ -652,3 +652,16 @@ public isolated function toExpString(float x, int? fractionDigits) returns strin
     'class: "org.ballerinalang.langlib.floatingpoint.ToExpString",
     name: "toExpString"
 } external;
+
+# Returns the average of its arguments.
+# Return NaN if there are no arguments,
+# ```ballerina
```
LakshanWeerasinghe, line 658:
```suggestion
#
# ```ballerina
```

langlib/lang.int/src/main/ballerina/int.bal
```diff
@@ -215,3 +215,18 @@ public isolated function fromHexString(string s) returns int|error = @java:Metho
     'class: "org.ballerinalang.langlib.integer.FromHexString",
     name: "fromHexString"
 } external;
+
+# Returns the average of its arguments.
+# The result is returned as a decimal.
```
LakshanWeerasinghe, line 220:
```suggestion
```
I think the we don't need to specifically say this.

langlib/lang.decimal/src/main/ballerina/decimal.bal
```diff
@@ -207,3 +207,17 @@ public isolated function fromString(string s) returns decimal|error = @java:Meth
...
 } external;
+
+# Returns the average of its arguments.
+#
+# ```ballerina
+# decimal:avg(10, 20, 30, 40) ⇒ 25.0
+# ```
+#
+# + n - first int value
+# + ns - other int values
```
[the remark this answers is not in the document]
KavinduZoysa: Fixed

langlib/lang.decimal/src/main/ballerina/decimal.bal
KavinduZoysa: Fixed

langlib/lang.float/src/main/ballerina/float.bal
KavinduZoysa: Fixed

langlib/lang.int/src/main/ballerina/int.bal
KavinduZoysa: Fixed

langlib/lang.decimal/src/main/java/org/ballerinalang/langlib/decimal/Avg.java
gimantha, line 25:
```suggestion
 * @since 2201.6.0
```

langlib/lang.float/src/main/java/org/ballerinalang/langlib/floatingpoint/Avg.java
```diff
@@ -0,0 +1,37 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.ballerinalang.langlib.floatingpoint;
+
+/**
+ * Native implementation of lang.float:avg(float...).
+ *
+ * @since 2201.5.0
```
gimantha, line 23:
```suggestion
 * @since 2201.6.0
```

langlib/lang.int/src/main/java/org/ballerinalang/langlib/integer/Avg.java
```diff
@@ -0,0 +1,36 @@
...
+ * under the License.
+ */
+package org.ballerinalang.langlib.integer;
+
+import io.ballerina.runtime.api.values.BDecimal;
+
+/**
+ * Native implementation of lang.int:avg(int, int...).
+ *
+ * @since 2201.5.0
```
gimantha, line 25:
```suggestion
 * @since 2201.6.0
```

langlib/lang.decimal/src/main/java/org/ballerinalang/langlib/decimal/Avg.java
KavinduZoysa: Fixed

langlib/lang.float/src/main/java/org/ballerinalang/langlib/floatingpoint/Avg.java
KavinduZoysa: Fixed

langlib/lang.int/src/main/java/org/ballerinalang/langlib/integer/Avg.java
KavinduZoysa: Fixed

LakshanWeerasinghe (approved):
LGTM!

PR #39659, opened 2023-02-20
Include Maven custom repo support

keizer619:
```
[[platform.java11.dependency]]
groupId="org.ballerinalang"
artifactId="ballerina-command-distribution"
version="0.8.14"

[[platform.java11.repository]]
id="github-update-tool"
url = "https://maven.pkg.github.com/ballerina-platform/ballerina-update-tool"
username = "{{username}}"
password = "{{password}}"
```


cli/ballerina-cli/src/test/resources/test-resources/validProjectWithCustomMavenRepo/Ballerina.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+org = "foo"
+name = "winery"
+version = "0.1.0"
+
+[[platform.java11.dependency]]
+groupId="org.ballerinalang"
```
hevayo, line 7: If the same package is in two different repos do we need to flag from which one to resolve ?

compiler/ballerina-lang/src/main/resources/ballerina-toml-schema.json
```diff
@@ -155,7 +155,7 @@
       "properties": {
         "java11": {
           "type": "object",
-          "additionalProperties": false,
+          "additionalProperties": true,
```
hevayo, line 158: Why do we need to allow additional properties ?

keizer619: This is not needed for java11.

cli/ballerina-cli/src/test/resources/test-resources/validProjectWithCustomMavenRepo/Ballerina.toml
keizer619: Usually we do not flag in maven repositories as it stops when resolved from one repository. In that case user has to keep the order of repositories

hevayo: Ok, shall we improve the test case to have dependency rom maven central as well/

hevayo (approved):

cli/ballerina-cli/src/test/resources/test-resources/validProjectWithCustomMavenRepo/Ballerina.toml
keizer619: Currently Maven central repo scenarios are covered in Bindgen scenario tests. Maven Central Repo and local .m2 is added as a default repo where user does not need to specify in the toml.
https://github.com/ballerina-platform/ballerina-lang/blob/master/misc/ballerina-bindgen/src/test/resources/mvn-test-resources/balProject/Ballerina.toml

IMS94 (approved):

PR #39678, opened 2023-02-21
Improve main function completion item

### LINKED ISSUE #38817: [Improvement]: Improve completions after qualifiers in module part
### Description

For an example, when we have already typed `public` keyword(or a visibility qualifier), the current completions we get are not that helpful for the users, and in some cases they can be incorrect as well. 

Consider following two scenarios,

1. 
Screenshot 2022-11-24 at 14 49 03
2. 
Screenshot 2022-11-24 at 14 49 12

This is not limited to the `public` keyword. If we type other qualifiers, for an example `isolated` keyword, these cases also can be improved. So in general we need to consider all qualifier types and find an approach to improve completions.

Screenshot 2022-11-24 at 14 50 17

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

-> Editor

_No response_

_No response_

_No response_

mindula:
## Purpose
This PR improves the completion item for main function.

Fixes #38817

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/FunctionCompletionItemBuilder.java
```diff
@@ -373,6 +378,40 @@ private static String getQualifiedFunctionName(String functionName, BallerinaCom
...
+            lastQualifier = CommonUtil.getLastQualifier(context, node);
+            if (lastQualifier.isPresent() || node.kind() == SyntaxKind.MODULE_PART) {
+                break;
+            }
+            node = node.parent();
+        }
+
+        CompletionItem completionItem = new CompletionItem();
+        String insertText;
+        if (lastQualifier.isPresent() && lastQualifier.get().text().contains(ItemResolverConstants.PUBLIC_KEYWORD)) {
```
IMS94, line 400: We can check `lastQualifier.get().kind() == SyntaxKind.PUBLIC_KEYWORD` instead of doing a string contains check

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MainFunctionCompletionItem.java
```diff
@@ -0,0 +1,40 @@
...
+import org.ballerinalang.langserver.commons.BallerinaCompletionContext;
+import org.ballerinalang.langserver.commons.completion.AbstractLSCompletionItem;
+import org.eclipse.lsp4j.CompletionItem;
+
+/**
+ * Completion item for main function.
+ *
+ * @since 2201.4.0
+ */
+public class MainFunctionCompletionItem extends AbstractLSCompletionItem {
```
IMS94, line 29: Shall we rename this class to `FunctionCompletionItem` because it doesn't do anything specific to the main function

language-server/modules/langserver-core/src/test/resources/completion/enum_decl_ctx/config/config5.json
```diff
@@ -137,7 +137,7 @@
       "detail": "Snippet",
       "sortText": "AA",
       "filterText": "public_function_main",
-      "insertText": "public function main() {\n\t${1}\n}",
+      "insertText": "function main() {\n\t${1}\n}",
```
IMS94, line 140: This case seems to be wrong. `public` should be available in the insert text here.

language-server/modules/langserver-core/src/test/resources/completion/module_part_context/config/module_level_after_configurable_config3.json
```diff
@@ -678,7 +678,7 @@
       "detail": "Snippet",
       "sortText": "D",
       "filterText": "public_function_main",
-      "insertText": "public function main() {\n\t${1}\n}",
+      "insertText": "function main() {\n\t${1}\n}",
```
IMS94, line 681: We should check this case too. `configurable <cursor>`, here we aren't adding the `public` keyword. I don't think we should suggest the main function altogether here. WDYT?

IMS94 (changes requested):
Added few comments

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/FunctionCompletionItemBuilder.java
mindula: done

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/MainFunctionCompletionItem.java
mindula: +1

language-server/modules/langserver-core/src/test/resources/completion/enum_decl_ctx/config/config5.json
mindula: Fixed with [f0d1cdc](https://github.com/ballerina-platform/ballerina-lang/pull/39678/commits/f0d1cdce329ee8533e9179772f75fb904a89f141)

language-server/modules/langserver-core/src/test/resources/completion/module_part_context/config/module_level_after_configurable_config3.json
mindula: +1

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/FunctionCompletionItemBuilder.java
```diff
@@ -373,6 +379,40 @@ private static String getQualifiedFunctionName(String functionName, BallerinaCom
...
+    public static CompletionItem buildMainFunction(BallerinaCompletionContext context) {
+        NonTerminalNode node = context.getNodeAtCursor();
+        Optional<Token> lastQualifier = Optional.empty();
+        while (node != null) {
+            lastQualifier = ModulePartNodeContextUtil.getLastQualifier(context, node);
+            if (lastQualifier.isPresent() || node.kind() == SyntaxKind.MODULE_PART) {
+                break;
+            }
+            node = node.parent();
+        }
```
malinthar, line 397: Shall we embed the parent visiting logic inside the  ModulePartNodeContextUtil.getLastQualifier method. Because ideally, we should not have to visit the parents to get the last qualifier. Are there any breaking cases if we do so?

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/FunctionCompletionItem.java
```diff
@@ -0,0 +1,40 @@
...
+import org.ballerinalang.langserver.commons.BallerinaCompletionContext;
+import org.ballerinalang.langserver.commons.completion.AbstractLSCompletionItem;
+import org.eclipse.lsp4j.CompletionItem;
+
+/**
+ * Completion item for main function.
+ *
+ * @since 2201.4.0
+ */
+public class FunctionCompletionItem extends AbstractLSCompletionItem {
```
malinthar, line 29: Having a `FunctionCompletionItem` can be confusing as we don't use this to only main function completion item. Is there a specific reason to use a separate class for this without using the FunctionCompletionItemBuilder other than the sorting purposes? If not shall we explore the possibility of using the `SnippetCompletionItem`

malinthar (commented):
Let's add a test covering the following case.

```Ballerina
public <cursor>

```

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/FunctionCompletionItem.java
IMS94: As I remember the reason was that we cannot match the constructor of `SnippetCompletionItem` here because we don't have a snippet block. `main()` function snippet is not static anymore due to handling the presence of `public` keyword.

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/builder/FunctionCompletionItemBuilder.java
IMS94: The module part context is very unpredictive and I think that's why we had to add this logic. @mindula can add why we needed the parent visit thing. I don't think we should make the parent visit thing common by moving to the util

malinthar: +1 to having the logic as is then

mindula: There are some cases where we need to visit the parents to get the last qualifier. For example, consider the following case,
```
public <cursor>
function name() {

}
```
when we try to get the last qualifier using ModulePartNodeContextUtil.getLastQualifier method for the above cursor position, it gives Optional.empty()

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/context/FunctionCompletionItem.java
mindula: It was originally named MainFunctionCompletionItem and @IMS94 suggested that it would be better to rename it to FunctionCompletionItem as it doesn't contain anything specific to the main function. WDYT?

malinthar: We can address this later as this is not a critical improvement

malinthar (dismissed):

IMS94 (approved):

Issue #39682, opened 2023-02-22
[Bug]: New function mocking fails with an NPE at moduleInit


gayaldassanayake:
### Description

Function mocking with new mocking framework fails when there is a module level function call for the function that is being mocked.

The stack trace is as below.

```
Compiling source
        gayaldassanayake/mocknpe:0.1.0

Running Tests

        mocknpe
error: dependant module initialization for test suite failed due to java.lang.NullPointerException
        at gayaldassanayake.mocknpe$test.0:$MOCK_clientFunction(mocknpe:1)
           gayaldassanayake.mocknpe.0:clientFunction(main:-1)
           gayaldassanayake:.<init>(mocknpe:2)
           gayaldassanayake:$moduleInit(mocknpe:-1)
           jdk.internal.reflect.NativeMethodAccessorImpl:invoke0(NativeMethodAccessorImpl.java:-2)
           jdk.internal.reflect.NativeMethodAccessorImpl:invoke(NativeMethodAccessorImpl.java:62)
           jdk.internal.reflect.DelegatingMethodAccessorImpl:invoke(DelegatingMethodAccessorImpl.java:43)
           java.lang.reflect.Method:invoke(Method.java:566)
           org.ballerinalang.test.runtime.entity.TesterinaFunction:lambda$runOnSchedule$0(TesterinaFunction.java:109)
           io.ballerina.runtime.internal.scheduling.SchedulerItem:execute(SchedulerItem.java:54)
           io.ballerina.runtime.internal.scheduling.Scheduler:run(Scheduler.java:324)
           io.ballerina.runtime.internal.scheduling.Scheduler:runSafely(Scheduler.java:291)
           java.lang.Thread:run(Thread.java:829)
error: there are test failures
```

However, when the function call is done within another function.
eg-
1. This causes the error. (Full code in the next section)
``` ballerina
string a = clientFunction();

function clientFunction() returns string {
    return "Hello from the original";
}
```

2. This does not.
``` ballerina
public function main() {
    string a = clientFunction();
}

function clientFunction() returns string {
    return "Hello from the original";
}
```

### Steps to Reproduce

Create a simple ballerina project with `main.bal` and `tests/test.bal` files as follows.

`main.bal`
``` ballerina
string a = clientFunction();

function clientFunction() returns string {
    return "Hello from the original";
}
```

`tests/test.bal`
``` ballerina
import ballerina/test;

@test:Mock {
    functionName: "clientFunction"
}
[...]

PR #39562, opened 2023-02-10
[master] Fix testerina-integration-test

Thevakumar-Luheerathan:
Fixes #39561

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #39684, opened 2023-02-22
Fix `ClassFormatError: Truncated class file` error when mocking in multi module packages.

gayaldassanayake:
## Purpose
Map `classVsMockFunctionsMap` in BTestMain keeps track of the class to mockfunction mapping. In the current implementation, we are not clearing it after running each module. So it tries to mock functions that are mocked by previous modules (and not mocked in the current module). 

As a result the modified class files becomes an empty string. This is because the function we are trying to mock is not available in the module. Hence a truncated class file error is received due to the modified class missing some of the byte code. 

Fixes https://github.com/wso2-enterprise/internal-support-ballerina/issues/266

## Approach
Clear `classVsMockFunctionsMap` after replacing the mock functions of each module. 

Note - tests/testerina-integration-test/src/test/resources/project-based-tests/function-mocking-tests-with-dependencies/modules/a_dependency/main.bal#18 needs to be uncommented to test the fix for the new mocking framework once https://github.com/ballerina-platform/ballerina-lang/issues/39682 is fixed.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/MockTest.java
+++ b/tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/MockTest.java
@@ -144,6 +144,19 @@ public class MockTest extends BaseTestCase {
         Assert.assertEquals(moduleObj.get("coveragePercentage").toString(), "57.14");
     }
 
+    @Test
+    public void testFunctionMockingInMultipleModulesWithDependencies() throws BallerinaTestException {
+        String msg1 = "function_mocking_with_dependencies.a_dependency\n\t\t[pass] test1\n\n\n\t\t" +
+                "1 passing\n\t\t0 failing\n\t\t0 skipped";
+        String msg2 = "function_mocking_with_dependencies.b_dependent\n\t\t[pass] test1\n\n\n\t\t" +
+                "1 passing\n\t\t0 failing\n\t\t0 skipped";
+        String[] args = mergeCoverageArgs(new String[]{"function-mocking-tests-with-dependencies"});
+        String output = balClient.runMainAndReadStdOut("test", args, new HashMap<>(), projectPath, false);
+        if (!output.contains(msg1) || !output.contains(msg2)) {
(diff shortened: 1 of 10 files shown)


tests/testerina-integration-test/src/test/java/org/ballerinalang/testerina/test/MockTest.java
```diff
@@ -144,6 +144,19 @@ public void testCoverageWithMocking() throws BallerinaTestException {
...
+    @Test
+    public void testFunctionMockingInMultipleModulesWithDependencies() throws BallerinaTestException {
+        String msg1 = "function_mocking_with_dependencies.a_dependency\n\t\t[pass] test1\n\n\n\t\t" +
+                "1 passing\n\t\t0 failing\n\t\t0 skipped";
+        String msg2 = "function_mocking_with_dependencies.b_dependent\n\t\t[pass] test1\n\n\n\t\t" +
+                "1 passing\n\t\t0 failing\n\t\t0 skipped";
+        String[] args = mergeCoverageArgs(new String[]{"function-mocking-tests-with-dependencies"});
+        String output = balClient.runMainAndReadStdOut("test", args, new HashMap<>(), projectPath, false);
+        if (!output.contains(msg1) || !output.contains(msg2)) {
+            Assert.fail("Test failed due to function mocking failure in test framework.\nOutput:\n" + output);
```
Dilhasha, line 156: Shall we update the test case to have the expected output in a file and compare in order to be consistent with the new integration test improvements?

gayaldassanayake: Added the output files for the test. As per an offline discussion with @Thevakumar-Luheerathan we can change the test case to read from the output file instead of the string in https://github.com/ballerina-platform/ballerina-lang/pull/39562

Thevakumar-Luheerathan (approved):

Issue #39699, opened 2023-02-22
[Bug]: Compiler crashes with 'NoSuchElementException' when executing map function on empty array


chiranSachintha:
### Description

$title,
```ballerina
import ballerina/io;

public function main() {
    var data = [].map( m => m + 1);
    io:println(data);
}
```
The program will crash with the following error message:
```
[2023-02-23 00:08:42,840] SEVERE {b7a.log.crash} - null 
java.util.NoSuchElementException
        at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:721)
        at java.base/java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:741)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.Types.validNumericStringOrXmlTypeExists(Types.java:4326)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.Types.validNumericTypeExists(Types.java:4387)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolResolver.getArithmeticOpsForTypeSets(SymbolResolver.java:1937)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.visit(TypeChecker.java:5015)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.visit(TypeChecker.java:213)
        at org.wso2.ballerinalang.compiler.tree.expressions.BLangBinaryExpr.accept(BLangBinaryExpr.java:72)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.checkExpr(TypeChecker.java:408)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.checkExpr(TypeChecker.java:345)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.populateArrowExprReturn(TypeChecker.java:6315)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.visit(TypeChecker.java:5529)
        at org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker.visit(TypeChecker.java:213)
        at org.wso2.ballerinalang.compiler.tree.expressions.BLangArrowFunction.accept(BLangArrowFunction.java:79)
[...]

Issue #39700, opened 2023-02-22
[Bug]: No error when for invalid additive expression when one operand is a union of members that belong to the same basic type


chiranSachintha:
### Description

The following program compiles without any errors, but it should be a compile-time error. The program attempts to concatenate a string and an int, which is not allowed. The compiler should have detected this type mismatch and thrown an error during the compilation process.

```ballerina
import ballerina/io;

public function main() {
    string:Char o = "s";
    var data = ["10", o].map( m => m + 1);
    io:println(data);
}
```

### Affected Version(s)

2201.5.0-SNAPSHOT

### Related area

-> Compilation

### Related issue(s) (optional)

https://github.com/ballerina-platform/ballerina-lang/issues/39680

MaryamZi:
This isn't an anonymous function issue. Seems to be a bug via `org.wso2.ballerinalang.compiler.semantics.analyzer.Types#validNumericTypeExists`.

```ballerina
public function main() {
    string|string:Char x = "";
    _ = x + 1;
}
```

chiranSachintha:
We need to fix the following kinds of scenarios as well. This should be a compile-time error. Currently, it is compiling without any errors.
```ballerina
import ballerina/io;

public function main() {
    var data = [<string|string>"10"].map( m => m + 1);
    io:println(data);
}
````

**Comment by MaryamZi** [comment]
```ballerina
public function main() {
    string|("a"|"b") a = "a";
    _ = a + 1; // no error
    _ = 1 + a;

    ("a"|"b")|string b = "a";
    _ = b + 1; // no error
    _ = 1 + b;

    string|string:Char c = "s";
    _ = c + 1; // no error
    _ = 1f + c;

    string|string d = "";
    _ = d + 1; // no error
    _ = 1 + d;

    (1|2)|int e = 1;
    _ = e + "";
    _ = "" + e; // no error

    xml:Element|xml:Text f = xml ``;
    _ = f + 1;
    _ = 1 + f; // no error
}
```

PR #39681, opened 2023-02-22
Fix incorrectly creating a union with one member as the inferred parameter type with tuples

MaryamZi:
## Purpose
$title, this was causing invalid analysis elsewhere where we assume a union to have > 1 member.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/39680

## Check List 
- [x] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #39702, opened 2023-02-23
Fix incorrectly creating a union with no members as the inferred parameter type with empty tuples

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/39699

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
@@ -581,7 +581,9 @@ public class TypeParamAnalyzer {
             tupleTypes.add(actualType.restType);
         }
 
-        BType type = tupleTypes.size() == 1 ? tupleTypes.iterator().next() : BUnionType.create(null, tupleTypes);
+        int size = tupleTypes.size();
+        BType type = size == 0 ? symTable.neverType :
+                (size == 1 ? tupleTypes.iterator().next() : BUnionType.create(null, tupleTypes));
         findTypeParam(loc, expType.eType, type, env, resolvedTypes, result);
     }
 
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/lambda/ArrowExprTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/lambda/ArrowExprTest.java
@@ -238,6 +238,15 @@ public class ArrowExprTest {
                                   "operator '+' not defined for 'string' and 'int'", 87, 25);
         BAssertUtil.validateError(resultSemanticsNegative, i++,
                                   "operator '+' not defined for 'int' and 'string'", 90, 25);
+        BAssertUtil.validateError(resultSemanticsNegative, i++,
+                                  "cannot define a variable of type 'never' or equivalent to type 'never'", 92, 16);
+        BAssertUtil.validateError(resultSemanticsNegative, i++,
+                                  "operator '+' not defined for 'never' and 'int'", 92, 21);
+//        https://github.com/ballerina-platform/ballerina-lang/issues/26191
+//        BAssertUtil.validateError(resultSemanticsNegative, i++,
+//                                  "cannot define a variable of type 'never' or equivalent to type 'never'", 95, 22);
+//        BAssertUtil.validateError(resultSemanticsNegative, i++,
+//                                  "operator '+' not defined for 'int' and 'never'", 95, 27);
         Assert.assertEquals(resultSemanticsNegative.getErrorCount(), i);
     }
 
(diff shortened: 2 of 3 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeParamAnalyzer.java
```diff
@@ -581,7 +581,9 @@ private void findTypeParamInTupleForArray(Location loc, BArrayType expType, BTup
             tupleTypes.add(actualType.restType);
         }
 
-        BType type = tupleTypes.size() == 1 ? tupleTypes.iterator().next() : BUnionType.create(null, tupleTypes);
+        int size = tupleTypes.size();
+        BType type = size == 0 ? symTable.neverType :
+                (size == 1 ? tupleTypes.iterator().next() : BUnionType.create(null, tupleTypes));
```
chiranSachintha, line 586: IMO, this is not the ideal approach to fix this issue. Consider the following examples, which will compile without any errors.
```ballerina
import ballerina/io;

public function main() {
    [string|string] k = ["10"];
    var data = k.map( m => m + 1);
    io:println(data);  //prints  ["101"]
}

```
```ballerina
import ballerina/io;

public function main() {
    [string|never] k = [];
    var data = k.map( m => m + 1);
    io:println(data); // prints ["1"]
}
```

chiranSachintha: Related to [this](https://github.com/ballerina-platform/ballerina-lang/pull/39681) PR as well.

MaryamZi: Aren't these due to https://github.com/ballerina-platform/ballerina-lang/issues/39700?

chiranSachintha: my concern is If we write something like this, it will still compile without any errors
```ballerina
import ballerina/io;

public function main() {
    var data = [<string|string>"10"].map( m => m + 1);
    io:println(data);
}
```
With your fix, we retrieve the type of the first element in the tuple in this instance and check if the operator for addition is available. In this case, the first operator is a union type(string), so follow the previous path.

MaryamZi: And that's still due to the same issue mentioned in https://github.com/ballerina-platform/ballerina-lang/issues/39700? The fix here is required irrespective of changes to `validNumericStringOrXmlTypeExists`. We shouldn't be creating unions with < 2 types.

chiranSachintha: So, if we fix that issue, do we need to special case when the type size is equal to 1? We can create a union type for that as well, right?

MaryamZi: A union means more than one type, so we still need to avoid creating a union when there's only one member.

chiranSachintha: We created a union type with one member(string) for the following scenario and set that type to the variable k.
```ballerina
public function main() {
    string|string k = "10";
}
```
Is it correct?

MaryamZi: No, it should either be `string` or the union of `string|string` which has a resultant type `string` (like we have effective type for intersections).

MaryamZi: > the union of `string|string` which has a resultant type `string` (like we have effective type for intersections)

which we don't have yet

chiranSachintha: Okay, I understand. That is the reason why this scenario compiles without any error after your changes also. Do we have an issue to track [this](https://github.com/ballerina-platform/ballerina-lang/pull/39702#discussion_r1115335605)?
```ballerina
import ballerina/io;

public function main() {
    var data = [<string|string>"10"].map( m => m + 1);
    io:println(data);
}
```

MaryamZi: The union revamp is something we've deferred. I believe fixing https://github.com/ballerina-platform/ballerina-lang/issues/39700 should fix this anyway. We may have to consider (incorrectly created) unions with one member.

chiranSachintha: Create https://github.com/ballerina-platform/ballerina-lang/issues/39714

chiranSachintha (approved):

Issue #39669, opened 2023-02-20
[Bug]: Error when using `nil-conditional-expr` with JSON type


chiranSachintha:
### Description

In the given program, when attempting to use the nil-coalescing operator ?: with the json type, an error message is generated stating that the operator cannot be applied to that type. The "json" type is a union of the following types: (), boolean, int, float, decimal, string, json[], and map<json>. It is expected that the program should work correctly with this type union.
```ballerina
function getJson(json i) returns json {
    return i ?: 10;
}

```
If we make the changes to the program as shown below, the program works correctly.
```ballerina
function getJson(json? i) returns json {
    return i ?: 10;
}

```

### Affected Version(s)

2201.5.0-SNAPSHOT

### Related area

-> Compilation

MaryamZi:
Happens for built-in unions that include nil - `anydata`, `any`, and `json`.

PR #39810, opened 2023-03-07
Allow using the nil conditional expressions with `json`, `anydata`, and `any`

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/39669

Note: since we cannot represent type - nil (a type by removing nil) at the moment for `json`, `anydata`, and `any`, the type of `L` in `L ?: R`, where `L` is originally of `json`, `anydata`, or `any`, is considered to be the original type itself.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
@@ -5109,16 +5109,22 @@ public class TypeChecker extends SimpleBLangNodeAnalyzer<TypeChecker.AnalyzerDat
         BType actualType = symTable.semanticError;
         if (lhsType != symTable.semanticError) {
             BType referencedType = Types.getReferredType(lhsType);
-            if (referencedType.tag == TypeTags.UNION && referencedType.isNullable()) {
-                BUnionType unionType = (BUnionType) referencedType;
-                LinkedHashSet<BType> memberTypes = unionType.getMemberTypes().stream()
-                        .filter(type -> type.tag != TypeTags.NIL)
-                        .collect(Collectors.toCollection(LinkedHashSet::new));
+            if (referencedType.isNullable()) {
+                if (referencedType.tag == TypeTags.UNION) {
+                    BUnionType unionType = (BUnionType) referencedType;
+                    LinkedHashSet<BType> memberTypes = unionType.getMemberTypes().stream()
+                            .filter(type -> type.tag != TypeTags.NIL)
+                            .collect(Collectors.toCollection(LinkedHashSet::new));
 
-                if (memberTypes.size() == 1) {
-                    actualType = memberTypes.toArray(new BType[0])[0];
+                    if (memberTypes.size() == 1) {
+                        actualType = memberTypes.toArray(new BType[0])[0];
+                    } else {
+                        actualType = BUnionType.create(null, memberTypes);
+                    }
                 } else {
-                    actualType = BUnionType.create(null, memberTypes);
+                    // This is a builtin union type (any, anydata, json), we use the type as is
+                    // since we don't have a way to represent the union - nil at the moment.
+                    actualType = referencedType;
                 }
             } else {
                 dlog.error(elvisExpr.pos, DiagnosticErrorCode.OPERATOR_NOT_SUPPORTED, OperatorKind.ELVIS,
--- a/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/elvis/ElvisExpressionTest.java
+++ b/tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/expressions/elvis/ElvisExpressionTest.java
@@ -236,6 +236,11 @@ public class ElvisExpressionTest {
         BRunUtil.invoke(compileResult, "testElvisExprWithIntersectionTypes");
     }
 
+    @Test
+    public void testElvisExprWithBuiltInNilableUnion() {
+        BRunUtil.invoke(compileResult, "testElvisExprWithBuiltInNilableUnion");
+    }
+
     @Test(description = "Negative test cases.")
     public void testElvisOperatorNegative() {
         int index = 0;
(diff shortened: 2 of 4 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5109,16 +5109,22 @@ public void visit(BLangElvisExpr elvisExpr, AnalyzerData data) {
...
+                    if (memberTypes.size() == 1) {
+                        actualType = memberTypes.toArray(new BType[0])[0];
+                    } else {
+                        actualType = BUnionType.create(null, memberTypes);
+                    }
                 } else {
-                    actualType = BUnionType.create(null, memberTypes);
+                    // This is a builtin union type (any, anydata, json), we use the type as is
+                    // since we don't have a way to represent the union - nil at the moment.
+                    actualType = referencedType;
```
chiranSachintha, line 5127: By setting the actual type as reference type, we are unable to perform this kind of assignment. Is there an issue for this?
```ballerina
type Foo boolean|int|float|decimal|string|json[]|map<json>;
function getJson(json i) {
    Foo k = i ?: 10;
}
```

MaryamZi: This is what I meant by the comment.

Doesn't work similarly with an is check

```ballerina
type Foo boolean|int|float|decimal|string|json[]|map<json>;

function getJson(json i) {
    Foo f;
    if i !is () {
        f = i;
    } else {
        f = 10;
    }
}
```

I don't think there's an issue specifically for this, but I believe we have issues related to narrowing.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5109,16 +5109,22 @@ public void visit(BLangElvisExpr elvisExpr, AnalyzerData data) {
         BType actualType = symTable.semanticError;
         if (lhsType != symTable.semanticError) {
             BType referencedType = Types.getReferredType(lhsType);
-            if (referencedType.tag == TypeTags.UNION && referencedType.isNullable()) {
```
chiranSachintha, line 5112: Not directly related to this PR. However, I believe it would be better if we could refactor this logic to be similar to this
```java
BType actualType = getActualType(lhsType, elvisExpr);

private BType getActualType(BType lhsType, BLangElvisExpr elvisExpr) {
    if (lhsType == symTable.semanticError) {
        return symTable.semanticError;
    }

    BType referencedType = Types.getReferredType(lhsType);

    if (!referencedType.isNullable()) {
        dlog.error(elvisExpr.pos, DiagnosticErrorCode.OPERATOR_NOT_SUPPORTED, OperatorKind.ELVIS, lhsType);
        return symTable.semanticError;
    }

    if (referencedType.tag != TypeTags.UNION) {
        // This is a builtin union type (any, anydata, json), we use the type as is
        // since we don't have a way to represent the union - nil at the moment.
        return referencedType;
    }

    BUnionType unionType = (BUnionType) referencedType;
    LinkedHashSet<BType> memberTypes = unionType.getMemberTypes().stream()
            .filter(type -> type.tag != TypeTags.NIL)
            .collect(Collectors.toCollection(LinkedHashSet::new));

    if (memberTypes.size() == 1) {
        return memberTypes.iterator().next();
    } else {
        return BUnionType.create(null, memberTypes);
    }
}
```

MaryamZi: I don't see a significant benefit (other than maybe length) by doing this. The name `getActualType` is ambiguous too, right? Doesn't quite convey anything.

chiranSachintha: Yes, the only benefit is that it makes the code easier to read. If so, we can keep the code as it is now.

chiranSachintha (approved):

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: In this particular case, I don't think it makes (much of) a difference in that aspect too.

chiranSachintha: Yes, I agree with you. I suggest this because I saw several nested if-else statements. (I tried to remove those things.)

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Was actually able to fix this for `json` and `anydata`, since they are modeled as unions. Updated the PR.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
MaryamZi: Ended up extracting out a method since the logic became lengthier with the latest changes.

compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -543,6 +543,9 @@ error.never.typed.var.def.not.allowed=\
 error.never.typed.object.field.not.allowed=\
   cannot define an object field of type ''never'' or equivalent to type ''never''
 
+error.nil.conditional.expr.not.yet.supported.with.nil=\
+  using ''?:'' with an expression of a type that is a subtype of nil is not yet supported
```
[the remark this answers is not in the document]
MaryamZi: Because the type can be a subtype like `never?` or a type reference type.

chiranSachintha (approved):

PR #39765, opened 2023-03-01
Correct the errors thrown when `bal test --rerun-failed` is used without a `rerun_test.json` file 

gayaldassanayake:
## Purpose
$subject

Fixes #39664

## Approach
If the rerun_test.json is missing, throw an error requesting user to run bal test and terminate without running the tests.

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #39665, opened 2023-02-20
Add diagnostics for single file legacy mocking

gayaldassanayake:
## Purpose
$ subject

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/39643

## Check List 
- [ ] Read the [Contributing Guide](https://github.com/ballerina-platform/ballerina-lang/blob/master/CONTRIBUTING.md)
- [ ] Updated Change Log
- [ ] Checked Tooling Support (#<Issue Number>)
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
  - [ ] Ballerina By Example Tests
- [ ] Increased Test Coverage   
- [ ] Added necessary documentation  
  - [ ] API documentation 
  - [ ] Module documentation in Module.md files
  - [ ] Ballerina By Examples

PR #39821, opened 2023-03-08
Fix `thenReturn(())`, `doNothing()` not returning a nil value

### LINKED ISSUE #32903: [Testerina] Function mocking `thenReturn` isnt compatible with `()` nil return value
**Description:**
Having `thenReturn(())` does not return a nil value but rather causes the inbuilt validation to fail as it considers this as "return value not being set". 

Same applies to `doNothing()` function which is supposed to skip execution entirely.

**Steps to reproduce:**

**Affected Versions:**
`beta2`

gayaldassanayake:
## Purpose
$subject

Fixes #32903

## Approach
Removed the validation that checks if returnVal == null and throws an error.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/mock/FunctionMock.java
+++ b/misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/mock/FunctionMock.java
@@ -56,10 +56,9 @@ public class FunctionMock {
         String originalClass = splitInfo[2];
 
         originalFunctionPackage = getQualifiedClassName(originalOrg, originalPackage, originalVersion, originalClass);
-        Object returnVal = null;
         for (String caseId : caseIds) {
             if (MockRegistry.getInstance().hasCase(caseId)) {
-                returnVal = MockRegistry.getInstance().getReturnValue(caseId);
+                Object returnVal = MockRegistry.getInstance().getReturnValue(caseId);
                 if (returnVal instanceof BString) {
                     if (returnVal.toString().contains(MockConstants.FUNCTION_CALL_PLACEHOLDER)) {
                         return callMockFunction(originalFunction, originalFunctionPackage,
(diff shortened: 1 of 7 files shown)


misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/mock/FunctionMock.java
```diff
@@ -73,15 +73,6 @@ public static Object mockHandler(BObject mockFuncObj, Object... args) {
                 break;
             }
         }
-        if (returnVal == null) {
```
Dilhasha, line 76: Can we still detect if there is no return value registered with this change?

i.e. when the mock function is defined but `thenReturn`/`doNothing` is not specified within a test case?

gayaldassanayake: It cannot differentiate between the two scenarios. Will work on a fix for that 👍

gayaldassanayake: Updated the logic to differentiate two scenarios and updated tests to match each case.

Dilhasha (approved):

gayaldassanayake:
To get the Windows checks passing, need to rebase once https://github.com/ballerina-platform/ballerina-lang/pull/39665 and https://github.com/ballerina-platform/ballerina-lang/pull/39765 are merged.

azinneera (dismissed):

azinneera (approved):

PR #39874, opened 2023-03-14
Bump dependencies if a higher distribution than `distribution-version` in Dependencies.toml is used

### LINKED ISSUE #39850: [Improvement]: Introduce distribution field to Dependencies.toml and bump dependencies if a higher distribution is used
### Description

Currently, the `Dependencies.toml` does not record the distribution version that is used to build the ballerina project. However it records the versions of dependencies that the previous build has used. Moreover, the dependencies are only bumped in patch versions from what is specified in `Dependencies.toml`.

### Describe your problem(s)

_No response_

### Describe your solution(s)

The above needs to be changed as follows.
1. `Dependencies.toml` should keep track of the distribution version used as follows.
    ``` toml
    [ballerina]
    dependencies-toml-version = "2"
    distribution-version = "2201.5.0"
    ``` 
    In order to be backward compatible, field `distribution-version` should be optional.
2. If a project is built using a newer distribution version than what is recored in the `Dependencies.toml`, the dependencies should be bumped to the latest minor version compatible with that distribution version. 
3. If the distribution version is not mentioned in the `Dependencies.toml`, the dependency versions should be bumped similar to (2).

### Related area

-> Compilation

_No response_

_No response_

_No response_

gayaldassanayake:
## Purpose
$ subject

Fixes #39850

## Approach
1. A field named `distribution-version` is introduced to Dependencies.toml. This contains the Ballerina version used in the build.
2. Update policy is changed to soft, iff the `distribution-version` in Dependencies.toml is lower than the distribution version of the current build.

## Remarks
TODOs
- [x] Add `distribution-version` field to Dependencies.toml
- [ ] Add test cases to validate the new field
- [ ] Add logic to set package locking to soft
- [ ] Add package resolution tests
- [ ] Add project-api-test tests

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/projects/DependencyManifest.java
```diff
@@ -35,14 +35,18 @@
 public class DependencyManifest {
 
     private final String dependenciesTomlVersion;
+    // TODO: can have a PackageVersion or String instead. Seems like semver is the best matching class. Need to check
+    private final SemanticVersion distributionVersion;
```
gayaldassanayake, line 39: can have a PackageVersion or String instead. Seems like semver is the best matching class. Any thoughts?

azinneera: +1 to have SemanticVersion.

gayaldassanayake:
Manually tested the following scenarios and the changes works as expected.

1. There is no Dep.toml and newly build ✅ 
2. Bump the version of a dir dependency( only patch should be present) and rerun with the same distribution version✅ 
3. Bump the version of a dir dependency(all patch, minor & next major should be present) and rerun with the same distribution version ✅ 
4. Bump the version of a dir dependency(only patch should be present) and run with a newer distribution version ✅ 
5. Bump the version of a dir dependency(all patch, minor & next major should be present) and run with a newer distribution version ✅ 
6. Bump the version of a dir dependency(all patch, minor & next major should be present) and run with a newer distribution version without the distribution-version field ✅ 
7. Bump the version of an indirect dependency( only patch should be present) and rerun with the same distribution version ✅ 
8. Bump the version of a indirect dependency(all patch, minor & next major should be present) and rerun with the same distribution version ✅ 
9. Bump the version of a indirect dependency(only patch should be present) and run with a newer distribution version ✅ 
10. Bump the version of a indirect dependency(all patch, minor & next major should be present) and run with a newer distribution version ✅ 
11. Bump the version of a indirect dependency(all patch, minor & next major should be present) and run with a newer distribution version without the distribution-version field ✅

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -485,11 +485,22 @@ private BlendedManifest createBlendedManifest(PackageContext rootPackageContext,
 
     private ResolutionOptions getResolutionOptions(PackageContext rootPackageContext,
                                                    CompilationOptions compilationOptions) {
+        boolean sticky = getSticky(rootPackageContext);
+        PackageLockingMode packageLockingMode;
+        SemanticVersion prevDistributionVersion = rootPackageContext.dependencyManifest().distributionVersion();
+        SemanticVersion currentDistributionVersion = SemanticVersion.from(RepoUtils.getBallerinaShortVersion());
+        if (!sticky && (null == prevDistributionVersion
+                || currentDistributionVersion.greaterThan(prevDistributionVersion))) {
```
azinneera, line 493: We must issue a warning diagnostic informing the user about the automatic update policy change.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/DependencyManifestBuilder.java
```diff
@@ -177,6 +181,33 @@ private String getDependenciesTomlVersion() {
...
+
+        TopLevelNode ballerinaEntries = tomlTableNode.entries().get("ballerina");
+        if (ballerinaEntries == null || ballerinaEntries.kind() == TomlType.NONE) {
+            return null;
+        }
+
+        if (ballerinaEntries.kind() == TomlType.TABLE) {
+            TomlTableNode ballerinaTableNode = (TomlTableNode) ballerinaEntries;
+            String distributionVersionString = getStringValueFromDependencyNode(ballerinaTableNode,
+                    "distribution-version");
```
azinneera, line 202: Let's make this a constant

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
gayaldassanayake: Should the warning diagnostic shown only if there is a distribution version increment (i.e. only if the if condition in L492 met)?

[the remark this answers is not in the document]
gayaldassanayake: The current PR logic sets the locking mode to SOFT for any distribution increment. Will fix that to match the requirement.

azinneera: Only if the condition is met. 
E.g:
`WARNING: Detected packages built with an older SL Update distribution. Dependencies will be updated to the latest versions compatible with the current distribution.
`

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -485,14 +486,41 @@ private BlendedManifest createBlendedManifest(PackageContext rootPackageContext,
 
     private ResolutionOptions getResolutionOptions(PackageContext rootPackageContext,
                                                    CompilationOptions compilationOptions) {
+        boolean sticky = getSticky(rootPackageContext);
+        PackageLockingMode packageLockingMode;
+        SemanticVersion prevDistributionVersion = rootPackageContext.dependencyManifest().distributionVersion();
+        SemanticVersion currentDistributionVersion = SemanticVersion.from(RepoUtils.getBallerinaShortVersion());
+        if (!sticky && (null == prevDistributionVersion
```
azinneera, line 493: Let's improve this logic to avoid printing the warning when the Dependencies.toml is not present. i.e We don't need to warn if it is a fresh project.

E.g:
```
if  not sticky
      if  prevDistributionVersion is null
            locking mode = SOFT
      else if isNewUpdateDistribution
            report warning
            locking mode = SOFT
      else 
            locking mode = MEDIUM
else 
      locking mode = MEDIUM
```

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticWarningCode.java
```diff
@@ -69,7 +69,9 @@ public enum DiagnosticWarningCode implements DiagnosticCode {
 
     INVALID_METADATA_ON_DUPLICATE_ENUM_MEMBER("BCE20405", "invalid.metadata.on.duplicate.enum.member"),
 
-    USAGE_OF_STRAND_ANNOTATION_WILL_BE_DEPRECATED("BCE20406", "strand.annotation.will.be.deprecated")
+    USAGE_OF_STRAND_ANNOTATION_WILL_BE_DEPRECATED("BCE20406", "strand.annotation.will.be.deprecated"),
+
+    BUILT_WITH_OLDER_SL_UPDATE_DISTRIBUTION("BCE20407", "built.wtih.older.sl.update.distribution")
```
azinneera, line 74: We maintain `ProjectDiagnosticErrorCode` class for Project API diagnostics. Lets move the diagnostic there.  All resolution diagnostics begin with `BCE50`.

https://github.com/ballerina-platform/ballerina-lang/blob/b43096b16218cd6c7f0158eb4582089690ddb09c/compiler/ballerina-lang/src/main/java/io/ballerina/projects/internal/ProjectDiagnosticErrorCode.java#L30-L37

The DiagnosticId for this warning will be `BCE5008`

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -485,14 +486,41 @@ private BlendedManifest createBlendedManifest(PackageContext rootPackageContext,
...
+                || isMinorOrMajorVersionIncrement(prevDistributionVersion, currentDistributionVersion))) {
+            packageLockingMode = PackageLockingMode.SOFT;
+            DiagnosticInfo diagnosticInfo = new DiagnosticInfo(
+                    DiagnosticWarningCode.BUILT_WITH_OLDER_SL_UPDATE_DISTRIBUTION.diagnosticId(),
+                    "WARNING: Detected packages built with an older SL Update distribution. " +
+                            "Dependencies will be updated to the latest versions compatible " +
+                            "with the current distribution.",
+                    DiagnosticWarningCode.BUILT_WITH_OLDER_SL_UPDATE_DISTRIBUTION.severity());
```
azinneera, line 501:
```suggestion
                    DiagnosticSeverity.WARNING);
```

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/GraphCommandTest.java
```diff
@@ -349,4 +366,27 @@ private void compileAndCacheTestDependencies() {
     private String readFormattedOutput() throws IOException {
         return readOutput(true).replaceAll("\n\t\n", "\n\n").replaceAll("\r", "").strip();
     }
+
+    private void replaceDependenciesTomlVersion(Path projectPath) throws IOException {
+        String currentDistrVersion = RepoUtils.getBallerinaShortVersion();
+        Path dependenciesTomlTemplatePath = projectPath.resolve("Dependencies-template.toml");
+        Path dependenciesTomlPath = projectPath.resolve("Dependencies.toml");
+
+        FileInputStream input = new FileInputStream(dependenciesTomlTemplatePath.toString());
```
azinneera, line 375: Let's use try-with-resources

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -485,14 +486,41 @@ private BlendedManifest createBlendedManifest(PackageContext rootPackageContext,
...
+        boolean sticky = getSticky(rootPackageContext);
+        PackageLockingMode packageLockingMode;
+        SemanticVersion prevDistributionVersion = rootPackageContext.dependencyManifest().distributionVersion();
+        SemanticVersion currentDistributionVersion = SemanticVersion.from(RepoUtils.getBallerinaShortVersion());
+        if (!sticky && (null == prevDistributionVersion
+                || isMinorOrMajorVersionIncrement(prevDistributionVersion, currentDistributionVersion))) {
+            packageLockingMode = PackageLockingMode.SOFT;
+            DiagnosticInfo diagnosticInfo = new DiagnosticInfo(
+                    DiagnosticWarningCode.BUILT_WITH_OLDER_SL_UPDATE_DISTRIBUTION.diagnosticId(),
+                    "WARNING: Detected packages built with an older SL Update distribution. " +
```
azinneera, line 498:
```suggestion
                    "Detected packages built with an older SL Update distribution. " +
```
`WARNING` will be printed by the `Diagnostic.toString()` method

azinneera: Let's improve this logic to print the warning only if the Dependencies.toml exists. i.e We don't need to warn if it is a fresh project.

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -485,14 +486,41 @@ private BlendedManifest createBlendedManifest(PackageContext rootPackageContext,
...
+                            "Dependencies will be updated to the latest versions compatible " +
+                            "with the current distribution.",
```
azinneera, line 500: @sameerajayasoma can this warning be improved?

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
sameerajayasoma: +1

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -485,14 +486,41 @@ private BlendedManifest createBlendedManifest(PackageContext rootPackageContext,
...
                 .setDumpRawGraphs(compilationOptions.dumpRawGraphs())
+                .setPackageLockingMode(packageLockingMode)
                 .build();
     }
 
+    private boolean isMinorOrMajorVersionIncrement(SemanticVersion prevDistributionVersion,
+                                                  SemanticVersion currentDistributionVersion) {
+        return currentDistributionVersion.greaterThan(prevDistributionVersion)
+                && (currentDistributionVersion.minor() != prevDistributionVersion.minor()
+                || currentDistributionVersion.major() != prevDistributionVersion.major());
```
azinneera, line 521: Major will always be equal for SL. So the condition would be:
```java
return (currentDistributionVersion.major() == prevDistributionVersion.major())
          && (currentDistributionVersion.minor() > prevDistributionVersion.minor())
```

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -485,14 +486,41 @@ private BlendedManifest createBlendedManifest(PackageContext rootPackageContext,
...
                 .setOffline(compilationOptions.offlineBuild())
-                .setSticky(getSticky(rootPackageContext))
+                .setSticky(sticky)
                 .setDumpGraph(compilationOptions.dumpGraph())
                 .setDumpRawGraphs(compilationOptions.dumpRawGraphs())
+                .setPackageLockingMode(packageLockingMode)
                 .build();
     }
 
+    private boolean isMinorOrMajorVersionIncrement(SemanticVersion prevDistributionVersion,
```
azinneera, line 517:
```suggestion
    private boolean isNewUpdateDistribution(SemanticVersion prevDistributionVersion,
```

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
sameerajayasoma: Detected an attempt to build this package using Swn Lake Update 5. However, this package was built using Swan Lake Update 4.1. To ensure compatibility, the `Dependencies.toml` file will be updated with the latest versions that are compatible with Update 5.

gayaldassanayake: So the warning message would be in the format,
> Detected an attempt to build this package using Swn Lake Update `<minor>.<patch>`. However, this package was built using Swan Lake Update `<minor>.<patch>`. To ensure compatibility, the Dependencies.toml file will be updated with the latest versions that are compatible with Update `<minor>.<patch>`.

Also in cases where Dependencies.toml doesn't contain `distribution-version` (update 4 or below), @azinneera suggested to use `...was built using Swan Lake Update 4 or below`. 

Hope it is fine.

azinneera: Let reword `below` to `an older Update`

compiler/ballerina-lang/src/main/java/io/ballerina/projects/PackageResolution.java
```diff
@@ -485,14 +486,56 @@ private BlendedManifest createBlendedManifest(PackageContext rootPackageContext,
...
+        String currentVersionForDiagnostic =
+                currentDistributionVersion.minor() + DOT + currentDistributionVersion.patch();
+        if (null != prevDistributionVersion) {
+            prevVersionForDiagnostic = prevDistributionVersion.minor()  + DOT + prevDistributionVersion.patch();
+        } else {
+            prevVersionForDiagnostic = "4 or an older Update";
+        }
+        DiagnosticInfo diagnosticInfo = new DiagnosticInfo(
+                ProjectDiagnosticErrorCode.BUILT_WITH_OLDER_SL_UPDATE_DISTRIBUTION.diagnosticId(),
+                "Detected an attempt to build this package using Swn Lake Update " + currentVersionForDiagnostic +
```
azinneera, line 529:
```suggestion
                "Detected an attempt to build this package using Swan Lake Update " + currentVersionForDiagnostic +
```

azinneera: Let's  append the patch version for only the patch distributions. `Swan Lake Update 5.0` looks a bit odd.

azinneera (approved):

PR #39934, opened 2023-03-21
Fix `tsymbol` getting `null` in first param in `function:call` lang-lib

### LINKED ISSUE #38932: [Bug]: Suggestions don't work for arguments in `function:call`
### Description

$title.

### Steps to Reproduce

```ballerina
public function main() {
    any|error johnName = function:call(ge<cursor>);
}

function getFullName(string firstName, string? lastName = ()) returns string {
    if lastName is () {
        return firstName;
    }

    return string `${firstName} ${lastName}`;
}
```

`getFullName` is not suggested.

```
	at io.ballerina.compiler.api.impl.symbols.BallerinaFunctionTypeSymbol.returnTypeDescriptor(BallerinaFunctionTypeSymbol.java:141)
	at org.ballerinalang.langserver.completions.providers.AbstractCompletionProvider.getAnonFunctionDefSnippet(AbstractCompletionProvider.java:770)
	at org.ballerinalang.langserver.completions.providers.AbstractCompletionProvider.expressionCompletions(AbstractCompletionProvider.java:582)
	at org.ballerinalang.langserver.completions.providers.context.FunctionCallExpressionNodeContext.getCompletions(FunctionCallExpressionNodeContext.java:56)
	at org.ballerinalang.langserver.completions.providers.context.FunctionCallExpressionNodeContext.getCompletions(FunctionCallExpressionNodeContext.java:40)
	at org.ballerinalang.langserver.completions.util.CompletionUtil.route(CompletionUtil.java:125)
	at org.ballerinalang.langserver.completions.util.CompletionUtil.getCompletionItems(CompletionUtil.java:81)
	at org.ballerinalang.langserver.completions.BallerinaCompletionExtension.execute(BallerinaCompletionExtension.java:65)
	at org.ballerinalang.langserver.completions.BallerinaCompletionExtension.execute(BallerinaCompletionExtension.java:40)
	at org.ballerinalang.langserver.LangExtensionDelegator.completion(LangExtensionDelegator.java:124)
	at org.ballerinalang.langserver.BallerinaTextDocumentService.lambda$completion$0(BallerinaTextDocumentService.java:149)
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:642)
	at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:479)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
```

### Affected Version(s)

2201.3.0-RC

_No response_

### Related area

-> Editor

_No response_

_No response_

_No response_

dulajdilshan:
## Purpose
$title

Fixes #38932

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
@@ -1420,8 +1420,13 @@ public class BIRPackageSymbolEnter {
                     return bMapType;
                 case TypeTags.INVOKABLE:
                     BInvokableType bInvokableType = new BInvokableType(null, null, null, null);
+                    bInvokableType.tsymbol = Symbols.createInvokableTypeSymbol(SymTag.FUNCTION_TYPE, flags,
+                            env.pkgSymbol.pkgID, bInvokableType,
+                            env.pkgSymbol.owner, symTable.builtinPos,
+                            COMPILED_SOURCE);
                     bInvokableType.flags = flags;
                     if (inputStream.readBoolean()) {
+                        // Return if an any function
                         return bInvokableType;
                     }
                     int paramCount = inputStream.readInt();
--- a/tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/expectedtype/FunctionCallExpressionTest.java
+++ b/tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/expectedtype/FunctionCallExpressionTest.java
@@ -85,6 +85,7 @@ public class FunctionCallExpressionTest {
                 {123, 20, TypeDescKind.INT},
                 {124, 22, TypeDescKind.INT},
                 {136, 32, TypeDescKind.STRING},
+                {144, 39, TypeDescKind.FUNCTION}
         };
     }
 }
--- a/tests/ballerina-compiler-api-test/src/test/resources/test-src/expected-type/function_call_expression_test.bal
+++ b/tests/ballerina-compiler-api-test/src/test/resources/test-src/expected-type/function_call_expression_test.bal
@@ -140,3 +140,7 @@ public function main() {
 function myFunction(string arg1, string arg2, int arg3 = 101) {
 
 }
+
+public function testLangLibFunctionCall() {
+    any|error johnName = function:call();
+}


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -1420,8 +1420,13 @@ public BType readType(int cpI) throws IOException {
                     return bMapType;
                 case TypeTags.INVOKABLE:
                     BInvokableType bInvokableType = new BInvokableType(null, null, null, null);
+                    bInvokableType.tsymbol = Symbols.createInvokableTypeSymbol(SymTag.FUNCTION_TYPE, flags,
```
chiranSachintha, line 1423: In the `setTSymbolForInvokableType`(L1437) method, no need to create a new type symbol again. We can remove that.

dulajdilshan: Fixed 8795ec02ca7e346039059139ce2b9f8364ba7c29

MaryamZi: We now pass a type to `createInvokableTypeSymbol` though? May be OK, but it could also have side effects. For now, shall we leave as is and move the symbol creation here to within the check for any function?

MaryamZi:
Shall we also update the PR title to reflect what's actually being fixed/updated?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
dulajdilshan: What do you mean by side effects? Isn't it redundant to have symbol creation in two places?

chiranSachintha: Yeah, I understand your question. However, even if we move within the block, there may still be issues if there are places that generate a `function type`(function type means any function type) without a type symbol(As we generate a type symbol for every `function type` from the BIR). In order to fix this properly, shouldn't we move the flag that checks for the presence of a type symbol to the beginning?(What I mean is that we only need to create a type symbol from the BIR if it is present when writing to the BIR)

MaryamZi: > However, even if we move within the block, there may still be issues if there are places that generate a function type(function type means any function type) without a type symbol(As we generate a type symbol for every function type from the BIR).

What kind of issue? Shouldn't every `function` type have a symbol?

> In order to fix this properly, shouldn't we move the flag that checks for the presence of a type symbol to the beginning?(What I mean is that we only need to create a type symbol from the BIR if it is present when writing to the BIR) 

Yes, but we can't do this atm because it is not how `function` was initially handled, right? IMO, it's OK to create a symbol for `function` like this since the attributes are fixed.

chiranSachintha: Every type should have a type symbol. 
After changing the logic as you mentioned, when reading function types from BIR, every `any function` type has a type symbol. However, in the normal compiling flow, there can be `any function` types without type symbols. my concern is whether this can cause any issues with the current compilation logic.

MaryamZi: What's the scenario where a `function` may not have a type symbol? How would adding the symbol here cause an issue?

dulajdilshan: > For example, there may be places that check if tsymbol.type is null and do something based on that. Previously it was null, but now you're setting the type.
> 
Previously it was null only for `any function` types. For other function types, we create the `tsymbol` using the invokable-type

> I don't think it's "redundant" if you move it to within the if block after checking for `function`.
I mean if we do so, we will be creating a tsymbol in `setTSymbolForInvokableType(bInvokableType, bInvokableType.retType);` and witihin the if-block, so wouldn't it be good if we have this above the if-block?

MaryamZi: > > For example, there may be places that check if tsymbol.type is null and do something based on that. Previously it was null, but now you're setting the type.
> 
> Previously it was null only for `any function` types. For other function types, we create the `tsymbol` using the invokable-type

Is this correct? `null` is passed for `type` in https://github.com/ballerina-platform/ballerina-lang/pull/39934/files#diff-75a4933bf7e446d096c3a9ce1c8e8abb41cd5497adba2c1a9c84093240169519L1157?

> 
> > I don't think it's "redundant" if you move it to within the if block after checking for `function`.
> 
> I mean if we do so, we will be creating a tsymbol in `setTSymbolForInvokableType(bInvokableType, bInvokableType.retType);` and witihin the if-block, so wouldn't it be good if we have this above the if-block?

It's only created once because we return within the if block? It's either within the if block (for `function`) or in `setTSymbolForInvokableType` for other functions?

dulajdilshan: Got it. Previously we haven't passed the type when creating the tsymbol.

dulajdilshan: I think passing `null` wouldn't break any flows. 73c243508dd413bbb93623cfed4ffa886c017dc3

MaryamZi (approved):

MaryamZi:
> Shall we also update the PR title to reflect what's actually being fixed/updated?

I would mention the actual change (e.g., Set the symbol for `function` in BIR symbol enter) and include "Fix NPE when getting completions in arguments in function:call()" in the description.

chiranSachintha (approved):

PR #39942, opened 2023-03-21
Support list constructor in group-by with sequence variables

KavinduZoysa:
## Purpose
$subject


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
```diff
@@ -1934,7 +1937,31 @@ public void visit(BLangLetExpression letExpr) {
 
     @Override
     public void visit(BLangListConstructorExpr listConstructorExpr) {
-        listConstructorExpr.exprs.forEach(this::acceptNode);
+        List<BLangExpression> expressions = listConstructorExpr.exprs;
```
ushirask, line 1940: Do we need to create a new variable here?

KavinduZoysa: Since it is used in 2 place, I think it is better to create a new variable

PR #39971, opened 2023-03-23
Fix compiler crashes for expressions in const annotations 

### LINKED ISSUE #38746: [Bug]: Compiler crashes for invalid non-const non-var ref expressions in const annotations
### Description

$title.

### Steps to Reproduce

```ballerina
public const annotation record {| int increment = 1; |} v25 on source type;

int x = 1;

@v25 {
    increment: x + 1
}
type Foo record {|
    int x;
|};
```

Happens with other kinds of expressions too. E.g., unary expression with a non-const expression (`-x`)

### Affected Version(s)

2201.3.0-RCs

_No response_

### Related area

-> Compilation

_No response_

_No response_

_No response_

ushirask:
Fixes #38746


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
```diff
@@ -234,8 +237,10 @@ public void visit(BLangRecordLiteral recordLiteral) {
                 } else {
                     continue;
                 }
-
+                BType tempCurrentConstSymbolType = this.currentConstSymbolType;
+                this.currentConstSymbolType = keyValuePair.getBType();
                 value = constructBLangConstantValue(keyValuePair.valueExpr);
+                this.currentConstSymbolType = tempCurrentConstSymbolType;
```
lochana-chathura, line 243: - We should deal with the global variables very carefully and avoid as much as possible in the compiler as one wrong assignment could mess up the whole compiler.

- Keeping two separate global variables for `constant.symbol` and `constant.symbol.type` anyway is not a good practice, as updating one place wouldn't update the other place in parallel. (i.e. updating `currentConstSymbolType` would not update `currentConstSymbol`'s type)

- Here, temporarily setting the value `this.currentConstSymbolType = keyValuePair.getBType();` is hacky and would be specific to this scenario only wouldn't solve this for non-record scenarios. 

As I understand the root cause of this issue is as follows.

At the `ConstantValueResolver` we expect to invoke the `void visit(BLangConstant constant) {}` visitor first, where we initialize the global variables. However, in this scenario, we invoke the `ConstantValueResolver` via the `constructBLangConstantValueWithExactType()` which bypass the said visit. 

The correct fix is, we need to initialize the `currentConstSymbol` at the beginning of `constructBLangConstantValueWithExactType()`. i.e. Need to add `this.currentConstSymbol = constantSymbol;` line.

lochana-chathura (changes requested):

compiler/ballerina-lang/src/main/java/org/ballerinalang/util/diagnostic/DiagnosticErrorCode.java
```diff
@@ -804,10 +804,10 @@ public enum DiagnosticErrorCode implements DiagnosticCode {
     SEQUENCE_VARIABLE_CAN_BE_USED_IN_SINGLE_ELEMENT_LIST_CTR_OR_FUNC_INVOCATION(
             "BCS4047", "seq.var.used.in.single.element.list.ctr.or.func.invocation"),
     SEQ_ARG_FOLLOWED_BY_ANOTHER_SEQ_ARG("BCS4048", "seq.arg.followed.by.another.seq.arg"),
-
     QUERY_CONSTRUCT_TYPES_CANNOT_BE_USED_WITH_COLLECT("BCS4049", "query.construct.types.cannot.be.used.with.collect"),
     VARIABLE_IS_SEQUENCED_MORE_THAN_ONCE("BCS4050", "variable.is.sequenced.more.than.once"),
-    INVALID_GROUPING_KEY_TYPE("BCS4051", "invalid.grouping.key.type")
+    INVALID_GROUPING_KEY_TYPE("BCS4051", "invalid.grouping.key.type"),
+    ANNOTATION_ELEMENT_SHOULD_BE_LITERAL("BCS4052", "annotation.element.should.be.a.literal")
```
lochana-chathura, line 810:
```suggestion
    ANNOTATION_ELEMENT_SHOULD_BE_LITERAL("BCE4052", "annotation.element.should.be.a.literal")
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SemanticAnalyzer.java
```diff
@@ -4416,7 +4416,7 @@ private void checkAnnotConstantExpression(BLangExpression expression) {
                 ((BLangListConstructorExpr) expression).exprs.forEach(this::checkAnnotConstantExpression);
                 break;
             default:
-                dlog.error(expression.pos, DiagnosticErrorCode.EXPRESSION_IS_NOT_A_CONSTANT_EXPRESSION);
+                dlog.error(expression.pos, DiagnosticErrorCode.ANNOTATION_ELEMENT_SHOULD_BE_LITERAL);
```
lochana-chathura, line 4419: Please refer to https://github.com/ballerina-platform/ballerina-lang/issues/38746#issuecomment-1594437887

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
```diff
@@ -814,7 +815,7 @@ private void addAssociatedTypeDefinition(BLangConstant constant, BIntersectionTy
 
     private BType createRecordType(BLangExpression expr, BConstantSymbol constantSymbol, Object value, Location pos,
                                    SymbolEnv env) {
-        if (expr.getKind() == NodeKind.SIMPLE_VARIABLE_REF) {
+        if (expr.getKind() == NodeKind.SIMPLE_VARIABLE_REF || expr.getKind() == NodeKind.UNARY_EXPR) {
```
lochana-chathura, line 818: This seems not to be the root cause for the unary case.

For `{increment: 1}` case we come to `createRecordType()` only once. However, for `{increment: +1}` case we come here twice and the crashing takes place at the second time. 

The reason we come here twice is that, at `evaluateUnaryOperator()` here 
https://github.com/ballerina-platform/ballerina-lang/blob/4339d5dea5f45efc2f4c2cd64c24758e7c431349/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java#L353  

we set the value type pair as follows. 
- value - 1
- type - currentConstSymbol.type = record { increment = 1 }

The type of 1 should be `int`. not the record type. 

That's the same reason `{increment: -1}` case works without this fix and we only crash for +1. 

I think we need to change the type to `value.type` here.

ushirask: Updated as suggested.

lochana-chathura (approved):

lochana-chathura (approved):

PR #40021, opened 2023-03-29
Verify member variable is public before mocking

### LINKED ISSUE #30893: [Testerina] When mocking a client and using `getMember()`, the functions of the client do not work
**Description:**
When mocking a client and using `getMember()`, a case is registered in the mock registry only for the member variable and not the functions of the client. 

This causes a `case not registered` error when attempting to use getters or setters of the client that happen to use that particular member variable. 

```
// Sample client that we can use for member access
public client class exClient {
    public int id;
    private string[] values = ["A", "B", "C", "D"];

    public function init(int id) {
        self.id = id;
    }

    public function getValue(int id) returns string {
        return self.values[id];
    }
}
```

In this case, calling `getValue()` does not work and the object mocking functionality throws an error since the function is not stubbed. 

**Describe your problem(s)**

**Describe your solution(s)**

Dilhasha:
## Purpose
$subject
Fixes #30893

## Approach
check the access modifiers before allowing to stub member variables allowing to mock only public member variables

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/mock/ObjectMock.java
+++ b/misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/mock/ObjectMock.java
@@ -20,6 +20,7 @@ package org.ballerinalang.testerina.natives.mock;
 import io.ballerina.runtime.api.PredefinedTypes;
 import io.ballerina.runtime.api.TypeTags;
 import io.ballerina.runtime.api.creators.ErrorCreator;
+import io.ballerina.runtime.api.flags.SymbolFlags;
 import io.ballerina.runtime.api.types.Field;
 import io.ballerina.runtime.api.types.MethodType;
 import io.ballerina.runtime.api.types.ObjectType;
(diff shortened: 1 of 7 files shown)


misc/testerina/modules/testerina-core/src/main/java/org/ballerinalang/testerina/natives/mock/ObjectMock.java
```diff
@@ -297,6 +303,10 @@ public static BError thenReturn(BObject caseObj) {
         return null;
     }
 
+    private static boolean validateFieldAccess(ObjectType objectType, String fieldName) {
```
ushirask, line 306: Shall we use a name like `validateFieldAccessIsPublic` or `isFieldPublic` for readability?

Dilhasha: Updated with 5b9bd8a

Dilhasha:
This change was merged with the PR https://github.com/ballerina-platform/ballerina-lang/pull/40251

PR #40064, opened 2023-04-03
Set the FIELD flag to fields with default values

MaryamZi:
## Purpose
$title. This also allows using quoted identifiers for builtin types in such fields.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/40065

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


MaryamZi:
> Adding the `field` flag for default values is okay. when it comes to [this](https://github.com/ballerina-platform/ballerina-lang/issues/40065) issue of checking whether a field is a member, is it correct to use the `field` flag? (I think I introduced this logic while fixing an issue related to variable shadowing) With this issue, I think we should use env details instead of the flag. WDYT?

Can you elaborate on why you think an env-based approach is better?

chiranSachintha:
> Can you elaborate on why you think an env-based approach is better?

We have added this `field` flag for the fields of records, objects, and tuples. When we visit a variable, to identify whether it is a field or another variable we can use the scope(env) of that instance(as we generate new env for records, objects and tuples). Currently, we used `field` flag to make this identification.

The reason why I think this way is that if we encounter an error indicating that a symbol has been re-declared due to a compiler bug, it should only occur if the environments are not defined correctly or something related to that. IMO encountering this type of error due to a missing flag is not correct.

MaryamZi:
> > Can you elaborate on why you think an env-based approach is better?
> 
> We have added this `field` flag for the fields of records, objects, and tuples. When we visit a variable, to identify whether it is a field or another variable we can use the scope(env) of that instance(as we generate new env for records, objects and tuples). Currently, we used `field` flag to make this identification.
> 
> The reason why I think this way is that if we encounter an error indicating that a symbol has been re-declared due to a compiler bug, it should only occur if the environments are not defined correctly or something related to that. IMO encountering this type of error due to a missing flag is not correct.

But ideally it shouldn't be missing, right? Assuming the flag is set, that's the more straightforward check?

Issue #24246, opened 2020-06-17
Enable readonly object checks for annotation value


KRVPerera:
**Description:**
This issue depends on StandardLibs changes: https://github.com/ballerina-platform/ballerina-lang/issues/24217

**Steps to reproduce:**
Create an object with invalid `readonly` values and try to use it in an annotation value. The compiler does not complain.

**Affected Versions:**
all

**Related Issues (optional):**
#24217 #24245

**Suggested Labels (optional):**
Area/Language

MaryamZi:
```ballerina
annotation record {| Foo f; |} annot on type;

class Foo {
    int i = 0;
}

@annot {
    f: new ()
}
type Bar record {|
    
|};
```

MaryamZi:
I believe we can fix this now since the stdlib scenarios were fixed. This may be a breaking change though.

MaryamZi:
Although https://github.com/ballerina-platform/ballerina-standard-library/issues/74 was fixed, seems like there are other mutable fields in HTTP annotations now - https://github.com/ballerina-platform/ballerina-standard-library/issues/4300.

PR #40074, opened 2023-04-03
Disallow annotations with mutable objects

MaryamZi:
## Purpose
$title. This was previously allowed because the std lib used such annotations.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/24246

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


MaryamZi:
Although https://github.com/ballerina-platform/ballerina-standard-library/issues/74 was fixed, seems like there are other mutable fields in HTTP annotations now - https://github.com/ballerina-platform/ballerina-standard-library/issues/4300.

tests/jballerina-unit-test/src/test/resources/test-src/annotations/annots_with_invalid_type.bal
```diff
@@ -20,3 +20,13 @@ type VAL true;
 type trueArray VAL[];
 
 annotation trueArray v2 on type;
+
+annotation record {| Foo f; |} v3 on type;
```
lochana-chathura, line 24: I personally prefer the current way (We use this style in nBallerina too). However, our formatter would format like this though. 
```suggestion
annotation record {|Foo f;|} v3 on type;
```

lochana-chathura (approved):
Looks Good.

tests/jballerina-unit-test/src/test/resources/test-src/annotations/annots_with_invalid_type.bal
MaryamZi: Shall we leave as is (the string representation also uses spaces)? 

I think it helps readbility for inclusive/exclusiveness too

```
annotation record {lName name;} a on type;
// vs
annotation record {|lName name;|} b on type;
```

vs

```
annotation record { lName name; } a on type;
// vs
annotation record {| lName name; |} b on type;
```

I'm not sure why we don't have spaces here, may have taken a similar approach to mapping constructors, but they aren't comparable IMO.

MaryamZi:
Changing PR state to draft since stdlib needs to be fixed.

PR #40114, opened 2023-04-06
Fix incorrect behaviour of member access in xml with negative index

### LINKED ISSUE #35980: Incorrect behavior of member access on xml with key value < 0
**Description:**
$title. 
As per the spec when the evaluated value of the container expression is an `xml` and `and k is < 0, then the evaluation completes abruptly with a panic`
(i.e. key-expression is evaluated to get a value k)

But ATM, this result is an empty xml value.

**Steps to reproduce:**
```ballerina
function init() {
    xml v = xml `This is an xml`;
    io:println(v[-1].toBalString()); // prints xml``, but should panic
}
```

**Affected Versions:**
Ballerina 2201.0.2

SasinduDilshara:
Fixes #35980 

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
@@ -522,10 +522,12 @@ public final class XmlItem extends XmlValue implements BXmlItem {
      */
     @Override
     public XmlValue getItem(int index) {
-        if (index != 0) {
+        if (index > 0) {
             return new XmlSequence();
+        } else if (index < 0) {
+            throw ErrorCreator.createError(BallerinaErrorReasons.XML_OPERATION_ERROR,
+                    StringUtils.fromString("xml index out of range: index " + index));
         }
-
         return this;
     }
 
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
@@ -29,8 +29,10 @@ import io.ballerina.runtime.api.values.BXmlSequence;
 import io.ballerina.runtime.internal.BallerinaXmlSerializer;
 import io.ballerina.runtime.internal.XmlFactory;
 import io.ballerina.runtime.internal.XmlValidator;
+import io.ballerina.runtime.internal.util.exceptions.BLangExceptionHelper;
 import io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons;
 import io.ballerina.runtime.internal.util.exceptions.BallerinaException;
+import io.ballerina.runtime.internal.util.exceptions.RuntimeErrors;
 import org.apache.axiom.om.OMElement;
 import org.apache.axiom.om.OMException;
 import org.apache.axiom.om.OMNode;
@@ -522,13 +524,13 @@ public final class XmlItem extends XmlValue implements BXmlItem {
      */
     @Override
     public XmlValue getItem(int index) {
-        if (index > 0) {
+        if (index == 0) {
+            return this;
+        } else if (index > 0) {
             return new XmlSequence();
-        } else if (index < 0) {
-            throw ErrorCreator.createError(BallerinaErrorReasons.XML_OPERATION_ERROR,
-                    StringUtils.fromString("xml index out of range: index " + index));
         }
-        return this;
+        throw BLangExceptionHelper.getRuntimeException(
+                RuntimeErrors.XML_SEQUENCE_INDEX_OUT_OF_RANGE, 1, index);
     }
 
     public int size() {
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
@@ -526,7 +526,8 @@ public final class XmlItem extends XmlValue implements BXmlItem {
     public XmlValue getItem(int index) {
         if (index == 0) {
             return this;
-        } else if (index > 0) {
+        }
+        if (index > 0) {
             return new XmlSequence();
         }
         throw BLangExceptionHelper.getRuntimeException(
--- a/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlNonElementItem.java
+++ b/bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlNonElementItem.java
@@ -27,7 +27,9 @@ import io.ballerina.runtime.api.values.BString;
 import io.ballerina.runtime.api.values.BXml;
 import io.ballerina.runtime.api.values.BXmlNonElementItem;
 import io.ballerina.runtime.internal.BallerinaXmlSerializer;
+import io.ballerina.runtime.internal.util.exceptions.BLangExceptionHelper;
 import io.ballerina.runtime.internal.util.exceptions.BallerinaErrorReasons;
+import io.ballerina.runtime.internal.util.exceptions.RuntimeErrors;
 import org.apache.axiom.om.OMAbstractFactory;
 import org.apache.axiom.om.OMFactory;
 import org.apache.axiom.om.OMNode;
@@ -144,11 +146,11 @@ public abstract class XmlNonElementItem extends XmlValue implements BXmlNonEleme
     public XmlValue getItem(int index) {
         if (index == 0) {
             return this;
-        } else if (index < 0) {
-            throw ErrorCreator.createError(BallerinaErrorReasons.XML_OPERATION_ERROR,
-                    StringUtils.fromString("xml index out of range: index " + index));
+        } else if (index > 0) {
+            return new XmlSequence();
         }
-        return new XmlSequence();
+        throw BLangExceptionHelper.getRuntimeException(
+                RuntimeErrors.XML_SEQUENCE_INDEX_OUT_OF_RANGE, 1, index);
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlNonElementItem.java
```diff
@@ -141,6 +144,9 @@ public BXml descendants() {
     public XmlValue getItem(int index) {
         if (index == 0) {
```
ushirask, line 145: Shall we do the `index > 0` here similar to in `XmlItem` to keep the functions consistent?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
```diff
@@ -522,10 +522,12 @@ public Object copy(Map<Object, Object> refs) {
      */
     @Override
     public XmlValue getItem(int index) {
-        if (index != 0) {
+        if (index > 0) {
             return new XmlSequence();
+        } else if (index < 0) {
```
suleka96, line 527: Shall we put the error case as the first case to facilitated early fail.

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-indexed-access.bal
```diff
@@ -88,6 +88,83 @@ function testLengthOfXMLSequence() returns [int, int, int, int] {
     return [x1.length(), x2.length(), x2[2].length(), x3.length()];
 }
 
+const i = -1;
+
+function testInvalidXMLAccessWithNegativeIndex() returns error? {
```
suleka96, line 93: Shall we also add several  xm subtype as sequence tests. `eg: xml<xml:Comment>` and so on

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlNonElementItem.java
KavinduZoysa: It will be more readable to arrange it like this,
```
if (index = 0) {
    return ...;
}
if (index > 0) {
    return ...;
}
throw Error...;
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlNonElementItem.java
```diff
@@ -141,6 +144,9 @@ public BXml descendants() {
...
             return this;
+        } else if (index < 0) {
+            throw ErrorCreator.createError(BallerinaErrorReasons.XML_OPERATION_ERROR,
+                    StringUtils.fromString("xml index out of range: index " + index));
```
LakshanWeerasinghe, line 149: Move this error message to `MessageBundle.properties`.

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-indexed-access.bal
```diff
@@ -88,6 +88,83 @@ function testLengthOfXMLSequence() returns [int, int, int, int] {
...
+
+    xml x9 = xml `<names><name>Sherlock Holmes</name><name>Sherlock Holmes</name></names>`;
+    e = trap x9[1][-1];
+    check assertErrorMessage(e, "Index -1 out of bounds for length 0");
+    e = trap x9[1][a];
+    check assertErrorMessage(e, "Index -1 out of bounds for length 0");
+    e = trap x9[0][i];
+    check assertErrorMessage(e, "xml index out of range: index -1");
+}
+
```
pcnfernando, line 167: Add few tests with indexes larger than the length and asserting empty xml

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
```diff
@@ -522,11 +524,13 @@ public Object copy(Map<Object, Object> refs) {
      */
     @Override
     public XmlValue getItem(int index) {
-        if (index != 0) {
+        if (index == 0) {
+            return this;
+        } else if (index > 0) {
```
sanjana, line 529: Can we get rid off the `else` here like this?
```suggestion
        } 
        
        if (index > 0) {
```

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlNonElementItem.java
```diff
@@ -141,8 +146,11 @@ public BXml descendants() {
     public XmlValue getItem(int index) {
         if (index == 0) {
             return this;
+        } else if (index > 0) {
```
sanjana, line 149: Can get rid off the else.
```suggestion
        } 
        
        if (index > 0) {
```

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-indexed-access.bal
```diff
@@ -88,6 +88,362 @@ function testLengthOfXMLSequence() returns [int, int, int, int] {
...
+    assert((<error> e).message(), errorMessage3);
+    e = trap x14[a];
+    assertTrue(e is error);
+    assert((<error> e).message(), errorMessage3);
+    e = trap x14[i];
+    assertTrue(e is error);
+    assert((<error> e).message(), errorMessage3);
+
+    xml x15 = xml `<name>Sherlock Holmes</name><name>Sherlock Holmes</name>`;
+    e = trap x15[-1];
```
sanjana, line 265: Better add test cased for `-2`, `-3` as well.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
SasinduDilshara: I think it is better to use if and else if rather than using two ifs

SasinduDilshara: are you suggesting to remove `else if` and add `if`?.
I think using `if` and `else if` will be better rather than using to `if`s.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlNonElementItem.java
SasinduDilshara: https://github.com/ballerina-platform/ballerina-lang/pull/40114/files#r1191913089

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-indexed-access.bal
SasinduDilshara: the check is `<0`, because of that isn't it ok to add only one negative number.
wdyt?

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlItem.java
SasinduDilshara: updated. 5ff4ac52aa7ec017f6582c9daaf91208f3566c3c

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/values/XmlNonElementItem.java
SasinduDilshara: Updated. 5ff4ac52aa7ec017f6582c9daaf91208f3566c3c

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-indexed-access.bal
SasinduDilshara: Added

suleka96: You have added `xml:Comment` which is xml singelton type tests not `xml<xml:Comment>` which is xml  sub type sequence tests.

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-indexed-access.bal
```diff
@@ -88,6 +88,362 @@ function testLengthOfXMLSequence() returns [int, int, int, int] {
     return [x1.length(), x2.length(), x2[2].length(), x3.length()];
 }
 
+const i = -1;
+
+function testInvalidXMLAccessWithNegativeIndex() returns error? {
+    int a = -1;
```
suleka96, line 94: Can we add some xml sub type sequence tests as well. eg: `xml<xml:Comment>`

pcnfernando: I believe there is https://github.com/ballerina-platform/ballerina-lang/pull/40114/files/5ff4ac52aa7ec017f6582c9daaf91208f3566c3c#diff-d48872f73bc16a73f08f8f39ee2ac3554cfff305bb4fa861a222befe454e4147R209

pcnfernando:
@suleka96 Merging due to code freeze. @SasinduDilshara  Please check up on https://github.com/ballerina-platform/ballerina-lang/pull/40114#discussion_r1197678589

tests/jballerina-unit-test/src/test/resources/test-src/types/xml/xml-indexed-access.bal
SasinduDilshara: yes.

PR #40164, opened 2023-04-07
Add example to int:range function

gimantha:
$Subject


langlib/lang.int/src/main/ballerina/int.bal
```diff
@@ -224,6 +224,23 @@ public isolated function fromHexString(string s) returns int|error = @java:Metho
 # When `step > 0`, the members of S that are `< rangeEnd` are returned in increasing order.
 # When `step < 0`, the members of S that are `> rangeEnd` are returned in decreasing order.
 # When `step = 0`, the function panics.
+#
+# ```ballerina
+# int[] evenNumbers = [];
+# foreach int i in int:range(0, 8, 2) {
+#     evenNumbers.push(i);
+# }
+# evenNumbers ⇒ [0, 2, 4, 6]
```
MaryamZi, line 233:
```suggestion
# evenNumbers ⇒ [0,2,4,6]
```

langlib/lang.int/src/main/ballerina/int.bal
```diff
@@ -224,6 +224,23 @@ public isolated function fromHexString(string s) returns int|error = @java:Metho
...
+#
+# int[] oddNumbers = [];
+# foreach int i in int:range(7, 0, -2) {
+#     oddNumbers.push(i);
+# }
+# oddNumbers ⇒ [7, 5, 3, 1]
```
MaryamZi, line 239:
```suggestion
# oddNumbers ⇒ [7,5,3,1]
```

MaryamZi: Since we're using toString. Other examples should have been the same.

MaryamZi (dismissed):

PR #40224, opened 2023-04-17
Fix type narrowing when `if` returns

### LINKED ISSUE #39914: [Bug]: Nil Lifting Fails Inside While Loop With Two Checks
### Description

Consider the code in the `steps to reproduce` section. This reports the following compilation error:

```
ERROR [main.bal:(13:9,13:22)] compound assignment not allowed with nullable operands
ERROR [main.bal:(13:9,13:22)] operator '+' not defined for 'string' and 'string:Char?'
```

But it is already returning for the `()` value from the above if condition. This error disappears if we remove the `else if` block is removed:

```ballerina
if char is () {
    return error("Failed");
}
word += char;
i = i + 1;
```

Or if the last logic is surrounded with an `else` block:

```ballerina
if char is () {
    return error("Failed");
} else if char !is DOT {
    return error("Failed");
} else {
    word += char;
}
i = i + 1;
```

### Steps to Reproduce

Use the following code:

```ballerina
const DOT = ".";

    int i = 0;
    string word = "";
    while i < 3 {
        string:Char? char = check getChar();
        if char is () {
            return error("Failed");
        } else if char !is DOT {
            return error("Failed");
        }
        word += char;
        i = i + 1;
    }
}

public function getChar() returns string:Char|error? {
    return check string:fromCodePointInt(89);
}
```

### Affected Version(s)

2201.4.0

_No response_

### Related area

-> Compilation

_No response_

_No response_

_No response_

ushirask:
Fixes #39914

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/statements/ifelse/if-stmt.bal
```diff
@@ -217,6 +217,47 @@ function testResetTypeNarrowingWithBlockStmt() {
...
+        word += char;
+    }
+
+    assertEquality("...", word);
+}
+
+public function getChar() returns string:Char|error? {
+    return check string:fromCodePointInt(89);
+}
+
```
prakanth97, line 260: Shall we add a negative test like below
```
    string|int|float? char = "a";
    string word = "";
    if char is () {
        return error("char is null");
    } else if char is int {
        return error("char is int");
    } else if char == "a" {

    }

    word += char; 
}
```

To show that type of `char` is narrowed to `string|float` properly

prakanth97: Also add below positive case as well
```
    string|int|float? char = "a";
    string word = "";
    if char is () {
        return error("char is null");
    } else if char is int {
        return error("char is int");
    }  else if char is float {
        return error("char is float");
    } 

    word += char; 
}
```

hasithaa:
Differed till the next RnD cycle (2024).

PR #40286, opened 2023-04-24
Fix xml parser error messages

### LINKED ISSUE #40127: [Bug]: Error message for invalid XML string contains dependency information
### Description

```bal
import ballerina/io;

    xml x1 = check xml:fromString("<!-- comments cannot have -- in it -->");
    io:println(x1);
}
```

Gives `error: failed to parse xml: [com.ctc.wstx.exc.WstxLazyException] String '--' not allowed in comment (missing '>'?)
 at [row,col {unknown-source}]: [1,10]`.

Currenlty to for the test cases we use `"com.sun.xml.internal.stream.XMLInputFactoryImpl"` and in the production we use `woodstox`. That's why we didn't catch this earlier. In `1.2.38` onwords we have used `woodstox` for testing as well hence updated the test cases in https://github.com/ballerina-platform/ballerina-lang/pull/40016

Need to fix this in SL and need to backport this to 1.2.x if possible

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

-> Runtime

_No response_

_No response_

_No response_

Nadeeshan96:
$title
Fixes #40127

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/XmlFactory.java
```diff
@@ -80,8 +79,6 @@ public static BXml parse(String xmlStr) {
 
             XmlTreeBuilder treeBuilder = new XmlTreeBuilder(xmlStr);
             return treeBuilder.parse();
-        } catch (BError e) {
-            throw e;
         } catch (Throwable e) {
             throw ErrorCreator.createError(StringUtils.fromString(("failed to parse xml: " + e.getMessage())));
```
gabilang, line 83: Shall we print the stack trace if the error message is null?

Nadeeshan96: Addressed with d04c6a717e035d66e5b732057429e1e20d22be8b.

Nadeeshan96: Reverted the change as the lines were not covered.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/XmlTreeBuilder.java
```diff
@@ -96,8 +96,8 @@ public XmlTreeBuilder(Reader stringReader) {
     }
 
     private void handleXMLStreamException(Exception e) {
-        // todo: do e.getMessage contain all the information? verify
-        throw new BallerinaException(e.getMessage(), e);
+        String reason = e.getCause() == null ? e.getMessage() : e.getCause().getMessage();
```
hasithaa, line 99: I have seen, the message is also `null` with some exceptions, we need to handle that case as well.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/XmlFactory.java
```diff
@@ -80,8 +79,6 @@ public static BXml parse(String xmlStr) {
 
             XmlTreeBuilder treeBuilder = new XmlTreeBuilder(xmlStr);
             return treeBuilder.parse();
-        } catch (BError e) {
```
hasithaa, line 83: I think this logic is correct. If you look at XMLTreeBuilder's `handleXMLStreamException` it is returning a BError value. With the removal of this, we are again creating another BError from another BError details. The other case is handing any throwable. Separating these two concerns is good.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/XmlTreeBuilder.java
Nadeeshan96: Will address this.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/XmlFactory.java
Nadeeshan96: In the `XMLFactory`, we have two prefixes: "failed to create xml: " and "failed to parse xml: " used to start the `BError` error messages. It seems the correct thing to do is not remove the `BError` case here and also keep only a single prefix "failed to parse xml: " in the `XMLFactory`. It will be a breaking change because some error messages will be changed. Shall I do like that?

Or I can keep like this by throwing a BError at a lower level without the message prefix, and adding the prefix for both the `BError` and `Throwable` here when creating a new BError.

Nadeeshan96: Went with the first approach as said in the review comment. Fixed with 30ffe3cdcee90ee72db3e1ade17408e58ab31ec4.

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/XmlFactory.java
```diff
@@ -97,10 +108,12 @@ public static BXml parse(InputStream xmlStream) {
         try {
             XmlTreeBuilder treeBuilder = new XmlTreeBuilder(new InputStreamReader(xmlStream));
             return treeBuilder.parse();
+        } catch (BError e) {
+            throw e;
         } catch (DeferredParsingException e) {
             throw ErrorCreator.createError(StringUtils.fromString((e.getCause().getMessage())));
         } catch (Throwable e) {
-            throw ErrorCreator.createError(StringUtils.fromString(("failed to create xml: " + e.getMessage())));
+            throw ErrorCreator.createError(StringUtils.fromString((PARSE_ERROR_PREFIX + e.getMessage())));
```
warunalakshitha, line 116: Do we need '(' here

Nadeeshan96: Will remove this.

HindujaB (dismissed):
LGTM

bvm/ballerina-runtime/src/main/java/io/ballerina/runtime/internal/XmlFactory.java
Nadeeshan96: Fixed with eb82f105a413cf5fec654f770a2802c197c35fa2.

warunalakshitha (approved):
LGTM

PR #40310, opened 2023-04-27
Support collect clause

### LINKED ISSUE #36811: Support for `collect` clause
**Description:**

The detailed explanation can be found in spec issues https://github.com/ballerina-platform/ballerina-spec/issues/1137 and https://github.com/ballerina-platform/ballerina-spec/issues/1144. 
In a summary,

- The `collect` clause is an alternative to `select`.

- The `collect` clause is roughly equivalent to doing a `group by` that groups all frames into a single group.
- If the aggregation function is unqualified, it should be looked up in langlib based on the type of aggregated variable.
- In the following example, since the variable `i` is aggregated, it is treated as `...i`. Also `sum` is treated as `int:sum`.
```ballerina
int tot = from var i in [1, 2, 3, 4] collect sum(i);
```
- Also in the above example, let's assume argument of sum is `i + 5`. In this case, the entire argument should be evaluated repeatedly for each element of `...i`.

Let's go through a few examples available in spec issues and figure out how to implement this feature.

ex: 1
```ballerina
var total = from var { cost } in orders collect sum(cost);
```
The variable `cost` is treated as `...const` and `sum` is treated as `int:sum()`. Therefore in the implementation, we can directly do something which is equivalent to `int:sum(...cost)`. But for the below example, this approach will not work.

ex: 2
```ballerina
decimal wageBill =
  from var { salary, bonus } in persons
  collect sum(salary + bonus);
```
In this example, we have to first evaluate `salary + bonus` for each index. Therefore we have to do,
```ballerina
    decimal wageBill = 0;
    foreach int i in 0...salary.length() {
       wageBill = wageBill.sum(salary[i] + bonus[i]); 
    }
```
Now we can change the implementation explained in example 1, to coincide with this example.

ex: 3
```ballerina
string nameList =
  from var { firstName, lastName } in persons
  collect ", ".join(firstName + " " + lastName);
```
Let's assume `string[] firstName = ["fn1", "fn2", "fn3"];` and `string[] lastName = ["ln1", "ln2", "ln3"];`. The expected output of above operation is `"fn1 ln1,fn2 ln2,fn3 ln3"`. So in the implementation, we can do something like this,
```ballerina
    string sentence = firstName[0] + " " + lastName[0];
    foreach int i in 1...firstName.length() - 1 {
        sentence = string:'join(",", sentence, firstName[i] + " " + lastName[i]);
    }
```

In general, when implementing we can follow the following steps.
1. Define a variable that is used to keep the aggregated result and initialize it using the first index (as done in example 3)
2. Within the loop call the aggregation function and update the defined variable in first step
3. Return the result

KavinduZoysa:
## Purpose
Fixes #36811


compiler/ballerina-lang/src/main/java/org/ballerinalang/model/clauses/CollectClauseNode.java
```diff
@@ -0,0 +1,36 @@
+/*
+ *  Copyright (c) 2023, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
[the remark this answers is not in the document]
KavinduZoysa: Fixed

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/NodeFactory.java
```diff
@@ -2067,6 +2067,18 @@ public static SelectClauseNode createSelectClauseNode(
         return stSelectClauseNode.createUnlinkedFacade();
     }
 
+    public static CollectClauseNode createCollectClauseNode(
+            Token collectKeyword,
+            ExpressionNode expression) {
+        Objects.requireNonNull(collectKeyword, "selectKeyword must not be null");
```
gimantha, line 2073:
```suggestion
        Objects.requireNonNull(collectKeyword, "select Keyword must not be null");
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/desugar/QueryDesugar.java
```diff
@@ -239,9 +242,10 @@ public class QueryDesugar extends BLangNodeVisitor {
     private SymbolEnv queryEnv;
     private boolean containsCheckExpr;
     private boolean withinQuery = false;
-    private boolean afterGroupBy = false;
     private boolean withinLambdaOrArrowFunc = false;
+    private boolean withinCollectClause = false;
```
gimantha, line 246: Is it possible to move these variables to `AnalyzerData` object?

KavinduZoysa: Still QueryDesugar does not follow that pattern.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/NodeFactory.java
KavinduZoysa: Changed

gimantha (approved):

Issue #40278, opened 2023-04-21
[Bug]: Toml quoted keys are not parsed properly


sameerajayasoma:
### Description

For each toml quoted key, I can see additional double quotes around it. 
```
[{"basePath":"/api/entitlement-service","url":["https://abc.com/api/entitlement-service"],"config":{""foo.bar.baz"":"value"}}]
```
Config.toml
```toml
[[endpoints]]
basePath = "/api/entitlement-service"
url = ["https://abc.com/api/entitlement-service"]
config = {"foo.bar.baz" = "value"}
```

### Steps to Reproduce

Use the below code snippet
```ballerina
type Endpoint record {|
    string basePath;
    string[] url;
    map<string> config;
|};

configurable Endpoint[] endpoints = [];
```

Config.toml
```toml
[[endpoints]]
basePath = "/api/entitlement-service"
url = ["https://abc.com/api/entitlement-service"]
config = {"foo.bar.baz" = "value"}
```

### Affected Version(s)

2201.5.0

### Related area

-> Compilation

sameerajayasoma:
This will be a blocker pretty soon. We need this to implement service registration and discovery proposal.

xlight05:
Had a look at this today.
From toml parsers side, As I remember, adding the quotes into quoted keys was intentional. This is due to complexities when retrieving via dotted keys. 

Consider the following sample - 
```toml
"foo.bar" = "value"

[foo]
bar = "value1"
```

Both the entries are different right? First one is a key value par in root level, and 2nd one is inside a table. But when we want to retrieve via dotted keys. we always have to use "foo.bar" for both cases. I believe that's why we made surrounding quotes required when we wanna retrieve quoted strings. This is the same behavior in parsers like Toml4j as well.


However, in `Config.toml` case this doesn't really matter. Had a chat with @HindujaB today. We'll introduce an API without quotes for the Configurables to use in the AST level.

HindujaB:
I have a minor concern. 
Since we support identifiers with special characters in Ballerina and if we have a configurable variable with quotes as follows,
(This is not yet supported) 
```ballerina
configurable string \"foo\.bar\" = ?;
```
What should be the `Config.toml` representation?

```toml
"foo.bar" = "hello-foo.bar"
```
or
```toml
"\"foo.bar\"" = "hello-foo.bar"
```
[...]

PR #40335, opened 2023-05-02
Add API to get unquoted name for key value pairs

xlight05:
## Purpose
$Subject

Part 1 fix of #40278

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlKeyNode.java
+++ b/misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlKeyNode.java
@@ -55,6 +55,14 @@ public class TomlKeyNode extends TomlNode {
         return String.join(".", list);
     }
 
+    public String originalName() {
+        List<String> list = new ArrayList<>(keys.size());
+        for (TomlKeyEntryNode keyEntryNode : keys) {
+            list.add(keyEntryNode.name().toString());
+        }
+        return String.join(".", list);
+    }
+
     @Override
     public String toString() {
         return "TomlKeyNode{" + "keys=" + name() + '}';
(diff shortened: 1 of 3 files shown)


misc/toml-parser/src/main/java/io/ballerina/toml/semantic/ast/TomlKeyNode.java
```diff
@@ -55,6 +55,14 @@ public String name() {
         return String.join(".", list);
     }
 
+    public String originalName() {
```
hasithaa, line 58: This is asking for trouble, Given now we have two methods `name` and `orignalName`, consumers of the API will get confused what is the correct method to use. This is not a good API design.

xlight05: This is done to make the names a bit consistent with other APIs in ballerina[1]

We can change the method name to unquotedName() to fit the use case. Shall I do that then? 

[1]https://github.com/ballerina-platform/ballerina-lang/blob/c3fdd7f3c41c01c155b8eaac4ad0d5a5f7ceabf1/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/model/symbols/BSymbol.java#L98

hasithaa: Yes, when I see those APIs now it is a bit confusing. Why can't we just return the actual value, instead of the literal value? Most of cases those two are equal. unless we use a quoted identifier.  Do we want to use literal value anywhere?

Issue #33323, opened 2021-10-21
`lineRange` is not set when parsing multiple statements with syntax API


madushajg:
**Description:**
When I parse multiple statements using `parseStatements()`, The `lineRange` hasn't been set in each `StatementNode` except the first `StatementNode`.

eg:
```java
String x = "string fName = \"Ted\"; string lName = \"Mosby\";";
NodeList<StatementNode> s = NodeParser.parseStatements(x);
s.get(1).lineRange();
```
Above will throws the below IOOB exception.
```
java.lang.IndexOutOfBoundsException: Index: '36', Size: '19'
```

**Affected Versions:**
2.0.0-beta.4-SNAPSHOT

lochana-chathura:
The reason for IOOB is the `ExternalTreeNodeList` created for the statement list does not have a parent.

If we consider a similar sample,
```java
TextDocument textDocument = TextDocuments.from("import foo; import bar;");
ModulePartNode s = SyntaxTree.from(textDocument).rootNode();
LineRange lineRange = s.imports().get(1).lineRange(); // Here the parent of imports().get(1) is ModulePartNode
```

```java
String y = "int a = 1; a = 2;";
NodeList<StatementNode> s = NodeParser.parseStatements(y);
LineRange lineRange = s.get(1).lineRange(); // Here the parent of s.get(1) is null.
```

When the parent of a `StatementNode` is null, `LineMap` is populated for the `StatementNode`, not for the whole input. Therefore, when trying to derive the `LineRange` from `TextRange` it causes IOOB.

In summary, we need to have a parent node for the `NodeList<StatementNode>`. It cannot exist alone.

rdulmina:
+1 for the approach

lochana-chathura:
[1] https://github.com/ballerina-platform/ballerina-lang/blob/8adfc0ac5b12d60c70d33893a15e49e7eba2920e/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/NonTerminalNode.java#L200

[2] https://github.com/ballerina-platform/ballerina-lang/blob/8adfc0ac5b12d60c70d33893a15e49e7eba2920e/compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/syntax/ExternalTreeNodeList.java#L56

One thing that I do not understand is the reason for not making `ExternalTreeNodeList` a parent.
 
[...]

PR #40339, opened 2023-05-02
Introduce an API for regexp pattern validation in NodeParser

### LINKED ISSUE #40318: [Improvement]: Requesting public API for supporting to check the Regex Pattern validity before code generation
### Description

We have a requirement to validate the given regex pattern before giving the tooling support to ballerina code generations. Therefore it would be great if we can have a public API from NodeParser including the validity of the pattern and the details of the errors. 

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

-> Compilation

_No response_

_No response_

_No response_

sanjana:
Fixes #40318

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


DamithaSenevirathne:
Shall we add `NodeParser` related testcases to validate this ?

IMO, we can use approach similar to following tests

compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/tree/nodeparser/ParseActionOrExpressionTest.java

sanjana:
> Shall we add `NodeParser` related testcases to validate this ?
> 
> IMO, we can use approach similar to following tests
> 
> compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/tree/nodeparser/ParseActionOrExpressionTest.java

Added in https://github.com/ballerina-platform/ballerina-lang/pull/40339/commits/b0536ab7e9d3ac45afa06f0d9725f331124cce88.

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/internal/parser/BallerinaParser.java
```diff
@@ -190,6 +190,20 @@ public STNode parseAsExpression() {
         return expr;
     }
 
+    /**
+     * Completely parses a given input as regexp.
```
LakshanWeerasinghe, line 194:
```suggestion
     * Completely parses a given input as a regexp.
```

compiler/ballerina-parser/src/test/java/io/ballerinalang/compiler/parser/test/tree/nodeparser/ParseRegexpPatternTest.java
```diff
@@ -0,0 +1,114 @@
...
+import io.ballerina.compiler.syntax.tree.Token;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import java.util.List;
+
+/**
+ * Test {@code parseRegexpPattern} method.
+ *
+ * @since 5.0.0
```
LakshanWeerasinghe, line 34: incorrect version

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/NodeParser.java
```diff
@@ -170,4 +173,17 @@ public static LetVariableDeclarationNode parseLetVarDeclaration(String text, boo
...
+
+    /**
+     * Parses the input as a regexp pattern.
+     *
+     * @param text the input
+     * @return an {@code ExpressionNode}
+     */
+    public static ExpressionNode parseRegexpPattern(String text) {
+        String regexpStr = RE_KEYWORD.stringValue() + BACKTICK_TOKEN.stringValue() + text
+                            + BACKTICK_TOKEN.stringValue();
```
lochana-chathura, line 185: This injecting part is troublesome.  For example, let's say the text is `[A-Z]` and we get the position of `]`. The correct position should be `4`. Since we inject a text wrapper inside the positions will be affected. 

Simply, input `[A-Z]` => output: ``Re `[A-Z]` `` which I think is not a suitable behavior for a public API. 

We can achieve the same output using our existing API,
 ``parseExpression(String.Format("`Re `%s`", "[A-Z]" ))``. 

Maybe we can have the above as a utility method in open API if the usages are prevalent. 

@rdulmina your thoughts on this?

lochana-chathura (changes requested):

compiler/ballerina-parser/src/main/java/io/ballerina/compiler/syntax/tree/NodeParser.java
lochana-chathura: Additionally, we can't expose an API where input and output both being `[A-Z]` since we have defined the content within the backticks as a `NodeList`. (I have described the reason in #33323)
https://github.com/ballerina-platform/ballerina-lang/blob/4df9b3836b0020fab1e06136a28265152d30b4b3/compiler/ballerina-treegen/src/main/resources/syntax_tree_descriptor.json#L2143-L2147

rdulmina: Yes, I think the user will have to use `parseExpression(String.Format("`Re `%s`", "[A-Z]" ))` for this use-case without a separate API.

pcnfernando: @lnash94 Would this do-able from the OpenAPI tooling side?

rdulmina: Since we always have the `re` prefix it will always generate regular expression. If the regexp is invalid it will anyway create a valid node with diagnostic.

lochana-chathura: you can simply use, `node.hasDiagnostics()` to see if if the regex is syntactically valid.

pcnfernando:
This fix is now invalid since we have a workaround. @sanjana please add the vertict and close the PR

sanjana:
As per the conclusions of the discussions, it is decided to use `parseExpression()` to parse a regexp expression, and use `node.hasDiagnostics()` to check for diagnostics. 
Therefore, the need for this API specifically for parsing regexp expressions is unnecessary. 
Hence closing this PR.

Issue #36810, opened 2022-07-01
Support for `group by` clause


KavinduZoysa:
**Description:**
$subject. 

Under this issue we need to make sure basic `group by` clause works.

ex:
```ballerina
   Order[] orders = [{buyer: "b1", price: 12}, {buyer: "b2", price: 13}, {buyer: "b3", price: 14}, {buyer: "b3", price: 15}];
   string[] ordersDup = from var {buyer, price} in orders
         group by buyer 
         select buyer; // @output ["b1", "b2", "b3"]
```

```ballerina
   Order[] orders = [{buyer: "b1", price: 12}, {buyer: "b2", price: 13}, {buyer: "b3", price: 14}, {buyer: "b3", price: 15}];
   string[] ordersDup = from var {buyer, price} in orders
         group by buyer 
         select {buyer, sum(price)}; // @output ["b1", "b2", "b3"]
```

As mentioned in the spec, this clause should be executed as follows.

*   Iterate over each input frame f:
    *   partition the input frames into groups, where two frames are in the same group if they have the same (==) value for all grouping-key variables
    *   within a group, frames are in incoming order
    *   groups are ordered by  earliest member in incoming order
*   for each group
    *   emit a frame that has
        *   each grouping key variable bound to the value common to the group
        *   other variables become indexed

KavinduZoysa:
I would suggest the design of  `_GroupByFunction` as follows.

```ballerina
class _GroupByFunction {
    *_StreamFunction;

    _Iterator? itr = ();
    # + _inputFrame - The input frame which came through the pipeline
    # + _frames - This represents the set of output frames that is created after grouping.
    #             For each `_inputFrame`, `_frames` should be updated.
    #             After this is updated for each `_inputFrame`, an iterator of this is created and
    #             each value is passed via pipeline.
    function (_Frame _inputFrame, _Frame[] _frames) returns error? groupByFunc;

    function init(function (_Frame _frame, _Frame[] frames) returns error? groupByFunc) {
        self.groupByFunc = groupByFunc;
        self.prevFunc = ();
    }

    public function process() returns _Frame|error? {
        // 1. If `itr` is (), `itr` should be created.
        //      1.1 Initialize an array of `_Frame` (let's call `_frames`), This is the 2nd argument for `groupByFunc`. 
        //          Each `_Frame` element of `_frames` corresponds to a single group of group by operation. 
        //          The `_Frame` element cosists of, 
[...]

PR #40353, opened 2023-05-05
Add more group-by clause tests

KavinduZoysa:
## Purpose
Related to #36810


compiler/ballerina-lang/src/main/resources/compiler.properties
```diff
@@ -1975,4 +1975,7 @@ error.seq.var.cannot.be.used.in.required.arg=\
   sequence variable cannot be used in a required argument
 
 error.seq.var.used.in.single.element.list.ctr.or.func.invocation=\
-  sequence variable can be used in a single element list constructor or function invocation
\ No newline at end of file
+  sequence variable can be used in a single element list constructor or function invocation
```
gimantha, line 1978: Shall we change this error message? when we say "can be" it may convey the idea it is not mandatory

gimantha (approved):

PR #40364, opened 2023-05-08
Fix invalid unused warning

KavinduZoysa:
## Purpose

Related to #36810

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/DataflowAnalyzer.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/DataflowAnalyzer.java
@@ -32,6 +32,7 @@ import org.ballerinalang.util.diagnostic.DiagnosticErrorCode;
 import org.ballerinalang.util.diagnostic.DiagnosticWarningCode;
 import org.wso2.ballerinalang.compiler.diagnostic.BLangDiagnosticLog;
 import org.wso2.ballerinalang.compiler.semantics.analyzer.cyclefind.GlobalVariableRefAnalyzer;
+import org.wso2.ballerinalang.compiler.semantics.model.Scope;
 import org.wso2.ballerinalang.compiler.semantics.model.SymbolEnv;
 import org.wso2.ballerinalang.compiler.semantics.model.SymbolTable;
 import org.wso2.ballerinalang.compiler.semantics.model.symbols.BConstantSymbol;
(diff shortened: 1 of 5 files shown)


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/DataflowAnalyzer.java
```diff
@@ -1621,6 +1622,22 @@ public void visit(BLangOrderByClause orderByClause) {
     @Override
     public void visit(BLangGroupByClause groupByClause) {
         groupByClause.groupingKeyList.forEach(value -> analyzeNode(value, env));
+        for (Map.Entry<Name, Scope.ScopeEntry> symbolEntry : groupByClause.env.scope.entries.entrySet()) {
+            BSymbol sym = null;
+            Location loc = null;
+            for (Map.Entry<BSymbol, Location> unusedLocalVariable : this.unusedLocalVariables.entrySet()) {
```
ushirask, line 1628: Should we have this as the outer loop since `unusedLocalVariables` is more likely to be empty?

KavinduZoysa: The purpose of the whole logic is to modify the `unusedLocalVariables`, if we move it to outer loops, it is a bit hard to modify `unusedLocalVariables` without concurrent modification exception. If you have any suggestions to do so, please comment.

ushirask:
```
for (Map.Entry<BSymbol, Location> unusedLocalVariable : this.unusedLocalVariables.entrySet()) {
        BSymbol unusedVarSymbol = unusedLocalVariable.getKey();
        Location loc = unusedLocalVariable.getValue();
        for (Map.Entry<Name, Scope.ScopeEntry> symbolEntry : groupByClause.env.scope.entries.entrySet()) {
            if (unusedVarSymbol.name.equals(symbolEntry.getKey())) {
                this.unusedLocalVariables.remove(unusedVarSymbol);
                this.unusedLocalVariables.put(symbolEntry.getValue().symbol, loc);
                break;
            }
        }    
}
```
Will something like this work?

ushirask: Even if we're using the code as you've given we can still move the 
```
this.unusedLocalVariables.remove(sym);
this.unusedLocalVariables.put(symbolEntry.getValue().symbol, loc);
```
before the `break` since it's the only time when `sym` would be not `null`

gimantha (approved):

Issue #40380, opened 2023-05-10
[Bug]: Non-isolated function pointers can be called within an isolated function


pcnfernando:
### Description

```ballerina
public isolated function foo() {
    var f = bar;
    string _ = f(); //should throw a compilation error here
}

function bar() returns string {
    return "bar";
}
```

Ref: https://ballerina.io/spec/lang/master/#isolated_functions

### Related area

-> Compilation

MaryamZi:
https://github.com/ballerina-platform/ballerina-lang/blob/v2201.5.0/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/IsolationAnalyzer.java#L2075

Seems to be a [regression](https://github.com/ballerina-platform/ballerina-lang/pull/36395/files#diff-90dc57beeefe7f36f406f57edecf4a23024fd420b8698934638fb5bf4672322dR2057) from the changes for worker/start action isolated support.

PR #40410, opened 2023-05-12
Fix isolated analysis for function pointer calls

MaryamZi:
## Purpose
$title.

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/40380

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


tests/jballerina-unit-test/src/test/resources/test-src/isolation-analysis/isolation_analysis.bal
```diff
@@ -543,6 +543,54 @@ function testIsolatedFunctionWithSelfAsCapturedVariable() {
     assertEquality(false, b.compare());
 }
 
+isolated function testIsolatedFPCallInIsolatedFunction() {
```
pcnfernando, line 546: Something not directly related to the change
```ballerina
isolated function dummyTest() {
    var f = abc;
    // f(); // @error invalid invocation of a non-isolated function in an 'isolated' function
    print(<any>f is isolated function); //true
}

function abc() {
}
```

Is it a conscious decision to throw the error even though inferring would resolve it as isolated?

tests/jballerina-unit-test/src/test/resources/test-src/isolation-analysis/isolation_analysis_negative.bal
```diff
@@ -330,3 +330,59 @@ object {} testInvalidAccessOfSelfAsCapturedVariableInIsolatedFunction = object {
         self.words.filter(
             isolated function (string word) returns boolean => self.length == word.length()).length();
 };
+
```
pcnfernando, line 333: Shouldn't below be allowed
```ballerina
isolated function dummyTest() {
    var x = [2, 4, 6, 8].iterator().next;
    _ = x(); //@error invalid invocation of a non-isolated function in an 'isolated' function

    _ = [2, 4, 6, 8].iterator().next(); //allowed
}
```

tests/jballerina-unit-test/src/test/resources/test-src/isolation-analysis/isolation_analysis.bal
MaryamZi: Yes. 

https://ballerina.io/spec/lang/master/#section_8.3.3

> This inference is purely an optimization to improve service concurrency and is subject to two constraints. First, any inferences do not affect the static types visible to other modules. Second, **anything explicitly declared as isolated must satisfy the requirements of this specification without without relying on inference.**

tests/jballerina-unit-test/src/test/resources/test-src/isolation-analysis/isolation_analysis_negative.bal
MaryamZi: A bound method is `isolated` only if the object AND the method are `isolated`. So the type inferred for `x` doesn't have isolated.

https://ballerina.io/spec/lang/master/#field-access-expr

> If T is a subtype of the object basic type, then T must have a member with name field-name. The field-access-expr is evaluated by first evaluating the expression to get a value obj. If the member is a field, then the result of the expression if the value of that field of obj and the static type of the field-access-expr is the type of that field of T. If that member is a method, then the result of the expression is a new function value that when called will call the method with self bound to obj; **the isolated bit of the new function value is set if and only if the isolated bits of both obj and the method are set;** the type descriptor of the new function value will have the same annotations as the type descriptor of the method.

An isolated method of a non-isolated object can access mutable fields of the object in an unsafe way.

pcnfernando (approved):
LGTM

Issue #40494, opened 2023-05-24
[2201.6.0-stage and 2201.6.x] Full build pipeline failure with query expressions


tharindu-nw:
**Description:**
Fails at nballerina level: https://github.com/ballerina-platform/ballerina-release/actions/runs/5060826646/jobs/9089852522

**Steps to reproduce:**

**Affected Versions:**

**OS, DB, other environment details and versions:**

**Related Issues (optional):**

**Suggested Labels (optional):**

**Suggested Assignees (optional):**

MaryamZi:
Simplified version from one failure in nBallerina.

```ballerina
type ModuleDecls [string, FuncDecl...];

type FuncDecl [string, Signature];

public type Signature [string, string[], string];

type ModuleDeclsMemo record {|
    readonly string id;
    map<Signature> funcs = {};
|};

public function main() {
    record {|
        table<ModuleDeclsMemo> key (id) decls = table[];
    |} sc = {};

    ModuleDecls[] decl = from var { id, funcs } in sc.decls
                        select [id, ...from var [name, sig] in funcs.entries()
                                    select <FuncDecl>[name, sig]];
}
```

SasinduDilshara:
This is fixed now with https://github.com/ballerina-platform/ballerina-lang/pull/40497

PR #40496, opened 2023-05-24
Revert "[Master] Fix the incorrect diagnostic in query expression"

SasinduDilshara:
Reverts ballerina-platform/ballerina-lang#37547

issue: https://github.com/ballerina-platform/ballerina-lang/issues/40494


IMS94:
@SasinduDilshara when reverting a change, we should revert individual commits that have caused the issue. In this case, I can only see a single commit that's manual.

And it includes removing some LS test cases we have added to create variable code action on select clauses. Ideal approach should be to disable those test cases, and open an issue to re-enable (or fix the original diagnostic issue) the test cases.

SasinduDilshara:
> @SasinduDilshara when reverting a change, we should revert individual commits that have caused the issue. In this case, I can only see a single commit that's manual.
> 
> And it includes removing some LS test cases we have added to create variable code action on select clauses. Ideal approach should be to disable those test cases, and open an issue to re-enable (or fix the original diagnostic issue) the test cases.

The initial PR have the changes related to those Ls test cases.
In here the test cases are based on the query expected type.
Because of we removing these changes, I think we need to revert those tests and add the changes from another PR. wdyt?

IMS94 (approved):

pcnfernando (approved):

pcnfernando:
Reopened https://github.com/ballerina-platform/ballerina-lang/issues/37450 and https://github.com/ballerina-platform/ballerina-lang/issues/37670

Issue #40445, opened 2023-05-17
[Bug]: `io.ballerina.projects.test.PackageResolutionIntegrationTests.testCase0013` is failing 


Thevakumar-Luheerathan:
### Description

When moving from update 5 to update 6. A warning saying that,

> WARNING [package_y] Detected an attempt to compile this package using Swan Lake Update 6. However, this package was built using Swan Lake Update 5. To ensure compatibility, execute `bal build` with Swan Lake Update 6 to update the dependencies with the latest compatible versions.

is added to the diagnostics. It eventually fails the diagnostic count assertion.

### Related area

-> Other Area

PR #40522, opened 2023-05-26
Add BuildCommand tests for dependency version resolution change and fix PackageResolutionIntegrationTests#13

### LINKED ISSUE #39850: [Improvement]: Introduce distribution field to Dependencies.toml and bump dependencies if a higher distribution is used

gayaldassanayake:
## Purpose
$subject

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/40445
Related to #39850

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


cli/ballerina-cli/src/test/resources/test-resources/dep-dist-version-projects/expected-dep-tomls/new-pkg.toml
```diff
@@ -0,0 +1,37 @@
+# AUTO-GENERATED FILE. DO NOT MODIFY.
+
+# This file is auto-generated by Ballerina for managing dependency versions.
+# It should not be modified by hand.
+
+[ballerina]
+dependencies-toml-version = "2"
+distribution-version = "2201.6.0-SNAPSHOT"
```
azinneera, line 8: Hardcoded distribution version

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/BuildCommandTest.java
```diff
@@ -888,14 +895,228 @@ public void testBuildBalProjectWithCustomMavenRepo() throws IOException {
...
+        // depA:1.0.0 -> depA:1.1.0
+        // depB:1.0.0 -> depB:1.1.0
+        Path actualDependenciesToml = projectPath.resolve("Dependencies.toml");
+        Path expectedDependenciesToml = testResources.resolve("dep-dist-version-projects").resolve("expected-dep-tomls")
+                .resolve("old-dist-precomp-pkg-with-no-sticky.toml");
+        Assert.assertTrue(actualDependenciesToml.toFile().exists());
+        assertTomlFilesEquals(actualDependenciesToml, expectedDependenciesToml);
+
+        // Revert the distribution version and dependency versions to the placeholder values
+        replaceDependenciesTomlContent(projectPath, RepoUtils.getBallerinaShortVersion(),
```
azinneera, line 1000: We can improve this to copy the project files to a temporary location and replacing them. This will avoid accidental committing of modified files.

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/BuildCommandTest.java
```diff
@@ -888,14 +895,228 @@ public void testBuildBalProjectWithCustomMavenRepo() throws IOException {
...
+        replaceDependenciesTomlContent(projectPath, "**INSERT_DISTRIBUTION_VERSION_HERE**", "2201.5.0");
+        System.setProperty("user.dir", projectPath.toString());
+        BuildCommand buildCommand = new BuildCommand(projectPath, printStream, printStream, false);
+
+        new CommandLine(buildCommand).parseArgs();
+        buildCommand.execute();
+        String buildLog = readOutput(true);
+        Assert.assertEquals(
+                buildLog.replaceAll("\r", ""),
+                getOutput("build-old-dist-precomp-proj-with-dep.txt")
```
azinneera, line 986: The output seems to be wrong. For the combination where `Dependencies.toml` exists and --sticky flag is given, we don't update automatically. Instead, we give a hint asking the user to set sticky to false.

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/BuildCommandTest.java
```diff
@@ -888,14 +895,228 @@ public void testBuildBalProjectWithCustomMavenRepo() throws IOException {
...
+        BuildCommand buildCommand = new BuildCommand(projectPath, printStream, printStream, false);
+
+        new CommandLine(buildCommand).parseArgs("--sticky");
+        buildCommand.execute();
+        String buildLog = readOutput(true);
+        Assert.assertEquals(
+                buildLog.replaceAll("\r", ""),
+                getOutput("build-old-dist-precomp-proj-with-dep.txt")
+                        .replaceAll("INSERT_NEW_DIST_VERSION_HERE", getNewVersionForOldDistWarning())
+                        .replaceAll("INSERT_OLD_DIST_VERSION_HERE", getOldVersionForOldDistWarning("2201.5.0")));
```
azinneera, line 1021: The output seems to be wrong. For the combination where Dependencies.toml exists and --sticky flag is given, we don't update automatically. Instead, we give a hint asking the user to set sticky to false.

gayaldassanayake: Here the Dependencies.toml sticks to the dependencies correctly. But the warning is inaccurate. will update the logic to have a hint as you suggested instead of the warning.

cli/ballerina-cli/src/test/resources/test-resources/dep-dist-version-projects/expected-dep-tomls/new-pkg.toml
gayaldassanayake: Will fix

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/BuildCommandTest.java
gayaldassanayake: The hint was produced correctly when there no distribution-version field in the Dependencies toml. It was not produced when there is a distribution-version field but it is an older version than what the user currently has. This is fixed now.

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/BuildCommandTest.java
azinneera: Then why do we need to revert the values?

gayaldassanayake: That is because we are using the same project for all the tests that are in the group `proj-with-deps-update-policy`. So we are reverting the changes to the test-package copied to the temporary directory after each test belonging to that group. This makes it ready for the next test.

azinneera (approved):

cli/ballerina-cli/src/test/java/io/ballerina/cli/cmd/BuildCommandTest.java
azinneera: Let's improve this to use separate projects for each test to avoid failures in the cleanup. Please create an issue to track it.

azinneera (approved):

PR #40589, opened 2023-06-05
Add check to filter listener from classes

### LINKED ISSUE #40551: JetStreamListener is listed under classes
## Description
The JetStreamListener is listed under classes in https://lib.ballerina.io/ballerinax/nats/latest 
Ideally, this should be under `Listeners`
Screenshot 2023-05-30 at 10 25 28

NipunaMadhushan:
## Purpose
Filter Listener classes from classes with a better approach.

Fixes #40551

## Approach
We cannot use `listener` as a qualifier in a class definition. Therefore the previous method to filter listener classes was to check the name of the class whether it is equal to `Listener`. But it will even filter other classes with name `Listener` as listeners and wouldn't filter listeners with other names.

These changes will check the necessary methods which should be included in a class to be a listener. The following methods should be included in a class to be a listener.
1. `attach`
2. `detach`
3. `'start`
4. `immediateStop`
5. `gracefulStop`

This approach will check for these methods to identify whether it is a listener or not.

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/misc/docerina/src/main/java/org/ballerinalang/docgen/Generator.java
+++ b/misc/docerina/src/main/java/org/ballerinalang/docgen/Generator.java
@@ -485,14 +485,14 @@ public class Generator {
         if (containsToken(classDefinitionNode.classTypeQualifiers(), SyntaxKind.CLIENT_KEYWORD)) {
             return new Client(name, description, isDeprecated, fields, functions, isReadOnly, isIsolated, isService);
         } else if (containsToken(classDefinitionNode.classTypeQualifiers(), SyntaxKind.LISTENER_KEYWORD)
-                || checkListenerModel(functions)) {
+                || isListenerModel(functions)) {
             return new Listener(name, description, isDeprecated, fields, functions, isReadOnly, isIsolated, isService);
         } else {
             return new BClass(name, description, isDeprecated, fields, functions, isReadOnly, isIsolated, isService);
         }
     }
 
-    private static boolean checkListenerModel(List<Function> classFunctions) {
+    private static boolean isListenerModel(List<Function> classFunctions) {
         boolean isStartIncluded = false;
         boolean isAttachIncluded = false;
         boolean isDetachIncluded = false;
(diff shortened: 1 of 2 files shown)


misc/docerina/src/main/java/org/ballerinalang/docgen/Generator.java
```diff
@@ -485,13 +485,38 @@ private static BClass getClassModel(ClassDefinitionNode classDefinitionNode, Sem
...
 
+    private static boolean isListenerModel(List<Function> classFunctions) {
+        boolean isStartIncluded = false;
+        boolean isAttachIncluded = false;
+        boolean isDetachIncluded = false;
+        boolean isGracefulStopIncluded = false;
+        boolean isImmediateStopIncluded = false;
+
+        for (Function function : classFunctions) {
+            if (function.name.equals("'start")) {
```
keizer619, line 503: @NipunaMadhushan Shall we move these into constants?

misc/docerina/src/main/java/org/ballerinalang/docgen/Generator.java
```diff
@@ -485,13 +485,38 @@ private static BClass getClassModel(ClassDefinitionNode classDefinitionNode, Sem
...
+                isStartIncluded = true;
```
keizer619, line 504: @NipunaMadhushan We can return true here and just keep `if` conditions and can avoid line 516

misc/docerina/src/main/java/org/ballerinalang/docgen/Generator.java
```diff
@@ -485,13 +485,38 @@ private static BClass getClassModel(ClassDefinitionNode classDefinitionNode, Sem
...
+            } else if (function.name.equals("attach")) {
+                isAttachIncluded = true;
+            } else if (function.name.equals("detach")) {
+                isDetachIncluded = true;
+            } else if (function.name.equals("gracefulStop")) {
+                isGracefulStopIncluded = true;
+            } else if (function.name.equals("immediateStop")) {
```
keizer619, line 511: @NipunaMadhushan Also just wondering is it enough to check the name or do we need to validate the signature too

sm1990 (dismissed):

anuruddhal (approved):

misc/docerina/src/main/java/org/ballerinalang/docgen/Generator.java
NipunaMadhushan: @keizer619 sure. Addressed the suggestions.

keizer619 (approved):

PR #40644, opened 2023-06-08
Make node clones once

### LINKED ISSUE #40658: [Bug]: OOM in binary expr with large number of operations
### Description

```ballerina
public function main() {

    decimal d1 = 0;
    decimal d2 = 0;
    decimal d3 = 0;
    decimal d4 = 0;
    decimal d5 = 0;
    decimal d6 = 0;
    decimal d7 = 0;
    decimal d8 = 0;
    decimal d9 = 0;
    decimal d10 = 0;
    decimal d11 = 0;
    decimal d12 = 0;
    decimal d13 = 0;
    decimal d14 = 0;
    decimal d15 = 0;
    decimal d16 = 0;
    decimal d17 = 0;
    decimal d18 = 0;
    decimal d19 = 0;
    decimal d20 = 0;
    decimal d21 = 0;
    decimal d22 = 0;
    decimal d23 = 0;
    decimal d24 = 0;

    decimal totalOtherCost = + d1 
                             + d2  
                             + d3 
                             + d4  
                             + d5 
                             + d6 
                             + d7 
                             + d8 
                             + d9 
                             + d10 
                             + d11 
                             + d12 
                             + d13  
                             + d14 
                             + d15  
                             + d16 
                             + d17  
                             + d18 
                             + d19  
                             + d20 
                             + d21  
                             + d22 
                             + d23  
                             + d24
                             ;
}
```

### Steps to Reproduce

_No response_

### Affected Version(s)

U5

_No response_

### Related area

-> Compilation

_No response_

_No response_

_No response_

KavinduZoysa:
## Purpose

 Fixes #40658


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5086,7 +5086,7 @@ private BType checkAndGetType(BLangExpression expr, SymbolEnv env, BLangBinaryEx
         this.dlog.mute();
 
         expr.cloneAttempt++;
-        BType exprCompatibleType = checkExpr(nodeCloner.cloneNode(expr), env, binaryExpr.expectedType, data);
+        BType exprCompatibleType = checkExpr(nodeCloner.cloneNodeOnce(expr), env, binaryExpr.expectedType, data);
```
MaryamZi, line 5089: Do you have a narrowed down sample for which the issue is reproducible?

The whole point of cloning and type-checking is to allow checking a node against different types. So we need a new clone. I'm not sure if this fix is correct.

KavinduZoysa: > Do you have a narrowed down sample for which the issue is reproducible?
> 

This is a narrowed example. 
https://github.com/wso2-enterprise/internal-support-ballerina/issues/344#issuecomment-1579917157
 In this example, we create cloned nodes for lhs exponentially.

MaryamZi: Do we know why that happens? If it is cloned only as intended, shouldn't cause an issue, right?

hasithaa: Don't we have to clean other usages of this? Add a test case, given in the example.

MaryamZi: Had a chat with Kavindu regarding this.




> The whole point of cloning and type-checking is to allow checking a node against different types. So we need a new clone.

For something like `A + B + C`, which is effectively a binary expression with the operands `A + B` and `C`, we clone `A + B` first, but then also clone `A` and `B` separately, With this fix we try to avoid cloning `A` and `B` if `A + B` is cloned.  

But, I think the correct fix is to clone only if required. For example, we don't have to clone `A + B` here, we only need to clone `A` and `B` if they are literals/require inference. We may be able to reuse `org.wso2.ballerinalang.compiler.semantics.analyzer.TypeChecker#requireTypeInference`, but we need to make sure that the type is propagated properly as required. 

This is also reproducible with a type cast expression.

```ballerina
public function main() {
    decimal totalOtherCost = <decimal>  
                             <decimal>  
                             <decimal> 
                             <decimal>  
                             <decimal> 
                             <decimal> 
                             <decimal> 
                             <decimal> 
                             <decimal> 
                             <decimal> 
                             <decimal> 
                             <decimal> 
                             <decimal>  
                             <decimal> 
                             <decimal>  
                             <decimal> 
                             <decimal>  
                             <decimal> 
                             <decimal>  
                             <decimal> 
                             <decimal>  
                             <decimal> 
                             <decimal>  
                             <decimal> 1;
}
```

MaryamZi: Btw, we did previously use `requireTypeInference` for type casts, but removed that check later in https://github.com/ballerina-platform/ballerina-lang/pull/30655. We need to check similar scenarios with the fix we come up for this.

KavinduZoysa: ACK. Change the logic to call the node cloner only for numeric literal.

I will check the type-cast scenario and change it too.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5079,6 +5066,47 @@ public void visit(BLangBinaryExpr binaryExpr, AnalyzerData data) {
         data.resultType = types.checkType(binaryExpr, actualType, data.expType);
     }
 
+    private BType getBinaryOperandType(BLangExpression operand, BLangBinaryExpr binaryExpr, BType referredExpType,
+                                       SymbolEnv env, AnalyzerData data) {
+        if ((referredExpType.tag == TypeTags.FLOAT || referredExpType.tag == TypeTags.DECIMAL ||
+                isOptionalFloatOrDecimal(referredExpType)) && isNumericLiteral(operand)) {
```
MaryamZi, line 5072: Can't we use `requireTypeInference` instead of `isNumericLiteral`?

KavinduZoysa: Tests are passing with `requireTypeInference`. But in that method, it returns true for other non numeric literal cases. Even if those case does not return true in binary-expr context, I think it is better to have a separate method. WDYT?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/TypeChecker.java
```diff
@@ -5079,6 +5066,47 @@ public void visit(BLangBinaryExpr binaryExpr, AnalyzerData data) {
...
+            return checkAndGetType(operand, env, binaryExpr, data);
+        }
+        if (isBinaryExpr(operand)) {
```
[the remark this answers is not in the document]
KavinduZoysa: Basically, if the expr is not a `numeric-literal` we have to type check using `no-type` as expected type. Then only we can get the error `operator not defined ...` other than `incompatible types ...`. That is the purpose of the last if block.
Let's say we have an expression like `(10.0 + 5) - 2`. In the first iteration, type checking happens for `10.0 + 5` with no type, and in 2nd iteration, it will not go to the first if block as expected because there is not an expected type. 
In order to solve this problem, I added the mentioned if block.

sanjana:
Please consider pulling the changes from #40271.

KavinduZoysa:
The current approach does not work for all cases. We have to extend this for unary-expr and type-cast-expr. We are closing this until we figure out another approach.

PR #40670, opened 2023-06-13
Fix heap dump path

NipunaMadhushan:
## Purpose
Currently the given heap dump path breaks 

Fixes https://github.com/ballerina-platform/ballerina-distribution/issues/4541

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/distribution/zip/jballerina/bin/bal
+++ b/distribution/zip/jballerina/bin/bal
@@ -251,7 +251,6 @@ BALLERINA_CLASSPATH=$BALLERINA_CLASSPATH:$BALLERINA_CLASSPATH_EXT
 CMD_LINE_ARGS="-Xbootclasspath/a:"$BALLERINA_XBOOTCLASSPATH" \
                -Xms256m -Xmx1024m \
                -XX:+HeapDumpOnOutOfMemoryError \
-               -XX:HeapDumpPath="''" \
                $JAVA_OPTS \
                -classpath "$BALLERINA_CLASSPATH" \
                -Dballerina.home="$BALLERINA_HOME" \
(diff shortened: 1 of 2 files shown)


keizer619 (approved):

distribution/zip/jballerina/bin/bal
```diff
@@ -251,7 +251,7 @@ BALLERINA_CLASSPATH=$BALLERINA_CLASSPATH:$BALLERINA_CLASSPATH_EXT
 CMD_LINE_ARGS="-Xbootclasspath/a:"$BALLERINA_XBOOTCLASSPATH" \
                -Xms256m -Xmx1024m \
                -XX:+HeapDumpOnOutOfMemoryError \
-               -XX:HeapDumpPath="$(pwd)" \
+               -XX:HeapDumpPath="''" \
```
hasithaa, line 254: Is this correct, In Windows this is empty and here it is not. 

Btw, this is an optional flag. If we want the default behavior, remove this flag. Fix this in `bal.bat` as well.

hasithaa (dismissed):
See inline comment.

distribution/zip/jballerina/bin/bal
NipunaMadhushan: @hasithaa I checked this locally. In Windows, it works when we give the path as `""`. But in Linux, it is working when we give `"''"`.

hasithaa: Don't we have to remove the windows one as well?

NipunaMadhushan: Yes. But have to check in a windows machine. By the time, I will remove from ubuntu since this causes an installer test failure.

NipunaMadhushan: @hasithaa Removed the configuration in windows as well after testing in windows.

keizer619 (approved):

hasithaa (approved):

PR #40766, opened 2023-06-19
Handle check-expr within group-by clause

### LINKED ISSUE #40762: [GroupBy][Bug]:Compiler doesn't validate whether an error handling mechanism is available for group by expression
### Description

```ballerina
public function main() {
    io:println(foo()); //prints the error correctly here
}

function foo() {
    _ = from int _ in [1, 2, 3]
    group by int item = check bar() // no error here even though no matching error handling mechanism
    select item;
}

function bar() returns int|error {
    return error("error");
}
```

### Steps to Reproduce

_No response_

### Affected Version(s)

2201.7.0

_No response_

### Related area

-> Compilation

_No response_

_No response_

_No response_

KavinduZoysa:
## Purpose

Fixes #40762

### DIFF
--- a/tests/jballerina-unit-test/src/test/resources/test-src/query/group_by_clause_negative_semantic.bal
+++ b/tests/jballerina-unit-test/src/test/resources/test-src/query/group_by_clause_negative_semantic.bal
@@ -1,0 +1,30 @@
+// Copyright (c) 2023 WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
+//
+// WSO2 LLC. licenses this file to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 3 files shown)


tests/jballerina-unit-test/src/test/resources/test-src/query/group_by_clause_negative_semantic.bal
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) 2023 WSO2 LLC. (http://www.wso2.org) All Rights Reserved.
```
gimantha, line 1:
```suggestion
// Copyright (c) 2023 WSO2 LLC. (http://www.wso2.org).
```

gimantha (approved):

Issue #2662, opened 2017-06-23
[Transactions] wrong invocation behaviour in transactions when first transaction is without commited block and transaction successful


erandidesilva:
**Sample code**

```
function insertEmployeeDataTwoTransactionsSqlError (json data) (json returnValues){
    sql:Parameter[] parameters = [];
    datatable df;
    sql:ClientConnector connectorInstance = connectorInit:init();
    string msg;
    returnValues = {"message1": "dummy", "message2": "dummy"};
    int i = 0;
    int length = jsons:getInt(data, "$.persons.length()");
    system:println("length" + length);
    try {
        transaction {
            //database insert on employees
            returnValues.message1 = "data inserted to employees";
            system:println("20a");
        }
        transaction {
           //database insert on persons that results in an sql error
        }committed{
            system:println("20b");
            returnValues.message2 = "data inserted to personslanka";
        } aborted{
            returnValues.message2 = "data not inserted to personslanka";
        }
    }catch (errors:Error ex) {
        system:println("20c");
        msg = "Error in database insertion. Data not inserted or retrieved.";
        ex = {msg:msg};
        throw ex;
    }
   return returnValues;
}
```

**Console output**

```
ballerina: deploying service(s) in 'scenario4'
ballerina: started server connector http-9090
length1
1
2
3
4
5
6
7
8
9
loop count in transaction1: 1
20a
20b

```
As shown above, on the successful completion of transaction 1, the commited block of the next transaction is invoked.

anupama-pathirage:
Not an issue.

Checked with the complete original scenario and the sample is working as expected. Reported behavior is due to the logic inside the transaction block.

PR #40901, opened 2023-06-30
Add missing qualifiers for the object type symbol when loaded through BIR

### LINKED ISSUE #40619: [Bug]: Missing Distinct Qualifier in ObjectTypeSymbol
### Description

The `distinct` qualifier of `Interceptor` type in the following code is missing in the `ObjectTypeSymbol`. Tried to access the qualifiers using `objectTypeSymbol.qualifiers()` API. Since the qualifier is missing, it cannot be validated at the compiler plugin.

```ballerina
public type Interceptor distinct service object {
    isolated remote function execute(Context context, Field 'field) returns anydata|error;
};

readonly service class ServiceInterceptor {
    *graphql:Interceptor;

    isolated remote function execute(graphql:Context context, graphql:Field 'field) returns anydata|error {
        anydata|error result = context.resolve('field);
        return result;
    }
}

service graphql:Service on new graphql:Listener(4000) {
    resource function get foo() returns graphql:Interceptor {
        return new ServiceInterceptor();
    }
}
```
This issue is similar to the https://github.com/ballerina-platform/ballerina-lang/issues/37560 which was fixed earlier. But it seems the fix is a partial fix. It doesn't capture the `distinct` qualifier in types like `Interceptors`. This doesn't work in u4, u5, u6. But this works with u5 latest timestamp.

### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

-> Compilation

_No response_

_No response_

_No response_

LakshanWeerasinghe:
## Purpose
$subject

With this fix `distinct` and `isolated` qualifiers will be read from the BIR and available in the BallerinaObjectTypeSymbol. This will cause to fail the test case in [#graphql](https://github.com/ballerina-platform/module-ballerina-graphql/blob/10326f1b6580b5b7d1e730e9d4418df8c823642c/compiler-plugin-tests/src/test/java/io/ballerina/stdlib/graphql/compiler/ServiceValidationTest.java#L248).

Fixes #40619

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
@@ -405,15 +405,15 @@ public class BIRPackageSymbolEnter {
         BInvokableType funcType = (BInvokableType) readBType(dataInStream);
         BInvokableSymbol invokableSymbol =
                 Symbols.createFunctionSymbol(flags, names.fromString(funcName), names.fromString(funcOrigName),
-                                             this.env.pkgSymbol.pkgID, funcType, this.env.pkgSymbol,
-                                             Symbols.isFlagOn(flags, Flags.NATIVE), pos, toOrigin(origin));
+                        this.env.pkgSymbol.pkgID, funcType, this.env.pkgSymbol,
+                        Symbols.isFlagOn(flags, Flags.NATIVE), pos, toOrigin(origin));
         invokableSymbol.source = pos.lineRange().fileName();
         invokableSymbol.retType = funcType.retType;
 
         Scope scopeToDefine = this.env.pkgSymbol.scope;
 
         boolean isResourceFunction = dataInStream.readBoolean();
-        
+
         if (this.currentStructure != null) {
             BType attachedType = Types.getReferredType(this.currentStructure.type);
 
@@ -583,8 +583,8 @@ public class BIRPackageSymbolEnter {
 
         defineMarkDownDocAttachment(symbol, docBytes);
         defineAnnotAttachmentSymbols(dataInStream,
-                                     (isClass || isEnum || symbol.tag == SymTag.TYPE_DEF) ? (Annotatable) symbol :
-                                             null);
+                (isClass || isEnum || symbol.tag == SymTag.TYPE_DEF) ? (Annotatable) symbol :
+                        null);
 
         if (type.tsymbol.name == Names.EMPTY && type.tag != TypeTags.INVOKABLE) {
             type.tsymbol.name = symbol.name;
@@ -704,7 +704,7 @@ public class BIRPackageSymbolEnter {
 
         for (int i = 0; i < attachPointCount; i++) {
             attachPoints.add(AttachPoint.getAttachmentPoint(getStringCPEntryValue(dataInStream),
-                                                            dataInStream.readBoolean()));
+                    dataInStream.readBoolean()));
         }
 
         BType annotationType = readBType(dataInStream);
@@ -1112,6 +1112,7 @@ public class BIRPackageSymbolEnter {
      * @since 0.970.0
      */
     private static class BIRPackageSymbolEnv {
+
         PackageID requestedPackageId;
         Map<Integer, byte[]> unparsedBTypeCPs = new HashMap<>();
         BPackageSymbol pkgSymbol;
@@ -1631,47 +1632,27 @@ public class BIRPackageSymbolEnter {
                     }
                     return finiteType;
                 case TypeTags.OBJECT:
-                    boolean service = inputStream.readByte() == 1;
-
                     pkgCpIndex = inputStream.readInt();
                     pkgId = getPackageId(pkgCpIndex);
 
                     String objName = getStringCPEntryValue(inputStream);
-                    var objFlags = (inputStream.readBoolean() ? Flags.CLASS : 0) | Flags.PUBLIC;
-                    objFlags = inputStream.readBoolean() ? objFlags | Flags.CLIENT : objFlags;
+                    long objSymFlags = inputStream.readLong();
                     BObjectTypeSymbol objectSymbol;
 
-                    if (Symbols.isFlagOn(objFlags, Flags.CLASS)) {
-                        objectSymbol = Symbols.createClassSymbol(objFlags, names.fromString(objName),
-                                                                 env.pkgSymbol.pkgID, null, env.pkgSymbol,
-                                                                 symTable.builtinPos, COMPILED_SOURCE, false);
+                    if (Symbols.isFlagOn(objSymFlags, Flags.CLASS)) {
+                        objectSymbol = Symbols.createClassSymbol(objSymFlags, names.fromString(objName),
+                                env.pkgSymbol.pkgID, null, env.pkgSymbol,
+                                symTable.builtinPos, COMPILED_SOURCE, false);
                     } else {
-                        objectSymbol = Symbols.createObjectSymbol(objFlags, names.fromString(objName),
-                                                                  env.pkgSymbol.pkgID, null, env.pkgSymbol,
-                                                                  symTable.builtinPos, COMPILED_SOURCE);
+                        objectSymbol = Symbols.createObjectSymbol(objSymFlags, names.fromString(objName),
+                                env.pkgSymbol.pkgID, null, env.pkgSymbol,
+                                symTable.builtinPos, COMPILED_SOURCE);
                     }
 
                     objectSymbol.scope = new Scope(objectSymbol);
                     BObjectType objectType;
                     // Below is a temporary fix, need to fix this properly by using the type tag
                     objectType = new BObjectType(objectSymbol);
-
-                    if (service) {
-                        objectType.flags |= Flags.SERVICE;
-                        objectSymbol.flags |= Flags.SERVICE;
-                    }
-                    if (isImmutable(flags)) {
-                        objectSymbol.flags |= Flags.READONLY;
-                    }
-                    if (Symbols.isFlagOn(flags, Flags.ANONYMOUS)) {
-                        objectSymbol.flags |= Flags.ANONYMOUS;
-                    }
-                    if (Symbols.isFlagOn(flags, Flags.DISTINCT)) {
-                        objectSymbol.flags |= Flags.DISTINCT;
-                    }
-                    if (Symbols.isFlagOn(flags, Flags.ISOLATED)) {
-                        objectSymbol.flags |= Flags.ISOLATED;
-                    }
                     objectType.flags = flags;
                     objectSymbol.type = objectType;
                     addShapeCP(objectType, cpI);
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
@@ -1666,6 +1666,12 @@ public class BIRPackageSymbolEnter {
                     if (Symbols.isFlagOn(flags, Flags.ANONYMOUS)) {
                         objectSymbol.flags |= Flags.ANONYMOUS;
                     }
+                    if (Symbols.isFlagOn(flags, Flags.DISTINCT)) {
+                        objectSymbol.flags |= Flags.DISTINCT;
+                    }
+                    if (Symbols.isFlagOn(flags, Flags.ISOLATED)) {
+                        objectSymbol.flags |= Flags.ISOLATED;
+                    }
                     objectType.flags = flags;
                     objectSymbol.type = objectType;
                     addShapeCP(objectType, cpI);
--- a/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
+++ b/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
@@ -405,15 +405,15 @@ public class BIRPackageSymbolEnter {
         BInvokableType funcType = (BInvokableType) readBType(dataInStream);
         BInvokableSymbol invokableSymbol =
                 Symbols.createFunctionSymbol(flags, names.fromString(funcName), names.fromString(funcOrigName),
-                        this.env.pkgSymbol.pkgID, funcType, this.env.pkgSymbol,
-                        Symbols.isFlagOn(flags, Flags.NATIVE), pos, toOrigin(origin));
+                                             this.env.pkgSymbol.pkgID, funcType, this.env.pkgSymbol,
+                                             Symbols.isFlagOn(flags, Flags.NATIVE), pos, toOrigin(origin));
         invokableSymbol.source = pos.lineRange().fileName();
         invokableSymbol.retType = funcType.retType;
 
         Scope scopeToDefine = this.env.pkgSymbol.scope;
 
         boolean isResourceFunction = dataInStream.readBoolean();
-
+        
         if (this.currentStructure != null) {
             BType attachedType = Types.getReferredType(this.currentStructure.type);
 
@@ -583,8 +583,8 @@ public class BIRPackageSymbolEnter {
 
         defineMarkDownDocAttachment(symbol, docBytes);
         defineAnnotAttachmentSymbols(dataInStream,
-                (isClass || isEnum || symbol.tag == SymTag.TYPE_DEF) ? (Annotatable) symbol :
-                        null);
+                                     (isClass || isEnum || symbol.tag == SymTag.TYPE_DEF) ? (Annotatable) symbol :
+                                             null);
 
         if (type.tsymbol.name == Names.EMPTY && type.tag != TypeTags.INVOKABLE) {
             type.tsymbol.name = symbol.name;
@@ -704,7 +704,7 @@ public class BIRPackageSymbolEnter {
 
         for (int i = 0; i < attachPointCount; i++) {
             attachPoints.add(AttachPoint.getAttachmentPoint(getStringCPEntryValue(dataInStream),
-                    dataInStream.readBoolean()));
+                                                            dataInStream.readBoolean()));
         }
 
         BType annotationType = readBType(dataInStream);
@@ -1112,7 +1112,6 @@ public class BIRPackageSymbolEnter {
      * @since 0.970.0
      */
     private static class BIRPackageSymbolEnv {
-
         PackageID requestedPackageId;
         Map<Integer, byte[]> unparsedBTypeCPs = new HashMap<>();
         BPackageSymbol pkgSymbol;
--- a/tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolBIRTest.java
+++ b/tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolBIRTest.java
@@ -58,11 +58,16 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
+import static io.ballerina.compiler.api.symbols.Qualifier.CLIENT;
+import static io.ballerina.compiler.api.symbols.Qualifier.DISTINCT;
+import static io.ballerina.compiler.api.symbols.Qualifier.ISOLATED;
+import static io.ballerina.compiler.api.symbols.Qualifier.SERVICE;
 import static io.ballerina.compiler.api.symbols.SymbolKind.ANNOTATION;
 import static io.ballerina.compiler.api.symbols.SymbolKind.CONSTANT;
 import static io.ballerina.compiler.api.symbols.SymbolKind.FUNCTION;
 import static io.ballerina.compiler.api.symbols.SymbolKind.MODULE;
 import static io.ballerina.compiler.api.symbols.SymbolKind.SERVICE_DECLARATION;
+import static io.ballerina.compiler.api.symbols.SymbolKind.TYPE;
 import static io.ballerina.compiler.api.symbols.SymbolKind.TYPE_DEFINITION;
 import static io.ballerina.compiler.api.symbols.SymbolKind.VARIABLE;
 import static io.ballerina.compiler.api.symbols.TypeDescKind.INT;
@@ -175,17 +180,15 @@ public class SymbolBIRTest {
 
         BallerinaModule fooModule = (BallerinaModule) symbolsInScope.stream()
                 .filter(sym -> sym.getName().get().equals("testproject")).findAny().get();
-        SemanticAPITestUtils.assertList(fooModule.functions(), List.of("loadHuman", 
+        SemanticAPITestUtils.assertList(fooModule.functions(), List.of("loadHuman",
                 "testAnonTypeDefSymbolsIsNotVisible", "add", "testFnA", "testFnB"));
         SemanticAPITestUtils.assertList(fooModule.constants(), List.of("RED", "GREEN", "BLUE", "PI", "TRUE", "FALSE"));
-        
-        SemanticAPITestUtils.assertList(fooModule.typeDefinitions(), List.of("HumanObj", "ApplicationResponseError", 
-                "Person", "BasicType", "Digit", "FileNotFoundError", "EofError", "Error", "Pet", "Student", "Cat", 
-                "Annot", "Detail", "Service", "FnTypeA", "FnTypeB", "Address"));
-
-        
         SemanticAPITestUtils.assertList(fooModule.classes(), List.of("PersonObj", "Dog", "EmployeeObj", "Human", 
                 "Client", "Response"));
+        SemanticAPITestUtils.assertList(fooModule.typeDefinitions(), List.of("HumanObj", "ApplicationResponseError",
+                "Person", "BasicType", "Digit", "FileNotFoundError", "EofError", "Error", "Pet", "Student", "Cat",
+                "Annot", "Detail", "Service", "FnTypeA", "FnTypeB", "Address", "InterceptorClient", 
+                "InterceptorService"));
         SemanticAPITestUtils.assertList(fooModule.enums(), List.of("Colour"));
 
         List<String> allSymbols = getSymbolNames(fooPkgSymbol, 0);
@@ -319,8 +322,8 @@ public class SymbolBIRTest {
 
         for (SymbolInfo val : expectedValues) {
             assertTrue(actualValues.stream()
-                               .anyMatch(sym -> val.equals(new SymbolInfo(sym.getName().get(), sym.kind()))),
-                       "Symbol not found: " + val);
+                            .anyMatch(sym -> val.equals(new SymbolInfo(sym.getName().get(), sym.kind()))),
+                    "Symbol not found: " + val);
 
         }
     }
@@ -361,6 +364,7 @@ public class SymbolBIRTest {
     }
 
     static class SymbolInfo {
+
         private String name;
         private SymbolKind kind;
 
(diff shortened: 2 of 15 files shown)


tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolBIRTest.java
```diff
@@ -360,6 +365,7 @@ private List<SymbolInfo> createSymbolInfoList(Object[][] infoArr) {
     }
 
     static class SymbolInfo {
+
```
lochana-chathura, line 368:
```suggestion
```

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -1674,6 +1674,12 @@ public BType readType(int cpI) throws IOException {
                     if (Symbols.isFlagOn(flags, Flags.ANONYMOUS)) {
                         objectSymbol.flags |= Flags.ANONYMOUS;
                     }
+                    if (Symbols.isFlagOn(flags, Flags.DISTINCT)) {
```
pcnfernando, line 1677: Can you add a few tests that loaded through BIR and check for type relationships?
eg:

```
import ballerina/io;

public type Foo distinct isolated client object {
    isolated function execute() returns anydata|error;
};

public type Bar distinct isolated client object {
    isolated function execute() returns anydata|error;
};

public function test(Foo foo, Bar bar) {
    io:println(foo is Bar); //incompatible types: 'Foo' will not be matched to 'Bar'
}
```

LakshanWeerasinghe: Added.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_negative.bal
```diff
@@ -0,0 +1,20 @@
...
+}
+
+function f3(objs:Xyz xyz) {
+    assertEquality(xyz is objs:Qux, false);
+}
+
+function f4(objs:Qux qux) {
+    assertEquality(qux is objs:Xyz, false);
+}
+
```
pcnfernando, line 18: Add another test that checks whether its of the same type. Although the actual test would not run, there wont be a compile time. Furthermore, can you add few positive tests as well that validates the types at runtime. Assigning it a any and validating later. Make sure add for both isolated/distinct together and seperately
eg:
```
function f5(objs:Qux qux) {
    assertEquality(qux is objs:Qux, true);
}
```

LakshanWeerasinghe: Added.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -1674,6 +1674,12 @@ public BType readType(int cpI) throws IOException {
...
+                        objectSymbol.flags |= Flags.DISTINCT;
+                    }
+                    if (Symbols.isFlagOn(flags, Flags.ISOLATED)) {
```
MaryamZi, line 1680: Did we check how https://github.com/ballerina-platform/ballerina-lang/issues/40619#issuecomment-1609088407 worked as expected without this?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
MaryamZi, line 1677: Wouldn't this approach still leave room for error? Can't we read/write the symbol flags and set that directly (if `flags` type flags are different)?

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_negative.bal
```diff
@@ -0,0 +1,20 @@
...
+function assertEquality(any|error actual, any|error expected) {
```
MaryamZi, line 19: Is this and calls necessary for this test?

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
```diff
@@ -0,0 +1,88 @@
+import bir/objs;
```
MaryamZi, line 1: Missing licence header.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
```diff
@@ -0,0 +1,88 @@
...
+        }
+    };
+    
+    objs:Qux qux = isolated service object {
+        public isolated function execute() returns anydata|error {
+            return "qux";
+        }
+    };
+    
+    f1(foo);
```
MaryamZi, line 64: I'd move assertions close to the variable declarations. Why do you need separate functions (e.g., f1, f2) for assertions btw?

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
```diff
@@ -0,0 +1,88 @@
+import bir/objs;
+
+function f1(objs:Foo foo) {
+    assertEquality(foo is objs:Foo, true);
```
MaryamZi, line 4: I'd add assertTrue, assertFalse functions.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
```diff
@@ -0,0 +1,88 @@
...
+    f3(xyz);
+    f4(qux);
+    f5(foo);
+    f6(baz);
+    f7(xyz);
+    f8(qux);
+    f9(foo);
+}
+
+function assertEquality(any|error actual, any|error expected) {
```
MaryamZi, line 75: Don't we only pass anydata values atm? I would change the param types and simplify the logic.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_projects/bir_test_project/Ballerina.toml
```diff
@@ -0,0 +1,5 @@
+[package]
+org= "bir"
```
MaryamZi, line 2:
```suggestion
org = "bir"
```

Fix in the rest also.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_projects/bir_test_project/Ballerina.toml
```diff
@@ -0,0 +1,5 @@
...
+name= "objs"
+version= "0.1.0"
+exported=["serv_classes"]
```
MaryamZi, line 5: Is this required?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
LakshanWeerasinghe: Yes, At the node level isolated [flag](https://github.com/ballerina-platform/ballerina-lang/blob/d08328e9b726c32553ce6bf9a5f2c70412a44e86/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java#L1677) information was available. This information was missing in the symbol and this symbol is used in the semantic API.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
LakshanWeerasinghe: Do you mean at the type writer in a place like [this](https://github.com/ballerina-platform/ballerina-lang/blob/d08328e9b726c32553ce6bf9a5f2c70412a44e86/compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/bir/writer/BIRTypeWriter.java#L449) we should write the symbol flags value to the BIR. In the BIRSymbolEnter we should read the flag values and set in the symbol, rather than reading flags one by one.

MaryamZi: Yeah?

MaryamZi: We need to check if this can cause any issues (not that I can think of any) since it will obviously mean more flags.

LakshanWeerasinghe: I'll try this approach and give an update.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_projects/bir_test_project/Ballerina.toml
LakshanWeerasinghe: We need to have a `Ballerina.toml`. But we don't need a value for the `exported` key.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_negative.bal
LakshanWeerasinghe: Yes. We need to check after the BIR changes whether there any issues w.r.t objectTypes.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
LakshanWeerasinghe: The approach you have suggested is working.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
```diff
@@ -0,0 +1,97 @@
...
+    
+    objs:Xyz xyz = isolated service object {
+        public isolated function execute() returns anydata|error {
+            return "xyz";
+        }
+    };
+    f2(xyz);
+    f5(xyz);
+    
+    objs:Qux qux = isolated service object {
```
SasinduDilshara, line 72: Shall we add a test with module level defn?.
Defining this in module level and test the  functionality

```
objs:Qux qux = isolated service object {
      public isolated function execute() returns anydata|error {
             return "qux";
         }
     };
```

LakshanWeerasinghe: Added.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_negative.bal
pcnfernando: I think Maryam is questioning the fact that since this specific sample fails at CT and the only validation happening when calling `assertEquality` is checking whether those object types are assignable to `any|error`; which is not actually needed.
I think you can update the samples in the below manner
```ballerina
function f1(objs:Bar bar) {
    _ = bar is objs:Foo;
}
```

LakshanWeerasinghe: Addressed [#90c5ee9](https://github.com/ballerina-platform/ballerina-lang/pull/40901/commits/90c5ee9af6208cf463782cea723e06a708c823fd).

tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolBIRTest.java
```diff
@@ -433,4 +438,30 @@ public void testIntersectionTypeDefSymbolPosBIR() {
...
+    public void testObjectTypeSymbolDefQualifiers() {
+        Project project = BCompileUtil.loadProject("test-src/object_symbol_qualifiers.bal");
+        Package currentPackage = project.currentPackage();
+        Document srcFile = getDocumentForSingleSource(project);
+        SemanticModel model = getDefaultModulesSemanticModel(project);
+        
+        Optional<Symbol> symbol = model.symbol(srcFile, from(19, 20));
+        assertTrue(symbol.isPresent());
+        assertEquals(symbol.get().kind(), TYPE);
+        assertEquals(((TypeReferenceTypeSymbol) symbol.get()).typeDescriptor().typeKind(), OBJECT);
```
pcnfernando, line 452: Move to separate variables when possible so that you can reuse

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -405,15 +405,15 @@ private void defineFunction(DataInputStream dataInStream) throws IOException {
         BInvokableType funcType = (BInvokableType) readBType(dataInStream);
         BInvokableSymbol invokableSymbol =
                 Symbols.createFunctionSymbol(flags, names.fromString(funcName), names.fromString(funcOrigName),
-                                             this.env.pkgSymbol.pkgID, funcType, this.env.pkgSymbol,
-                                             Symbols.isFlagOn(flags, Flags.NATIVE), pos, toOrigin(origin));
+                        this.env.pkgSymbol.pkgID, funcType, this.env.pkgSymbol,
```
pcnfernando, line 408: Don't re-format existing code that you are not updating

tests/ballerina-compiler-api-test/src/test/java/io/ballerina/semantic/api/test/SymbolBIRTest.java
LakshanWeerasinghe: Fixed in [#2662](https://github.com/ballerina-platform/ballerina-lang/pull/40901/commits/26627582f1e188891880ffecb62c539c855b662f)

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
LakshanWeerasinghe: Reverted in [#0763](https://github.com/ballerina-platform/ballerina-lang/pull/40901/commits/076383542895d77de1ad438372cf151195500550)

pcnfernando:
@LakshanWeerasinghe please create an issue for this on [std-lib](https://github.com/ballerina-platform/ballerina-standard-library) requesting to change their test with this fix?

pcnfernando (dismissed):

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_projects/bir_test_project/Ballerina.toml
MaryamZi: Yes, I meant this line. :)

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/bala/object/ObjectInBalaTest.java
```diff
@@ -546,6 +547,27 @@ public void testObjectReferingTypeFromBalaNegative() {
...
+        int i = 0;
+        BAssertUtil.validateError(result, i++, 
+                "incompatible types: 'bir/objs:0.1.0:Bar' will not be matched to 'bir/objs:0.1.0:Foo'", 20, 9);
+        BAssertUtil.validateError(result, i++, 
+                "incompatible types: 'bir/objs:0.1.0:Foo' will not be matched to 'bir/objs:0.1.0:Bar'", 24, 9);
+        BAssertUtil.validateError(result, i++, 
+                "incompatible types: 'bir/objs:0.1.0:Xyz' will not be matched to 'bir/objs:0.1.0:Qux'", 28, 9);
+        BAssertUtil.validateError(result, i, 
+                "incompatible types: 'bir/objs:0.1.0:Qux' will not be matched to 'bir/objs:0.1.0:Xyz'", 32, 9);
+        Assert.assertEquals(result.getErrorCount(), 4);
```
MaryamZi, line 562:
```suggestion
        BAssertUtil.validateError(result, i++, 
                "incompatible types: 'bir/objs:0.1.0:Qux' will not be matched to 'bir/objs:0.1.0:Xyz'", 32, 9);
        Assert.assertEquals(result.getErrorCount(), i);
```

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_negative.bal
```diff
@@ -0,0 +1,33 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import bir/objs;
+
+function f1(objs:Bar bar) {
+    _ = bar is objs:Foo;
```
MaryamZi, line 20: Can't we just do

```suggestion
    objs:Foo _ = bar;
```

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_negative.bal
```diff
@@ -0,0 +1,33 @@
...
+    _ = foo is objs:Bar;
+}
+
+function f3(objs:Xyz xyz) {
+    _ = xyz is objs:Qux;
+}
+
+function f4(objs:Qux qux) {
+    _ = qux is objs:Xyz;
+}
```
MaryamZi, line 33: How about also testing with an object defined in the current module vs one from the imported module, where the qualifiers are incompatible?

MaryamZi: Can add positive tests also.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
```diff
@@ -0,0 +1,104 @@
...
+    assertFalse(foo is objs:Bar);
+}
+
+function testObjectTypeAssignability() {
+    objs:Foo foo = isolated client object {
+         public isolated function execute() returns anydata|error {
+           return "foo";
+        }
+    };
+    f1(foo);
```
MaryamZi, line 59: Do we really need the separate functions? Can't we do the checks here itself? That would help with readability IMO, since we won't have to go to and from functions.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
```diff
@@ -0,0 +1,104 @@
...
+import bir/objs;
+
+objs:Qux quxMod = isolated service object {
+    public isolated function execute() returns anydata|error {
+         return "qux";
+    }
+};
+
+function f1(objs:Foo foo) {
+    assertTrue(foo is objs:Foo);
```
MaryamZi, line 26: I would also add some compile-time positive tests like

```ballerina
T _ = t; // where `t` will be of a different type
```

Multiple combinations
- source and target from a imported module
- source from current module, target from imported module
- source from imported module, target from current module

MaryamZi: These combinations can also be covered in runtime positive (`is`) tests.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_projects/bir_test_project/main.bal
```diff
@@ -0,0 +1,19 @@
+public type Foo distinct isolated client object {
```
MaryamZi, line 1: Licence header?

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_projects/bir_test_project/Ballerina.toml
```diff
@@ -0,0 +1,4 @@
+[package]
+org=  "bir"
```
MaryamZi, line 2:
```suggestion
org = "bir"
```

Fix other lines also.

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/bala/object/ObjectInBalaTest.java
```diff
@@ -546,6 +547,27 @@ public void testObjectReferingTypeFromBalaNegative() {
...
+    }
+
+    @Test
+    public void testObjectTypeAssignability() {
```
MaryamZi, line 566: Can we be more descriptive here, mentioning flags/qualifiers?

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
LakshanWeerasinghe: These functions were added to address this [#suggestion](https://github.com/ballerina-platform/ballerina-lang/pull/40901#discussion_r1257754638). By removing the functions we cannot achieve the assigning validation.

MaryamZi: Why can't we do

```suggestion
    assertTrue(<any> foo is objs:Foo);
```

or 

```suggestion
    any a = foo;
    assertTrue(a is objs:Foo);
```

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
LakshanWeerasinghe: Added in [#050dc4](https://github.com/ballerina-platform/ballerina-lang/pull/40901/commits/050dc4b510554b9df7e59929bf5265976524497b).

tests/jballerina-unit-test/src/test/java/org/ballerinalang/test/bala/object/ObjectInBalaTest.java
LakshanWeerasinghe: Added in [#050dc4](https://github.com/ballerina-platform/ballerina-lang/pull/40901/commits/050dc4b510554b9df7e59929bf5265976524497b)

MaryamZi:
The GraphQL (level 7) build is failing? Is this being looked into (where it has to be fixed)?

LakshanWeerasinghe:
> The GraphQL (level 7) build is failing? Is this being looked into (where it has to be fixed)?

Yes. It's tracked by this [#task](https://github.com/ballerina-platform/ballerina-standard-library/issues/4672). The failure is due to the correct qualifier list is getting from the SemanticAPI so the validation test is failing.

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_negative.bal
```diff
@@ -0,0 +1,41 @@
...
+function f3(objs:Xyz xyz) {
+    objs:Qux _ = xyz;
+}
+
+function f4(objs:Qux qux) {
+    objs:Xyz _ = qux;
+}
+
+function f5(Foo foo) {
+    objs:Foo _ = foo;
```
MaryamZi, line 40: Can we check the other way too? IMO, those are more useful, since when both the types are read from the BIR, flags were ignored for both, so doesn't make a difference?

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
```diff
@@ -0,0 +1,98 @@
...
+        public isolated function execute() returns anydata|error {
+            return "quxx";
+        }
+    };
+    
+    objs:Muxx _ = quxx;
+    Quxx _ = quxx;
+    assertTrue(quxx is Quxx);
+    
+    Quxx quxxMod = isolated client object {
```
MaryamZi, line 72:
```suggestion
    Quxx _ = isolated client object {
```

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_projects/bir_test_project/main.bal
```diff
@@ -0,0 +1,43 @@
...
+    public isolated function execute() returns anydata|error;
+};
+
+public type Quxx isolated client object {
+    public isolated function execute() returns anydata|error;
+};
+
+public type Muxx isolated client object {
+    public isolated function execute() returns anydata|error;
+};
```
MaryamZi, line 43:
```suggestion
};

```

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_negative.bal
LakshanWeerasinghe: Updated.

MaryamZi:
> > The GraphQL (level 7) build is failing? Is this being looked into (where it has to be fixed)?
> 
> Yes. It's tracked by this [#task](https://github.com/ballerina-platform/ballerina-standard-library/issues/4672). The failure is due to the correct qualifier list is getting from the SemanticAPI so the validation test is failing.

Is there a PR ready for this?

tests/jballerina-unit-test/src/test/resources/test-src/bala/test_bala/object/test_bir_positive.bal
```diff
@@ -0,0 +1,98 @@
...
+
+function testObjectTypeAssignabilityWithQualifiers() {
+    objs:Foo foo = isolated client object {
+         public isolated function execute() returns anydata|error {
+           return "foo";
+        }
+    };
+    assertTrue(foo is objs:Foo);
+    assertTrue(<any>foo is objs:Foo);
+    assertFalse(<any>foo is objs:Bar);
```
MaryamZi, line 37: Shall we also add false tests (both ways) for when one type is in the current module and the other in a different module?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/BIRPackageSymbolEnter.java
```diff
@@ -1631,41 +1631,27 @@ public BType readType(int cpI) throws IOException {
...
+                                symTable.builtinPos, COMPILED_SOURCE);
                     }
 
                     objectSymbol.scope = new Scope(objectSymbol);
                     BObjectType objectType;
                     // Below is a temporary fix, need to fix this properly by using the type tag
                     objectType = new BObjectType(objectSymbol);
-
-                    if (service) {
-                        objectType.flags |= Flags.SERVICE;
```
rdulmina, line 1660: Don't we need to set the object type flags? Since we include `service` flag in the type previously there may be depending usages.

LakshanWeerasinghe: There is no changed has been done to the `objectType.flags` value. If you look at the line 1655 we are explicitly set the flags value. This line is anyway has no impact.

rdulmina: Oh I c its fine then

MaryamZi (dismissed):

MaryamZi (approved):

pcnfernando (approved):

PR #40964, opened 2023-07-10
[Master] Reduce creating unwanted type definitions for const annotations

### LINKED ISSUE #40914: [Improvement]: Remove unnecessary type definition creation for interop function definitions
### Description

$title

If an interop function is defined in Ballerina it creates unnecessary type definitions by the front-end per each interop function.

For xample, langlibs have high no. of interop functions. (eg: float langlib nearly 35 and it has 70 BLangTypeDefinitions in the BLangPackage - 2 per one interop function).

image
image

Hence we have high no. of unnecessary typedesc classes, value classes, frame classes that result in unnecessarily large jar size.

### Describe your problem(s)

E.g.:

```bal
import ballerina/jballerina.java;

public isolated function abs(float x) returns float = @java:Method {
    'class: "org.ballerinalang.langlib.floatingpoint.Abs",
    name: "abs"
} external;

public isolated function max(float... xs) returns float = @java:Method {
    'class: "org.ballerinalang.langlib.floatingpoint.Max",
    name: "max"
} external;

public isolated function absBal(float x) returns float {
    return x > 0.0 ? x : -x;
}
```

For the first 2 interop functions, the front-end creates records {class:.., name:..} and 2 type definitions for each interop function.
image

### Describe your solution(s)

We see the type definitions created for all interop functions are mostly the same. We can try removing all of them or keep at most just 2, thus removing unnecessary typeDefs and attached functions in the BIRPackage. 

### Related area

-> Compilation

_No response_

_No response_

_No response_

prakanth97:
## Purpose

Fixes #40914 

## Approach

- According to the spec, information related to source annotation shouldn't be propagated to runtime. So, we removed all the type definitions created for constant values of source annotation.
- Remove unnecessary type definitions created for constant annotations. It reduced by approximately half.

## Samples
For the following code, the jar size is reduced from 9.773571 Mb to 7.729598 Mb (20.913 %). Tried the native image size for same code in mac is reduced from 41.151857 Mb to 31.244385 Mb (24.075 %).
```
public function main() {
    int numbers = addNumbers(1, 2);
}

function addNumbers(int i, int j) returns int {
    return i + j;
}
```

## Check List 
- [ ] Updated Change Log
- [ ] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
```diff
@@ -579,6 +581,10 @@ BLangConstantValue constructBLangConstantValueWithExactType(BLangExpression expr
 
         this.anonTypeNameSuffixes = anonTypeNameSuffixes;
         updateConstantType(constantSymbol, expression, env);
+        BLangTypeDefinition typeDefinition = createdTypeDefinitions.get(constantSymbol.type.tsymbol);
+        if (isSourceOnlyAnon && typeDefinition != null) {
```
ushirask, line 585: If we do
```suggestion
        BLangTypeDefinition typeDefinition = isSourceOnlyAnon ? createdTypeDefinitions.get(constantSymbol.type.tsymbol) : null;
        if (typeDefinition != null) {
```
we can avoid unnecessary `get` to the map but might affect readability. WDY think?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/util/Flags.java
```diff
@@ -82,6 +82,7 @@ public class Flags {
     public static final long ENUM_MEMBER = INFER << 1;                          //  41
     public static final long QUERY_LAMBDA = ENUM_MEMBER << 1;                   //  42
     public static final long EFFECTIVE_TYPE_DEF = QUERY_LAMBDA << 1;            //  43
+    public static final long SOURCE_ANNOTATION = EFFECTIVE_TYPE_DEF << 1;       // 44
```
ushirask, line 85: Incorrect formatting.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
prakanth97: +1

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
```diff
@@ -579,6 +581,11 @@ BLangConstantValue constructBLangConstantValueWithExactType(BLangExpression expr
 
         this.anonTypeNameSuffixes = anonTypeNameSuffixes;
         updateConstantType(constantSymbol, expression, env);
+        BLangTypeDefinition typeDefinition = isSourceOnlyAnon ? createdTypeDefinitions.get(constantSymbol.type.tsymbol)
+                : null;
+        if (typeDefinition != null) {
```
lochana-chathura, line 586: the reason to have two if conditions (at L584 and L586) is `createdTypeDefinitions.get(constantSymbol.type.tsymbol)` could also be `null`?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
```diff
@@ -579,6 +581,11 @@ BLangConstantValue constructBLangConstantValueWithExactType(BLangExpression expr
...
+            typeDefinition.symbol.flags |= Flags.SOURCE_ANNOTATION;
+        }
```
hasithaa, line 588: A better code, 
```
Optional.ofNullable(isSourceOnlyAnon ? createdTypeDefinitions.get(constantSymbol.type.tsymbol) : null)
        .ifPresent(typeDefinition -> typeDefinition.symbol.flags |= Flags.SOURCE_ANNOTATION);
```

hasithaa:
Can you add a test case, where you have a record type definition which used for both annotation purposes and used for normal value creations?

prakanth97:
> c6e9721

Added [c6e9721](https://github.com/ballerina-platform/ballerina-lang/pull/40964/commits/c6e9721c1a9efe4eca66031669b2a432a72fcdff)

prakanth97:
> Can you add a test case, where you have a record type definition which used for both annotation purposes and used for normal value creations?

Added [c6e9721](https://github.com/ballerina-platform/ballerina-lang/pull/40964/commits/c6e9721c1a9efe4eca66031669b2a432a72fcdff)

hasithaa (approved):

tests/jballerina-unit-test/src/test/resources/test-src/typedefs/type-definitions.bal
```diff
@@ -338,6 +338,29 @@ public function returnTupleWithSingletonType2() returns X2 {
     return ["x", 2, ()];
 }
 
+type Data record {|
+    string value;
+|};
+
+const annotation Data dataAnon on type;
```
MaryamZi, line 345: Anon or Anot?

tests/jballerina-unit-test/src/test/resources/test-src/typedefs/type-definitions.bal
```diff
@@ -338,6 +338,29 @@ public function returnTupleWithSingletonType2() returns X2 {
...
+}
+type Person2 record {|
+    string name;
+|};
+
+function testAnnotWithRecordTypeDefinition() {
+    Person2 foo = {name: "James"};
+    typedesc<any> t = typeof foo;
+    Data? annon = t.@dataAnon;
+    assertEquality("{\"value\":\"T1\"}", annon.toString());
```
MaryamZi, line 358: Generally better to assert the value rather than the string representation IMO (since it'll take types also into account).

tests/jballerina-unit-test/src/test/resources/test-src/typedefs/type-definitions.bal
```diff
@@ -338,6 +338,29 @@ public function returnTupleWithSingletonType2() returns X2 {
...
+
+    Data data = {value: "T2"};
+    assertEquality("{\"value\":\"T2\"}" , data.toString());
```
MaryamZi, line 361: Is this to test the availability of Data at runtime? If so, aren't tests that actually validate against `Data` better? E.g., positive and negative `is` checks. Can also use a record with a default, use cloneWithType with it with a value that doesn't have a field for the field with the default, and test that the default is used.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/SymbolResolver.java
```diff
@@ -2588,6 +2589,15 @@ public void populateAnnotationAttachmentSymbol(BLangAnnotationAttachment annotat
                                                                                  constantSymbol);
     }
 
+    private boolean isSourceAnonOnly(Set<AttachPoint> attachPoints) {
```
MaryamZi, line 2592: Annot?

MaryamZi: Need to change where called also.

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
```diff
@@ -667,8 +672,8 @@ private void updateConstantType(BConstantSymbol symbol, BLangExpression expr, Sy
             return;
         }
 
-        if (resolvedType.getKind() == TypeKind.INTERSECTION && isListOrMapping(type.tag)) {
-            expr.setBType(((BIntersectionType) resolvedType).effectiveType);
+        if (resolvedType.getKind() == TypeKind.RECORD && isListOrMapping(type.tag)) {
```
MaryamZi, line 675: Was this intersection only for records? Not for lists (tuples)?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
```diff
@@ -846,9 +851,23 @@ private BType createRecordType(BLangExpression expr, BConstantSymbol constantSym
...
+        updateRecordFields(recordType, pos, env);
+        recordType.tsymbol.flags |= Flags.READONLY;
+        recordType.flags |= Flags.READONLY;
+        return recordType;
+    }
+
+    private void updateRecordFields(BRecordType recordType, Location pos, SymbolEnv env) {
+        BTypeSymbol structureSymbol = recordType.tsymbol;
+        for (BField field : recordType.fields.values()) {
+            field.type = ImmutableTypeCloner.getImmutableType(pos, types, field.type, env,
```
MaryamZi, line 863: Within this we still create type defs, right? E.g., mapping within mapping. Shall we create an issue to avoid this also as a future improvement?

compiler/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/semantics/analyzer/ConstantValueResolver.java
prakanth97: Since this fix is only for the constant annotation. Then record would be the only possible case not tuple. Right?

MaryamZi: I was thinking about when there's a list as a member (field) of the annotation? I don't think we get here for that though.

```ballerina
const annotation map<map<string>[]> annot on source var;

@annot {
    x: [{a: "hello", b: "world"}]
}
var val = 1;
```

PR #41083, opened 2023-07-25
Fix sorting in Record type descriptor node context

### LINKED ISSUE #40617: [Improvement]: Improve sorting of trie completion items
### Description

Consider the following capture

image

The `log: setOutput()` completion item depicted above should not be given the highest priority in the above context. basic types and other user-defined types should be given the highest priority. 

### Describe your problem(s)

```ballerina
import ballerina/log;
import ballerina/http;

final http:Client clientEP = check new ("http://localhost:9090/");

type Person record {|
    string name;
    int age;
    s<cursor>
|};

type Params record {|

|};

    // http:Response res = check clientEP->/test;
    // log:printInfo("response obtained", statuscode = res.statusCode, payload = check res.getTextPayload());
    test("hello", 1, 2);
    testIncludedRecordParams();
}

public function test(string arg1, int... restArg) {

}

public function testIncludedRecordParams() {

}

```

### Describe your solution(s)

_No response_

### Related area

-> Editor

_No response_

_No response_

_No response_

mindula:
## Purpose
This PR fixes the sorting order in record type descriptor node context.

Fixes #40617

## Samples
Screenshot 2023-07-25 at 14 01 39

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [x] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [ ] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples

### DIFF
--- a/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
+++ b/language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
@@ -308,6 +308,9 @@ public abstract class AbstractCompletionProvider<T extends Node> implements Ball
                 completionItem.setInsertText(moduleName + ":" + insertText);
                 completionItem.setLabel(moduleName + ":" + label);
                 completionItems.add(lsCompletionItem);
+                if (completionItem.getFilterText() != null) {
+                    completionItem.setFilterText(moduleName + "_" + completionItem.getFilterText());
+                }
             }
         });
         return completionItems;
--- a/language-server/modules/langserver-core/src/test/java/org/ballerinalang/langserver/completion/RecordTypeDescriptorTest.java
+++ b/language-server/modules/langserver-core/src/test/java/org/ballerinalang/langserver/completion/RecordTypeDescriptorTest.java
@@ -17,7 +17,11 @@
  */
 package org.ballerinalang.langserver.completion;
 
+import org.ballerinalang.langserver.commons.workspace.WorkspaceDocumentException;
 import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.io.IOException;
 
 /**
  * Record Type Descriptor Context tests.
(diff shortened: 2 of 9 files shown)


malinthar (commented):
IIUC, this change does not change the sorting order. It rather fixes the invalid sorting caused when the filter text is incorrect, right? If so, can you mention the order of the completion items(kinds)

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
```diff
@@ -308,6 +308,9 @@ private List<LSCompletionItem> getTypeItems(BallerinaCompletionContext context)
                 completionItem.setInsertText(moduleName + ":" + insertText);
                 completionItem.setLabel(moduleName + ":" + label);
                 completionItems.add(lsCompletionItem);
+                if (completionItem.getFilterText() != null) {
+                    completionItem.setFilterText(moduleName + "_" + completionItem.getFilterText());
```
KavinduZoysa, line 312: Can't we set this filter text(i.e: filter text with module name) inside `getCompletionItemList` function(L:299)?

mindula:
> IIUC, this change does not change the sorting order. It rather fixes the invalid sorting caused when there is a filter text available and the filtertexts of some completion items are incorrect, right? If so, can you mention the order of the completion items(kinds)

Yes this fix doesn't specifically change the sorting order, but due to the change in the filter text there is some sort of a change in the sorting. You can refer to https://github.com/ballerina-platform/ballerina-lang/issues/40617#issuecomment-1649889052.
The order of the completion items is as follows,
1. Types defined in the same module
2. Types visible from other modules (StrandData, Thread in lang.value)
3. Visible Modules
4. Constant
5. Enums
6. Enum Member
7. Basic Types (boolean, int, string, etc.)
8. Type Descriptor snippets (record snippet, object snippet)
9. keywords (true, false, record, object)

language-server/modules/langserver-core/src/main/java/org/ballerinalang/langserver/completions/providers/AbstractCompletionProvider.java
mindula: No we can't because we only want to set the filter text for the types of external modules.

KavinduZoysa (approved):

malinthar (approved):

PR #41173, opened 2023-08-08
Introduce new `isEnum()` and `getEnumSymbol()` APIs in `UnionTypeSymbol` to distinguish and get enum-symbol in union-type descriptors

### LINKED ISSUE #38179: [Bug]: Unable to distinguish enum symbol from union-type symbol in Semantic API
### Description

Consider following code

```ballerina
public enum Baz {
    # This is open
    OPEN,
    # This is close
    CLOSE
}

public type State Baz;
```

`TypeReferenceTypeSymbol typeReferenceTypeSymbol = <State>;` // I have set the State as typereferenceTypeSymbol
calling `typeReferenceTypeSymbol.typeDescriptor()` returns `UnionTypeSymbol` but it should return `EnumSymbol`

### Steps to Reproduce

_No response_

### Affected Version(s)

_No response_

_No response_

### Related area

-> Compilation

_No response_

_No response_

_No response_

dulajdilshan:
## Purpose
$title

Fixes #38179

## Check List 
- [ ] Updated Change Log
- [x] Added necessary tests
  - [ ] Unit Tests
  - [ ] Spec Conformance Tests
  - [ ] Integration Tests
- [x] Increased Test Coverage   
  - [ ] API documentation 
  - [ ] Ballerina By Examples


compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaUnionTypeSymbol.java
```diff
@@ -147,6 +154,32 @@ public String signature() {
...
+
+        BTypeSymbol tsymbol = this.getBType().tsymbol;
+        if (tsymbol.getKind() != SymbolKind.ENUM) {
+            return Optional.empty();
+        }
+
+        SymbolFactory symbolFactory = SymbolFactory.getInstance(this.context);
+        this.enumSymbol = symbolFactory.createEnumSymbol((BEnumSymbol) tsymbol, tsymbol.getName().value);;
+
+        return Optional.ofNullable(this.enumSymbol);
```
LakshanWeerasinghe, line 180:
```suggestion
        return Optional.of(this.enumSymbol);
```
At this point can the enumSymbol be null?

dulajdilshan: Thanks for noticing!

compiler/ballerina-lang/src/main/java/io/ballerina/compiler/api/impl/symbols/BallerinaUnionTypeSymbol.java
```diff
@@ -147,6 +154,32 @@ public String signature() {
...
+    public boolean isEnum() {
+        if (this.enumSymbol != null) {
+            return true;
+        }
+
+        return this.getBType().tsymbol.getKind() == SymbolKind.ENUM;
+    }
+
+    @Override
+    public Optional<EnumSymbol> getEnumSymbol() {
```
MaryamZi, line 167: I'm not sure this is ideal. isEnum kinda makes sense, but why do we need to allow retrieving the enum symbol like this? In a way, it's like having a reference to the definition from the type symbol.

dulajdilshan: Yes. It's kinda exposing the information related to the enum symbol. Currently we may not have such things, but  having this may allow us to expose those infromation right?

dulajdilshan: and this is not a way of referencing. For referencing we have a structure as TypeRefType

MaryamZi: I meant having `enumSymbol` in `BallerinaUnionTypeSymbol`. This isn't quite right, IMO. The type doesn't need to how it was defined.



> Yes. It's kinda exposing the information related to the enum symbol. Currently, we may not have such things, but having this may allow us to expose that information right?

I don't think that's the correct way. If someone wants enum symbol they can still get it from the model, right? I don't think the type symbol should expose it.

dulajdilshan: >I meant having enumSymbol in BallerinaUnionTypeSymbol. This isn't quite right, IMO. The type doesn't need to how it was defined.
yeah. What if we have introduced some attributes only related to enums and the ballerina extension developer needs to get them? We can't expose that only-enum information via the BallerinaUnionTypeSymbol right?

> I don't think that's the correct way. If someone wants enum symbol they can still get it from the model, right? I don't think the type symbol should expose it.

How can they get an enum symbol without the location?

MaryamZi: > > I meant having enumSymbol in BallerinaUnionTypeSymbol. This isn't quite right, IMO. The type doesn't need to how it was defined.
> 
> yeah. What if we have introduced some attributes only related to enums and the ballerina extension developer needs to get them? We can't expose that only-enum information via the BallerinaUnionTypeSymbol right?

Can you give an example for this?

> 
> > I don't think that's the correct way. If someone wants enum symbol they can still get it from the model, right? I don't think the type symbol should expose it.
> 
> How can they get an enum symbol without the location?

Can't they filter out from module symbols by name?

MaryamZi: > We can't expose that only-enum information via the BallerinaUnionTypeSymbol right?

Yes, we can't. I can't think of such info anyway. We shouldn't mix up enums and types. An enum is just another way to define a union of string singletons. From a type symbol point of view, they are basically the same.

dulajdilshan: > Can you give an example for this?

what about annotation attachments? We shouldn't link them with the UnionTypeSymbol right?

> Can't they filter out module symbols by name?

Yeah, we can, but we need the name to filter out right? What if we need to get the enum that is in another module, and we have referenced an enum value (e.x: OPEN) and we need to know the other sibling values of the enum (which is referenced in new type?

```ballerina
//  defined in another module `test0.module1`'
public enum State {
    OPEN,
    CLOSED
}

// main.bal
import test0.module1;

type ConnectionState State;

function test() {
    string x = module1.OPEN;
}
```
Let's say there is a ballerina extension writer who needs to get only the sibling values of `OPEN`.
 
Also, isn't it costly to use the filter when there are so many symbols?

> Yes, we can't. I can't think of such info anyway. We shouldn't mix up enums and types. An enum is just another way to define a union of string singletons. From a type symbol point of view, they are basically the same.

Yes. This is true, but here we are just showing the relationship with the associated enum defined somewhere

MaryamZi: > what about annotation attachments? We shouldn't link them with the UnionTypeSymbol right?

```ballerina
annotation annot on type;

const RED = "RED";
const BLUE = "BLUE";

@annot
type Colour RED|BLUE;
```

is the same as

```ballerina
annotation annot on type;

@annot
enum Colour {
    RED,
    BLUE
}
```

so we shouldn't have to special case anything for enums.



> Yeah, we can, but we need the name to filter out right? 

If you have the union type you have the name?

> What if we need to get the enum that is in another module, and we have referenced an enum value (e.x: OPEN) and we need to know the other sibling values of the enum (which is referenced in new type?

How does adding the enum symbol to the union type help with this? The union type corresponds to the enum, not the member?

> Also, isn't it costly to use the filter when there are so many symbols?

It's just filtering based on a string? Anyway, the API being well-defined is more important IMO.

dulajdilshan:
Closing due to this: https://github.com/ballerina-platform/ballerina-lang/issues/38179#issuecomment-1705990925

PR #41184, opened 2023-08-09
Bump `mime4j` and `woodstox` transitive dependency versions 

rdulmina:
## Purpose
Axiom was using mime4j `0.7.2` which has a vulnerability, hence constrained to use `0.8.9`. `org.codehaus.woodstox:woodstox-core-asl` is also used by axiom, hence excluded the transitive dependency since we are using `fasterxml:woodstox-core`.


build.gradle
```diff
@@ -69,3 +69,7 @@ dependencies {
         //implementation 'org.springframework:spring-web:5.0.2.RELEASE'
     }
 }
+
+allprojects {
```
keizer619, line 73: We can remove this right?

keizer619 (approved):

build.gradle
rdulmina: Resolving this as discussed offline

PR #41231, opened 2023-08-17
Add create function CA in anonymous functions

### LINKED ISSUE #38744: [Improvement]: Add missing create function code action in serveral contexts
### Description

case1: 

create_function

Here we should get the create function code action to create the undefined function.

```Ballerina
       string[] results = []
       int[] res = results.'map((result) => getIntValue(result));

```

Case 2
create_function2

```Ballerina
type A record {|
    int field1;
|};

type B record {|
    int field2;
|};

public function main() {
    A[] values = [];

    B[] result = from A item in values
        select getB(item);
}

```
### Describe your problem(s)

_No response_

### Describe your solution(s)

_No response_

### Related area

-> Editor

_No response_

_No response_

_No response_

KavinduZoysa:
## Purpose

Fixes #38744


mindula:
If we consider the scenario mentioned in the issue, we get the `Change variable type` code action and when we apply it, we change the type to `$CompilationError$[]`. Will you be able to look into it? In those kind of scenarios, i think we should not suggest the code action.

https://github.com/ballerina-platform/ballerina-lang/assets/61020198/8004b9bb-20ca-496a-8724-f902e9e3eb85

KavinduZoysa:
> 

@mindula, I don't think we need to put in extra effort to disallow this. After we make the following change[1] from the front end side, those should be working as expected.
[1]: https://github.com/ballerina-platform/ballerina-lang/pull/41231#issuecomment-1695039394

malinthar (approved):
LGTM!

